PR #1263, opened 2023-09-19
Paywalls: Add product_name variable support

tonidero:
### Description
This PR adds support for the `product_name` variable in paywalls


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -97,7 +97,7 @@ class VariableProcessorTest {
 
     @Test
     fun `process variables processes product_name`() {
-        expectVariablesResult("{{ product_name }}", "PRODUCT_NAME")
+        expectVariablesResult("{{ product_name }}", "Annual")
```
NachoSoto, line 100: Just a single test here is good (same as `VariablesTest`).
I added a secondary `PackageVariablesTests` to check more specific details based on each package type.
For this I tested all types in iOS:
```swift
func testProductName() {
    expect(TestData.weeklyPackage.productName) == "Weekly"
    expect(TestData.monthlyPackage.productName) == "Monthly"
    expect(TestData.threeMonthPackage.productName) == "3 months"
    expect(TestData.sixMonthPackage.productName) == "6 months"
    expect(TestData.annualPackage.productName) == "Annual"
    expect(TestData.lifetimePackage.productName) == "Lifetime"
}
```

Which was useful then to also verify localization.

We can add that later but wanted to bring it up so it's on the radar :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -24,7 +24,7 @@ internal class VariableDataProvider(
     }
 
     fun productName(rcPackage: Package): String {
-        return "PRODUCT_NAME"
+        return rcPackage.product.title
```
NachoSoto, line 27: Is this localized?

NachoSoto (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
tonidero: The localization here comes from Google. Similar to iOS I think. So depending on the Google account country, Google will return us a different title. I don't think we can do much ourselves here.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
tonidero: Right, we need to add more test data with the different package types. Right now we only have annual, monthly and weekly. But we can do this 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
NachoSoto: Yeah perfect 👍🏻  that's what I was wondering.

PR #1264, opened 2023-09-19
Paywalls: Add support for `sub_period` variable

tonidero:
### Description
This PR adds support for the `sub_period` variable. In order to do that, some refactors were made to make sure we could test the result.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
@@ -1,6 +1,8 @@
 package com.revenuecat.purchases.ui.revenuecatui.data.processed
 
 import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.PackageType
+import com.revenuecat.purchases.ui.revenuecatui.R
 import com.revenuecat.purchases.ui.revenuecatui.helpers.ApplicationContext
 import java.util.Locale
 
@@ -28,7 +30,17 @@ internal class VariableDataProvider(
     }
 
     fun periodName(rcPackage: Package): String {
-        return "PERIOD_NAME"
+        val stringId = when (rcPackage.packageType) {
+            PackageType.LIFETIME -> R.string.lifetime
+            PackageType.ANNUAL -> R.string.annual
+            PackageType.SIX_MONTH -> R.string.semester
+            PackageType.THREE_MONTH -> R.string.quarter
+            PackageType.TWO_MONTH -> R.string.bimonthly
+            PackageType.MONTHLY -> R.string.monthly
+            PackageType.WEEKLY -> R.string.weekly
+            PackageType.UNKNOWN, PackageType.CUSTOM -> null
+        }
+        return stringId?.let { applicationContext.getString(it) } ?: ""
     }
 
     fun subscriptionDuration(rcPackage: Package): String? {
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactoryTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactoryTest.kt
@@ -3,9 +3,10 @@ package com.revenuecat.purchases.ui.revenuecatui.data.processed
 import android.net.Uri
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.PackageType
 import com.revenuecat.purchases.ui.revenuecatui.PaywallViewMode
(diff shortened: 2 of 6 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -28,7 +30,17 @@ internal class VariableDataProvider(
...
+            PackageType.LIFETIME -> R.string.lifetime
+            PackageType.ANNUAL -> R.string.annual
+            PackageType.SIX_MONTH -> R.string.semester
+            PackageType.THREE_MONTH -> R.string.quarter
+            PackageType.TWO_MONTH -> R.string.bimonthly
+            PackageType.MONTHLY -> R.string.monthly
+            PackageType.WEEKLY -> R.string.weekly
+            PackageType.UNKNOWN, PackageType.CUSTOM -> null
+        }
+        return stringId?.let { applicationContext.getString(it) } ?: ""
```
tonidero, line 43: I'm defaulting to an empty string, which I think is the same we do in iOS.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactoryTest.kt
```diff
@@ -87,14 +81,20 @@ internal class TemplateConfigurationFactoryTest {
 
     private fun getPackageInfo(rcPackage: Package): TemplateConfiguration.PackageInfo {
         val localizedConfiguration = TestData.template2.configForLocale(Locale.US)!!
+        val periodName = when(rcPackage.packageType) {
+            PackageType.ANNUAL -> "Annual"
+            PackageType.MONTHLY -> "Monthly"
+            PackageType.WEEKLY -> "Weekly"
+            else -> error("Unknown package type ${rcPackage.packageType}")
```
tonidero, line 88: I didn't like copying this here... But I thought it was better to kind of duplicate this logic than use the real code during the tests.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -101,8 +95,20 @@ class VariableProcessorTest {
...
+    @Test
+    fun `process variables processes sub_period monthly`() {
+        rcPackage = TestData.Packages.monthly
+        expectVariablesResult("{{ sub_period }}", "Monthly")
+    }
+
+    @Test
+    fun `process variables processes custom period`() {
+        rcPackage = TestData.Packages.annual.copy(packageType = PackageType.CUSTOM)
+        expectVariablesResult("{{ sub_period }}", "")
```
tonidero, line 111: Just want to confirm this is what we expect @NachoSoto

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
NachoSoto: 👍🏻

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
NachoSoto: Yeah 👍🏻  I don't have a test for this in iOS but this is fine.
The idea is to have validations in the frontend / backend to prevent users from using these with custom pacakges.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -13,45 +15,131 @@ import java.util.Locale
 @RunWith(AndroidJUnit4::class)
 class VariableProcessorTest {
 
+    private val usLocale = Locale.US
+    private val esLocale = Locale("es", "ES")
```
NachoSoto, line 19: Cool 👍🏻 

Don't have to do it now,  but I guess we're close to adding Spanish strings and testing them? :)

NachoSoto (approved):

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
tonidero: I mean, we could potentially run instrumentation tests (that run on an actual device), but I'm not sure it's worth it for this since we need to run them on firebase test lab and are much slower.

PR #1266, opened 2023-09-20
Paywalls: Adds some initial spanish strings and some initial preview

tonidero:
### Description
This PR adds the spanish version of the existing hardcoded strings and adds a preview for spanish locale

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
@@ -100,6 +100,14 @@ internal object TestData {
                 offerDetailsWithIntroOffer = "{{ total_price_and_per_month }} after {{ sub_offer_duration }} trial",
                 offerName = "{{ sub_period }}",
             ),
+            "es_ES" to PaywallData.LocalizedConfiguration(
+                title = "Título en español",
+                subtitle = "Un lorem ipsum en español que es más largo para mostrar un subtítulo multilinea.",
+                callToAction = "Suscribete for {{ price_per_period }}",
+                offerDetails = "{{ total_price_and_per_month }}",
+                offerDetailsWithIntroOffer = "{{ total_price_and_per_month }} con {{ sub_offer_duration }} de prueba",
+                offerName = "{{ sub_period }}",
+            ),
         ),
         assetBaseURL = Constants.assetBaseURL,
     )
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
@@ -19,8 +19,7 @@ internal object TemplateConfigurationFactory {
         val paywallTemplate = PaywallTemplate.fromId(paywallData.templateName)
             ?: throw IllegalArgumentException("Unknown template ${paywallData.templateName}")
 
-        val localizedConfiguration = paywallData.configForLocale(locale)
-            ?: error("No configuration found for locale $locale")
+        val localizedConfiguration = paywallData.localizedConfiguration
         val packageIds = paywallData.config.packages
         require(packageIds.isNotEmpty()) { "No packages ids found in paywall data" }
         require(packages.isNotEmpty()) { "No packages found in offering" }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
@@ -139,7 +139,8 @@ private fun SelectPackageButton(
     }
 }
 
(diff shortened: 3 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
```diff
@@ -19,8 +19,7 @@ internal object TemplateConfigurationFactory {
         val paywallTemplate = PaywallTemplate.fromId(paywallData.templateName)
             ?: throw IllegalArgumentException("Unknown template ${paywallData.templateName}")
 
-        val localizedConfiguration = paywallData.configForLocale(locale)
-            ?: error("No configuration found for locale $locale")
+        val localizedConfiguration = paywallData.localizedConfiguration
```
tonidero, line 22: This uses the application context locale. Which seems to work on device, but doesn't work on the Android studio preview. To fix that, we need to use the activity context to get the locale and calculate it on the fly. This is rearchitecturing things a bit so will work on this on a different PR.

NachoSoto: It's weird that this is ignoring the `locale` parameter though, which means that we could end up with inconsistent values here and when calling `ProcessedLocalizedConfiguration` inside of `PackageConfigurationFactory`, right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
```diff
@@ -166,7 +166,8 @@ private fun SelectPackageButton(
 private val PaywallViewState.Loaded.selectedLocalization: ProcessedLocalizedConfiguration
     get() = selectedPackage.localization
 
-@Preview(showBackground = true)
+@Preview(showBackground = true, locale = "en-rUS")
+@Preview(showBackground = true, locale = "es-rES")
```
NachoSoto, line 170: This is nice 👍🏻 

What's the `r` supposed to be for?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
```diff
@@ -59,6 +59,14 @@ internal object TestData {
                 offerDetailsWithIntroOffer = "{{ total_price_and_per_month }} after {{ sub_offer_duration }} trial",
                 offerName = "{{ sub_period }}",
             ),
+            "es_ES" to PaywallData.LocalizedConfiguration(
+                title = "Título en español",
+                subtitle = "Un lorem ipsum en español que es más largo para mostrar un subtítulo multilinea.",
+                callToAction = "Suscribete for {{ price_per_period }}",
```
NachoSoto, line 65: 😇 
```suggestion
                callToAction = "Suscríbete for {{ price_per_period }}",
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
tonidero: It's for `region`. From the docs, that's the way if you want to specify region. We could also just specify the language, but this seemed more realistic

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
tonidero: Yeah, sorry I posted this comment before discovering the issue was with how we get the localized configuration, which is fixed in #1267 . So please ignore this

NachoSoto: Makes sense. So you can revert this?

tonidero: I can, though Im changing it in #1267 again to this, since we perform some calculations there to use the list of locales there. So I think we can keep it as it is now. Lmk if you would prefer me to revert this to minify the changes

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
tonidero: Sorry, didn't post the docs link: https://developer.android.com/jetpack/compose/tooling/previews#preview-locale

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
NachoSoto: Okay that's fine then 👍🏻

vegaro (approved):

PR #1267, opened 2023-09-20
Fix locale selection logic

tonidero:
### Description
This fixes the locale selection logic to work with Android studio preview and make it more robust.

image


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -41,52 +43,60 @@ data class PaywallData(
...
-            return localization.values.first()
+            localization.entries.first().let { localization ->
+                return (getLocaleForLabel(localization.key) to localization.value)
+            }
         }
 
+    @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
+    private fun getLocaleForLabel(label: String): Locale {
+        // Parsing language requires it to be in the form "en-US", even though Locale.toString() returns "en_US"
+        return Locale.forLanguageTag(label.replace("_", "-"))
```
tonidero, line 78: I also had to do this to make it work as expected. The normal constructor of locale, has 2 parameters for language and country. If we want to parse a language tag, that didn't work that well in the previews.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -41,52 +43,60 @@ data class PaywallData(
...
      * @note This allows searching by `Locale` with only language code and missing region (like `en`, `es`, etc).
      *
      * @return [LocalizedConfiguration] for the given [Locale], if found.
      */
+    @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
     fun configForLocale(requiredLocale: Locale): LocalizedConfiguration? {
         return localization[requiredLocale.toString()]
             ?: localization.entries.firstOrNull { (localeKey, _) ->
-                @Suppress("UsePropertyAccessSyntax")
-                Locale(localeKey).languageWithFallback == requiredLocale.languageWithFallback
```
tonidero, line 73: This was wrong, since it could potentially end up comparing `isO3Language` with `language`

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -41,52 +43,60 @@ data class PaywallData(
     /**
      * Returns the [LocalizedConfiguration] to be used based on the current locale.
      */
-    val localizedConfiguration: LocalizedConfiguration
+    val localizedConfiguration: Pair<Locale, LocalizedConfiguration>
```
tonidero, line 46: I changed this to also return the locale selected. This is a bit weird as a public API, but I think we should use this instead of the default locale in case the default locale doesn't match the one we use for the strings. This way we can make sure to use the same locale everywhere.

NachoSoto: Makes sense. I'd update the docstring above though.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -41,52 +43,60 @@ data class PaywallData(
...
 
             for (i in 0 until preferredLocales.size()) {
                 preferredLocales.get(i)?.let { locale ->
-                    configForLocale(locale)?.let { localizedConfiguration ->
-                        return localizedConfiguration
+                    // This is a hack to make android studio previews work. In previews, the locale language method
+                    // gives a different result than expected (for example, "es-es" instead of "es").
+                    // In order to fix that, we convert the string to a locale and we parse that as a locale so the
+                    // language and country are correctly parsed.
+                    val localeToCheck = Locale.forLanguageTag(locale.toString())
```
NachoSoto, line 57: I would extract this to a `Locale` extension and add tests specifically for it.
This is solving a very specific problem (looking up a locale with `es-es`, for example), and putting the logic here seems like the wrong place for it.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
NachoSoto, line 78: Ditto here. What do you think about moving it to `LocaleExtensions` and adding tests for it?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -41,52 +43,60 @@ data class PaywallData(
...
             ?: localization.entries.firstOrNull { (localeKey, _) ->
-                @Suppress("UsePropertyAccessSyntax")
-                Locale(localeKey).languageWithFallback == requiredLocale.languageWithFallback
+                val locale = getLocaleForLabel(localeKey)
+                try {
+                    locale.isO3Language == requiredLocale.isO3Language
+                } catch (e: MissingResourceException) {
+                    errorLog("Locale $this can't obtain ISO3 language code ($e). Falling back to language.")
+                    locale.language == requiredLocale.language
+                }
```
NachoSoto, line 96: This is similar to the iOS version now:
```
private extension Locale {

    func sharesLanguageCode(with other: Locale) -> Bool {
        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
            return self.language.languageCode == other.language.languageCode
        } else {
            return self.languageCode == other.languageCode
        }
    }

}
```

I'd do the same and make it a method on `Locale` though, something like `Locale.sharesLanguageCodeWith(other: Locale)`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -18,7 +17,6 @@ internal fun Offering.toPaywallViewState(variableDataProvider: VariableDataProvi
             paywallData = paywallData,
             packages = availablePackages,
             activelySubscribedProductIdentifiers = emptySet(), // TODO-PAYWALLS: Check for active subscriptions
-            locale = Locale.getDefault(), // TODO-PAYWALLS: Use the correct locale
```
NachoSoto, line 21: Awesome

NachoSoto (commented):
Looks great 👍🏻 just some suggestions to make these "hacks" more explicit as Locale extensions and adding tests for them.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
```diff
@@ -14,12 +13,11 @@ internal object TemplateConfigurationFactory {
         paywallData: PaywallData,
         packages: List<Package>,
         activelySubscribedProductIdentifiers: Set<String>,
-        locale: Locale,
```
NachoSoto, line 17: I see you removed this 👍🏻

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
tonidero: This is tricky... This is part of the `purchases` module, so we would need to add the extensions there. However, `forLanguageTag` uses `API 21+`. Having said that, I think it's ok to fallback to the original locale in this case, since this is mostly done to fix previews.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/1267/commits/d01bb5546931c21d7bf4a0bd26d48a44270031ee

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/1267/commits/d01bb5546931c21d7bf4a0bd26d48a44270031ee

purchases/src/main/kotlin/com/revenuecat/purchases/utils/LocaleExtensions.kt
```diff
@@ -0,0 +1,31 @@
...
+// This is a hack to make android studio previews work. In previews, the locale language method
+// gives a different result than expected (for example, "es-es" instead of "es").
+// In order to fix that, we convert the string to a locale and we parse that as a locale so the
+// language and country are correctly parsed.
+@RequiresApi(Build.VERSION_CODES.LOLLIPOP)
+internal fun Locale.convertToCorrectlyFormattedLocale(): Locale {
+    return toString().toLocale()
+}
+
+@RequiresApi(Build.VERSION_CODES.LOLLIPOP)
```
vegaro, line 18: We have a util `Locale.toBCP47()` in `utils.kt` that we could use to make this also work pre-lollipop

vegaro: Nevermind, we have a min sdk of 24 in the UI module, which is where this is going to be used

tonidero: Oh yeah, I didn't know we had that. In any case, here we are converting a string to a locale, not the other way around, so we can't use that I think?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
vegaro: right, good catch

purchases/src/test/java/com/revenuecat/purchases/paywalls/PaywallDataTest.kt
```diff
@@ -77,8 +59,6 @@ class PaywallDataTest {
             assertThat(darkColors.accent2?.stringRepresentation).isEqualTo("#FF00FF")
         }
 
-        mockLocale("gl_ES", "glg")
```
vegaro, line 80: Oh so we don't need to mock this anymore because we catch in `sharedLanguageCodeWith` and use language to compare right?

vegaro (approved):

purchases/src/test/java/com/revenuecat/purchases/paywalls/PaywallDataTest.kt
tonidero: Well, actually I believe the issues you were running into are mostly fixed by using the `Locale.forLanguageTag` constructor. Before, we were doing `Locale(localeKey)` where `localeKey` was a tag but a language code is expected there. After using that, the tests were fixed without having to mock.

tonidero: You're right! Fixed in https://github.com/RevenueCat/purchases-android/pull/1267/commits/eeef67697e675bce67a0d96dbcea451c86e6175a

purchases/src/test/java/com/revenuecat/purchases/utils/LocaleExtensionsTest.kt
```diff
@@ -0,0 +1,59 @@
+package com.revenuecat.purchases.utils
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.Locale
+
+@RunWith(AndroidJUnit4::class)
+class LocaleExtensionsTest {
```
NachoSoto, line 10: ❤️

PR #1269, opened 2023-09-20
Paywalls: Implement `price_per_period` variable support

tonidero:
### Description
This PR adds support for the `price_per_period` variable in the paywalls. The period is calculated using a `MeasureFormat` which is only available in Android API 24+ so we've increased the min sdk version for now.

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
@@ -81,7 +81,12 @@ class VariableProcessorTest {
 
     @Test
     fun `process variables processes price_per_period`() {
-        expectVariablesResult("{{ price_per_period }}", "PRICE_PER_PERIOD")
+        expectVariablesResult("{{ price_per_period }}", "$67.99/yr")
+    }
+
+    @Test
+    fun `process variables processes price_per_period localized in spanish`() {
+        expectVariablesResult("{{ price_per_period }}", "$67.99/a", esLocale)
     }
 
     @Test
--- a/examples/paywall-tester/build.gradle
+++ b/examples/paywall-tester/build.gradle
@@ -10,7 +10,7 @@ android {
 
     defaultConfig {
         applicationId "com.revenuecat.paywall_tester"
-        minSdk 21
+        minSdk 24
         versionCode paywallTesterVersionCode as Integer
         versionName paywallTesterVersionName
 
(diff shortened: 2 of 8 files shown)


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -81,7 +81,12 @@ class VariableProcessorTest {
 
     @Test
     fun `process variables processes price_per_period`() {
-        expectVariablesResult("{{ price_per_period }}", "PRICE_PER_PERIOD")
+        expectVariablesResult("{{ price_per_period }}", "$67.99/yr")
+    }
+
+    @Test
+    fun `process variables processes price_per_period localized in spanish`() {
+        expectVariablesResult("{{ price_per_period }}", "$67.99/a", esLocale)
```
tonidero, line 89: Honestly, I don't like this abbreviation that much 🤔. IMO it may be even more confusing. I'm wondering whether we should use the full word, or just have our own localization for this... On the other hand, this is what the android formatter does, so might be ok...

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -81,7 +81,12 @@ class VariableProcessorTest {
 
     @Test
     fun `process variables processes price_per_period`() {
-        expectVariablesResult("{{ price_per_period }}", "PRICE_PER_PERIOD")
+        expectVariablesResult("{{ price_per_period }}", "$67.99/yr")
```
tonidero, line 84: We should add tests for other period units. Will add those for this and other tests in a followup PR.

NachoSoto (approved):

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
NachoSoto: Have you checked the other `FormatWidth`s? If you look at iOS `Localization.abbreviatedUnitLocalizedString` I have some heuristics to get an abbreviation that matches an expected length based on `unitAbbreviationLengthPriorities`.

You can iterate on this in a follow-up PR, but that's how I managed to get desirable abbreviations in different languages.

tonidero: I did check and in spanish at least it went from `1 a.` to `1 año`. There was nothing in between... But yeah, we can improve these as you did in iOS later.

PR #1270, opened 2023-09-20
Created enum with variable names

vegaro:
Extracted variable names to an enum

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
@@ -31,22 +31,22 @@ internal object VariableProcessor {
         rcPackage: Package,
         locale: Locale,
     ): String? {
-        return when (variableName) {
-            VariableName.APP_NAME.identifier -> variableDataProvider.applicationName
-            VariableName.PRICE.identifier -> variableDataProvider.localizedPrice(rcPackage)
-            VariableName.PRICE_PER_PERIOD.identifier -> variableDataProvider.localizedPricePerPeriod(rcPackage)
-            VariableName.TOTAL_PRICE_AND_PER_MONTH.identifier -> variableDataProvider.localizedPriceAndPerMonth(
+        return when (VariableName.valueOfIdentifier(variableName)) {
+            VariableName.APP_NAME -> variableDataProvider.applicationName
+            VariableName.PRICE -> variableDataProvider.localizedPrice(rcPackage)
+            VariableName.PRICE_PER_PERIOD -> variableDataProvider.localizedPricePerPeriod(rcPackage)
+            VariableName.TOTAL_PRICE_AND_PER_MONTH -> variableDataProvider.localizedPriceAndPerMonth(
                 rcPackage,
             )
-            VariableName.PRODUCT_NAME.identifier -> variableDataProvider.productName(rcPackage)
-            VariableName.SUB_PERIOD.identifier -> variableDataProvider.periodName(rcPackage)
-            VariableName.SUB_PRICE_PER_MONTH.identifier -> variableDataProvider.localizedPricePerMonth(
+            VariableName.PRODUCT_NAME -> variableDataProvider.productName(rcPackage)
+            VariableName.SUB_PERIOD -> variableDataProvider.periodName(rcPackage)
+            VariableName.SUB_PRICE_PER_MONTH -> variableDataProvider.localizedPricePerMonth(
                 rcPackage,
                 locale,
             )
-            VariableName.SUB_DURATION.identifier -> variableDataProvider.subscriptionDuration(rcPackage)
-            VariableName.SUB_OFFER_DURATION.identifier -> variableDataProvider.introductoryOfferDuration(rcPackage)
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
```diff
@@ -32,20 +32,38 @@ internal object VariableProcessor {
...
+            VariableName.SUB_OFFER_DURATION.identifier -> variableDataProvider.introductoryOfferDuration(rcPackage)
+            VariableName.SUB_OFFER_PRICE.identifier -> variableDataProvider.localizedIntroductoryOfferPrice(rcPackage)
             else -> {
                 Logger.e("Unknown variable: $variableName")
                 null
             }
         }
     }
+
+    enum class VariableName(val identifier: String) {
```
NachoSoto, line 57: <3

Can this be `private`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
```diff
@@ -32,20 +32,38 @@ internal object VariableProcessor {
...
+        PRICE("price"),
+        PRICE_PER_PERIOD("price_per_period"),
+        TOTAL_PRICE_AND_PER_MONTH("total_price_and_per_month"),
+        PRODUCT_NAME("product_name"),
+        SUB_PERIOD("sub_period"),
+        SUB_PRICE_PER_MONTH("sub_price_per_month"),
+        SUB_DURATION("sub_duration"),
+        SUB_OFFER_DURATION("sub_offer_duration"),
+        SUB_OFFER_PRICE("sub_offer_price"),
+    }
```
tonidero, line 68: We could create a static method `valueOf` or something like that that gives you the appropriate enum value. That way we can have a more exhaustive `when` above

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
vegaro: hmmm I believe so yes, I am using it for validation but so far only from this class

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
vegaro: Actually, I can just use Enum's `valueOf` right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
```diff
@@ -31,21 +31,46 @@ internal object VariableProcessor {
...
+        SUB_PERIOD("sub_period"),
+        SUB_PRICE_PER_MONTH("sub_price_per_month"),
+        SUB_DURATION("sub_duration"),
+        SUB_OFFER_DURATION("sub_offer_duration"),
+        SUB_OFFER_PRICE("sub_offer_price"),
+        ;
+        
+        companion object {
+            fun valueOfIdentifier(identifier: String): VariableName? {
+                return values().find { it.identifier == identifier }
```
NachoSoto, line 72: Sorry I think I commented on the commit directly.

Is there a reason this has to be linear?
In Swift I would build a `Map<String, VariableName>` to avoid this, seems like an easy win?

vegaro: how would you ensure the map is updated if new values are added to the enum?

tonidero: I guess we could build the map with the `values` method once. Small win I guess.

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
NachoSoto: > how would you ensure the map is updated if new values are added to the enum?

By constructing it automatically:
https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/PackageType.swift#L105-L108

NachoSoto: I think it's worth it. If you have maybe on average 6 variables in your strings, that's iterating the list of variables 6 times for no reason.

aboedo: +1

vegaro: Sure. I didn't think initially to do it with an associateBy and build the map dynamically. It makes sense!

PR #1272, opened 2023-09-20
`Paywalls`: added transition when loading images

NachoSoto:
This makes async loading a bit smoother and consistent with iOS.


aboedo (approved):

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -18,7 +20,10 @@ internal fun RemoteImage(
     contentDescription: String? = null,
 ) {
     AsyncImage(
-        model = urlString,
+        model = ImageRequest.Builder(LocalContext.current)
+            .data(urlString)
+            .crossfade(durationMillis = UIConstant.defaultAnimationDurationMillis)
+            .build(),
```
tonidero, line 26: I wonder if we should add a placeholder... but this is fine for now

PR #1274, opened 2023-09-20
`Paywalls`: new `PaywallIcon`

NachoSoto:
Screenshot 2023-09-20 at 13 53 59


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
```diff
@@ -0,0 +1,169 @@
...
+        painter = painterResource(id = icon.drawable()),
+        contentDescription = null,
+        tint = tintColor,
+    )
+}
+
+/**
+ * An icon that can be displayed inside a RevenueCat paywall.
+ */
+enum class PaywallIconName(val value: String) {
```
tonidero, line 39: Should this be `internal`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
```diff
@@ -0,0 +1,169 @@
...
+    SHARE("share"),
+    SMARTPHONE("smartphone"),
+    STACKED_BAR("stacked_bar"),
+    STARS("stars"),
+    SUBTRACT("subtract"),
+    TICK("tick"),
+    TRANSFER("transfer"),
+    TWO_WAY_ARROWS("two_way_arrows"),
+    KEY("key"),
+    WARNING("warning"),
```
tonidero, line 85: I believe the strings always match the enum but uppercased right? If so, we could just remove the string value and in the `fromValue`, we can do `PaywallIconName.valueOf(value.uppercased())` I think. Or create a map from the lowercased strings to the enums as well, to avoid linear lookup

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
```diff
@@ -0,0 +1,169 @@
...
+            TRANSFER -> R.drawable.transfer
+            TWO_WAY_ARROWS -> R.drawable.two_way_arrows
+            KEY -> R.drawable.vpn_key
+        }
+    }
+}
+
+@Preview(showBackground = true, widthDp = 300)
+@Composable
+internal fun PaywallIconPreview() {
```
tonidero, line 150: 🎉

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
```diff
@@ -0,0 +1,169 @@
...
+internal fun PaywallIcon(
+    icon: PaywallIconName,
+    modifier: Modifier = Modifier,
+    tintColor: Color,
+) {
+    Icon(
+        modifier = Modifier
+            .aspectRatio(1.0f)
+            .fillMaxSize()
+            .then(modifier),
```
tonidero, line 29: So this would apply the modifiers in the `modifier` parameter AFTER the aspectRation/fillMaxSize modifiers. That might be ok depending on the usage. Most times, when there is a `modifier` parameter, I've seen it used instead of the `Modifier` name, so this would look like:
```
        modifier = modifier
            .aspectRatio(1.0f)
            .fillMaxSize(),
```
But again, this might be correct if we need to apply the modifiers in the parameters last.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
```diff
@@ -0,0 +1,169 @@
...
+            red = Random.nextInt(0, 256),
+            green = Random.nextInt(0, 256),
+            blue = Random.nextInt(0, 256),
+        )
+    }
+
+    LazyVerticalGrid(columns = GridCells.Adaptive(40.dp)) {
+        items(icons.size) {
+            Box(modifier = Modifier.background(randomColor())) {
+                PaywallIcon(icon = icons[it], tintColor = Color.Black)
```
tonidero, line 165: I guess there is a chance the background color is also black? I guess it's probably not a big deal though.

tonidero (approved):
Left some comments but nothing blocking. I would suggest removing the strings for the enum if it's possible though, to avoid repetition there.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
NachoSoto: We need to expose this for https://linear.app/revenuecat/issue/PWL-194/icons /cc @vegaro
So we can iterate through the icons inside of `features` and verify that they're all valid.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
NachoSoto: Let me check 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
NachoSoto: Sorry I read your first comment as "private", I keep forgetting that Kotlin's default is `public` and not `internal` like in Swift.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallIcon.kt
NachoSoto: Done!

PR #1275, opened 2023-09-21
Add `awaitRestore` to customEntitlementComputation library

tonidero:
### Description
We moved the restore purchases functionality to the customEntitlementComputation library in #1211 but we didn't make the coroutines version available as well. This fixes that.


purchases/src/test/java/com/revenuecat/purchases/PurchasesCoroutinesCommonTest.kt
```diff
@@ -229,4 +231,113 @@ internal class PurchasesCoroutinesCommonTest : BasePurchasesTest() {
     }
 
     // endregion
+
+    // region awaitRestore
```
tonidero, line 235: This is just moving the tests. No changes.

vegaro (approved):

PR #1281, opened 2023-09-21
`Paywalls`: `InternalPaywallView` now takes a `PaywallViewMode` parameter

NachoSoto:
It was hardcoded to `FULL_SCREEN`, but we'll need to be able to override it.


tonidero (approved):
Yup, thanks!

PR #1261, opened 2023-09-19
Creates default PaywallData

vegaro:
Creates a default PaywallData in case data doesn't pass validation

PR #1273, opened 2023-09-20
PaywallData validation

vegaro:
Add validation to paywall data and creates default template if validation fails

PR #1282, opened 2023-09-21
`Paywalls`: created `LoadingPaywallView`

NachoSoto:
This uses `PaywallData.default` (from #1261) to be able to display a loading view just like in iOS.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
@@ -9,7 +9,6 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.lifecycle.ViewModel
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.PackageType
@@ -141,7 +140,7 @@ private object LoadingPaywallConstants {
 
 private class LoadingViewModel(
     state: PaywallViewState,
-) : ViewModel(), PaywallViewModel {
+) : PaywallViewModel {
     override val state: StateFlow<PaywallViewState>
         get() = _state.asStateFlow()
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
@@ -1,0 +1,147 @@
+package com.revenuecat.purchases.ui.revenuecatui
+
+import android.content.Context
+import androidx.compose.foundation.layout.Box
+import androidx.compose.material3.ColorScheme
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalContext
+import androidx.lifecycle.ViewModel
+import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.PackageType
+import com.revenuecat.purchases.models.Period
+import com.revenuecat.purchases.models.Price
+import com.revenuecat.purchases.paywalls.PaywallData
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewModel
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewState
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.PaywallTemplate
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.TemplateConfiguration
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.VariableDataProvider
+import com.revenuecat.purchases.ui.revenuecatui.data.testdata.MockApplicationContext
+import com.revenuecat.purchases.ui.revenuecatui.data.testdata.TestStoreProduct
+import com.revenuecat.purchases.ui.revenuecatui.extensions.createDefault
+import com.revenuecat.purchases.ui.revenuecatui.helpers.toAndroidContext
+import com.revenuecat.purchases.ui.revenuecatui.helpers.toPaywallViewState
+import com.revenuecat.purchases.ui.revenuecatui.templates.Template2
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import java.net.URL
+
+@Composable
+internal fun LoadingPaywallView(mode: PaywallViewMode) {
+    val paywallData: PaywallData = PaywallData.createDefault(
+        LoadingPaywallConstants.packages,
+        MaterialTheme.colorScheme
+    )
+
+    val offering = Offering(
+        identifier = LoadingPaywallConstants.offeringIdentifier,
+        serverDescription = "Loading paywall",
+        metadata = emptyMap(),
+        availablePackages = LoadingPaywallConstants.packages,
+        paywall = paywallData
+    )
+
+    val viewModel = LoadingViewModel(mode, offering, paywallData)
+
+    val state = offering.toPaywallViewState(
+        VariableDataProvider(LocalContext.current.applicationContext.toAndroidContext()),
+        mode,
+        paywallData,
+        LoadingPaywallConstants.template
+    )
+
+    when (state) {
+        is PaywallViewState.Error,
+        is PaywallViewState.Loading -> Box {}
+        is PaywallViewState.Loaded -> Template2(
+            state = state,
+            viewModel = viewModel
+        )
+    }
+}
+
+private object LoadingPaywallConstants {
+    const val offeringIdentifier = "loading_offering"
+
+    val template = PaywallTemplate.TEMPLATE_2
+
+    val packages = listOf(
+        Package(
+            identifier = "weekly",
+            packageType = PackageType.WEEKLY,
+            offering = this.offeringIdentifier,
+            product = TestStoreProduct(
+                id = "com.revenuecat.weekly",
+                title = "Weekly",
+                price = Price(formatted = "$1.99", currencyCode = "USD", amountMicros = 1_990_000),
+                description = "Weekly",
+                period = Period(value = 1, unit = Period.Unit.WEEK, iso8601 = "P1W"),
+            ),
+        ),
+        Package(
+            identifier = "monthly",
+            packageType = PackageType.MONTHLY,
+            offering = this.offeringIdentifier,
+            product = TestStoreProduct(
+                id = "com.revenuecat.monthly",
+                title = "Monthly",
+                price = Price(formatted = "$5.99", currencyCode = "USD", amountMicros = 5_990_000),
+                description = "Monthly",
+                period = Period(value = 1, unit = Period.Unit.MONTH, iso8601 = "P1M"),
+            ),
+        ),
+        Package(
+            identifier = "annual",
+            packageType = PackageType.ANNUAL,
+            offering = this.offeringIdentifier,
+            product = TestStoreProduct(
+                id = "com.revenuecat.annual",
+                title = "Annual",
+                price = Price(formatted = "$15.99", currencyCode = "USD", amountMicros = 5_990_000),
+                description = "Annual",
+                period = Period(value = 1, unit = Period.Unit.YEAR, iso8601 = "P1Y"),
+            ),
+        )
+    )
+}
+
+private class LoadingViewModel(
+    mode: PaywallViewMode = PaywallViewMode.default,
+    offering: Offering,
+    paywall: PaywallData
+) : ViewModel(), PaywallViewModel {
+    override val state: StateFlow<PaywallViewState>
+        get() = _state.asStateFlow()
+
+    override fun refreshStateIfLocaleChanged() = Unit
+    override fun refreshStateIfColorsChanged(colorScheme: ColorScheme) = Unit
+
+    private val _state =
+        MutableStateFlow(
+            offering.toPaywallViewState(
+                VariableDataProvider(MockApplicationContext()),
+                mode,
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.helpers.toPaywallViewState
+import com.revenuecat.purchases.ui.revenuecatui.templates.Template2
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import java.net.URL
+
+@Composable
+internal fun LoadingPaywallView(mode: PaywallViewMode) {
+    val provider = VariableDataProvider(LocalContext.current.applicationContext.toAndroidContext())
```
NachoSoto, line 29: This is duplicated so I'd like to avoid it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,120 @@
...
+    val offering = LoadingPaywallConstants.offering
+
+    Template2(
+        state = offering.toPaywallViewState(
+            provider,
+            LoadingPaywallConstants.paywallData!!,
```
NachoSoto, line 35: We won't need `!!` once I actually define the data below.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,120 @@
...
+        ),
+        viewModel = LoadingViewModel(mode, offering)
+    )
+}
+
+private object LoadingPaywallConstants {
+    private const val offeringIdentifier = "loading_offering"
+
+    // TODO: use .createDefault()
+    val paywallData: PaywallData? = null
```
NachoSoto, line 46: Just need to wait for #1261 to be able to use it :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.helpers.toAndroidContext
+import com.revenuecat.purchases.ui.revenuecatui.helpers.toPaywallViewState
+import com.revenuecat.purchases.ui.revenuecatui.templates.Template2
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import java.net.URL
+
+@Composable
+internal fun LoadingPaywallView(mode: PaywallViewMode) {
```
NachoSoto, line 28: This whole implementation is pretty much the same as in iOS, creating fake data and reusing the rest of the code! :D

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
```diff
@@ -27,7 +28,7 @@ import java.net.URL
 
 internal object TestData {
     val template1 = PaywallData(
-        templateName = "1", // TODO-PAYWALLS: use enum
+        templateName = PaywallTemplate.TEMPLATE_1.id,
```
NachoSoto, line 31: Using these now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
```diff
@@ -283,7 +284,7 @@ private object Constants {
     val assetBaseURL = URL("https://assets.pawwalls.com")
 }
 
-private data class TestStoreProduct(
+internal data class TestStoreProduct(
```
NachoSoto, line 287: We might want to move this to the main SDK for everyone to use? It's been super helpful in iOS.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -14,19 +14,25 @@ internal fun Offering.toPaywallViewState(
...
-    }
+
+    return toPaywallViewState(variableDataProvider, paywallData, mode)
+}
+
+internal fun Offering.toPaywallViewState(
+    variableDataProvider: VariableDataProvider,
+    paywallData: PaywallData,
+    mode: PaywallViewMode,
+): PaywallViewState.Loaded {
```
NachoSoto, line 25: @vegaro this is what I was telling you about :) thanks to your refactor in #1273 (I'll wait for that to be finished before merging this) we can create a `PaywallViewState.Loaded` here since we can guarantee we have everything ready (useful for the default / loading paywalls).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -14,19 +14,25 @@ internal fun Offering.toPaywallViewState(
...
-            mode = mode,
-            paywallData = paywallData,
-            packages = availablePackages,
-            activelySubscribedProductIdentifiers = emptySet(), // TODO-PAYWALLS: Check for active subscriptions
-        )
-        PaywallViewState.Loaded(
-            templateConfiguration = templateConfiguration,
-            selectedPackage = templateConfiguration.packages.default,
-        )
-    } catch (e: Exception) {
```
NachoSoto, line 29: I removed this to compile for now. #1273 is also getting rid of it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,120 @@
...
+    val provider = VariableDataProvider(LocalContext.current.applicationContext.toAndroidContext())
+    val offering = LoadingPaywallConstants.offering
+
+    Template2(
```
NachoSoto, line 32: An upcoming PR will actually make this look like a "placeholder" like in iOS.

NachoSoto:
@RevenueCat/cashnip this is ready for review!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,148 @@
...
+            offering.toPaywallViewState(
+                VariableDataProvider(MockApplicationContext()),
+                mode,
+                template = LoadingPaywallConstants.template,
+                validatedPaywallData = paywall,
+            ),
+        )
+
+    override fun selectPackage(packageToSelect: TemplateConfiguration.PackageInfo) {
+        error("Not supported")
```
tonidero, line 134: Hmm what would happen if people tapped on these buttons while they are loading? Would it crash? I think we probably want to do nothing on these.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    val offering = Offering(
+        identifier = LoadingPaywallConstants.offeringIdentifier,
+        serverDescription = "Loading paywall",
+        metadata = emptyMap(),
+        availablePackages = LoadingPaywallConstants.packages,
+        paywall = paywallData,
+    )
+
+    val viewModel = LoadingViewModel(mode, offering, paywallData)
```
tonidero, line 47: Classes that inherit from `ViewModel` should not be instantiated using the constructor. Instead you could use the `viewModel` function to get an instance. However, you would need to create a factory to pass this data to the VM, similar to how we create the main view model currently

Alternatively, we could just make this VM not inherit from `ViewModel`. The main advantage of using that is that you get a data container that survives configuration changes automatically... But I don't think that's that important for this. If we see performance issues at some point, we can reconsider.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,148 @@
...
+    override val state: StateFlow<PaywallViewState>
+        get() = _state.asStateFlow()
+
+    override fun refreshStateIfLocaleChanged() = Unit
+    override fun refreshStateIfColorsChanged(colorScheme: ColorScheme) = Unit
+
+    private val _state =
+        MutableStateFlow(
+            offering.toPaywallViewState(
+                VariableDataProvider(MockApplicationContext()),
```
tonidero, line 126: Hmm I don't think we should use `MockApplicationContext` in production, even if it's not really going to be visible... Maybe we can pass the real application context to the view model?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    val template = PaywallTemplate.TEMPLATE_2
+
+    val packages = listOf(
+        Package(
+            identifier = "weekly",
+            packageType = PackageType.WEEKLY,
+            offering = this.offeringIdentifier,
+            product = TestStoreProduct(
+                id = "com.revenuecat.weekly",
```
tonidero, line 78: I wonder how many devs have a weekly option... Just wondering whether we should just have 2 options for the loading screen, and remove the weekly one.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -0,0 +1,148 @@
...
+    const val offeringIdentifier = "loading_offering"
+
+    val template = PaywallTemplate.TEMPLATE_2
+
+    val packages = listOf(
+        Package(
+            identifier = "weekly",
+            packageType = PackageType.WEEKLY,
+            offering = this.offeringIdentifier,
+            product = TestStoreProduct(
```
tonidero, line 77: Hmm not a fan of using `TestStoreProduct` in production... But I guess we would have to do something similar in any case for the loading screen following this route...

I'm wondering whether we would like to have a completely separate loading screen which doesn't need actual data... But I guess that would mean duplicating a lot of the UI layout, so probably not be worth it. I think we can go with this for now.

tonidero (commented):
Left a few comments but looking good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
NachoSoto: TIL, thanks! I removed the `ViewModel` parent class.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
NachoSoto: Fixed.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
NachoSoto: Ultimately you won't see them. You think having 2 in the loading screen makes more sense than 3?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
NachoSoto: Yeah it's slightly weird, but are you worried about something in particulaar?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
tonidero: Yeah that's what I was thinking... But maybe we can leave the 3 for now, if it's what we do in iOS.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
tonidero: Well, mostly worried about using these test models in production... I think it's a bit of a smell TBH... But yeah, as long as we make sure it's limited to the `LoadingPaywallView`, it should be ok for now I think. We can reconsider about adding a separate layout for the loading view later if needed.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
NachoSoto: Yeah. Ultimately there's no right answer though.

tonidero (approved):
This looks good for now. I'm not completely sold on using `TestStoreProduct` in production, but I think we can iterate on that later if needed.

NachoSoto:
Thanks, yeah. I'm thinking we could have a separate `PackageConfiguration` that allows us to represent these fake packages instead.

PR #1283, opened 2023-09-21
Paywalls: Add support for `sub_offer_duration` variable in paywalls

tonidero:
### Description
Adds support for `sub_offer_duration` variable in paywalls. Makes sure we use the same intro offer than with `sub_offer_price` in case there are multiple (currently we only display one)

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
@@ -175,14 +175,14 @@ class VariableProcessorTest {
     fun `process variables processes sub_offer_duration`() {
         expectVariablesResult("{{ sub_offer_duration }}", "1 month")
         expectVariablesResult("{{ sub_offer_duration }}", "1 month", rcPackage = TestData.Packages.bimonthly)
-        expectVariablesResult("{{ sub_offer_duration }}", "1 month", rcPackage = TestData.Packages.quarterly)
+        expectVariablesResult("{{ sub_offer_duration }}", "2 weeks", rcPackage = TestData.Packages.quarterly)
     }
 
     @Test
     fun `process variables processes sub_offer_duration for spanish locale`() {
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -173,7 +173,21 @@ class VariableProcessorTest {
...
     fun `process variables processes sub_offer_duration`() {
-        expectVariablesResult("{{ sub_offer_duration }}", "INT_OFFER_DURATION")
+        expectVariablesResult("{{ sub_offer_duration }}", "1 month")
+        expectVariablesResult("{{ sub_offer_duration }}", "1 month", rcPackage = TestData.Packages.bimonthly)
+        expectVariablesResult("{{ sub_offer_duration }}", "2 weeks", rcPackage = TestData.Packages.quarterly)
+    }
+
+    @Test
+    fun `process variables processes sub_offer_duration for spanish locale`() {
+        expectVariablesResult("{{ sub_offer_duration }}", "1 mes", esLocale)
```
NachoSoto, line 183: Relying on the default package is confusing here, maybe always pass an explicit value?

vegaro (approved):

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
vegaro: I had to look up the default value yes, maybe making it more explicit is better

tonidero: I added the default value in the tests where we test a bunch of packages in the followup PR: https://github.com/RevenueCat/purchases-android/pull/1285

I thought about making the parameter not optional, but then we have to specify the package in a bunch of extra tests which seemed like a lot of repetition.

tonidero:
Going to merge this since I addressed the comments in the followup PR.

PR #1284, opened 2023-09-21
`Paywalls`: implemented `LoadingPaywallView` with `placeholder`

NachoSoto:
This uses https://github.com/google/accompanist/blob/main/placeholder/. It's deprecated but might come to Compose itself in the future. This is just a temporary loading screen, we can decide whether to keep it, implement it in another way, or remove it in the future.

image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
@@ -6,6 +6,7 @@ import androidx.compose.material3.ColorScheme
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.tooling.preview.Preview
@@ -15,6 +16,9 @@ import com.revenuecat.purchases.PackageType
 import com.revenuecat.purchases.models.Period
 import com.revenuecat.purchases.models.Price
 import com.revenuecat.purchases.paywalls.PaywallData
+import com.revenuecat.purchases.ui.revenuecatui.composables.Fade
+import com.revenuecat.purchases.ui.revenuecatui.composables.PlaceholderDefaults
+import com.revenuecat.purchases.ui.revenuecatui.composables.placeholder
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewModel
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewState
 import com.revenuecat.purchases.ui.revenuecatui.data.processed.PaywallTemplate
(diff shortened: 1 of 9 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
```diff
@@ -62,7 +69,11 @@ internal fun Template2(state: PaywallViewState.Loaded, viewModel: PaywallViewMod
 }
 
 @Composable
-private fun Template2MainContent(state: PaywallViewState.Loaded, viewModel: PaywallViewModel) {
+private fun Template2MainContent(
+    state: PaywallViewState.Loaded,
+    viewModel: PaywallViewModel,
+    childModifier: Modifier,
```
tonidero, line 75: I wonder if we should rename to `contentModifier` or `loadingContentModifier` so it's clearer.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
```diff
@@ -126,7 +140,7 @@ private fun SelectPackageButton(
     }
     val border = if (isSelected) null else BorderStroke(2.dp, colors.text1)
     Button(
-        modifier = Modifier.fillMaxWidth(),
+        modifier = Modifier.fillMaxWidth().then(childModifier),
```
tonidero, line 143: You could do 
```suggestion
        modifier = childModifier.fillMaxWidth(),
```
I think? We should see how it looks though.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
```diff
@@ -29,13 +34,23 @@ internal fun LoadingPaywallView(mode: PaywallViewMode) {
     val provider = VariableDataProvider(LocalContext.current.applicationContext.toAndroidContext())
     val offering = LoadingPaywallConstants.offering
 
+    val placeholder = Modifier.placeholder(
+        visible = true,
+        highlight = com.revenuecat.purchases.ui.revenuecatui.composables.Fade(
```
tonidero, line 39: We should try to use a normal import

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
NachoSoto: 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
NachoSoto: Yeah that's simpler 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywallView.kt
NachoSoto: Oopss thanks

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/placeholder.kt
```diff
@@ -0,0 +1,326 @@
...
+        infiniteRepeatable(
+            animation = tween(durationMillis = 1700, delayMillis = 200),
+            repeatMode = RepeatMode.Restart,
+        )
+    }
+}
+
+/**
+ * From https://github.com/google/accompanist/blob/main/placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
+ * See https://github.com/google/accompanist/pull/1672. That's deprecated, so we're copying it here for now.
```
tonidero, line 66: 👍

tonidero (approved):
I'd love to get this in if it works well. We should make sure it works well on different devices. But I think the code looks good for now.

PR #1287, opened 2023-09-22
Paywalls: Fix state update upon locale changes

tonidero:
### Description
I noticed the locale changes weren't working as expected. It changed the system driven locales but not our computed strings. This was because the code that was meant to update the state upon locale changes was wrong. We were using `remember` to store the last locales, but that doesn't survive configuration changes like the view model does. So I moved the logic to the view model.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
@@ -3,12 +3,7 @@ package com.revenuecat.purchases.ui.revenuecatui
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.LocalContext
-import androidx.core.os.LocaleListCompat
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewModel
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
```diff
@@ -47,15 +44,6 @@ internal fun InternalPaywallView(
     }
 }
 
-@Composable
-private fun updateStateIfLocaleChanged(viewModel: PaywallViewModel) {
-    var locale by remember { mutableStateOf(LocaleListCompat.getDefault()) }
```
tonidero, line 52: As mentioned in the description, this doesn't survive configuration changes. So we were always getting the latest locales upon a locale change.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -55,19 +56,24 @@ internal class PaywallViewModelImpl(
     override val state: StateFlow<PaywallViewState>
         get() = _state.asStateFlow()
     private val _state: MutableStateFlow<PaywallViewState>
+    private val _lastLocaleList = MutableStateFlow(LocaleListCompat.getDefault())
```
NachoSoto, line 59: `LocaleListCompat.getDefault()` is repeated 3 times in this file, maybe extract it?

NachoSoto (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -55,19 +56,24 @@ internal class PaywallViewModelImpl(
...
 
     init {
+        Logger.e("TEST: INITIALIZING PAYWALL VIEW MODEL")
```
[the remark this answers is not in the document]
tonidero: Nope 😓 Thanks!

PR #1209, opened 2023-08-24
Support for Billing Client 6

MarkVillacampa:
# Bumping Billing client to 6.0.0
- Deprecates...
  - ProrationMode -> ReplacementMode
  - GoogleProrationMode -> GoogleReplacementMode
- Temporarily disables DEFERRED product change until backend support is ready.

PR #1289, opened 2023-09-25
Made some classes and functions internal

vegaro:
In preparation for BC6 support, I've modified the visibility of some classes, constants and functions to make them internal. Most of them are in the amazon package (they used to be public to be able to call them from other modules)

Note this is based off #1209

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/logUtils.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/logUtils.kt
@@ -30,7 +30,6 @@ internal fun infoLog(message: String) {
     logIfEnabled(LogLevel.INFO, currentLogHandler::i, message)
 }
 
-// Public because it's been used in the hybrids
 internal fun warnLog(message: String) {
     logIfEnabled(LogLevel.WARN, currentLogHandler::w, message)
 }


purchases/src/main/kotlin/com/revenuecat/purchases/common/logUtils.kt
```diff
@@ -31,7 +31,7 @@ internal fun infoLog(message: String) {
 }
 
 // Public because it's been used in the hybrids
-fun warnLog(message: String) {
+internal fun warnLog(message: String) {
```
vegaro, line 34: I think we can make this one internal even if it's been used in the hybrids

MarkVillacampa: We're still using this in PHC, will it work after setting it to internal?

https://github.com/RevenueCat/purchases-hybrid-common/blob/db9cb9090cbe2959408c32d2ee122fc10f1af621/android/src/main/java/com/revenuecat/purchases/hybridcommon/common.kt#L381

vegaro: No... it's only two logs though... maybe we can do something so we can hide this in this SDK 🤔

tonidero: Well, if we make it internal, we can still log things in PHC/Hybrids... The consequence of that however is that those logs won't be redirected through the log handler in case the developer has set any. If we want to bypass that, we should wrap the log handler in PHC to set it in Android but also to redirect the messages in PHC. But honestly, if those logs are not very important, I would be ok if they are not redirected for now.

vegaro: Yes, I agree, I'd rather make this internal so we don't expose this to devs for no reason

vegaro:
Let's keep this open until the base branch is ready to go

MarkVillacampa:
I've cherry-picked the commit in this PR into the #bc6-support-new branch because, after force-pushing that branch, there were a bunch of conflicts here.

PR #1290, opened 2023-09-25
in app messages support

aboedo:
This PR provides support for in app messages for declined payments with BillingClient. They can be shown using `Purchases.showDeclinedPaymentMessageIfNeeded`. It also provides an API in `PurchasesConfiguration` to display those messages automatically.

image

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/strings/BillingStrings.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/strings/BillingStrings.kt
@@ -23,8 +23,8 @@ internal object BillingStrings {
     const val ILLEGAL_STATE_EXCEPTION_WHEN_CONNECTING = "There was an IllegalStateException when connecting to " +
         "BillingClient. This has been reported to occur on Samsung devices on unknown circumstances.\nException: %s"
     const val BILLING_CONNECTION_ERROR_INAPP_MESSAGES = "Error connecting to billing client to display " +
-        "inapp messages: %s"
+        "in-app messages: %s"
     const val BILLING_INAPP_MESSAGE_NONE = "No Google Play in-app message was available."
-    const val BILLING_INAPP_MESSAGE_UPDATE = "Subscription status was updated from In-App Message."
+    const val BILLING_INAPP_MESSAGE_UPDATE = "Subscription status was updated from in-app message."
     const val BILLING_INAPP_MESSAGE_UNEXPECTED_CODE = "Unexpected billing code: %s"
 }


purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -130,6 +130,17 @@ class Purchases internal constructor(
         purchasesOrchestrator.removeUpdatedCustomerInfoListener()
     }
 
+    /**
+     * Google Play only, no-op for Amazon.
+     * If the user has had a payment declined, this will show a toast notification notifying them and
+     * providing instructions for recovery of the subscription.
+     *
+     * https://developer.android.com/google/play/billing/subscriptions#in-app-messaging
```
aboedo, line 138: we should replace this with a rev.cat link that we own

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -57,6 +62,17 @@ open class PurchasesConfiguration(builder: Builder) {
...
         }
 
+        /**
+         * Enable this setting to show a toast with recovery options for users who have had a declined payment
+         * automatically. https://developer.android.com/google/play/billing/subscriptions#in-app-messaging
+         *
+         * If this setting is disabled, you can show the toast by calling
+         * [Purchases.showDeclinedPaymentMessageIfNeeded]
+         */
+        fun showDeclinedPaymentMessagesAutomatically(showDeclinedPaymentMessagesAutomatically: Boolean) = apply {
```
aboedo, line 72: not married to the naming

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -753,6 +756,25 @@ internal class BillingWrapper(
 
     override fun isConnected(): Boolean = billingClient?.isReady ?: false
 
+    override fun showInAppMessagesIfNeeded(activity: Activity) {
```
aboedo, line 759: getting the activity might actually be tricky here

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -753,6 +756,25 @@ internal class BillingWrapper(
...
+        billingClient?.showInAppMessages(activity,
+            inAppMessageParams
+        ) { inAppMessageResult ->
+            if (inAppMessageResult.responseCode == InAppMessageResult.InAppMessageResponseCode.NO_ACTION_NEEDED) {
+                // No in-app message was available.
+                // TODO: log
+            } else if (inAppMessageResult.responseCode
+                == InAppMessageResult.InAppMessageResponseCode.SUBSCRIPTION_STATUS_UPDATED
+            ) {
+                // TODO: post receipt
```
aboedo, line 773: we'd need to do a syncPurchases

tonidero: Hmm do we need to? If I remember correctly, when a purchase recovers, it would still be the same token, so it's already sent and we shouldn't need to do anything right?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
tonidero: I've been thinking about it... From the documentation, this seems to be the only use case for inapp messages right now, but the naming in Google Play seems more generic, so I wonder if we should keep it generic as well and have it be `showInAppMesssagesAutomatically`... Will keep it like this for now since it's clearer what's going to be doing for now, but we might need to rename it in the future if Google adds more functionality.

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Done

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -753,6 +757,37 @@ internal class BillingWrapper(
 
     override fun isConnected(): Boolean = billingClient?.isReady ?: false
 
+    override fun showInAppMessagesIfNeeded(activity: Activity) {
+        val inAppMessageParams = InAppMessageParams.newBuilder()
+            .addInAppMessageCategoryToShow(InAppMessageParams.InAppMessageCategoryId.TRANSACTIONAL)
+            .build()
+        val weakActivity = WeakReference(activity)
```
tonidero, line 764: I'm using a weak reference here since I want to avoid any chance of leaking the activity if there is a problem connecting to the billing client.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -38,6 +40,9 @@ open class PurchasesConfiguration(builder: Builder) {
         @set:JvmSynthetic @get:JvmSynthetic
         internal var observerMode: Boolean = false
 
+        @set:JvmSynthetic @get:JvmSynthetic
+        internal var showDeclinedPaymentMessagesAutomatically: Boolean = false
```
tonidero, line 44: Should we make this default to `true` already?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: However, we probably do need to get the new customer info from the backend. I forgot that. Adding

tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/1290/commits/a8eb16f5b5c433520dac67a131e41e84ecf98fc1

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -57,6 +62,18 @@ open class PurchasesConfiguration(builder: Builder) {
             this.appUserID = appUserID
         }
 
+        /**
+         * Enable this setting to show a toast with recovery options for users who have had a declined payment
+         * automatically. Default is disabled.
+         * For more info: https://rev.cat/googleplayinappmessaging
+         *
+         * If this setting is disabled, you can show the toast by calling
```
vegaro, line 70: I think this is technically a snackbar not a toast, right?

purchases/src/main/kotlin/com/revenuecat/purchases/strings/BillingStrings.kt
```diff
@@ -22,4 +22,9 @@ internal object BillingStrings {
     const val BILLING_CLIENT_RETRY = "Retrying BillingClient connection after backoff of %s milliseconds."
     const val ILLEGAL_STATE_EXCEPTION_WHEN_CONNECTING = "There was an IllegalStateException when connecting to " +
         "BillingClient. This has been reported to occur on Samsung devices on unknown circumstances.\nException: %s"
+    const val BILLING_CONNECTION_ERROR_INAPP_MESSAGES = "Error connecting to billing client to display " +
+        "inapp messages: %s"
+    const val BILLING_INAPP_MESSAGE_NONE = "No Google Play in-app message was available."
+    const val BILLING_INAPP_MESSAGE_UPDATE = "Subscription status was updated from In-App Message."
```
vegaro, line 28: we are writing in-app in three different ways in this strings file `inapp`, `in-app`, `In-App`

vegaro (approved):

MarkVillacampa (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
tonidero: Yeah, will just make it default to true. We shouldn't release tomorrow's automatic release though with this.

tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/1290/commits/31570b6fee3bcb3498d213352c1c45ef5d83dac2

PR #1292, opened 2023-09-26
Add preview for default data

vegaro:
A bit hacky but this way we can see the preview of the default data

image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
@@ -3,6 +3,7 @@ package com.revenuecat.purchases.ui.revenuecatui
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ReadOnlyComposable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalInspectionMode
@@ -67,4 +68,5 @@ private fun getPaywallViewModel(
 }
 
 @Composable
+@ReadOnlyComposable
 private fun isInPreviewMode() = LocalInspectionMode.current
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
```diff
@@ -130,17 +133,25 @@ private fun AppIcon(
...
-        context.applicationInfo.loadIcon(packageManager)
-    }
 
-    Image(
-        bitmap = appIconResId.toBitmap(config = Bitmap.Config.ARGB_8888).asImageBitmap(),
-        contentDescription = null,
-        modifier = modifier,
-        contentScale = ContentScale.Crop,
-    )
+    if (LocalInspectionMode.current) {
```
tonidero, line 137: Should we reuse the `isInPreviewMode` composable? We might want to extract it to another file for that.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
```diff
@@ -60,6 +61,10 @@ private fun getPaywallViewModel(
             offering,
             listener,
             MaterialTheme.colorScheme,
+            preview = isInPreviewMode(),
         ),
     )
 }
+
+@Composable
```
tonidero, line 69: I think you can add a `@ReadOnlyComposable` annotation here for some small optimizations.

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/HelperFunctions.kt
```diff
@@ -0,0 +1,9 @@
+package com.revenuecat.purchases.ui.revenuecatui.helpers
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.ui.platform.LocalInspectionMode
+
+@Composable
+@ReadOnlyComposable
+internal fun isInPreviewMode() = LocalInspectionMode.current
```
NachoSoto, line 9: Can this be a lazy constant, so we don't evaluate it multiple times?

tonidero: Hmm I'm not sure you can declare a constant and mark it as `@Composable`, which is needed to use `LocalInspectionMode.current`. In any case, this composable shouldn't be recomputed unless the local composition value changes thanks to `@ReadOnlyComposable`.

NachoSoto: Oooh nice gotcha

PR #1293, opened 2023-09-27
Future-proof in-app message APIs in case new types are added

tonidero:
### Description
Followup to #1290.

This improves the API to future-proof them in case Google adds new in-app message types in the future. This is not a breaking change since the previous APIs haven't been released yet.


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -191,8 +192,8 @@ internal class PurchasesOrchestrator constructor(
     }
 
     override fun onActivityStarted(activity: Activity) {
-        if (appConfig.showDeclinedPaymentMessagesAutomatically) {
-            showDeclinedPaymentMessageIfNeeded(activity)
+        if (appConfig.showInAppMessagesAutomatically) {
+            showInAppMessagesIfNeeded(activity, InAppMessageType.values().toList())
```
tonidero, line 196: When showing messages automatically, we show all types

MarkVillacampa (approved):

PR #1294, opened 2023-09-27
Template 3

vegaro:
Preview with a lot of features
Screenshot 2023-09-27 at 19 42 37

Preview with three features
Screenshot 2023-09-27 at 19 42 47

Screenshot from emulator
Screenshot 2023-09-27 at 19 43 16

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewMode.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewMode.kt
@@ -9,11 +9,4 @@ internal enum class PaywallViewMode {
     companion object {
         val default = FULL_SCREEN
     }
-
-    fun shouldDisplayIcon(): Boolean {
-        return when (this) {
-            FULL_SCREEN -> true
-            FOOTER, FOOTER_CONDENSED -> false
-        }
-    }
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewMode.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewMode.kt
@@ -9,4 +9,11 @@ internal enum class PaywallViewMode {
     companion object {
         val default = FULL_SCREEN
     }
+
+    fun shouldDisplayIcon(): Boolean {
+        return when (this) {
+            FULL_SCREEN -> true
+            FOOTER, FOOTER_CONDENSED -> false
+        }
+    }
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
@@ -151,18 +151,23 @@ internal object TestData {
         ),
         localization = mapOf(
             "en_US" to PaywallData.LocalizedConfiguration(
-                title = "RevenueCat Paywalls automatically check for Introductory Offer eligibility",
+                title = "How your free trial works",
                 callToAction = "Continue",
                 features = listOf(
                     PaywallData.LocalizedConfiguration.Feature(
-                        content = "Content for Feature 1 in English",
-                        iconID = "attach_money",
-                        title = "Title for Feature 1 in English",
+                        title = "Today",
+                        content = "Full access to 1000+ workouts plus free meal plan worth {{ price }}.",
+                        iconID = "tick",
                     ),
                     PaywallData.LocalizedConfiguration.Feature(
-                        content = "Content for Feature 2 in English",
-                        iconID = "search",
-                        title = "Title for Feature 2 in English",
+                        title = "Day 7",
+                        content = "Get a reminder about when your trial is about to end.",
+                        iconID = "notification",
+                    ),
+                    PaywallData.LocalizedConfiguration.Feature(
+                        title = "Day 14",
+                        content = "You'll automatically get subscribed. Cancel anytime before if you didn't love our app.",
+                        iconID = "attachment",
                     ),
                 ),
                 offerDetails = "{{ total_price_and_per_month }}",
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
@@ -116,6 +116,51 @@ internal object TestData {
         ),
         assetBaseURL = Constants.assetBaseURL,
     )
+    val template3 = PaywallData(
+        templateName = "3", // TODO-PAYWALLS: use enum
+        config = PaywallData.Configuration(
+            packages = listOf(
+                PackageType.WEEKLY.identifier!!,
+                PackageType.MONTHLY.identifier!!,
+                PackageType.ANNUAL.identifier!!,
+            ),
+            defaultPackage = PackageType.MONTHLY.identifier!!,
+            images = Constants.images,
+            blurredBackgroundImage = true,
+            displayRestorePurchases = true,
+            termsOfServiceURL = URL("https://revenuecat.com/tos"),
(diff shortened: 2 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewMode.kt
```diff
@@ -9,4 +9,11 @@ internal enum class PaywallViewMode {
     companion object {
         val default = FULL_SCREEN
     }
+
+    fun shouldDisplayIcon(): Boolean {
```
tonidero, line 13: This seems unused?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
```diff
@@ -116,6 +116,115 @@ internal object TestData {
         ),
         assetBaseURL = Constants.assetBaseURL,
     )
+    val template3 = PaywallData(
+        templateName = "3", // TODO-PAYWALLS: use enum
```
tonidero, line 120: Since the original data is a string, it might be difficult to use the enum here? Though if we end up moving the enum to the main sdk, then that will be different

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
```diff
@@ -116,6 +116,115 @@ internal object TestData {
         ),
         assetBaseURL = Constants.assetBaseURL,
     )
+    val template3 = PaywallData(
```
tonidero, line 119: I wonder if we should move each template test data to its own file, to avoid this file getting huge with all templates

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -0,0 +1,199 @@
...
+}
+
+@Composable
+private fun Features(
+    features: List<PaywallData.LocalizedConfiguration.Feature>,
+    colors: TemplateConfiguration.Colors,
+) {
+    if (features.isEmpty()) return
+
+    Column(
```
tonidero, line 106: I wonder if we should use a `LazyColumn` here... For most cases, this is enough, but if someone ever added a ton of features, `LazyColumn` might make more sense... I think it's ok to keep this though.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -0,0 +1,199 @@
...
+        modifier = Modifier.verticalScroll(rememberScrollState())
```
tonidero, line 107: Hmm I see we have the scroll at the `Fetures` composable... Should this scroll be at the `Template3MainContent` level (or its container)? Just wondering if we change to landscape, how will this look...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -0,0 +1,199 @@
...
+    feature: PaywallData.LocalizedConfiguration.Feature,
+    colors: TemplateConfiguration.Colors,
+) {
+    Row(
+        verticalAlignment = Alignment.Top,
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(horizontal = UIConstant.defaultHorizontalPadding),
+    ) {
+        feature.iconID?.let { PaywallIconName.fromValue(it) }?.let { icon ->
```
tonidero, line 131: Hmm is there a chance of some features having icons and some not? Asking since in that case, the rows text might not be aligned. This might be an ok edge case though.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -0,0 +1,199 @@
...
+            Text(
+                style = MaterialTheme.typography.bodyLarge,
+                fontWeight = FontWeight.Bold,
+                textAlign = TextAlign.Start,
+                text = feature.title,
+                color = colors.text1,
+            )
+            feature.content?.let { content ->
+                Text(
+                    modifier = Modifier.padding(top = Template3UIConstants.iconPadding * 2),
```
tonidero, line 158: Slightly strange to use the icon padding for the text... But it's probably ok. We can polish all this later.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -0,0 +1,199 @@
...
+                    text = content,
+                    color = colors.text2,
+                )
+            }
+        }
+    }
+}
+
+@Composable
+private fun IconImage(iconUri: Uri?) {
```
tonidero, line 171: Seems like this is pretty similar to the one in Template2. Could it be extracted and shared?

tonidero (approved):
Left a few comments but nothing blocking!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/TestData.kt
vegaro: Yes, that's mainly why I was asking, because I tried using the enum here but realized we need to move the class to the other module

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
vegaro: The frontend always asks for an icon. Not sure if we need to take care of that, but maybe we should have a backup icon just in case?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
vegaro: Good idea!

vegaro: Done in [`57b973c` (#1294)](https://github.com/RevenueCat/purchases-android/pull/1294/commits/57b973ce5d7bb9c3e822968918e4b1f91f5f3e37)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
vegaro: In iOS it's at the Features level, only the features scroll.

I haven't tested the landscape, I will

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
tonidero: If it's validated in the frontend, then I think it's fine to leave as is.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -0,0 +1,174 @@
...
+                    style = MaterialTheme.typography.bodyMedium,
+                    fontWeight = FontWeight.Normal,
+                    textAlign = TextAlign.Start,
+                    text = content,
+                    color = colors.text2,
+                )
+            }
+        }
+    }
+}
```
NachoSoto, line 167: Just ~160 lines :D love it

NachoSoto (commented):
Looks awesome!

PR #1295, opened 2023-09-27
Better handling of packages not found for id error

vegaro:
We had a `require` in `createPackageConfiguration` that would throw if the selected package in the template configuration is not available. I changed the error handling with using `Result`

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.ui.revenuecatui.data.processed
 
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.paywalls.PaywallData
+import com.revenuecat.purchases.ui.revenuecatui.errors.PackageConfigurationError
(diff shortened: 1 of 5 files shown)


tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -25,7 +26,9 @@ internal object PackageConfigurationFactory {
             }
             rcPackage
         }
-        require(filteredRCPackages.isNotEmpty()) { "No packages found for ids $filter" }
+        if (filteredRCPackages.isEmpty()) {
+            return Result.failure(PackageConfigurationError("No packages found for ids $filter"))
```
tonidero, line 30: I do wonder if we should fallback to the default configuration in this case... But for now, I think it's ok to just display an error.

vegaro: That's a good point, but which package do we pick? The first one available?

tonidero: Hmm yeah... I guess the current if it exists and if not, fallback to the first... but I think this is also ok

PR #1297, opened 2023-09-28
Paywalls: Add API to display paywall as a composable dialog

tonidero:
### Description
This adds a new composable `PaywallDialog` that can be used to display the paywall as a dialog. It also modifies paywall tester to provide options on how we want to display the paywall from the offerings tab.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
@@ -56,6 +56,9 @@ private fun getPaywallViewModel(
 ): PaywallViewModel {
     val applicationContext = LocalContext.current.applicationContext
     return viewModel<PaywallViewModelImpl>(
+        // We need to pass the key in order to create different view models for different offerings when
+        // trying to load different paywalls for the same view model store owner.
+        key = offering?.identifier,
         factory = PaywallViewModelFactory(
             applicationContext.toAndroidContext(),
             mode,
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -22,6 +22,11 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
 import com.revenuecat.purchases.ui.revenuecatui.helpers.computeWindowWidthSizeClass
 import kotlinx.coroutines.launch
 
+/**
+ * Composable offering a dialog screen Paywall UI configured from the RevenueCat dashboard.
+ * This dialog will be shown as a full screen dialog in compact devices and a normal dialog othewise.
+ * @param paywallDialogOptions The options to configure the PaywallDialog and what to do on dismissal.
+ */
 @Composable
 fun PaywallDialog(
     paywallDialogOptions: PaywallDialogOptions,
@@ -43,7 +48,9 @@ fun PaywallDialog(
                 }
                 if (shouldDisplayDialog) {
                     Logger.d("Displaying paywall dialog according to display logic")
-                } else Logger.d("Not displaying paywall dialog according to display logic")
+                } else {
+                    Logger.d("Not displaying paywall dialog according to display logic")
+                }
             }
         }
     }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -1,0 +1,80 @@
+package com.revenuecat.purchases.ui.revenuecatui
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.Scaffold
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.window.Dialog
+import androidx.compose.ui.window.DialogProperties
+import androidx.window.core.layout.WindowWidthSizeClass
+import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.PurchasesException
+import com.revenuecat.purchases.awaitCustomerInfo
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+import com.revenuecat.purchases.ui.revenuecatui.helpers.computeWindowWidthSizeClass
+import kotlinx.coroutines.launch
+
+@Composable
+fun PaywallDialog(
+    paywallDialogOptions: PaywallDialogOptions,
+) {
+    val shouldDisplayBlock = paywallDialogOptions.shouldDisplayBlock
+    var shouldDisplayDialog by remember { mutableStateOf(shouldDisplayBlock == null) }
+    if (shouldDisplayBlock != null) {
+        LaunchedEffect(paywallDialogOptions) {
+            launch {
+                shouldDisplayDialog = try {
+                    // TODO-PAYWALLS: This won't receive updates in case the customer info changes and starts/stops
+                    // being eligible to display the paywall dialog. We would need to support multiple customer info
+                    // listeners to refresh this.
+                    val customerInfo = Purchases.sharedInstance.awaitCustomerInfo()
+                    shouldDisplayBlock.invoke(customerInfo)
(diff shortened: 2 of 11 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -0,0 +1,80 @@
...
+    val shouldDisplayBlock = paywallDialogOptions.shouldDisplayBlock
+    var shouldDisplayDialog by remember { mutableStateOf(shouldDisplayBlock == null) }
+    if (shouldDisplayBlock != null) {
+        LaunchedEffect(paywallDialogOptions) {
+            launch {
+                shouldDisplayDialog = try {
+                    // TODO-PAYWALLS: This won't receive updates in case the customer info changes and starts/stops
+                    // being eligible to display the paywall dialog. We would need to support multiple customer info
+                    // listeners to refresh this.
+                    val customerInfo = Purchases.sharedInstance.awaitCustomerInfo()
```
tonidero, line 38: I didn't fully like this here so I was thinking about moving it to the view model. But then, I didn't want to have to create the view model at this point (since we might not even display the dialog)... I thought it was ok for now to keep this. Lmk if you would prefer me to move it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -0,0 +1,80 @@
...
+                .padding(paddingValues),
+        ) {
+            PaywallView(paywallDialogOptions.toPaywallViewOptions())
+        }
+    }
+}
+
+@Composable
+@ReadOnlyComposable
+private fun shouldUsePlatformDefaultWidth(): Boolean {
```
tonidero, line 75: This basically will display a full screen dialog for phones and a normal dialog for tablets and foldables. I think it looks ok for now but we can revisit later and polish this logic.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewListener.kt
```diff
@@ -12,5 +12,5 @@ interface PaywallViewListener {
     fun onRestoreStarted() {}
     fun onRestoreCompleted(customerInfo: CustomerInfo) {}
     fun onRestoreError(error: PurchasesError) {}
-    fun onDismissed() {}
+    fun onCloseButtonPressed() {}
```
tonidero, line 15: I'm changing this listener method, since the `onDismissed` only made sense when displaying as a dialog and the dialog options will already have a separate callback for that. 
I'm making showing the close button optional (currently not implemented) and having a callback here for when it's tapped.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -51,7 +51,9 @@ internal fun Template1(state: PaywallViewState.Loaded, viewModel: PaywallViewMod
         modifier = Modifier.fillMaxSize(),
         horizontalAlignment = Alignment.CenterHorizontally,
     ) {
-        Template1MainContent(state)
+        Column(modifier = Modifier.weight(1f)) {
+            Template1MainContent(state)
+        }
```
tonidero, line 56: This fixes an issue that happened in some devices, specially in landscape where the purchase button and footer weren't visible. By adding a weight to the main content, it has less priority than the button/footer so those will always show.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template2.kt
```diff
@@ -63,9 +62,15 @@ internal fun Template2(state: PaywallViewState.Loaded, viewModel: PaywallViewMod
         Column(
             modifier = Modifier.fillMaxSize(),
         ) {
-            Spacer(modifier = Modifier.weight(1f))
-            Template2MainContent(state, viewModel)
-            Spacer(modifier = Modifier.weight(1f))
+            Box(
+                modifier = Modifier.weight(1f).padding(
```
tonidero, line 66: Similar to the above, but for template 2.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywallView.kt
```diff
@@ -55,6 +55,9 @@ private fun getPaywallViewModel(
 ): PaywallViewModel {
     val applicationContext = LocalContext.current.applicationContext
     return viewModel<PaywallViewModelImpl>(
+        // We need to pass the key in order to create different view models for different offerings when
+        // trying to load different paywalls for the same view model store owner.
+        key = offering?.identifier,
```
tonidero, line 60: This has an important drawback which is that, if a user displays multiple different offerings as dialogs, the viewmodels for all of them will remain in memory... I'm not sure if this is the best way to fix this, but it should work for now.

NachoSoto: I think that makes sense. It's also probably extremely uncommon.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -0,0 +1,80 @@
...
+) {
+    val shouldDisplayBlock = paywallDialogOptions.shouldDisplayBlock
+    var shouldDisplayDialog by remember { mutableStateOf(shouldDisplayBlock == null) }
+    if (shouldDisplayBlock != null) {
+        LaunchedEffect(paywallDialogOptions) {
+            launch {
+                shouldDisplayDialog = try {
+                    // TODO-PAYWALLS: This won't receive updates in case the customer info changes and starts/stops
+                    // being eligible to display the paywall dialog. We would need to support multiple customer info
+                    // listeners to refresh this.
```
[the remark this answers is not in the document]
NachoSoto: Yeah that's how it works in iOS. We just want to do one initial check. I'd remove the TODO, since we wouldn't want to get notified of customer info changes and hide the paywall.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
NachoSoto: This is fine. That's exactly how `PresentingPaywallModifier` works.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -0,0 +1,80 @@
...
+                    val customerInfo = Purchases.sharedInstance.awaitCustomerInfo()
+                    shouldDisplayBlock.invoke(customerInfo)
+                } catch (e: PurchasesException) {
+                    Logger.e("Error fetching customer info to display paywall dialog", e)
+                    false
+                }
+                if (shouldDisplayDialog) {
+                    Logger.d("Displaying paywall dialog according to display logic")
+                } else Logger.d("Not displaying paywall dialog according to display logic")
```
NachoSoto, line 46: Nit: weird that you have {} for the `if` but not the `else`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.window.core.layout.WindowWidthSizeClass
+import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.PurchasesException
+import com.revenuecat.purchases.awaitCustomerInfo
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+import com.revenuecat.purchases.ui.revenuecatui.helpers.computeWindowWidthSizeClass
+import kotlinx.coroutines.launch
+
+@Composable
+fun PaywallDialog(
```
NachoSoto, line 26: Can you add a docstring to this?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
NachoSoto: Would adding the weight to the outer colum work too?

NachoSoto (approved):
Looks great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
tonidero: No, if we apply the weight to the outer column, that means that it won't apply to the inner views. Weights work with siblings by allowing you to set a relative weight between them to use the available space. In case one or more of the siblings don't have a weight attribute, those will have priority over those that do have it, which is what I'm doing here (give priority to the button + footer over the main content to display on the screen)

Issue #1288, opened 2023-09-23
Performance issue in retryBillingServiceConnectionWithExponentialBackoff


rantianhua:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
It looks like a dead loop in retrying BillingClient connection.
```
2023-09-23 13:04:55.112 28721-29147 [Purchases] - DEBUG                        D  ℹ️ Saving tokens []
2023-09-23 13:04:55.113 28721-29147 [Purchases] - DEBUG                        D  ℹ️ Tokens already posted: []
2023-09-23 13:04:55.117 28721-29150 [Purchases] - DEBUG                        D  ℹ️ Cleaning previously sent tokens
2023-09-23 13:04:55.117 28721-29150 [Purchases] - DEBUG                        D  ℹ️ Tokens already posted: []
2023-09-23 13:04:55.118 28721-29150 [Purchases] - DEBUG                        D  ℹ️ Saving tokens []
2023-09-23 13:04:55.118 28721-29150 [Purchases] - DEBUG                        D  ℹ️ Tokens already posted: []
2023-09-23 13:04:55.119 28721-29150 [Purchases] - DEBUG                        D  ℹ️ No pending purchases to sync
2023-09-23 13:04:55.119 28721-29133 [Purchases] - DEBUG                        D  ℹ️ Cleaning previously sent tokens
2023-09-23 13:04:55.120 28721-29133 [Purchases] - DEBUG                        D  ℹ️ Tokens already posted: []
2023-09-23 13:04:55.120 28721-29133 [Purchases] - DEBUG                        D  ℹ️ Saving tokens []
2023-09-23 13:04:55.121 28721-29133 [Purchases] - DEBUG                        D  ℹ️ Tokens already posted: []
2023-09-23 13:04:55.121 28721-29133 [Purchases] - DEBUG                        D  ℹ️ No pending purchases to sync
2023-09-23 13:04:55.122 28721-29147 [Purchases] - DEBUG                        D  ℹ️ No pending purchases to sync
[...]

PR #1300, opened 2023-09-29
Optimize billing client connection retries

tonidero:
### Description
This is a possible approach to deal with #1288 
Seems like Google may be calling the `onBillingServiceDisconnected` callback multiple times consecutively. In those cases we are scheduling retries in the future for each of those calls which adds some overheard to that bug in Google's side. In order to minimize that overhead, this PR suggests having only a maximum of 1 retry scheduled, which should reduce the overhead a lot.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -93,6 +93,8 @@ internal class BillingWrapper(
     // how long before the data source tries to reconnect to Google play
     private var reconnectMilliseconds = RECONNECT_TIMER_START_MILLISECONDS
 
+    private var reconnectionAlreadyScheduled = false
+
     class ClientFactory(private val context: Context) {
         @UiThread
         fun buildClient(listener: com.android.billingclient.api.PurchasesUpdatedListener): BillingClient {
@@ -121,6 +123,8 @@ internal class BillingWrapper(
             if (billingClient == null) {
                 billingClient = clientFactory.buildClient(this)
             }
+
+            reconnectionAlreadyScheduled = false
 
             billingClient?.let {
                 if (!it.isReady) {
@@ -730,9 +734,7 @@ internal class BillingWrapper(
     }
 
     override fun onBillingServiceDisconnected() {
-        mainHandler.post {
-            log(LogIntent.DEBUG, BillingStrings.BILLING_SERVICE_DISCONNECTED.format(billingClient.toString()))
-        }
+        log(LogIntent.DEBUG, BillingStrings.BILLING_SERVICE_DISCONNECTED.format(billingClient?.toString()))
         retryBillingServiceConnectionWithExponentialBackoff()
     }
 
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -730,9 +734,7 @@ internal class BillingWrapper(
     }
 
     override fun onBillingServiceDisconnected() {
-        mainHandler.post {
-            log(LogIntent.DEBUG, BillingStrings.BILLING_SERVICE_DISCONNECTED.format(billingClient.toString()))
-        }
```
tonidero, line 735: I removed wrapping this log to the main thread... Seemed like unneeded overhead.

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -93,6 +93,8 @@ internal class BillingWrapper(
     // how long before the data source tries to reconnect to Google play
     private var reconnectMilliseconds = RECONNECT_TIMER_START_MILLISECONDS
 
+    private var reconnectionAlreadyScheduled = false
```
tonidero, line 96: I guess the ideal way of doing this would be to have a queue of jobs, and check if it's empty... But this seemed like a simpler solution for now. Lmk what you think.

tonidero:
Bump @RevenueCat/coresdk

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
vegaro: Yeah, not sure why this would be posted to the main thread

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -743,12 +745,17 @@ internal class BillingWrapper(
...
+            log(LogIntent.DEBUG, BillingStrings.BILLING_CLIENT_RETRY_ALREADY_SCHEDULED)
+        } else {
+            log(LogIntent.DEBUG, BillingStrings.BILLING_CLIENT_RETRY.format(reconnectMilliseconds))
+            reconnectionAlreadyScheduled = true
+            startConnectionOnMainThread(reconnectMilliseconds)
+            reconnectMilliseconds = min(
+                reconnectMilliseconds * 2,
+                RECONNECT_TIMER_MAX_TIME_MILLISECONDS,
+            )
+        }
```
vegaro, line 758: I am wondering if we should change these logs from debug to warning? That way they would show up in crash reporting tools logs?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -743,12 +745,17 @@ internal class BillingWrapper(
      * This prevents ANRs, see https://github.com/android/play-billing-samples/issues/310
      */
     private fun retryBillingServiceConnectionWithExponentialBackoff() {
```
vegaro, line 747: this is not really exponential anymore since it's only one retry?

vegaro (commented):
I think this makes sense to me!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Hmm, I guess it depends on how it's configured but yeah, considering it shouldn't happen often, I think that's a good idea.

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Well, it's exponential, but you can only keep on retry scheduled at the same time. So only after the currently scheduled retry finishes, will there be another retry with a longer backoff. In theory, we should only connect once, so I think it's ok?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/1300/commits/4bb42b6acf9400295b2dfe0ee25b1f83418db2dc

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
vegaro: Ahh that's an important detail, right. I misunderstood what this was doing, that's kinda why I asked lol. Makes sense yes.

vegaro (approved):

PR #1307, opened 2023-10-04
Refactor checking for available packages when creating package configuration

vegaro:
There shouldn't be any behavior change, but this refactor makes it more clear what happens when none of the configured packages are available.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
@@ -10,24 +10,26 @@ internal object PackageConfigurationFactory {
     @Suppress("LongParameterList")
     fun createPackageConfiguration(
         variableDataProvider: VariableDataProvider,
-        packages: List<Package>,
+        availablePackages: List<Package>,
         activelySubscribedProductIdentifiers: Set<String>,
-        filter: List<String>,
+        packageIdsInConfig: List<String>,
         default: String?,
         localization: PaywallData.LocalizedConfiguration,
         configurationType: PackageConfigurationType,
         locale: Locale,
     ): Result<TemplateConfiguration.PackageConfiguration> {
-        val packagesById = packages.associateBy { it.identifier }
-        val filteredRCPackages = filter.mapNotNull {
-            val rcPackage = packagesById[it]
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -10,24 +10,26 @@ internal object PackageConfigurationFactory {
...
             if (rcPackage == null) {
                 Logger.d("Package with id $it not found. Ignoring.")
             }
             rcPackage
-        }
+        }.takeUnless { it.isEmpty() } ?: availablePackages
+
         if (filteredRCPackages.isEmpty()) {
-            return Result.failure(PackageConfigurationError("No packages found for ids $filter"))
+            // This wont' happen because availablePackages won't be empty. Offerings can't have empty available packages
```
tonidero, line 31:
```suggestion
            // This won't happen because availablePackages won't be empty. Offerings can't have empty available packages
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
```diff
@@ -28,9 +25,9 @@ internal object TemplateConfigurationFactory {
         val createPackageResult =
             PackageConfigurationFactory.createPackageConfiguration(
                 variableDataProvider = variableDataProvider,
-                packages = packages,
+                availablePackages = availablePackages,
                 activelySubscribedProductIdentifiers = activelySubscribedProductIdentifiers,
-                filter = packageIds,
+                packageIdsInConfig = paywallData.config.packages,
```
tonidero, line 30: I wonder if we should rename the property `packages` to `packageIds`, even if the backend response remains the same... But that can be done separately if needed.

tonidero (approved):
Much clearer, thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactory.kt
vegaro: That will make things clearer. I will open a PR

PR #1308, opened 2023-10-04
Fix template test data

vegaro:
We were passing `PaywallTemplate` enum `name` instead of `id` when building `TestData`


tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/testdata/templates/Template1TestData.kt
```diff
@@ -9,7 +9,7 @@ import java.net.URL
 
 internal val TestData.template1: PaywallData
     get() = PaywallData(
-        templateName = PaywallTemplate.TEMPLATE_1.name,
+        templateName = PaywallTemplate.TEMPLATE_1.id,
```
NachoSoto, line 12: Yikes, this is so easy to get wrong, considering we call this template _name_. I guess a consequence of enums having an implicit "name" property.

PR #1311, opened 2023-10-04
`Paywalls`: use `IntroEligibilityStateView`

NachoSoto:
Also added the missing offer details to template 3:
Screenshot 2023-10-04 at 14 34 58


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -67,6 +69,17 @@ internal fun Template3(
         ) {
             Template3MainContent(state)
         }
+        Column(
+            modifier = Modifier.align(Alignment.CenterHorizontally),
```
tonidero, line 73: Hmm could we apply this alignment to the parent column? That way we avoid this column

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
NachoSoto: That doesn't work because I'd have to use `ColumnScope`.

PR #1312, opened 2023-10-04
`Paywalls`: Markdown support

NachoSoto:
This uses https://github.com/commonmark/commonmark-java for parsing, and the rendering implementation is derived from https://github.com/ErikHellman/MarkdownComposer/blob/master/app/src/main/java/se/hellsoft/markdowncomposer/MarkdownComposer.kt with a lot changes to be able to use `Markdown()` in the same way we use `Text()`, therefore being able to replace our usages everywhere.

Screenshot 2023-10-05 at 12 27 37

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/IntroEligibilityStateView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/IntroEligibilityStateView.kt
@@ -1,6 +1,5 @@
 package com.revenuecat.purchases.ui.revenuecatui.composables
 
-import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.TextStyle
@@ -30,7 +29,7 @@ internal fun IntroEligibilityStateView(
         textWithNoIntroOffer ?: textWithIntroOffer ?: ""
     }
 
-    Text(
+    Markdown(
         text,
         color = color,
         style = style,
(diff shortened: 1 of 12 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/IntroEligibilityStateView.kt
```diff
@@ -30,7 +29,7 @@ internal fun IntroEligibilityStateView(
         textWithNoIntroOffer ?: textWithIntroOffer ?: ""
     }
 
-    Text(
+    Markdown(
```
NachoSoto, line 32: Example of how trivial it is to switch over.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/testdata/templates/Template1TestData.kt
```diff
@@ -38,11 +38,11 @@ internal val TestData.template1: PaywallData
         ),
         localization = mapOf(
             "en_US" to PaywallData.LocalizedConfiguration(
-                title = "Ignite your child's curiosity",
-                subtitle = "Get access to all our educational content trusted by thousands of parents.",
-                callToAction = "Subscribe for {{ sub_price_per_month }}",
-                callToActionWithIntroOffer = "Purchase for {{ sub_price_per_month }} per month",
-                offerDetails = "{{ sub_price_per_month }} per month",
+                title = "Ignite your _child_'s curiosity",
```
NachoSoto, line 41: Changing some of our data to test it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template3.kt
```diff
@@ -138,7 +138,8 @@ private fun Feature(
         verticalAlignment = Alignment.Top,
         modifier = Modifier
             .fillMaxWidth()
-            .padding(horizontal = UIConstant.defaultHorizontalPadding),
+            .padding(horizontal = UIConstant.defaultHorizontalPadding)
+            .padding(top = Template3UIConstants.iconPadding * 2),
```
NachoSoto, line 142: Moved the padding to the row instead.

vegaro (approved):

tonidero (approved):

PR #1313, opened 2023-10-05
Parse `{{ sub_offer_price_2 }}` and `{{ sub_offer_duration_2 }}` variables

vegaro:
Add support for `{{ sub_offer_price_2 }}` and `{{ sub_offer_duration_2 }}`

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
@@ -33,7 +33,7 @@ internal class VariableDataProvider(
     }
 
     fun localizedIntroductoryOfferPrice(rcPackage: Package): String? {
-        return getIntroPhaseToApply(rcPackage)?.price?.formatted
+        return getFreeOrIntroPhaseToApply(rcPackage)?.price?.formatted
     }
 
     fun productName(rcPackage: Package): String {
@@ -60,7 +60,7 @@ internal class VariableDataProvider(
     }
 
     fun introductoryOfferDuration(rcPackage: Package, locale: Locale): String? {
-        return getIntroPhaseToApply(rcPackage)?.billingPeriod?.localizedPeriod(locale)
+        return getFreeOrIntroPhaseToApply(rcPackage)?.billingPeriod?.localizedPeriod(locale)
     }
 
     fun localizedPricePerPeriod(rcPackage: Package, locale: Locale): String {
@@ -81,10 +81,13 @@ internal class VariableDataProvider(
         return "$pricePerPeriod ($pricePerMonth/$unit)"
     }
 
-    private fun getIntroPhaseToApply(rcPackage: Package): PricingPhase? {
-        // TODO-PAYWALLS: Decide how this will look like for products that have both a free trial
-        // and a discounted introductory price.
+    private fun getFreeOrIntroPhaseToApply(rcPackage: Package): PricingPhase? {
         val option = rcPackage.product.defaultOption
         return option?.freePhase ?: option?.introPhase
+    }
+
+    private fun getIntroPhaseToApply(rcPackage: Package): PricingPhase? {
+        val option = rcPackage.product.defaultOption
+        return option?.introPhase
     }
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
@@ -32,12 +32,12 @@ internal class VariableDataProvider(
         return rcPackage.product.formattedPricePerMonth(locale)
     }
 
-    fun localizedFreeOrIntroductoryOfferPrice(rcPackage: Package): String? {
-        return getFreeOrIntroPhaseToApply(rcPackage)?.price?.formatted
+    fun localizedFirstIntroductoryOfferPrice(rcPackage: Package): String? {
+        return getFirstIntroOfferToApply(rcPackage)?.price?.formatted
     }
 
-    fun localizedIntroductoryOfferPrice(rcPackage: Package): String? {
-        return getIntroPhaseToApply(rcPackage)?.price?.formatted
+    fun localizedSecondIntroductoryOfferPrice(rcPackage: Package): String? {
+        return getSecondIntroOfferToApply(rcPackage)?.price?.formatted
     }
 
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
```diff
@@ -86,6 +69,34 @@ internal object VariableProcessor {
...
+        VariableName.PRODUCT_NAME -> variableDataProvider.productName(rcPackage)
+        VariableName.SUB_PERIOD -> variableDataProvider.periodName(rcPackage)
+        VariableName.SUB_PRICE_PER_MONTH -> variableDataProvider.localizedPricePerMonth(
+            rcPackage,
+            locale,
+        )
+
+        VariableName.SUB_DURATION -> variableDataProvider.subscriptionDuration(rcPackage, locale)
+        VariableName.SUB_OFFER_DURATION -> variableDataProvider.introductoryOfferDuration(rcPackage, locale)
+        VariableName.SUB_OFFER_DURATION_2 -> variableDataProvider.introductoryOfferDuration(rcPackage, locale)
```
tonidero, line 95: Should this pick only the intro phase and not the free phase and then the intro phase? Currently both this and sub_offer_duration will return the same value right?

vegaro: Oh right, if `SUB_OFFER_DURATION` picks `option?.introPhase`, `SUB_OFFER_DURATION_2` will also pick `option?.introPhase`. Is that what you mean?

vegaro: oh wait, nevermind, this is the duration 😮‍💨

tonidero: Well yeah, in that case we should probably pick the pricing phases directly and get them in order. The first one would likely be the free trial, the second one the intro price

vegaro: I realized `option.introPhase` and `option.freePhase` also check if the amounts are 0L, so I think it's better to use those

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -81,10 +89,16 @@ internal class VariableDataProvider(
...
-        // TODO-PAYWALLS: Decide how this will look like for products that have both a free trial
-        // and a discounted introductory price.
+    private fun getFirstIntroOfferToApply(rcPackage: Package): PricingPhase? {
         val option = rcPackage.product.defaultOption
         return option?.freePhase ?: option?.introPhase
     }
+
+    private fun getSecondIntroOfferToApply(rcPackage: Package): PricingPhase? {
+        val option = rcPackage.product.defaultOption
+        if (option?.freePhase != null) {
```
tonidero, line 99: Well, I think this should work as long as google doesn't allow to have an intro price before a free trial, which they currently don't and would also be very weird 😅 . So I think it's ok.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -226,6 +287,37 @@ class VariableProcessorTest {
...
+
+    // region sub_offer_price_2
+
+    @Test
+    fun `process variables processes sub_offer_price_2 for products with no intro price`() {
+        expectVariablesResult("{{ sub_offer_price_2 }}", "", rcPackage = TestData.Packages.annual)
+    }
+
+    @Test
+    fun `process variables processes sub_offer_price_2 for intro prices`() {
```
tonidero, line 300: Should this test be something like `process variables processes sub_offer_price_2 for intro prices when free trial available`? Since it should be null for intro prices when there is no free trial available I would say.

tonidero (approved):
Looks great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
vegaro: Yeah, I thought the same, but I checked and added a intro offer, then a trial, and the trial is still applied first. This is their preview

Screenshot 2023-10-06 at 16 06 14

And yes, that would be strange lol

vegaro:
I broke some tests, checking!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
vegaro: Check [9ea9167](https://github.com/RevenueCat/purchases-android/pull/1313/commits/9ea916718705fa928c82e36f9d107346cf5113c9). I fixed tests. Names were weird or didn't make sense lol

tonidero (approved):
New names make sense 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
```diff
@@ -86,6 +69,34 @@ internal object VariableProcessor {
         }
     }
 
+    private fun processVariable(
```
NachoSoto, line 72: 👍🏻

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorTest.kt
```diff
@@ -201,6 +235,39 @@ class VariableProcessorTest {
         expectVariablesResult("{{ sub_offer_duration }}", "", rcPackage = TestData.Packages.monthly)
     }
 
+    // endregion
+
+    // region sub_offer_duration_2
+
+    @Test
+    fun `process variables processes sub_offer_duration_2 for products with both free trial and intro price`() {
```
NachoSoto, line 243: Nice

NachoSoto (approved):
Awesome!

PR #1317, opened 2023-10-06
Paywalls: Add support to launch paywall as activity

tonidero:
### Description
Adds a `PaywallActivityLauncher` class that devs can instantiate in their activity/fragment `onCreate` in order to launch the `PaywallActivity` which is a wrapper around the full screen `PaywallView`. Then, they receive the result in a `PaywallResultHandler` interface.

Once they want to launch the activity, they can call the `launch` method in that `PaywallActivityLauncher`.


ui/revenuecatui/build.gradle
```diff
@@ -53,6 +62,8 @@ dependencies {
     implementation libs.androidx.lifecycle.viewmodel.compose
     implementation libs.coil.compose
     implementation libs.coil.svg
+    implementation libs.activity.compose
+    implementation libs.androidx.fragment.ktx
```
tonidero, line 66: I've been thinking about splitting all this into a different module, so we don't clutter the `ui:revenuecat` dependency for people not using activities/fragments. The idea would be to have a different dependency for people using activities fragments with all these codes and dependencies.

The down side is to further split into more modules, and we might need to expose some things like the `setOfferingId` in the `PaywallViewOptions`... But I think it might be ok?

ui/revenuecatui/build.gradle
```diff
@@ -24,6 +25,9 @@ android {
 
     defaultConfig {
         minSdkVersion 24 // MeasureFormat requires API 24
+        vectorDrawables {
+            useSupportLibrary true
+        }
```
tonidero, line 30: This was added by the wizard when I created the activity. Need to test whether it's actually needed

ui/revenuecatui/build.gradle
```diff
@@ -32,6 +36,11 @@ android {
     composeOptions {
         kotlinCompilerExtensionVersion '1.4.0-alpha02'
     }
+    packaging {
+        resources {
+            excludes += '/META-INF/{AL2.0,LGPL2.1}'
+        }
+    }
```
tonidero, line 43: Same, added by the wizard. This shouldn't really hurt at all.

ui/revenuecatui/build.gradle
vegaro: I would keep it in the same module, so we don't end up with many modules again. But it's true `androidx.activity:activity-compose` is 1.1 MB, which is not small compared to the 300 KB the billing client takes

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
```diff
@@ -1,17 +1,25 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResult
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler
+
+class MainActivity : ComponentActivity(), PaywallResultHandler {
+    private lateinit var paywallActivityLauncher: PaywallActivityLauncher
 
-class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        paywallActivityLauncher = PaywallActivityLauncher(this, this)
```
vegaro, line 22: What if we make PaywallResultHandler a lambda like this:

```
typealias PaywallResultHandler = (PaywallResult) -> Unit
```

Then in the constructor of the launcher, we can handle the `ActivityResultCallback` there directly:

```
    constructor(activity: ComponentActivity, resultHandler: PaywallResultHandler) {
        activityResultLauncher = activity.registerForActivityResult(PaywallContract()) { paywallResult ->
            resultHandler(it)
        }
    }
```

Then, the activity can be started like this:

```
paywallActivityLauncher = PaywallActivityLauncher(this) { result ->
    Log.e("PaywallsTester", "LAUNCH PAYWALL RESULT: $result")
}
```

This way, folks don't have to make their Activity conform to `PaywallResultHandler` (`ActivityResultCallback`) and override `onActivityResult`.

tonidero: I thought about that... My reasoning to go this way is in case we ever want to add another callback or anything, this would be more extensible even though slightly harder to use... Maybe it's ok though. We can discuss that

vegaro: For adding another callback, we could have it as:

```
    constructor(activity: ComponentActivity, resultHandler: PaywallResultHandler, anotherCallback: NewCallback) {
```

And then we could do ourselves the complicated part, which is figuring out which callback to call using the activity result code. Otherwise devs would have to handle that

tonidero: Well, that's not as extensible IMO... Another point to consider is that this `PaywallActivityLauncher` needs to be created during the `onCreate`, it can't be created after (it will fail at runtime if you try). In that case, you would have a callback at the activity level, not inside a method, so you still need to reference some other class there. I think for a class variable, an interface is a more common pattern... What do you think?

vegaro: Makes sense. Using the interface is actually more flexible + it's using a "system" interface, that devs will probably be familiar with. Let's do the activity

ui/revenuecatui/build.gradle
NachoSoto: Is 1MB much these days? 😅

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewOptions.kt
```diff
@@ -5,25 +5,32 @@ import com.revenuecat.purchases.Offering
 class PaywallViewOptions(builder: Builder) {
 
     val offering: Offering?
+    internal val offeringId: String?
```
NachoSoto, line 8: How come this is needed if we already have `Offering`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -0,0 +1,71 @@
...
+import androidx.compose.ui.Modifier
+import com.revenuecat.purchases.CustomerInfo
+import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesErrorCode
+import com.revenuecat.purchases.models.StoreTransaction
+import com.revenuecat.purchases.ui.revenuecatui.PaywallView
+import com.revenuecat.purchases.ui.revenuecatui.PaywallViewListener
+import com.revenuecat.purchases.ui.revenuecatui.PaywallViewOptions
+
+internal class PaywallActivity : ComponentActivity(), PaywallViewListener {
```
NachoSoto, line 22: It might be nice to add a docstring to this.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallContract.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import androidx.activity.result.contract.ActivityResultContract
+import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesErrorCode
+
+internal class PaywallContract : ActivityResultContract<PaywallActivityArgs, PaywallResult>() {
```
NachoSoto, line 11: I'd love a docstring here too

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallResult.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import android.os.Parcelable
+import com.revenuecat.purchases.CustomerInfo
+import com.revenuecat.purchases.PurchasesError
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Result of the paywall activity.
+ */
+sealed class PaywallResult : Parcelable {
```
NachoSoto, line 11: Is this usable from Java too?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewOptions.kt
tonidero: We can't pass an `Offering` object to the new activity since it's not serializable. As a fallback, in order to display a specific offering in the activity, we allow to pass an offering id and load the offering again from the paywall view itself.

I'm not exposing that API publicly right now, so devs still need to pass the full `Offering` object normally, this would only be used for activities

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallResult.kt
tonidero: It's usable yes, though you don't really have a nice switch as you do in Kotlin, you basically have to check the type and perform the downcast to the corresponding type. It's really a much nicer API in Kotlin, but still usable in Java.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewOptions.kt
NachoSoto: This is a weird definition of the options then.
What do you think about turning this into an enum:
```kotlin
sealed class Selection {
    data class Offering(val offering: Offering) : Selection()
    data class OfferingIdentifier(val id: String) : Selection()
    object None : Selection()
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallResult.kt
NachoSoto: 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallViewOptions.kt
tonidero: Hmm well, I'm currently not exposing the option to pass the identifier to the paywallView publicly. I guess we could hold that type internally to decide what to do though, since they are indeed exclusive. Will do that!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/1317/commits/09c51a8fd2ddd35d952d1d2910f31cbd280b3f9b

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallContract.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/1317/commits/09c51a8fd2ddd35d952d1d2910f31cbd280b3f9b

ui/revenuecatui/build.gradle
tonidero: Well, usually it's much less than that for devs minifying their apps. It's true that we're using a ton of dependencies in the revenuecatui module, so I wasn't sure about trying to optimize it a bit... But yeah, we can keep it in one module for now and avoid having a ton of modules 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -0,0 +1,43 @@
...
+     */
+    constructor(fragment: Fragment, resultHandler: PaywallResultHandler) {
+        activityResultLauncher = fragment.registerForActivityResult(PaywallContract(), resultHandler)
+    }
+
+    /**
+     * Launch the paywall activity.
+     * @param offering The offering to be shown in the paywall. If null, the current offering will be shown.
+     */
+    fun launch(offering: Offering? = null) {
```
vegaro, line 40: should we add `@JvmOverloads` here?

vegaro (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Yup, good point! Haven't been adding API tests yet so I missed that

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/1317/commits/6826ef782fd523f79d47306107c008b28d24beb6

tonidero:
Will merge this, if there are any further comments I can address separately

PR #1318, opened 2023-10-06
Log errors when displaying default paywall

vegaro:
We can't really detect debug builds like we do in iOS, so this PR just adds some debug logs for validation errors that create the default template.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
@@ -1,8 +1,35 @@
 package com.revenuecat.purchases.ui.revenuecatui.errors
 
+import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.ui.revenuecatui.strings.PaywallValidationErrorStrings
+
 sealed class PaywallValidationError : Throwable() {
-    object MissingPaywall : PaywallValidationError()
-    data class InvalidTemplate(val templateName: String) : PaywallValidationError()
-    data class InvalidVariables(val unrecognizedVariables: Set<String>) : PaywallValidationError()
-    data class InvalidIcons(val invalidIcons: Set<String>) : PaywallValidationError()
+
+    abstract fun associatedErrorString(offering: Offering): String
+
+    object MissingPaywall : PaywallValidationError() {
+        override fun associatedErrorString(offering: Offering): String {
+            return PaywallValidationErrorStrings.MISSING_PAYWALL.format(offering.identifier)
+        }
+    }
+
+    data class InvalidTemplate(val templateName: String) : PaywallValidationError() {
+        override fun associatedErrorString(offering: Offering): String {
+            return PaywallValidationErrorStrings.INVALID_TEMPLATE_NAME.format(templateName)
+        }
+    }
+
+    data class InvalidVariables(val unrecognizedVariables: Set<String>) : PaywallValidationError() {
+        override fun associatedErrorString(offering: Offering): String {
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -24,14 +25,18 @@ internal fun Offering.validatedPaywall(
             PaywallData.createDefault(packages = availablePackages, currentColorScheme),
             PaywallData.defaultTemplate,
             PaywallValidationError.MissingPaywall,
-        )
+        ).also {
+            Logger.d(PaywallValidationErrorStrings.MISSING_PAYWALL.format(this.identifier))
```
tonidero, line 29: Hmm I wonder if this should be a warning... I think displaying the default paywall shouldn't be a fallback but shouldn't be relied on normally... Not sure if there will be use cases where people don't want to configure anything in the dashboard though.

Same for the others, I might do them warnings.

tonidero (approved):
Looks good, just a comment

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
vegaro: This is a good point. In iOS we display them in the view, which is a pretty big warning I would say. What do you think @NachoSoto

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -45,14 +50,23 @@ private fun PaywallData.validate(): Result<PaywallTemplate> {
 
     val invalidVariablesError = localizedConfiguration.validateVariables()
     if (invalidVariablesError != null) {
-        return Result.failure(invalidVariablesError)
+        return Result.failure<PaywallTemplate>(invalidVariablesError).also {
+            val joinedUnrecognizedVariables = invalidVariablesError.unrecognizedVariables.joinToString()
+            Logger.d(PaywallValidationErrorStrings.INVALID_VARIABLES.format(joinedUnrecognizedVariables))
```
NachoSoto, line 55: This is kind duplicated in all of these `return`s.
Should we log it in a single line in `PaywallViewModel.calculateState` instead, when `PaywallValidationError` isn't `null`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
```diff
@@ -1,8 +1,35 @@
...
+import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.ui.revenuecatui.strings.PaywallValidationErrorStrings
+
 sealed class PaywallValidationError : Throwable() {
-    object MissingPaywall : PaywallValidationError()
-    data class InvalidTemplate(val templateName: String) : PaywallValidationError()
-    data class InvalidVariables(val unrecognizedVariables: Set<String>) : PaywallValidationError()
-    data class InvalidIcons(val invalidIcons: Set<String>) : PaywallValidationError()
+
+    abstract fun associatedErrorString(offering: Offering): String
```
tonidero, line 8: Hmm instead of doing this an abstract function, you could implement it and use a `when(this) to return the appropriate string for each sealed class type.

tonidero (approved):

PR #1320, opened 2023-10-06
`Paywalls`: internal beta setup

NachoSoto:


examples/MagicWeather/app/src/main/java/com/revenuecat/sample/data/Constants.kt
```diff
@@ -7,15 +7,15 @@ object Constants {
 
      Modify this property to reflect your Google Play app's API key.
      */
-    const val GOOGLE_API_KEY = "googl_api_key"
+    const val GOOGLE_API_KEY = "goog_cRtaFpfwfcQYPXkIEqFUkVszUfN"
```
tonidero, line 10: Oops

build.gradle
```diff
@@ -4,6 +4,8 @@ buildscript {
 }
 
 plugins {
+    id 'maven-publish'
```
tonidero, line 7: Hmm I think this shouldn't be needed? The `alias libs.plugins.mavenPublish apply false` adds the plugin without applying, so each plugin needs to apply it later

ui/revenuecatui/build.gradle
```diff
@@ -1,4 +1,6 @@
 plugins {
+    id 'maven-publish'
```
tonidero, line 2: This shouldn't be needed either I think, just uncommenting the code below should apply the plugin

examples/MagicWeather/app/src/main/java/com/revenuecat/sample/data/Constants.kt
NachoSoto: Pushed it on purpose for testers to be able to run it

build.gradle
NachoSoto: Maybe not, but we're not merging this PR anyway.

NachoSoto:
Closing this.

PR #1321, opened 2023-10-06
`Paywalls`: log error when failing to load images

NachoSoto:


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -28,6 +30,14 @@ internal fun RemoteImage(
         imageLoader = LocalContext.current.getRevenueCatUIImageLoader(),
         modifier = modifier,
         contentScale = contentScale,
+        onState = {
+            when (it) {
+                is AsyncImagePainter.State.Error -> {
+                    Logger.e("Error loading image from '$urlString': ${it.result}")
```
tonidero, line 36: I guess in this case the image is not part of the layout at all right?

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
NachoSoto: I'm not 100% sure, but probably depends on the modifiers.

PR #1322, opened 2023-10-06
`Paywalls`: fixed `PaywallDialog.setRequiredEntitlementIdentifier`

NachoSoto:
Reverse logic discovered during testing.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -37,11 +37,7 @@ fun PaywallDialog(
         LaunchedEffect(paywallDialogOptions) {
             launch {
                 shouldDisplayDialog = try {
-                    // TODO-PAYWALLS: This won't receive updates in case the customer info changes and starts/stops
```
NachoSoto, line 40: I removed this cause we don't really need to worry about it

tonidero (approved):

PR #1324, opened 2023-10-08
Add revenuecatui gradle.properties to specify name of dependency

tonidero:
### Description
Adding a `gradle.properties` file to the `ui:revenuecatui` module to specify the name of the dependency. With this, developers will need to add the dependency like: `implementation "com.revenuecat.purchases:purchases-ui:VERSION_NUMBER"`

### DIFF
--- a/ui/revenuecatui/gradle.properties
+++ b/ui/revenuecatui/gradle.properties
@@ -1,0 +1,4 @@
+POM_NAME=revenuecat-ui
+# We default to deploy the noop version since release is the noop version.
+POM_ARTIFACT_ID=revenuecat-ui
+POM_PACKAGING=aar
--- a/ui/revenuecatui/gradle.properties
+++ b/ui/revenuecatui/gradle.properties
@@ -1,4 +1,4 @@
-POM_NAME=revenuecat-ui
+POM_NAME=purchases-ui
 # We default to deploy the noop version since release is the noop version.
-POM_ARTIFACT_ID=revenuecat-ui
+POM_ARTIFACT_ID=purchases-ui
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_NAME=revenuecat-ui
+# We default to deploy the noop version since release is the noop version.
+POM_ARTIFACT_ID=revenuecat-ui
```
tonidero, line 3: I prefer this name than the one we currently use which is `com.revenuecat.purchases:revenuecatui` since it's more in line with our amazon and debug view dependencies. I think we should get this merged before having people test the library with new apps at least so they use the correct dependency name, cc @NachoSoto

vegaro: The debug view one is `purchases-debug-view` and the amazon one is `purchases-amazon`. Do we need the `revenuecat` part? In iOS the regular SDK is called RevenueCat, but we are still calling the android one `purchases-android`

I wonder if just `ui` works too since it's part of `com.revenuecat.purchases:`

tonidero: Hmm honestly I was between `revenuecat-ui` and `purchases-ui`. While we are under the `com.revenuecat.purchases` group, I think we need a bit more detail in case we ever have multiple ui libraries (or even right now with debugview and the paywall). Do you think `purchases-ui` might be better @vegaro ?

NachoSoto (approved):

vegaro (approved):

PR #1325, opened 2023-10-09
[Do not merge] PoC of overriding fonts through the theme

tonidero:
### Description
This is a small PoC of how it looks like to override the fonts so they are used in the paywall:
image


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/theme/Type.kt
```diff
@@ -2,19 +2,50 @@ package com.revenuecat.paywallstester.ui.theme
...
+        letterSpacing = 0.sp
+    ),
+    titleLarge = TextStyle(
+        fontFamily = superMarioFontFamily,
+        fontWeight = FontWeight.Normal,
+        fontSize = 22.sp,
+        lineHeight = 28.sp,
+        letterSpacing = 0.sp
+    ),
+    titleMedium = TextStyle(
```
tonidero, line 41: This might be a bit confusing... Basically the developer needs to override the specific styles we use within the paywall in order to use their typography everywhere, so maybe we should keep a list of these somewhere?

aboedo:
OMG that font 😂

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/theme/Type.kt
```diff
@@ -2,19 +2,50 @@ package com.revenuecat.paywallstester.ui.theme
 
 import androidx.compose.material3.Typography
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.Font
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.sp
+import com.revenuecat.paywallstester.R
+
+val superMarioFontFamily =
```
aboedo, line 11: I think we should just keep this as the standard font for all paywalls

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/theme/Type.kt
aboedo: we can make it a part of the API, like we did for iOS. So that if you want to use a custom font provider, it has to define what each of these (titleMedium, titleLarge, etc) is

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/theme/Type.kt
tonidero: 🤣 I would totally do this, but we would need to find versions with smaller weights and cursive :P. That's not supported in this font unfortunately

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/theme/Type.kt
tonidero: I was trying to avoid that since it means we need to override the theme for the paywall, instead of relying on the overall theme... but yeah, that might be the best way to make sure this can be customized. I will write up something so it can be part of the API.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
```diff
@@ -35,6 +41,24 @@ class MainActivity : ComponentActivity(), PaywallResultHandler {
...
+                override fun getFont(type: TypographyType): PaywallFontFamily? {
+                    return PaywallFontFamily(
+                        fonts = listOf(
+                            PaywallFont.GoogleFont(
+                                fontName = "Lobster Two",
+                                fontProvider = GoogleFontProvider(R.array.com_google_android_gms_fonts_certs),
+                            ),
+                            PaywallFont.ResourceFont(
+                                resourceId = R.font.super_mario,
+                                fontStyle = FontStyle.Italic,
```
tonidero, line 56: Note how by specifying the fontStyle to italic, this has more preference than the other font for italic, so we only use it there.

tonidero:
Going to close this in favor of #1358

PR #1326, opened 2023-10-09
Fix material theme references to use Material3 versions

tonidero:
### Description
Make sure we use material3 all accross the paywall tester app and the revenuecatui library. Also some linter fixes in `Markdown`


vegaro (approved):
I wonder if there is some sort of linter rule that could warn us of the wrong import

PR #1332, opened 2023-10-10
`Paywalls`: `PaywallColor` supports RGBA

NachoSoto:
Turns out it already did, so I just wrote some tests.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallColor.kt
```diff
@@ -68,10 +68,3 @@ data class PaywallColor(
         },
     )
 }
-
-/**
- * Represents a color scheme.
- */
-enum class ColorScheme {
```
NachoSoto, line 75: Wasn't used.

tonidero (approved):

PR #1333, opened 2023-10-10
Expose new `TestStoreProduct`

NachoSoto:
This can be used to create mock data for tests or Jetpack Compose previews.


purchases/src/main/kotlin/com/revenuecat/purchases/models/TestStoreProduct.kt
```diff
@@ -0,0 +1,95 @@
...
+    override val subscriptionOptions: SubscriptionOptions?
+        get() = buildSubscriptionOptions()
+    override val defaultOption: SubscriptionOption?
+        get() = subscriptionOptions?.defaultOffer
+    override val purchasingData: PurchasingData
+        get() = object : PurchasingData {
+            override val productId: String
+                get() = id
+            override val productType: ProductType
+                get() = type
```
NachoSoto, line 29: This was wrong in the existing implementation, now correctly derived from the above logic

purchases/src/test/java/com/revenuecat/purchases/PurchaseParamsTest.kt
```diff
@@ -44,6 +46,23 @@ class PurchaseParamsTest {
         assertThat(purchaseProductParams.presentedOfferingIdentifier).isEqualTo(STUB_OFFERING_IDENTIFIER)
     }
 
+    @Test
+    fun `Initializing with TestStoreProduct throws error`() {
```
NachoSoto, line 50: Added this test to ensure that attempting to purchase a `TestStoreProduct` fails early.

purchases/src/main/kotlin/com/revenuecat/purchases/models/TestStoreProduct.kt
```diff
@@ -0,0 +1,95 @@
...
+ * This can be used to create mock data for tests or Jetpack Compose previews.
+ */
+data class TestStoreProduct(
+    override val id: String,
+    override val title: String,
+    override val description: String,
+    override val price: Price,
+    override val period: Period?,
+    private val freeTrialPeriod: Period? = null,
+    private val introPrice: Price? = null,
```
tonidero, line 16: I guess these 2 fields would only apply to subscriptions, so I wonder if we should make the API receive a sealed class for Subscription/InApp to support these 2... Anyway I think it's ok for now

purchases/src/main/kotlin/com/revenuecat/purchases/models/TestStoreProduct.kt
```diff
@@ -0,0 +1,95 @@
+package com.revenuecat.purchases.models
+
+import com.revenuecat.purchases.ProductType
+
+/**
+ * A test-only [StoreProduct] implementation.
+ * This can be used to create mock data for tests or Jetpack Compose previews.
+ */
+data class TestStoreProduct(
```
tonidero, line 9: I think we can probably remove some implementations of `StoreProduct` from the tests with this, but we can leave that for another PR.

tonidero (approved):
Makes sense! This should help avoiding a ton of duplication for tests and paywalls 🙌

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/models/TestStoreProduct.kt
NachoSoto: That actually makes sense.

Since I'm merging this in `paywalls` we have time to improve this.

PR #1334, opened 2023-10-10
`Paywalls`: disallow purchasing currently subscribed products

NachoSoto:
In the future we'll need to also update the UI for this (https://linear.app/revenuecat/issue/PWL-243/sdk-update-templates-to-reflect-currently-subscribed-products)

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -61,15 +61,15 @@ internal class PaywallViewModelImpl(
     preview: Boolean = false,
 ) : ViewModel(), PaywallViewModel {
 
-    val variableDataProvider = VariableDataProvider(applicationContext, preview)
+    private val variableDataProvider = VariableDataProvider(applicationContext, preview)
 
     override val state: StateFlow<PaywallViewState>
         get() = _state.asStateFlow()
 
-    val mode: PaywallViewMode
+    private val mode: PaywallViewMode
         get() = options.mode
 
-    private val _state: MutableStateFlow<PaywallViewState>
+    private val _state: MutableStateFlow<PaywallViewState> = MutableStateFlow(PaywallViewState.Loading)
     private val _lastLocaleList = MutableStateFlow(getCurrentLocaleList())
     private val _colorScheme = MutableStateFlow(colorScheme)
 
@@ -77,12 +77,7 @@ internal class PaywallViewModelImpl(
         get() = options.listener
 
     init {
-        val selectedOffering = options.offeringSelection.offering
-        _state = MutableStateFlow(PaywallViewState.Loading)
-
-        if (selectedOffering == null) {
-            updateState()
-        }
+        updateState()
     }
 
     override fun refreshStateIfLocaleChanged() {
@@ -104,6 +99,7 @@ internal class PaywallViewModelImpl(
             is PaywallViewState.Loaded -> {
                 currentState.selectPackage(packageToSelect)
             }
+
             else -> {
                 Logger.e("Unexpected state trying to select package: $currentState")
             }
@@ -164,7 +160,7 @@ internal class PaywallViewModelImpl(
                 var currentOffering = options.offeringSelection.offering
                 if (currentOffering == null) {
                     val offerings = Purchases.sharedInstance.awaitOfferings()
-                    options.offeringSelection.offeringIdentifier?.let { offerings[it] }
+                    currentOffering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
                         ?: offerings.current
                 }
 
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -114,7 +114,10 @@ internal class PaywallViewModelImpl(
         val activity = context.getActivity() ?: error("Activity not found")
         when (val currentState = _state.value) {
             is PaywallViewState.Loaded -> {
-                purchasePackage(activity, currentState.selectedPackage.rcPackage)
+                val selectedPackage = currentState.selectedPackage
+                if (!selectedPackage.currentlySubscribed) {
```
tonidero, line 118: I think we need to pass in the currently subscribed product ids when creating the `TemplateConfiguration`, so this won't work yet... But I don't think it hurts either.

tonidero: Also, until we change the UI, should we log a message when the user is subscribed?

tonidero (approved):
Mostly, I would add a log for now. Other than that, looks good

NachoSoto:
@vegaro @tonidero requesting review again cause I've changed this significantly to compute `activelySubscribedProductIdentifiers`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -59,41 +61,41 @@ internal class PaywallViewModelImpl(
...
-
     init {
-        _state = MutableStateFlow(offering?.calculateState() ?: PaywallViewState.Loading)
-        if (offering == null) {
-            updateOffering()
+        val selectedOffering = options.offeringSelection.offering
+        _state = MutableStateFlow(PaywallViewState.Loading)
+
+        if (selectedOffering == null) {
+            updateState()
```
NachoSoto, line 84: I've simplified this a lot because now we need to fetch `Offerings` and `CustomerInfo` (like in iOS). So there's now only a single path: `updateState` does all the logic

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfigurationFactoryTest.kt
```diff
@@ -31,7 +31,9 @@ internal class TemplateConfigurationFactoryTest {
             paywallViewMode,
             TestData.template2,
             listOf(TestData.Packages.weekly, TestData.Packages.monthly, TestData.Packages.annual),
-            emptySet(),
+            setOf(
+                TestData.Packages.monthly.product.id
```
NachoSoto, line 35: Added this to test the value.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -59,41 +61,41 @@ internal class PaywallViewModelImpl(
     preview: Boolean = false,
 ) : ViewModel(), PaywallViewModel {
 
-    private val variableDataProvider = VariableDataProvider(applicationContext, preview)
```
tonidero, line 62: Hmm any reason we removed this `private`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -162,38 +160,56 @@ internal class PaywallViewModelImpl(
...
+        return LocaleListCompat.getDefault()
+    }
+}
 
-        error?.let { validationError ->
-            Logger.w(validationError.associatedErrorString(this))
-            Logger.w(PaywallValidationErrorStrings.DISPLAYING_DEFAULT)
-        }
-        return toPaywallViewState(variableDataProvider, mode, displayablePaywall, template)
+@Suppress("TooGenericExceptionCaught")
```
tonidero, line 193: I think this line is probably not needed anymore? Since we are not catching any errors.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -59,41 +61,41 @@ internal class PaywallViewModelImpl(
     preview: Boolean = false,
 ) : ViewModel(), PaywallViewModel {
 
-    private val variableDataProvider = VariableDataProvider(applicationContext, preview)
+    val variableDataProvider = VariableDataProvider(applicationContext, preview)
 
     override val state: StateFlow<PaywallViewState>
         get() = _state.asStateFlow()
+
+    val mode: PaywallViewMode
```
tonidero, line 69: Same here, I think this is only used inside this class?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: Hmm I'm actually wondering, if `selectedOffering` is not null, where do we set the `_state` to the final value?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -162,38 +160,56 @@ internal class PaywallViewModelImpl(
...
         viewModelScope.launch {
             try {
-                val offerings = Purchases.sharedInstance.awaitOfferings()
-                val currentOffering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
-                    ?: offerings.current
+                var currentOffering = options.offeringSelection.offering
+                if (currentOffering == null) {
+                    val offerings = Purchases.sharedInstance.awaitOfferings()
+                    options.offeringSelection.offeringIdentifier?.let { offerings[it] }
+                        ?: offerings.current
```
tonidero, line 170: Hmm I think we're not using the result from this? I think we want to save this into `currentOffering` correct?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Hmm good catch, it was because `calculateState` was defined outside but I put it inside now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Fixed

NachoSoto: I was missing `fileprivate` 😅 but I just moved `calculateState` inside

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Hmmm good catch, fixed.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: OMG yeah 🤦🏻  we really need tests for this class.

NachoSoto:
@tonidero thanks for the thorough review! I fixed everything now.
I'm gonna add tests for the view model today.

tonidero (approved):
Looks great!

PR #1335, opened 2023-10-10
`Paywalls`: new `PaywallActivityLauncher.launchIfNeeded` methods

NachoSoto:
This mimics the equivalent API in `PaywallDialogOptions` (and like in iOS).

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
@@ -4,7 +4,10 @@ import androidx.activity.ComponentActivity
 import androidx.activity.result.ActivityResultCallback
 import androidx.activity.result.ActivityResultLauncher
 import androidx.fragment.app.Fragment
+import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayBlockForEntitlementIdentifier
+import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayPaywall
 
 /**
  * Implement this interface to receive the result of the paywall activity.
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/HelperFunctions.kt
```diff
@@ -3,7 +3,33 @@ package com.revenuecat.purchases.ui.revenuecatui.helpers
...
+    }
+
+    return shouldDisplayDialog
+}
+
+internal fun shouldDisplayBlockForEntitlementIdentifier(entitlement: String): ((CustomerInfo) -> Boolean) {
+    return { customerInfo ->
+        customerInfo.entitlements[entitlement]?.isActive != true
+    }
+}
```
NachoSoto, line 35: I've extracted these to share this code.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -41,4 +44,33 @@ class PaywallActivityLauncher {
     fun launch(offering: Offering? = null) {
         activityResultLauncher.launch(PaywallActivityArgs(offeringId = offering?.identifier))
     }
+
+    /**
+     * Launch the paywall activity if the current user does not have [requiredEntitlementIdentifier] active.
+     * @param offering The offering to be shown in the paywall. If null, the current offering will be shown.
+     */
+    @JvmOverloads
+    suspend fun launchIfNeeded(
```
NachoSoto, line 53: Not sure if this is the best way to do this (exposing a `suspend` method). Feedback welcome.

tonidero: Hmm well, I imagine many people using activities/fragments are still not using coroutines, so offering a wrapper would be better... However, we would only be able to launch this coroutine from a global scope unless we ask developers to give us the launcher activity/fragment/view model to act as the scope...

I see 2 options here:
- Store a weak reference to the activity/fragment on the constructor so we can use that to launch the coroutine in that scope, and offer a non-coroutine version of these methods.
- Have a non-coroutine version of `shouldDisplayPaywall` that returns a callback so we can support this use case. We can wrap that callback in a coroutine for the current version, same way we do in `CoroutinesExtensionsCommon`.

Of those 2, I think the second might be the cleanest, but the first one is probably ok.
Any other ideas @vegaro ?

NachoSoto: Thanks! Just did option 2 :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -41,4 +44,35 @@ class PaywallActivityLauncher {
     fun launch(offering: Offering? = null) {
         activityResultLauncher.launch(PaywallActivityArgs(offeringId = offering?.identifier))
     }
+
+    /**
+     * Launch the paywall activity if the current user does not have [requiredEntitlementIdentifier] active.
+     * @param offering The offering to be shown in the paywall. If null, the current offering will be shown.
```
tonidero, line 50: We should add the docs for the `requiredEntitlementIdentifier` parameter.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -41,4 +44,35 @@ class PaywallActivityLauncher {
...
+    ) {
+        launchIfNeeded(
+            offering = offering,
+            shouldDisplayBlock = shouldDisplayBlockForEntitlementIdentifier(requiredEntitlementIdentifier),
+        )
+    }
+
+    /**
+     * Launch the paywall activity based on whether the result of [shouldDisplayBlock] is true.
+     * @param offering The offering to be shown in the paywall. If null, the current offering will be shown.
```
tonidero, line 65: Same here about the docs for the new parameter.

tonidero (approved):
Looks great!

PR #1337, opened 2023-10-10
`Paywalls`: animate package selection transitions

NachoSoto:
https://github.com/RevenueCat/purchases-android/assets/685609/f7d337bb-613f-47c1-8429-13a270171236

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
@@ -1,5 +1,7 @@
 package com.revenuecat.purchases.ui.revenuecatui
 
+import androidx.compose.animation.core.LinearOutSlowInEasing
+import androidx.compose.animation.core.tween
 import androidx.compose.ui.unit.dp
 
 internal object UIConstant {
@@ -12,4 +14,9 @@ internal object UIConstant {
     val defaultPackageBorderWidth = 1.5.dp
 
     const val defaultAnimationDurationMillis = 200
+
+    fun <T> defaultAnimation() = tween<T>(
+        durationMillis = 200,
+        easing = LinearOutSlowInEasing,
+    )
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/IntroEligibilityStateView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/IntroEligibilityStateView.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.revenuecatui.composables
 
+import androidx.compose.animation.Crossfade
 import androidx.compose.runtime.Composable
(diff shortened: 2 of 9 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewState.kt
```diff
@@ -11,12 +12,12 @@ internal sealed class PaywallViewState {
     data class Error(val errorMessage: String) : PaywallViewState()
     data class Loaded(
         val templateConfiguration: TemplateConfiguration,
-        val selectedPackage: TemplateConfiguration.PackageInfo,
+        val selectedPackage: MutableState<TemplateConfiguration.PackageInfo>,
```
tonidero, line 15: Hmm I was doubting whether exposing a `MutableState` to the View model and the data layer was a good idea, but seems to be accepted from what I've seen.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/IntroEligibilityStateView.kt
```diff
@@ -33,14 +35,20 @@ internal fun IntroEligibilityStateView(
...
 
-    Markdown(
-        text,
-        color = color,
-        style = style,
-        fontWeight = fontWeight,
-        textAlign = textAlign,
-        modifier = modifier,
-    )
+    Crossfade(
```
tonidero, line 38: I was wondering whether we should optimize and only apply the Crossfade in templates that allow to select a package... But I think it should be ok since we're not changing the text anyway in those.

tonidero (approved):
I would love to figure out why the animation seems to take longer than 200ms... Other than that, this looks good

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
```diff
@@ -12,4 +14,9 @@ internal object UIConstant {
     val defaultPackageBorderWidth = 1.5.dp
 
     const val defaultAnimationDurationMillis = 200
+
+    fun <T> defaultAnimation() = tween<T>(
+        durationMillis = 200,
```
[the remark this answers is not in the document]
NachoSoto: Yeah 2 seconds. I made it slower to see it in detail.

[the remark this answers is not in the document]
tonidero: Ah makes sense. Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewState.kt
NachoSoto: And looks like it's the only way to be able to derive animations from it. I tried making it private and exposing only the value, but that doesn't work.

PR #1328, opened 2023-10-10
Paywalls: Support custom fonts through FontProvider

tonidero:
### Description
This adds a way for devs to give us a `FontProvider`, which will be used to override the font family in our views when a font family is provided. If it's not provided, the current material3 theme's font will be used.

PR #1338, opened 2023-10-11
Paywalls: Support Google fonts and font families with multiple fonts

tonidero:
### Description
Followup to #1328 

This PR adds support for Google fonts when launching the paywall as an activity and modifies the API to better support both these and bundled fonts and font families with multiple fonts.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
@@ -48,24 +48,22 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
 
     private fun getFontProvider(): FontProvider? {
         val googleFontProviders = mutableMapOf<GoogleFontProvider, GoogleFont.Provider>()
-        val fontsMap = getArgs()?.fonts?.mapValues { fontFamilyMap ->
-            fontFamilyMap.value?.let { fontFamily ->
-                val fonts = fontFamily.fonts.map { font ->
-                    when (font) {
-                        is PaywallFont.ResourceFont -> Font(font.resourceId, font.fontWeight, font.fontStyle)
-                        is PaywallFont.GoogleFont -> {
-                            val googleFontProvider = font.fontProvider
-                            val provider = googleFontProviders.getOrElse(googleFontProvider) {
-                                val googleProvider = googleFontProvider.toGoogleProvider()
-                                googleFontProviders[googleFontProvider] = googleProvider
-                                googleProvider
-                            }
-                            Font(GoogleFont(font.fontName), provider, font.fontWeight, font.fontStyle)
+        val fontsMap = getArgs()?.fonts?.mapValues { (_, fontFamily) ->
+            val fonts = fontFamily?.fonts?.map { font ->
+                when (font) {
+                    is PaywallFont.ResourceFont -> Font(font.resourceId, font.fontWeight, font.fontStyle)
+                    is PaywallFont.GoogleFont -> {
+                        val googleFontProvider = font.fontProvider
+                        val provider = googleFontProviders.getOrElse(googleFontProvider) {
+                            val googleProvider = googleFontProvider.toGoogleProvider()
+                            googleFontProviders[googleFontProvider] = googleProvider
+                            googleProvider
                         }
+                        Font(GoogleFont(font.fontName), provider, font.fontWeight, font.fontStyle)
                     }
                 }
-                FontFamily(fonts)
             }
+            fonts?.let { FontFamily(it) }
         } ?: return null
         return object : FontProvider {
             override fun getFont(type: TypographyType): FontFamily? {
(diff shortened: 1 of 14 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/PaywallFontFamily.kt
```diff
@@ -0,0 +1,10 @@
+package com.revenuecat.purchases.ui.revenuecatui.fonts
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Represents a font family. You can add one ore more [PaywallFont] with different weights and font styles.
+ */
+@Parcelize
+data class PaywallFontFamily(val fonts: List<PaywallFont>) : Parcelable
```
tonidero, line 10: We could probably not have this and just have a list of fonts... But seemed like a better idea to have a wrapper to mimick what it looks like for Google.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -43,10 +47,25 @@ internal class PaywallActivity : ComponentActivity(), PaywallViewListener {
     }
 
     private fun getFontProvider(): FontProvider? {
-        return getArgs()?.fonts?.let { fonts ->
-            val fontsMap = fonts.mapValues { entry ->
-                entry.value?.let { fontRes ->
-                    ResourcesCompat.getFont(this, fontRes)?.let { FontFamily(it) }
+        return getArgs()?.fonts?.let { fontFamilyMap ->
+            val googleFontProviders = mutableMapOf<GoogleFontProvider, GoogleFont.Provider>()
```
tonidero, line 51: This is a small optimization... Honestly, it's probably not needed, but didn't want to create the provider multiple times.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/PaywallFont.kt
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Represents a font. You can create either a [GoogleFont] or a [ResourceFont].
+ */
+sealed class PaywallFont : Parcelable {
+    /**
+     * Represents a downloadable Google Font.
+     */
+    @Parcelize
+    data class GoogleFont(
```
tonidero, line 18: I actually wondered whether we wanted to offer a similar API for compose... But I think it's ok to leave the responsibility of building the FontFamily (from any source) to the developer

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -43,10 +47,25 @@ internal class PaywallActivity : ComponentActivity(), PaywallViewListener {
...
+            val fontsMap = fontFamilyMap.mapValues { entry ->
```
vegaro, line 52: I would destructure here:

```
            val fontsMap = fontFamilyMap.mapValues { (_, value) ->
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/GoogleFontProvider.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.os.Parcelable
+import androidx.annotation.ArrayRes
+import androidx.compose.ui.text.googlefonts.GoogleFont
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Represents a Google font provider.
+ */
+@Parcelize
+data class GoogleFontProvider(
```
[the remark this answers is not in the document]
vegaro: Oh and it's parcelable, is that the reason?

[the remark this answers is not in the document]
tonidero: Yeah, `GoogleFont.Provider` is not parcelable... I thought about writing a custom parceler, but we don't have access to the underlying data.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/FontProvider.kt
```diff
@@ -6,6 +6,7 @@ import com.revenuecat.purchases.ui.revenuecatui.PaywallView
 /**
  * Implement this interface to provide custom fonts to the [PaywallView]. If you don't, the current material3 theme
  * typography will be used.
+ * If you only want to use a single FontFamily for all text styles use [CustomFontProvider].
```
NachoSoto, line 9: 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/PaywallFont.kt
NachoSoto: What would that look like?

NachoSoto (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/PaywallFont.kt
tonidero: Basically it would be to allow using a `ParcelableFontProvider` in `PaywallViewOptions`, so both using a GoogleFont or a `ResourceFont` are allowed... But again, I think if they are using a custom font from those places, it's easier for them to just pass us the `FontFamily` directly in compose world, where there is no need for this data to be parcelable.

PR #1342, opened 2023-10-12
`Paywalls`: improved default paywall

NachoSoto:
## Changes:

- Added background image
- Improved colors

## Before:

Screenshot 2023-10-12 at 14 35 17

## After:

Screenshot 2023-10-12 at 14 31 19

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallBackground.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallBackground.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.revenuecatui.composables
 
+import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PaywallDataExtensions.kt
```diff
@@ -82,25 +82,23 @@ private val PaywallData.Companion.defaultTemplateBaseURL: URL
 private fun PaywallData.Companion.defaultColors(
     currentColorScheme: ColorScheme,
 ): PaywallData.Configuration.ColorInformation {
+    val colors = getThemeColors(currentColorScheme)
     return PaywallData.Configuration.ColorInformation(
-        light = getThemeColors(background = PaywallColor(colorInt = Color.White.toArgb()), currentColorScheme),
```
NachoSoto, line 86: I removed these hardcoded colors because we could get it with `currentColorScheme.background`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallBackground.kt
```diff
@@ -1,34 +1,51 @@
...
+        .conditional(templateConfiguration.configuration.blurredBackgroundImage) {
+            // TODO-PAYWALLS: backwards compatibility for blurring
+            blur(BackgroundUIConstants.blurSize, edgeTreatment = BlurredEdgeTreatment.Unbounded)
+                .alpha(BackgroundUIConstants.blurAlpha)
+        }
+
+    if (templateConfiguration.configuration.images.background == PaywallData.defaultBackgroundPlaceholder) {
+        Image(
+            modifier = modifier,
+            painter = painterResource(id = R.drawable.default_background),
```
vegaro, line 32: Something I talked with @tonidero is what if we had a generated background instead of a this default jpg with mountains and a lake. We could maybe do a gradient from background to foreground or something that doesn't' take space in the SDK and works fine in all device dimensions.

Not really feedback for this PR, just something I wanted to write somewhere lol

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PaywallDataExtensions.kt
```diff
@@ -82,25 +82,23 @@ private val PaywallData.Companion.defaultTemplateBaseURL: URL
...
-private fun getThemeColors(
-    background: PaywallColor,
-    currentColorScheme: ColorScheme,
-): PaywallData.Configuration.Colors {
+private fun getThemeColors(currentColorScheme: ColorScheme): PaywallData.Configuration.Colors {
     return PaywallData.Configuration.Colors(
-        background = background,
-        text1 = currentColorScheme.primary.asPaywallColor(),
+        background = currentColorScheme.background.asPaywallColor(),
+        text1 = currentColorScheme.inverseSurface.asPaywallColor(),
```
vegaro, line 97: Oh nice, the text was actually hard to see due to this. Now it looks good

vegaro (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallBackground.kt
NachoSoto: > works fine in all device dimensions.

Well the image is blurred so it doesn't matter if it's squashed in different aspect ratios. But yeah if we can make something that looks nice to replace it I'm all for it.

PR #1345, opened 2023-10-13
Fix loading another template in Paywalls screen

vegaro:
There was a bug when loading a different template after dismissing a paywall. The offering would not change and would always load the first offering loaded

It took me a bit to realize what's going on, but we have this piece of code in the InternalPaywallView

```
@Composable
private fun getPaywallViewModel(
    options: PaywallViewOptions,
): PaywallViewModel {
    return viewModel<PaywallViewModelImpl>(
        // We need to pass the key in order to create different view models for different offerings when
        // trying to load different paywalls for the same view model store owner.
        key = options.offeringSelection.offeringIdentifier,
        factory = PaywallViewModelFactory(
            applicationContext.toAndroidContext(),
            options,
            MaterialTheme.colorScheme,
            preview = isInPreviewMode(),
        ),
    )
}
```

The key used to identify the viewmodel is the offering identifier, and for the offerings we create in the SamplePaywalls class, the offering identifier is always the same: `offering`

This was causing the InternalPaywallView to always load the same ViewModel, the first one that was created.

I updated the code to create a unique identifier for every sample offering.

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/SamplePaywalls.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/SamplePaywalls.kt
@@ -13,7 +13,7 @@ import java.net.URL
 class SamplePaywallsLoader {
     fun offeringForTemplate(template: SamplePaywalls.SampleTemplate): Offering {
         return Offering(
-            SamplePaywalls.offeringIdentifier,
+            "$SamplePaywalls.offeringIdentifier_${template.id}",
             SamplePaywalls.offeringIdentifier,
             emptyMap(),
             SamplePaywalls.packages,


NachoSoto (approved):
Haha I was looking into it and this was my theory, but couldn't find where we'd be using the identifier as a "key".

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/SamplePaywalls.kt
```diff
@@ -13,7 +13,7 @@ import java.net.URL
 class SamplePaywallsLoader {
     fun offeringForTemplate(template: SamplePaywalls.SampleTemplate): Offering {
         return Offering(
-            SamplePaywalls.offeringIdentifier,
+            "$SamplePaywalls.offeringIdentifier_${template.id}",
```
NachoSoto, line 16: `SampleTemplate` no longer has an ID

vegaro: #1347

PR #1346, opened 2023-10-13
`Paywalls`: handle "action in progress" state

NachoSoto:
https://github.com/RevenueCat/purchases-android/assets/685609/1c13a65e-a630-477c-b82e-b00a98b9fdc6

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
@@ -1,26 +1,53 @@
 package com.revenuecat.purchases.ui.revenuecatui.composables
 
+import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
 import com.revenuecat.purchases.ui.revenuecatui.UIConstant
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewModel
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewState
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.TemplateConfiguration
+import com.revenuecat.purchases.ui.revenuecatui.data.testdata.MockViewModel
+import com.revenuecat.purchases.ui.revenuecatui.data.testdata.TestData
 import com.revenuecat.purchases.ui.revenuecatui.extensions.introEligibility
 
 @Composable
 internal fun PurchaseButton(
     state: PaywallViewState.Loaded,
+    viewModel: PaywallViewModel,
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
```diff
@@ -1,28 +1,55 @@
...
+        colors = state.templateConfiguration.getCurrentColors(),
+        packages = state.templateConfiguration.packages,
+        selectedPackage = state.selectedPackage,
+        viewModel = viewModel,
+        childModifier = childModifier,
+    )
+}
+
+@Composable
+private fun PurchaseButton(
```
NachoSoto, line 47: I extracted this so the actual definition only requires the raw types and not the whole state.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/Animations.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.graphics.Color
+import com.revenuecat.purchases.ui.revenuecatui.UIConstant
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewModel
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewState
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.TemplateConfiguration
+
+@Composable
+internal fun PaywallViewModel.packageButtonActionInProgressOpacityAnimation(): Float {
```
NachoSoto, line 14: I extracted these so they can be used by all templates.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -63,7 +68,11 @@ internal class PaywallViewModelImpl(
 
     override val state: StateFlow<PaywallViewState>
         get() = _state.asStateFlow()
+    override val actionInProgress: State<Boolean>
+        get() = derivedStateOf { _actionInProgress.value }
```
tonidero, line 72: Hmm I think `derivedStateOf` here is not needed? I think this could be:
```suggestion
        get() = _actionInProgress
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -122,13 +133,18 @@ internal class PaywallViewModelImpl(
     }
 
     override fun restorePurchases() {
+        if (_actionInProgress.value) { return }
```
tonidero, line 136: Maybe we should log something in these cases?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -157,6 +176,8 @@ internal class PaywallViewModelImpl(
             } catch (e: PurchasesException) {
                 listener?.onPurchaseError(e.error)
             }
+
+            _actionInProgress.value = false
```
tonidero, line 180: Hmm I guess depending on the implementation, the paywall might be dismissed by the time we execute this, but I think it should be ok even if that's the case.

tonidero (approved):
Looks great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Oh good call 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Okay 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Yeah 👍🏻

PR #1348, opened 2023-10-13
`Paywalls`: support `{{ sub_duration_in_months }}`

NachoSoto:


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -102,3 +107,13 @@ internal class VariableDataProvider(
         return null
     }
 }
+
+@SuppressWarnings("MagicNumber")
+private fun Period.normalizedMonths(): Period {
+    return if (unit == Period.Unit.YEAR) {
```
tonidero, line 113: Hmm I guess this only will work for year... We should be careful if we have weeks/days.... Maybe add a comment to this method?

tonidero (approved):

PR #1350, opened 2023-10-16
Paywalls: Add restore paywall callbacks

tonidero:
### Description
We were not calling the restore callbacks. This fixes that.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -124,10 +124,13 @@ internal class PaywallViewModelImpl(
     override fun restorePurchases() {
```
NachoSoto, line 124: It would be nice to have tests for `PaywallViewModel` too (example: https://github.com/RevenueCat/purchases-ios/blob/main/Tests/RevenueCatUITests/Purchasing/PurchaseHandlerTests.swift#L55)

tonidero: Absolutely we should. I'm taking note so I will add them tomorrow.

PR #1358, opened 2023-10-17
Add custom font example to paywalls screen

tonidero:
### Description
This enables using a custom font from the `PaywallsScreen` in the paywalls tester app


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -22,18 +22,27 @@ import androidx.compose.ui.tooling.preview.Preview
...
 import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
+import com.revenuecat.purchases.ui.revenuecatui.fonts.CustomFontProvider
+import com.revenuecat.purchases.ui.revenuecatui.fonts.FontProvider
 
 @Composable
 fun PaywallsScreen(
     samplePaywallsLoader: SamplePaywallsLoader = SamplePaywallsLoader(),
 ) {
     var displayPaywallDialogOffering by remember { mutableStateOf<Offering?>(null) }
+    var fontProvider by remember { mutableStateOf<FontProvider?>(null) }
```
NachoSoto, line 40: At least in SwiftUI the way to do this would be to combine the state in a common type. Would that be too verbose in Kotlin?

NachoSoto (approved):
Awesome!!

NachoSoto:
Want to add an example to the docs now? 🙏🏻

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
tonidero: Well, we could definitely make a `sealed class` containing all this state, which should be cleaner 👍 Will do that change.

tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/1358/commits/7766689ef6edae5c94c6a01f29c5164d2dcaa2ab. Do you want to take another look @NachoSoto ?

PR #1360, opened 2023-10-18
Paywalls: Fix paywall compose previews

tonidero:
### Description
The template previews were broken after some recent changes to the VM that made it so the state didn't change to the loaded state in the view model after some operations happened in the SDK, which wasn't available. This just uses a mock view model for the templates moving forward.


vegaro (approved):

NachoSoto:
I'll merge this cause I need it 😇

PR #1304, opened 2023-10-03
Add purchases.setOnesignalUserID

Raquel10-RevenueCat:
### Motivation
OneSignal is migrating to a new User Centric API (v11.0). Adding a new reserved attribute to provide the new OneSignal User Id.

### Description
* Added $onesignalUserId reserved attribute with setter methods

PR #1361, opened 2023-10-18
[AUTOMATIC] Release/7.1.0

RCGitBot:

### New Features
* Add purchases.setOnesignalUserID (#1304) via Raquel Diez (@Raquel10-RevenueCat)


NachoSoto:
Should we skip this for now? I feel like it will be confusing since we have a 7.1.0-beta.1 already.

tonidero:
Hmm seems like something we probably want to ship, but I think it's not urgent. @Raquel10-RevenueCat let us know if there are any time constraints on shipping this. I think it won't take long until we have paywalls out in a stable release and we can ship this with that at that point. 

Will close for now, but we can reopen if needed.

PR #1363, opened 2023-10-18
`Paywalls`: improve handling of lifetime/custom packages

NachoSoto:
This uses the existing `TemplateConfiguration.PackageInfo.currentlySubscribed` to prevent users from subscribing to a lifetime subscription they already own.

Better consumable support will come in the future.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.ui.revenuecatui.data.processed
 
 import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.PackageType
 import com.revenuecat.purchases.paywalls.PaywallData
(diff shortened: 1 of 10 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -32,10 +34,22 @@ internal object PackageConfigurationFactory {
...
+            val currentlySubscribed = when (it.packageType) {
+                PackageType.ANNUAL,
+                PackageType.SIX_MONTH,
+                PackageType.THREE_MONTH,
+                PackageType.TWO_MONTH,
+                PackageType.MONTHLY,
+                PackageType.WEEKLY,
+                -> activelySubscribedProductIdentifiers.contains(it.product.id)
+                PackageType.LIFETIME -> nonSubscriptionProductIdentifiers.contains(it.product.id)
+                PackageType.UNKNOWN, PackageType.CUSTOM -> false
```
tonidero, line 46: Should we move `PackageType.CUSTOM` above, so we also disallow purchasing for existing subscriptions in custom packages?

tonidero (approved):
Just a comment but looks good

PR #1364, opened 2023-10-18
`Paywalls`: improve `LoadingPaywall`

NachoSoto:
Added alpha and corner radius to the placeholder items. Also changed how `Footer` is implemented to only make the `Text` a placeholder.

Screenshot 2023-10-18 at 15 12 42


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -67,7 +67,7 @@ private fun Footer(
     val context = LocalContext.current
 
     Row(
-        modifier = childModifier
+        modifier = Modifier
```
NachoSoto, line 70: Move this to each `Text`.

PR #1366, opened 2023-10-19
Paywalls: Support footer in template 1

tonidero:
### Description
Template 1 did not support footer presentation.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
@@ -48,7 +48,7 @@ internal fun Footer(
     templateConfiguration: TemplateConfiguration,
     viewModel: PaywallViewModel,
     childModifier: Modifier = Modifier,
-    allPlansTapped: () -> Unit = {},
+    allPlansTapped: (() -> Unit)? = null,
 ) {
     Footer(
         mode = templateConfiguration.mode,
@@ -68,7 +68,7 @@ private fun Footer(
     colors: TemplateConfiguration.Colors,
     viewModel: PaywallViewModel,
     childModifier: Modifier = Modifier,
-    allPlansTapped: () -> Unit = {},
+    allPlansTapped: (() -> Unit)? = null,
 ) {
     val context = LocalContext.current
 
@@ -83,7 +83,7 @@ private fun Footer(
     ) {
         val color = colors.text1
 
-        if (mode == PaywallMode.FOOTER_CONDENSED) {
+        if (mode == PaywallMode.FOOTER_CONDENSED && allPlansTapped != null) {
             Button(
                 color = color,
                 childModifier = childModifier,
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -83,7 +83,7 @@ private fun Footer(
     ) {
         val color = colors.text1
 
-        if (mode == PaywallMode.FOOTER_CONDENSED) {
+        if (mode == PaywallMode.FOOTER_CONDENSED && allPlansTapped != null) {
```
tonidero, line 86: Since template 1 doesn't support the condensed footer version (it's already a single option to begin with), I made the parameter nullable, and use that to detect whether to show the button or not. This is the easiest way I think to not show the condensed button in templates where it's not supported.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -53,12 +55,10 @@ import com.revenuecat.purchases.ui.revenuecatui.extensions.introEligibility
 @Composable
 internal fun Template1(state: PaywallState.Loaded, viewModel: PaywallViewModel) {
     Column(
-        modifier = Modifier.fillMaxSize(),
+        modifier = Modifier.fillMaxWidth(),
         horizontalAlignment = Alignment.CenterHorizontally,
     ) {
-        Column(modifier = Modifier.weight(1f)) {
```
tonidero, line 59: I had to change the layout a bit to support footer.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -70,52 +70,55 @@ private fun ColumnScope.Template1MainContent(state: PaywallState.Loaded) {
...
+    if (state.isInFullScreenMode) {
+        Column(
             modifier = Modifier
-                .padding(horizontal = UIConstant.defaultHorizontalPadding),
+                .fillMaxWidth()
+                .verticalScroll(rememberScrollState())
+                .fillMaxHeight()
+                .weight(1f),
+            horizontalAlignment = Alignment.CenterHorizontally,
+            verticalArrangement = Arrangement.Center,
```
tonidero, line 81: Note that I removed the space between in the vertical arrangement. Instead, I'm relying on the `Spacer` inside the column. Those should stretch and contract depending on the available space. If it gets to too much content, it should become scrollable.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
NachoSoto: That makes sense. Looking at iOS I did the same thing.

NachoSoto (approved):
Awesome 👏🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
tonidero: There are slight differences in the space between elements, since we are now relying on the spacers instead of the verticalArrangment spacebetween, but I think it looks mostly the same

tonidero: But yeah, I will try to get some snapshot tests running tomorrow 🤞

PR #1369, opened 2023-10-19
Add `StoreProduct.pricePerMonth`

NachoSoto:
This is a counterpart to the existing `formattedPricePerMonth`. It becomes the canonical implementation returning a `Price` instead.

This will be used for Paywalls to calculate discounts between packages.


tonidero (approved):
LGTM!

PR #1349, opened 2023-10-13
`Paywalls`: template 4

NachoSoto:
https://github.com/RevenueCat/purchases-android/assets/664544/0e6745dc-dac0-49b4-82e6-310111b574f8

Discount labels are automatically resized to fit

Screenshot 2023-10-25 at 16 24 48

Screenshot 2023-10-25 at 16 25 40

For three or less packages, this is how it looks

Screenshot 2023-10-25 at 16 29 59

PR #1370, opened 2023-10-19
`Paywalls`: calculate `discountRelativeToMostExpensivePerMonth`

NachoSoto:
Depends on #1369.
This is needed for template 4 (#1349).

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
@@ -86,15 +86,8 @@ internal object PackageConfigurationFactory {
 
     private fun mostExpensivePricePerMonth(packages: List<Package>): Price? {
         return packages
-            .asSequence()
-            .map { it.product }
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -60,4 +66,29 @@ internal object PackageConfigurationFactory {
...
+            .asSequence()
+            .map { it.product }
+            .mapNotNull { product ->
+                product.pricePerMonth()?.let {
+                    Pair(product, it)
+                }
+            }
+            .maxByOrNull { it.second.amountMicros }
+            ?.second
+    }
```
vegaro, line 81: Can this be simplified to?

```
    private fun mostExpensivePricePerMonth(packages: List<Package>): Price? {
        return packages
            .mapNotNull { it.product.pricePerMonth() }
            .maxByOrNull { it.amountMicros }
    }
```

NachoSoto: Yes 🙌🏻 changed

PR #1375, opened 2023-10-19
Paywalls: Small optimization purchase button

tonidero:
### Description
This changes the animations for the purchase button to use `AnimatedVisibility` which actually removes the composable when the animation ends, making sure the progress indicator is not rendered without need, causing CPU usage

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
@@ -1,8 +1,11 @@
 package com.revenuecat.purchases.ui.revenuecatui.composables
 
-import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
```diff
@@ -86,33 +77,59 @@ private fun PurchaseButton(
...
 }
 
+@Composable
+private fun ButtonLabel(
+    shouldShow: Boolean,
+    selectedPackage: TemplateConfiguration.PackageInfo,
+    colors: TemplateConfiguration.Colors,
+) {
+    val localization = selectedPackage.localization
+    AnimatedVisibility(
```
NachoSoto, line 96: <3 this is so much better

NachoSoto (approved):

NachoSoto:
This is a huge find, thank you @tonidero! 🙏🏻

NachoSoto (changes requested):
Actually I think we need to use the existing implementation

I was in the process of changing `ConsistentPackageContentView` to use `AnimatedVisibility` as well, but I realized that it removes elements from the layout. It's important that we don't do that because we need them to be hidden and contribute to the overall layout so that package transitions are consistent and don't move content around.

tonidero:
Oh hmm then maybe we can use it only for the progress indicator, which was causing the issue? Will confirm tomorrow, thanks for catching that?

NachoSoto:
I noticed that the purchase button was changing height now because of that as well.
I'm trying to find another solution for this 🙏🏻  no worries though, we'll figure this out! Thanks for finding the source of the performance issue.

NachoSoto:
@tonidero new approach: https://github.com/RevenueCat/purchases-android/pull/1376

tonidero:
Closing in favor of #1376

PR #1376, opened 2023-10-19
`Paywalls`: optimize `PurchaseButton`

NachoSoto:
Extracted part of #1375. This uses `AnimatedVisibility` on the `CircularProgressIndicator` so that when it's not visible it's actually removed and the animation doesn't take CPU time. It uses `Modifier.matchParentSize` to ensure that when it's presented it doesn't contribute to the layout and abruptly resize the `PurchaseButton`.


tonidero (approved):
Makes sense, thanks for fixing that!

PR #1377, opened 2023-10-19
`Paywalls`: optimize `AdaptiveComposable`

NachoSoto:
This fixes the performance issue where it was constantly laying out, consuming about 10% CPU non-stop.

With this change, after the initial layout, CPU usage is 0% on a paywall.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/AdaptiveComposable.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/AdaptiveComposable.kt
@@ -27,14 +27,16 @@ internal fun RowScope.AdaptiveComposable(composables: List<@Composable () -> Uni
         modifier = Modifier
             .fillMaxWidth()
             .onGloballyPositioned { coordinates ->
-                maxSize = coordinates.size.width
+                if (coordinates.size.width != maxSize) {
+                    maxSize = coordinates.size.width
+                }
             }
             .align(Alignment.CenterVertically),
     ) {
         // Offscreen measurement
         Box(
             Modifier
-                .graphicsLayer(alpha = 0.2f)
+                .graphicsLayer(alpha = 0.0f)
                 .fillMaxWidth()
                 .wrapContentWidth(unbounded = true)
                 .align(Alignment.Center),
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/AdaptiveComposable.kt
```diff
@@ -27,23 +27,27 @@ internal fun RowScope.AdaptiveComposable(composables: List<@Composable () -> Uni
...
+                    maxSize = coordinates.size.width
+                }
             }
             .align(Alignment.CenterVertically),
     ) {
         // Offscreen measurement
         Box(
             Modifier
-                .graphicsLayer(alpha = 0.2f)
+                .graphicsLayer(alpha = 0.0f)
```
NachoSoto, line 39: This didn't have an impact, but I realized it was the wrong value.

NachoSoto:
@tonidero this fixes it :) thanks for pointing out the issue.

My only concern with this, is that we might need to reset `viewSizes` if something changes like locale or font settings? Or would compose recreate the composables for us?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/AdaptiveComposable.kt
```diff
@@ -27,23 +27,27 @@ internal fun RowScope.AdaptiveComposable(composables: List<@Composable () -> Uni
...
                 .align(Alignment.Center),
         ) {
             composables.forEachIndexed { index, view ->
                 Box(
                     Modifier.layout { measurable, _ ->
-                        val placeable = measurable.measure(Constraints())
-                        viewSizes[index] = placeable.width
+                        if (viewSizes[index] == 0) {
+                            viewSizes[index] = measurable.measure(Constraints()).width
+                        }
```
tonidero, line 49: As you said, I think this will cause us to not recompute the `viewSizes` when something changes (for example a locale change) since the remembered state will be kept unless the whole composable tree changes. I believe the recomposition was caused by setting a value in `viewSizes` right? If so, could we just modify this block with:
```suggestion
                        val measuredWidth =  measurable.measure(Constraints()).width
                        if (viewSizes[index] != measuredWidth) {
                            viewSizes[index] = measuredWidth
                        }
```

tonidero:
> My only concern with this, is that we might need to reset viewSizes if something changes like locale or font settings? Or would compose recreate the composables for us?

I tested it out on this branch (changing locale and font size) and seems we are recomposing it correctly. Since those changes are configuration changes, the compose tree is recreated, so the remembered state is lost. This could indeed be a problem if this state lived in the view model which survives configuration changes.

tonidero (approved):
Nice fix!

tonidero:
This was actually also causing the emerge tools snapshot tests to get stuck. Glad we could fix it! 🙌

PR #1378, opened 2023-10-19
`Paywalls`: `PurchaseButton` supports gradients

NachoSoto:
See also https://github.com/RevenueCat/purchases-ios/pull/3121

Screenshot 2023-10-19 at 15 31 42

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
@@ -4,6 +4,7 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
@@ -16,10 +17,14 @@ import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.ReadOnlyComposable
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.SolidColor
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
```diff
@@ -113,6 +123,19 @@ private fun PurchaseButton(
     }
 }
 
+@ReadOnlyComposable
```
NachoSoto, line 126: I _think_ this is correct? 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfiguration.kt
```diff
@@ -44,7 +44,7 @@ internal data class TemplateConfiguration(
         val text2: Color,
         val callToActionBackground: Color,
         val callToActionForeground: Color,
-        val callToActionSecondaryBackground: Color,
+        val callToActionSecondaryBackground: Color?,
```
NachoSoto, line 47: We can't default to the other color so we know if it was set or not.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
```diff
@@ -74,10 +79,15 @@ private fun PurchaseButton(
         )
 
         Button(
-            modifier = Modifier.fillMaxWidth(),
+            modifier = Modifier
+                .fillMaxWidth()
+                .background(
+                    brush = buttonBrush(colors),
+                    shape = ButtonDefaults.shape
```
NachoSoto, line 86: I love that Compose exposes this.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
tonidero: Yeah, since this composable doesn't modify the composable tree, I think it can be a readonly composable 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/ColorsFactory.kt
```diff
@@ -23,7 +22,7 @@ internal object ColorsFactory {
             text2 = Color(text2ColorInt),
             callToActionBackground = Color(callToActionBackgroundColorInt),
             callToActionForeground = Color(callToActionForegroundColorInt),
-            callToActionSecondaryBackground = Color(callToActionSecondaryBackgroundColorInt),
+            callToActionSecondaryBackground = callToActionSecondaryBackgroundColorInt?.let(::Color),
```
tonidero, line 25: Hmm do we need to change `callToActionSecondaryBackgroundColorInt` to not fallback to `callToActionBackgroundColorInt`? Otherwise, it will always have a value I think?

tonidero (approved):
Left a comment that I think we should address but looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/ColorsFactory.kt
NachoSoto: Yup I changed that above

PR #1379, opened 2023-10-20
`Paywalls`: optimize backwards compatible blurring

NachoSoto:
This accomplishes 2 things:
- Optimizes blurring algorithm by only blurring a smaller version of the image
- Fixes the blur not having a large enough radius (because of the `25` limit by `RenderScript`

### Before:
Blur before

### After:
Blur after


joshdholtz (approved):
Looks good to me 🤷‍♂️

PR #1380, opened 2023-10-20
`Paywalls`: fix backwards-compatible blurring of default paywall background

NachoSoto:
When displaying the default template and using `Image` we weren't using the `transformation` for backwards compatible blurring.

This adds a new `LocalImage` that reuses the same code as `RemoteImage` to be able to render images from a resource using Coil, which also has the benefit of asynchronous loading, caching, etc.

Screenshot 2023-10-20 at 13 12 29

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.ui.revenuecatui.composables
 
 import android.content.Context
+import androidx.annotation.DrawableRes
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


aboedo (approved):

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -24,10 +45,41 @@ internal fun RemoteImage(
...
+private sealed class ImageSource {
+    data class Local(@DrawableRes val resource: Int) : ImageSource() {
+        override val data: Any = resource
+    }
+    data class Remote(val urlString: String) : ImageSource() {
+        override val data: Any = urlString
+    }
+
+    abstract val data: Any
+}
```
vegaro, line 68: beautiful

vegaro (approved):

PR #1382, opened 2023-10-20
`Paywalls`: fixed incorrect background on footer

NachoSoto:
### Before:
Screenshot 2023-10-20 at 14 06 30

### After:
Rounded corner after

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -2,7 +2,6 @@ package com.revenuecat.purchases.ui.revenuecatui
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -69,7 +68,6 @@ private fun LoadedPaywall(state: PaywallState.Loaded, viewModel: PaywallViewMode
                         ),
                     )
                     .background(backgroundColor)
-                    .padding(top = UIConstant.footerRoundedBorderHeight)
```
NachoSoto, line 72: We can't add a padding here because it wouldn't apply the clip to the entire thing.

aboedo (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
tonidero: Hmm interesting, I'm not sure when that broke, but it did apply the clip to the padding before :man_shrugging:

tonidero (approved):

PR #1383, opened 2023-10-20
`Paywalls`: improve error handling

NachoSoto:
This makes `PaywallOptions.Builder` require a `dismissRequest`.
We can get away with not requiring it in iOS because we rely on `@Environment(\.dismiss)`.

This allows us to handle errors in an uniform way and always dismiss the paywall when something fails.

Note that this is only when `PaywallViewModel.state` becomes `Error`. Paywall validation simply logs errors and displays the default template.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -3,13 +3,19 @@ package com.revenuecat.purchases.ui.revenuecatui
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallState
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallViewModel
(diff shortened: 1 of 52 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -33,16 +36,20 @@ internal fun InternalPaywall(
...
+        viewModel.refreshStateIfColorsChanged(MaterialTheme.colorScheme)
 
         when (val state = viewModel.state.collectAsState().value) {
             is PaywallState.Loading -> {
                 LoadingPaywall(mode = options.mode)
             }
 
             is PaywallState.Error -> {
-                Text(text = "Error: ${state.errorMessage}")
+                LoadingPaywall(mode = options.mode)
```
NachoSoto, line 47: Displaying this so that when loading fails we don't make the content disappears.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -26,19 +26,23 @@ internal sealed class OfferingSelection {
 class PaywallOptions(builder: Builder) {
 
     internal val offeringSelection: OfferingSelection
-    val shouldDisplayDismissButton: Boolean
+    private val shouldDisplayDismissButton: Boolean
```
tonidero, line 29: This is a small API change... But I think it's probably ok since we're still on beta. We haven't implemented this yet as well.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -26,19 +26,23 @@ internal sealed class OfferingSelection {
...
         this.offeringSelection = builder.offeringSelection
         this.shouldDisplayDismissButton = builder.shouldDisplayDismissButton
         this.fontProvider = builder.fontProvider
         this.listener = builder.listener
+        this.dismissRequest = builder.dismissRequest
     }
 
-    class Builder {
+    class Builder(
+        internal val dismissRequest: () -> Unit,
```
tonidero, line 44: Considering this is now mandatory, I think we should remove the `onCloseButtonPressed` on the `PaywallListener`. Wdyt?

tonidero (approved):
Just a comment, but looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
vegaro: I agree, I think having both is confusing

NachoSoto: Ooh I missed that yeah

PR #1386, opened 2023-10-24
`Paywalls`: `PaywallDialogOptions` no longer requires `dismissRequest`

NachoSoto:
`PaywallDialog` works by presenting paywalls automatically based on a provided condition. Because of this, it should also know how to dismiss itself. I changed it to implement this automatically, but also provide a way to optionally pass a block for it.

Also updated the docs to reflect the new APIs: https://github.com/RevenueCat/revenuecat-docs/pull/450

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
@@ -26,13 +26,10 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.revenuecat.paywallstester.MainActivity
-import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.Offerings
-import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.ui.revenuecatui.PaywallDialog
 import com.revenuecat.purchases.ui.revenuecatui.PaywallDialogOptions
-import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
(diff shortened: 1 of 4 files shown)


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -124,17 +121,9 @@ private fun OfferingsListScreen(
...
-                dismissRequest = {
-                    displayPaywallDialogOffering = null
-                },
-            )
+            PaywallDialogOptions.Builder()
+                .setDismissRequest { displayPaywallDialogOffering = null }
                 .setOffering(displayPaywallDialogOffering)
-                .setListener(object : PaywallListener {
-                    override fun onPurchaseCompleted(customerInfo: CustomerInfo, storeTransaction: StoreTransaction) {
-                        displayPaywallDialogOffering = null
```
NachoSoto, line 135: This is done by `PaywallViewModelImpl` now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -38,24 +38,26 @@ fun PaywallDialog(
         }
     }
     if (shouldDisplayDialog) {
```
tonidero, line 40: I'm thinking, since we are going to be handling the visibility ourselves, we should remember this parameter through configuration changes. For example, when going between portrait/landscape, changing dark mode or locale,...

I haven't tested it but if I'm not wrong, currently, if we display the paywall, the user dismisses it, and then there is one of those configuration changes, it would reappear. I think we probably want to avoid that by making that state survive configuration changes. One of the options is to use [rememberSaveable](https://developer.android.com/jetpack/compose/state#state-in-composables), instead of `remember`, that's probably the quickest. The other would be moving this state to a viewModel, which might be harder in this case since we don't have a viewModel for the dialog currently.

tonidero (commented):
I only have that concern. We should test to confirm my theory but other than that, this looks good.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
NachoSoto: That's a really good point. I just read about the difference and that makes sense 👍🏻 
I just changed this.

tonidero (approved):
Looks great!

PR #1387, opened 2023-10-24
`Paywalls`: support for Markdown links

NachoSoto:

PR #1388, opened 2023-10-24
`Paywalls`: use `LocalUriHandler` for opening links

NachoSoto:
Follow up to #1387.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
@@ -26,6 +26,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalUriHandler
 import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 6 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -70,7 +70,7 @@ private fun Footer(
     childModifier: Modifier = Modifier,
     allPlansTapped: (() -> Unit)? = null,
 ) {
-    val context = LocalContext.current
+    val uriHandler = LocalUriHandler.current
```
tonidero, line 73: Oh nice! Yeah, I didn't know this existed, makes sense 👍

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -152,11 +147,6 @@ internal class PaywallViewModelImpl(
         }
     }
 
-    override fun openURL(url: URL, context: Context) {
```
tonidero, line 155: There is one downside of moving this outside the viewModel, which is that if we want to have events or some other logic when navigating to links, having it as part of the compose tree might not be as easy to test or handle... But we can do that later if we have to.

PR #1392, opened 2023-10-25
`Paywalls`: fixed accessibility across templates

NachoSoto:
### Improvements:
- `PurchaseButton` is now accessible and will only read the relevant text
- `Footer` is fully accessible and will read the largest text even if it's not displayed
- `Template2` provides context on what package is selected
- `Template3` combines feature elements into a single accessibility element

https://github.com/RevenueCat/purchases-android/assets/685609/77bae51f-0f52-4c57-b642-62aefa3fc8b7


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -173,11 +179,18 @@ private fun RowScope.Button(
         TextButton(
             onClick = action::invoke,
             contentPadding = PaddingValues(4.dp),
-            modifier = Modifier.align(CenterHorizontally),
+            modifier = Modifier
+                .align(CenterHorizontally)
+                .semantics(mergeDescendants = true) {
```
tonidero, line 184: 👍

tonidero (approved):
Love it!

PR #1394, opened 2023-10-26
`Paywalls`: fix template 3 offer details color

NachoSoto:


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/OfferDetails.kt
```diff
@@ -20,6 +22,7 @@ import com.revenuecat.purchases.ui.revenuecatui.extensions.introEligibility
 @Composable
 internal fun OfferDetails(
     state: PaywallState.Loaded,
+    color: (TemplateConfiguration.Colors) -> Color = { it.text1 }
```
tonidero, line 25: Hmm well, I usually prefer to avoid lambdas for composables whenever possible since it avoids using compose optimizations to avoid recomposing compose whenever the parameters are immutable/stable... 

Having said that this is probably ok I think. Shouldn't cause any performance issues

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/OfferDetails.kt
NachoSoto: Hmm yeah let me change this.

PR #1397, opened 2023-10-26
`Paywalls`: fix `Modifier.conditional`

NachoSoto:


NachoSoto:
Oh wait I guess this was right, and it's easier to use this way.

PR #1401, opened 2023-10-26
`Paywalls`: convert empty strings to `null`

NachoSoto:
Equivalent to https://github.com/RevenueCat/purchases-ios/pull/2818

`IntroEligibilityStateView` relies on strings being `null` to determine what string is available, therefore we need to make sure we don't have empty strings in `PaywallData`.

### DIFF
--- a/purchases/src/test/resources/paywalldata-sample1.json
+++ b/purchases/src/test/resources/paywalldata-sample1.json
@@ -27,7 +27,9 @@
     "es_ES": {
       "title": "Tienda",
       "call_to_action": "Comprar",
+      "call_to_action_with_multiple_intro_offers": " ",
       "offer_details_with_intro_offer": " ",
+      "offer_details_with_multiple_intro_offers": "",
       "offer_name": "{{ period }}",
       "features": [
         {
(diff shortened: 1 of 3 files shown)


NachoSoto:
Bump @vegaro @tonidero

purchases/src/test/resources/paywalldata-sample1.json
```diff
@@ -27,7 +27,9 @@
     "es_ES": {
       "title": "Tienda",
       "call_to_action": "Comprar",
+      "call_to_action_with_multiple_intro_offers": " ",
       "offer_details_with_intro_offer": " ",
+      "offer_details_with_multiple_intro_offers": "",
       "offer_name": "{{ period }}",
       "features": [
         {
```
joshdholtz, line 35: Do we want to add a `"subtitle": ""` in here to test the new `EmptyStringToNullSerializer` on `subtitle`?

joshdholtz (changes requested):

purchases/src/test/resources/paywalldata-sample1.json
NachoSoto: I wanted to test that missing fields decode as null still instead of crashing as well

joshdholtz (approved):
LGTM 🚀

tonidero (approved):

PR #1403, opened 2023-10-27
Fix navigation after closing paywall in Paywalls screen

vegaro:
@aboedo noticed that when navigating to the Template 2 full screen from the Paywalls tab, going back, then navigating again was not working. 

@tonidero took a very quick look and seems that we are not calling the PaywallDialog onDismiss callback when going back, so the PaywallsScreen thinks that there is still a paywall displaying so it doesn't work.

I added the missing call to the `dismissRequest` callback

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -38,7 +38,10 @@ fun PaywallDialog(
         }
     }
     if (shouldDisplayDialog) {
-        val dismissRequest = { shouldDisplayDialog = false }
+        val dismissRequest = {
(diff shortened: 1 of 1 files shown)


tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -38,7 +38,10 @@ fun PaywallDialog(
...
     }
     if (shouldDisplayDialog) {
-        val dismissRequest = { shouldDisplayDialog = false }
+        val dismissRequest = {
+            paywallDialogOptions.dismissRequest?.invoke()
+            shouldDisplayDialog = false
+        }
 
         Dialog(
             onDismissRequest = dismissRequest,
```
NachoSoto, line 47: It makes sense to pass this here, but this is also it makes sense to pass this here, but this is modifying line 50 too

PR #1408, opened 2023-10-27
Fix onDismiss getting called twice in PaywallDialogs

vegaro:
We call `paywallDialogOptions.dismissRequest?.invoke()` inside `toPaywallOptions` so since we were passing 

```
val dismissRequest = {
            paywallDialogOptions.dismissRequest?.invoke()
            shouldDisplayDialog = false
}
```

the `paywallDialogOptions.dismissRequest?.invoke()` piece was getting called twice when calling `paywallDialogOptions.toPaywallOptions(dismissRequest)`

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -39,16 +39,18 @@ fun PaywallDialog(
         }
     }
     if (shouldDisplayDialog) {
-        val dismissRequest = {
-            paywallDialogOptions.dismissRequest?.invoke()
-            shouldDisplayDialog = false
-        }
-
         Dialog(
-            onDismissRequest = dismissRequest,
+            onDismissRequest = {
+                shouldDisplayDialog = false
+                paywallDialogOptions.dismissRequest?.invoke()
+            },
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -39,16 +39,18 @@ fun PaywallDialog(
...
+                shouldDisplayDialog = false
+                paywallDialogOptions.dismissRequest?.invoke()
+            },
             properties = DialogProperties(usePlatformDefaultWidth = shouldUsePlatformDefaultWidth()),
         ) {
-            DialogScaffold(paywallDialogOptions.toPaywallOptions(dismissRequest))
+            DialogScaffold(
+                paywallDialogOptions.toPaywallOptions(dismissRequest = {
+                    shouldDisplayDialog = false
+                }),
```
NachoSoto, line 52: I think this would be simpler keeping the `onDismissRequest` like it was before #1403:
```kotlin
dismissRequest = {
  shouldDisplayDialog = false
}
```

Then you can pass it here, and for the `Dialog` above just do:
```kotlin
onDismissRequest = {
  dismissRequest()
  paywallDialogOptions.dismissRequest?.invoke()
}
```

What do you think?

NachoSoto:
Thanks for fixing this 👍🏻 just one suggestion to simplify the code.

NachoSoto (approved):

PR #1411, opened 2023-10-27
`Paywalls`: fix `MarkdownText` stealing touches from package buttons

NachoSoto:


tonidero (approved):
Ohh makes sense. Thanks for fixing it!

PR #1413, opened 2023-10-27
`Paywalls`: footer view should not render background image

NachoSoto:
Unless it needs to be blur, to match iOS behavior.

Thanks to @lburdock for finding this discrepancy!


tonidero (approved):
Ohh makes sense 👍

PR #1416, opened 2023-10-30
Paywalls: Fix purchase button text flashing when changing selected package

tonidero:
### Description
In some cases, the purchase button text doesn't change when selecting different packages. However, we were still animating the text resulting in a small flash in the purchase button when changing the selected package.

This makes it so, if the purchase button text doesn't change between all the packages, we don't animate it.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/ConsistentPackageContentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/ConsistentPackageContentView.kt
@@ -31,15 +31,20 @@ internal fun ConsistentPackageContentView(
 internal fun ConsistentPackageContentView(
     packages: List<TemplateConfiguration.PackageInfo>,
     selected: TemplateConfiguration.PackageInfo,
+    shouldAnimate: Boolean = true,
     creator: @Composable (TemplateConfiguration.PackageInfo) -> Unit,
 ) {
     Box {
         packages.forEach { packageItem ->
-            val opacity by animateFloatAsState(
-                targetValue = if (packageItem.rcPackage == selected.rcPackage) 1.0f else 0.0f,
-                animationSpec = UIConstant.defaultAnimation(),
-                label = "ConsistentPackageContentView",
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/ConsistentPackageContentView.kt
```diff
@@ -31,15 +31,20 @@ internal fun ConsistentPackageContentView(
 internal fun ConsistentPackageContentView(
     packages: List<TemplateConfiguration.PackageInfo>,
     selected: TemplateConfiguration.PackageInfo,
+    shouldAnimate: Boolean = true,
```
tonidero, line 34: I thought about changing the implementation to just have a single text view instead of multiple as we have now, but that seemed a more complex implementation... So I decided to just disable the animation.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
```diff
@@ -183,3 +184,27 @@ private fun PurchaseButtonPreview() {
...
+    val firstText = with(all.first()) {
+        introEligibilityText(
+            introEligibility,
+            localization.callToAction,
+            localization.callToActionWithIntroOffer,
+            localization.callToActionWithMultipleIntroOffers,
+        )
+    }
+
+    all.forEach {
```
NachoSoto, line 198: Maybe skip the first?

NachoSoto (approved):
Yessss 👏🏻🙌🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
tonidero: Done in https://www.npmjs.com/package/@revenuecat/purchases-capacitor#setloglevel

PR #1417, opened 2023-10-30
Paywalls: Fix wrong view model data after configuration changes

tonidero:
### Description
I noticed that after a configuration change (for example moving from landscape to portrait), the close button didn't work as expected on the dialogs. I tracked it down to the fact that the view model kept the first version of the dismiss callback, so it wasn't calling the new one created after the configuration change.

I decided to remove the current approach of using keys to create individual view models and instead I'm adding a method to the view model that gets called every time we try to get the view model. If we detect that the PaywallOptions have changed, we will update it on the view model, ensuring we have the latest version of the callbacks.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -60,6 +60,30 @@ class PaywallOptions(builder: Builder) {
...
+        if (shouldDisplayDismissButton != other.shouldDisplayDismissButton) return false
+        if (fontProvider != other.fontProvider) return false
+        if (listener != other.listener) return false
+        if (mode != other.mode) return false
+        if (dismissRequest != other.dismissRequest) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
```
NachoSoto, line 77: If we made `PaywallOptions` a `data class` would we get this for free?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -87,6 +87,13 @@ internal class PaywallViewModelImpl(
         updateState()
     }
 
+    fun updateOptions(options: PaywallOptions) {
+        if (this.options != options) {
+            this.options = options
+            updateState()
```
NachoSoto, line 93: `updateState` only uses `options.offeringSelection`, so it seems silly to  fetch everything if we only change something else like `shouldDisplayDismissButton`.

Suggestion: compare only `options.offeringSelection` here, but the risk is that if `updateState` uses something else in the future, it would break.
So to deal with that, we can extract a `createState` method like `suspend createState(offeringSelection: OfferingSelection): PaywallState`.

Then make `updateState` simply do:
```kotlin
private fun updateState() {
    viewModelScope.launch {
         _state.value = createState(options.offeringSelection)
    }
}
```

And we can do the exact same here, therefore ensuring that if `createState` ever needs more data from `options`, that's explicit here.

What do you think?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
NachoSoto: Though my other suggestion might remove the need for this too.

NachoSoto (commented):
Thanks for fixing this! I have a suggestion to simplify this.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: Actually, we also use the `mode` (it's masked behind a computed property in `PaywallViewModelImpl`), and `options.shouldDisplayDismissButton` in `calculateState`. So we would need those 3.

The other fields are the dismissRequest callback, the listener and the fontProvider... So I kinda think it's worth to just compare everything and if anything changes, just recompute the state... Wdyt?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
tonidero: > If we made PaywallOptions a data class would we get this for free?

I was trying to avoid it, since it's not recommended to expose data classes... But honestly, it's reached a point where things are much easier, so I think I will do the change yeah.

tonidero: Made the change to data classes https://github.com/RevenueCat/purchases-android/pull/1417/commits/e262e9af760c79fd29776ccd00700132bdf61852

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
NachoSoto: Oh that's right. This is fine then :+1:

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
NachoSoto: 👍🏻 
Why is it not recommended?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialogOptions.kt
```diff
@@ -6,23 +6,23 @@ import com.revenuecat.purchases.ui.revenuecatui.fonts.FontProvider
 import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayBlockForEntitlementIdentifier
 
 @ExperimentalPreviewRevenueCatUIPurchasesAPI
-class PaywallDialogOptions(builder: Builder) {
+data class PaywallDialogOptions internal constructor(
```
NachoSoto, line 9: Much better!

NachoSoto (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
tonidero: Mentioned it over slack but this article has some of the drawbacks: https://jakewharton.com/public-api-challenges-in-kotlin/

PR #1419, opened 2023-10-31
Paywalls: Improve PaywallDialog look on tablets

tonidero:
### Description
I've been testing the paywall dialog in tablets. I noticed it mostly looks fine on landscape, but not so great on portrait. So I added a max height to the dialog so it was not so stretched

| Before | After |
|-------|-------|
| image | image |
| image | image |
| | image |
| | image |
| | image |
| | image |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -1,7 +1,10 @@
 package com.revenuecat.purchases.ui.revenuecatui
 
+import android.content.res.Configuration
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -12,13 +15,18 @@ import androidx.compose.runtime.mutableStateOf
...
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.window.Dialog
 import androidx.compose.ui.window.DialogProperties
 import androidx.window.core.layout.WindowWidthSizeClass
 import com.revenuecat.purchases.ui.revenuecatui.helpers.computeWindowWidthSizeClass
 import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayPaywall
 import kotlinx.coroutines.launch
 
+private object UIDialogConstants {
+    const val MAX_HEIGHT_PERCENTAGE_TABLET = 0.85f
```
NachoSoto, line 27: Nice!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -69,6 +81,21 @@ private fun DialogScaffold(paywallOptions: PaywallOptions) {
     }
 }
 
+@Composable
+@ReadOnlyComposable
+private fun getDialogMaxHeightPercentage(): Float {
+    val orientation = LocalConfiguration.current.orientation
+    if (orientation == Configuration.ORIENTATION_LANDSCAPE) {
+        return 1f
+    }
```
NachoSoto, line 90: I see what you're trying to do here. I wonder if a better approach to this to also work with multi-tasking (side by side apps, whatever Android calls it) is to check the application window aspect ratio?

Though I guess even on that case this implementation makes sense.

NachoSoto (approved):
<3

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
tonidero: Right, that's a good point... With the current implementation, when using multitasking, if the view is mostly square, the dialog won't use the full height, but it would look better if it did. Will change that 👍

tonidero: After the change:
image

vegaro (approved):

PR #1426, opened 2023-11-01
`StoreProduct`: new `pricePerWeek` and `pricePerYear`

NachoSoto:
This brings parity with `purchases-ios` and will be used for a new paywalls variable.

### DIFF
--- a/api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
+++ b/api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
@@ -28,7 +28,9 @@ final class StoreProductAPI {
         final String description = product.getDescription();
         final Period period = product.getPeriod();
         final String formattedPricePerMonth = product.formattedPricePerMonth(locale);
+        final Price pricePerWeek = product.pricePerWeek(locale);
         final Price pricePerMonth = product.pricePerMonth(locale);
+        final Price pricePerYear = product.pricePerYear(locale);
 
         SubscriptionOptions subscriptionOptions = product.getSubscriptionOptions();
         SubscriptionOption defaultOption = product.getDefaultOption();
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreProduct.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreProduct.kt
@@ -2,6 +2,8 @@ package com.revenuecat.purchases.models
 
 import com.revenuecat.purchases.ProductType
 import com.revenuecat.purchases.utils.pricePerMonth
+import com.revenuecat.purchases.utils.pricePerWeek
+import com.revenuecat.purchases.utils.pricePerYear
 import java.util.Locale
 
 /**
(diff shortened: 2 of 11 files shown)


purchases/src/test/java/com/revenuecat/purchases/utils/PriceExtensionsPricePerMonthTest.kt
```diff
@@ -0,0 +1,42 @@
...
+import junitparams.Parameters
+import org.junit.Test
+import org.junit.experimental.categories.Category
+import org.junit.runner.RunWith
+
+@RunWith(JUnitParamsRunner::class)
+@Category(AndroidJUnit4::class)
+class PriceExtensionsPricePerMonthTest: PriceExtensionsPricePerPeriodTest() {
+    @Test
+    @Parameters(
```
NachoSoto, line 16: Parameterized tests FTW. It allows us to write repeated tests without duplicating code, and they work great with the IDE:

Screenshot 2023-11-01 at 09 46 20

api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
```diff
@@ -28,7 +28,9 @@ static void check(final StoreProduct product) {
         final String description = product.getDescription();
         final Period period = product.getPeriod();
         final String formattedPricePerMonth = product.formattedPricePerMonth(locale);
+        final Price pricePerWeek = product.pricePerWeek(locale);
         final Price pricePerMonth = product.pricePerMonth(locale);
+        final Price pricePerYear = product.pricePerYear(locale);
```
tonidero, line 33: We should also add versions without the optional parameter to these (I see we didn't have them for the `pricePerMonth` and `formattedPricePerMonth`...), same in the Kotlin API tests.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreProduct.kt
```diff
@@ -91,6 +93,18 @@ interface StoreProduct {
...
 
+    /**
+     * Null for INAPP products. The price of the [StoreProduct] in the given locale in a weekly recurrence.
+     * This means that, for example, if the period is monthly, the price will be divided by 4.
+     * It uses a currency formatter to format the price in the given locale.
+     * Note that this value may be an approximation.
+     * For Google subscriptions, this value will use the basePlan to calculate the value.
+     * @param locale Locale to use for formatting the price. Default is the system default locale.
+     */
+    fun pricePerWeek(locale: Locale = Locale.getDefault()): Price? {
```
tonidero, line 104: We should add a `@JvmOverloads` here I think, so thee default value works in Java. Same in `pricePerYear`

purchases/src/test/java/com/revenuecat/purchases/utils/PriceExtensionsPricePerMonthTest.kt
tonidero: Nice! yeah, I once tried to update to Juni5 to use them, but they were not compatible with Robolectric, which we use in a bunch of our tests... But adding a library like this for tests makes sense 👍

tonidero (approved):
Looks great!

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreProduct.kt
NachoSoto: Oh it's because `@JvmOverloads` isn't allowed on `interface` methods.

NachoSoto: We can add those explicitly overloaded in a separate PR.

PR #1433, opened 2023-11-03
Fix RevenueCatUI API issues and add API tests 

tonidero:
### Description
This adds API tests for the paywalls module. In the process of writing those API tests I found 2 issues that I wanted to get fixed while the API is experimental:
- The `requiredEntitlementIdentifier` parameter in the `PaywallActivityLauncher.launchIfNeeded` method was after some optional parameters, making it mandatory to use named parameters if you didn't want to pass the other optional parameters.
- In Java, it wasn't possible to create a `PaywallFont` since it was using value classes that are not available in Java.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -73,9 +73,9 @@ class PaywallActivityLauncher {
      */
     @JvmOverloads
     fun launchIfNeeded(
+        requiredEntitlementIdentifier: String,
         offering: Offering? = null,
         fontProvider: ParcelizableFontProvider? = null,
-        requiredEntitlementIdentifier: String,
```
tonidero, line 78: Honestly I wasn't sure about doing this now, but I think it's better now that the API is still experimental. If we don't put this mandatory parameter first, users are forced to use named parameters if they only want to set the `requiredEntitlementIdentifier` but not the optional parameters.

api-tester/build.gradle
```diff
@@ -17,7 +17,7 @@ android {
     }
 
     defaultConfig {
-        minSdkVersion 21 // Compose requires minSdkVersion 21
+        minSdkVersion 24 // RevenueCat UI requires 24
```
tonidero, line 20: Had to update the api tests in order to test the revenuecatui module. I think it should be fine, since our API shouldn't depend on the sdk version

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fonts/PaywallFont.kt
```diff
@@ -33,9 +33,9 @@ sealed class PaywallFont : Parcelable {
         val fontWeight: FontWeight = FontWeight.Normal,
         /**
          * The style of the font, normal or italic. The system uses this to match a font to a font request.
+         * We use int instead of [FontStyle] because [FontStyle] is not compatible with Java.
          */
-        @TypeParceler<FontStyle, FontStyleParceler>()
-        val fontStyle: FontStyle = FontStyle.Normal,
+        val fontStyle: Int = FontStyle.Normal.value,
```
tonidero, line 38: This is another breaking change in kotlin... Basically I'm doing this since it made it impossible to create instances of these classes in Java, since `FontStyle` is not available in Java.

api-tester/build.gradle
NachoSoto: Hmm but this helped guarantee that we didn't break this compatibility. Should we create a separate module for `revenuecatui` API tests?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
NachoSoto: Yes, better now than never! I don't think there are that many users yet so let's fix these things while we have time.

NachoSoto (commented):
Awesome thanks for doing this! I figured we'd catch little things like that <3

Just one suggestion, but let me know if that's more trouble than it's worth.

api-tester/build.gradle
tonidero: Hmm I see your point, but it's difficult to change the minSdkVersion accidentally 🤞, and the main sdk shouldn't compile already if there is any issue that would be caught by the API tester. So I think it's ok to keep it in the same module? cc @vegaro. In case you have other thoughts.

NachoSoto (approved):

PR #1435, opened 2023-11-03
Improve fullscreen templates in landscape orientation

tonidero:
### Description
This modifies templates 1 and 5 so the top images don't use the full height of the screen when using devices in horizontal. This is not a problem when presenting as a dialog, since the width is limited, but if using the composables directly, it might be problematic.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -14,12 +14,12 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.window.Dialog
 import androidx.compose.ui.window.DialogProperties
 import androidx.window.core.layout.WindowWidthSizeClass
 import com.revenuecat.purchases.ui.revenuecatui.helpers.computeWindowWidthSizeClass
 import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayPaywall
+import com.revenuecat.purchases.ui.revenuecatui.helpers.windowAspectRatio
 import kotlinx.coroutines.launch
 
 private object UIDialogConstants {
@@ -84,8 +84,7 @@ private fun DialogScaffold(paywallOptions: PaywallOptions) {
 @Composable
 @ReadOnlyComposable
 private fun getDialogMaxHeightPercentage(): Float {
-    val aspectRatio = LocalConfiguration.current.screenHeightDp.toFloat() / LocalConfiguration.current.screenWidthDp
-    if (aspectRatio < UIDialogConstants.MAX_ASPECT_RATIO_TO_APPLY_MAX_HEIGHT) {
+    if (windowAspectRatio() < UIDialogConstants.MAX_ASPECT_RATIO_TO_APPLY_MAX_HEIGHT) {
         return 1f
     }
     return computeWindowWidthSizeClass().let {
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
@@ -2,5 +2,5 @@ package com.revenuecat.purchases.ui.revenuecatui.extensions
 
 import androidx.compose.ui.unit.IntSize
 
-val IntSize.aspectRatio: Float
+internal val IntSize.aspectRatio: Float
     get() = width.toFloat() / height
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
@@ -1,0 +1,6 @@
+package com.revenuecat.purchases.ui.revenuecatui.extensions
+
+import androidx.compose.ui.unit.IntSize
+
+val IntSize.aspectRatio: Float
+    get() = width.toFloat() / height
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/HelperFunctions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/HelperFunctions.kt
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.ui.revenuecatui.helpers
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.ReadOnlyComposable
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.LocalInspectionMode
 import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.Purchases
@@ -13,6 +14,12 @@ import kotlin.coroutines.suspendCoroutine
 @Composable
 @ReadOnlyComposable
 internal fun isInPreviewMode() = LocalInspectionMode.current
+
+@Composable
+@ReadOnlyComposable
+internal fun windowAspectRatio(): Float {
+    return LocalConfiguration.current.screenHeightDp.toFloat() / LocalConfiguration.current.screenWidthDp
+}
 
 /**
  * Evaluates [shouldDisplayBlock] with the current CustomerInfo to determine if a paywall should be displayed.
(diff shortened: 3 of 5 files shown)


tonidero:
Screenshots change in: https://github.com/RevenueCat/purchases-android-snapshots/pull/3. Will hold this until the base PR is ready. snapshot tests will fail until that PR is merged.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -130,10 +134,17 @@ private fun ColumnScope.Template1MainContent(state: PaywallState.Loaded) {
...
+            val aspectRatio = windowAspectRatio()
+            val screenHeight = LocalConfiguration.current.screenHeightDp
             RemoteImage(
                 urlString = uri.toString(),
                 modifier = Modifier
-                    .aspectRatio(ratio = 1.2f),
+                    .conditional(aspectRatio > 1f) {
+                        aspectRatio(ratio = 1.2f)
+                    }
+                    .conditional(aspectRatio <= 1f) {
```
NachoSoto, line 145: Oooh so nice

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -170,10 +173,16 @@ private fun ColumnScope.Template5MainContent(
 @Composable
 private fun HeaderImage(uri: Uri?) {
     uri?.let {
+        val aspectRatio = windowAspectRatio()
```
NachoSoto, line 176: What if the paywall isn't displayed full screen and is instead a dialog?

NachoSoto (commented):
Awesome! Just one question.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
tonidero: Hmm I was assuming this would pick the window size of the dialog, not the screen, but I was wrong, 
so this change would still apply there. As long as the width of the screen is bigger than the height, we will set the image as a percentage of the screen, instead of using an aspect ratio. This is not ideal, so trying to figure out a solution. Great catch! Thanks!

tonidero: I changed the algorithm to use the template size instead of the window size, so we don't care if it's displayed as a dialog, as full screen, or anything the developer wants to do. This will cause a recomposition unfortunately... But I think it's the best solution

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -127,13 +138,20 @@ private fun ColumnScope.Template1MainContent(state: PaywallState.Loaded) {
...
+                        fillMaxWidth().height(screenHeight.dp / 2f)
```
vegaro, line 153: Does this keep the aspect ratio? The `fillMaxWidth()` will fill the parent and the height will be reduced? So a portrait image will look squashed right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -168,12 +176,18 @@ private fun ColumnScope.Template5MainContent(
...
+        val aspectRatio = template5Size.height.toFloat() / template5Size.width.toFloat()
         RemoteImage(
             urlString = uri.toString(),
             modifier = Modifier
-                .aspectRatio(ratio = Template5UIConstants.headerAspectRatio),
+                .conditional(aspectRatio > 1f) {
+                    aspectRatio(ratio = Template5UIConstants.headerAspectRatio)
+                }
+                .conditional(aspectRatio <= 1f) {
+                    fillMaxHeight(Template5UIConstants.percentageScreenImageInLandscape).fillMaxWidth()
```
vegaro, line 189: same question here about the aspect ratio

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
tonidero: No it won't. However, we are using `ContentScale.Crop` here, so the image won't look squashed but it will get cut, so depending on the image it might not look great.

I agree, this might not look great in some scenarios, but usually, this shouldn't be a problem as long as the width is not much larger than the height. So, for developers presenting as a dialog, this shouldn't be an issue I think (the dialog will not use the full available width in these scenarios.

Do you have any suggestions to handle this instead? This is the screenshots change after this: https://github.com/RevenueCat/purchases-android-snapshots/pull/3

vegaro: I was missing the `ContentScale.Crop` 😄 

I was worried the image would look squashed, but it won't, so this should be fine

vegaro (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/HelperFunctions.kt
```diff
@@ -14,6 +15,12 @@ import kotlin.coroutines.suspendCoroutine
 @ReadOnlyComposable
 internal fun isInPreviewMode() = LocalInspectionMode.current
 
+@Composable
+@ReadOnlyComposable
+internal fun windowAspectRatio(): Float {
+    return LocalConfiguration.current.screenHeightDp.toFloat() / LocalConfiguration.current.screenWidthDp
```
NachoSoto, line 21: Nice: maybe extract `LocalConfiguration.current` in a `val` here to avoid calling it twice?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -84,8 +84,7 @@ private fun DialogScaffold(paywallOptions: PaywallOptions) {
 @Composable
 @ReadOnlyComposable
 private fun getDialogMaxHeightPercentage(): Float {
-    val aspectRatio = LocalConfiguration.current.screenHeightDp.toFloat() / LocalConfiguration.current.screenWidthDp
-    if (aspectRatio < UIDialogConstants.MAX_ASPECT_RATIO_TO_APPLY_MAX_HEIGHT) {
+    if (windowAspectRatio() < UIDialogConstants.MAX_ASPECT_RATIO_TO_APPLY_MAX_HEIGHT) {
```
NachoSoto, line 87: > I changed the algorithm to use the template size instead of the window size, so we don't care if it's displayed as a dialog, as full screen, or anything the developer wants to do. This will cause a recomposition unfortunately... But I think it's the best solution

Isn't this still the window size?

tonidero: Yeah, but in this case, we can indeed use the window size I think, since it's what's going to determine whether we use a full height dialog or not, so in this case, we do know it's going to be displayed as a dialog, and we already have some logic so we don't display it as a full screen dialog for MEDIUM and EXPANDED devices.

Also note that this is just a refactor, but the logic hasn't changed in this PR.

NachoSoto: Oooh right right 👍🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -127,13 +138,20 @@ private fun ColumnScope.Template1MainContent(state: PaywallState.Loaded) {
 }
 
 @Composable
-private fun HeaderImage(uri: Uri?) {
+private fun HeaderImage(uri: Uri?, template1Size: IntSize) {
```
NachoSoto, line 141: Maybe 
```suggestion
private fun HeaderImage(uri: Uri?, templateSize: IntSize) {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -168,12 +176,18 @@ private fun ColumnScope.Template5MainContent(
 }
 
 @Composable
-private fun HeaderImage(uri: Uri?) {
+private fun HeaderImage(uri: Uri?, template5Size: IntSize) {
     uri?.let {
+        val aspectRatio = template5Size.height.toFloat() / template5Size.width.toFloat()
```
NachoSoto, line 181: Maybe we can add an `val IntSize.aspectRatio` extension?

NachoSoto (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
```diff
@@ -0,0 +1,6 @@
+package com.revenuecat.purchases.ui.revenuecatui.extensions
+
+import androidx.compose.ui.unit.IntSize
+
+val IntSize.aspectRatio: Float
```
NachoSoto, line 5: `internal`?

PR #1436, opened 2023-11-06
Paywalls: Store paywall events on disk and API (1)

tonidero:
### Description
This adds support for storing paywall events into disk. In followup PRs, we will send these events to our servers for analytics and use these events to indicate whether a purchase came from our paywalls.
- [ ] Send events to server
- [ ] Use events to detect whether a purchase came from paywall.

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -21,6 +21,7 @@ import com.revenuecat.purchases.models.BillingFeature
 import com.revenuecat.purchases.models.GoogleProrationMode
 import com.revenuecat.purchases.models.InAppMessageType
 import com.revenuecat.purchases.models.StoreProduct
+import com.revenuecat.purchases.paywalls.events.PaywallEvent
 import com.revenuecat.purchases.strings.BillingStrings
 import com.revenuecat.purchases.strings.ConfigureStrings
 import java.net.URL
@@ -448,6 +449,15 @@ class Purchases internal constructor(
      */
     fun invalidateCustomerInfoCache() {
         purchasesOrchestrator.invalidateCustomerInfoCache()
+    }
+
+    /**
+     * Used by `RevenueCatUI` to keep track of [PaywallEvent]s.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    @JvmSynthetic
+    fun track(paywallEvent: PaywallEvent) {
+        purchasesOrchestrator.track(paywallEvent)
     }
 
     // region Subscriber Attributes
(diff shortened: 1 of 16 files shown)


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -450,6 +451,15 @@ class Purchases internal constructor(
         purchasesOrchestrator.invalidateCustomerInfoCache()
     }
 
+    /**
+     * Used by `RevenueCatUI` to keep track of [PaywallEvent]s.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    @JvmSynthetic
```
tonidero, line 458: Made this synthetic since this is really only meant for our use in `RevenueCatUI` which uses Kotlin.

NachoSoto: TIL but makes sense.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -252,12 +255,30 @@ internal class PurchasesFactory(
...
             return Purchases(purchasesOrchestrator)
         }
     }
 
+    private fun createPaywallEventsManager(
+        context: Context,
+        identityManager: IdentityManager,
+        eventsDispatcher: Dispatcher,
+    ): PaywallEventsManager? {
+        return if (isAndroidNOrNewer()) {
```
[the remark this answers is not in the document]
NachoSoto: Maybe add a comment mentioning that?

purchases/src/test/java/com/revenuecat/purchases/paywalls/events/PaywallEventSerializationTests.kt
```diff
@@ -0,0 +1,67 @@
...
+            ),
+            type = PaywallEventType.IMPRESSION,
+        ),
+        userID = "testAppUserId",
+    )
+
+    @Test
+    fun `can encode paywall event correctly`() {
+        val eventString = PaywallEventsManager.json.encodeToString(event)
+        assertThat(eventString).isEqualTo(
```
NachoSoto, line 39: I imagine you compared these with iOS 👍🏻

NachoSoto (approved):
Awesome!

purchases/src/test/java/com/revenuecat/purchases/paywalls/events/PaywallEventSerializationTests.kt
tonidero: These are different than in iOS, but it's the way they are stored in disk, so it shouldn't matter if they are different I think. In iOS, the encoding is pretty different, but all the data is there.

NachoSoto: Oh yeah of course never mind

PR #1437, opened 2023-11-06
`Offering`: restore constructor with no `PaywallData`

NachoSoto:
We were missing the constructors in the API tester so we didn't realize that we removed the previous constructor with no `PaywallData`.
This was caught by PHC: https://github.com/RevenueCat/purchases-hybrid-common/pull/544


tonidero (approved):
Ahh right. Thanks for fixing it!

PR #1442, opened 2023-11-07
Paywalls: Support sending paywall events to servers (2)

tonidero:
### Description
This follows up on #1436. 

Here we are adding the code to send paywall events to our servers. This is not connected with the work done in #1436 yet. That will be done in followup PRs.
- [ ] Actually read events from disk and use this to send them to servers.
- [ ] Use paywall events to detect whether a purchase came from a paywall.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -60,7 +60,7 @@ internal typealias IdentifyCallback = Pair<(CustomerInfo, Boolean) -> Unit, (Pur
 internal typealias DiagnosticsCallback = Pair<(JSONObject) -> Unit, (PurchasesError, Boolean) -> Unit>
 
 /** @suppress */
-internal typealias PaywallEventsCallback = Pair<() -> Unit, (PurchasesError) -> Unit>
+internal typealias PaywallEventsCallback = Pair<() -> Unit, (PurchasesError, Boolean) -> Unit>
 
 /** @suppress */
 internal typealias ProductEntitlementCallback = Pair<(ProductEntitlementMapping) -> Unit, (PurchasesError) -> Unit>
@@ -465,7 +465,7 @@ internal class Backend(
     fun postPaywallEvents(
         paywallEventRequest: PaywallEventRequest,
         onSuccessHandler: () -> Unit,
-        onErrorHandler: (PurchasesError) -> Unit,
+        onErrorHandler: (error: PurchasesError, shouldMarkAsSynced: Boolean) -> Unit,
     ) {
         val body = PaywallEventsManager.json.encodeToJsonElement(paywallEventRequest).asMap()
         if (body == null) {
@@ -474,6 +474,7 @@ internal class Backend(
                     PurchasesErrorCode.UnknownError,
                     "Error encoding paywall event request",
                 ).also { errorLog(it) },
+                true,
             )
             return
         }
@@ -492,7 +493,7 @@ internal class Backend(
                 synchronized(this@Backend) {
                     paywallEventsCallbacks.remove(paywallEventRequest.cacheKey)
                 }?.forEach { (_, onErrorHandler) ->
-                    onErrorHandler(error)
+                    onErrorHandler(error, true)
                 }
             }
 
@@ -503,7 +504,7 @@ internal class Backend(
                     if (result.isSuccessful()) {
                         onSuccessHandler()
                     } else {
-                        onErrorHandler(result.toPurchasesError())
+                        onErrorHandler(result.toPurchasesError(), RCHTTPStatusCodes.isSynced(result.responseCode))
                     }
                 }
             }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -5,6 +5,8 @@
 
 package com.revenuecat.purchases.common
 
+import android.os.Build
+import androidx.annotation.RequiresApi
 import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.PostReceiptInitiationSource
 import com.revenuecat.purchases.PurchasesError
@@ -16,8 +18,12 @@ import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMap
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.models.GoogleReplacementMode
 import com.revenuecat.purchases.models.PricingPhase
+import com.revenuecat.purchases.paywalls.events.PaywallEventRequest
+import com.revenuecat.purchases.paywalls.events.PaywallEventsManager
 import com.revenuecat.purchases.strings.NetworkStrings
+import com.revenuecat.purchases.utils.asMap
 import com.revenuecat.purchases.utils.filterNotNullValues
+import kotlinx.serialization.json.encodeToJsonElement
 import org.json.JSONArray
 import org.json.JSONException
 import org.json.JSONObject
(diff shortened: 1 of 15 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -94,6 +103,9 @@ internal class Backend(
     @get:Synchronized @set:Synchronized
     @Volatile var diagnosticsCallbacks = mutableMapOf<CallbackCacheKey, MutableList<DiagnosticsCallback>>()
 
+    @get:Synchronized @set:Synchronized
+    @Volatile var paywallEventsCallbacks = mutableMapOf<CallbackCacheKey, MutableList<PaywallEventsCallback>>()
```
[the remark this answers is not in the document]
NachoSoto: 👍🏻  yeah long overdue separation as we did in iOS

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -441,14 +453,73 @@ internal class Backend(
...
+                return httpClient.performRequest(
+                    appConfig.paywallEventsURL,
+                    Endpoint.PostPaywallEvents,
+                    body,
+                    postFieldsToSign = null,
+                    backendHelper.authenticationHeaders,
+                )
+            }
+
+            override fun onError(error: PurchasesError) {
```
NachoSoto, line 492: Also outside the scope of this PR, but there is a lot of repetition on how we make requests as well that would be useful to abstract.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -441,14 +453,73 @@ internal class Backend(
...
+                ).also { errorLog(it) },
+                true,
+            )
+            return
+        }
+        val call = object : Dispatcher.AsyncCall() {
+            override fun call(): HTTPResult {
+                return httpClient.performRequest(
+                    appConfig.paywallEventsURL,
+                    Endpoint.PostPaywallEvents,
```
NachoSoto, line 485: It took me a while to find where this was defined. I feel like `paywallEventsURL` should be part of the `Endpoint` configuration. What do you think? Otherwise it would be possible (at compile time) to make a request to conflicting endpoints / domains.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCHTTPStatusCodes.kt
```diff
@@ -11,4 +11,9 @@ internal object RCHTTPStatusCodes {
 
     fun isSuccessful(statusCode: Int) = statusCode < BAD_REQUEST
     fun isServerError(statusCode: Int) = statusCode >= ERROR
+
+    // Note: this means that all 4xx (except 404) are considered as successfully synced.
+    // The reason is because it's likely due to a client error, so continuing to retry
+    // won't yield any different results and instead kill pandas.
+    fun isSynced(statusCode: Int) = isSuccessful(statusCode) || !(isServerError(statusCode) || statusCode == NOT_FOUND)
```
NachoSoto, line 18: I'm surprised this didn't exist yet. We use it in iOS for marking attributes as synced.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/JsonElementExtensions.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.doubleOrNull
+import kotlinx.serialization.json.floatOrNull
+import kotlinx.serialization.json.intOrNull
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.longOrNull
+
+fun JsonElement.asMap(): Map<String, Any?>? {
```
NachoSoto, line 17: This can be `internal`?

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionBackendIntegrationTest.kt
```diff
@@ -157,4 +161,41 @@ internal class ProductionBackendIntegrationTest: BaseBackendIntegrationTest() {
...
+                displayMode = "footer",
+                darkMode = true,
+                localeIdentifier = "en_US",
+            )
+        ))
+        ensureBlockFinishes { latch ->
+            backend.postPaywallEvents(
+                request,
+                onSuccessHandler = {
+                    latch.countDown()
```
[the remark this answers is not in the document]
NachoSoto: We use snapshot tests in iOS to verify the request format. Would it be possible to at least verify the body of the request?

purchases/src/test/java/com/revenuecat/purchases/common/networking/EndpointTest.kt
```diff
@@ -56,6 +57,13 @@ class EndpointTest {
         assertThat(endpoint.getPath()).isEqualTo(expectedPath)
     }
 
+    @Test
+    fun `Paywall events has correct path`() {
+        val endpoint = Endpoint.PostPaywallEvents
+        val expectedPath = "/events"
+        assertThat(endpoint.getPath()).isEqualTo(expectedPath)
+    }
+
```
NachoSoto, line 66: Ideally this verifies the domain too if/when that's part of `Endpoint`.

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/BaseBackendIntegrationTest.kt
```diff
@@ -72,6 +72,7 @@ internal abstract class BaseBackendIntegrationTest {
     ) {
         appConfig = mockk<AppConfig>().apply {
             every { baseURL } returns URL("https://api.revenuecat.com")
+            every { paywallEventsURL } returns URL("https://api-paywalls.revenuecat.com/")
```
NachoSoto, line 75: This could use `AppConfig` instead of duplicating it?

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Absolutely agree.

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/BaseBackendIntegrationTest.kt
tonidero: Yeah, just have to expose it, but might be better.

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionBackendIntegrationTest.kt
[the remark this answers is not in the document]
tonidero: I do this in the unit tests, not the backend integration tests: https://github.com/RevenueCat/purchases-android/pull/1442/files#diff-611e526013c5cd77f7141eb4ece99c74f81ecd9477bf84bf0f528e506ad40199R95. Lmk if you think we should still do that here... Might be tricky to intercept though, so I kinda think this is enough

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Hmm we could certainly move it there, but I also feel with the upcoming work for adding fallback base URLs, it might be better to keep separate... Honestly, I could see it going either way.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCHTTPStatusCodes.kt
tonidero: Hmm that's slightly different in Android for subscriber attributes... There, we would mark them as synced if it's not a server error and we receive a backend error code 24 (UnsupportedError), but we don't mark them as synced with 404... I guess we could add that as well?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/JsonElementExtensions.kt
tonidero: 🤦 Yup thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: I'm gonna hold on this until we actually rewrite the networking code to not increase the size of this PR even more.

NachoSoto: Sounds good, and that's a good point.

NachoSoto (approved):

PR #1443, opened 2023-11-07
Remove side effect from setting purchasesUpdatedListener

vegaro:
I removed a side effect on setting `purchasesUpdatedListener`. I don't think we should be starting/ending the connection there and we should do it independently

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/BasePurchasesTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/BasePurchasesTest.kt
@@ -41,6 +41,7 @@ import io.mockk.every
 import io.mockk.just
 import io.mockk.mockk
 import io.mockk.mockkStatic
+import io.mockk.runs
 import io.mockk.slot
 import io.mockk.unmockkStatic
 import org.junit.After
@@ -154,6 +155,10 @@ internal open class BasePurchasesTest {
             } answers {
                 purchasesUpdatedListener = null
             }
+
+            every {
+                startConnectionOnMainThread()
+            } just runs
         }
     }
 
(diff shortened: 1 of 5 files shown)


purchases/src/test/java/com/revenuecat/purchases/BasePurchasesTest.kt
```diff
@@ -154,6 +155,10 @@ internal open class BasePurchasesTest {
             } answers {
                 purchasesUpdatedListener = null
             }
+
+            every {
+                startConnectionOnMainThread()
+            } just runs
```
tonidero, line 161: Total nitpick, but I've been using `Runs` (uppercase) instead of `runs`. Do you have any preference?

tonidero (approved):
I think this makes sense. Have you tested the case of closing purchases then configuring again? We only reset the listener when closing purchases, so it's the only point where I think it could be problematic. Not sure how many people use that, but might be good to confirm it reconnects ok.

purchases/src/test/java/com/revenuecat/purchases/BasePurchasesTest.kt
vegaro: oh I have no preference, I will change it to the uppercase version 😄

vegaro:
> I think this makes sense. Have you tested the case of closing purchases then configuring again? We only reset the listener when closing purchases, so it's the only point where I think it could be problematic. Not sure how many people use that, but might be good to confirm it reconnects ok.

I haven't tested it, I will. I noticed though, that the `close` function does call `endConnection` after setting the listener to null, so I don't think there's any difference there:

```
fun close() {
        purchasesUpdatedListener = null
        endConnection()
}
```

Before this change, we were actually calling `endConnection` twice when calling `close` (one when setting the listener to null as a side effect and one directly in `close`)

vegaro:
It works @tonidero , I configured, then ended connection then configured and it works fine

PR #1444, opened 2023-11-07
Add retries to queryProductDetailsAsync

vegaro:
Separate `queryProductDetailsAsync` into a `QueryProductDetailsUseCase` that's easier to test and work with.

Added retry mechanisms for different errors.

This is still WIP until we determine which architecture we want to use.


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -78,6 +77,7 @@ internal class BillingWrapper(
     @Suppress("unused")
     private val diagnosticsTrackerIfEnabled: DiagnosticsTracker?,
     private val dateProvider: DateProvider = DefaultDateProvider(),
+    private val queryProductDetailsUseCase: QueryProductDetailsUseCase,
```
tonidero, line 80: Should we add a default value so we don't need need to pass it from the constructor everywhere else?

tonidero (commented):
This will simplify things so much 😊

aboedo (commented):
This is looking great! The only thing I'm wondering is how you're envisioning minimizing duplication here. Will each interaction with BC have its own corresponding UseCase class? How will they share logic for retries / etc? 
If they end up following a similar pattern, I think we should be able to unify the logic quite a bit between different use cases

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -0,0 +1,151 @@
...
+                    onReceive(storeProducts)
+                }
+
+                BillingResponse.ServiceDisconnected -> {
+                    retryConnection(productType, productIds, onReceive, onError)
+                }
+
+                BillingResponse.NetworkError,
+                BillingResponse.ServiceUnavailable,
+                BillingResponse.Error,
```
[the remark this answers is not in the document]
vegaro: I actually thought of creating sealed classes `Retryable`, `NonRetryable`, `SimpleRetry`, etc I am not sure we'll be able to represent things like "retry only for purchases", like for example `NETWORK_ERROR` on `consumeAsync`

vegaro:
@aboedo, that's a good point. I was envisioning each function to be its own use case, but I don't see an easy way of sharing logic, like for `BILLING_UNAVAILABLE`, `FEATURE_NOT_SUPPORTED`, `USER_CANCELED`, `DEVELOPER_ERROR` and `ITEM_UNAVAILABLE` they all should just forward the error.

Maybe this is not the right approach then 🤔

vegaro:
@aboedo maybe I can do something with an abstract class that shares most of the logic, but also lets implement custom logic if required 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -159,7 +159,7 @@ internal class BillingWrapper(
     }
 
     @Synchronized
-    private fun executeRequestOnUIThread(request: (PurchasesError?) -> Unit) {
+    fun executeRequestOnUIThread(request: (PurchasesError?) -> Unit) {
```
tonidero, line 162: Hmm do we need to change visibility here?

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -0,0 +1,129 @@
...
+        )
+        received.takeUnless { it.isEmpty() }?.forEach {
+            log(LogIntent.PURCHASE, OfferingStrings.LIST_PRODUCTS.format(it.productId, it))
+        }
+
+        val storeProducts = received.toStoreProducts()
+        onReceive(storeProducts)
+    }
+
+    override fun forwardError(billingResult: BillingResult, onError: PurchasesErrorCallback) {
```
tonidero, line 82: I'm wondering, would it be possible to move this to the `UseCase` base class? We could change the logs so we can print something generic and just provide a name to each use case so it can be printed in these logs... Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/UseCase.kt
```diff
@@ -0,0 +1,80 @@
...
+        billingResult: BillingResult,
+        response: T,
+    ) {
+        when (BillingResponse.fromCode(billingResult.responseCode)) {
+            BillingResponse.OK -> {
+                onOk(response)
+            }
+
+            BillingResponse.ServiceDisconnected -> {
+                run()
```
tonidero, line 61: A couple things:
- Should we log something in this case? 
- We don't seem to have any tests covering this behavior and some other errors here. Should we add them? 
- I guess we will have infinite retries here right? I think that's intended, but wanted to make sure.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/UseCase.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.revenuecat.purchases.common.LogIntent
+import com.revenuecat.purchases.common.log
+import com.revenuecat.purchases.google.BillingResponse
+import com.revenuecat.purchases.strings.BillingStrings
+import java.io.PrintWriter
+import java.io.StringWriter
+
+private const val MAX_RETRIES_DEFAULT = 3
+
+abstract class UseCase<T>(
```
tonidero, line 16: Should be `internal`

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingResponse.kt
```diff
@@ -0,0 +1,42 @@
+package com.revenuecat.purchases.google
+
+import com.android.billingclient.api.BillingClient.BillingResponseCode
+
+sealed class BillingResponse {
```
tonidero, line 5: Should be `internal`.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/UseCase.kt
```diff
@@ -0,0 +1,80 @@
...
+
+            BillingResponse.ServiceDisconnected -> {
+                run()
+            }
+
+            BillingResponse.NetworkError,
+            BillingResponse.ServiceUnavailable,
+            BillingResponse.Error,
+            -> {
+                if (retryAttempt < maxRetries) {
```
tonidero, line 68: It's nice we don't need to reset the `retryAttempt` since this use case is created for each request 👍

tonidero (commented):
Mostly, I think we should add some tests for the new behavior. Other than that, looks great!! Love this new direction.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/UseCase.kt
vegaro: Oh nice, I hadn't noticed that, so that's a nice side effect yes

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/UseCase.kt
vegaro: > I guess we will have infinite retries here right? I think that's intended, but wanted to make sure.

Yes, but... it's very strange that we end up here because we are checking if the service is ready before executing anything. We still need to cover it, in case it gets disconnected while this is executing though.

So `run` will re add it to the queue via `executeRequestOnUIThread`, and it will retry the reconnection via `startConnection`. It's actually intended we readd the whole request to the queue, so we don't end up locking the main thread in an infinite loop.

purchases/src/test/java/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
```diff
@@ -0,0 +1,621 @@
...
+
+        val productIDs = setOf("product_a")
+
+        var receivedList: List<StoreProduct>? = null
+        wrapper.queryProductDetailsAsync(
+            ProductType.SUBS,
+            productIDs, {
+                receivedList = it
+            }, {
+                AssertionsForClassTypes.fail("shouldn't be an error")
```
tonidero, line 161: Hmm couldn't this be `Assertions.fail`? I have never used `AssertionsForClassTypes` 😅

tonidero (approved):
Looks great to me!

purchases/src/test/java/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
vegaro: oh yes, what's this!!! Android Studio did something very weird when I moved things here, I've had to remove a lot of these implicit imports

PR #1445, opened 2023-11-08
Paywalls: Add paywall events flush logic and tests (3)

tonidero:
### Description
This adds the code to the `PaywallEventsManager` class to actually send the events to the backend and clear them locally once finished.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -255,7 +255,7 @@ internal class PurchasesFactory(
                 postPendingTransactionsHelper,
                 syncPurchasesHelper,
                 offeringsManager,
-                createPaywallEventsManager(application, identityManager, eventsDispatcher),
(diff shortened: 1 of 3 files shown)


NachoSoto (approved):
Looks great!
I want to make sure this handles (and tests) a specific use case that's tested in iOS:

See `PaywallEventStoreTests.testFetchEventsWithUnrecognizedLines`:
That's basically testing that this code:
```swift
return try await self.handler.readLines()
    .prefix(count)
    .compactMap { try? PaywallEventSerializer.decode($0) }
```
Is written like that, instead of
```swift
return try await self.handler.readLines()
    .compactMap { try? PaywallEventSerializer.decode($0) }
    .prefix(count)
```

Notice `prefix` happens first

So that if invalid lines are found, they still get removed from the store.

tonidero:
Nope, I didn't have that logic. Thanks for mentioning it! I will fix that and add some tests tomorrow

tonidero:
Made changes to handle this in https://github.com/RevenueCat/purchases-android/pull/1451

PR #1446, opened 2023-11-08
Paywalls: Move logic for events file helper to common generic class (4)

tonidero:
### Description
I was working on sharing logic between diagnostics and paywall events, but there were quite a few differences that made this tricky. Diagnostics doesn't use kotlin serialization, it anonymizes the events, and has a bunch of other logic that we need to remove to be able to share most of the code with paywall events.

In this first PR, I abstracted most of the logic of `PaywallEventsFileHelper` and `DiagnosticsFileHelper` to a common `EventsFileHelper`.


purchases/src/main/kotlin/com/revenuecat/purchases/utils/EventsFileHelper.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.revenuecat.purchases.common.FileHelper
+import com.revenuecat.purchases.common.verboseLog
+import org.json.JSONObject
+import java.util.stream.Stream
+
+@RequiresApi(Build.VERSION_CODES.N)
+internal open class EventsFileHelper<T : Event> (
+    private val fileHelper: FileHelper,
+    private val filePath: String,
+    private val stringToEventConverter: ((String) -> T)? = null,
```
tonidero, line 14: Had to make this nullable since for diagnostics we don't parse the file back to `DiagnosticEntry` objects, as we do for `PaywallStoredEvent`. So had to add another read method that reads the file as JSONObject. 

The alternative was to make diagnostics use kotlin serialization, but as I started on that, I noticed I would have to rewrite a lot of things and didn't want to spend a lot of time there.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsFileHelper.kt
```diff
@@ -23,32 +21,4 @@ internal class DiagnosticsFileHelper(
     fun isDiagnosticsFileTooBig(): Boolean {
```
tonidero, line 21: This is the only method remaining here. I want to finish paywall events before dedicating more time to rewriting the whole thing for diagnostics.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -16,7 +17,7 @@ import kotlin.time.Duration
  */
 internal class DiagnosticsTracker(
     private val appConfig: AppConfig,
-    private val diagnosticsFileHelper: DiagnosticsFileHelper,
+    private val diagnosticsFileHelper: EventsFileHelper<DiagnosticsEntry>,
```
vegaro, line 20: Love it

vegaro (commented):
I think this makes sense as a first refactor

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
NachoSoto: +1

purchases/src/main/kotlin/com/revenuecat/purchases/utils/Event.kt
```diff
@@ -0,0 +1,5 @@
+package com.revenuecat.purchases.utils
+
+interface Event {
```
NachoSoto, line 3: This is `public`?
Also I'd add a docstring

purchases/src/main/kotlin/com/revenuecat/purchases/utils/EventsFileHelper.kt
NachoSoto: Maybe
```suggestion
    private val eventDeserializer: ((String) -> T)? = null,
```

And I would leave a comment for how this behaves if it's `null`.

NachoSoto (approved):

PR #1447, opened 2023-11-08
Paywalls: Track paywall events

tonidero:
### Description
This PR adds tracking of the `paywall_impression`, `paywall_close` and `paywall_cancel` events to the paywalls.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -47,7 +47,7 @@ internal fun InternalPaywall(
     viewModel: PaywallViewModel = getPaywallViewModel(options),
 ) {
     BackHandler {
-        viewModel.close()
+        viewModel.closePaywall()
     }
     PaywallTheme(fontProvider = options.fontProvider) {
         viewModel.refreshStateIfLocaleChanged()
@@ -63,7 +63,7 @@ internal fun InternalPaywall(
             LoadingPaywall(
                 mode = options.mode,
                 shouldDisplayDismissButton = options.shouldDisplayDismissButton,
-                onDismiss = viewModel::close,
+                onDismiss = viewModel::closePaywall,
             )
         }
 
@@ -118,7 +118,7 @@ private fun LoadedPaywall(state: PaywallState.Loaded, viewModel: PaywallViewMode
             },
     ) {
         TemplatePaywall(state = state, viewModel = viewModel)
-        CloseButton(state.shouldDisplayDismissButton, viewModel::close)
+        CloseButton(state.shouldDisplayDismissButton, viewModel::closePaywall)
     }
 }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -1,9 +1,11 @@
 package com.revenuecat.purchases.ui.revenuecatui
 
+import androidx.activity.compose.BackHandler
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.background
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.AlertDialog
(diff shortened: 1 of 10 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -50,15 +50,19 @@ fun PaywallDialog(
         val dismissRequest = {
             shouldDisplayDialog = false
         }
+        val paywallOptions = paywallDialogOptions.toPaywallOptions(dismissRequest)
+
+        val viewModel = getPaywallViewModel(options = paywallOptions)
```
tonidero, line 55: I needed the viewmodel in order to track this event when dismissing the dialog correctly. It shouldn't cause duplications since we don't have keys anymore and it's the same model.

tonidero (commented):
Need to add tests and test it manually a bit more, but seems to be working.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -44,9 +46,13 @@ internal fun InternalPaywall(
     options: PaywallOptions,
     viewModel: PaywallViewModel = getPaywallViewModel(options),
 ) {
+    viewModel.trackPaywallImpressionIfNeeded()
+    BackHandler {
+        viewModel.close()
```
NachoSoto, line 51: Is `viewModel.close` called twice now with this?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -83,6 +93,9 @@ internal class PaywallViewModelImpl(
     private val mode: PaywallMode
         get() = options.mode
 
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
```
NachoSoto, line 96: Maybe `PaywallViewModel` can be marked with this instead so we don't add the `@OptIn` in so many methods?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
tonidero: No, the `BackHandler` is only called when the user presses the back button/performs the back navigation gesture.

We only call `close` manually when the user presses the `X` button or dismisses the PaywallDialog (by tapping outside it).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -50,15 +50,19 @@ fun PaywallDialog(
...
 
         Dialog(
             onDismissRequest = {
                 dismissRequest()
+                viewModel.close()
```
tonidero, line 60: I would like to add tests to make sure we're calling these appropriately, but those would be UI tests, not snapshot tests, which might be a bit harder to write... Will try to write these tests after.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
NachoSoto: Oh so we weren't calling that then, good catch

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -262,4 +288,53 @@ internal class PaywallViewModelImpl(
...
+            type = eventType,
+        )
+
+        purchases.track(event)
+    }
+
+    @Suppress("ReturnCount")
+    private fun createEventData(): PaywallEvent.Data? {
+        val currentState = state.value
+        if (currentState !is PaywallState.Loaded) {
```
NachoSoto, line 321: TIL `!is`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -120,8 +135,9 @@ internal class PaywallViewModelImpl(
         }
     }
 
-    override fun closeButtonPressed() {
-        Logger.d("Paywalls: Close button pressed.")
+    override fun close() {
```
NachoSoto, line 138: Maybe?
```suggestion
    override fun closePaywall() {
```

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
```diff
@@ -432,4 +543,36 @@ class PaywallViewModelTest {
...
+        verify(exactly = 0) {
+            purchases.track(
+                withArg {
+                    assertThat(it.type).isEqualTo(eventType)
+                }
+            )
+        }
+    }
+
+    private fun delayGettingOfferings() {
```
NachoSoto, line 571:
```suggestion
    private fun delayFetchingOfferings() {
```

PR #1204, opened 2023-08-18
Optimize SDK initialization when requests executed before any activity starts

tonidero:
### Description
We discovered that the SDK might take a while to respond to offerings/customer info operations if they are started before any `Activity` has had the chance to start. This is because the SDK considers the app to be in the background, and we have some optimizations in place where we add some jittering to backend requests performed in the background

The approach in this PR is to make it so:
- If a request without a jitter delay is started after a request with jitter (but before it's completed), we move the callbacks to use the response from the request without jittering.
- If a request with jittering is started after a request without jitter (but before it's completed), we will not add the request with jitter but just reuse the existing request without jittering.

Note that with this approach, we might be duplicating requests in these situations... But I think it should be ok, lmk if you think otherwise

PR #1453, opened 2023-11-09
Lower request jitter log level from warning to debug

tonidero:
### Description
These logs were added in https://github.com/RevenueCat/purchases-android/pull/1204. These logs might happen primarily if the developer calls getCustomerInfo/getOfferings before any activity has started. In those situations, we have to handle it so the request does not wait for the whole jitter.

There is nothing much the developer can do about this and it's not actually an error, so lowering the log level from warning to debug for these.

Fixes https://github.com/RevenueCat/react-native-purchases/issues/757


NachoSoto (approved):
🙏🏻

PR #1460, opened 2023-11-10
Add retries to queryPurchases

vegaro:
Adds retries to queryPurchases by creating a `QueryPurchasesUseCase`


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
```diff
@@ -37,15 +37,16 @@ internal abstract class BillingClientUseCase<T>(
     }
 
     abstract fun executeAsync()
-    abstract fun onOk(received: T)
```
vegaro, line 40: Had to make some changes to the base class because queryPurchases chains two calls to the BillingClient

purchases/src/test/java/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCaseTest.kt
```diff
@@ -314,9 +314,9 @@ internal class QueryPurchaseHistoryUseCaseTest: BaseBillingUseCaseTest() {
                 timesExecutedInMainThread++
 
                 queryPurchaseHistoryStubbing answers {
-                    if (timesExecutedInMainThread == 0) {
+                    if (timesExecutedInMainThread == 1) {
```
vegaro, line 309: This was actually 1, because it is actually increased the first time it's executed

tonidero (approved):
Left a question, but looking good!

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
```diff
@@ -35,7 +35,7 @@ internal class GetBillingConfigUseCase(
                     )
                     return@getBillingConfigAsync
                 }
-                processResult(result, config)
+                processResult(result, config, ::onOk)
```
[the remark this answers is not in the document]
vegaro: oh yes, it's not needed because it's the default of the function

```
    fun processResult(
        billingResult: BillingResult,
        response: T,
        onSuccess: (T) -> Unit = ::onOk,
        onError: (BillingResult) -> Unit = ::forwardError,
    ) {
```

I added that `onSuccess` parameter everywhere, then the default and forgot to remove the parameter here. It's not needed

PR #1463, opened 2023-11-13
Mark purchases.setOnesignalUserID as deprecated

Raquel10-RevenueCat:
We need to support OneSignal aliases as a prerequisite to allow using the setOneSignalUserId function in the intended way. Supporting OneSignal aliases is not yet prioritized, so we will mark setOneSignalUserId as deprecated until this functionality is fully supported.

### Checklist

### DIFF
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
@@ -85,6 +85,7 @@ final class PurchasesAPI {
         purchases.setFBAnonymousID("");
         purchases.setMparticleID("");
         purchases.setOnesignalID("");
+        purchases.setOnesignalUserID("");
         purchases.setAirshipChannelID("");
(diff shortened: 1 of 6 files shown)


NachoSoto:
This is already part of the public API so removing it is a breaking change unfortunately.

NachoSoto (dismissed):
We'll need to figure out an alternative. Maybe document that it's not fully supported by the backend yet?

NachoSoto (approved):

NachoSoto:
@Raquel10-RevenueCat thanks for updating this, looks good! 🙌🏻

@RevenueCat/coresdk thoughts on the label? Should we make this a `feat` since I believe the right thing is to make it a minor bump?

tonidero (approved):

tonidero:
> Should we make this a feat since I believe the right thing is to make it a minor bump?

Hmm, considering we are just going back to a previously deprecated (not actually marked as deprecated API), I think it's ok to be a patch? Basically this is only really deprecating an API.

PR #1464, opened 2023-11-13
`Paywalls`: improve image loading

NachoSoto:
#### Added memory cache:

We currently only had disk cache. This should improve image loading performance. Note that `CoilImageDownloader` is untouched because it would have its own memory cache.

#### Work around corrupted images in cache:

Currently if you end up with a corrupted image in the cache, it would continue failing to load forever.
Ideally Coil handles this, but with this implementation we'll attempt to load them again with a write-only cache policy. If that succeeds, the cache will get updated with the correct image.


tonidero (approved):
Looks great! 👍

PR #1465, opened 2023-11-14
`Paywalls`: fix template 5 header aspect ratio

NachoSoto:
### Before:

image

### After

Screenshot 2023-11-13 at 18 10 09

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
@@ -182,10 +182,10 @@ private fun HeaderImage(uri: Uri?, templateSize: IntSize) {
         RemoteImage(
             urlString = uri.toString(),
             modifier = Modifier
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -182,10 +182,10 @@ private fun HeaderImage(uri: Uri?, templateSize: IntSize) {
         RemoteImage(
             urlString = uri.toString(),
             modifier = Modifier
-                .conditional(aspectRatio > 1f || templateSize == IntSize.Zero) {
+                .conditional(aspectRatio < 1f || templateSize == IntSize.Zero) {
```
tonidero, line 185: Maybe
```suggestion
                .conditional(aspectRatio <= 1f || templateSize == IntSize.Zero) {
```

tonidero (approved):
Applied the fix to template 1 in #1466. But yeah, great catch!

tonidero:
Actually, I wonder about why the snapshot tests are not failing... seems like there might be something wrong there... Will take a look.

PR #1418, opened 2023-10-30
[WIP] `Paywalls`: override locale

NachoSoto:

PR #1467, opened 2023-11-14
Paywalls: Fix purchasing regression by providing real activity

tonidero:
### Description
With #1418, we are overriding the context in order to provide our own locale to make sure the screen has the same locale in the whole screen. However, with the current implementation, the context becomes a `ContextImpl` which can't be converted to the original activity. This was causing a crash when starting the purchase process

This PR fixes the issue by providing a new `LocalActivity` local composition that is performed before the overriden context takes place so we can access the activity in our composable tree.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -125,6 +127,7 @@ private fun LoadedPaywall(state: PaywallState.Loaded, viewModel: PaywallViewMode
         val configuration = state.configurationWithOverriddenLocale()
 
         CompositionLocalProvider(
+            LocalActivity provides LocalContext.current.getActivity(),
```
tonidero, line 130: This shouldn't cause any leaks since any changes to the `LocalContext.current` would cause a recomposition to the composition local provider

tonidero:
Need to fix the tests

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/HelperFunctions.kt
```diff
@@ -11,6 +14,12 @@ import com.revenuecat.purchases.getCustomerInfoWith
 import kotlin.coroutines.resume
 import kotlin.coroutines.suspendCoroutine
 
+/**
+ * CompositionLocal containing the current Activity.
+ * Needs to be provided, it's null by default.
+ */
+internal val LocalActivity: ProvidableCompositionLocal<Activity?> = compositionLocalOf { null }
```
NachoSoto, line 21: Oooh I've been meaning to figure out how to do this 👌🏻

NachoSoto (commented):
Awesome thanks for fixing this! 👏🏻 

A couple of extra things that would be nice:
- Can we remove that `Context.getActivity`?
- Would it be possible to add a test that reproduces the crash? I'm worried something like this can happen again and we wouldn't catch purchases crashing

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -208,3 +209,19 @@ private fun ErrorDialog(
         },
     )
 }
+
+/**
+ * Returns the activity from a given context. Most times, the context itself will be
+ * an activity, but in the case it's not, it will iterate through the context wrappers until it
+ * finds one that is an activity.
+ */
+private fun Context.getActivity(): Activity? {
```
NachoSoto, line 218: 👍🏻

NachoSoto (approved):

PR #1468, opened 2023-11-14
Prevents queryPurchases from being called twice

vegaro:
We are getting this exception:

```
Exception java.lang.IllegalStateException:
  at kotlin.coroutines.SafeContinuation.resumeWith (SafeContinuationJvm.kt:44)
  at <private>
  at com.revenuecat.purchases.CustomerInfoHelper$getCustomerInfoFetchOnly$1$1.invoke (CustomerInfoHelper.kt:107)
  at com.revenuecat.purchases.CustomerInfoHelper$getCustomerInfoFetchOnly$1$1.invoke (CustomerInfoHelper.kt:107)
  at com.revenuecat.purchases.CustomerInfoHelper.dispatch$lambda$0 (CustomerInfoHelper.kt:197)
  at android.os.Handler.handleCallback (Handler.java:942)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loopOnce (Looper.java:240)
  at android.os.Looper.loop (Looper.java:351)
  at android.app.ActivityThread.main (ActivityThread.java:8381)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:584)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1013)
  ```
  
  I haven't been able to reproduce it but it looks like `getCustomerInfoFetchOnly` is calling the same continuation twice. If `queryPurchases` callback gets called twice, it would cause the same exception. We have other functions that are called twice, so I have added a check to prevent `queryPurchases` callback from being called twice.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -925,7 +925,7 @@ internal class BillingWrapper(
             if (hasResponded.getAndSet(true)) {
                 log(
                     LogIntent.GOOGLE_ERROR,
-                    OfferingStrings.EXTRA_QUERY_PRODUCT_DETAILS_RESPONSE.format(billingResult.responseCode),
+                    OfferingStrings.EXTRA_QUERY_PURCHASES_RESPONSE.format(billingResult.responseCode),
                 )
                 return@queryPurchasesAsync
             }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -64,6 +64,7 @@ import java.io.StringWriter
 import java.lang.ref.WeakReference
 import java.util.Date
 import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.math.min
 import kotlin.time.Duration
 
@@ -486,7 +487,7 @@ internal class BillingWrapper(
                     )
                     return@withConnectedClient
                 }
-                this.queryPurchasesAsyncWithTracking(
+                this.queryPurchasesAsyncWithTrackingEnsuringOneResponse(
                     BillingClient.ProductType.SUBS,
                     querySubsPurchasesParams,
                 ) querySubPurchasesAsync@{ activeSubsResult, activeSubsPurchases ->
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt [hunk 1]
```diff
@@ -893,20 +891,17 @@ internal class BillingWrapper(
...
-                synchronized(this@BillingWrapper) {
-                    if (hasResponded) {
-                        log(
-                            LogIntent.GOOGLE_ERROR,
-                            RestoreStrings.EXTRA_QUERY_PURCHASE_HISTORY_RESPONSE.format(billingResult.responseCode),
-                        )
-                        return@queryPurchaseHistoryAsync
-                    }
-                    hasResponded = true
+                if (hasResponded.getAndSet(true)) {
```
vegaro, line 899: I think this is cleaner than the `synchronized(this@BillingWrapper)`

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -2711,6 +2710,75 @@ class BillingWrapperTest {
...
+                fail("shouldn't be an error")
+            }
+        )
+
+        assertThat(numCallbacks).isEqualTo(1)
+    }
+
+    @Test
+    fun `queryPurchasesAsync only calls one response when BillingClient responds twice from different threads`() {
+        val numCallbacks = AtomicInteger(0)
```
vegaro, line 2743: This AtomicInteger was tricky, had to make it atomic otherwise tests failed on CI. Made the change in the other tests that are similar to this one just in case

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -919,13 +914,21 @@ internal class BillingWrapper(
         }
     }
 
-    private fun BillingClient.queryPurchasesAsyncWithTracking(
+    private fun BillingClient.queryPurchasesAsyncWithTrackingEnsuringOneResponse(
         @BillingClient.ProductType productType: String,
         queryParams: QueryPurchasesParams,
         listener: PurchasesResponseListener,
     ) {
+        val hasResponded = AtomicBoolean(false)
```
tonidero, line 922: Oh right, so this is the main change right? We weren't handling multiple results from the billing client on the `queryPurchases` method?

tonidero (approved):
I think the changes make sense!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
vegaro: Yes exactly. I haven't been able to reproduce it but forcing the callback in this function to execute twice will render the same exception. I looked at the other places where we could be triggering the callback twice and it's only in the backend call, but didn't see anything strange there.

NachoSoto:
Looks like a test is failing in `main`: https://circleci.com/gh/RevenueCat/purchases-android/32281
I wonder if that test is flaky?

PR #1469, opened 2023-11-14
Paywalls: Support condensed footer in template 4

tonidero:
### Description
Template 4 did not support condensed footer mode. This adds it with a simple slide up-down animation.

https://github.com/RevenueCat/purchases-android/assets/808417/59025a19-59c7-4669-b1be-93843ab1290c


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template4.kt
```diff
@@ -141,26 +151,44 @@ private fun Template4MainContent(
...
-                textWithNoIntroOffer = it.localization.offerDetails,
-                textWithIntroOffer = it.localization.offerDetailsWithIntroOffer,
-                textWithMultipleIntroOffers = it.localization.offerDetailsWithMultipleIntroOffers,
-                eligibility = it.introEligibility,
-                color = colors.text1,
-                style = MaterialTheme.typography.bodyMedium,
-                textAlign = TextAlign.Center,
-                modifier = Modifier.fillMaxWidth(),
-            )
+        Column(horizontalAlignment = Alignment.CenterHorizontally) {
```
tonidero, line 158: I needed to wrap these into another column like this to avoid the outer column `verticalArrangement` spacing. Seems like the `AnimatedVisibility` jumps when trying to animate an item disappearing and that spacing going away. I'm mimicking that by manually handling the animation of a custom `Spacer` here.

NachoSoto: > Seems like the AnimatedVisibility jumps when trying to animate an item disappearing and that spacing going away

Yeah that's why we can't use it in many place where e need spacing to remain consistent.

NachoSoto (approved):

PR #1477, opened 2023-11-15
Fix flaky test in BillingWrapperTest

vegaro:
`queryPurchasesAsync` is called twice in the function, so 4 would come from calls to `onQueryPurchasesResponse` and 1 for the `onSuccess`

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
@@ -2776,7 +2776,7 @@ class BillingWrapperTest {
             }
         )
 
-        lock.await(2, TimeUnit.SECONDS)
+        lock.await(300, TimeUnit.MILLISECONDS)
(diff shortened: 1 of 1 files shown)


purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -2773,7 +2776,7 @@ class BillingWrapperTest {
             }
         )
 
-        lock.await()
+        lock.await(2, TimeUnit.SECONDS)
```
tonidero, line 2779: It would probably be fine here with 200-300ms... But I think it's ok. If it's not done in 300ms, it shouldn't be done in 2s either, so this should fail as well.

tonidero (approved):

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
vegaro: yes, makes sense. I will update!

PR #1481, opened 2023-11-16
Update Billing client to 6.1.0

tonidero:
### Description
I run some basic tests to make sure common use cases work (subscribe, unsubscribe and upgrade). Everything seems to work fine. We can add support for the new APIs in future PRs.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
@@ -73,7 +73,7 @@ class GoogleDeviceIdentifiersFetcherTests {
             mockContext = mockContext,
             expectedAdID = "12345",
             expectedAndroidID = "androidid",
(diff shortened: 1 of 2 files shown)


vegaro (approved):

purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
```diff
@@ -73,7 +73,7 @@ class GoogleDeviceIdentifiersFetcherTests {
             mockContext = mockContext,
             expectedAdID = "12345",
             expectedAndroidID = "androidid",
-            expectedException = GooglePlayServicesRepairableException(1, "error", null)
+            expectedException = GooglePlayServicesRepairableException(1, "error", mockk())
```
tonidero, line 76: Looks like Google changed this from being nullable to non-nullable in 6.1.0, which is a breaking change 😬...

NachoSoto: 👏🏻

PR #1482, opened 2023-11-16
Add method to get storefront for Google and Amazon stores

tonidero:
### Description
This PR adds a new method to the `BillingAbstract` to obtain the country code of the store, now that it's been added in Google's BC 6.1.0.

In this PR, we only log the country as soon as we connect but we plan to send this data to the backend in future PRs.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
@@ -1,0 +1,48 @@
+package com.revenuecat.purchases.google.usecase
+
+import com.android.billingclient.api.BillingClient
+import com.android.billingclient.api.BillingConfig
+import com.android.billingclient.api.GetBillingConfigParams
+import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesErrorCallback
+import com.revenuecat.purchases.PurchasesErrorCode
+import com.revenuecat.purchases.common.LogIntent
+import com.revenuecat.purchases.common.log
+import com.revenuecat.purchases.strings.BillingStrings
+import com.revenuecat.purchases.strings.OfferingStrings
+import java.util.concurrent.atomic.AtomicBoolean
+
+internal class GetBillingConfigUseCase(
+    val onReceive: (BillingConfig) -> Unit,
+    val onError: PurchasesErrorCallback,
+    val withConnectedClient: (BillingClient.() -> Unit) -> Unit,
+    executeRequestOnUIThread: ((PurchasesError?) -> Unit) -> Unit,
+) : BillingClientUseCase<BillingConfig?>(onError, executeRequestOnUIThread) {
+    override val errorMessage: String
+        get() = "Error getting billing config"
+
+    override fun executeAsync() {
+        withConnectedClient {
+            val hasResponded = AtomicBoolean(false)
+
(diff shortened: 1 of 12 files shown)


joshdholtz (approved):
😍

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+    val onError: PurchasesErrorCallback,
+    val withConnectedClient: (BillingClient.() -> Unit) -> Unit,
+    executeRequestOnUIThread: ((PurchasesError?) -> Unit) -> Unit,
+) : BillingClientUseCase<BillingConfig?>(onError, executeRequestOnUIThread) {
+    override val errorMessage: String
+        get() = "Error getting billing config"
+
+    override fun executeAsync() {
+        withConnectedClient {
+            val hasResponded = AtomicBoolean(false)
```
vegaro, line 26: Nice! better be safe than sorry lol

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+                        OfferingStrings.EXTRA_GET_BILLING_CONFIG_RESPONSE.format(result.responseCode),
+                    )
+                    return@getBillingConfigAsync
+                }
+                processResult(result, config)
+            }
+        }
+    }
+
+    override fun onOk(received: BillingConfig?) {
```
vegaro, line 41: oh interesting, so BillingClient can return an OK with a null BillingConfig!

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+) : BillingClientUseCase<BillingConfig?>(onError, executeRequestOnUIThread) {
+    override val errorMessage: String
+        get() = "Error getting billing config"
+
+    override fun executeAsync() {
+        withConnectedClient {
+            val hasResponded = AtomicBoolean(false)
+
+            getBillingConfigAsync(GetBillingConfigParams.newBuilder().build()) { result, config ->
+                if (hasResponded.getAndSet(true)) {
```
vegaro, line 29: Worth it to change to main thread? we are not doing anything critical so I think it's fine to stay in this thread if it's a background thread

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
tonidero: Yeah, it's weird but it's nullable in the API 🤷‍♂️ . So I'm forced to perform this extra mapping to also error if this is null.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
tonidero: Right... in the followup PR we are storing stuff in the cache, but other than that we are just logging things, so I think it's ok to stay in the same thread. Lmk if you think otherwise and I can change it in a followup PR.

PR #1483, opened 2023-11-16
Use state listener in Amazon store

tonidero:
### Description
We were not calling the BillingAbstract `stateListener` after it was connected. This adds a call to make sure we call it and keep the same behavior as on Google. Currently, we were just syncing pending purchases in Google, which shouldn't matter much if we do it there for Amazon since we're also doing on app foreground.


joshdholtz (approved):
👍

vegaro (approved):
Makes sense! As you said, if we are already doing it `onAppForeground`, this should be safe

PR #1484, opened 2023-11-16
Add `X-Storefront` header to requests

tonidero:
### Description
This adds a new `X-Storefront` header to all backend requests.  This will allow the backend to know the store country for users.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -9,7 +9,6 @@ import android.os.Build
 import androidx.annotation.VisibleForTesting
 import com.revenuecat.purchases.Store
 import com.revenuecat.purchases.VerificationResult
-import com.revenuecat.purchases.common.caching.DeviceCache
 import com.revenuecat.purchases.common.diagnostics.DiagnosticsTracker
 import com.revenuecat.purchases.common.networking.ETagManager
 import com.revenuecat.purchases.common.networking.Endpoint
@@ -20,6 +19,7 @@ import com.revenuecat.purchases.common.networking.RCHTTPStatusCodes
 import com.revenuecat.purchases.common.verification.SignatureVerificationException
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.common.verification.SigningManager
+import com.revenuecat.purchases.interfaces.StorefrontProvider
 import com.revenuecat.purchases.strings.NetworkStrings
 import com.revenuecat.purchases.utils.filterNotNullValues
 import org.json.JSONException
@@ -43,7 +43,7 @@ internal class HTTPClient(
     private val eTagManager: ETagManager,
     private val diagnosticsTrackerIfEnabled: DiagnosticsTracker?,
     val signingManager: SigningManager,
-    private val deviceCache: DeviceCache,
+    private val storefrontProvider: StorefrontProvider,
     private val dateProvider: DateProvider = DefaultDateProvider(),
     private val mapConverter: MapConverter = MapConverter(),
 ) {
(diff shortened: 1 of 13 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/GetBillingConfigUseCase.kt
```diff
@@ -42,6 +44,7 @@ internal class GetBillingConfigUseCase(
         if (received == null) {
             onError(PurchasesError(PurchasesErrorCode.StoreProblemError, BillingStrings.BILLING_CONFIG_NULL_ON_SUCCESS))
         } else {
+            deviceCache.setStorefront(received.countryCode)
```
tonidero, line 47: This will be updated every time we connect to the billing client. Having said that, there is a chance of the cache to be outdated if it loses connection at some point, but in that case, we won't be able to get offerings or query purchases, so it should be fine on most cases.

joshdholtz: Yeah yeah, that's totally fine!

joshdholtz (approved):
Thank you so much for doing this! 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -36,11 +37,13 @@ import java.net.URLConnection
 import java.util.Date
 import kotlin.time.Duration
 
+@Suppress("LongParameterList")
 internal class HTTPClient(
     private val appConfig: AppConfig,
     private val eTagManager: ETagManager,
     private val diagnosticsTrackerIfEnabled: DiagnosticsTracker?,
     val signingManager: SigningManager,
+    private val deviceCache: DeviceCache,
```
NachoSoto, line 46: Instead of injecting this huge type, how about injecting a `UserContextProvider` interface?

Otherwise ripping out DeviceCache from here in the future might get harder.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -177,6 +180,17 @@ internal open class DeviceCache(
...
 
+    @Synchronized
+    fun setStorefront(countryCode: String) {
+        verboseLog(BillingStrings.BILLING_STOREFRONT_CACHING.format(countryCode))
+        preferences.edit().putString(storefrontCacheKey, countryCode).apply()
+    }
+
+    @Synchronized
+    fun getStorefront(): String? {
+        return preferences.getString(storefrontCacheKey, null)
```
NachoSoto, line 191: I'd add a debug log if fetching this doesn't have a cache yet so we know if there are race conditions.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Hmm what else would you add to the `UserContextProvider`? If it would be only for getting the storefront, I would probably have something more specific right now, like `StorefrontProvider` or something like that? Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
tonidero: Yeah, this should only happen the first time before the app has connecting to the billing client in theory, but it doesn't hurt 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/1484/commits/2efb8b52dfdc744ebb41ab481ef1c23423bed78a

NachoSoto (approved):

PR #1487, opened 2023-11-17
ConsumePurchaseUseCase

vegaro:
Created ConsumePurchaseUseCase with retries

Based off #1490

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/ConsumePurchaseUseCase.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/ConsumePurchaseUseCase.kt
@@ -1,0 +1,45 @@
+package com.revenuecat.purchases.google.usecase
+
+import com.android.billingclient.api.BillingClient
+import com.android.billingclient.api.ConsumeParams
+import com.revenuecat.purchases.PostReceiptInitiationSource
+import com.revenuecat.purchases.PurchasesErrorCallback
+
+internal class ConsumePurchaseUseCaseParams(
+    val purchaseToken: String,
+    val initiationSource: PostReceiptInitiationSource,
+    appInBackground: Boolean,
+) : UseCaseParams(appInBackground)
+
+internal class ConsumePurchaseUseCase(
+    private val useCaseParams: ConsumePurchaseUseCaseParams,
+    val onReceive: (purchaseToken: String) -> Unit,
+    val onError: PurchasesErrorCallback,
+    val withConnectedClient: (BillingClient.() -> Unit) -> Unit,
+    executeRequestOnUIThread: ExecuteRequestOnUIThreadFunction,
+) : BillingClientUseCase<String>(useCaseParams, onError, executeRequestOnUIThread) {
+
+    override val backoffForNetworkErrors: Boolean
+        get() = when (useCaseParams.initiationSource) {
+            PostReceiptInitiationSource.RESTORE,
+            PostReceiptInitiationSource.PURCHASE,
+            -> false
+            PostReceiptInitiationSource.UNSYNCED_ACTIVE_PURCHASES -> true
+        }
+
+    override val errorMessage: String
+        get() = "Error consuming purchase"
+
+    override fun executeAsync() {
+        withConnectedClient {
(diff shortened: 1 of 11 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/ConsumePurchaseUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+    val withConnectedClient: (BillingClient.() -> Unit) -> Unit,
+    executeRequestOnUIThread: ExecuteRequestOnUIThreadFunction,
+) : BillingClientUseCase<String>(useCaseParams, onError, executeRequestOnUIThread) {
+
+    override val backoffForNetworkErrors: Boolean
+        get() = when (useCaseParams.initiationSource) {
+            PostReceiptInitiationSource.RESTORE,
+            PostReceiptInitiationSource.PURCHASE,
+            -> false
+            PostReceiptInitiationSource.UNSYNCED_ACTIVE_PURCHASES -> true
```
tonidero, line 27: Hmm I've been thinking... Could it happen that we try to retry consuming a purchase with backoff, but then the user tries to restore in between, succeeds and we consume twice? I think that should be a no-op but it brings to mind whether we should retry consuming the purchase at all... Wdyt?

vegaro: @tonidero just tested it and it returns `ITEM_NOT_OWNED` which in this case will send a callback to the `onError`. 

What do you think? I think that's fine no?

tonidero: @vegaro Hmm I'm still wondering if it wouldn't be fine to just not retry consuming purchases and avoid this scenario entirely, specially considering we would retry on the next app foreground... It would also simplify the code I think. 

But if you feel strongly we should retry here, I would be ok with that.

purchases/src/test/java/com/revenuecat/purchases/google/usecase/ConsumePurchaseUseCaseTest.kt
```diff
@@ -0,0 +1,1244 @@
...
+        useCase.run()
+
+        assertThat(capturedDelays.size).isEqualTo(12)
+        assertThat(capturedDelays.last()).isCloseTo(RETRY_TIMER_MAX_TIME_MILLISECONDS, Offset.offset(1000L))
+        assertThat(receivedError).isNotNull
+        assertThat(receivedError!!.code).isEqualTo(PurchasesErrorCode.StoreProblemError)
+    }
+
+    @Test
+    fun `If service returns SERVICE_UNAVAILABLE, don't retry and error if user in session for unsynced active purchases`() {
```
tonidero, line 1015: Some of these tests are kinda duplicated among the use case tests... I guess we could try with paremeterized tests but not sure if it's worth trying to extract them for now.

tonidero (commented):
Code looks good, but want to perform some testing before approving

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/ConsumePurchaseUseCase.kt
vegaro: I think if there's a `NETWORK_ERROR`, which is transient, it's fine to retry specially when purchasing. Imagine a user is purchasing and right when we consume there's a `NETWORK_ISSUE`, they would need to reopen the app to retry consuming it. Retrying twice could maybe recover and consume the purchase, without the need to reopen.

Also, if after a failure to consume, when reopening the Billing client is still giving NETWORK_ERROR, we will retry with backoff. At some point in the future, the backoff retry could succeed. They can still retry by calling restore or reopening the app. One of those could succeed and give more opportunities to be able to consume the purchase.

tonidero: Ok makes sense. Thanks for explaining!

tonidero (approved):
Did some quick OTP tests and seems to be working fine!

PR #1488, opened 2023-11-20
Check diagnostics file size in background thread

tonidero:
### Description
We were checking the diagnostics file size in the main thread which was causing a disk read violation. This moves that operation to the diagnostics thread

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
@@ -45,10 +45,12 @@ internal class DiagnosticsSynchronizer(
     }
 
     fun clearDiagnosticsFileIfTooBig() {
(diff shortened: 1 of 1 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
```diff
@@ -45,10 +45,12 @@ internal class DiagnosticsSynchronizer(
     }
 
     fun clearDiagnosticsFileIfTooBig() {
-        if (diagnosticsFileHelper.isDiagnosticsFileTooBig()) {
-            verboseLog("Diagnostics file is too big. Deleting it.")
-            diagnosticsTracker.trackMaxEventsStoredLimitReached()
-            resetDiagnosticsStatus()
+        enqueue {
```
tonidero, line 48: It should be fine to enqueue since we only have a single diagnostics file, where the track methods are called. So if we have to delete the diagnostics file, it should happen before we add any more events.

vegaro (approved):

PR #1489, opened 2023-11-20
Initialize shared preferences lazily

tonidero:
### Description
We detected some `DiskReadViolation` caused by initilizing shared preferences. This should have a very low impact on initialization time, but might be good to delay that so it's not caused until actually needed. We currently have 3 shared preferences files, 2 of them are only used later on and can be initialized later.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
@@ -21,11 +21,12 @@ import java.util.stream.Collectors
  */
 @RequiresApi(Build.VERSION_CODES.N)
 internal class DiagnosticsSynchronizer(
+    context: Context,
     private val diagnosticsFileHelper: DiagnosticsFileHelper,
     private val diagnosticsTracker: DiagnosticsTracker,
     private val backend: Backend,
     private val diagnosticsDispatcher: Dispatcher,
-    private val sharedPreferences: SharedPreferences,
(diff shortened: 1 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
```diff
@@ -21,11 +21,12 @@ import java.util.stream.Collectors
  */
 @RequiresApi(Build.VERSION_CODES.N)
 internal class DiagnosticsSynchronizer(
+    context: Context,
     private val diagnosticsFileHelper: DiagnosticsFileHelper,
     private val diagnosticsTracker: DiagnosticsTracker,
     private val backend: Backend,
     private val diagnosticsDispatcher: Dispatcher,
-    private val sharedPreferences: SharedPreferences,
+    private val sharedPreferences: Lazy<SharedPreferences> = lazy { initializeSharedPreferences(context) },
```
tonidero, line 29: No need to initialize shared preferences here if not using diagnostics.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -49,7 +49,8 @@ internal data class HTTPResultWithETag(
 }
 
 internal class ETagManager(
-    private val prefs: SharedPreferences,
+    context: Context,
+    private val prefs: Lazy<SharedPreferences> = lazy { initializeSharedPreferences(context) },
```
tonidero, line 53: For etags we are only actually needing this for backend requests.

vegaro (approved):
Makes sense

PR #1490, opened 2023-11-22
Add retries with backoff and different treatment to SERVICE_UNAVAILABLE

vegaro:
Decided to split this up since it was adding up when working on the ConsumePurchaseUseCase

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -193,8 +193,8 @@ internal class BillingWrapper(
         productType: ProductType,
         productIds: Set<String>,
         appInBackground: Boolean,
-        onError: PurchasesErrorCallback,
         onReceive: StoreProductsCallback,
+        onError: PurchasesErrorCallback,
     ) {
         log(LogIntent.DEBUG, OfferingStrings.FETCHING_PRODUCTS.format(productIds.joinToString()))
         val useCase = QueryProductDetailsUseCase(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -94,7 +94,7 @@ internal class BillingWrapper(
     private val purchaseContext = mutableMapOf<String, PurchaseContext>()
 
     private val serviceRequests =
-        ConcurrentLinkedQueue<Pair<(connectionError: PurchasesError?) -> Unit, Long>>()
+        ConcurrentLinkedQueue<Pair<(connectionError: PurchasesError?) -> Unit, Long?>>()
 
     // how long before the data source tries to reconnect to Google play
     private var reconnectMilliseconds = RECONNECT_TIMER_START_MILLISECONDS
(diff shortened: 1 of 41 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
```diff
@@ -13,21 +13,36 @@ import com.revenuecat.purchases.google.toHumanReadableDescription
 import com.revenuecat.purchases.strings.BillingStrings
 import java.io.PrintWriter
 import java.io.StringWriter
+import kotlin.math.min
+
+internal typealias ExecuteRequestOnUIThreadFunction = (delay: Long, onError: (PurchasesError?) -> Unit) -> Unit
```
tonidero, line 18: NItpick: We could change the name `delay` -> `delayInMillis`?

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
```diff
@@ -96,4 +110,45 @@ internal abstract class BillingClientUseCase<T>(
...
+        } else {
+            onError(billingResult)
+        }
+    }
+
+    private fun backoffOrErrorIfUseInSession(
+        onError: (BillingResult) -> Unit,
+        billingResult: BillingResult,
+    ) {
+        if (!useCaseParams.appInBackground) {
```
tonidero, line 132: I guess this will pick the value when the request is initiated. As in, if the user backgrounds/foregrounds the app, this won't be updated. I guess this is probably ok though... I do wonder if there is a way to reset the request when this state changes, but maybe we can do that later.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -23,20 +22,21 @@ import java.util.Date
 import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.time.Duration
 
-internal data class QueryProductDetailsUseCaseParams(
```
tonidero, line 26: Hmm it's sad we won't get the `data class` anymore with its benefits... I wonder if we should make `UseCaseParams` an interface instead of a class. Then, these data classes just need to add the `override` for that property I think? We would need to create a new class for `GetBillingConfigUseCase` since that didn't have parameters before, but I think that's preferrable. Thoughts?

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
tonidero: Nitpick, inverse the condition here?

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
vegaro: perfect, I will do that

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -164,9 +170,9 @@ internal class BillingWrapper(
     }
 
     @Synchronized
-    private fun executeRequestOnUIThread(request: (PurchasesError?) -> Unit) {
+    private fun executeRequestOnUIThread(delayMilliseconds: Long = 0, request: (PurchasesError?) -> Unit) {
```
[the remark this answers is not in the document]
vegaro: ok, let's do that. I assumed post and postDelayed 0 is the same but let's do it just in case.

tonidero (approved):
It's a lot of changes, but I tested it for a bit and seems to work fine. As we talked, it might be a good idea to separate the `PurchasesState` to a different class that can be injected instead of passing the background state everywhere like that, but we can do that later.

PR #1493, opened 2023-11-23
Refactor findPurchaseInPurchaseHistory

vegaro:
Refactor findPurchaseInPurchaseHistory to use `QueryPurchaseHistoryUseCase`


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -461,33 +459,33 @@ internal class BillingWrapper(
...
-                        if (purchaseHistoryRecordWrapper != null) {
-                            onCompletion(purchaseHistoryRecordWrapper)
-                        } else {
-                            val message = PurchaseStrings.NO_EXISTING_PURCHASE.format(productId)
-                            val error = PurchasesError(PurchasesErrorCode.PurchaseInvalidError, message)
-                            onError(error)
-                        }
+        log(LogIntent.DEBUG, RestoreStrings.QUERYING_PURCHASE_WITH_TYPE.format(productId, productType.name))
+        productType.toGoogleProductType()?.let { googleProductType ->
+            QueryPurchaseHistoryUseCase(
```
tonidero, line 464: I wonder if we should move tests using `findPurchaseInPurchaseHistory` also to this use case's test class...

tonidero (approved):
Nice cleanup!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
vegaro: Let's do it

tonidero (approved):

PR #1495, opened 2023-11-23
Refactor QueryPurchasesUseCase and adds QueryPurchasesByTypeUseCase

vegaro:
We are using `BillingClient.queryPurchases` for `getPurchaseType`. I've realized the easiest to reuse `QueryPurchasesUseCase` for both was to split it into a `QueryPurchasesByTypeUseCase` that can get only purchases of a certain type.

This has the consequence that in `QueryPurchasesUseCase` we are now calling another `BillingClientUseCase` twice, meaning that we are posting to the main thread twice. I actually think this is an advantage so we split the work and treat the retries independently, but let me know what you think.


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchasesUseCase.kt
```diff
@@ -45,142 +26,58 @@ internal class QueryPurchasesUseCase(
...
+                productType = ProductType.SUBS,
+            ),
+            onSuccess = onQuerySubscriptionsSuccess,
+            onError = onQuerySubscriptionsError,
+            withConnectedClient = withConnectedClient,
+            executeRequestOnUIThread = executeRequestOnUIThread,
+        ).run()
     }
 
     override fun executeAsync() {
```
vegaro, line 64: I mentioned it in the description of the PR, but this change now has the consequence of posting to the main thread twice, once for SUBS and once for INAPPS. I don't think it's a bad change, but maybe we want to consider removing this UseCase and moving its "logic" (calling the `QueryPurchasesByTypeUseCase` twice) to the BillingWrapper

tonidero: Hmm I guess a few things to be aware of are:
- There is a chance of this request to get delayed if something else is added to the `serviceRequests` in between both calls. Probably not a big deal
- If the billing client disconnects in between, we might need to reconnect. As well, probably not a big deal. 

One thing to be aware of though, if we try to execute the request with a delay, the delay would be applied twice, once for querying subs and once for inapps. I don't think we are adding a delay to these right now, so we should be ok I think.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCase.kt
```diff
@@ -59,7 +59,7 @@ internal class QueryPurchaseHistoryUseCase(
                     }
                 }
             } ?: run {
-                errorLog(PurchaseStrings.INVALID_PRODUCT_TYPE.format("getPurchaseType"))
+                errorLog(PurchaseStrings.INVALID_PRODUCT_TYPE.format("queryPurchaseHistory"))
```
tonidero, line 62: Oh nice catch!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -502,51 +497,45 @@ internal class BillingWrapper(
...
+        productType: String,
+        purchaseToken: String,
+        listener: (ProductType) -> Unit,
+        resultHandler: (Boolean) -> Unit,
+    ) {
+        QueryPurchasesByTypeUseCase(
+            QueryPurchasesByTypeUseCaseParams(
+                dateProvider,
+                diagnosticsTrackerIfEnabled,
+                appInBackground = false,
```
tonidero, line 526: We're hardcoding this to false right now. Are you planning to pass it in on another PR?

tonidero (approved):
Left some questions but I think it should be safe to merge

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
vegaro: Oh this is not correct yeah. I will fix it.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchasesUseCase.kt
vegaro: > If the billing client disconnects in between, we might need to reconnect. As well, probably not a big deal.

The good thing about this is we already have the result of one of the calls, so only one of the calls needs to be retried. This can also happen right now btw, and we would have to retry both

> I don't think we are adding a delay to these right now, so we should be ok I think.

Yes... We are not applying backoff here so I think it should be fine yes.

vegaro: Oh actually, we would backoff if the app is in background and we get a ServiceUnavailable

tonidero: Ohhh right... I guess in that case this could get delayed quite a bit.

Then again, thinking about it, if this is in the background, it might be ok? As long as, once we come back to foreground, we don't have to wait the whole time (so we can enqueue multiple instances of this request and it's not an issue).

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
vegaro: ok, there's a bit of an issue here...

`getStoreTransaction` gets called from `onPurchasesUpdated`, which is a billing client interface, so we don't have access to `appInBackground`....

I think it's time to inject that `StateProvider`

PR #1496, opened 2023-11-24
Avoid creating threads when dispatching tasks

tonidero:
### Description
We were creating threads every time we added a task to the executor service. This shouldn't be a problem if tasks ended early and not too many tasks were in the queue, but it's not very efficient. This is another implementation to propagate exceptions to the main thread so they can be thrown and not swallowed when they happen on the thread pool.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt
@@ -62,23 +62,24 @@ internal open class Dispatcher(
     ) {
         synchronized(this.executorService) {
             if (!executorService.isShutdown) {
+                val commandHandlingExceptions = {
+                    try {
+                        command.run()
+                    } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                        errorLog("Exception running command: $e")
+                        mainHandler?.post {
+                            e.cause?.let { throw it }
+                        }
+                    }
+                }
                 if (delay != Delay.NONE && executorService is ScheduledExecutorService) {
                     var delayToApply = (delay.minDelay.inWholeMilliseconds..delay.maxDelay.inWholeMilliseconds).random()
                     if (runningIntegrationTests) {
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt
```diff
@@ -59,28 +62,24 @@ internal open class Dispatcher(
...
-                    } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
-                        e.cause?.let { throw it }
+                    executorService.submit {
+                        try {
+                            command.run()
+                        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                            errorLog("Exception running command: $e")
+                            mainHandler?.post {
+                                e.cause?.let { throw it }
+                            }
```
tonidero, line 79: Actually, need to do this as well when there is a delay

purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt
```diff
@@ -59,28 +62,25 @@ internal open class Dispatcher(
...
         synchronized(this.executorService) {
             if (!executorService.isShutdown) {
-                val future = if (delay != Delay.NONE && executorService is ScheduledExecutorService) {
+                val commandHandlingExceptions = Runnable {
+                    try {
+                        command.run()
+                    } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                        errorLog("Exception running command: $e")
+                        mainHandler?.post {
+                            throw e.cause ?: e
```
vegaro, line 71: can you explain this? why not just throw e?

tonidero: Probably not needed, I was mostly copying here what we had previously... In my tests, the cause was sometimes null, so I fallback to the original exception, but we could probably just propagate the original exception

vegaro: Yes, I am not sure why we were doing `e.cause` 🤔

vegaro (approved):

PR #1498, opened 2023-11-27
Fix integration tests

tonidero:
### Description
Looks like integration tests broke in the recent billing wrapper refactors. We didn't notice since these run nightly on main. This fixes them


vegaro (approved):
Thanks for fixing these

PR #1499, opened 2023-11-27
Fix white line around buttons in Template 4

vegaro:
From:

Screenshot 2023-11-27 at 13 47 34

To:

Screenshot 2023-11-27 at 13 47 18

I fixed it by adding the border to the surrounding box and a very small padding to the button. I think the line comes originally from antialising.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template4.kt
```diff
@@ -274,10 +274,20 @@ private fun SelectPackageButton(
...
+                .border(
+                    width = UIConstant.defaultPackageBorderWidth,
+                    color = mainColor,
+                    shape = RoundedCornerShape(UIConstant.defaultCornerRadius),
+                ),
+        ) {
             Button(
                 modifier = modifier
                     .alpha(buttonAlpha)
+                    .padding(1.dp) // Trick to prevent white line around the button border
```
tonidero, line 290: I wonder if it would work with 1px instead of 1dp... You can convert pixels to dp with this I think: https://developer.android.com/reference/kotlin/androidx/compose/ui/unit/Density#(kotlin.Int).toDp()

vegaro: it worked! perfect

PR #1504, opened 2023-11-29
Improve price per month accuracy for weekly subscriptions

aboedo:
We were calculating the price per month for weekly products as price * 4, but there are on average 4.345 weeks in a month, so this updates the value accordingly.


aboedo:
looks like snapshots are failing but I can't see them, I imagine we had a value that was calculated off of the weekly in the snapshot?

purchases/src/test/java/com/revenuecat/purchases/models/PeriodTest.kt
```diff
@@ -9,7 +9,7 @@ import org.junit.runner.RunWith
 @RunWith(AndroidJUnit4::class)
 class PeriodTest {
     companion object {
-        private val MAX_OFFSET = Offset.offset(0.0000001)
+        private val MAX_OFFSET = Offset.offset(0.0001)
```
aboedo, line 12: don't love reducing accuracy, but having a value this precise would probably risk our tests failing from floating point precision errors

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
```diff
@@ -31,9 +31,9 @@ data class Period(
         private const val DAYS_PER_WEEK = 7.0
         private const val DAYS_PER_MONTH = 30.0
         private const val DAYS_PER_YEAR = 365.0
-        private const val WEEKS_PER_MONTH = 4.0
         private const val WEEKS_PER_YEAR = 52.14
```
vegaro, line 34: Should we also calculate this one so we have the same precision?

vegaro (approved):
Makes sense! Just that little question

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
aboedo: I was shooting for "least amount of changes possible" but we could.

It'd be nice to update them all really to also take into account leap years

tonidero (approved):

tonidero:
> looks like snapshots are failing but I can't see them, I imagine we had a value that was calculated off of the weekly in the snapshot?

@aboedo I generated snapshots [here](https://github.com/RevenueCat/purchases-android-snapshots/pull/9/files) and noticed that the weekly price was actually more expensive than the others with this new calculation, which caused a changed in the discount percentages. I thought to merge the new snapshots, but for now I decided to slightly lower the price of the weekly product so it's not more expensive than the monthly one. I [pushed that change](https://github.com/RevenueCat/purchases-android/pull/1504/commits/8323803d308bedee169254eae159d94910add21b) in this branch. Lmk if you would prefer to update the snapshots and we can do that.

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
tonidero: Will make that change separately.

PR #1507, opened 2023-11-30
Paywalls: Auto-close paywall activity if restore grants required entitlement identifier

tonidero:
### Description
Users could potentially restore their purchases and get entitlements granted but the paywall activity was not dismissing. This will add that functionality in the case where users use the `requiredEntitlementIdentifier` parameter to launch the activity.

This adds a new `PaywallResult` value, `Restored`. This will be returned if the last action the user performed was a restore.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
@@ -101,8 +101,8 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
     }
 
     override fun onRestoreCompleted(customerInfo: CustomerInfo) {
-        val requiredEntitlementIdentifier = getArgs()?.requiredEntitlementIdentifier ?: return
         setResult(RESULT_OK, createResultIntent(PaywallResult.Restored(customerInfo)))
+        val requiredEntitlementIdentifier = getArgs()?.requiredEntitlementIdentifier ?: return
         if (customerInfo.entitlements.active.containsKey(requiredEntitlementIdentifier)) {
             finish()
         }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
@@ -102,8 +102,8 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
 
     override fun onRestoreCompleted(customerInfo: CustomerInfo) {
         val requiredEntitlementIdentifier = getArgs()?.requiredEntitlementIdentifier ?: return
+        setResult(RESULT_OK, createResultIntent(PaywallResult.Restored(customerInfo)))
         if (customerInfo.entitlements.active.containsKey(requiredEntitlementIdentifier)) {
-            setResult(RESULT_OK, createResultIntent(PaywallResult.Restored(customerInfo)))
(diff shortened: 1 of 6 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -100,6 +100,14 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
...
     }
 
+    override fun onRestoreCompleted(customerInfo: CustomerInfo) {
+        setResult(RESULT_OK, createResultIntent(PaywallResult.Restored(customerInfo)))
+        val requiredEntitlementIdentifier = getArgs()?.requiredEntitlementIdentifier ?: return
+        if (customerInfo.entitlements.active.containsKey(requiredEntitlementIdentifier)) {
+            finish()
+        }
+    }
+
```
vegaro, line 110: hmmm this is making me think... if anyone is already doing this (finishing the activity when they get a `PaywallResult.Restored`, this will break their implementation. Right? Are we doing this in iOS?

tonidero: `PaywallResult.Restored` is a new sealed class element. Until now, I don't think users could detect whether the activity had restored or not.

As for iOS. I do think we have a `didFinishRestoringWith` delegate developers can use for this but I don't think it autocloses. That's something developers would need to handle themselves.

vegaro: 😮‍💨 ok, should have kept reading

sorry for the noise

vegaro: I think we can autoclose and if anyone complains, we can add a setting somehow. I think autoclosing is a better default experience

tonidero: Well, in iOS I think we are not autoclosing for now. Not on a normal purchase nor on a restore. We talked about this before, and we felt it was more in line with iOS to not autodismiss. CC @aboedo @NachoSoto for confirmation.

But in android, since the developer doesn't have good access to the paywall activity, nor a way to commuinicate with it once the paywall activity is launched, we need to handle it for them.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -70,12 +70,19 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
+        val shouldDisplayBlock = shouldDisplayBlockForEntitlementIdentifier(requiredEntitlementIdentifier)
+        shouldDisplayPaywall(shouldDisplayBlock) { shouldDisplay ->
+            if (shouldDisplay) {
+                activityResultLauncher.launch(
+                    PaywallActivityArgs(
+                        requiredEntitlementIdentifier = requiredEntitlementIdentifier,
+                        offeringId = offering?.identifier,
+                        fontProvider = fontProvider,
+                        shouldDisplayDismissButton = shouldDisplayDismissButton,
+                    ),
```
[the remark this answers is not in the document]
tonidero: Yeah... and we can't really return an activity directly, since we only could have an intent. The system will be in charge of creating the activity.

NachoSoto (commented):
This matches the behavior in iOS 👍🏻

PR #1509, opened 2023-11-30
Paywalls: Add `PaywallFooterView`

tonidero:
### Description
This adds a new view `PaywallFooterView` which acts as a wrapper for the `PaywallFooter` composable. It has some limitations:
- Condensed form not supported
- Need to set most of the attributes on the construction of the view.

This view will be used in the hybrids in order to support footer view.
image

### DIFF
--- a/examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
+++ b/examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
@@ -19,7 +19,7 @@
                 android:layout_height="wrap_content"/>
     </LinearLayout>
 
-    <com.revenuecat.purchases.ui.revenuecatui.fragment.PaywallFooterView
+    <com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView
             android:id="@+id/paywall_footer_view"
             android:layout_marginTop="-20dp"
             android:fontFamily="@font/lobster_two"
--- a/examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
+++ b/examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
@@ -7,6 +7,7 @@
         tools:context=".PaywallFooterViewActivity">
 
     <LinearLayout
+            android:background="@android:color/holo_blue_light"
             android:layout_weight="1"
             android:layout_width="match_parent"
             android:layout_height="0dp"
@@ -20,7 +21,9 @@
 
     <com.revenuecat.purchases.ui.revenuecatui.fragment.PaywallFooterView
             android:id="@+id/paywall_footer_view"
+            android:background="@android:color/holo_blue_light"
             android:fontFamily="@font/lobster_two"
+            android:layout_marginBottom="-10dp"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"/>
 
--- a/examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
+++ b/examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:tools="http://schemas.android.com/tools"
+        android:orientation="vertical"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        tools:context=".PaywallFooterViewActivity">
+
+    <LinearLayout
+            android:layout_weight="1"
+            android:layout_width="match_parent"
+            android:layout_height="0dp"
+            android:orientation="vertical"
+            android:gravity="center">
+        <TextView
+                android:text="Paywall Content"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"/>
+    </LinearLayout>
+
+    <com.revenuecat.purchases.ui.revenuecatui.fragment.PaywallFooterView
+            android:id="@+id/paywall_footer_view"
+            android:fontFamily="@font/lobster_two"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"/>
+
+</LinearLayout>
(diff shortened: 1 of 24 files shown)


examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
```diff
@@ -0,0 +1,30 @@
...
+        <TextView
+                android:text="Paywall Content"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"/>
+    </LinearLayout>
+
+    <com.revenuecat.purchases.ui.revenuecatui.fragment.PaywallFooterView
+            android:id="@+id/paywall_footer_view"
+            android:background="@android:color/holo_blue_light"
+            android:fontFamily="@font/lobster_two"
```
tonidero, line 25: Right now I'm hardcoding the font, so we can test it

examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
```diff
@@ -0,0 +1,30 @@
...
+            android:gravity="center">
+        <TextView
+                android:text="Paywall Content"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"/>
+    </LinearLayout>
+
+    <com.revenuecat.purchases.ui.revenuecatui.fragment.PaywallFooterView
+            android:id="@+id/paywall_footer_view"
+            android:background="@android:color/holo_blue_light"
```
tonidero, line 24: Note how I'm setting the same background color as the view above here. This is so it looks like the view above continues behind the footer. But the positioning of the views will depend on the developer with this approach

ui/revenuecatui/src/main/res/values/paywall_view_attrs.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <declare-styleable name="PaywallFooterView">
+        <attr name="android:fontFamily" />
+        <attr name="offeringIdentifier" format="string" />
```
tonidero, line 5: These are the parameters that can be passed from the xml. This is limited, some of them can only be passed through the view.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/fragment/PaywallFooterView.kt
```diff
@@ -0,0 +1,137 @@
...
+     */
+    fun setPaywallListener(listener: PaywallListener?) {
+        this.listener = listener
+    }
+
+    /**
+     * Sets a dismiss handler which will be called when the user successfully purchases or if there is an error
+     * loading the offerings and the user clicks through the error dialog.
+     */
+    fun setDismissHandler(dismissHandler: (() -> Unit)?) {
```
tonidero, line 81: I wondered whether to remove this for the footer view, since there is already a purchase completed listener in the `PaywallListener`. But there can still be an error loading the offerings and we show an error dialog in that case, In that case, we want to dismiss it, and this would be the only way to do it.

examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
```diff
@@ -0,0 +1,29 @@
...
+        android:orientation="vertical"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        tools:context=".PaywallFooterViewActivity">
+
+    <LinearLayout
+            android:background="@android:color/holo_blue_light"
+            android:layout_weight="1"
+            android:layout_width="match_parent"
+            android:layout_height="0dp"
```
vegaro, line 13: Why is this 0dp?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
```diff
@@ -29,28 +30,38 @@ import androidx.compose.ui.tooling.preview.Preview
 fun PaywallFooter(
     options: PaywallOptions,
     condensed: Boolean = false,
-    mainContent: @Composable (PaddingValues) -> Unit,
+    mainContent: @Composable ((PaddingValues) -> Unit)? = null,
```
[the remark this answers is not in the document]
vegaro: I am a bit confused about this change. When would someone not want to set something? This is a footer anywyas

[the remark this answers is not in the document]
tonidero: Right... The main reason for this change was that the `Column` composable inside has a `fillMaxSize` modifier, which we don't want when we want to render only the footer view. In native android, we probably want people to pass us a composable, but when using the `PaywallFooterView`, we can't easily pass composables as a parameter to the view (most people using the view are probably not using compose).

An alternative to this would be to add a parameter, `useMaxSize` or something like that that tells us whether this composable should use all available space. I kinda didn't want to add another parameter if possible, but it might be less error prone than making the content nullable... wdyt?

examples/paywall-tester/src/main/res/layout/activity_paywall_footer_view.xml
tonidero: If you set the `layout_weight` parameter, you usually need to set one of the dimensions with 0dp so it can grow from that dimension. This is basically telling it to use all the available space after removing the size of other sibling views that don't use `layout_weight`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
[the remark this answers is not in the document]
vegaro: Hmm, I see... I am seeing now that `mainContent` is a confusing parameter. It makes it feel as if it's the content of the Footer, and not whatever is _behind_ the Footer

I am honestly leaning towards creating two views, the current `PaywallFooter` should be something like `PaywallFooterWithScreenContent` and a new `PaywallFooter` shouldn't paint anything else and not accept a `mainContent`

We can discuss this in our weekly sync I think

vegaro (approved):

PR #1510, opened 2023-12-01
Paywalls: Add `PaywallView`

tonidero:
### Description
Full screen counterpart to #1509. This adds a new view `PaywallView` which wraps the `Paywall` composable so it's usable in the view system. 

The view will be used in the hybrids.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
@@ -137,7 +137,7 @@ class PaywallView : FrameLayout {
                 recycle()
             }
         }
-        fontFamilyId?.let {
+        fontFamilyId?.takeIf { it > 0 }?.let {
             val typeface = ResourcesCompat.getFont(context, it)
             if (typeface == null) {
                 Logger.e("Font given for PaywallView not found")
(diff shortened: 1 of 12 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -0,0 +1,151 @@
...
+                shouldShowCloseButton = if (hasValue(R.styleable.PaywallView_shouldDisplayDismissButton)) {
+                    getBoolean(R.styleable.PaywallView_shouldDisplayDismissButton, false)
+                } else {
+                    null
+                }
+            } finally {
+                recycle()
+            }
+        }
+        fontFamilyId?.let {
```
vegaro, line 140:
```suggestion
        fontFamilyId?.takeIf { it > 0 }?.let {
```

vegaro (approved):
There is some code shared between `PaywallView` and `PaywallFooterView`. I was wondering if it would be worth it to extract it to a common class. For example, the bug with the font family resource id being 0 could have been fixed in both views at the same time if we had that code somewhere else that's common for both. But I am not sure what's the best strategy to share that code.

tonidero:
> There is some code shared between `PaywallView` and `PaywallFooterView`. I was wondering if it would be worth it to extract it to a common class. For example, the bug with the font family resource id being 0 could have been fixed in both views at the same time if we had that code somewhere else that's common for both. But I am not sure what's the best strategy to share that code.

Ohh right... There is one more parameter currently in the full screen view than that footer view, and this uses ids, but I will think of ways to share that, thanks!

vegaro:
@tonidero can you take a look at the merge conflicts resolution?

tonidero:
Took a look, makes sense! :shipit:

NachoSoto:
Awesome!

PR #1511, opened 2023-12-01
Paywalls: Remove `PaywallActivity` theme to pickup application's theme by default

tonidero:
### Description
Remove `PaywallActivity`'s theme in order to pickup the application's theme if set.


vegaro (approved):
Makes sense!!

PR #1519, opened 2023-12-06
Paywalls: fix empty description when using custom package type and Offer Period

aboedo:
Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/3495

If you select a custom package type and set Offer Period, the paywall would have an empty entry for that package.

image

| Before | After |
| :-: | :-: |
| Screenshot_20231206-144952 | Screenshot_20231206-151434 |


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -50,6 +50,9 @@ internal class VariableDataProvider(
     }
 
     fun periodName(rcPackage: Package): String? {
+        if (rcPackage.packageType == PackageType.CUSTOM) {
```
NachoSoto, line 53: Ditto to iOS: `UNKNOWN` should use this too.

aboedo: updated here as well, thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -50,6 +50,10 @@ internal class VariableDataProvider(
     }
 
     fun periodName(rcPackage: Package): String? {
```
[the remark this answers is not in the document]
NachoSoto: Yeah that would be the same approach as in iOS 👍🏻

PR #1521, opened 2023-12-07
Wait for tests to pass when deploying

vegaro:
I noticed it's possible to deploy without waiting for all tests to pass. It's probably an edge case but it's possible we merge something into `main` that breaks unit tests then start a release from that broken commit.

I added a new hold job to force the release in case we actually don't want to wait for tests to pass. The name is pretty clear `force-release-without-waiting-for-tests` so I think people will understand that it's skipping tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -600,18 +600,6 @@ workflows:
     jobs:
       - revenuecat/danger
 
-  build-test:
-    when:
-      not:
-        equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
-    jobs:
-      - test
-      - detekt
-      - assemble-purchase-tester
-      - run-backend-integration-tests
-      - verify-revenuecatui-snapshots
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -650,15 +644,26 @@ workflows:
...
             - run-firebase-tests-purchases-custom-entitlement-computation-integration-test
           <<: *release-branches
+      - force-release-without-waiting-for-tests:
+          type: approval
+          <<: *release-branches
       - revenuecat/tag-current-branch:
           requires:
             - hold
           <<: *release-branches
+      - revenuecat/tag-current-branch:
```
tonidero, line 663: Haven't checked, but I guess it's ok to have 2 of the same job in the same workflow but with different requirements?

tonidero (approved):
I guess this new workflow will end up pretty big... but I think it should be fine 👍

PR #1526, opened 2023-12-14
Add revenuecatui docs to reference docs

tonidero:
### Description
Add the `revenuecatui` module to generated documentation. We already had a multi-module setup, so it was straightforward to add it.

image


NachoSoto (approved):
Yay!! 😻

PR #1528, opened 2023-12-18
Disable close button when action is in progress

vegaro:
Equivalent of https://github.com/RevenueCat/purchases-ios/pull/3474

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
@@ -73,6 +73,7 @@ internal fun LoadingPaywall(
         validatedPaywallData = paywallData,
         template = LoadingPaywallConstants.template,
         shouldDisplayDismissButton = shouldDisplayDismissButton,
+        actionInProgress = false,
     )
 
     when (state) {
@@ -110,7 +111,7 @@ private fun LoadingPaywall(
                     ),
             )
         }
-        CloseButton(state.shouldDisplayDismissButton, onDismiss)
+        CloseButton(state.shouldDisplayDismissButton, state.actionInProgress, onDismiss)
     }
 }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
@@ -110,7 +110,11 @@ private fun LoadingPaywall(
                     ),
             )
         }
-        CloseButton(state.shouldDisplayDismissButton, viewModel.actionInProgress.value, onDismiss)
+        CloseButton(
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -271,6 +271,7 @@ internal class PaywallViewModelImpl(
             validatedPaywallData = displayablePaywall,
             template = template,
             shouldDisplayDismissButton = options.shouldDisplayDismissButton,
+            actionInProgress = actionInProgress.value,
```
vegaro, line 274: I added this here, then realized that we need to update the state in `verifyNoActionInProgressOrStartAction()`, where `actionInProgress` is updated. I am not sure there's a more efficient way of doing it. The alternative would be to update the button wherever the `PaywallListener` is implemented

What do you think @tonidero

tonidero: Hmm we're already exposing `actionInProgress` from the `PaywallViewModel` as a different `State`. Couldn't we just use that in the `InternalPaywall` directly? I believe in that case this wiring won't be necessary? Do lmk if I'm missing something!

vegaro: oh that's right, I missed we were exposing that as a state already

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
```diff
@@ -110,7 +110,7 @@ private fun LoadingPaywall(
                     ),
             )
         }
-        CloseButton(state.shouldDisplayDismissButton, onDismiss)
+        CloseButton(state.shouldDisplayDismissButton, viewModel.actionInProgress.value, onDismiss)
```
NachoSoto, line 113: Maybe add labels to the parameters here and in `InternalPaywall`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/CloseButton.kt
```diff
@@ -10,11 +10,16 @@ import androidx.compose.ui.res.painterResource
...
+fun BoxScope.CloseButton(
+    shouldDisplayDismissButton: Boolean,
+    actionInProgress: Boolean,
+    onClick: () -> Unit,
+) {
     if (shouldDisplayDismissButton) {
         IconButton(
             onClick = onClick,
             modifier = Modifier.align(Alignment.TopStart),
+            enabled = actionInProgress.not(),
```
NachoSoto, line 22: So nice

NachoSoto (approved):
Let's release this?

tonidero (approved):
Looks great!

PR #1529, opened 2023-12-20
Paywalls: improve Chinese localization

aboedo:
Android equivalent to https://github.com/RevenueCat/purchases-ios/pull/3489/

I noticed that we don't have a separate file for Chinese Simplified on Android? Is this intentional?

### DIFF
--- a/ui/revenuecatui/src/main/res/values-zh-rHK/strings.xml
+++ b/ui/revenuecatui/src/main/res/values-zh-rHK/strings.xml
@@ -2,20 +2,20 @@
 <resources>
     <string name="OK">好的</string>
     <string name="privacy">隱私</string>
-    <string name="privacy_policy">隱私政策</string>
+    <string name="privacy_policy">隱私權政策</string>
     <string name="restore">恢復</string>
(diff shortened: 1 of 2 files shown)


tonidero (approved):
This looks good to me! As for why we don't have simplified chinese... I'm not sure... @NachoSoto ?

NachoSoto:
I automated the import of languages from iOS. Let me check.

NachoSoto:
I'm not familiar, would it be `zh`?
Screenshot 2023-12-15 at 15 44 29

NachoSoto (approved):

greenietea (approved):

greenietea:
> I'm not familiar, would it be `zh`? Screenshot 2023-12-20 at 10 18 51

both traditional and simplified are represented as zh. 

are we meant to also have another file named: `ui/revenuecatui/src/main/res/values-zh-rHans/strings.xml` where we keep the simplified version?

aboedo:
I can tackle adding it as a separate PR, thanks!

PR #1537, opened 2023-12-22
Paywalls: add Chinese from mainland China

aboedo:
Added Chinese from mainland China. 

Notes: 

- I_have_no_idea_what_im_doing.gif
- There isn't an exact equivalent to iOS's zh_Hans in Android Studio, it allows you to set `zh_CN` and `zh_SG`. ChatGPT suggested adding those two as the equivalent. I added only one, under the assumption that zh_SG would automatically pick up zh_CN just like what happens in different variations of Spanish or English. 
- I added the entries that we have on iOS for Chinese Simplified, updating format where needed.

### DIFF
--- a/ui/revenuecatui/src/main/res/values-zh-rCN/strings.xml
+++ b/ui/revenuecatui/src/main/res/values-zh-rCN/strings.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="OK">好的</string>
+    <string name="privacy">隐私</string>
+    <string name="privacy_policy">隐私政策</string>
+    <string name="restore">恢复</string>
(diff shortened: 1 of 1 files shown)


tonidero (approved):

vegaro (approved):

aboedo:
@RevenueCat/coresdk one thing I realized is that we don't have an android equivalent to this iOS key: https://github.com/RevenueCat/purchases-ios/blob/main/RevenueCatUI/Resources/zh-Hans.lproj/Localizable.strings#L5 do we not provide this behavior by default?

NachoSoto (approved):
Makes sense 🙏🏻

NachoSoto:
And yeah I think we don't have that alert in Android.

PR #1538, opened 2023-12-26
`Paywalls`: `PurchaseCancelledError` sends `onPurchaseCancelled` instead of an error

NachoSoto:
Cancelling the purchase modal shouldn't display an error dialog.

This is a very confusing experience, especially because that error wouldn't be localized.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
@@ -10,6 +10,7 @@ interface PaywallListener {
     fun onPurchaseStarted(rcPackage: Package) {}
     fun onPurchaseCompleted(customerInfo: CustomerInfo, storeTransaction: StoreTransaction) {}
     fun onPurchaseError(error: PurchasesError) {}
+    fun onPurchaseCancelled() {}
     fun onRestoreStarted() {}
     fun onRestoreCompleted(customerInfo: CustomerInfo) {}
     fun onRestoreError(error: PurchasesError) {}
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -205,9 +205,11 @@ internal class PaywallViewModelImpl(
             } catch (e: PurchasesException) {
                 if (e.code == PurchasesErrorCode.PurchaseCancelledError) {
                     trackPaywallCancel()
+                    listener?.onPurchaseCancelled()
```
vegaro, line 208: Awesome

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
```diff
@@ -10,6 +10,7 @@ interface PaywallListener {
     fun onPurchaseStarted(rcPackage: Package) {}
     fun onPurchaseCompleted(customerInfo: CustomerInfo, storeTransaction: StoreTransaction) {}
     fun onPurchaseError(error: PurchasesError) {}
+    fun onPurchaseCancelled() {}
```
vegaro, line 13: This is a breaking change, but it's marked as experimental, so it's probably fine to release as minor

vegaro (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
vegaro: Actually, it has a default implementation so we are good!

NachoSoto: Talked about this offline. This has a default implementation so it's optional for developers.

tonidero: Hmm while this is an easier API, I wonder if it might be confusing since `PurchasesError` also has the `PurchaseCancelledError` code, which will be unused now...

So this could be considered a breaking change for people that have been listening to that instead.

NachoSoto: That's fair, but luckily Android paywalls is still marked as experimental so we could fix this on time 😇

PR #1542, opened 2024-01-02
PaywallActivityLauncher: Add callback indicating whether the paywall was displayed when presented conditionally to an entitlement identifier

tonidero:
### Description
In the hybrid SDKs, we want to return a result from the `presentPaywall`/`presentPaywallIfNeeded` methods with the result of the paywall. However, we can't do that easily since `presentPaywallIfNeeded` might not present the paywall at all, and there was no way to know whether the paywall was displayed in that case until now.

This adds a new parameter to the `launchIfNeeded` methods that receive a `requiredEntitlementIdentifier` that will indicate whether the paywall was displayed. Note that this isn't added to the `launchIfNeeded` method that received a block of code, since the client should know in that case whether the paywall was displayed or not, depending on the result of that callback.

### DIFF
--- a/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
+++ b/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
@@ -108,9 +108,9 @@ private class PaywallActivityLauncherAPI {
     }
 
     fun checkPaywallDisplayCallback() {
+        @Suppress("EmptyFunctionBlock")
         val paywallDisplayCallback = object : PaywallDisplayCallback {
-            override fun onPaywallShouldDisplay(shouldDisplayPaywall: Boolean) {
-            }
+            override fun onPaywallShouldDisplay(shouldDisplayPaywall: Boolean) {}
         }
     }
 }
--- a/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
+++ b/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
@@ -7,6 +7,7 @@ import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityLauncher
+import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallDisplayCallback
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallResultHandler
 import com.revenuecat.purchases.ui.revenuecatui.fonts.ParcelizableFontProvider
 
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -18,6 +18,14 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayPaywall
 @ExperimentalPreviewRevenueCatUIPurchasesAPI
 interface PaywallResultHandler : ActivityResultCallback<PaywallResult>
 
+/**
+ * Implement this interface to receive whether the paywall was displayed when it depends on a condition.
+ */
+@ExperimentalPreviewRevenueCatUIPurchasesAPI
+interface PaywallDisplayCallback {
+    fun onPaywallShouldDisplay(shouldDisplayPaywall: Boolean)
```
NachoSoto, line 26: This sounds very confusing. The parameter is named as a question instead of a noun, as if it was a function. How about:
```suggestion
    fun onPaywallDisplay(wasDisplayed: Boolean)
```

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
```diff
@@ -90,4 +106,11 @@ private class PaywallActivityLauncherAPI {
             true
         }
     }
+
+    fun checkPaywallDisplayCallback() {
+        @Suppress("EmptyFunctionBlock")
+        val paywallDisplayCallback = object : PaywallDisplayCallback {
+            override fun onPaywallShouldDisplay(shouldDisplayPaywall: Boolean) {}
```
vegaro, line 113: Not needed but wanted to suggest it just in case you think it's a good idea. What do you think about returning a `DisplayPaywallResult` in case we want to add more stuff to the result in the future?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Yeah, I was unsure naming this... `onPaywallDisplay` is not fully correct, since it's possible the paywall was not displayed so I had some back and forth in my mind on this... What do you think of `fun onPaywallDisplayResult(wasDisplayed: Boolean)`?

vegaro (approved):
LGTM, apart from @NachoSoto comment about naming

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
tonidero: Hmm it's a good point... Do you have any idea of anything else we might want to return? I think right now a simple boolean suffices so I would prefer not to complicate the API without need.

vegaro: Sounds good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: I did this rename in https://github.com/RevenueCat/purchases-android/pull/1542/commits/2fad81932bee82b4e94091d5d65e2ac252ff52e1. Lmk what you think!

vegaro (approved):

PR #1545, opened 2024-01-03
Remove scroll in Template 2 and 5 in Footer mode

vegaro:
This was making https://github.com/RevenueCat/react-native-purchases/pull/799/files#diff-8eaf252c074773fc7f002e444b255a9c2c60b23a2c75448799b85abb2f509bf9R20 crash when measuring the size giving this error:

image

I removed the scroll in Footer mode because it's not needed


tonidero (approved):
Makes sense. Thanks for fixing this!

PR #1551, opened 2024-01-08
`Paywalls`: improve landscape layout on all templates

NachoSoto:
## Snapshots:

https://github.com/RevenueCat/purchases-android-snapshots/pull/10/files


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
```diff
@@ -1,6 +1,26 @@
...
+
+/**
+ * @return: Modifier with a callback indicating whether landscape layout should be used.
+ */
+@SuppressLint("ModifierFactoryExtensionFunction")
+internal fun PaywallState.Loaded.onLandscapeLayoutChanged(
+    changed: (Boolean) -> Unit,
+) = Modifier
+    .then(
+        // TODO: vertical compact?
```
NachoSoto, line 24: @tonidero I'd love your feedback on this.

On iOS we don't use the aspect ratio to determine landscape layout, but instead only use it if the "vertical size is compact" (which is basically only true for phones + landscape, but not landscape tablets).

Should we have some threshold screen height to make this false so landscape tablets still use the portrait layout?

NachoSoto:
Also @tonidero looks like the landscape snapshot tests don't have the new layout. Is that because of the limitation we talked a about where the snapshot is taken before the re-layout when the `remember` value changes?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
```diff
@@ -1,6 +1,26 @@
...
     get() = width.toFloat() / height
+
+internal val IntSize.isLandscape: Boolean
+    get() = width > height
+
+/**
+ * @return: Modifier with a callback indicating whether landscape layout should be used.
+ */
+@SuppressLint("ModifierFactoryExtensionFunction")
+internal fun PaywallState.Loaded.onLandscapeLayoutChanged(
```
tonidero, line 20: Note that changing from/to landscape in a device will trigger a configuration change and recreate the activity, and the whole UI will be recomputed (only the view model will survive). So not sure if this is needed?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensions.kt
tonidero: Hmm in Android, we have [window size classes](https://github.com/RevenueCat/purchases-android/blob/79e9c6bc6e9536dab13296723a603d28da186692/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/WindowHelper.kt#L11). We do use that to calculate whether to display as a full screen dialog or not when presenting the paywall as a dialog. I think we might be able to use something like that?

In that example, I'm using the width size class, but we should also be able to calculate the height size class. As for the accuracy... It's provided by Google, so it should work, but we should test on different devices just in case.

tonidero:
> Also @tonidero looks like the landscape snapshot tests don't have the new layout. Is that because of the limitation we talked a about where the snapshot is taken before the re-layout when the `remember` value changes?

Since we are using `onGloballyPositioned` to recalculate things, that's a possible explanation yeah... But as I mentioned in one comment, I think I would try to avoid having to recompute that on the fly, since changes in rotation will recreate the activity and recalculate all the composables anyway.

NachoSoto:
@tonidero this is ready again. Just gotta fix a small bug on template 4 and update snapshots.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/snapshottests/BasePaparazziTest.kt
```diff
@@ -25,7 +25,7 @@ data class TestConfig(
  *
  * ### Automation:
  * - To run them locally you need:
- * `bundle exec fastlane verify_revenuecatui_snapshot_tests`
+ * `bundle exec fastlane verify_revenuecatui_snapshots`
```
NachoSoto, line 28: This was wrong

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/IntSizeExtensionsTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import androidx.compose.ui.unit.IntSize
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.data.Offset
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class IntSizeExtensionsTest {
```
NachoSoto, line 11: Added missing tests.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -451,8 +511,8 @@ private fun Template5PaywallPreview() {
 }
 
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
-@Preview(showBackground = true, locale = "en-rUS")
-@Preview(showBackground = true, locale = "es-rES")
+@Preview(showBackground = true, locale = "en-rUS", group = "footer")
```
NachoSoto, line 514: This allows showing only a subset of previews which is super useful.

tonidero: Nice! We definitely had a ton of previews, this will help!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/WindowHelper.kt
```diff
@@ -1,18 +1,69 @@
...
+    return shouldUseLandscapeLayout(mode = templateConfiguration.mode)
+}
+
+@Composable
+@ReadOnlyComposable
+private fun shouldUseLandscapeLayout(mode: PaywallMode): Boolean {
+    return shouldUseLandscapeLayout(mode, computeWindowHeightSizeClass())
+}
+
+@VisibleForTesting
```
tonidero, line 40: This could be:
```suggestion
@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
This way the IDE will warn if we try to use it.

tonidero (approved):
Nice work on this!

PR #1554, opened 2024-01-11
Fix AmazonConfiguration for Java projects

vegaro:
For some strange reason, trying to do the following from a JAVA class wouldn't work:

```
 PurchasesConfiguration.Builder builder = new AmazonConfiguration.Builder(this, "");
``` 

This either:

```
 AmazonConfiguration.Builder builder = new AmazonConfiguration.Builder(this, "");
 builder.build();
 ```
 
I tried a bunch of things and realized a couple things:
- Moving `PurchasesConfiguration.Builder` out of`PurchasesConfiguration` fixes it
- The code works if used within the same project, but breaks as soon as it is published to a Maven repository (local or remote). Looks related to the compilation.
- Tried making Builder `inner` Kotlin classes, but those are final and not able to instantiate outside of the class
- The decompiled Kotlin generated code looks exactly the same as the one in this PR, so I am not sure what can be going on.

### CODE REVIEW

File: feature/amazon/src/main/java/com/revenuecat/purchases/amazon/AmazonConfiguration.java [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+import androidx.annotation.NonNull;
+
+import com.revenuecat.purchases.PurchasesConfiguration;
+import com.revenuecat.purchases.Store;
+
+/**
+ * Holds parameters to initialize the SDK for the Amazon Store. Create an instance of this class using the [Builder]
+ * and pass it to [Purchases.configure].
+ */
+public final class AmazonConfiguration extends PurchasesConfiguration {
```
tonidero, line 27: Can you add this to the api tests, since we didn't have it before?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -153,6 +173,9 @@ open class PurchasesConfiguration(builder: Builder) {
             this.dangerousSettings = dangerousSettings
         }
 
+        /**
+         * Creates a [PurchasesConfiguration] instance with the specified properties.
+         */
```
tonidero, line 178: Thanks for adding these!

tonidero (approved):
Just the API tests but looks good!

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
```diff
@@ -153,4 +158,18 @@ private class PurchasesAPI {
         val customerInfo: CustomerInfo = logInResult.customerInfo
         LogInResult(customerInfo, created)
     }
+
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    fun checkAmazonConfiguration(context: Context, executorService: ExecutorService) {
```
NachoSoto, line 163: Is this for diagnostics? Maybe move that to a separate method? Otherwise if we accidentally make the whole Amazon configuration experimental we might not notice 🙈

api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
```diff
@@ -105,4 +111,17 @@ static void checkConfiguration(final Context context,
...
+                                         final ExecutorService executorService) {
+        PurchasesConfiguration amazonConfiguration = new AmazonConfiguration.Builder(context, "")
+                .appUserID("")
+                .observerMode(true)
+                .observerMode(false)
+                .service(executorService)
+                .diagnosticsEnabled(true)
+                .entitlementVerificationMode(EntitlementVerificationMode.INFORMATIONAL)
+                .showInAppMessagesAutomatically(true)
+                .build();
```
NachoSoto, line 125: <3

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -153,6 +173,9 @@ open class PurchasesConfiguration(builder: Builder) {
...
         open fun build(): PurchasesConfiguration {
```
NachoSoto, line 179: This could be final?

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
vegaro: good catch

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
vegaro: It might be open for tests? let me check

vegaro: actually, it looks like it can be final, but.... that would be a breaking change 😕

NachoSoto: Oh right. It's fine.

PR #1556, opened 2024-01-12
Fix paywall dialog composable sizing issues

tonidero:
### Description
As I was testing #1551, I noticed that the dialog size wasn't correct. The width of the device sometimes was `MEDIUM` on a phone, causing it to present the dialog in a narrowed version. After more testing, I noticed that it was always one of the dimensions that were not `COMPACT` on my emulator, so I decided to check both. If one of the dimensions is `COMPACT`, we would present the dialog as full screen.

| Before  | After |
| ------------- | ------------- |
| image |  image |


aboedo (approved):

NachoSoto (approved):
Makes sense!

PR #1567, opened 2024-01-25
Add name property to StoreProduct

tonidero:
### Description
Currently `StoreProduct`, `GoogleStoreProduct`, `AmazonStoreProduct` and `TestStoreProduct` expose the `title` of the product. With Google, they have 2 fields for products:
- The [name](https://developer.android.com/reference/com/android/billingclient/api/ProductDetails#getName()) which actually has the product name
- The [title](https://developer.android.com/reference/com/android/billingclient/api/ProductDetails#getTitle()) which has the product name + app name

We were only exposing the `title`. We probably should also expose the `name`. This means that for `AmazonStoreProduct`, the title will be the same as the `name`, but I think that's fine as long as it's documented.

Additionally, this changes our paywalls to use the `name` property instead of the `title`, since it's more appropriate for the paywalls.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -46,7 +46,7 @@ internal class VariableDataProvider(
     }
 
     fun productName(rcPackage: Package): String {
-        return rcPackage.product.title
+        return rcPackage.product.name
```
tonidero, line 49: This is a behavior change, so the product names don't include the app name.

purchases/src/test/java/com/revenuecat/purchases/amazon/helpers/dummies.kt
```diff
@@ -55,8 +55,10 @@ fun stubStoreProductForAmazon(
         get() = type
     override val price: Price
         get() = price
+    override val name: String
+        get() = "An awesome name"
     override val title: String
-        get() = "An awesome title"
+        get() = "An awesome name (App name)"
```
NachoSoto, line 61: 👌🏻

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
NachoSoto: But also a bug fix :)

NachoSoto (approved):
Thanks for doing this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
NachoSoto: Actually what do you think of putting this in a separate PR so we can cool out that we fixed the variable?

tonidero: Yeah, I can do that, will need to update a couple tests but that's ok, will just put them back in that PR 👍

tonidero: Separated in https://github.com/RevenueCat/purchases-android/pull/1569

Issue #1566, opened 2024-01-25
Crash in 7.3.5


zenyagami:
- [ x] I have updated Purchases SDK to the latest version
- [x ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x ] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
There is a crash in some devices, apparently in Tablets
`Context android.app.ContextImpl@61fd4b4 is not a UiContext`
I found that is in this code
`val metrics = WindowMetricsCalculator.getOrCreate().computeCurrentWindowMetrics(context)`  apparently `computeCurrentWindowMetrics` need to be "UI context" and seems that is not always the case
here is a stack trace
```
Fatal Exception: java.lang.IllegalArgumentException: Context android.app.ContextImpl@61fd4b4 is not a UiContext
       at androidx.window.layout.util.ContextUtils.unwrapUiContext$window_release(ContextUtils.kt:52)
       at androidx.window.layout.WindowMetricsCalculatorCompat.computeCurrentWindowMetrics(WindowMetricsCalculatorCompat.kt:67)
       at com.revenuecat.purchases.ui.revenuecatui.helpers.WindowHelperKt.getScreenSize(WindowHelper.kt:69)
       at com.revenuecat.purchases.ui.revenuecatui.helpers.WindowHelperKt.windowSizeClass(WindowHelper.kt:55)
       at com.revenuecat.purchases.ui.revenuecatui.helpers.WindowHelperKt.computeWindowHeightSizeClass(WindowHelper.kt:25)
       at com.revenuecat.purchases.ui.revenuecatui.helpers.WindowHelperKt.shouldUseLandscapeLayout(WindowHelper.kt:44)
       at com.revenuecat.purchases.ui.revenuecatui.helpers.WindowHelperKt.shouldUseLandscapeLayout(WindowHelper.kt:38)
       at com.revenuecat.purchases.ui.revenuecatui.templates.Template5Kt.Template5(Template5.kt:100)
       at com.revenuecat.purchases.ui.revenuecatui.InternalPaywallKt.TemplatePaywall(InternalPaywall.kt:155)
[...]

PR #1570, opened 2024-01-25
Use activity instead of context to calculate window size

tonidero:
### Description
This is an attempt on dealing with #1566. We haven't been able to repro, but we can use the underlying activity from the context given by `LocalContext.current` in order to properly calculate the window size.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/WindowHelper.kt
```diff
@@ -59,14 +58,14 @@ private fun windowSizeClass(): WindowSizeClass {
 @ReadOnlyComposable
 @Composable
 private fun getScreenSize(): Pair<Float, Float> {
-    return if (isInPreviewMode()) {
+    val activity = LocalActivity.current
+    return if (isInPreviewMode() || activity == null) {
```
NachoSoto, line 62: Oooh good idea using this as a fallback too.

NachoSoto (approved):

vegaro (approved):

PR #1576, opened 2024-01-26
`OfferingsFactory`: debug logs when creating `Offerings`

NachoSoto:
While debugging an issue, I realized that the SDK never prints anything after fetching products.
This should provide more visibility on the `Offerings` initialization.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
@@ -52,6 +52,7 @@ internal class OfferingsFactory(
                                     ),
                                 )
                             } else {
+                                log(LogIntent.DEBUG, OfferingStrings.CREATED_OFFERINGS.format(offerings))
                                 onSuccess(offerings)
                             }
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
```diff
@@ -52,6 +52,7 @@ internal class OfferingsFactory(
                                     ),
                                 )
                             } else {
+                                log(LogIntent.DEBUG, OfferingStrings.CREATED_OFFERINGS.format(offerings))
```
tonidero, line 55: Should this be verbose? Seems like this might be a big log in some cases.

tonidero (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
NachoSoto, line 55:
```suggestion
                                log(LogIntent.VERBOSE, OfferingStrings.CREATED_OFFERINGS.format(offerings))
```

PR #1579, opened 2024-01-26
`Paywalls`: improve `PaywallData.configForLocale()` disambiguation

NachoSoto:
iOS equivalent: https://github.com/RevenueCat/purchases-ios/pull/3605

See also https://medium.com/fabiohub/chinese-locale-in-android-part-2-d1cfe87b3fb2:

> In order to support zh-rHK, zh-rHK_#Hant and zh-rHK_#Hans to fall back correctly on both pre & post Android 7.0 devices, you need these resources:
> - `zh-rHK` (Traditional Chinese)
> - `zh-rTW` (Traditional Chinese)
> - `zh` (Simplified Chinese)

simplified
traditional


tonidero (approved):
Annoying we have to add this logic... But looks good!

NachoSoto:
Yeah I forgot to mention that the method in iOS does this automatically, but I couldn't get these tests to pass without encoding that.

PR #1582, opened 2024-01-29
Paywall views: Support updating options state after initial layout

tonidero:
### Description
This will make it so calling the setOfferingId method after the initial layout updates the state causing a recomposition of the paywall.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
@@ -55,7 +55,11 @@ class PaywallView : FrameLayout {
         init(context, null)
     }
 
-    private var updateOfferingId: () -> Unit = {}
+    private val paywallOptionsState = mutableStateOf(
+        PaywallOptions.Builder {
+            dismissHandler?.invoke()
+        }.build(),
+    )
     private var offeringId: String? = null
     private var fontProvider: FontProvider? = null
     private var dismissHandler: (() -> Unit)? = null
@@ -95,34 +99,29 @@ class PaywallView : FrameLayout {
      */
     fun setOfferingId(offeringId: String?) {
         this.offeringId = offeringId
-        this.updateOfferingId()
-        invalidate()
+        val offeringSelection = if (offeringId == null) {
+            OfferingSelection.None
+        } else {
+            OfferingSelection.OfferingId(offeringId)
+        }
+        paywallOptionsState.value = paywallOptionsState.value.copy(offeringSelection = offeringSelection)
     }
 
     private fun init(context: Context, attrs: AttributeSet?) {
         parseAttributes(context, attrs)
+        paywallOptionsState.value = PaywallOptions.Builder { dismissHandler?.invoke() }
+            .setListener(internalListener)
+            .setFontProvider(fontProvider)
+            .setOfferingId(offeringId)
+            .setShouldDisplayDismissButton(shouldDisplayDismissButton ?: false)
+            .build()
         addView(
             ComposeView(context).apply {
                 setContent {
-                    var paywallOptions by remember {
-                        mutableStateOf(PaywallOptions.Builder { dismissHandler?.invoke() }
-                            .setListener(internalListener)
-                            .setFontProvider(fontProvider)
-                            .setOfferingId(offeringId)
-                            .setShouldDisplayDismissButton(shouldDisplayDismissButton ?: false)
-                            .build())
+                    val paywallOptions by remember {
+                        paywallOptionsState
                     }
-                    updateOfferingId = {
-                        val newOfferingId = offeringId
-                        if (newOfferingId != null) {
-                            paywallOptions = paywallOptions.copy(
-                                offeringSelection = OfferingSelection.OfferingId(newOfferingId)
-                            )
-                        }
-                    }
-                    Paywall(
-                        options = paywallOptions,
-                    )
+                    Paywall(paywallOptions)
                 }
             },
         )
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -89,11 +89,26 @@ class PaywallView : FrameLayout {
      */
     fun setOfferingId(offeringId: String?) {
         this.offeringId = offeringId
+        recreateComposeView()
         invalidate()
     }
 
     private fun init(context: Context, attrs: AttributeSet?) {
         parseAttributes(context, attrs)
+        recreateComposeView()
```
tonidero, line 98: This is difficult to review, but I've basically extracted adding the compose view to a function

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -89,6 +95,7 @@ class PaywallView : FrameLayout {
      */
     fun setOfferingId(offeringId: String?) {
         this.offeringId = offeringId
+        this.updateOfferingId()
         invalidate()
```
vegaro, line 99: I don't think we need this anymore, I tested it without the invalidate and it works

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -97,12 +104,22 @@ class PaywallView : FrameLayout {
...
-                        .build()
+                    var paywallOptions by remember {
+                        mutableStateOf(PaywallOptions.Builder { dismissHandler?.invoke() }
+                            .setListener(internalListener)
+                            .setFontProvider(fontProvider)
+                            .setOfferingId(offeringId)
+                            .setShouldDisplayDismissButton(shouldDisplayDismissButton ?: false)
+                            .build())
+                    }
+                    updateOfferingId = {
```
Jjastiny, line 115: nit: Sorry to bother you guys again, but I highly recommend staying away from unstable lambdas, as it may cause unforeseen recompositions and performance hits.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -97,12 +104,22 @@ class PaywallView : FrameLayout {
...
+                        val newOfferingId = offeringId
+                        if (newOfferingId != null) {
+                            paywallOptions = paywallOptions.copy(
+                                offeringSelection = OfferingSelection.OfferingId(newOfferingId)
+                            )
+                        }
+                    }
```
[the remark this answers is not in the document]
Jjastiny: @vegaro 
I think that'll help our team a lot since we won't have to fork off this repository for fontProvider changes :D

vegaro:
Took another approach in https://github.com/RevenueCat/purchases-android/pull/1586

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
tonidero: Hi @Jjastiny, I believe in this case it should be fine since we are only storing a lambda to trigger the update from outside the compose context, but I agree it's probably not the cleanest approach. I just pushed another approach inspired by your approach that also seems to work fine. cc @vegaro

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -88,24 +97,29 @@ class PaywallView : FrameLayout {
...
      */
     fun setOfferingId(offeringId: String?) {
-        this.offeringId = offeringId
-        invalidate()
+        val offeringSelection = if (offeringId == null) {
+            OfferingSelection.None
+        } else {
+            OfferingSelection.OfferingId(offeringId)
+        }
+        paywallOptionsState.value = paywallOptionsState.value.copy(offeringSelection = offeringSelection)
```
tonidero, line 105: There is quite a bit of repetition on these, we probably need to refactor these to avoid it.

vegaro (approved):
Tested it in react native and it works perfectly

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -88,24 +97,29 @@ class PaywallView : FrameLayout {
...
-                        .setListener(internalListener)
-                        .setFontProvider(fontProvider)
-                        .setOfferingId(offeringId)
-                        .setShouldDisplayDismissButton(shouldDisplayDismissButton ?: false)
-                        .build()
-                    Paywall(
-                        options = paywallOptions,
-                    )
+                    val paywallOptions by remember {
+                        paywallOptionsState
```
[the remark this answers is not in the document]
tonidero: Hi @Jjastiny, I believe the changes in this PR would update the state every time the `setOfferingId` method is called, even if it's called multiple times. Can you explain why you think why this would only work once?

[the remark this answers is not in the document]
Jjastiny: @tonidero 
hhm yeah, I stand corrected. It does look like it's recomposing for me. Sorry for the misunderstanding!

PR #1584, opened 2024-01-29
`Paywalls`: replace `TextAlign.Left` with `TextAlign.Start` to better support RTL

NachoSoto:
This is equivalent in LTR languages, but more correct for RTL.


tonidero (approved):
Good call, thanks!

PR #1585, opened 2024-01-29
`Paywalls`: fix template 5 title alignment

NachoSoto:
We didn't notice in our tests because of the large string, but it was vertically centered instead of left aligned like it should be (that's how it's designed in the dashboard and iOS).


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -242,6 +242,8 @@ private fun Title(
         textAlign = TextAlign.Left,
         text = state.selectedLocalization.title,
         color = state.templateConfiguration.getCurrentColors().text1,
+        modifier = Modifier
+            .fillMaxWidth(),
```
tonidero, line 246: I believe it should but just to check, this look good on landscape?

tonidero (approved):
Thanks for fixing this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
NachoSoto: Yeah 👍🏻

PR #1588, opened 2024-02-01
Exposing fontprovider

Jjastiny:
### Checklist

### Motivation
- It's not possible for us to change the font family in react-native-purchase.
- My hope was to expose this font provider methods for the RN Bridge so that we can at least set custom font family.

### Description
- Exposing fontProviders for PaywallFooterView and PaywallView.
- theme composition local provider should update this autmoatically.

Need to add something like this on the React Native side.
If user did link the font via
```xml
<font-family xmlns:app="http://schemas.android.com/apk/res/android">
        <font app:fontStyle="normal" android:fontWeight="200" android:font="@font/montserrat" />
</font-family>
```

```kotlin
    ReactFontManager.getInstance().addCustomFont(this, "Montserrat", R.font.montserrat);
```

We could just bring the font family 
```kotlin
    @ReactProp(name= "fontFamily")
    fun setFontFamily(view: PaywallFooterView, fontFamily: String?){
        val typeface = ReactFontManager.getInstance()
            .getTypeface(fontFamily /** Montserrat in this case. **/, ReactFontManager.TypefaceStyle.NORMAL, assets)
        
        view.setFontProvider(CustomFontProvider(FontFamily(typeface)))
    }
```

Screenshot 2024-01-26 at 10 04 03 PM


Jjastiny:
@tonidero I've changed the PaywallView to use AbstractComposeView and expose fontProvider!

tonidero (approved):
Looks great! Thanks for contributing! I will make sure this gets merged and released on the next version and that you appear as the contributor!

PR #1589, opened 2024-02-01
[EXTERNAL] Exposing fontprovider setter in `PaywallView` and `PaywallFooterView` by @Jjastiny

tonidero:
External contribution from #1588 

### Checklist

### Motivation
- It's not possible for us to change the font family in react-native-purchase.
- My hope was to expose this font provider methods for the RN Bridge so that we can at least set custom font family.

### Description
- Exposing fontProviders for PaywallFooterView and PaywallView.
- theme composition local provider should update this autmoatically.

Need to add something like this on the React Native side.
If user did link the font via
```xml
<font-family xmlns:app="http://schemas.android.com/apk/res/android">
        <font app:fontStyle="normal" android:fontWeight="200" android:font="@font/montserrat" />
</font-family>
```

```kotlin
    ReactFontManager.getInstance().addCustomFont(this, "Montserrat", R.font.montserrat);
```

We could just bring the font family 
```kotlin
    @ReactProp(name= "fontFamily")
    fun setFontFamily(view: PaywallFooterView, fontFamily: String?){
        val typeface = ReactFontManager.getInstance()
            .getTypeface(fontFamily /** Montserrat in this case. **/, ReactFontManager.TypefaceStyle.NORMAL, assets)
        
        view.setFontProvider(CustomFontProvider(FontFamily(typeface)))
    }
```

Screenshot 2024-01-26 at 10 04 03 PM

### DIFF
--- a/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallFooterViewAPI.kt
+++ b/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallFooterViewAPI.kt
@@ -2,7 +2,7 @@ package com.revenuecat.apitester.kotlin.revenuecatui
 
 import android.content.Context
 import android.util.AttributeSet
-import android.widget.FrameLayout
+import androidx.compose.ui.platform.AbstractComposeView
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
@@ -14,7 +14,7 @@ import com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView
 private class PaywallFooterViewAPI {
 
     fun checkType(context: Context) {
-        val paywallFooterView: FrameLayout = PaywallFooterView(context)
+        val paywallFooterView: AbstractComposeView = PaywallFooterView(context)
     }
 
     @Suppress("LongParameterList")
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
@@ -120,7 +120,7 @@ class PaywallView : AbstractComposeView {
 
     private fun init(context: Context, attrs: AttributeSet?) {
         parseAttributes(context, attrs)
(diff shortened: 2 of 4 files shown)


api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallFooterViewAPI.kt
```diff
@@ -14,7 +15,7 @@ import com.revenuecat.purchases.ui.revenuecatui.views.PaywallFooterView
 private class PaywallFooterViewAPI {
 
     fun checkType(context: Context) {
-        val paywallFooterView: FrameLayout = PaywallFooterView(context)
+        val paywallFooterView: AbstractComposeView = PaywallFooterView(context)
```
tonidero, line 18: This is a breaking change... But feels that it should be ok... Any thoughts?

vegaro: If it works I think we should go for it, it will simplify the views

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -113,16 +126,6 @@ class PaywallView : FrameLayout {
             .setOfferingId(initialOfferingId)
             .setShouldDisplayDismissButton(shouldDisplayDismissButton ?: false)
             .build()
-        addView(
```
vegaro, line 116: Can we change line 123 to be `paywallOptions = PaywallOptions.Builder`?

vegaro (approved):

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallFooterViewAPI.kt
NachoSoto: Paywalls is still experimental so :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -133,4 +136,12 @@ class PaywallView : FrameLayout {
         this.initialFontProvider = fontProvider
         this.shouldDisplayDismissButton = shouldDisplayDismissButton
     }
+
+    @Composable
+    override fun Content() {
```
NachoSoto, line 141: This is much nicer

PR #1593, opened 2024-02-05
Fix `DEFERRED` purchases when passing old product id in invalid format

tonidero:
`DEFERRED` is not currently supported in the latest SDK release, but on 6.9.5, the latest version it's supported on, the `DEFERRED` purchase might not work if the developer gives as an old product id in the format `productId:basePlanId`. Reason for it is that for deferred purchases, we used the given old product id as the key of the callback, which doesn't match the one returned once the purchase is completed, so the purchase is successful, but the value is never returned.

This applies a patch so we make sure to use only the `productId` as the `oldProductIdentifier` even if the developer gives us the other format (which can be obtained from several places in our SDK). This will be release as a hotfix and we will make hotfixes of the hybrids as well

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -931,7 +931,7 @@ internal class PurchasesOrchestrator constructor(
                     // This is so, in case a developer gives us an oldProductId in the form "oldProductId:basePlanId",
                     // we only use the oldProductId part as the callback key. Otherwise, it won't be able to find the
                     // callback when the purchase is completed.
-                    oldProductId.split(":")[0]
+                    oldProductId.substringBefore(Constants.SUBS_ID_BASE_PLAN_ID_SEPARATOR)
                 } else {
                     purchasingData.productId
                 }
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -927,8 +927,14 @@ internal class PurchasesOrchestrator constructor(
             if (!state.purchaseCallbacksByProductId.containsKey(purchasingData.productId)) {
                 // When using DEFERRED proration mode, callback needs to be associated with the *old* product we are
                 // switching from, because the transaction we receive on successful purchase is for the old product.
-                val productId =
-                    if (googleProrationMode == GoogleProrationMode.DEFERRED) oldProductId else purchasingData.productId
+                val productId = if (googleProrationMode == GoogleProrationMode.DEFERRED) {
+                    // This is so, in case a developer gives us an oldProductId in the form "oldProductId:basePlanId",
+                    // we only use the oldProductId part as the callback key. Otherwise, it won't be able to find the
+                    // callback when the purchase is completed.
+                    oldProductId.substringBefore(Constants.SUBS_ID_BASE_PLAN_ID_SEPARATOR)
```
tonidero, line 934: Note that we will need to be aware of this once we add back support for deferred purchases in `main` @RevenueCat/catforms

MarkVillacampa (approved):
Good catch!

PR #1595, opened 2024-02-05
Release/6.9.6

tonidero:
### Bugfixes
* Fix `DEFERRED` purchases when passing old product id in invalid format (#1593) via Toni Rico (@tonidero)


tonidero:
Failure deploying happened AFTER the main SDK was deployed. It failed with a 502 while deploying the debugview library... I believe there is no need to deploy that, so will create the github release manually and merge this. cc @RevenueCat/coresdk

PR #1598, opened 2024-02-05
Revert making PaywallFooterView an AbstractComposeView

vegaro:
https://github.com/RevenueCat/purchases-android/pull/1589  made `PaywallFooterView` an `AbstractComposeView`, which broke React Native's implementation because we override `onMeasure` which is final in `AbstractComposeView`


vegaro:
cc: @Jjastiny since you added this it might be of your interest

tonidero (approved):
LGTM!

Jjastiny:
@vegaro Gotcha, I'll take a look. maybe another way to measure on the creation.

PR #1600, opened 2024-02-05
`Paywalls`: add test coverage for locales with different region

NachoSoto:
Test coverage to ensure that Android isn't affected by https://github.com/RevenueCat/purchases-ios/pull/3633.

I've also extracted `getDefaultLocales` and  `localizedConfiguration(List<Locale>)` so we can test those.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/utils/LocaleExtensionsTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/utils/LocaleExtensionsTest.kt
@@ -51,6 +51,14 @@ class LocaleExtensionsTest {
     }
 
     @Test
+    fun `sharedLanguageCodeWith - returns true without region`() {
+        val locale = Locale("en")
+        val otherLocale = Locale("en", "IN")
+        assertThat(locale.sharedLanguageCodeWith(otherLocale)).isTrue
+        assertThat(otherLocale.sharedLanguageCodeWith(locale)).isTrue
+    }
+
+    @Test
     fun `sharedLanguageCodeWith - returns false when locales don't share language code`() {
         val locale = Locale("es", "ES")
         val otherLocale = Locale("en", "ES")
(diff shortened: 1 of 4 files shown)


purchases/src/test/java/com/revenuecat/purchases/utils/LocaleExtensionsTest.kt
```diff
@@ -70,4 +78,11 @@ class LocaleExtensionsTest {
         val otherLocale = Locale("zh-Hans")
         assertThat(locale.sharedLanguageCodeWith(otherLocale)).isFalse
     }
+
+    @Test
+    fun `getDefaultLocales returns the correct list`() {
+        assertThat(
+            getDefaultLocales().map { it.toString() }
+        ).isEqualTo(listOf("en_US"))
```
tonidero, line 86: Hmm if I'm not wrong, this test will only pass on devices where the default locale is `en_US`?

tonidero (approved):
Just a question but looks good!

purchases/src/test/java/com/revenuecat/purchases/utils/LocaleExtensionsTest.kt
NachoSoto: True. Do you have any suggestions that wouldn't lead to false positives (like overriding the locale ourselves).

tonidero: Hmm difficult to do without mocking, which would lose the purpose of the test... Honestly the value of the test is limited since the unit tests don't run in an android image but on the JVM virtual machine (in some cases, there are some differences)

Having said that, I think we can either keep it as is, even if it might fail in some devices, or just return early if the locale is not `Locale.US` so we don't test in those devices. Feels like the first option might be safer but we should add a comment so when it fails, we are aware of this

NachoSoto: Yeah ideally we only run this on CI, where I think it's safe to assume it's that locale. But we don't have a way to detect that, right?

tonidero: Not right now no... We could potentially pass a parameter in CI for that through gradle. That could be an option as well yeah

NachoSoto: Let me add a comment for now then.

PR #1604, opened 2024-02-06
Add `getCurrentOffering(forPlacement: String)` to `Offerings`

joshdholtz:
### Motivation

New function to get `Offering` by a placement identifier

### Description

New `getCurrentOfferingForPlacement(placementId: String): Offering?` method on `Offerings`
- Added demo button for this in paywalls example app
- Added demo button for this in purchase tester app

```kotlin
// Example usage
offerings.getCurrentOfferingForPlacement("onboarding")?.let {
  showPaywall(it)
}
```

#### Logic flow

1. Returns an `Offering` if the placement identifier has a value
2. Checks if the placement identifier key exists in the `placements` dictionary
    - If the key exists (and value is `null`)
          1. Returns `null` if the placement identifier key exists in the dictionary but has a `null` value (this is intentional by the developer to NOT have an `Offering` for this placement identifier)
    - If the key doesn't exist
          1. Return the `Offering` for `fallback_offering_id` (if its not null)

#### New placements object in offerings response

```js
{
    // other offerings stuff
    "placements": {
        "fallback_offering_id": "offering_a", // can be null
        "offering_ids_by_placement": {
            "onboarding_start": null,
            "onboarding_end": "offering_b",
        }
  }
}
```

#### Mapping `JSONObject.NULL` to `null`

Placements is the first thing that I think is allowed to have a null value where JSONObject.toMap() was needed. This function, even though allowed for an optional T type with Map<String, T>, it didn't' actually properly map a value to null. Instead, it kept it as a JSONObject.NULL

Due to Kotlin typecasting, it wasn't really possible check if T was T or T? so JSONObject.NULL couldn't be mapped safely to null.

This feels not as nice because its a whole new method, but its very explicit in what it does... but I'm also happy if we have a better name than this 😛

#### Purchase Tester

https://github.com/RevenueCat/purchases-android/assets/401294/6384d93d-ef1c-44de-b151-299f612b7514

#### Paywall Tester

https://github.com/RevenueCat/purchases-android/assets/401294/3477a014-7303-42c2-aaf8-8211584e5ee7

### DIFF
--- a/api-tester/src/main/java/com/revenuecat/apitester/java/PresentedOfferingContextAPI.java
+++ b/api-tester/src/main/java/com/revenuecat/apitester/java/PresentedOfferingContextAPI.java
@@ -6,9 +6,10 @@ import com.revenuecat.purchases.PresentedOfferingContext;
 final class PresentedOfferingContextAPI {
     static void check(final PresentedOfferingContext presentedOfferingContext) {
         final String offeringIdentifier = presentedOfferingContext.getOfferingIdentifier();
+        final String placementIdentifier = presentedOfferingContext.getPlacementIdentifier();
     }
 
-    static void checkConstructor(final String offeringIdentifier) {
-        final PresentedOfferingContext presentedOfferingContext = new PresentedOfferingContext(offeringIdentifier);
+    static void checkConstructor(final String offeringIdentifier, final String placementIdentifier) {
+        final PresentedOfferingContext presentedOfferingContext = new PresentedOfferingContext(offeringIdentifier, placementIdentifier);
     }
 }
--- a/api-tester/src/main/java/com/revenuecat/apitester/java/PresentedOfferingContextAPI.java
+++ b/api-tester/src/main/java/com/revenuecat/apitester/java/PresentedOfferingContextAPI.java
@@ -6,7 +6,6 @@ import com.revenuecat.purchases.PresentedOfferingContext;
 final class PresentedOfferingContextAPI {
     static void check(final PresentedOfferingContext presentedOfferingContext) {
         final String offeringIdentifier = presentedOfferingContext.getOfferingIdentifier();
-        final String placementIdentifier = presentedOfferingContext.getPlacementIdentifier();
     }
 
     static void checkConstructor(final String offeringIdentifier, final String placementIdentifier) {
--- a/api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
+++ b/api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
@@ -43,7 +43,7 @@ final class StoreProductAPI {
         AmazonStoreProduct underlyingAmazonProduct = AmazonStoreProductKt.getAmazonProduct(product);
 
         final StoreProduct copyWithOfferingId = product.copyWithOfferingId("offeringId");
-        final StoreProduct copyWithContext = product.copyWithPresentedOfferingContext(new PresentedOfferingContext("offeringId"));
+        final StoreProduct copyWithContext = product.copyWithPresentedOfferingContext(new PresentedOfferingContext("offeringId", null));
     }
 
     static void check(final ProductType type) {
--- a/api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
+++ b/api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
@@ -43,7 +43,7 @@ final class StoreProductAPI {
         AmazonStoreProduct underlyingAmazonProduct = AmazonStoreProductKt.getAmazonProduct(product);
 
         final StoreProduct copyWithOfferingId = product.copyWithOfferingId("offeringId");
-        final StoreProduct copyWithContext = product.copyWithPresentedOfferingContext(new PresentedOfferingContext("offeringId", null));
+        final StoreProduct copyWithContext = product.copyWithPresentedOfferingContext(new PresentedOfferingContext("offeringId", "placementId"));
     }
 
     static void check(final ProductType type) {
--- a/api-tester/src/main/java/com/revenuecat/apitester/kotlin/PresentedOfferingContextAPI.kt
+++ b/api-tester/src/main/java/com/revenuecat/apitester/kotlin/PresentedOfferingContextAPI.kt
@@ -8,7 +8,7 @@ private class PresentedOfferingContextAPI {
         val offeringIdentifier: String = presentedOfferingContext.offeringIdentifier
     }
 
-    fun checkConstructor(offeringId: String) {
-        val presentedOfferingContext = PresentedOfferingContext(offeringId)
+    fun checkConstructor(offeringId: String, placementId: String?) {
+        val presentedOfferingContext = PresentedOfferingContext(offeringId, placementId)
     }
 }
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
@@ -1,5 +1,5 @@
 package com.revenuecat.paywallstester
 
 object Constants {
-    const val GOOGLE_API_KEY = "goog_uArAGtKQJeuXMWoSNXCAZBzTepD"
+    const val GOOGLE_API_KEY = "API_KEY"
 }
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
@@ -1,5 +1,5 @@
 package com.revenuecat.paywallstester
 
 object Constants {
-    const val GOOGLE_API_KEY = "API_KEY"
+    const val GOOGLE_API_KEY = "goog_uArAGtKQJeuXMWoSNXCAZBzTepD"
 }
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallTesterApp.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallTesterApp.kt
@@ -21,6 +21,7 @@ fun PaywallTesterApp(
     AppNavHost(navController = navController)
 }
 
+@Suppress("LongMethod")
 @Composable
 private fun AppNavHost(
     navController: NavHostController,
@@ -46,11 +47,23 @@ private fun AppNavHost(
                             .plus("?${PaywallScreenViewModel.FOOTER_CONDENSED_KEY}=true"),
                     )
                 },
+                navigateToPaywallByPlacementScreen = { placementId ->
+                    navController.navigate(
+                        AppScreen.PaywallByPlacement.route
+                            .plus("/$placementId"),
+                    )
+                },
             )
         }
         composable(
             route = AppScreen.Paywall.route.plus("/{${PaywallScreenViewModel.OFFERING_ID_KEY}}"),
             arguments = listOf(navArgument(PaywallScreenViewModel.OFFERING_ID_KEY) { type = NavType.StringType }),
+        ) {
+            PaywallScreen(dismissRequest = navController::popBackStack)
+        }
+        composable(
+            route = AppScreen.PaywallByPlacement.route.plus("/{${PaywallScreenViewModel.PLACEMENT_ID_KEY}}"),
+            arguments = listOf(navArgument(PaywallScreenViewModel.PLACEMENT_ID_KEY) { type = NavType.StringType }),
         ) {
             PaywallScreen(dismissRequest = navController::popBackStack)
         }
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
@@ -213,7 +213,6 @@ private fun OfferingsListScreen(
                             Purchases.sharedInstance.getOfferingsWith {
                                 it.getCurrentOfferingForPlacement(placementId)?.let { offering ->
                                     showDialog.value = false
-//                                    placementOffering.value = offering
                                     tappedOnNavigateToOfferingByPlacement(placementId)
                                 } ?: run {
                                     noPlacementFoundMessage.value = "No offering found for placement '$placementId'"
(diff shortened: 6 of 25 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -24,4 +25,49 @@ data class Offerings(
...
+    val fallbackOfferingId: String?,
+    val offeringIdsByPlacement: Map<String, String?>,
+)
+
+private fun Offering.withPlacement(placement: String): Offering {
+    val placementPackages = this.availablePackages.map {
+        Package(
+            it.identifier,
+            it.packageType,
+            it.product,
```
joshdholtz, line 60: Need to copy/set placement into all of the products

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallTesterApp.kt
```diff
@@ -46,6 +47,12 @@ private fun AppNavHost(
                             .plus("?${PaywallScreenViewModel.FOOTER_CONDENSED_KEY}=true"),
                     )
                 },
+                navigateToPaywallByPlacementScreen = { placementId ->
```
tonidero, line 50: While I'm ok leaving this in paywall tester, I wonder if it would make more sense to add this to purchase tester... I guess ideally we would unify both, but that might be a bigger effort 😅

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
```diff
@@ -1,5 +1,5 @@
 package com.revenuecat.paywallstester
 
 object Constants {
-    const val GOOGLE_API_KEY = "API_KEY"
+    const val GOOGLE_API_KEY = "goog_uArAGtKQJeuXMWoSNXCAZBzTepD"
```
tonidero: Need to remove?

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -129,6 +174,60 @@ private fun OfferingsListScreen(
...
+
+                    // Submit Button
+                    Button(
+                        onClick = {
+                            val placementId = placementIdentifier.value
+
+                            Purchases.sharedInstance.getOfferingsWith {
+                                it.getCurrentOfferingForPlacement(placementId)?.let { offering ->
+                                    showDialog.value = false
+//                                    placementOffering.value = offering
```
tonidero: Need to remove?

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -24,4 +25,59 @@ data class Offerings(
...
+     */
+    fun getCurrentOfferingForPlacement(placementId: String): Offering? {
+        val placements = this.placements ?: run {
+            return null
+        }
+
+        val offeringForPlacement = placements.offeringIdsByPlacement[placementId]?.let { getOffering(it) }
+        val fallbackOffering = placements.fallbackOfferingId?.let { getOffering(it) }
+
+        val showNoOffering = placements.offeringIdsByPlacement.containsKey(placementId)
```
tonidero, line 42: Nitpick but this only needs to be calculated if the offering by placement is null right? Should we do an if-else instead to avoid computing these when not needed?

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -24,4 +25,59 @@ data class Offerings(
...
+    }
+}
+
+/**
+ * This class contains all the offerings by placement configured in RevenueCat dashboard.
+ * For more info see https://www.revenuecat.com/docs/targeting
+ * @property fallbackOfferingId The optional offering identifier to fallback on if the placement isn't found.
+ * @property offeringIdsByPlacement Dictionary of all offering identifiers keyed by their placement identifier.
+ */
+data class Placements(
```
tonidero, line 55: Hmm do we need to expose this class? If so we need to add api tests... but I think we should try to avoid exposing it if possible

purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
```diff
@@ -49,7 +52,24 @@ internal abstract class OfferingParser {
...
+                ?.toMap<String?>()
+                ?.replaceJsonNullWithKotlinNull()
+
+            return@let offeringIdsByPlacement?.let {
+                Placements(
+                    fallbackOfferingId = fallbackOfferingId,
+                    offeringIdsByPlacement = offeringIdsByPlacement,
+                )
+            } ?: run {
+                errorLog(OfferingStrings.PLACEMENTS_EMPTY_ERROR)
```
tonidero, line 67: I'm unsure of this situation... will it only happen when there is a placements key but not an offering_ids_by_placement key? I guess that that should never happen but should be ok to log it. Maybe leave a comment to expect this situation to never happen?

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -24,4 +25,59 @@ data class Offerings(
...
+    val fallbackOfferingId: String?,
+    val offeringIdsByPlacement: Map<String, String?>,
+)
+
+private fun Offering.withPlacement(placementId: String): Offering {
+    val updatedAvailablePackages = this.availablePackages.map {
+        val context = PresentedOfferingContext(
+            offeringIdentifier = it.presentedOfferingContext.offeringIdentifier,
+            placementIdentifier = placementId,
+        )
```
tonidero, line 65: I think you can substitute this using the copy constructor... so something like: `it.presentedOfferingContext.copy(placementIdentifier = placementId)`

tonidero (commented):
Left a few comments but looking good!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallTesterApp.kt
joshdholtz: Ah yeah yeah, I can also add to purchase tester! It's just more fun to show paywalls but totally makes sense there too 😛

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
joshdholtz: Oops nope! This should be internal 😊

purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
joshdholtz: This should never happen because its not allowed on the backend. I didn't want to _not_ catch it but it could be unclear why things aren't working if it does happen 🤷 

I will leave a comment!

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
joshdholtz: Oh, nice! Didn't know dataclasses had that 😅

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
joshdholtz: Yeah yeah, this could debatably go place the`if (showNoOffering)` line below! I pulled this out because I wanted to be explicit with the variable name on what this logic was doing 🤷‍♂️  The method will only get called explicitly be the developer and its a O(1) lookup so it shouldn't be anything too bad for unnecessary computing.

But I'm happy to move it in and add a comment if you'd think that would be better 😊

tonidero: NABD either way, just to avoid a couple extra ops, so I'm ok with this.

api-tester/src/main/java/com/revenuecat/apitester/java/PresentedOfferingContextAPI.java
```diff
@@ -6,9 +6,10 @@
 final class PresentedOfferingContextAPI {
     static void check(final PresentedOfferingContext presentedOfferingContext) {
         final String offeringIdentifier = presentedOfferingContext.getOfferingIdentifier();
+        final String placementIdentifier = presentedOfferingContext.getPlacementIdentifier();
     }
 
-    static void checkConstructor(final String offeringIdentifier) {
-        final PresentedOfferingContext presentedOfferingContext = new PresentedOfferingContext(offeringIdentifier);
+    static void checkConstructor(final String offeringIdentifier, final String placementIdentifier) {
+        final PresentedOfferingContext presentedOfferingContext = new PresentedOfferingContext(offeringIdentifier, placementIdentifier);
```
tonidero, line 13: We should add these to the kotlin API tests as well?

api-tester/src/main/java/com/revenuecat/apitester/java/StoreProductAPI.java
```diff
@@ -43,7 +43,7 @@ static void check(final StoreProduct product) {
         AmazonStoreProduct underlyingAmazonProduct = AmazonStoreProductKt.getAmazonProduct(product);
 
         final StoreProduct copyWithOfferingId = product.copyWithOfferingId("offeringId");
-        final StoreProduct copyWithContext = product.copyWithPresentedOfferingContext(new PresentedOfferingContext("offeringId"));
+        final StoreProduct copyWithContext = product.copyWithPresentedOfferingContext(new PresentedOfferingContext("offeringId", null));
```
tonidero, line 46: Is this needed? I would think not, but just to check.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -129,6 +174,59 @@ private fun OfferingsListScreen(
...
     }
+
+    if (showDialog.value) {
+        Dialog(
+            onDismissRequest = {
+                showDialog.value = false
+                noPlacementFoundMessage.value = null
+                placementIdentifier.value = ""
+            },
+            properties = DialogProperties(usePlatformDefaultWidth = false),
```
tonidero, line 185: Hmm is this needed? This is usually used if you want to make the dialog full screen, and not sure if we want to do that for this one?

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -129,6 +174,59 @@ private fun OfferingsListScreen(
                 .build(),
         )
     }
+
+    if (showDialog.value) {
+        Dialog(
```
tonidero, line 179: Might be a good idea to extract this dialog to a different composable.

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -9,6 +9,7 @@ package com.revenuecat.purchases
 data class Offerings(
     val current: Offering?,
     val all: Map<String, Offering>,
+    val placements: Placements?,
```
tonidero: Hmm interesting... since the public constructor here has `Placements` which is `private`, I would think it shouldn't compile here?

I think the way to do this is to make the constructor for `Offerings` `internal` and then add another constructor without a `placements` parameter to keep retrocompastibility.

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -24,4 +25,56 @@ data class Offerings(
...
+    }
+}
+
+/**
+ * This class contains all the offerings by placement configured in RevenueCat dashboard.
+ * For more info see https://www.revenuecat.com/docs/targeting
+ * @property fallbackOfferingId The optional offering identifier to fallback on if the placement isn't found.
+ * @property offeringIdsByPlacement Dictionary of all offering identifiers keyed by their placement identifier.
+ */
+private data class Placements(
```
tonidero: I think this should be `internal`, not `private`, since it's used in the `OfferingParser`?

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -10,6 +10,11 @@ data class Offerings(
     val current: Offering?,
     val all: Map<String, Offering>,
 ) {
+    internal var placements: Placements? = null
+
+    internal constructor(current: Offering?, all: Map<String, Offering>, placements: Placements?) : this(current, all) {
```
tonidero, line 15: Oh just saw you doing this. I would make this the other way instead. Something like this:
```
data class Offerings internal constructor(
    val current: Offering?,
    val all: Map<String, Offering>,
    internal val placements: Placements? = null,
) {
   constructor(current: Offering?, all: Map<String, Offering>): this(current, all, null))
   ....
}
```
That way it's clearer to see all the parameters in the main constructor which is internal.

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
tonidero: See my comment in: https://github.com/RevenueCat/purchases-android/pull/1604#discussion_r1496127913

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
joshdholtz: This actually is 🤷‍♂️  It resizes the model when placement is not found so things refit 👇 

Screenshot 2024-02-20 at 12 20 09 PM

BTW, I'm _really_ bad at jetpack compose so any suggestions are appreciated 😁

api-tester/src/main/java/com/revenuecat/apitester/java/PresentedOfferingContextAPI.java
tonidero, line 13: We should add one where the `placementIdentifier` is not passed

api-tester/src/main/java/com/revenuecat/apitester/kotlin/PresentedOfferingContextAPI.kt
```diff
@@ -8,7 +8,7 @@ private class PresentedOfferingContextAPI {
         val offeringIdentifier: String = presentedOfferingContext.offeringIdentifier
     }
 
-    fun checkConstructor(offeringId: String) {
-        val presentedOfferingContext = PresentedOfferingContext(offeringId)
+    fun checkConstructor(offeringId: String, placementId: String?) {
+        val presentedOfferingContext = PresentedOfferingContext(offeringId, placementId)
```
tonidero, line 12: Same here, we should add one where `placementId` is not passed

tonidero (approved):
Looks great!

PR #1605, opened 2024-02-07
Fix paywall POST receipt data for paywall revision

joshdholtz:
### Motivation

The API is expecting the `revision` key instead of `paywall_revision` for paywall data in POST receipt. The backend has a fix to support both but changing this for consistency between SDKs.

### Description

- Renamed `paywall_revision` to `revision` in `PaywallPostReceiptData`
- Added `offering_id` to `PaywallPostReceiptData`

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallPostReceiptData.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallPostReceiptData.kt
@@ -19,6 +19,8 @@ internal data class PaywallPostReceiptData(
     val darkMode: Boolean,
     @SerialName("locale")
     val localeIdentifier: String,
+    @SerialName("offering_id")
+    val offeringId: String,
 ) {
     companion object {
         val json = Json.Default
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallPostReceiptData.kt
```diff
@@ -11,8 +11,8 @@ import kotlinx.serialization.json.encodeToJsonElement
 internal data class PaywallPostReceiptData(
     @SerialName("session_id")
     val sessionID: String,
-    @SerialName("paywall_revision")
-    val paywallRevision: Int,
+    @SerialName("revision")
+    val revision: Int,
```
tonidero, line 15: So we only need to change this one right? Not the `PaywallBackendEvent` one?

tonidero (approved):
Just a question, but looks good!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallPostReceiptData.kt
tonidero: Chatted in person, and this is ok. The other does expect `paywall_revision`.

NachoSoto (approved):

NachoSoto:
I suggest changing this from `fix` to `refactor` since it's more of an underlying change that users don't care about IMO.

joshdholtz:
@tonidero @NachoSoto I added another commit here that adds `offering_id` to paywall data as well

NachoSoto (approved):

PR #1606, opened 2024-02-08
Fix serialization proguard rules

vegaro:
- Added rules for version 1.4.1 https://github.com/Kotlin/kotlinx.serialization/blob/v1.4.1/README.md#android which are now included in the latest versions of kotlin serialization but starting from 1.5.0. We are still in 1.4.1 for kotlin compatibility reasons

- Added rules that fix issues when targetting Android SDK 34 and compiling with < 34 (like using version 33). There's been a breaking change in Android 14 (they added ClassValue) and copying what the [behavior changes website mentions](https://developer.android.com/about/versions/14/behavior-changes-14#core-libraries):

> The problem originates with a Kotlin library that changes runtime behaviour based on whether Class.forName("java.lang.ClassValue") returns a class or not. If your app was developed against an older version of the runtime without the java.lang.ClassValue class available, then these optimizations might remove the computeValue method from classes derived from java.lang.ClassValue.

This was causing some functions to be missing from the apk when compiling the app with R8 enabled. I compared both apps (compiling against 33 and 34) and realized  `ClassValueParametrizedCache` and `ClassValueCache` were missing their `computeValue` methods. So I generated methods for those. There's some discussion about this in this issue as well, and the solution they suggest is the same https://github.com/Kotlin/kotlinx.serialization/issues/2119

### DIFF
--- a/purchases/consumer-rules.pro
+++ b/purchases/consumer-rules.pro
@@ -6,3 +6,49 @@
 -dontwarn coil.**
 # Adding temporarily to fix issue after adding kotlin serialization
 -dontwarn java.lang.ClassValue
+
+# START Keep kotlinx.serialization annotations.
+
+# Keep `Companion` object fields of serializable classes.
+# This avoids serializer lookup through `getDeclaredClasses` as done for named companion objects.
+-if @kotlinx.serialization.Serializable class **
+-keepclassmembers class <1> {
+    static <1>$Companion Companion;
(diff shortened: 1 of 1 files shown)


tonidero (approved):
Amazing work finding the issue! 🎉

NachoSoto:
🎉👏🏻🙌🏻

PR #1608, opened 2024-02-09
Add setDisplayDismissButton to PaywallView

vegaro:
WIP while I take a look at safe area issues


tonidero (approved):
Code changes look good... Lmk if you would like a re-review!

PR #1610, opened 2024-02-12
Paywalls: Call `PaywallDialog` dismiss handler after successful restore if needed

tonidero:
### Description
When using `PaywallDialog` composable, if the user restored and matched the `requiredEntitlementIdentifier` or condition to display the paywall given by the developer, we were not dismissing the paywall automatically.

This fixes that, and auto-dismisses the paywall after a successful restore if it matches the given condition by the developer.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -61,13 +61,14 @@ internal interface PaywallViewModel {
 }
 
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class, ExperimentalPreviewRevenueCatPurchasesAPI::class)
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LongParameterList")
 internal class PaywallViewModelImpl(
     override val resourceProvider: ResourceProvider,
     private val purchases: PurchasesType = PurchasesImpl(),
     private var options: PaywallOptions,
     colorScheme: ColorScheme,
     private var isDarkMode: Boolean,
+    private var shouldDisplayBlock: ((CustomerInfo) -> Boolean)?,
     preview: Boolean = false,
 ) : ViewModel(), PaywallViewModel {
     private val variableDataProvider = VariableDataProvider(resourceProvider, preview)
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -61,13 +61,14 @@ internal interface PaywallViewModel {
...
 @OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class, ExperimentalPreviewRevenueCatPurchasesAPI::class)
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LongParameterList")
 internal class PaywallViewModelImpl(
     override val resourceProvider: ResourceProvider,
     private val purchases: PurchasesType = PurchasesImpl(),
     private var options: PaywallOptions,
     colorScheme: ColorScheme,
     private var isDarkMode: Boolean,
+    private var shouldDisplayBlock: ((CustomerInfo) -> Boolean)?,
```
tonidero, line 71: Note that we don't update this after it's set. This means that if the developer changes between different required entitlement ids/condition blocks/having conditions and not, the paywall will keep using the first one passed (note that the view model can survive the paywall closing when presented as a dialog, unless using the navigation component.

I'm not sure how big of a problem that would be though...

NachoSoto: Should this be immutable then?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -170,6 +171,11 @@ internal class PaywallViewModelImpl(
                 val customerInfo = purchases.awaitRestore()
                 Logger.i("Restore purchases successful: $customerInfo")
                 listener?.onRestoreCompleted(customerInfo)
+                shouldDisplayBlock?.let {
+                    if (!it(customerInfo)) {
+                        options.dismissRequest()
+                    }
+                }
```
NachoSoto, line 178: Nit: maybe add some line breaks in this method? There's 5 different completely things being done, so I'd separate them a bit.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -170,6 +171,11 @@ internal class PaywallViewModelImpl(
                 val customerInfo = purchases.awaitRestore()
                 Logger.i("Restore purchases successful: $customerInfo")
                 listener?.onRestoreCompleted(customerInfo)
+                shouldDisplayBlock?.let {
+                    if (!it(customerInfo)) {
+                        options.dismissRequest()
```
NachoSoto, line 176: I would add some logs here like we do when we determine if the paywall should be presented.

NachoSoto (approved):

PR #1612, opened 2024-02-14
Refactor `presentedOfferingIdentifier` into `presentedOfferingContext` object

tonidero:
### Description
With some new incoming features, we will need to pass more context information about where the package/product was obtained from. This PR prepares for that by adding a new class `PresentedOfferingContext` which right now only holds the `presentedOfferingIdentifier` but that we can add more info later on that allows to pass more information to our servers.

#### TODO
- [ ] Fix API tests
- [ ] Fix and add tests

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
@@ -7,14 +7,41 @@ import com.revenuecat.purchases.models.StoreProduct
  * @property identifier Unique identifier for this package. Can be one a predefined package type or a custom one.
  * @property packageType Package type for the product. Will be one of [PackageType].
  * @property product [StoreProduct] assigned to this package.
- * @property offering offering this package was returned from.
+ * @property offering offering Id this package was returned from.
+ * @property presentedOfferingContext [PresentedOfferingContext] from which this package was obtained.
  */
 data class Package(
     val identifier: String,
     val packageType: PackageType,
     val product: StoreProduct,
-    val offering: String,
-)
+    val presentedOfferingContext: PresentedOfferingContext,
+) {
+    @Deprecated(
+        "Use constructor with presentedOfferingContext instead",
+        ReplaceWith(
+            "Package(identifier, packageType, product, offering, " +
+                "PresentedOfferingContext(offeringIdentifier = offering))",
+        ),
+    )
+    constructor(
+        identifier: String,
+        packageType: PackageType,
+        product: StoreProduct,
+        offering: String,
+    ) : this(
+        identifier,
+        packageType,
+        product,
+        PresentedOfferingContext(offeringIdentifier = offering),
+    )
+
+    @Deprecated(
+        "Use presentedOfferingContext.offeringIdentifier instead",
+        ReplaceWith("presentedOfferingContext.offeringIdentifier"),
+    )
+    val offering: String
+        get() = presentedOfferingContext.offeringIdentifier ?: ""
+}
 
 /**
  *  Enumeration of all possible Package types.
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonStoreProduct.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonStoreProduct.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.amazon
 
+import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.ProductType
 import com.revenuecat.purchases.models.Period
 import com.revenuecat.purchases.models.Price
@@ -8,7 +9,7 @@ import com.revenuecat.purchases.models.SubscriptionOption
 import com.revenuecat.purchases.models.SubscriptionOptions
 import org.json.JSONObject
 
-data class AmazonStoreProduct(
+data class AmazonStoreProduct @JvmOverloads constructor(
 
     /**
      * The productId
@@ -76,11 +77,9 @@ data class AmazonStoreProduct(
     val originalProductJSON: JSONObject,
 
     /**
-     * The offering ID this `AmazonStoreProduct` was returned from.
-     *
-     * Null if not using RevenueCat offerings system, or if fetched directly via `Purchases.getProducts`
+     * The context from which this product was obtained.
      */
-    override val presentedOfferingIdentifier: String? = null,
+    override val presentedOfferingContext: PresentedOfferingContext = PresentedOfferingContext(),
 ) : StoreProduct {
 
     @Deprecated(
@@ -120,6 +119,44 @@ data class AmazonStoreProduct(
         presentedOfferingIdentifier = presentedOfferingIdentifier,
     )
 
+    @Deprecated(
+        "Replaced with constructor that takes a presentedOfferingContext",
+        ReplaceWith(
+            "AmazonStoreProduct(productId, type, name, title, description, period, price, " +
+                "subscriptionOptions, defaultOption, iconUrl, freeTrialPeriod, originalProductJSON, " +
+                "PresentedOfferingContext(presentedOfferingIdentifier))",
+        ),
+    )
+    constructor(
+        id: String,
+        type: ProductType,
+        name: String,
+        title: String,
+        description: String,
+        period: Period?,
+        price: Price,
+        subscriptionOptions: SubscriptionOptions?,
+        defaultOption: SubscriptionOption?,
+        iconUrl: String,
+        freeTrialPeriod: Period?,
+        originalProductJSON: JSONObject,
+        presentedOfferingIdentifier: String? = null,
+    ) : this(
+        id = id,
+        type = type,
+        name = name,
+        title = title,
+        description = description,
+        period = period,
+        price = price,
+        subscriptionOptions = subscriptionOptions,
+        defaultOption = defaultOption,
+        iconUrl = iconUrl,
+        freeTrialPeriod = freeTrialPeriod,
+        originalProductJSON = originalProductJSON,
+        presentedOfferingContext = PresentedOfferingContext(presentedOfferingIdentifier),
+    )
+
     /**
      * Contains only data that is required to make the purchase.
      */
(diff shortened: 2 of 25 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
```diff
@@ -7,14 +7,41 @@ import com.revenuecat.purchases.models.StoreProduct
...
+        product,
+        PresentedOfferingContext(offeringIdentifier = offering),
+    )
+
+    @Deprecated(
+        "Use presentedOfferingContext.offeringIdentifier instead",
+        ReplaceWith("presentedOfferingContext.offeringIdentifier"),
+    )
+    val offering: String
+        get() = presentedOfferingContext.offeringIdentifier ?: ""
```
tonidero, line 43: I made the `presentedOfferingContext` have a nullable `offeringIdentifier` instead of making the `presentedOfferingContext` nullable in StoreProduct/SubscriptionOption in case we could have non-offering related context, which is probably not the case right now... But this forces us to default to something here as a fallback. It should never be null right now though in the `Package`.

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/handler/PurchaseHandler.kt
```diff
@@ -29,7 +30,6 @@ internal class PurchaseHandler(
 ) : PurchaseResponseListener {
 
     private val productTypes = mutableMapOf<String, ProductType>()
-    private val presentedOfferingsByProductIdentifier = mutableMapOf<String, String?>()
```
tonidero, line 32: I removed this because it seems it wasn't used. Seems we were passing the offering id in the `AmazonBilling` class instead.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
```diff
@@ -70,9 +71,9 @@ data class StoreTransaction(
     val originalJson: JSONObject,
 
     /**
-     * Offering that was presented when making the purchase. Always null for restored purchases.
+     * Context of the offering that was presented when making the purchase.
      */
-    val presentedOfferingIdentifier: String?,
+    val presentedOfferingContext: PresentedOfferingContext?,
```
tonidero, line 76: I had doubts on this... Note that it's nullable here but not nullable in the `Package`/`StoreProduct`/`SubscriptionOption`. It will be null here when it can't find any information about the context (usually restores and such) vs having a context with empty data when the origin of the purchase is through a product/we don't have offering data.

Lmk if you think we should standardize this somehow and we can chat, but I thought it was ok like this for now.

purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
joshdholtz: I like this! I think this makes sense 👍

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
joshdholtz: I think this is fine to be nullable here! Forcing an object before the transaction makes it a cleaner (especially when copying the object down into subscription options. I probably would have done the same here in making it nullable in the transaction 👍

joshdholtz (commented):
This is looking good! Thank you so much for jumping on this so quickly 👍

purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
vegaro: hmm I think it makes sense for `StoreProduct/SubscriptionOption`, but with this there's no way to indicate that a package will always have an offering and a package is always going to have an offering. Do we need to replace `offering` in `Package` with `presentedOfferingContext`? I think we can leave this as is `offering: String` and only replace `presentedOfferingIdentifier` in `StoreProduct` and `SubscriptionOption`? We access the `presentedOfferingContext` through the products, and not the packages.

tonidero: Hmm it's a good point... the only difference is going to be that the offering is not nullable in the Package... However, It feels nicer to me if we have the same format in all 3 places... If we don't have this data at the `Package` level we will need to access it through the StoreProduct when purchasing, which seems slightly less clean... 

What do you think about leaving the offering property undeprecated here in the Package? It would be duplicated with the one in `presentedOfferingContext` but that way we can keep the non-nullability here

vegaro: I guess my question is if we need a `presentedOfferingContext` at all in `Package`. Is it ever going to be different than the `offering` value? I guess we can add a different way of getting Packages in the future, which would indeed make offering not nullable and create the need of a `presentedOfferingContext`, but maybe we can add that property whenever we need. I just think adding it right now makes the API confusing and doesn't add value. Maybe I am missing context

vegaro: Actually... What if we make it a sealed class that only has an implementation right now? That way we can keep the original nullabilities. Can you remind me what other type of data we are adding to the `PresentedOfferingContext`?

tonidero: @joshdholtz I was chatting live with @vegaro about this one... In the end, we believe it would be better to make the `offeringIdentifier` not nullable inside the `PresentedOfferingContext` so we can make the `PresentedOfferingContext` not nullable in `Package` but nullable in `StoreProduct`/`SubscriptionOption`. Do you think we will (now or in the future) have data like this without having an offering id? 

My original reason to make it nullable was to provide more flexibility in the future in case we have some of this data without an offering id, but I don't think that would happen for now at least, so it makes sense to better reflect the current model.

Let us know what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/models/GoogleStoreProduct.kt
```diff
@@ -1,10 +1,21 @@
...
 
 import com.android.billingclient.api.ProductDetails
+import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.ProductType
 import java.util.Locale
 
-data class GoogleStoreProduct(
+data class GoogleStoreProduct
+@JvmOverloads
+@Deprecated(
```
tonidero, line 10: Quick note about this deprecation. I was running into a lot of issues trying to add the new parameter while trying to keep retrocompatibility... In the end, I'm deprecating the main constructor and adding a new one which is internal only. 

Reason was that there were a lot of overload conflicts, specially in Java, where String can be null and it was making the API tests fail. 

IMO, we should make all these constructors internal in the future. If we want to allow devs to create them, then I would suggest using the `Builder` pattern instead... Much easier to add/deprecate things and allows us to keep flexibility for the constructor.

purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/1612/commits/9eaf8a42fb199168f586f2f9c1e80b53f46860ca

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/handler/PurchaseHandler.kt
```diff
@@ -38,26 +38,25 @@ internal class PurchaseHandler(
         activity: Activity,
         appUserID: String,
         storeProduct: StoreProduct,
-        presentedOfferingIdentifier: String?,
+        presentedOfferingContext: PresentedOfferingContext?,
```
joshdholtz, line 41: Is this not being used in this method? 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -779,10 +780,7 @@ internal class BillingWrapper(
 
             getPurchaseType(purchase.purchaseToken) { type ->
                 completion(
-                    purchase.toStoreTransaction(
-                        type,
-                        context?.presentedOfferingId,
-                    ),
+                    purchase.toStoreTransaction(type),
```
joshdholtz, line 783: I might have missed this but this no need any presented offering information anymore? 🤔

joshdholtz (approved):
This looks good to me! Two minor questions but otherwise LGTM 💪 Thanks so much for doing this!

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendTest.kt
```diff
@@ -50,6 +51,7 @@ import io.mockk.unmockkObject
 import io.mockk.verify
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Fail.fail
+import org.checkerframework.checker.optional.qual.Present
```
vegaro, line 54: Is this intended?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: This wasn't needed since it would only enter here if `context` was null. From a few lines above (line 774), you can see that if context.productType is not null, it will return, and `productType` is not null inside the context, so it would only continue here if context was null (it was a warning from Android Studio)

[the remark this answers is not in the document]
tonidero: See response in https://github.com/RevenueCat/purchases-android/pull/1612/files/9eaf8a42fb199168f586f2f9c1e80b53f46860ca#r1491165230

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/handler/PurchaseHandler.kt
tonidero: Nope, it was actually unused. I removed part of it but forgot to remove the parameter here. Thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/Package.kt
```diff
@@ -7,14 +7,41 @@ import com.revenuecat.purchases.models.StoreProduct
...
     val packageType: PackageType,
     val product: StoreProduct,
-    val offering: String,
-)
+    val presentedOfferingContext: PresentedOfferingContext,
+) {
+    @Deprecated(
+        "Use constructor with presentedOfferingContext instead",
+        ReplaceWith(
+            "Package(identifier, packageType, product, offering, " +
```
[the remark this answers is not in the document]
tonidero: Good catch!

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonStoreProduct.kt
```diff
@@ -117,7 +118,45 @@ data class AmazonStoreProduct(
...
+        title: String,
+        description: String,
+        period: Period?,
+        price: Price,
+        subscriptionOptions: SubscriptionOptions?,
+        defaultOption: SubscriptionOption?,
+        iconUrl: String,
+        freeTrialPeriod: Period?,
+        originalProductJSON: JSONObject,
+        presentedOfferingIdentifier: String,
```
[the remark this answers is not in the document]
vegaro: Makes sense!

purchases/src/test/java/com/revenuecat/purchases/amazon/ReceiptToStoreTransactionTest.kt
```diff
@@ -337,7 +342,7 @@ class ReceiptToStoreTransactionTest {
         val receipt = dummyReceipt()
         val storeTransaction = receipt.toStoreTransaction(
             productId = "sku",
-            presentedOfferingIdentifier = null,
+            presentedOfferingContext = presentedOfferingContext,
```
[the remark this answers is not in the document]
tonidero: Ohh yeah, I missed that... I think it doesn't really matter in this test, but better to keep the behavior in this PR. Thanks!

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendTest.kt
tonidero: Nope, unused import... We should run detekt in the tests as well, but that will probably be a lot of changes 😅

vegaro (approved):

PR #1618, opened 2024-02-16
Fix load shedder integration tests

vegaro:
Fix compilation in load shedder integration tests


tonidero (approved):
Thanks for fixing that!

Issue #1621, opened 2024-02-20
[Amazon] Android 14 devices crash with receiver not exported


snapwoodapps:
My app targets Android 14 (targetSdkVersion 34).  [Android 14 requires runtime registered receivers to specify whether they are exported or not](https://developer.android.com/about/versions/14/behavior-changes-14#runtime-receivers-exported).

I'm seeing the following crash in the field when Amazon users install the app on their Android 14 phones and attempt to purchase.  

```
Fatal Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{com.myapp.sample/com.revenuecat.purchases.amazon.purchasing.ProxyAmazonBillingActivity}: java.lang.SecurityException: com.myapp.sample: One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3822)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3963)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:103)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:139)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:96)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2468)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:205)
       at android.os.Looper.loop(Looper.java:294)
       at android.app.ActivityThread.main(ActivityThread.java:8248)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)

Caused by java.lang.SecurityException: com.myapp.sample: One of RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED should be specified when a receiver isn't being registered exclusively for system broadcasts
       at android.os.Parcel.createExceptionOrNull(Parcel.java:3057)
       at android.os.Parcel.createException(Parcel.java:3041)
       at android.os.Parcel.readException(Parcel.java:3024)
       at android.os.Parcel.readException(Parcel.java:2966)
[...]

PR #1622, opened 2024-02-21
Fix runtime crash when using amazon and targetting android 14

tonidero:
### Description
As reported in #1621, the app crashed when using amazon and targetting android 14 in an android 14 device and starting a purchase. This fixes the issue.


gradle/libs.versions.toml
```diff
@@ -19,7 +19,7 @@ coroutines = "1.6.4"
 androidxNavigation = "2.5.3"
 appcompat = "1.4.1"
 dokka = "1.8.10"
-androidxCore = "1.8.0"
+androidxCore = "1.12.0"
```
tonidero, line 22: Didn't see anything concerning in these and allows us to use the new `ContextCompat.registerReceiver`

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/purchasing/ProxyAmazonBillingDelegate.kt
```diff
@@ -19,9 +21,10 @@ internal class ProxyAmazonBillingDelegate {
     internal var broadcastReceiver: ProxyAmazonBillingActivityBroadcastReceiver? = null
     private val filter = ProxyAmazonBillingActivityBroadcastReceiver.newPurchaseFinishedIntentFilter()
 
+    @SuppressLint("WrongConstant")
```
tonidero, line 24: This was very weird... The linter was complaining that we were passing the wrong type when passing `ContextCompat.RECEIVER_EXPORTED` even though it's the expected type...

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -490,7 +489,7 @@ internal class BillingWrapper(
         )
     }
 
-    @VisibleForTesting(otherwise = PRIVATE)
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
tonidero, line 492: This broke compilation after the update for some reason... In any case, this is how we do it everywhere else.

vegaro (approved):

PR #1623, opened 2024-02-21
Change default paywall background with a lower quality image

tonidero:
### Description
Equivalent to iOS's https://github.com/RevenueCat/purchases-ios/pull/3379

Deals with: https://github.com/RevenueCat/purchases-flutter/issues/982

This image is used in the default paywall when there is a problem with the configured paywall in the dashboard or there is no paywall configured. This image is blurred, so the quality is not really important here, and it was a pretty big image.


joshdholtz (approved):
:shipit:

NachoSoto:
♥️

PR #1626, opened 2024-02-22
New `syncAttributesAndOfferingsIfNeeded` method

joshdholtz:
### Motivation

Adding new method to manually sync attributes and fetch fresh offerings 

### Description

- New `Purchases.shared.syncAttributesAndOfferingsIfNeeded()` method
    - Syncs subscriber attributes and then fetches fresh offerings from network
    - This is designed to be a blocking call since developer will want fresh offerings after syncing subscriber attributes
    - The method is rate limited to 5 calls per minute (new `RateLimiter` class()

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
@@ -100,7 +100,7 @@ suspend fun Purchases.awaitSyncPurchases(): CustomerInfo {
  * be called when using Targeting Rules with Custom Attributes. Any subscriber attributes should be set before
  * calling this method to ensure the returned offerings are applied with the latest subscriber attributes.
  *
- * * This method is rate limited to once per minute.
+ * This method is rate limited to 5 calls per minute. It will log a warning and return offerings cache when reached.
  *
  * Refer to [the guide](https://www.revenuecat.com/docs/tools/targeting) for more targeting information
  * For more offerings information, see [getOfferings]
@@ -113,7 +113,7 @@ suspend fun Purchases.awaitSyncPurchases(): CustomerInfo {
  */
 @JvmSynthetic
 @Throws(PurchasesException::class)
-suspend fun Purchases.awaitSyncAttributesAndOfferingsIfNeeded(): Offerings? {
+suspend fun Purchases.awaitSyncAttributesAndOfferingsIfNeeded(): Offerings {
     return suspendCoroutine { continuation ->
         syncAttributesAndOfferingsIfNeededWith(
             onSuccess = continuation::resume,
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
@@ -94,3 +94,30 @@ suspend fun Purchases.awaitSyncPurchases(): CustomerInfo {
         )
     }
 }
+
+/**
+ * Syncs subscriber attributes and then fetches the configured offerings for this user. This method is intended to
+ * be called when using Targeting Rules with Custom Attributes. Any subscriber attributes should be set before
+ * calling this method to ensure the returned offerings are applied with the latest subscriber attributes.
+ *
+ * * This method is rate limited to once per minute.
+ *
+ * Refer to [the guide](https://www.revenuecat.com/docs/tools/targeting) for more targeting information
+ * For more offerings information, see [getOfferings]
+ *
+ * Coroutine friendly version of [Purchases.syncAttributesAndOfferingsIfNeeded].
+ *
+ * @throws [PurchasesException] with the first [PurchasesError] if there's an error syncing attributes
+ * or fetching offerings.
+ * @returns The [Offerings] fetched after syncing attributes.
+ */
+@JvmSynthetic
+@Throws(PurchasesException::class)
+suspend fun Purchases.syncAttributesAndOfferingsIfNeeded(): Offerings? {
+    return suspendCoroutine { continuation ->
+        syncAttributesAndOfferingsIfNeededWith(
+            onSuccess = continuation::resume,
+            onError = { continuation.resumeWithException(PurchasesException(it)) },
+        )
+    }
+}
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
@@ -113,7 +113,7 @@ suspend fun Purchases.awaitSyncPurchases(): CustomerInfo {
  */
 @JvmSynthetic
 @Throws(PurchasesException::class)
-suspend fun Purchases.syncAttributesAndOfferingsIfNeeded(): Offerings? {
+suspend fun Purchases.awaitSyncAttributesAndOfferingsIfNeeded(): Offerings? {
     return suspendCoroutine { continuation ->
         syncAttributesAndOfferingsIfNeededWith(
             onSuccess = continuation::resume,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -42,6 +42,7 @@ import com.revenuecat.purchases.interfaces.PurchaseCallback
 import com.revenuecat.purchases.interfaces.PurchaseErrorCallback
 import com.revenuecat.purchases.interfaces.ReceiveCustomerInfoCallback
 import com.revenuecat.purchases.interfaces.ReceiveOfferingsCallback
+import com.revenuecat.purchases.interfaces.SyncAttributesAndOfferingsCallback
 import com.revenuecat.purchases.interfaces.SyncPurchasesCallback
 import com.revenuecat.purchases.interfaces.UpdatedCustomerInfoListener
 import com.revenuecat.purchases.models.BillingFeature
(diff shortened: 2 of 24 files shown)


purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -94,3 +94,30 @@ suspend fun Purchases.awaitSyncPurchases(): CustomerInfo {
...
+ *
+ * Coroutine friendly version of [Purchases.syncAttributesAndOfferingsIfNeeded].
+ *
+ * @throws [PurchasesException] with the first [PurchasesError] if there's an error syncing attributes
+ * or fetching offerings.
+ * @returns The [Offerings] fetched after syncing attributes.
+ */
+@JvmSynthetic
+@Throws(PurchasesException::class)
+suspend fun Purchases.awaitSyncAttributesAndOfferingsIfNeeded(): Offerings? {
```
tonidero, line 116: Hmm should the return type be optional?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1154,6 +1186,8 @@ internal class PurchasesOrchestrator constructor(
 
         const val frameworkVersion = Config.frameworkVersion
 
+        const val lastSyncAttributesAndOfferingsTimeout = 60000
```
tonidero, line 1189: Nitpick, but we could use the `Duration` types for these comparisons, so this could for example be `60.seconds`, and then in the other places we could also convert to `Duration`, to be more clear on the type.

purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
```diff
@@ -67,4 +67,8 @@ enum class PurchasesErrorCode(val code: Int, val description: String) {
         36,
         "Request failed signature verification. Please see https://rev.cat/trusted-entitlements for more info.",
     ),
+    SyncingAttributesRateLimitReached(
+        37,
```
tonidero, line 71: Is this error number correct? Note that it needs to match iOS so it's the same in the hybrids.

tonidero (commented):
Have a meeting now, but will continue reviewing later, looking good though!

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
joshdholtz: Ohhhh, good point! This should not 😅

purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
MarkVillacampa: I think error number 37 already exists in iOS and is equivalent to number 36 on iOS so we already have a mismatch.

joshdholtz: It looks like everything is off-by-one between iOS and Android error codes? 🤔

MarkVillacampa: I think we just skip error numbers in one platform if it doesn't apply e.g. there is no error 26 `productDiscountMissingIdentifierError` in Android, but there is in iOS.

vegaro: Ugh, it's broken yes. `SignatureVerificationError` should be 37 in Android. I would say for simplicity  make `SyncingAttributesRateLimitReached` 40 in both platforms and we fix `SignatureVerificationError` and `featureNotAvailableInCustomEntitlementsComputationMode` later by making them 38 and 39 🤷‍♂️ 

Btw, I would love to bring this up in the next SDK guild meeting because it's one of the most bug prone thing in the SDKs we have, and I am not sure how to solve it

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/SyncAttributesAndOfferingsCallback.kt
```diff
@@ -0,0 +1,9 @@
+package com.revenuecat.purchases.interfaces
+
+import com.revenuecat.purchases.Offerings
+import com.revenuecat.purchases.PurchasesError
+
+interface SyncAttributesAndOfferingsCallback {
+    fun onSuccess(offerings: Offerings)
+    fun onError(error: PurchasesError)
+}
```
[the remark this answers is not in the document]
joshdholtz: Oooooops! Forgot about that one 😅 Thank you!

purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
joshdholtz: @vegaro Make it a shared CI step that compares and fails if different? 🤷‍♂️

vegaro: Yes, we are going to have to do something on CI, a test. I think that's the only option. Maybe in PHC? But that would too late, after releasing the natives. 

Let's talk about it yes haha

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
tonidero: Hmm this is still returning a nullable value?

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
```diff
@@ -26,9 +26,10 @@ internal class OfferingsManager(
         appInBackground: Boolean,
         onError: ((PurchasesError) -> Unit)? = null,
         onSuccess: ((Offerings) -> Unit)? = null,
+        fetchCurrent: Boolean = false,
     ) {
         val cachedOfferings = offeringsCache.cachedOfferings
-        if (cachedOfferings == null) {
+        if (cachedOfferings == null || fetchCurrent) {
             log(LogIntent.DEBUG, OfferingStrings.NO_CACHED_OFFERINGS_FETCHING_NETWORK)
```
tonidero, line 33: I would update the logs for each of these scenarios so it's not confusing

purchases/src/main/kotlin/com/revenuecat/purchases/utils/RateLimiter.kt
```diff
@@ -0,0 +1,28 @@
+package com.revenuecat.purchases.utils
+
+internal class RateLimiter(val maxCalls: Int, val periodSeconds: Int) {
+    companion object {
+        const val MILLISECONDS_PER_SECOND = 1000
+    }
+
+    private val maxCallInclusive: Int
+        get() = maxCalls + 1
```
tonidero, line 9: Hmm I think this can be
```suggestion
    private val maxCallInclusive = maxCalls + 1
```

purchases/src/main/kotlin/com/revenuecat/purchases/utils/RateLimiter.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    private val callTimestamps: LongArray = LongArray(maxCallInclusive)
+    private var index = 0
+
+    @Synchronized
+    fun shouldProceed(): Boolean {
```
tonidero, line 15: ❤️

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -214,6 +217,37 @@ internal class PurchasesOrchestrator constructor(
...
+        if (!lastSyncAttributesAndOfferingsRateLimiter.shouldProceed()) {
+            log(
+                LogIntent.WARNING,
+                SyncAttributesAndOfferingsStrings.RATE_LIMIT_REACHED.format(
+                    lastSyncAttributesAndOfferingsRateLimiter.maxCalls,
+                    lastSyncAttributesAndOfferingsRateLimiter.periodSeconds,
+                ),
+            )
+
+            getOfferings(receiveOfferingsCallback)
```
tonidero, line 242: Just to confirm, this will return the offerings from cache, not from network right?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -214,6 +217,37 @@ internal class PurchasesOrchestrator constructor(
...
+            return
+        }
+
+        subscriberAttributesManager.synchronizeSubscriberAttributesForAllUsers(appUserID) {
+            getOfferings(receiveOfferingsCallback, fetchCurrent = true)
```
tonidero, line 247: I'm wondering if we should fallback to the cached offerings if the `getOfferings` call fails... Considering we fallback to the cached offerings before, I think it would be better... and maybe add a log.

tonidero (commented):
Looking great! I think we are close :)

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
joshdholtz: Correct correct!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
joshdholtz: I _think_ this already happens in `fetchAndCacheOfferings()` 🤔 If there is a server error, `offeringsCache.cachedOfferingsResponse` is used. Its not the "in memory" cache but it should be the same thing. Thoughts?

tonidero: Oh right 🤦 . Please ignore me :P

PR #1628, opened 2024-02-26
Add targeting to `PresentedOfferingContext`

joshdholtz:
### Motivation

Get more context with targeting information 

### Description

- Parse targeting information into `Offerings`
- Add targeting context into `Offerings.current`
- Add targeting context into `getCurrentOffering(forPlacement: String)`
- Post targeting context in post receipts

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
@@ -46,43 +46,46 @@ data class Offerings internal constructor(
 
         return offering?.withPresentedContext(placementId = placementId, targeting = this.targeting)
     }
+
+    /**
+     * This class contains information about the targeting that presented the offering.
+     * @property revision The revision of the targeting that presented the offering.
+     * @property ruleId The rule id of the targeting that presented the offering.
+     */
+    internal data class Targeting(
+        val revision: Int,
+        val ruleId: String,
+    )
+
+    /**
+     * This class contains all the offerings by placement configured in RevenueCat dashboard.
+     * For more info see https://www.revenuecat.com/docs/targeting
+     * @property fallbackOfferingId The optional offering identifier to fallback on if the placement isn't found.
+     * @property offeringIdsByPlacement Dictionary of all offering identifiers keyed by their placement identifier.
+     */
+    internal data class Placements(
+        val fallbackOfferingId: String?,
+        val offeringIdsByPlacement: Map<String, String?>,
+    )
 }
 
-/**
- * This class contains information about the targeting that presented the offering.
- * @property revision The revision of the targeting that presented the offering.
- * @property ruleId The rule id of the targeting that presented the offering.
- */
-internal data class Targeting(
-    val revision: Int,
-    val ruleId: String,
-)
-
-/**
- * This class contains all the offerings by placement configured in RevenueCat dashboard.
- * For more info see https://www.revenuecat.com/docs/targeting
- * @property fallbackOfferingId The optional offering identifier to fallback on if the placement isn't found.
- * @property offeringIdsByPlacement Dictionary of all offering identifiers keyed by their placement identifier.
- */
-internal data class Placements(
-    val fallbackOfferingId: String?,
-    val offeringIdsByPlacement: Map<String, String?>,
-)
-
-internal fun Offering.withPresentedContext(placementId: String?, targeting: Targeting?): Offering {
+internal fun Offering.withPresentedContext(placementId: String?, targeting: Offerings.Targeting?): Offering {
     val updatedAvailablePackages = this.availablePackages.map {
-        val context = it.presentedOfferingContext.copy(
-            placementIdentifier = placementId,
-            targetingRevision = targeting?.revision,
-            targetingRuleId = targeting?.ruleId,
+        val oldContext = it.presentedOfferingContext
+
+        val newContext = oldContext.copy(
+            placementIdentifier = placementId ?: oldContext.placementIdentifier,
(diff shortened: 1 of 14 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
```diff
@@ -70,7 +79,12 @@ internal abstract class OfferingParser {
             }
         }
 
-        return Offerings(offerings[currentOfferingID], offerings, placements)
+        return Offerings(
+            current = offerings[currentOfferingID]?.withPresentedContext(null, targeting),
```
tonidero, line 83: Is it possible to use both placements and targeting at the same time? Just making sure since this will override the placementId to null, if it has one.

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -58,9 +69,13 @@ internal data class Placements(
     val offeringIdsByPlacement: Map<String, String?>,
 )
 
-private fun Offering.withPlacement(placementId: String): Offering {
+internal fun Offering.withPresentedContext(placementId: String?, targeting: Targeting?): Offering {
```
tonidero, line 72: You could potentially pass the current values as default values for these parameters I think? That way we would avoid overwriting the wrong value accidentally.

purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
```diff
@@ -52,6 +54,13 @@ internal abstract class OfferingParser {
             }
         }
 
+        val targeting: Targeting? = offeringsJson.optJSONObject("targeting")?.let {
+            val revision = it.getInt("revision")
+            val ruleId = it.getString("rule_id")
```
tonidero, line 59: We could write some tests in the `OfferingsFactoryTest` to actually perform the parsing of these I think

purchases/src/main/kotlin/com/revenuecat/purchases/PresentedOfferingContext.kt
```diff
@@ -17,4 +17,14 @@ data class PresentedOfferingContext(
...
+
+    /**
+     * The revision of the targeting used to obtain this object.
+     */
+    val targetingRevision: Int? = null,
+
+    /**
+     * The rule id from the targeting used to obtain this object.
+     */
+    val targetingRuleId: String? = null,
```
tonidero, line 29: I was also wondering whether we should have one property of the `Targeting` type, instead of these 2 separately... We could make it internal in order to not expose the type publicly, Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -43,10 +44,20 @@ data class Offerings internal constructor(
...
+        return offering?.withPresentedContext(placementId = placementId, targeting = this.targeting)
     }
 }
 
+/**
+ * This class contains information about the targeting that presented the offering.
+ * @property revision The revision of the targeting that presented the offering.
+ * @property ruleId The rule id of the targeting that presented the offering.
+ */
+internal data class Targeting(
```
tonidero, line 56: I'm slightly concerned this name might be too generic... Maybe we could rename `OfferingTargeting`? Or make it an inner class inside `Offering`, so it would be `Offering.Targeting`?

tonidero (commented):
Looking good! Just left some comments

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
joshdholtz: I'll rename this to `TargetedOffering` which will match the naming of the backend 👍

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -43,31 +44,48 @@ data class Offerings internal constructor(
...
+}
+
+internal fun Offering.withPresentedContext(placementId: String?, targeting: Offerings.Targeting?): Offering {
     val updatedAvailablePackages = this.availablePackages.map {
-        val context = it.presentedOfferingContext.copy(placementIdentifier = placementId)
-        val product = it.product.copyWithPresentedOfferingContext(context)
+        val oldContext = it.presentedOfferingContext
+
+        val newContext = oldContext.copy(
+            placementIdentifier = placementId ?: oldContext.placementIdentifier,
```
tonidero, line 77: This means this method can't set the value to null, even if given as a parameter... My other suggestion was to pass the original value of the offering as default values of the parameters of this method. Then, if null is passed, we know it's explicit, and we set it here. NABD though, we probably don't really need to set it to null if it has a value I think?

tonidero (approved):
I'm just wondering if it's worth trying to make the new types internal... But it should also be ok as it is.

purchases/src/main/kotlin/com/revenuecat/purchases/PresentedOfferingContext.kt
```diff
@@ -17,4 +17,23 @@ data class PresentedOfferingContext(
...
-) : Parcelable
+
+    /**
+     * The targeting context used to obtain this object.
+     */
+    val targetingContext: TargetingContext? = null,
+) : Parcelable {
+
+    @Parcelize
+    data class TargetingContext(
```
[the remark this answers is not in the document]
joshdholtz: I actually tried doing that (because it does seem better for us and the end user) but it ended up getting really messy 😬 I'll try doing it again though!

PR #1634, opened 2024-03-06
Fix amazon zero price

vegaro:
We were never posting prices for Amazon purchases, because when we try to find the purchased product, we where using the `subscriptionOptions` property, which is always null for Amazon

This fix is a bit hacky right now, but we need to ship it quickly

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
@@ -35,7 +35,8 @@ internal class PostTransactionWithProductDetailsHelper(
                         val purchasedStoreProduct =
                             // Amazon purchases don't have subscription options
                             // Amazon is the only store that has marketplace
-                            // We want Google restores to enter in this if, that's why we need to filter Amazon only
+                            // We want Google restores to enter in this if condition,
+                            // that's why we need to filter Amazon only
                             if (transaction.type == ProductType.SUBS && transaction.marketplace == null) {
                                 storeProducts.firstOrNull { product ->
                                     product.subscriptionOptions?.let { subscriptionOptions ->
@@ -47,7 +48,7 @@ internal class PostTransactionWithProductDetailsHelper(
                                     product.id == transaction.productIds.firstOrNull()
                                 }
                             }
-                        debugLog("Purchased store product: $purchasedStoreProduct")
+                        debugLog("Store product found for transaction: $purchasedStoreProduct")
                         postReceiptHelper.postTransactionAndConsumeIfNeeded(
                             purchase = transaction,
                             storeProduct = purchasedStoreProduct,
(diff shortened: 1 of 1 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
```diff
@@ -31,18 +32,22 @@ internal class PostTransactionWithProductDetailsHelper(
...
                     onReceive = { storeProducts ->
-                        val purchasedStoreProduct = if (transaction.type == ProductType.SUBS) {
-                            storeProducts.firstOrNull { product ->
-                                product.subscriptionOptions?.let { subscriptionOptions ->
-                                    subscriptionOptions.any { it.id == transaction.subscriptionOptionId }
-                                } ?: false
+                        val purchasedStoreProduct =
+                            // Amazon purchases don't have subscription options
+                            // Amazon is the only store that has marketplace
+                            // We want Google restores to enter in this if, that's why we need to filter Amazon only
```
joshdholtz, line 38: Should this "enter in this if subs"?

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
```diff
@@ -31,18 +32,22 @@ internal class PostTransactionWithProductDetailsHelper(
...
+                                    product.id == transaction.productIds.firstOrNull()
+                                }
                             }
-                        } else {
-                            storeProducts.firstOrNull { product ->
-                                product.id == transaction.productIds.firstOrNull()
-                            }
-                        }
-
+                        debugLog("Purchased store product: $purchasedStoreProduct")
```
joshdholtz, line 50: This can happen in restore too, right? If so, should this log be changed?

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
```diff
@@ -31,18 +32,22 @@ internal class PostTransactionWithProductDetailsHelper(
...
-                        val purchasedStoreProduct = if (transaction.type == ProductType.SUBS) {
-                            storeProducts.firstOrNull { product ->
-                                product.subscriptionOptions?.let { subscriptionOptions ->
-                                    subscriptionOptions.any { it.id == transaction.subscriptionOptionId }
-                                } ?: false
+                        val purchasedStoreProduct =
+                            // Amazon purchases don't have subscription options
+                            // Amazon is the only store that has marketplace
+                            // We want Google restores to enter in this if, that's why we need to filter Amazon only
+                            if (transaction.type == ProductType.SUBS && transaction.marketplace == null) {
```
vegaro, line 39: I thought about just removing the else and checking if `purchasedStoreProduct` is null and in that case doing:

```
storeProducts.firstOrNull { product ->
    product.id == transaction.productIds.firstOrNull()
}
```

But I worry about restored purchases, which also don't have subscription options, and right now we are also getting a `purchasedStoreProduct` for them. That might be another bug, or expected behavior. 

So I went with checking `transaction.marketplace == null` which is a proxy for checking if the store is Google. So if it's Amazon, or consumables, it will go into the else.

joshdholtz (approved):
Two small comments but otherwise looks good 👍

aboedo (approved):

farmerbb:
Thank you very much for getting this fixed!

PR #1638, opened 2024-03-07
Release 6.9.7

vegaro:
### Bugfixes
* Fix amazon zero price (#1634) via Cesar de la Vega (@vegaro)


.circleci/config.yml
```diff
@@ -225,7 +225,7 @@ jobs:
           command: sudo pip install awscli
       - run:
           name: Deploy to S3
-          command: aws s3 sync ~/project/docs/6.9.6 s3://purchases-docs/android/6.9.6 --delete
+          command: aws s3 sync ~/project/docs/6.9.7 s3://purchases-docs/android/6.9.7 --delete
       - run:
           name: Update index.html
           command: aws s3 cp ~/project/docs/index.html s3://purchases-docs/android/index.html
```
tonidero, line 231: Hmm I just noticed but would this make the 6.9.7 docs the default ones? If so, we would need to either fix it so we don't overwrite the file if the versions is older than the one already there, or overwrite this file after publishing this

tonidero (approved):
Just a concern... but maybe we can fix the file after releasing this?

joshdholtz (approved):

.circleci/config.yml
tonidero: Yup! We can remove that for now 👍

PR #1648, opened 2024-03-15
Create android test apps to test Android SDK regressions

tonidero:
### Description
We recently updated a library that actually made us incompatible with apps targeting android 33. We shouldn't inadvertently change that, so we are adding these simple apps which only need to be compiled to verify we are still compatible with the target versions for `purchases-android` and `purchases-ui`.


settings.gradle
```diff
@@ -14,3 +14,5 @@ include ':api-tester'
 include 'ui:debugview'
 include ':ui:revenuecatui'
 include ':examples:paywall-tester'
+include ':test-apps:testpurchasesandroidcompatibility'
+include ':test-apps:testpurchasesuiandroidcompatibility'
```
tonidero, line 18: Note that I've included them in the main project instead of depending on the deployed build, so it's easier to use.

test-apps/testpurchasesandroidcompatibility/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+    alias(libs.plugins.kotlin.android)
+}
+
+android {
+    namespace = "com.revenuecat.testpurchasesandroidcompatibility"
+    compileSdk = 33 // Keeping it at 33 to test compatibility with purchases-android
+
+    defaultConfig {
+        applicationId = "com.revenuecat.testpurchasesandroidcompatibility"
+        minSdk = 19
```
tonidero, line 12: This will also test the minSdk version compatibility

test-apps/testpurchasesandroidcompatibility/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+        targetSdk = 33 // Keeping it at 33 to test compatibility with purchases-android
```
tonidero, line 13: If I lower this to 32, this fails to compile, as expected

.circleci/config.yml
```diff
@@ -171,6 +171,12 @@ jobs:
       - android/accept-licenses
       - android-dependencies
       - android/restore-build-cache
+      - run:
+          name: Verify purchases-android SDK compatibility
```
aboedo, line 175: wonder if we should also just put "SDK level 33" here so it's very obvious exactly what went wrong if this fails and you don't have context

aboedo (approved):

.circleci/config.yml
tonidero: Hmm right... I'm concerned that we might forget about updating this in the future, when we need to update the target sdk... But maybe that's ok we can work it so it makes sense 👍

PR #1652, opened 2024-03-20
Add space to title in Template 5

vegaro:
If there's no header image in template 5, the close button overlaps the title

Screenshot 2024-03-20 at 13 12 06

Adds some space if there's no header image:
Screenshot 2024-03-20 at 14 48 12

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
@@ -20,7 +20,7 @@ internal object UIConstant {
     // But lower than 1.0f
     const val halfWeight = 0.5f
 
-    val closeButtonSize = 24.dp
+    val iconButtonSize = 48.dp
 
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
```diff
@@ -20,6 +20,8 @@ internal object UIConstant {
     // But lower than 1.0f
     const val halfWeight = 0.5f
 
+    val closeButtonSize = 24.dp
```
tonidero, line 23: FFTI but should we use this to set the size of the close button as well? I see this is set in the drawable, but maybe it's worth setting it on code to make sure it matches?

tonidero (approved):
Just a comment, but looks good. Thanks for fixing this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/UIConstant.kt
vegaro: good idea!

PR #1658, opened 2024-03-25
Create Amazon diagnostic Counters

vegaro:
Based off #1651 

Creates Counters for posting Amazon diagnostics. This PR doesn't send the counters, that will go in a separate PR.


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -16,6 +16,7 @@ import kotlin.time.Duration
  * This class is the entry point for all diagnostics tracking. It contains all information for all events
  * sent and their properties. Use this class if you want to send a a diagnostics entry.
  */
+@Suppress("TooManyFunctions")
```
vegaro, line 19: We should extract Google and Amazon events at some point. I think this is fine for now.

tonidero (approved):

PR #1665, opened 2024-04-02
Support sending only diagnostics events and update data to raw data

tonidero:
### Description
We're planning to only send diagnostics events and have the backend be in charge of analyzing the events and redirect them to Grafana/other consumers as needed. This removes the `Counter`/`Histogram` concepts from the SDK and updates the existing events to send the raw data, instead of some discretized version of that data.


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsAnonymizer.kt
```diff
@@ -6,22 +6,8 @@ internal class DiagnosticsAnonymizer(
     private val anonymizer: Anonymizer,
 ) {
     fun anonymizeEntryIfNeeded(diagnosticsEntry: DiagnosticsEntry): DiagnosticsEntry {
```
tonidero, line 8: Note that I currently left the name as `DiagnosticsEntry` instead of renaming to `DiagnosticsEvent`. I think it's ok to keep it as is, but lmk if you prefer a different name.

tonidero:
Depends on backend changes, but we can review this already.

aboedo (approved):
image

gotta love that ratio

PR #1666, opened 2024-04-03
`Amazon`: Add getAmazonLWAConsentStatus method to support Quick Subscribe

MarkVillacampa:
Exposes the `lwaConsentStatus` user data field from the Amazon SDK. Necessary to support the [QuickSubscribe](https://developer.amazon.com/docs/in-app-purchasing/iap-quicksubscribe.html) feature.


feature/amazon/build.gradle
```diff
@@ -27,5 +27,5 @@ android {
 dependencies {
     implementation project(":purchases")
 
-    implementation libs.amazon.appstore.sdk
+    // implementation libs.amazon.appstore.sdk
```
MarkVillacampa, line 30: Version 3.0.5 of the amazon appstore sdk if not yet available on Maven Central, so we need to import the .aar manually

https://central.sonatype.com/artifact/com.amazon.device/amazon-appstore-sdk

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -763,6 +763,18 @@ class Purchases internal constructor(
...
+    /**
+     * Note: This method only works for the Amazon Appstore. There is no Google equivalent at this time.
+     * Calling from a Google-configured app will always return AmazonLWAConsentStatus.UNAVAILABLE.
+     *
+     * Get the Login with Amazon consent status for the current user. Used to implement Quick Subscribe.
+     *
+     * @param [onSuccess] Called when the consent status was successfully fetched.
+     * @param [onError] Called when there was an error fetching the consent status.
+     */
+    fun getAmazonLWAConsentStatus(onSuccess: (AmazonLWAConsentStatus) -> Unit, onError: PurchasesErrorCallback) {
```
tonidero, line 775: Need to add this and the enum to the API tests.

tonidero, line 775: We could also add a coroutine version in `coroutinesExtensions.kt`.

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt
```diff
@@ -18,7 +19,7 @@ internal class DefaultPurchasingServiceProvider : PurchasingServiceProvider {
     }
 
     override fun getUserData(): RequestId {
-        return PurchasingService.getUserData()
+        return PurchasingService.getUserData(UserDataRequest.newBuilder().setFetchLWAConsentStatus(true).build())
```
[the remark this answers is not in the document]
tonidero: I guess not but just to check, this shouldn't cause any behavior changes for users not using Quick Subscribe right? Just not sure why this is needed, if there is no behavior change 🤔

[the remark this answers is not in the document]
MarkVillacampa: I *think* it's possible the `lwaConsentStatus` field is not filled/requested unless this flag is set to true. This might be for backwards compatibility with older versions of the SDK.

https://developer.amazon.com/docs/in-app-purchasing/iap-quicksubscribe.html#implement-getuserdata-changes-and-integrate-login-with-amazon

jitpack.yml
```diff
@@ -0,0 +1,2 @@
+jdk:
+  - openjdk17
```
MarkVillacampa, line 2: jitpack builds with java 8, but we require a newer version

settings.gradle
```diff
@@ -7,12 +7,5 @@ pluginManagement {
...
-include ':integration-tests'
 include ':purchases'
-include 'examples:purchase-tester'
-include ':api-tester'
-include 'ui:debugview'
-include ':ui:revenuecatui'
-include ':examples:paywall-tester'
-include ':test-apps:testpurchasesandroidcompatibility'
-include ':test-apps:testpurchasesuiandroidcompatibility'
+include ':ui:revenuecatui'
```
MarkVillacampa, line 11: I had to remove all test modules because they cannot be built by jitpack

purchases/libs/appstore-sdk-3.0.5.jar
tonidero, line 1: NABD if we are not going to merge this, but this should ideally go in the `feature/amazon` module.

examples/purchase-tester/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="com.google.android.gms.permission.AD_ID"/>
+    <uses-feature android:name="amazon.lwa.quicksignup.supported"/>
```
MarkVillacampa, line 5: This is needed to test Quick Subscribe in LaT

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
```diff
@@ -90,6 +92,12 @@ class OverviewFragment : Fragment(), OfferingCardAdapter.OfferingCardAdapterList
         viewModel.retrieveCustomerInfo()
 
         Purchases.sharedInstance.getOfferingsWith(::showError, ::populateOfferings)
+
+        Purchases.sharedInstance.getAmazonLWAConsentStatus({
+            Log.i("PurchaseTester", "AmazonLWAConsentStatus Success: $it")
+        }, {
+            Log.e("PurchaseTester", "AmazonLWAConsentStatus Error: $it")
+        })
```
MarkVillacampa, line 100: Figured we dont want a UI for this since there is no need to need to check more than once per app launch

vegaro: Not sure if there's an easy way to only do this for Amazon configurations? I think in `OfferingFragment.onCreateView` we check if it's play store, so maybe we could do something similar over here

vegaro: Unless we want to see the `UNAVAILABLE` log when configured for the play store

purchases/src/main/kotlin/com/revenuecat/purchases/AmazonLWAConsentStatus.kt
```diff
@@ -0,0 +1,16 @@
...
+enum class AmazonLWAConsentStatus {
+    /**
+     * User has provided consent to access data.
+     */
+    CONSENTED,
+
+    /**
+     * Customer hasn't provided consent or the consent has expired.
+     */
+    UNAVAILABLE,
```
vegaro, line 15: The docs in this enum use both `Customer` and `User`, we should be consistent

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -741,6 +742,17 @@ class Purchases internal constructor(
         purchasesOrchestrator.getProductsOfTypes(productIds.toSet(), setOf(ProductType.INAPP), callback)
     }
 
+    /**
+     * Note: This method only works for the Amazon Appstore. There is no Google equivalent at this time.
+     * Calling from a Google-configured app will always return AmazonLWAConsentStatus.UNAVAILABLE.
+     *
+     * Get the Login with Amazon consent status for the current user. Used to implement Quick Subscribe.
```
vegaro, line 749: Maybe we should link here to docs?

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt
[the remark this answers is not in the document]
vegaro: Feels like there should be a test making sure getUserData is called with this builder

purchases/src/main/kotlin/com/revenuecat/purchases/AmazonLWAConsentStatus.kt
```diff
@@ -0,0 +1,16 @@
+package com.revenuecat.purchases
+
+/**
+ * Specifies the user consent status for Login with Amazon
+ */
+enum class AmazonLWAConsentStatus {
```
tonidero, line 6: Need to add this enum to API tests

purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
```diff
@@ -205,6 +219,23 @@ fun Purchases.syncAttributesAndOfferingsIfNeededWith(
...
 }
 
+/**
+ * Note: This method only works for the Amazon Appstore. There is no Google equivalent at this time.
+ * Calling from a Google-configured app will always return AmazonLWAConsentStatus.UNAVAILABLE.
+ *
+ * Get the Login with Amazon consent status for the current user. Used to implement Quick Subscribe.
+ *
+ * @param [onSuccess] Called when the consent status was successfully fetched.
+ * @param [onError] Called when there was an error fetching the consent status.
```
tonidero, line 229: Nitpick, but we should order these in the order of the params, so we should switch these
```suggestion
 * @param [onError] Called when there was an error fetching the consent status.
 * @param [onSuccess] Called when the consent status was successfully fetched.
```

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
```diff
@@ -361,6 +363,42 @@ internal class AmazonBilling(
         }
     }
 
+    override fun getAmazonLWAConsentStatus(
```
tonidero, line 366: Looks like we need to add tests for this one

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
```diff
@@ -129,6 +141,7 @@ private class PurchasesAPI {
         val customerInfo4: CustomerInfo = purchases.awaitRestore()
         val customerInfo5: CustomerInfo = purchases.awaitSyncPurchases()
         var offerings: Offerings = purchases.awaitSyncAttributesAndOfferingsIfNeeded()
+        var consentStatus = purchases.getAmazonLWAConsentStatus()
```
tonidero, line 144: We should add the type here
```suggestion
        var consentStatus: AmazonLWAConsentStatus = purchases.getAmazonLWAConsentStatus()
```

tonidero (commented):
Looks pretty good! I think once we have tests for the new amazon method we should be good

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt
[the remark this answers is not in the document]
MarkVillacampa: Hmm how could you test that? that's the direct call to Amazon's SDK.

Internally, the implementation of `getUserData` without parameters just calls the builder without setting the `setFetchLWAConsentStatus` flag.

Screenshot 2024-06-18 at 12 31 22

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
MarkVillacampa: Changed. It's not the cleanest way but we're now checking we're using the SDK in Amazon mode before checking consent status.

purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
MarkVillacampa: Eneded up changing the order of the parameters instead of the documentation.

tonidero: Hmm I think our usual convention is to put the `onError` first and `onSuccess` after, so I would actually rollback that change and change the documentation, so it's consistent.

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
```diff
@@ -115,6 +123,10 @@ private class PurchasesAPI {
             onError = { _: PurchasesError -> },
             onSuccess = { _: Offerings -> },
         )
+        purchases.getAmazonLWAConsentStatus(
```
tonidero, line 126: Need to add the `getAmazonLWAConsentStatusWith` variant here as well I think.

purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
```diff
@@ -205,6 +219,27 @@ fun Purchases.syncAttributesAndOfferingsIfNeededWith(
...
+ * account creation using Quick Subscribe.
+ *
+ * For more information, check the documentation:
+ * https://developer.amazon.com/docs/in-app-purchasing/iap-quicksubscribe.html
+ *
+ * @param [onSuccess] Called when the consent status was successfully fetched.
+ * @param [onError] Called when there was an error fetching the consent status.
+ */
+@Suppress("unused")
+fun Purchases.getAmazonLWAConsentStatus(
```
tonidero, line 236: Usually for these, we rename the method with the `With` suffix like:
```suggestion
fun Purchases.getAmazonLWAConsentStatusWith(
```

tonidero (approved):
Just a couple extra comments about the order of params and the method name but looks good to me!

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt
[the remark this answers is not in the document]
vegaro: I mean that we should check we are passing a builder that has  `setFetchLWAConsentStatus` set to true to `PurchasingService.getUserData` 

That way we find out if someone removes it and changes this function back to

```
 override fun getUserData(): RequestId {
        return PurchasingService.getUserData()
 }
 ```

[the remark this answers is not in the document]
Comment (tonidero) [reply purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt:22]: You can potentially mock static/builder methods, similar to how we do for Google here: https://github.com/RevenueCat/purchases-android/blob/5deff6cbebe4712ef3633053ceb6d217179f4fa6/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt#L325

File: purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt (hunk 11 above)
Comment (vegaro) [reply purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt:233]: The order of the parameters was my reasoning, and it the reason for having the `onSuccess` at the end is so the `onSuccess` is the one that gets the lambda conversion, which I think it's cleaner.

```
fun Purchases.getAmazonLWAConsentStatus(::onError) {
	// do something
}
```

Better than:

```
fun Purchases.getAmazonLWAConsentStatus(onSuccess = { 
 // do something on success
} ) {
	// do something on error
}
```

**Review by vegaro — APPROVED** [review]
Other than Toni's comments and the test to make sure we are passing the correct parameters to `PurchasingService`, this looks good.

File: purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt (hunk 3 above)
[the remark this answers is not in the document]
Comment (vegaro) [reply purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt:22]: Thanks for sending that Toni. I thought my original comment was confusing and could mean "test the insights of Amazon to make sure the method is called with the right parameters". 

But yes I'm meaning make sure the builder passed has `setFetchLWAConsentStatus` set. I thought you could mock `PurchasingService` and capture the parameter passed to `getUserData`, but just checked and looks like `UserDataRequest` doesn't expose `fetchLWAConsentStatus`, which makes it a bit more complicated...

Feel free to ignore if it's too complicated, but I think you should be able to test it by doing what Toni mentions and mocking the static

```
 mockkStatic(UserDataRequest::class)
 val mockBuilder = mockk<UserDataRequest.Builder>(relaxed = true)
  every {
      UserDataRequest.newBuilder()
  } returns mockBuilder
```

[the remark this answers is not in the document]
Comment (MarkVillacampa) [reply purchases/src/main/kotlin/com/revenuecat/purchases/amazon/DefaultPurchasingServiceProvider.kt:22]: I get it now! Added a test specifically for testing `DefaultPurchasingServiceProvider`. Testing it directly seemed like the easiest way and didn't feel like creating a new file for it was needed but let me know if you'd improve the test in some way!

File: purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt (hunk 11 above)
Comment (MarkVillacampa) [reply purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt]: Good catch I didn't notice the naming or the convention in the other methods, fixed!

File: purchases/src/test/java/com/revenuecat/purchases/amazon/AmazonBillingTest.kt [hunk 16]
```diff
@@ -1117,6 +1116,31 @@ class AmazonBillingTest {
         assertThat(consent).isEqualTo(AmazonLWAConsentStatus.CONSENTED)
     }
 
+    @Test
+    fun `calling getUserData() in DefaultPurchasingServiceProvider sets LWA flag in builder`() {
+        setup()
+
+        mockkStatic(UserDataRequest::class)
+        mockkStatic(PurchasingService::class)
```
tonidero, line 1124: Need to add a `tearDown` method to unmock these like:
```
@After
fun tearDown() {
    clearAllMocks()
}
```
Otherwise, the mockk will live through different tests.

tonidero (approved):

purchases/src/test/java/com/revenuecat/purchases/amazon/AmazonBillingTest.kt
MarkVillacampa: added 👍

PR #1670, opened 2024-04-08
Remove confusing placements error log

tonidero:
### Description
This removes an error log which was happening even in normal situations. The response from the backend may have a `placements` key with only a fallback, but no `offering_ids_by_placement`, which was logging it as an error, which it wasn't


joshdholtz (approved):
Ship it 🚢

PR #1671, opened 2024-04-10
Add offline entitlements diagnostics events

tonidero:
### Description
This adds some new events to get some data into offline entitlements when diagnostics is enabled:
- `entered_offline_entitlements_mode`
- `error_entering_offline_entitlements_mode`

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
@@ -390,7 +390,7 @@ class DiagnosticsTrackerTest {
     fun `trackErrorEnteringOfflineEntitlementsMode tracks correct data for one time purchase error`() {
         val expectedProperties = mapOf(
             "offline_entitlement_error_reason" to "one_time_purchase_found",
-            "error_message" to "There was a problem with the operation. Looks like we don't support that yet. Check the underlying error for more details. Underlying error: Offline entitlements are not supported for one time purchases. Found one time purchases.",
+            "error_message" to "There was a problem with the operation. Looks like we don't support that yet. Check the underlying error for more details. Underlying error: Offline entitlements are not supported for one time purchases. Found one time purchases. See for more info: https://rev.cat/offline-entitlements",
         )
         every { diagnosticsFileHelper.appendEvent(any()) } just Runs
         diagnosticsTracker.trackErrorEnteringOfflineEntitlementsMode(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/strings/OfflineEntitlementsStrings.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/strings/OfflineEntitlementsStrings.kt
@@ -12,7 +12,7 @@ internal object OfflineEntitlementsStrings {
     const val ALREADY_CALCULATING_OFFLINE_CUSTOMER_INFO = "Already calculating offline customer info for %s."
     const val ERROR_PARSING_PRODUCT_ENTITLEMENT_MAPPING = "Error parsing cached product entitlement mapping: %s"
     const val OFFLINE_ENTITLEMENTS_UNSUPPORTED_INAPP_PURCHASES = "Offline entitlements are not supported for " +
-        "one time purchases. Found one time purchases."
+        "one time purchases. Found one time purchases. See for more info: https://rev.cat/offline-entitlements"
     const val PRODUCT_ENTITLEMENT_MAPPING_REQUIRED = "Product entitlement mapping is required for offline " +
         "entitlements. Skipping offline customer info calculation."
 }


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -199,6 +202,36 @@ internal class DiagnosticsTracker(
...
+        val reason = if (isOneTimePurchaseFoundError) {
+            "one_time_purchase_found"
+        } else {
+            "unknown"
+        }
+        val event = DiagnosticsEntry(
+            name = DiagnosticsEntryName.ERROR_ENTERING_OFFLINE_ENTITLEMENTS_MODE,
+            properties = mapOf(
+                "offline_entitlement_error_reason" to reason,
+                "error_message" to "${error.message}. Underlying error: ${error.underlyingErrorMessage}",
```
tonidero, line 227: This won't actually be tracked in Grafana, but I thought it would still be useful to send it.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -199,6 +202,36 @@ internal class DiagnosticsTracker(
...
+        val event = DiagnosticsEntry(
+            name = DiagnosticsEntryName.ENTERED_OFFLINE_ENTITLEMENTS_MODE,
+            properties = mapOf(),
+        )
+        trackEvent(event)
+    }
+
+    fun trackErrorEnteringOfflineEntitlementsMode(error: PurchasesError) {
+        val isOneTimePurchaseFoundError = error.code == PurchasesErrorCode.UnsupportedError &&
+            error.underlyingErrorMessage == OfflineEntitlementsStrings.OFFLINE_ENTITLEMENTS_UNSUPPORTED_INAPP_PURCHASES
```
tonidero, line 217: I thought about making this a parameter instead of relying on processing the error... But this cleans it up a bit IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
aboedo: at the very least it'll be useful in elastic if we're looking through logs of a single customer

purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
```diff
@@ -346,4 +349,63 @@ class DiagnosticsTrackerTest {
...
+                    event.properties == expectedProperties
+            })
+        }
+    }
+
+    @Test
+    fun `trackErrorEnteringOfflineEntitlementsMode tracks correct data for unknown error`() {
+        val expectedProperties = mapOf(
+            "offline_entitlement_error_reason" to "unknown",
+            "error_message" to "Unknown error.. Underlying error: test error message",
```
aboedo, line 372:
```suggestion
            "error_message" to "Unknown error. Underlying error: test error message",
```

purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
```diff
@@ -346,4 +349,63 @@ class DiagnosticsTrackerTest {
...
+                    event.properties == expectedProperties
+            })
+        }
+    }
+
+    @Test
+    fun `trackErrorEnteringOfflineEntitlementsMode tracks correct data for one time purchase error`() {
+        val expectedProperties = mapOf(
+            "offline_entitlement_error_reason" to "one_time_purchase_found",
+            "error_message" to "There was a problem with the operation. Looks like we doesn't support that yet. Check the underlying error for more details.. Underlying error: Offline entitlements are not supported for active inapp purchases. Found active inapp purchases.",
```
aboedo, line 393:
```suggestion
            "error_message" to "There was a problem with the operation. Looks like we don't support that yet. Check the underlying error for more details. Underlying error: Offline entitlements are not supported for active inapp purchases. Found active inapp purchases.",
```

aboedo: I'd also skip the "looks like we don't support that yet" bit entirely, I'm not sure we'll ever be able to support one-times in the first place. 
Also... I guess active inapp purchases is technically correct, but like, that sounds like any purchases, right? I'd maybe change the copy to `one time purchases` for clarity

aboedo: it'd be great to add a link to documentation too, we usually use the Rebrandly links for that kind of thing. It helps in case we ever want to document this very specifically, if we have a Rebrandly in place we can update where it points to later on

aboedo (approved):
looking good, minor comments

purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
tonidero: > I'd also skip the "looks like we don't support that yet" bit entirely

That part is actually more generic, and it's the error message we display every time we get the `UnsupportedError` code... I could change it, but I would prefer to leave that part as is, since it might apply to something else... But happy to revamp this in our future better error handling exploration!

PR #1674, opened 2024-04-11
Paywalls: Allow closed button color to be configured

joshdholtz:
### Motivation

Allow dialog paywall view to have a custom close button color from the dashboard

### Description

- New `close_button` color


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/CloseButton.kt
```diff
@@ -25,7 +27,11 @@ fun BoxScope.CloseButton(
                 onClick = onClick,
                 enabled = actionInProgress.not(),
             ) {
-                Icon(painter = painterResource(id = R.drawable.close), contentDescription = null)
+                Icon(
+                    painter = painterResource(id = R.drawable.close),
+                    contentDescription = null,
+                    tint = color ?: Color.Black,
```
tonidero, line 33: Can we default this to `LocalContentColor.current`? This is what was before, so we don't change the color for users inadvertently.

tonidero (approved):
Looks great! Just a comment.

PR #1697, opened 2024-05-08
Support non consumable products

tonidero:
### Description
This brings support to non-consumables in android. The backend will let us know whether we need to consume a product after posting it or not.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
@@ -7,6 +7,7 @@ import com.revenuecat.purchases.common.PostReceiptDataErrorCallback
 import com.revenuecat.purchases.common.PostReceiptErrorHandlingBehavior
 import com.revenuecat.purchases.common.ReceiptInfo
 import com.revenuecat.purchases.common.caching.DeviceCache
+import com.revenuecat.purchases.common.networking.PostReceiptResponse
 import com.revenuecat.purchases.common.offlineentitlements.OfflineEntitlementsManager
 import com.revenuecat.purchases.models.StoreProduct
 import com.revenuecat.purchases.models.StoreTransaction
@@ -52,9 +53,9 @@ internal class PostReceiptHelper(
             storeUserID,
             marketplace,
             initiationSource,
-            onSuccess = {
+            onSuccess = { postReceiptResponse ->
                 deviceCache.addSuccessfullyPostedToken(purchaseToken)
-                onSuccess(it)
+                onSuccess(postReceiptResponse.customerInfo)
             },
             onError = { backendError, errorHandlingBehavior, _ ->
                 if (errorHandlingBehavior == PostReceiptErrorHandlingBehavior.SHOULD_BE_CONSUMED) {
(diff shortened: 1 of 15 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -102,13 +103,23 @@ internal class PostReceiptHelper(
...
             marketplace = purchase.marketplace,
             initiationSource = initiationSource,
-            onSuccess = { info ->
-                billing.consumeAndSave(finishTransactions, purchase, initiationSource)
-                onSuccess?.let { it(purchase, info) }
+            onSuccess = { postReceiptResponse ->
+                // Currently we only support a single token per postReceipt call but multiple product Ids.
+                // The backend would fail if given more than one product id (multiline purchases which are
+                // not supported) so it's safe to pickup the first one.
+                // We would need to refactor this if/when we support multiple tokens per call.
```
aboedo, line 110: I guess one important thing to keep in mind here is that if we did add backend-side support, this SDK would remain unchanged in the wild. We can update it, but many devs will continue using this version which mistakenly refers to our backend not working

aboedo: It does make me wonder, given that consumption really happens at the token level, whether we should have had the consumption field be top-level in post-receipt response so that it unequivocally tells you whether or not to consume

tonidero: I mean, this SDK won't work with multiline purchases. As you said, currently, we only support sending one token per request. The idea of this response was to future-proof it so, in case we add support for multiple tokens/products per request, we don't need to change the response for this data.

> the consumption field be top-level in post-receipt response

We can do that, but as I mentioned, if we add support for multiple tokens per request, we would need to change this part of the response. We might be doing this too early, but I liked giving it more flexibility like this, however, it's completely unneeded right now yeah. Also, if we did that, I would still add it into a `purchased_product` field or similar so it's clear that it's information about the product, but only have one product in the response, in case we want to add other info about the product in the future.

Lmk if you feel strongly about it and I can still change the backend side.

tonidero:
Backend has been merged. Would appreciate another review for this!

purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
```diff
@@ -1470,6 +1482,114 @@ class PostReceiptHelperTest {
...
+                finishTransactions = true,
+                purchase = mockStoreTransaction,
+                shouldConsume = false,
+                initiationSource = initiationSource
+            )
+        }
+    }
+
+    @Test
+    fun `postTransactionAndConsumeIfNeeded tries to consume products if product data not available`() {
```
aboedo, line 1540: 👍

aboedo (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/strings/PurchaseStrings.kt
```diff
@@ -7,6 +7,8 @@ internal object PurchaseStrings {
         "BillingClient is not connected yet."
     const val CONSUMING_PURCHASE = "Consuming purchase with token %s"
     const val CONSUMING_PURCHASE_ERROR = "Error consuming purchase. Will retry next queryPurchases. %s"
+    const val NOT_CONSUMING_IN_APP_PURCHASE_ACCORDING_TO_BACKEND = "Not consuming in-app purchase according to" +
+        " server configuration."
```
aboedo, line 11: let's expand on the implications or maybe add a link to docs

aboedo (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/strings/PurchaseStrings.kt
```diff
@@ -8,7 +8,8 @@ internal object PurchaseStrings {
     const val CONSUMING_PURCHASE = "Consuming purchase with token %s"
     const val CONSUMING_PURCHASE_ERROR = "Error consuming purchase. Will retry next queryPurchases. %s"
     const val NOT_CONSUMING_IN_APP_PURCHASE_ACCORDING_TO_BACKEND = "Not consuming in-app purchase according to" +
-        " server configuration."
+        " server configuration. This is expected for non-consumable products. User won't be able to purchase this" +
+        " product again."
```
JayShortway, line 12: Not sure how these are surfaced, but maybe: 
 ```suggestion
        " server configuration. This is expected for non-consumable products. Users won't be able to purchase this" +
        " product again."
``` 
Or:
```suggestion
        " server configuration. This is expected for non-consumable products. The user won't be able to purchase this" +
        " product again."
```
What do you think?

tonidero: These are logcat logs, as long as the developer has set a specific `LogLevel` like this: https://www.revenuecat.com/docs/test-and-launch/debugging. And yeah, I think that's fine 👍

JayShortway: Ah thanks for the context!

PR #1706, opened 2024-05-27
Upgrade Kotlin to 2.0.0

vegaro:
- [x] Upgrade Kotlin to 2.0.0
- [ ] Check hybrids keep working
  - [ ] Flutter
  - [x] Unity
- [x] Upgrade Compose to compatible version
- [x] Check Android manifest differences (compared purchase tester generated AndroidManifest.xml before and after)
- [ ] Check full dependency tree changes
- [x] Dokka keeps working
- [ ] Check that clients on older version of compose can still use paywalls

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ plugins {
 //    The request for this plugin could not be satisfied because the plugin is already on the classpath with an
 //    unknown version, so compatibility cannot be checked.
     alias libs.plugins.kotlin.parcelize apply false
+    alias libs.plugins.compose.compiler apply false
 }
 
 dependencies {
--- a/examples/paywall-tester/build.gradle
+++ b/examples/paywall-tester/build.gradle
@@ -46,7 +46,7 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
     kotlinOptions {
-        jvmTarget = '19'
+        jvmTarget = '1.8'
     }
     viewBinding {
         enabled = true
--- a/examples/paywall-tester/build.gradle
+++ b/examples/paywall-tester/build.gradle
@@ -1,6 +1,7 @@
 plugins {
     alias(libs.plugins.android.application)
     alias(libs.plugins.kotlin.android)
+    alias libs.plugins.compose.compiler
 }
 
 apply from: "$rootProject.projectDir/base-application.gradle"
@@ -45,16 +46,10 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
     kotlinOptions {
-        jvmTarget = '1.8'
-    }
-    buildFeatures {
-        compose true
+        jvmTarget = '19'
     }
     viewBinding {
         enabled = true
-    }
-    composeOptions {
-        kotlinCompilerExtensionVersion '1.4.0-alpha02'
     }
     packaging {
         resources {
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,8 +1,8 @@
 [versions]
 agp = "8.1.3"
-kotlin = "1.7.21"
+kotlin = "2.0.0"
 # Can't update until we use more recent kotlin. 1.5.0 uses Kotlin 1.8.0
-kotlinxSerializationJSON = "1.4.1"
+kotlinxSerializationJSON = "1.6.3"
 billing = "6.2.1"
 lifecycle = "2.5.0"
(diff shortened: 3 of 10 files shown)


examples/paywall-tester/build.gradle
```diff
@@ -45,17 +46,11 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
     kotlinOptions {
-        jvmTarget = '1.8'
-    }
-    buildFeatures {
-        compose true
+        jvmTarget = '19'
```
JayShortway, line 49: This seems quite high and doesn't correspond to the `targetCompatibility` set above. Is this needed for anything? (Also in `api-tester/build.gradle`.)

build.gradle
```diff
@@ -16,6 +16,7 @@ plugins {
 //    The request for this plugin could not be satisfied because the plugin is already on the classpath with an
 //    unknown version, so compatibility cannot be checked.
     alias libs.plugins.kotlin.parcelize apply false
+    alias libs.plugins.compose.compiler apply false
```
JayShortway, line 19: Nice improvement!

gradle/libs.versions.toml
```diff
@@ -109,9 +110,9 @@ robolectric = { module = "org.robolectric:robolectric", version.ref = "roboelect
 
 tink = { module = "com.google.crypto.tink:tink-android", version.ref = "tink" }
 
-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
+coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version = "1.8.1" }
 
-kotlin-bom = "org.jetbrains.kotlin:kotlin-bom:1.8.0"
+kotlin-bom = "org.jetbrains.kotlin:kotlin-bom:2.0.0"
```
JayShortway, line 115:
```suggestion
kotlin-bom = { module = "org.jetbrains.kotlin:kotlin-bom", version.ref = "kotlin" }
```

JayShortway:
Is my assumption correct that updating to Kotlin 2.0.0 means implementing apps will need to at least be on 1.8.0?

JayShortway:
Looks good, nice update! Just a few minor things.

examples/paywall-tester/build.gradle
vegaro: hmm, good point, I saw it in a snippet in https://developer.android.com/jetpack/androidx/releases/compose-kotlin#declaring_dependencies 

what do you think? if I understand correctly, this is the version of the jvm our code should be compatible with? So maybe we should keep it as 1.8?

vegaro:
> Is my assumption correct that updating to Kotlin 2.0.0 means implementing apps will need to at least be on 1.8.0?

yes

examples/paywall-tester/build.gradle
JayShortway: Yeah that's how I understand it as well. [Java 17 is the latest supported version on Android 14](https://developer.android.com/build/jdks#compileSdk), and [desugaring is only available for Java 11 & 8](https://developer.android.com/studio/write/java8-support#library-desugaring). I have no idea why that snippet sets it to 19. 😕  
  
I think keeping it at 1.8 (8) is the safest indeed, unless we need any of the new Java features.

tonidero:
Looks good so far! We should also check that clients on older version of compose can still use paywalls (or at least to figure out if there is a min version)

JayShortway:
Just came across [this](https://kotlinlang.org/docs/whatsnew20.html#new-directory-for-kotlin-data-in-gradle-projects), might be good to add a line to `.gitignore`? 

> With this change, you may need to add the `.kotlin` directory to your project's `.gitignore` file.

vegaro:
alright, after a bunch of testing

Looks like we can only upgrade kotlin to 1.8 and 2.0 even 1.9 are going to have to wait. The main reasoning is:
- Capacitor uses kotlin 1.8
- React Native latest version (0.74) uses 1.9, but that’s very recent
- React Native 0.72 uses 1.7, and I’ve managed to build an app in this version if we upgrade to 1.8. If we upgrade to 1.9 it starts breaking. I also tried with 0.71 (using 1.6) and it also worked . If anyone has issues, we at least know that 0.73 should work fine, so we can recommend to upgrade.

I will close the 2.0 PR and open one upgrading to 1.8

PR #1709, opened 2024-05-29
Fix non-consumables not getting acknowledged

vegaro:
After #1697 , we stopped consuming non-consumables if the backend indicates to not consume them. Non-consumables still need to be acknowledged, otherwise they get refunded, this PR fixes that.

For reference, https://developer.android.com/google/play/billing/integrate#non-consumable-products

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -371,23 +371,21 @@ internal class BillingWrapper(
                     initiationSource,
                     onConsumed = deviceCache::addSuccessfullyPostedToken,
                 )
-            } else if (finishTransactions) {
+            } else if (finishTransactions && !alreadyAcknowledged) {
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Ohh thanks for fixing this!! We should confirm that the otp can still be consumed after it's been acknowledged just in case a developer wants to make a non-consumable a consumable

vegaro:
It is possible to consume after only acknowledging. But that made me think of something else, that it is that we are saving tokens in the cache even for non-consumables. That means that when calling `queryPurchases`, since that token is in the cache, we will post it to the backend, and we will never know if the product has been changed to be a consumable and needs to be consumed. The only way we will consume that purchase after being made a consumable is if the app is reinstalled (and the cache cleaned) or if the user restores purchases (where we don't check the token cache).

tonidero:
Yeah that's a good point, I considered it but seems like an edge case to me. So this would affect devs that turn a non-consumable into a consumable and want the products to be consumed immediately. And for those, we can ask them to perform a restore as you said... Do you think we should support this use case?

PR #1708, opened 2024-05-28
Deprecate observerMode in favor of purchasesAreCompletedBy

JayShortway:
## Changed
- Adds `PurchasesAreCompletedBy` enum.
- Adds `PurchasesConfiguration.Builder.purchasesAreCompletedBy()`
- Deprecates `PurchasesConfiguration.Builder.observerMode()`, with a nice `ReplaceWith` action.
- Adds `Purchases.syncAmazonPurchase()`.
- Deprecates `Purchases.syncObserverModeAmazonPurchase`, with a nice `ReplaceWith` action.
- Deprecates `Purchases.finishTransactions`, with an okay `ReplaceWith` action.
- Adds API tests for `PurchasesAreCompletedBy`.
- Adds API tests for `purchasesAreCompletedBy()`.
- PaywallsTester, PurchaseTester and the MagicWeather samples use `purchasesAreCompletedBy()` instead of `observerMode()`.
- Adds some unit tests for `purchasesAreCompletedBy()`.

## Not changed
This PR does not remove all internal mentions of the terms "observer mode" and "finish transactions" from the codebase. Will do that in a follow-up. 

## Blockers
- [x] Get consensus that this is the route we want to take. 
    - [x] Enum instead of boolean? 👉 enum
    - [x] Deprecate or remove straight away? 👉 deprecate
    - [x] Decide on a target branch (`8.0.0-dev` or `main`). 👉 main
- [x] Wait for `/finishing-transactions` to go live, as it is mentioned in the Magic Weather samples. See https://github.com/RevenueCat/docusaurus/pull/261 👉 moved to another PR: https://github.com/RevenueCat/purchases-android/pull/1721

PR #1710, opened 2024-05-29
Remove internal observermode mentions

JayShortway:
**Note:** this is a draft, because it is based on #1708, which is also still a draft.

This removes all internal mentions of the term "observer mode", except when still necessary to test deprecated APIs.


purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
```diff
@@ -86,7 +86,7 @@ internal class AmazonBilling(
     constructor(
```
JayShortway, line 86: The comment above this constructor says it is used via reflection. I couldn't find any reflective usage of it. Would love for someone to confirm that 😄

JayShortway:
How deep do we want to go with `PurchasesAreCompletedBy`? In my opinion, at some point components just need to know the relevant interpretation of it, without knowing all enum cases. For most components, that means knowing whether to finish transactions or not. 

In practice this means I added `PurchasesAreCompletedBy` to the `AppConfig` constructor, but not to anything "below" that. 

What are your thoughts?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesAreCompletedBy.kt
```diff
@@ -20,3 +20,12 @@ enum class PurchasesAreCompletedBy {
...
 }
+
+/**
+ * Whether the Purchases SDK should finish transactions.
+ */
+internal val PurchasesAreCompletedBy.finishTransactions: Boolean
+    get() = when (this) {
+        PurchasesAreCompletedBy.REVENUECAT -> true
+        PurchasesAreCompletedBy.MY_APP -> false
+    }
```
JayShortway, line 31: What are your thoughts on this? Intended as a single source of truth. Related to https://github.com/RevenueCat/purchases-android/pull/1710#issuecomment-2147057244.

tonidero: This is nice! 👍 I like this pattern a lot :) It could also be a parameter for the enum, but IMO this is cleaner.

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
tonidero: I believe that refers to this: https://github.com/RevenueCat/purchases-android/blob/b8bae4ad69e52ab5c98dcfc1e6d82715eecf4ed5/purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt#L33. Before, we actually used reflection but we refactored it a while back so the comment seems outdated.

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
```diff
@@ -580,9 +580,9 @@ internal class AmazonBilling(
         purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
     }
 
-    private fun checkObserverMode(): Boolean {
-        return if (observerMode) {
-            log(LogIntent.AMAZON_WARNING, AmazonStrings.WARNING_AMAZON_OBSERVER_MODE)
+    private fun finishingTransactionsIsDisabled(): Boolean {
```
tonidero, line 583: Nitpick but I would maybe rename to `shouldFinishTransactions` even if we have to negate the usages... This could also work but the naming makes me think of a property rather than a function, so maybe we can just convert it to a computed property instead.

tonidero (approved):
Looks great!

> How deep do we want to go with PurchasesAreCompletedBy?

Well, I think this level is ok for now... I guess ideally we change to using that throughout the codebase, but we can do that in different steps/PRs if needed.

> We could opt to rename finishTransactions to something like completePurchases or finishPurchases

IMO, we can try to use the same nomenclature/type throughout the codebase, so if we use `PurchasesAreCompletedBy`, I think that should be enough IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesAreCompletedBy.kt
JayShortway: Thanks! Yea I think this is cleaner too. 🙂 Influenced by [extension-oriented design](https://elizarov.medium.com/extension-oriented-design-13f4f27deaee).

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
JayShortway: Yup, you're totally right. The idea behind the name was fluent if-statements: 
```kotlin
if (finishingTransactionsIsDisabled()) // then
```  
But it's better not to have negatives ("disabled") in names. Renamed!

PR #825, opened 2023-02-28
Purchase.Builder

beylmk:
Adding the `isPersonalizedPrice` parameter to the purchase functions acted as a bit of a forcing function for us to refactor all of the purchase methods. See that conversation [here](https://github.com/RevenueCat/purchases-android/pull/796).

Ultimately, what was decided is that we should use the builder pattern for purchases. Now, rather than having 8+ different purchase functions, we have 1, with a Builder for options like `oldProductId`, `prorationMode`, and `isPersonalizedPrice`. The builder can be constructed with either a `StoreProduct`, `Package`, or `SubscriptionOption`, to cover all of the purchaseable entities.

Making this change means we need to unify `ProductChangeCallback` and `PurchaseCallback`, since the same method could be used for a normal purchase or a product change. This means making a single interface that has a nullable `StoreTransaction`.

This is a WIP, but would like to get eyes on it early in the process, as it touches a lot of code.

Open Questions:
* I'd love feedback on how to handle deprecation of the callbacks. It'd be great to keep the name `PurchaseCallback` and just make the `StoreTransaction` nullable, but I don't know of an easy way to notate that with annotations... I think in the past we've just fully deprecated and added something new, but I don't want to give up the good name. Perhaps `PurchaseListener` works, but we already did a big refactor to make listener -> callback.
* Should we allow all proration modes when the deprecated versions of the purchase methods are called?

PR #1711, opened 2024-05-29
SDK-3469 Remove deprecated purchase methods

JayShortway:
This PR leaves just `purchase(PurchaseParams)` as the only way to purchase.

## Breaking changes
- Removes `purchasePackage()`
- Removes `purchaseProduct()`
- Removes `purchasePackageWith()`
- Removes `purchaseProductWith()`
- Removes `UpgradeInfo`


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedOfferingFragment.kt
```diff
@@ -1,266 +0,0 @@
...
-import com.revenuecat.purchases.models.GoogleReplacementMode
-import com.revenuecat.purchases.models.StoreProduct
-import com.revenuecat.purchases.models.StoreTransaction
-import com.revenuecat.purchases.purchasePackageWith
-import com.revenuecat.purchases.purchaseProductWith
-import com.revenuecat.purchases_sample.R
-import com.revenuecat.purchases_sample.databinding.FragmentOfferingBinding
-
-@SuppressWarnings("TooManyFunctions")
-class DeprecatedOfferingFragment : Fragment(), DeprecatedPackageCardAdapter.PackageCardAdapterListener {
```
JayShortway, line 32: This file was deleted because of the following snippet in `OverviewFragment`:  
```kotlin
val directions = OverviewFragmentDirections.actionOverviewFragmentToOfferingFragment(offering.identifier)
// uncomment line below and comment line above to test deprecated purchase methods
// val directions = OverviewFragmentDirections.
// actionOverviewFragmentToDeprecatedOfferingFragment(offering.identifier)
```  
Let me know if that was the wrong conclusion I took!

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
```diff
@@ -114,7 +114,7 @@ internal class PurchasesTest : BasePurchasesTest() {
     // region purchasing
 
     @Test
-    fun `deprecated upgrade defaults ProrationMode to null if not passed`() {
+    fun `upgrade defaults ReplacementMode to WITHOUT_PRORATION if not passed`() {
```
JayShortway, line 117: I changed this test to assert the current behavior, assuming that that behavior is correct.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -223,142 +221,6 @@ class Purchases internal constructor(
...
-     *
-     * @param [activity] Current activity
-     * @param [storeProduct] The StoreProduct of the product you wish to purchase
-     * @param [callback] The PurchaseCallback that will be called when purchase completes.
-     */
-    @Deprecated(
-        "Use purchase() and PurchaseParams.Builder instead",
-        ReplaceWith("purchase()"),
-    )
-    fun purchaseProduct(
```
tonidero, line 280: IMO I would leave these functions that don't have a `UpgradeInfo`. I think some people might still be using those, and we should try to minimize the changes if possible. Lmk if you think otherwise though!

JayShortway: Haha I'm fine with both, but I went off of @vegaro's [comment](https://github.com/RevenueCat/purchases-android/pull/1707#discussion_r1618415370): 
> only leave the `purchase` function that takes `PurchaseParams.Builder` 
  
I'm happy to just do what the 2 of you agree on 😄

vegaro: These were deprecated in the current major https://github.com/RevenueCat/purchases-android/pull/825, that's where my suggestion for removal came from. If we don't remove them now, they will live for another major. Current major was released in September, I think 9 months is enough time to migrate to the `purchase` function?

tonidero: Right, but I think it shouldn't be a big problem to keep those around right? And I think as long as it's not a problem, we should keep the deprecated functions. If at some point it becomes a problem, we can reconsider.

tonidero: Agreed, I added it to our sync. Let's discuss it then!

vegaro: I spoke to @tonidero yesterday and have concluded that it is probably better to keep them, so we focus on the developers and not on our own development experience. We can always remove them in the future if they become a pain to maintain. I think it's still worth to discuss in the meeting today so we all align for the future though.

Sorry for the back and forth, and for making you work on something that is getting scratched 😢

JayShortway: No worries, it's good exercise! 😄 I'll await today's meeting just in case, and will make the changes afterwards.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedOfferingFragment.kt
JayShortway: No longer deleted.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: This is now done!

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedPackageCardAdapter.kt
```diff
@@ -38,12 +39,13 @@ class DeprecatedPackageCardAdapter(
             binding.currentPackage = currentPackage
             binding.isSubscription = product.type == ProductType.SUBS
             binding.isActive = activeSubscriptions.contains(product.id)
+            // Upgrades are no longer possible with deprecated methods.
```
tonidero, line 42: Honestly, I think it might make sense to just remove these deprecated classes now... I believe it was mostly used to test during the migration to the new ReplacementMode, but I think it wouldn't be very useful now... But this can also happen separately.

tonidero (approved):
This looks great!

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedPackageCardAdapter.kt
JayShortway: Alright! My vote would be to do it separately indeed, to avoid this PR from creeping up in scope.

JayShortway: Also quicker v8 beta 🚀 😄

PR #1715, opened 2024-05-30
Fix purchase tester offerings not reloading

tonidero:
### Description
I've run into some issues with purchase tester for a while where the offerings were not updated after changing screens or purchasing. This simplifies the reloading so it loads always. There might be more performant options, but considering this is a test app, I didn't want to dedicate too much time to it.


JayShortway (approved):
If it works, it works! 

I was also thinking that _if_ this were a Compose app, it would be a lot easier to do quick changes to it, as Compose generally requires less code to write. But yea, migrating to Compose is a big change haha.

tonidero:
> I was also thinking that _if_ this were a Compose app, it would be a lot easier to do quick changes to it, as Compose generally requires less code to write. But yea, migrating to Compose is a big change haha. 

Yup definitely, I think moving to Compose would be great but we can tackle that when we have time

PR #1717, opened 2024-05-30
Support pending prepaid purchases

tonidero:
### Description
PR to support pending prepaid purchase as a new feature added in BC7. This will be enabled by default but can be disabled through an option in `PurchasesConfiguration`: `pendingTransactionsForPrepaidPlansEnabled `.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
@@ -19,7 +19,7 @@ open class PurchasesConfiguration(builder: Builder) {
     val diagnosticsEnabled: Boolean
     val dangerousSettings: DangerousSettings
     val verificationMode: EntitlementVerificationMode
-    val pendingPrepaidSubscriptionsEnabled: Boolean
+    val pendingTransactionsForPrepaidPlansEnabled: Boolean
 
     init {
         this.context = builder.context
@@ -32,7 +32,7 @@ open class PurchasesConfiguration(builder: Builder) {
         this.verificationMode = builder.verificationMode
         this.dangerousSettings = builder.dangerousSettings
         this.showInAppMessagesAutomatically = builder.showInAppMessagesAutomatically
-        this.pendingPrepaidSubscriptionsEnabled = builder.pendingPrepaidSubscriptionsEnabled
+        this.pendingTransactionsForPrepaidPlansEnabled = builder.pendingTransactionsForPrepaidPlansEnabled
     }
 
     @Suppress("TooManyFunctions")
@@ -66,7 +66,7 @@ open class PurchasesConfiguration(builder: Builder) {
         internal var dangerousSettings: DangerousSettings = DangerousSettings()
 
         @set:JvmSynthetic @get:JvmSynthetic
-        internal var pendingPrepaidSubscriptionsEnabled: Boolean = true
+        internal var pendingTransactionsForPrepaidPlansEnabled: Boolean = false
 
         /**
          * A unique id for identifying the user
@@ -180,12 +180,12 @@ open class PurchasesConfiguration(builder: Builder) {
         }
 
         /**
-         * Disable this setting if you don't want to allow pending purchases for prepaid subscriptions (only supported
+         * Enable this setting if you want to allow pending purchases for prepaid subscriptions (only supported
          * in Google Play). Note that entitlements are not granted until payment is done.
-         * Default is enabled.
+         * Default is disabled.
          */
-        fun pendingPrepaidSubscriptionsEnabled(pendingPrepaidSubscriptionsEnabled: Boolean) = apply {
-            this.pendingPrepaidSubscriptionsEnabled = pendingPrepaidSubscriptionsEnabled
+        fun pendingTransactionsForPrepaidPlansEnabled(pendingTransactionsForPrepaidPlansEnabled: Boolean) = apply {
+            this.pendingTransactionsForPrepaidPlansEnabled = pendingTransactionsForPrepaidPlansEnabled
         }
 
         /**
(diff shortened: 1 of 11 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -173,6 +179,15 @@ open class PurchasesConfiguration(builder: Builder) {
             this.dangerousSettings = dangerousSettings
         }
 
+        /**
+         * Disable this setting if you don't want to allow pending purchases for prepaid subscriptions (only supported
+         * in Google Play). Note that entitlements are not granted until payment is done.
+         * Default is enabled.
+         */
+        fun pendingPrepaidSubscriptionsEnabled(pendingPrepaidSubscriptionsEnabled: Boolean) = apply {
+            this.pendingPrepaidSubscriptionsEnabled = pendingPrepaidSubscriptionsEnabled
```
tonidero, line 188: Note that this won't be used in the `customEntitlementComputationMode` flavor

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -173,6 +179,15 @@ open class PurchasesConfiguration(builder: Builder) {
             this.dangerousSettings = dangerousSettings
         }
 
+        /**
+         * Disable this setting if you don't want to allow pending purchases for prepaid subscriptions (only supported
+         * in Google Play). Note that entitlements are not granted until payment is done.
+         * Default is enabled.
+         */
+        fun pendingPrepaidSubscriptionsEnabled(pendingPrepaidSubscriptionsEnabled: Boolean) = apply {
```
JayShortway, line 187: I noticed Google calls this feature "pending transactions for prepaid plans", e.g. [here](https://developer.android.com/google/play/billing/subscriptions#pending). What do you think about something like `pendingTransactionsForPrepaidPlansEnabled()`?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -173,6 +179,15 @@ open class PurchasesConfiguration(builder: Builder) {
             this.dangerousSettings = dangerousSettings
         }
 
+        /**
+         * Disable this setting if you don't want to allow pending purchases for prepaid subscriptions (only supported
+         * in Google Play). Note that entitlements are not granted until payment is done.
+         * Default is enabled.
```
JayShortway, line 185: Would a developer who pays no attention to this and just updates to `8.0.0` notice any differences from this being enabled?

JayShortway (approved):
Makes a lot of sense! Just a suggestion and a question.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
tonidero: Well, main difference is that users can pay for prepaid plans (if they have any) after. But in any case, entitlements won't be granted until it's payed. So it just provides more flexibility for payment for the end customer. I believe it's the most sensible default. I can't really think of why you would like to disable this, but wanted to provide the option just in case.

JayShortway: Alright, yea makes sense. Thanks for the additional context! Just wanted to make sure nothing unexpected happens to developers 🙂

tonidero: Yup! It's good to double check :)

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -62,6 +65,9 @@ open class PurchasesConfiguration(builder: Builder) {
         @set:JvmSynthetic @get:JvmSynthetic
         internal var dangerousSettings: DangerousSettings = DangerousSettings()
 
+        @set:JvmSynthetic @get:JvmSynthetic
+        internal var pendingPrepaidSubscriptionsEnabled: Boolean = true
```
aboedo, line 69: maybe we can default to false for now while the FF is disabled by default?

aboedo: it's default off for BillingClient after all (in that you need to call something explicitly to enable it)

aboedo: I think updating to default to true in the future would be fine and wouldn't require a major since it won't have any API changes

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
aboedo: +1. 

As a side note, if you say `pendingTransactionsForPrepaidPlansEnabled` 3 times in a row quickly I think that nets you an honorary Master's Degree in Diction

aboedo (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/1717/commits/67a43a68ca9aad093b14cd717f942a85c4ede392

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/1717/commits/67a43a68ca9aad093b14cd717f942a85c4ede392

PR #1719, opened 2024-05-30
Remove `force-release-without-waiting-for-tests` job

vegaro:
This was causing the release PRs to never turn green completely

CleanShot 2024-05-30 at 16 34 07@2x

Also, it breaks "Rerun from fails" in CircleCI since there's a approval job, which causes the workflow to never fail completely if another job fails.


tonidero (approved):
Yesssss 🙌

PR #1720, opened 2024-05-30
Added new paywall template to support multiple tiered subscriptions

joshdholtz:
### Motivation

Template 5 but with multiple tiers 😊 

Supports...
- [x] Portrait
- [x] Landscape
- [x] Footer
- [ ] Accessibility

### Description

- Updated `PackageConfigurationType ` for `MULTITIER`
- Updated `PaywallTemplate` for `TEMPLATE_7`
- Updated `PaywallData` with...
  - `localizationByTier`
- Updated `PaywallData.Configuration` with...
  - `packages` defaults to `emptyList()
  - New `imagesByTier: Map<String, ColorInformation>?`
  - New `colorsByTier: Map<String, Images>?`
  - New `tiers: List<Tier>?`
- Updated `PaywallData.Configuration.ColorInfomation` with...
  - `tierControlBackground: PaywallColor?`
  - `tierControlForeground: PaywallColor?`
  - `tierControlSelectedBackground: PaywallColor?`
  - `tierControlSelectedForeground: PaywallColor?`

#### Screenshots

| Portrait | Landscape | Footer | Footer Condensed |
|--------|--------|--------|--------|
| Screenshot 2024-06-17 at 10 41 36 AM | Screenshot 2024-06-17 at 10 36 19 AM | Screenshot 2024-06-17 at 10 40 29 AM | Screenshot 2024-06-17 at 10 40 33 AM | 

#### Video (with animations)

https://github.com/RevenueCat/purchases-android/assets/401294/b06d0dd5-3449-44d9-ad6f-9894bc0edc61


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -194,6 +244,14 @@ data class PaywallData(
                 get() = listOfNotNull(header, background, icon)
         }
 
+        @Serializable
+        data class Tier(
+            val id: String,
+            val packages: List<String>,
+            @SerialName("default_package")
+            val defaultPackage: String,
```
tonidero, line 252: Nitpick I wonder if we should rename to `defaultPackageIdentifier`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -9,34 +9,150 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
...
+                        defaultPackage = packageInfo.default,
+                        packages = packageInfo.all,
+                    )
+                }
+
+                Result.success(
+                    Pair(
+                        locale,
+                        TemplateConfiguration.PackageConfiguration.MultiTier(
+                            firstTier = allTierInfos.first(),
```
tonidero, line 136: Hmm maybe we want to deal with this later, but it would be good to have a way to specify the initial tier from the dashboard?

joshdholtz: I agree with this! I'll add it to the schema and then we can add a dashboard setting for it later 😇

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
joshdholtz: I think this makes it clearer! Should I also rename the `defaultPackage` that is on `Configuration`? 🤔

tonidero: Ohhh right.... I mean, that would be a breaking change... we could probably deprecate it and add a new one, but I think it might be ok to keep it like this then... We can consider deprecating and renaming separately, and it would be good to be consistent.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -88,15 +87,46 @@ data class PaywallData(
...
 
+    val tieredLocalizedConfiguration: Pair<Locale, Map<String, LocalizedConfiguration>>
+        @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
+        get() {
+            return tieredConfigForLocales(locales = getDefaultLocales())
+        }
+
+    @VisibleForTesting
+    @RequiresApi(Build.VERSION_CODES.LOLLIPOP)
+    fun tieredConfigForLocales(locales: List<Locale>): Pair<Locale, Map<String, LocalizedConfiguration>> {
```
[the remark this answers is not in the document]
joshdholtz: Ah, I need to add a test for this I think!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -82,15 +82,43 @@ data class PaywallData(
...
      */
     @Serializable
     data class Configuration(
         /**
          * The list of package identifiers this paywall will display.
+         *
+         * This defaults to an empty list for multi-tier configurations.
          */
-        @SerialName("packages") val packageIds: List<String>,
+        @SerialName("packages") val packageIds: List<String> = emptyList(),
```
joshdholtz, line 121: This now defaults to `emptyList()` since its not used by multi-tier paywalls

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -56,13 +56,14 @@ internal fun Footer(
...
     childModifier: Modifier = Modifier,
+    colors: TemplateConfiguration.Colors = templateConfiguration.getCurrentColors(),
     allPlansTapped: (() -> Unit)? = null,
 ) {
     Column {
         Footer(
             mode = templateConfiguration.mode,
             configuration = templateConfiguration.configuration,
-            colors = templateConfiguration.getCurrentColors(),
+            colors = colors,
```
joshdholtz, line 66: Footer colors depends on current tier colors for new template need to manually pass in the current tier colors

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -50,18 +51,37 @@ internal fun Offering.validatedPaywall(
...
-        return Result.failure(invalidIconsError)
+            val invalidIconsError = localizedConfiguration.validateIcons()
+            if (invalidIconsError != null) {
+                return Result.failure(invalidIconsError)
+            }
+        }
+        PackageConfigurationType.MULTITIER -> {
+            val (_, localizedConfigurationByTier) = tieredLocalizedConfiguration
+
+            localizedConfigurationByTier.entries.forEach { (_, localizedConfiguration) ->
```
joshdholtz, line 73: I probably need to validate tier count here 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
```diff
@@ -0,0 +1,669 @@
...
+@Composable
+internal fun Template7(
+    state: PaywallState.Loaded,
+    viewModel: PaywallViewModel,
+) {
+    val packagesConfig = state.templateConfiguration.packages
+    require(packagesConfig is TemplateConfiguration.PackageConfiguration.MultiTier) {
+        // This should never happen
+        "The configuration is not MultiTier"
+    }
```
joshdholtz, line 102: Should make sure that this case is cover in the validate check for multi-tier

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/MultiTierTemplateConfigurationFactoryTest.kt
```diff
@@ -0,0 +1,210 @@
...
+        )
+
+        val expectedConfiguration = TemplateConfiguration.PackageConfiguration.MultiTier(
+            firstTier = basicTier,
+            allTiers = listOf(basicTier, standardTier, premiumTier)
+        )
+
+        assertThat(packageConfiguration.firstTier).isEqualTo(expectedConfiguration.firstTier)
+//        assertThat(packageConfiguration.default).isEqualTo(expectedConfiguration.default)
+//        assertThat(packageConfiguration.all).containsExactly(*expectedConfiguration.all.toTypedArray())
```
joshdholtz, line 197: Ahhhh, need to re-enable these tests!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -60,25 +176,47 @@ internal object PackageConfigurationFactory {
             )
         }
 
-        val firstPackage = packageInfos.first()
-        val defaultPackage = packageInfos.firstOrNull { it.rcPackage.identifier == default } ?: firstPackage
+        return Pair(locale, packageInfos)
```
vegaro, line 179: Same tiny suggestion here about using the `to` notation, but feel free to ignore

```suggestion
        return locale to packageInfos
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
```diff
@@ -0,0 +1,132 @@
...
+    val height = 40.dp
+    val selectedTierPadding = 2.dp
+}
+
+@Composable
+@Suppress("LongParameterList")
+internal fun SelectedTierView(
+    selectedTier: TemplateConfiguration.TierInfo,
+    backgroundSelectedColor: Color,
+    foregroundSelectedColor: Color,
```
vegaro, line 46: There are a lot of places in this PR where there's a background and a foreground color being passed around. Maybe we could create a data class that holds a foreground and a background color? I think that will prevent issues, with passing one instead of another to functions, and also reduce parameters in a bunch of functions

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
```diff
@@ -0,0 +1,132 @@
...
+    ) {
+        val optionWidth = with(density) { (totalWidthPx / tiers.size).toDp() }
+        val indicatorOffset by animateDpAsState(targetValue = optionWidth * selectedIndex, label = "tier_switcher")
+
+        Box(
+            Modifier
+                .offset(x = indicatorOffset)
+                .fillMaxHeight()
+                .width(optionWidth)
+                .padding(2.dp)
```
vegaro, line 99: This could be `selectedTierPadding`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -9,34 +9,150 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
...
+                                all = packageInfos,
+                            ),
+                        ),
+                    ),
+                )
+            }
+            PackageConfigurationType.MULTITIER -> { this
+                val tiers = paywallData.config.tiers ?: run {
+                    return Result.failure(PackageConfigurationError("No tier found for $packageIdsInConfig"))
+                }
```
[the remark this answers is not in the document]
joshdholtz: I think so since I'm early returning! Unless there is another way? 🤔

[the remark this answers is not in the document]
vegaro: I mean that I think you can just return without the run?

```
val tiers = paywallData.config.tiers 
    ?: return Result.failure(PackageConfigurationError("No tier found for $packageIdsInConfig"))
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/TemplateConfiguration.kt
```diff
@@ -52,14 +64,52 @@ internal data class TemplateConfiguration(
...
+        data class Single(val singlePackage: PackageInfo) : PackageConfiguration() {
+            override val all: List<PackageInfo>
+                get() = listOf(singlePackage)
+            override val default: PackageInfo
+                get() = singlePackage
+        }
+
+        data class Multiple(val multiPackage: MultiPackage) : PackageConfiguration() {
+            override val all: List<PackageInfo>
+
```
vegaro, line 100: extra line

```suggestion
```

vegaro (approved):
good job on this! it may be worth it adding some tests around the new PaywallData functions. Codecov is complaining they are not covered.

Other than that, I've noticed there's a tiny stroke around the pill of the tier, which I think it's intentional, but in this case it looks strange due to the color selection 😺

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
```diff
@@ -0,0 +1,132 @@
...
+            },
+    ) {
+        val optionWidth = with(density) { (totalWidthPx / tiers.size).toDp() }
+        val indicatorOffset by animateDpAsState(targetValue = optionWidth * selectedIndex, label = "tier_switcher")
+
+        Box(
+            Modifier
+                .offset(x = indicatorOffset)
+                .fillMaxHeight()
+                .width(optionWidth)
```
tonidero, line 98: Hmm just to check, have you tried assigning a percentage of the available size to the size here? So having this be `fillMaxWidth(1f/tiers.size)`. Not sure if that would work, but it would avoid the recomposition when the size is calculated.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -9,34 +9,196 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
...
+                id = tier.id,
+                name = tierName,
+                defaultPackage = packageInfo.default,
+                packages = packageInfo.all,
+            )
+        }
+
+        return Result.success(
+            locale to TemplateConfiguration.PackageConfiguration.MultiTier(
+                firstTier = allTierInfos.first(),
```
tonidero, line 129: Just checking, but in the end we're allowing to select the default tier in a separate PR?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
tonidero: Agreed

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
tonidero: So validate that there is at least 2+ tiers? Not sure if we could allow 1... Doesn't make much sense, but shouldn't break anything so 🤷

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
tonidero: I wonder if we need to make the transition between colors when switching tiers smoother... Feels slightly jumpy now... But maybe we should check with design to make sure.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
```diff
@@ -0,0 +1,668 @@
...
+        mutableStateOf(state.templateConfiguration.mode != PaywallMode.FOOTER_CONDENSED)
+    }
+
+    var selectedTier by remember {
+        mutableStateOf(firstTier)
+    }
+
+    val colorForTier = state.templateConfiguration.getCurrentColorsForTier(tier = selectedTier)
+
+    Column(
```
tonidero, line 117: Wonder if we should also try to animate all the main content when switching tiers for a smoother transition since everything can pretty much change... Might be slightly tricky though, depending on what we're animating...

tonidero (commented):
Looks great! Amazing job on this! Also checked out recompositions and it's pretty good performance wise. Left a few comments, but should be close

tonidero:
One other thing, should the package selector be anchored to the bottom of the screen, together with the purchase button? Seems it would look nicer than tied to the main content.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
joshdholtz: Seems like this works! 😊

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
joshdholtz: Added some animations and it feels soooooo much better!

vegaro:
I was going to mention the same as @tonidero . I also don't think Google will like that the selected package details is behind the button, like the subscribe button is visible but it's possible the price is covered, like in this screenshot

CleanShot 2024-06-19 at 12 22 17@2x

joshdholtz:
> I was going to mention the same as @tonidero . I also don't think Google will like that the selected package details is behind the button, like the subscribe button is visible but it's possible the price is covered, like in this screenshot
> 
> CleanShot 2024-06-19 at 12 22 17@2x

Yeaaaahhh, this makes sense! I'll talk to the team about this tomorrow to see what we want to do about it 🤔 Thanks for bringing this up!

PR #1721, opened 2024-06-03
Update observermode docs URLs

JayShortway:
Updates URLs referenced in the docs.


vegaro (approved):
Sweet. Thanks for doing this!

PR #1725, opened 2024-06-05
Stops collecting the ANDROID_ID.

JayShortway:
### Main
- `GoogleDeviceIdentifiersFetcher` no longer collects the `ANDROID_ID`.
### Others
- Removes `ReservedSubscriberAttribute.ANDROID_ID`.
- Removes `SubscriberAttributeKey.DeviceIdentifiers.AndroidID`.
### Tests
- `GoogleDeviceIdentifiersFetcherTests` asserts the `ANDROID_ID` is not collected.
- `SubscriberAttributesManagerTests` no longer returns the `ANDROID_ID` in its mocked `DeviceIdentifiersFetcher` response.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
@@ -56,8 +56,9 @@ class GoogleDeviceIdentifiersFetcherTests {
             assertThat(advertisingID).isNotNull()
             assertThat(advertisingID).isEqualTo("12345")
 
-            val androidID = identifiers[SubscriberAttributeKey.DeviceIdentifiers.AndroidID.backendKey]
-            assertThat(androidID).isEqualTo("androidid")
+            // Assert we don't collect the ANDROID_ID
+            assertThat(identifiers.containsKey("\$androidId")).isFalse()
+            assertThat(identifiers.containsValue("androidid")).isFalse()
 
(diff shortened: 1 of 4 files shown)


purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
```diff
@@ -244,8 +251,9 @@ class GoogleDeviceIdentifiersFetcherTests {
             val advertisingID = identifiers[SubscriberAttributeKey.DeviceIdentifiers.GPSAdID.backendKey]
             assertThat(advertisingID).isNull()
 
-            val androidID = identifiers[SubscriberAttributeKey.DeviceIdentifiers.AndroidID.backendKey]
-            assertThat(androidID).isEqualTo("androidid")
+            // Assert we don't collect the ANDROID_ID
+            assertThat(identifiers.containsKey("\$androidId")).isFalse()
+            assertThat(identifiers.containsValue("androidid")).isFalse()
```
tonidero, line 256: Since this is repeated a few times, we can just extract it to a method I think. That way we also save the comments

tonidero (approved):
Looks good! Just a suggestion

purchases/src/test/java/com/revenuecat/purchases/google/attribution/GoogleDeviceIdentifiersFetcherTests.kt
JayShortway: Yea I'm not opposed to that! Didn't do it at first because I saw some repetition in this test already, which might have been deliberate. It's extracted now!

PR #1728, opened 2024-06-05
Diagnostics: sync only on first activity foregrounded

tonidero:
### Description
As I was testing some things, I noticed we are syncing diagnostics on SDK initialization, which is ok most times, but we don't want to sync when the app is woken by notifications or alarms. Only when the app is open. This moves the syncing process to avoid those extra requests.


vegaro (approved):
Makes sense!

PR #1733, opened 2024-06-05
Offer overrides

joshdholtz:
### Motivation

_Coming soon_

### Description

_Coming soon_


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
joshdholtz, line 1: Made changes here because the tier switcher was trying to expand MAX MAX MAX height which was really bad for footers

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template4.kt
```diff
@@ -279,7 +279,7 @@ private fun SelectPackageButton(
 
     val discountText = packageInfo.localizedDiscount(viewModel.resourceProvider)
 
-    if (packageInfo.discountRelativeToMostExpensivePerMonth != null) {
+    if (packageInfo.localization.offerBadge != null) {
         columnModifier = columnModifier.background(mainColor)
     }
```
joshdholtz, line 284: Relative discount now counts from new logic in localization

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template5.kt
```diff
@@ -448,10 +446,9 @@ private fun CheckmarkBox(isSelected: Boolean, colors: TemplateConfiguration.Colo
 @Composable
 private fun RowScope.DiscountBanner(
     state: PaywallState.Loaded,
-    resourceProvider: ResourceProvider,
     packageInfo: TemplateConfiguration.PackageInfo,
 ) {
-    val text = packageInfo.localizedDiscount(resourceProvider)?.uppercase() ?: return
+    val text = packageInfo.localization.offerBadge?.uppercase() ?: return
```
joshdholtz, line 451: Relative discount now comes from new logic in localization

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -451,6 +451,9 @@ data class PaywallData(
         @SerialName("tier_name")
         @Serializable(with = EmptyStringToNullSerializer::class)
         val tierName: String? = null,
+
+        @SerialName("offer_overrides")
+        val offerOverrides: Map<String, OfferOverride> = emptyMap(),
```
tonidero, line 456: Since these seem to be mostly overriding other existing values I was wondering whether it would make sense to just have the backend override these values in the existing fields instead of having the SDK process these overrides, so we have less logic on the SDKs (This does not apply to the badge which is a new field but could be added with the rest of localizations)... But maybe I'm missing something/don't want to add this logic to the backend, so I'm ok with this!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/snapshottests/TemplateSnapshotTest.kt
```diff
@@ -27,6 +27,8 @@ class TemplateSnapshotTest(private val testConfig: TemplateTestConfig): BasePapa
             TestData.template3Offering,
             TestData.template4Offering,
             TestData.template5Offering,
+            TestData.template7Offering,
+            TestData.template7CustomPackageOffering,
```
tonidero, line 31: 🎉

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
```diff
@@ -82,7 +83,7 @@ internal fun TierSwitcher(
 ) {
     val selectedIndex = tiers.indexOf(selectedTier)
     var totalWidthPx by remember { mutableStateOf(0) }
-    var totalHeightPx by remember { mutableStateOf(0) }
+    var totalHeightRowDp by remember { mutableStateOf(40.dp) }
```
tonidero, line 86: Any reason for this to start at 40.dp?

tonidero (approved):
Just left some questions but looks good to me!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
joshdholtz: @tonidero Oh, for snapshot tests 😅 Otherwise the highlighted background selector box thing doesn't show because its height 0 🤷‍♂️  It's _nearly_ instant when running by a real person snapshots weren't working

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
joshdholtz: @tonidero The backend right now only sends a single name and details to use across every package 😬 We really should have designed it this way from the start to make it more flexible but sadly did not 🤷‍♂️

tonidero: Ohhhh ok makes sense yeah. Thanks for explaining!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/TierSwitcher.kt
tonidero: Yeah snapshot tests don't seem to take the snapshot correctly when there are post layout measurements. Makes sense then!

joshdholtz: I maybe should have added that as a comment 😅 I'll do that in the `multi-tier-paywall` branch before I merge it into `main` 💪

PR #1735, opened 2024-06-06
SDK-3477 Show a toast when a paywall link is clicked without a browser installed

JayShortway:
As the title says.  

~~**Note:** ready for review, but still a draft until the copy is final and translations are added.~~ 

### Blockers
- [x] Copy review
- [x] Translations

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
@@ -130,9 +130,11 @@ private fun Footer(
                 R.string.terms_and_conditions,
                 R.string.terms,
             ) {
-                val msg = context.getString(R.string.no_browser_cannot_open_terms)
-                Toast.makeText(context, msg, Toast.LENGTH_SHORT).show()
-                uriHandler.openUriOrElse(it.toString()) { Logger.w(msg) }
+                uriHandler.openUriOrElse(it.toString()) {
+                    val msg = context.getString(R.string.no_browser_cannot_open_link)
+                    Toast.makeText(context, msg, Toast.LENGTH_SHORT).show()
+                    Logger.w(msg)
+                }
             }
 
             if (configuration.privacyURL != null) {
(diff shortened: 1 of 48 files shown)


ui/revenuecatui/src/main/res/values/strings.xml
```diff
@@ -18,4 +18,6 @@
     <string name="monthly">Monthly</string>
     <string name="weekly">Weekly</string>
     <string name="lifetime">Lifetime</string>
+    <string name="no_browser_cannot_open_terms">No browser installed. Terms and conditions could not be opened.</string>
+    <string name="no_browser_cannot_open_privacy">"No browser installed. Privacy policy could not be opened."</string>
```
JayShortway, line 22: Let me know what you think of this copy. I'll add translations after a second pair of eyes gives the thumbs up 🙂

tonidero (approved):
Love it!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -145,7 +150,9 @@ private fun Footer(
                 R.string.privacy,
             ) {
                 uriHandler.openUriOrElse(it.toString()) {
```
tonidero, line 152: We could potentially rename this function, so it's clear the `else` should only be called when there is no browser. But NABD.

JayShortway: Hmm what name would you suggest? It was inspired by the stdlib's `Map.getOrElse()`, which I think is similar in spirit. ("Do this thing (get / openUri), or if that fails for whatever reason, do this other thing (orElse).")

tonidero: Well, honestly I don't like any of the namings I'm coming up with.... But maybe redefine the function to:
```
fun UriHandler.openUri(uri: String, browserNotFoundAction: () -> Unit) {
```

Alternatively, we can keep it as it is, but catch all exceptions, and pass them as a parameter to the `fallbackAction` like:
```
fun UriHandler.openUriOrElse(uri: String, fallbackAction: (Exception) -> Unit) {
```
And changing the body to catch all exceptions. 

But as I said, I'm also ok with what it is, just that it could be confused as it being called in other exceptions.

JayShortway: Interesting idea! But I think my current preference is to leave it as is. The `Exception` is also kind-of an implementation detail of the function, as the same functionality could have been achieved without a try-catch, but by querying the `PackageManager` for Activities that can resolve the implicit `Intent`.  
  
It's an easy-to-reverse decision if we end up hating the name 😄

tonidero: Yeah I think that's fine 👍

PR #1737, opened 2024-06-06
Support BC7

tonidero:
### Description
This PR adds support for BC7 and related new features.

#### New APIs:
- `SubscriptionOption.installmentInfo` which contains information about installment plans you might want to display in your paywall.
- `PurchasesConfiguration.pendingTransactionsForPrepaidPlansEnabled` allows to enable pending prepaid subscriptions.

#### Breaking changes
- Increase min Sdk to 21. This was increased in BC7
- Remove `UpgradeInfo`
- Remove all `purchasePackage` and `purchaseProduct` methods using `UpgradeInfo`. If you want to perform upgrades/downgrades, please migrate to use the `purchase(PurchaseParams)` method, which has options to perform that operation.


joshdholtz (approved):
:shipit:

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -277,6 +277,7 @@ class Purchases internal constructor(
             val configuration = PurchasesConfiguration.Builder(context, apiKey)
                 .appUserID(appUserID)
                 .dangerousSettings(DangerousSettings(customEntitlementComputation = true))
+                .pendingTransactionsForPrepaidPlansEnabled(true)
```
vegaro, line 280: why is this always true?

tonidero: I believe we want to enable it by default for companies using `customEntitlementComputation`. We can provide an option to disable it later if needed though, but we try to keep the API very small in the `customEntitlementComputation` flavor I think.

README.md
```diff
@@ -51,7 +51,7 @@ Or view / build our Android sample app:
 ## Requirements
 - Java 8+
 - Kotlin 1.5.0+
```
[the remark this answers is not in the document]
tonidero: Ohh good catch! Will update that since I'm writing the migration docs.

PR #1743, opened 2024-06-07
Remove `ExperimentalPreviewRevenueCatUIPurchasesAPI` annotations

vegaro:
Making Paywalls not experimental

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -347,6 +347,7 @@ implementation 'com.revenuecat.purchases:purchases-ui:7.1.0'
 ```
 3. Display a paywall:
 ```kotlin
+@OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
 @Composable
 private fun LockedScreen() {
     YourContent()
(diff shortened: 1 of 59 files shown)


CHANGELOG.md
```diff
@@ -360,11 +359,6 @@ private fun LockedScreen() {
 }
 ```
 
-> **Note**
-> Android paywalls is currently behind an experimental flag: `ExperimentalPreviewRevenueCatUIPurchasesAPI`.
-> It is safe to release app updates with it. We guarantee that paywalls will continue to work and any changes will always be backwards compatible.
-> They are stable, but migration steps may be required in the future. We'll do our best to minimize any changes you have to make.
```
tonidero, line 366: Hmm should we remove this from the CHANGELOG? I think I would keep it and just add the change in the CHANGELOG for the new version.

tonidero (approved):
We should add it to the migration guide. But looks great!

CHANGELOG.md
vegaro: ugh my bad, this is me being stupid and not checking where I was removing this

PR #1745, opened 2024-06-10
Add retries upon Google's SERVICE_UNAVAILABLE error when app in foreground

tonidero:
### Description
We detected this error could happen consistently when acknowledging pending prepaid purchases after payment is performed when the app is closed. This adds some retries to this error, but much smaller than the one that we have when the app is in background (15min vs 4 seconds).

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
@@ -72,7 +72,7 @@ internal abstract class BillingClientUseCase<T>(
             }
 
             BillingResponse.ServiceUnavailable -> {
-                backoffOrErrorIfUseInSession(onError, billingResult)
+                backoffOrErrorIfManyRetries(onError, billingResult)
             }
 
             BillingResponse.NetworkError,
@@ -125,19 +125,14 @@ internal abstract class BillingClientUseCase<T>(
         }
     }
 
-    private fun backoffOrErrorIfUseInSession(
+    private fun backoffOrErrorIfManyRetries(
         onError: (BillingResult) -> Unit,
         billingResult: BillingResult,
     ) {
-        if (useCaseParams.appInBackground) {
-            log(LogIntent.GOOGLE_WARNING, BillingStrings.BILLING_SERVICE_UNAVAILABLE_BACKGROUND)
-            if (retryBackoffMilliseconds < RETRY_TIMER_MAX_TIME_MILLISECONDS) {
-                retryWithBackoff()
-            } else {
-                onError(billingResult)
-            }
+        log(LogIntent.GOOGLE_WARNING, BillingStrings.BILLING_SERVICE_UNAVAILABLE.format(useCaseParams.appInBackground))
+        if (retryBackoffMilliseconds < RETRY_TIMER_MAX_TIME_MILLISECONDS) {
+            retryWithBackoff()
         } else {
-            log(LogIntent.GOOGLE_ERROR, BillingStrings.BILLING_SERVICE_UNAVAILABLE_FOREGROUND)
             onError(billingResult)
         }
     }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
@@ -7,20 +7,23 @@ import com.revenuecat.purchases.PurchasesErrorCallback
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.log
+import com.revenuecat.purchases.common.min
 import com.revenuecat.purchases.google.BillingResponse
 import com.revenuecat.purchases.google.billingResponseToPurchasesError
 import com.revenuecat.purchases.google.toHumanReadableDescription
 import com.revenuecat.purchases.strings.BillingStrings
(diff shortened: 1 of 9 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
```diff
@@ -20,6 +20,7 @@ internal typealias ExecuteRequestOnUIThreadFunction = (delayInMillis: Long, onEr
 private const val MAX_RETRIES_DEFAULT = 3
 private const val RETRY_TIMER_START_MILLISECONDS = 878L // So it gets close to 15 minutes in last retry
 internal const val RETRY_TIMER_MAX_TIME_MILLISECONDS = 1000L * 60L * 15L // 15 minutes
+internal const val RETRY_TIMER_MAX_TIME_MILLISECONDS_FOREGROUND = 1000L * 4L // 4 seconds
```
tonidero, line 23: This gives for 3 retries and up to ~6s delay, which I think is more than enough. We should fail if it's failing more than that I think.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
```diff
@@ -20,6 +20,7 @@ internal typealias ExecuteRequestOnUIThreadFunction = (delayInMillis: Long, onEr
 private const val MAX_RETRIES_DEFAULT = 3
 private const val RETRY_TIMER_START_MILLISECONDS = 878L // So it gets close to 15 minutes in last retry
 internal const val RETRY_TIMER_MAX_TIME_MILLISECONDS = 1000L * 60L * 15L // 15 minutes
```
JayShortway, line 22: Should we also name this `RETRY_TIMER_MAX_TIME_MILLISECONDS_BACKGROUND` for consistency, or is this not always used for _just_ the background?

JayShortway (commented):
Looks good with my limited context. Just had a minor suggestion. I'll let someone with more context on the retry logic approve!

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
tonidero: Yeah I originally thought to rename that, but it's also used in some other retries which are background/foreground independent, like network errors...

But I can try to put a better name so it's not confusing 👍. I'm thinking `RETRY_TIMER_MAX_TIME_SERVICE_UNAVAILABLE_FOREGROUND` and change the type to `Duration` to avoid having the `milliseconds` suffix (still a bit of a mouthful, but this is currently only used for that specific error)

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
```diff
@@ -125,19 +126,19 @@ internal abstract class BillingClientUseCase<T>(
...
         billingResult: BillingResult,
     ) {
-        if (useCaseParams.appInBackground) {
-            log(LogIntent.GOOGLE_WARNING, BillingStrings.BILLING_SERVICE_UNAVAILABLE_BACKGROUND)
-            if (retryBackoffMilliseconds < RETRY_TIMER_MAX_TIME_MILLISECONDS) {
-                retryWithBackoff()
-            } else {
-                onError(billingResult)
-            }
+        log(LogIntent.GOOGLE_WARNING, BillingStrings.BILLING_SERVICE_UNAVAILABLE.format(useCaseParams.appInBackground))
```
vegaro, line 133: nitpick I know this was here before, but I think this log should go in line 76. This function could be call in cases different to `billing unavailable`

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
tonidero: Hmm I think this function is only called when we receive `ServiceUnavailable` from google though? (I might be missing something)

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
tonidero: Tried to improve it here: https://github.com/RevenueCat/purchases-android/pull/1745/commits/aa29f7eac7c8f98d0683821ea3be27d297ed7d0e

JayShortway: Alright, yea good improvement! Instead of `_MILLISECONDS` we could also use `_MILLIS` but that has to be a convention we'd choose to follow.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/BillingClientUseCase.kt
vegaro: Yes, it's only called from there, but this log is the only thing that makes it not reusable from another case, so what I meant is that if someone calls the function _in the future_ and doesn't look at the implementation, this log will be wrong.

You can leave it as is. I was triggered by the fact that the name of the function is `backoffOrErrorIfManyRetries` and not `backoffOrErrorIfManyRetriesWhenBillingUnavailable`

tonidero: Oh right, I just renamed to try to make it clearer. Thanks!

PR #1749, opened 2024-06-11
Remove unneeded Android Lollipop version checks

tonidero:
### Description
With version 8.0.0 and BC7, we bumped the min sdk to 21 (since it was required by BC7). This removes some checks we had for versions lower than that which shouldn't happen anymore.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
@@ -22,57 +22,6 @@ internal fun Context.getLocale(): Locale? =
         resources.configuration.locale
     }
 
-internal fun Locale.toBCP47(): String {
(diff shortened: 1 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
```diff
@@ -22,57 +22,6 @@ internal fun Context.getLocale(): Locale? =
...
-    val bcp47Tag = StringBuilder(language)
-    if (region.isNotEmpty()) {
-        bcp47Tag.append(separator).append(region)
-    }
-    if (variant.isNotEmpty()) {
-        bcp47Tag.append(separator).append(variant)
-    }
-
-    return bcp47Tag.toString()
-}
```
vegaro, line 74: yayyyy

vegaro (approved):

PR #1750, opened 2024-06-11
Paywalls: Add bottom padding when there are no buttons in footer

tonidero:
### Description
See https://github.com/RevenueCat/purchases-flutter/issues/1078

There is no bottom margin if there is no button present in the footer. This adds a small padding there.

| Before  |  After  |
|---|---|
| image | image |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
@@ -71,7 +71,7 @@ internal fun Footer(
     }
 }
 
-@Suppress("LongParameterList", "LongMethod")
+@Suppress("LongParameterList", "LongMethod", "CyclomaticComplexMethod")
 @Composable
 private fun Footer(
     mode: PaywallMode,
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -83,18 +83,31 @@ private fun Footer(
...
+                end = UIConstant.defaultHorizontalPadding,
+                bottom = bottomMargin,
+            ),
         horizontalArrangement = Arrangement.Center,
         verticalAlignment = Alignment.CenterVertically,
     ) {
         val color = colors.text1
 
-        if (mode == PaywallMode.FOOTER_CONDENSED && allPlansTapped != null) {
+        if (shouldDisplayAllPlansButton && allPlansTapped != null) {
```
tonidero, line 110: This is a bit stupid but need to add the `allPlansTapped != null` check to make sure the compiler picks up it's not null within the if. We don't have Kotlin 2 😢

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt
```diff
@@ -71,7 +71,7 @@ internal fun Footer(
     }
 }
 
-@Suppress("LongParameterList", "LongMethod")
+@Suppress("LongParameterList", "LongMethod", "CyclomaticComplexMethod")
```
tonidero, line 74: We can probably refactor this composable.... But for now, I think it's ok.

vegaro (approved):

PR #1754, opened 2024-06-18
Update test lane to require active entitlements

tonidero:
### Description
Now we have an active token in production that won't expire, so we can require an active entitlement for integration tests. We just need to update some ENVs in CI. This is just a cleanup


aboedo (approved):

joshdholtz (approved):
:shipit:

PR #1756, opened 2024-06-19
Support ISO8601-2 period durations

### LINKED ISSUE #1753: Package billingPeriod "value" incorrect: seems to be iso8601-1 instead of iso8601-2
From https://community.revenuecat.com/sdks-51/package-billingperiod-value-incorrect-seems-to-be-iso8601-1-instead-of-iso8601-2-4655

I seem to be getting an incorrect value for my free trial billingPeriod on Android.

I thought it would be good to ask here whether I’m understanding this correctly and/or whether to open an issue on GitrHub.

I’ve set the trial period to 90 days in the Play Console, but in the returned package I’m getting 86 days.
Looking at the iso8601 value, 84 days corresponds to 12 weeks so it looks like it’s ignoring the 6D.
I’ve just learned that iso8601-1 doesn’t allow any other values if W(eeks) is specified but iso8601-2 does. 
It seems like, although P12W6D is correct (84 days + 6 days = 90 days) the ‘value’ field is showing an iso8601-1 interpretation (84 days only) rather than iso8601-2.

 
image

---

I created a simple test in our test suite and it looks like we are parsing `P12W6D` as 12 weeks (omitting the days). I am not sure how it's getting converted to 86 days in the screenshot since the `Period.create("P12W6D")` function is creating a 12 week Period, but in any case it should be creating a 90 days Period

tonidero:
### Description
This will fix #1753

In cases where the period we get from the store is in the format ISO8601-2, that is, has multiple units, our current parsing code wouldn't parse it correctly. I changed to use the smaller unit sent from the store and convert it to that unit.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/models/PeriodTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/models/PeriodTest.kt
@@ -53,6 +53,22 @@ class PeriodTest {
     }
 
     @Test
+    fun `create period creates expected period class for multiple different units`() {
+        val period = Period.create("P12W6D")
+        assertThat(period.value).isEqualTo(90)
+        assertThat(period.unit).isEqualTo(Period.Unit.DAY)
+        assertThat(period.iso8601).isEqualTo("P12W6D")
+    }
+
+    @Test
+    fun `create period creates expected period class for multiple different units with min unit months`() {
+        val period = Period.create("P2Y6M")
+        assertThat(period.value).isEqualTo(30)
+        assertThat(period.unit).isEqualTo(Period.Unit.MONTH)
+        assertThat(period.iso8601).isEqualTo("P2Y6M")
+    }
+
+    @Test
     fun `valueInMonths is correct for days`() {
         val period = Period.create("P3D")
         assertThat(period.valueInMonths).isCloseTo(0.1, MAX_OFFSET)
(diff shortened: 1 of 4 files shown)


purchases/src/test/java/com/revenuecat/purchases/models/PeriodTest.kt
```diff
@@ -52,6 +52,22 @@ class PeriodTest {
         assertThat(period.iso8601).isEqualTo("P3M")
     }
 
+    @Test
+    fun `create period creates expected period class for multiple different units`() {
+        val period = Period.create("P12W6D")
```
JayShortway, line 57: Nice! Does this mean we support both ISO 8601-1 and ISO 8601-2 now? If so, maybe we can specify that in the documentation of the `Period.create()` function? (Is that meant to be used by developers?)

JayShortway:
I saw this comment above `String.toPeriod()`:
>// Would use Duration.parse but only available API 26 and up

We can use that (and [other `java.time` APIs](https://developer.android.com/studio/write/java8-support-table) such as `java.time.Period.parse()`) if we [set up desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring). Not something for this PR of course, but maybe something we can look into.

JayShortway (approved):
Good stuff! Added a minor documentation suggestion.

purchases/src/test/java/com/revenuecat/purchases/models/PeriodTest.kt
tonidero: Yup! It should support both. And yeah, currently that doesn't have documentation, I think the only reason to keep it public is to help with devs creating instances for tests but shouldn't really be used in a production environment... So I will add some notes about that. Good catch!

vegaro (approved):
Thanks for taking care of this! I still don't understand how the dev was getting 84 days instead of 12 weeks in the JSON 🤔

tonidero:
> I still don't understand how the dev was getting 84 days instead of 12 weeks in the JSON 🤔

Yup that's pretty strange and wasn't able to reproduce... But I wanted to get the underlying issue for now.

PR #1757, opened 2024-06-19
Makes `ITEM_NOT_OWNED` message more clear when restoring and acknowledging/consuming

vegaro:
Cleans up logs when acknowledging or consuming and getting an `ITEM_NOT_OWNED`. Also removed some duplication of the logs

-------

Before:
CleanShot 2024-06-19 at 12 30 21@2x

After:
CleanShot 2024-06-19 at 12 29 26@2x

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/AcknowledgePurchaseUseCase.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/AcknowledgePurchaseUseCase.kt
@@ -44,19 +44,20 @@ internal class AcknowledgePurchaseUseCase(
                 processResult(
                     billingResult,
                     useCaseParams.purchaseToken,
-                    onError = {
+                    onError = { errorBillingResult ->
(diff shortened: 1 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/AcknowledgePurchaseUseCase.kt
```diff
@@ -39,6 +44,21 @@ internal class AcknowledgePurchaseUseCase(
                 processResult(
                     billingResult,
                     useCaseParams.purchaseToken,
+                    onError = {
+                        val underlyingErrorMessage: String
+                        if (it.responseCode == BillingClient.BillingResponseCode.ITEM_NOT_OWNED &&
```
tonidero, line 49: Nitpick but maybe add a name to the type in the lambda? Like `billingResult` (Same in the `ConsumePurchaseUseCase`)

tonidero (approved):
Nice! This was very annoying. Thanks for fixing!

PR #1762, opened 2024-06-24
Adds .aiexclude

JayShortway:
Adds `.aiexclude` to exclude files with sensitive info from being accessed by Gemini in Android Studio. More info can be read [here](https://developer.android.com/studio/preview/gemini/aiexclude).

### DIFF
--- a/.aiexclude
+++ b/.aiexclude
@@ -1,2 +1,3 @@
 fastlane/.env
 local.properties
+Constants.kt
--- a/.aiexclude
+++ b/.aiexclude
@@ -1,0 +1,2 @@
+fastlane/.env
+local.properties


.aiexclude
```diff
@@ -0,0 +1,2 @@
+fastlane/.env
+local.properties
```
tonidero, line 2: I wonder if we should also include all `Constants.kt` files since we usually have API keys there... NABD if they are leaked but might be worth adding just in case. The most important info lives on these files yeah.

tonidero (approved):
Nice! Thanks for adding this! Just a suggestion but looks good!

.aiexclude
JayShortway: Good point! I added `Constants.kt` so any files with that name, now or in the future, will be excluded. I figured that's acceptable, but let me know if you rather have all current `Constants.kt` files specified by path.

tonidero: No, I think this is good 👍 Thanks!

PR #1751, opened 2024-06-12
Add `DEFERRED` replacement mode

tonidero:
### Description
Deferred replacement mode was removed in V7 of the SDK. This brings it back.

PR #1766, opened 2024-06-27
Use correct product context with `DEFERRED` purchases

tonidero:
### Description
Another followup to #1751 and #1764 

When making `DEFERRED` purchases, Google returns the result using the old product id. We need to cache all the context information using the old product id so we can attribute it correctly.


vegaro (approved):
Makes sense!

PR #1768, opened 2024-07-01
Use trimmed API key

tonidero:
### Description
I noticed that trusted entitlements may fail if passing an API key with extra whitespaces. This is because it's part of the data we are verifying and the backend uses the trimmed version so there is a mismatch. This trims the API key we use in the SDK, so it matches what the backend uses in those cases. Any real API key shouldn't contain any whitespaces so I think it should be fine but lmk what you think

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
@@ -14,9 +14,6 @@ open class PurchasesConfiguration(builder: Builder) {
     val apiKey: String
     val appUserID: String?
 
-    internal val trimmedApiKey: String
-        get() = apiKey.trim()
-
     @Deprecated(
         "observerMode is being deprecated in favor of purchasesAreCompletedBy.",
         ReplaceWith(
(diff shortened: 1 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -14,6 +14,9 @@ open class PurchasesConfiguration(builder: Builder) {
     val apiKey: String
     val appUserID: String?
 
+    internal val trimmedApiKey: String
+        get() = apiKey.trim()
```
vegaro, line 18: can we just trim the apiKey when setting it in the Builder? That way we don't need to store another API key and we can guarantee `apiKey` will always be trimmed

tonidero: Yeah, I thought about doing this to avoid modifying the input in the `PurchasesConfiguration`.... But I think it should be ok to do that and simplifies things. Will do that, thanks!

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/1768/commits/6db9240af99f1bd0ea4fa083ee911892d70ea2b5

vegaro (approved):
Thanks for making that change!!

Issue #1629, opened 2024-02-29
ANRs when calling Purchases.configure() during Application.onCreate()


snapwoodapps:
I'm seeing in the field several dozen ANRs happening inside Purchases.configure.  I am calling configure during Application.onCreate(), which matches your examples. 

For reference, I have 16 products defined for Google Play where this is failing and 13 products defined for Amazon.  These are part of 8 offerings, most with 3 packages.  I don't know if the quantity here is impacting the configure time, but it doesn't seem like much compared to other apps.

Here are a sampling of my ANRs.  Each ANR exception is a bit different, likely due to how far configure got in reconstituting its data before the system killed it with the ANR hammer.  The second one is especially concerning to me since it shows json parsing happening on the main thread, which implies I/O and that is generally a bad idea on the main thread.

1.
>at com.revenuecat.purchases.PurchasesFactory.createPurchases (PurchasesFactory.kt:166)
  at com.revenuecat.purchases.PurchasesFactory.createPurchases$default (PurchasesFactory.kt:58)
  at com.revenuecat.purchases.Purchases$Companion.configure (Purchases.kt:852)

2.
>at kotlinx.serialization.modules.SerializersModuleKt.<clinit> (SerializersModule.kt:80)
  at kotlinx.serialization.modules.SerializersModuleKt.getEmptySerializersModule (SerializersModule.kt)
  at kotlinx.serialization.modules.SerializersModuleBuildersKt.EmptySerializersModule (SerializersModuleBuilders.kt:40)
  at kotlinx.serialization.json.Json$Default.<init> (Json.kt:71)
  at kotlinx.serialization.json.Json$Default.<init> (Json.kt)
  at kotlinx.serialization.json.Json.<clinit> (Json.kt)
  at kotlinx.serialization.json.JsonKt.Json$default (Json.kt:197)
  at com.revenuecat.purchases.common.OfferingParser.<clinit> (OfferingParser.kt:21)
  at com.revenuecat.purchases.OfferingParserFactory.createOfferingParser (OfferingParserFactory.kt:12)
  at com.revenuecat.purchases.PurchasesFactory.createPurchases (PurchasesFactory.kt:218)
  at com.revenuecat.purchases.PurchasesFactory.createPurchases$default (PurchasesFactory.kt:58)
  at com.revenuecat.purchases.Purchases$Companion.configure (Purchases.kt:852)

3.
[...]

PR #1772, opened 2024-07-02
Attempt to fix ANRs by moving some tasks during configure to background

vegaro:
## Description
This PR addresses a significant number of ANRs (Application Not Responding errors) reported, such as in issue #1629.

## Problem
The method `IdentityManager.configure` is currently running on the main thread and accessing `SharedPreferences`. Accessing `SharedPreferences` can be costly, especially if the XML file storing the preferences is large. This cost is particularly high during the initial accesses, as the XML file needs to be read and parsed, which can block the main thread and lead to ANRs.

## Solution
**Custom Dispatcher**: This PR introduces a custom dispatcher to the `IdentityManager`, allowing configure to run on a background thread. By moving this operation off the main thread, we avoid blocking it with the potentially expensive SharedPreferences access.
**Main Thread Optimization**: The remaining setup in `PurchasesOrchestrator` has been deferred to execute after configure completes, ensuring the main thread remains unblocked and improving overall app responsiveness.

## Notes
While this fix should mitigate the reported ANRs, it's possible there are other underlying causes contributing to these issues. This PR serves as a good initial step towards improving performance.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -95,10 +95,6 @@ internal class PurchasesFactory(
                 createEventsExecutor(),
                 runningIntegrationTests = runningIntegrationTests,
             )
-            val identityManagerDispatcher = Dispatcher(
-                createDefaultExecutor(),
-                runningIntegrationTests = runningIntegrationTests,
-            )
 
             var diagnosticsFileHelper: DiagnosticsFileHelper? = null
             var diagnosticsHelper: DiagnosticsHelper? = null
@@ -183,7 +179,7 @@ internal class PurchasesFactory(
                 offeringsCache,
                 backend,
                 offlineEntitlementsManager,
-                identityManagerDispatcher,
+                dispatcher,
             )
 
             val customerInfoUpdateHandler = CustomerInfoUpdateHandler(
@@ -290,6 +286,7 @@ internal class PurchasesFactory(
                 createPaywallEventsManager(application, identityManager, eventsDispatcher, backend),
                 paywallPresentedCache,
                 purchasesStateProvider,
+                dispatcher,
             )
 
             return Purchases(purchasesOrchestrator)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -95,6 +95,10 @@ internal class PurchasesFactory(
                 createEventsExecutor(),
                 runningIntegrationTests = runningIntegrationTests,
             )
+            val identityManagerDispatcher = Dispatcher(
+                createDefaultExecutor(),
+                runningIntegrationTests = runningIntegrationTests,
+            )
 
             var diagnosticsFileHelper: DiagnosticsFileHelper? = null
             var diagnosticsHelper: DiagnosticsHelper? = null
@@ -179,6 +183,7 @@ internal class PurchasesFactory(
                 offeringsCache,
                 backend,
                 offlineEntitlementsManager,
+                identityManagerDispatcher,
             )
 
             val customerInfoUpdateHandler = CustomerInfoUpdateHandler(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -286,7 +286,7 @@ internal class PurchasesFactory(
                 createPaywallEventsManager(application, identityManager, eventsDispatcher, backend),
                 paywallPresentedCache,
                 purchasesStateProvider,
-                dispatcher,
+                dispatcher = dispatcher,
             )
 
             return Purchases(purchasesOrchestrator)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -73,7 +73,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.time.Duration.Companion.seconds
 
 @Suppress("LongParameterList", "LargeClass", "TooManyFunctions")
-internal class PurchasesOrchestrator constructor(
+internal class PurchasesOrchestrator(
     private val application: Application,
     backingFieldAppUserID: String?,
     private val backend: Backend,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -780,7 +780,6 @@ internal class PurchasesOrchestrator(
     //endregion
 
     // region Private Methods
-    @Synchronized
     private fun enqueue(command: () -> Unit) {
         dispatcher.enqueue({ command() }, Delay.NONE)
     }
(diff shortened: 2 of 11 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -20,14 +22,15 @@ import com.revenuecat.purchases.subscriberattributes.caching.SubscriberAttribute
...
     private val deviceCache: DeviceCache,
     private val subscriberAttributesCache: SubscriberAttributesCache,
     private val subscriberAttributesManager: SubscriberAttributesManager,
     private val offeringsCache: OfferingsCache,
     private val backend: Backend,
     private val offlineEntitlementsManager: OfflineEntitlementsManager,
+    private val dispatcher: Dispatcher,
 ) {
 
     val currentAppUserID: String
```
tonidero, line 36: This could potentially fail now since setting the user id happens asynchronously now right?

vegaro: hmm right... actually... any call in this class right? for example `currentUserIsAnonymous`. 

I think we need to make all async and make sure they are all enqueued in the same dispatcher, right?

tonidero: Yup pretty much, which seems it would be a pretty big undertaking...

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -194,22 +197,24 @@ internal class PurchasesOrchestrator constructor(
...
-            )
-        }
-        offeringsManager.onAppForeground(identityManager.currentAppUserID)
-        postPendingTransactionsHelper.syncPendingPurchaseQueue(allowSharingPlayStoreAccount)
-        synchronizeSubscriberAttributesIfNeeded()
-        offlineEntitlementsManager.updateProductEntitlementMappingCacheIfStale()
-        flushPaywallEvents()
-        if (firstTimeInForeground && isAndroidNOrNewer()) {
-            diagnosticsSynchronizer?.syncDiagnosticsFileIfNeeded()
+        enqueue {
```
vegaro, line 200: I don't think there should be any issue with enqueuing all this right? It is all pretty much updating caches, which is already changing into another thread to make the API call. This way we guarantee the accesses to the deviceCache are in the background

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -49,10 +55,16 @@ internal class IdentityManager(
             ?: deviceCache.getLegacyCachedAppUserID()
             ?: generateRandomID()
         log(LogIntent.USER, IdentityStrings.IDENTIFYING_APP_USER_ID.format(appUserIDToUse))
-        deviceCache.cacheAppUserID(appUserIDToUse)
-        subscriberAttributesCache.cleanUpSubscriberAttributeCache(appUserIDToUse)
-        deviceCache.cleanupOldAttributionData()
-        invalidateCustomerInfoAndETagCacheIfNeeded(appUserIDToUse)
+
+        val cacheEditor = deviceCache.startEditing()
```
vegaro, line 59: This change (chaining the editions) will probably have marginal effects, but I don't think it hurts

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -135,11 +147,17 @@ internal class IdentityManager(
         }
     }
 
-    private fun invalidateCustomerInfoAndETagCacheIfNeeded(appUserID: String) {
+    private fun invalidateCustomerInfoAndETagCacheIfNeeded(
+        appUserID: String,
+        cacheEditor: SharedPreferences.Editor,
+    ) {
+        if (backend.verificationMode == SignatureVerificationMode.Disabled) {
```
tonidero, line 154: I guess this is to avoid the `getCachedCustomerInfo` below from happening when it's disabled right? Makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -49,10 +55,16 @@ internal class IdentityManager(
...
-        invalidateCustomerInfoAndETagCacheIfNeeded(appUserIDToUse)
+
+        val cacheEditor = deviceCache.startEditing()
+        deviceCache.cacheAppUserID(appUserIDToUse, cacheEditor)
+        subscriberAttributesCache.cleanUpSubscriberAttributeCache(appUserIDToUse, cacheEditor)
+        invalidateCustomerInfoAndETagCacheIfNeeded(appUserIDToUse, cacheEditor)
+        cacheEditor.apply()
+
+        enqueue {
+            deviceCache.cleanupOldAttributionData()
```
tonidero, line 66: Hmm doing this on a thread could be dangerous I think? As in, this will remove all attribution, so there is a small chance of actual new data making it in before we delete it, then deleting it

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -775,6 +780,10 @@ internal class PurchasesOrchestrator constructor(
     //endregion
 
     // region Private Methods
+    @Synchronized
```
tonidero, line 783: NABD but I think this doesn't need to be synchronized?

tonidero (commented):
Good ideas! I just left some concerns but I think this should improve things a lot.

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
vegaro: exactly! I noticed we were getting the customer info all the time! I will add a comment here to make sure it's clear

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
vegaro: note this is attribution data and not subscriber attributes. Not sure if you were at RevenueCat when we replaced  attribution data with subscriber attributes. Attribution data is deprecated and we don't actually ever post anything in the key. I thought we were still using this for the device identifiers, but we are using subscriber attributes for that.

This function is just cleaning up the caches in case someone is upgrading from an old version of the SDK that used attribution data

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
vegaro: Right! Not sure why I added this lol

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
tonidero: Ohhh that makes total sense then! Thanks for explaining!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -179,6 +179,7 @@ internal class PurchasesFactory(
                 offeringsCache,
                 backend,
                 offlineEntitlementsManager,
+                dispatcher,
```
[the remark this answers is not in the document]
vegaro: I agree we should rename it. I had originally created a new backgroundTaskDispatcher before I noticed we could reuse this one. Is that naming good?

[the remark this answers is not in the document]
tonidero: Yeah, I think that could be enough as a "generic" dispatcher 👍

M4x1n33 (approved):

PR #1776, opened 2024-07-03
[WIP] paywall custom purchase

jamesrb1:
Callbacks are added to the `PaywallViewModel` in a new model struct `MyAppPurchaseLogic`, via its`PaywallOptions`.

When the VM's purchase/restore methods are called, the `MyAppPurchaseLogic` struct (which holds the purchase/restore handlers) is passed down to the purchase orchestrator that calls the customer handlers if `finishTransactions` is false, skipping its built-in logic.


jamesrb1:
Created a new branch from earlier in this series of commits based off advice received here https://revenuecat.slack.com/archives/C03HPN9UE80/p1720051171704469

PR #1777, opened 2024-07-04
Paywalls can use custom in-app purchase/restore code

jamesrb1:
This allows someone who's adopting RevenueCat paywalls to use their own app purchase/restore logic rather than RevenueCat's.

This is done by implementing  `MyAppPurchaseLogic`, which has two functions that implement the custom purchase and restore functions.

```kotlin
interface MyAppPurchaseLogic {

    suspend fun performPurchase(activity: Activity, rcPackage: Package): MyAppPurchaseResult

    suspend fun performRestore(customerInfo: CustomerInfo): MyAppRestoreResult

}
```
Users can also use completion handlers by implementing this abstract class instead:

```kotlin
abstract class MyAppPurchaseLogicCompletion : MyAppPurchaseLogic {

    abstract fun performPurchaseWithCompletion(activity: Activity, rcPackage: Package, completion: (MyAppPurchaseResult) -> Unit)

    abstract fun performRestoreWithCompletion(completion: (MyAppRestoreResult) -> Unit)

    final override suspend fun performPurchase(activity: Activity, rcPackage: Package): MyAppPurchaseResult =
        suspendCoroutine { continuation ->
            performPurchaseWithCompletion(activity, rcPackage) { result ->
                continuation.resume(result)
            }
        }

    final override suspend fun performRestore(customerInfo: CustomerInfo): MyAppRestoreResult =
        suspendCoroutine { continuation ->
            performRestoreWithCompletion { result ->
                continuation.resume(result)
            }
        }
}
```

An nullable instance of this interface is now part of `PaywallOptions` and `PaywallDialogOptions`, and settable via the builders. These classes are used when creating a Paywall. 

Builder function:
```kotlin
fun setMyAppPurchaseLogic(myAppPurchaseLogic: MyAppPurchaseLogic?) = apply {
    this.myAppPurchaseLogic = myAppPurchaseLogic
}
```
When a `MyAppPurchaseLogic` is provided, the paywall viewmodel will use this logic instead of RevenueCat's _if_ `Purchases` is configured such that `PurchasesAreCompletedBy.MY_APP`. This happens in `PaywallViewModelImpl`, methods `awaitRestorePurchases` and `awaitRestorePurchases`.

If a `PaywallViewModel` is constructed _without_ being provided with an instance of MyAppPurchaseLogic when PurchasesAreCompletedBy.MY_APP, it will throw:

```kotlin
// called by constructor
private fun validateState() {
    if (purchases.purchasesAreCompletedBy == PurchasesAreCompletedBy.MY_APP && options.myAppPurchaseLogic == null) {
        error(
            "myAppPurchaseLogic is null, but is required when purchases.purchasesAreCompletedBy is .MY_APP."
        )
    }
}
```

PaywallViewModel` refactored a bit to make awaitable versions of purchase and restore functions. This has made the PaywallViewModel.kt diff a bit messy, but the change here was breaking it up so that the awaitable versions wouldn't create a co-routine context, and the code that checks what `purchasesAreCompletedBy` is, and executing either the customer code or (as previous) the RC code to do the purchase/restore.

To make the paywalls tester app run the custom code blocks, make these temporary additional changes:

image


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -45,6 +49,17 @@ fun PaywallsScreen(
     LazyColumn {
         items(SamplePaywalls.SampleTemplate.values()) { template ->
             val offering = samplePaywallsLoader.offeringForTemplate(template)
+            val myAppPurchaseLogic = MyAppPurchaseLogic(
+                performPurchase = { _, _ ->
+                    println("Hello from performPurchase!")
```
tonidero, line 54: Should we remove these? If we want to keep logs, we should use `Log.d` instead, so it's better formatted with logcat.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Code for handling in-app purchases and restorations directly by the application rather than by RevenueCat.
+ * These methods are called by a RevenueCat Paywall only when `Purchases.purchasesAreCompletedBy` is `MY_APP`.
+ *
+ * @property performPurchase Performs an in-app purchase.
+ * @property performRestore Perform a purchase restore for the given customer.
+ */
+class MyAppPurchaseLogic(
+    val performPurchase: suspend ((Activity, Package) -> MyAppPurchaseResult),
+    val performRestore: suspend ((CustomerInfo) -> MyAppRestoreResult),
```
tonidero, line 25: I was wondering, until now, we've been using lambda functions instead of coroutines for things like this (and then wrap those into async APIs in some cases. Is there a reason we wanted to go coroutine-first here? 

FWIW, we've discussed a bit about going with coroutines moving forward, but we should make sure that we keep compatibility for people not using coroutines and we keep a similar API.

IMO, for now, we could have this be an interface with these 2 as functions, that need to be implemented by the developer. This is also the approach we follow with the `PaywallListener`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+ * Represents a purchase made in the application.
+ *
+ * @property purchase The actual purchase object from the store.
+ * @property productType The type of product being purchased. Defaults to [ProductType.UNKNOWN].
+ * @property presentedOfferingContext Context of the offering presented to the user, if applicable.
+ * @property subscriptionOptionId The subscription option ID from Google, if applicable.
+ * @property replacementMode The replacement mode for Google subscriptions, if applicable.
+ */
+class MyAppPurchase(
+    val purchase: Purchase,
```
tonidero, line 81: This is only compatible with Google, not Amazon. However, I believe, we don't need this object, since our SDK should automatically pick purchases up and send them to our servers I believe. We do need to know whether the purchase was successful or not to display a success/error state in the paywall though. Lmk if I'm missing something though!

tonidero (commented):
I'm suggesting a couple changes:
- Make the `MyAppPurchaseLogic` an interface instead of using coroutines.
- I believe we can remove the `MyAppPurchase` class entirely (do lmk if I'm missing something)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Represents a purchase made in the application.
+ *
+ * @property purchase The actual purchase object from the store.
+ * @property productType The type of product being purchased. Defaults to [ProductType.UNKNOWN].
+ * @property presentedOfferingContext Context of the offering presented to the user, if applicable.
+ * @property subscriptionOptionId The subscription option ID from Google, if applicable.
+ * @property replacementMode The replacement mode for Google subscriptions, if applicable.
```
JayShortway, line 78: I think some of these can be deduced internally by the SDK, instead of having the developer provide them back to us. 
  
For instance:
- `productType` ➡️ `package.product.type`
- `presentedOfferingContext` ➡️ `package.presentedOfferingContext`
- maybe more? (cc @tonidero)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -200,32 +244,92 @@ internal class PaywallViewModelImpl(
...
+
+        finishAction()
+    }
+
+    private fun validateState() {
+        if (purchases.purchasesAreCompletedBy == PurchasesAreCompletedBy.MY_APP && options.myAppPurchaseLogic == null) {
+            error(
+                "myAppPurchaseLogic is null, but is required when purchases.purchasesAreCompletedBy is .MY_APP.",
+            )
+        }
```
JayShortway, line 331: The built-in `check` function has the same behavior, throwing `IllegalStateException`. Alternatively you can use `require` instead of `check`, which throws an `IllegalArgumentException`. Both can be considered correct in this case imo.
```suggestion
        check(
            purchases.purchasesAreCompletedBy != PurchasesAreCompletedBy.MY_APP || options.myAppPurchaseLogic != null
        ) { "myAppPurchaseLogic is null, but is required when purchases.purchasesAreCompletedBy is MY_APP." }
```  
  
   
I would also be fine to inline this in the constructor, as that's really the only place we should do this check. Having it as a separate function gives the impression that we might want to reuse this and call this in other places too. So you'd get: 

```kotlin
init {
    check(...) {...}
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -145,47 +157,79 @@ internal class PaywallViewModelImpl(
...
-                        Logger.d("Dismissing paywall after restore since display condition has not been met")
-                        options.dismissRequest()
+            when (purchases.purchasesAreCompletedBy) {
+                PurchasesAreCompletedBy.MY_APP -> {
+                    checkNotNull(customRestoreHandler) {
+                        "myAppPurchaseLogic must not be null when purchases.purchasesAreCompletedBy " +
+                            "is PurchasesAreCompletedBy.MY_APP"
+                    }
+                    val customerInfo = purchases.awaitCustomerInfo()
+                    when (val result = customRestoreHandler.invoke(customerInfo)) {
```
JayShortway, line 188: `invoke` can be omitted: 
 ```suggestion
                     when (val result = customRestoreHandler(customerInfo)) {
```

File: ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt [hunk 7]
```diff
@@ -200,32 +244,92 @@ internal class PaywallViewModelImpl(
...
+
+            val customPurchaseHandler = myAppPurchaseLogic?.performPurchase
+
+            when (purchases.purchasesAreCompletedBy) {
+                PurchasesAreCompletedBy.MY_APP -> {
+                    checkNotNull(customPurchaseHandler) {
+                        "myAppPurchaseLogic must not be null when purchases.purchasesAreCompletedBy " +
+                            "is PurchasesAreCompletedBy.MY_APP"
+                    }
+                    when (val result = customPurchaseHandler.invoke(activity, packageToPurchase)) {
```
JayShortway, line 277: Same here, `invoke` can be omitted: 
```suggestion
                    when (val result = customPurchaseHandler(activity, packageToPurchase)) {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.json.JSONObject
+
+/**
+ * Code for handling in-app purchases and restorations directly by the application rather than by RevenueCat.
+ * These methods are called by a RevenueCat Paywall only when `Purchases.purchasesAreCompletedBy` is `MY_APP`.
+ *
+ * @property performPurchase Performs an in-app purchase.
+ * @property performRestore Perform a purchase restore for the given customer.
+ */
+class MyAppPurchaseLogic(
```
JayShortway, line 23: It would be great if this implemented an interface `PurchaseLogic`. That way, we could have our own internal implementation of it (e.g. `RevenueCatPurchaseLogic`) which contains the logic that exists now (before this PR). 

That would make especially the `PaywallViewModel` a bit more straightforward imo. `performPurchase()` and `performRestore()` would simply check which `PurchaseLogic` to forward the calls to. 

Just an idea that _seems_ feasible, but let me know if it isn't!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
```diff
@@ -92,13 +105,159 @@ class PaywallViewModelTest {
...
+        every { purchases.purchasesAreCompletedBy } returns PurchasesAreCompletedBy.MY_APP
+
+        try {
+            create(
+                activeSubscriptions = setOf(TestData.Packages.weekly.product.id)
+            )
+            fail("Expected IllegalStateException to be thrown")
+        } catch (e: IllegalStateException) {
+            // Success: caught the expected exception
+        }
```
JayShortway, line 258: I would either add `@Suppress("SwallowedException")` right above the `try` to satisfy Detekt, or if you want to go all-in on AssertJ, you can do: 
```suggestion
        assertThatExceptionOfType(IllegalStateException::class.java).isThrownBy {
            create(
                activeSubscriptions = setOf(TestData.Packages.weekly.product.id)
            )
        }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -54,8 +58,12 @@ internal interface PaywallViewModel {
      */
     fun purchaseSelectedPackage(activity: Activity?)
 
+    suspend fun awaitPurchaseSelectedPackage(activity: Activity?)
+
     fun restorePurchases()
 
+    suspend fun awaitRestorePurchases()
```
JayShortway, line 65: Are these (`await*()`) functions added only for tests? If so, I don't think we need that. See next comment 👇

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
```diff
@@ -92,13 +105,159 @@ class PaywallViewModelTest {
...
+        coVerify { myAppPurchaseLogic.performPurchase(any(), any()) }
+
+        verifyOrder {
+            listener.onPurchaseStarted(any())
+            listener.onPurchaseCompleted(customerInfo, any())
+        }
+
+        assertThat(model.actionInProgress.value).isFalse
+        assertThat(dismissInvoked).isTrue
+    }
```
JayShortway, line 193: We can modify this test, and tests like it, so we don't need the `await*()` call. 

1. Add an observable boolean using `MutableStateFlow`, initialized to `false`.
2. Construct a `MyAppPurchaseLogic` normally, without mocking. In `performPurchase`, set the observable boolean to `true`.
3. Call `purchaseSelectedPackage()` instead of `awaitPurchaseSelectedPackage()`.
4. Wait for the first emitted item from the observable boolean that is `true`. 

Altogether that makes it look something like this: 

```kotlin
fun `Calls custom purchase logic when purchasesAreCompletedBy == MY_APP`() = runTest {
    every { purchases.purchasesAreCompletedBy } returns PurchasesAreCompletedBy.MY_APP
    
    val purchase = stubGooglePurchase()
    val myAppPurchase = MyAppPurchase(purchase)
    
    val customPurchaseCalled = MutableStateFlow(false)
    val myAppPurchaseLogic = MyAppPurchaseLogic(
        performPurchase = { _, _ ->
            customPurchaseCalled.value = true
            MyAppPurchaseResult.Success(myAppPurchase)
        },
        performRestore = { error("Not expected to be called") }
    )
    
    val model = create(
        customPurchaseLogic = myAppPurchaseLogic,
        activeSubscriptions = setOf(TestData.Packages.weekly.product.id)
    )
    
    model.purchaseSelectedPackage(activity)
    customPurchaseCalled.first { it }
    verifyOrder {
        listener.onPurchaseStarted(any())
        listener.onPurchaseCompleted(customerInfo, any())
    }
    
    assertThat(model.actionInProgress.value).isFalse
    assertThat(dismissInvoked).isTrue
}
```

JayShortway (commented):
Good stuff! 🙌  Great job decoupling all this logic! 
  
I added quite an extensive review, but a lot of it is "FYI" and "tips & tricks". Let me know if you want less of that 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
JayShortway: Or we could fully remove the `MyAppPurchase` class as Toni suggests. If that's possible, that would make the API even simpler.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
JayShortway: Just double checking, with this: 
>Right, in this case I'm actually thinking that to make it simpler on developers, we want to get the response on a callback itself. 
  
are you referring to the `completion` callback in your suggested code block? 

I was actually going to suggest the option of making them blocking with return values (as I believe the BillingClient's `launchBillingFlow()` is blocking as well?): 
```diff
interface MyAppPurchaseLogic {
-    fun performPurchase(activity: Activity, rcPackage: Package, completion: (MyAppPurchaseResult) -> Unit)
+    fun performPurchase(activity: Activity, rcPackage: Package): MyAppPurchaseResult
-    fun performRestore(completion: (MyAppRestoreResult) -> Unit)
+    fun performRestore(): MyAppRestoreResult
}
``` 
  
If **not** doing that ☝️ is what you meant, I believe you 😄 You have more context around these APIs and how they're typically used.

tonidero: Actually `launchBillingFlow` is asynchronous. As in, it doesn't block the thread, but launches a new window/activity on top to display the purchase flow.

So yeah, we shouldn't make them blocking, and providing a completion block/asynchronous function would help in that sense.

Honestly, it would look much nicer with the coroutine approach, but I'm not sure of suddenly changing the API style for this... I could be convinced otherwise though.

JayShortway: >Actually `launchBillingFlow` is asynchronous.
  
You're right! 

>Honestly, it would look much nicer with the coroutine approach, but I'm not sure of suddenly changing the API style for this... I could be convinced otherwise though.

Well, one could argue that `PurchaseLogic` is a different beast from `PaywallListener`. The former implements long-running operations, while the latter is just notifying you of events that happened. Following that argument it is not super strange that they have different APIs. 

If we're doing `suspend` functions, how are we going to do that in `PaywallActivity`, which could be used from Java? I guess we could use a [`lifecycleScope`](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope) to call the `suspend` functions, and turn them into callbacks for Java. (`lifecycleScope` is part of `androidx.lifecycle:lifecycle-runtime-ktx`, which we are already (transitively) depending on.)

Not sure where I stand yet haha. I can agree to both I think. `suspend` functions are definitely an elegant solution.

jamesrb1: Thank you for this vigorous discussion! I will look for direction, whatever the conclusion is.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -45,6 +49,17 @@ fun PaywallsScreen(
     LazyColumn {
         items(SamplePaywalls.SampleTemplate.values()) { template ->
             val offering = samplePaywallsLoader.offeringForTemplate(template)
+            val myAppPurchaseLogic = MyAppPurchaseLogic(
```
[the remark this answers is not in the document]
jamesrb1: Done and done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+    val performPurchase: suspend ((Activity, Package) -> MyAppPurchaseResult),
+    val performRestore: suspend ((CustomerInfo) -> MyAppRestoreResult),
+)
+
+/**
+ * Represents the result of a purchase attempt made by custom app-based code (not RevenueCat).
+ */
+sealed class MyAppPurchaseResult {
```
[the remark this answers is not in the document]
jamesrb1: Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @property error Details of the error that occurred.
+     */
+    data class Error(val error: PurchasesError) : MyAppPurchaseResult()
+}
+
+/**
+ * Represents the result of a restore purchases attempt.
+ */
+sealed class MyAppRestoreResult {
```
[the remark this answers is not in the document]
jamesrb1: Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,118 @@
...
+    private val presentedOfferingContext: PresentedOfferingContext? = null,
+    private val subscriptionOptionId: String? = null,
+    private val replacementMode: GoogleReplacementMode? = null,
+) {
+    /**
+     * Converts the [MyAppPurchase] instance to a [StoreTransaction] object.
+     *
+     * @return A [StoreTransaction] representing the purchase.
+     */
+    fun toStoreTransaction(): StoreTransaction = StoreTransaction(
```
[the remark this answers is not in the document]
jamesrb1: Yes, and done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: I was aware of `check`/`require`, but I didn't find the boolean condition as easy to understand and immediately see as correct (or not), and `require` I thought threw the wrong exception type. So I'd leave this as-is, but will change it if you like!

I agree on moving it directly to the constructor, will do once I hear back on the above.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: Yes, they were added specifically for tests. If I'd been aware of that testing flow before I'd refactored I probably wouldn't have bothered. But now it's done, I think it makes for cleaner test code, and having await variants available doesn't seem like a bad thing to me generally. So at this point, I'd be inclined to leave it as-is, but I will revert if you'd prefer it as it was before. Let me know!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
jamesrb1: Interesting! `listener.onPurchaseCompleted(customerInfo, any())` is the last call to be made, after `performPurchase` has run - would it need to toggle a flag there instead?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
jamesrb1: Yes I think this is a good idea, but will wait until I have direction on whether to co-routines or not.

jamesrb1:
Thank you for the thorough reviews! And I'm happy to learn "FYI / tips & tricks" anytime, don't hold back!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
aboedo: > I didn't find the boolean condition as easy to understand and immediately see as correct (or not)

Could you clarify this? Is it because you'd be checking for the opposite condition?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
aboedo: if these methods are public and only for tests then it'd be great to find a way to do without them

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
aboedo: Yeah, going with ~observer mode~ Purchases Completed By My App ™️ doesn't mean we would miss them, so we might be fine without having to do the MyAppPurchase dance. 

Slightly more nuance with Amazon in that we still need you to directly report the purchases to us. But that is kind of independent of this, right? We'd still require you to call the Amazon-specific method? 

cc @vegaro

jamesrb1: The _only_ reason for this is to enable calling the PaywallListener purchaseCompleted callback (as discussed/ing in Slack, at meeting).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: I think these are internal. The class is `internal class PaywallViewModelImpl` and these functions are part of the interface `internal interface PaywallViewModel`. It also looks like I could take them out of the interface and then mark them as `internal` right in the class.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: I just found the if condition here matches better how I think:

"if purchases are completed by my app, and my app's purchase logic does not exist, that's an error":

`purchases.purchasesAreCompletedBy == PurchasesAreCompletedBy.MY_APP && options.myAppPurchaseLogic == null`

vs 

"check that purchases are not completed by my app, or that my app's purchase logic exists, otherwise that's an error":

`check(purchases.purchasesAreCompletedBy != PurchasesAreCompletedBy.MY_APP || options.myAppPurchaseLogic != null)`. 

`check` is more concise and meant to do in one step what I've done in two, and normally should be used, but given the conditions that are being checked here I had to think harder to convince myself of the correctness of the code when using `check`. But maybe others find the `check` code fine to read and/or prefer it because it's more concise, in which case I'm happy to change it!

(This reminds me a bit of [this exchange](https://github.com/RevenueCat/purchases-ios/pull/3823#discussion_r1575364771) when choosing to use `if` when `guard` might have been technically preferred but not quite as readable.)

JayShortway: Yea that's a good argument! The fewer brain cpu cycles we have to spend reading the code, the better. 

You could still do this ofc:
```kotlin
check(!(purchases.purchasesAreCompletedBy == PurchasesAreCompletedBy.MY_APP && options.myAppPurchaseLogic == null))
```
But I can agree to the `if` being clearer, because now we have a sneaky `!` in there that's easy to miss.

JayShortway: What do you think about my suggestion to inline it in the constructor? So in case of the `if`:

```kotlin
init {
    if (...) {
        error(...)
    }
}
```

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
JayShortway: Ah yes! In that case a minimal diff from the existing test would be something like: 

```diff
@Test
fun `Calls custom purchase logic when purchasesAreCompletedBy == MY_APP`() = runTest {
+   val purchaseCompleted = MutableStateFlow(false)
    every { purchases.purchasesAreCompletedBy } returns PurchasesAreCompletedBy.MY_APP
+   every { listener.onPurchaseCompleted(any(), any()) }.answers {
+        purchaseCompleted.value = true
+   }

    val myAppPurchaseLogic = mockk<MyAppPurchaseLogic>(relaxed = true)

    val purchase = stubGooglePurchase()
    val myAppPurchase = MyAppPurchase(purchase)

    coEvery { myAppPurchaseLogic.performPurchase(any(), any()) } returns MyAppPurchaseResult.Success(myAppPurchase)

    val model = create(
        customPurchaseLogic = myAppPurchaseLogic,
        activeSubscriptions = setOf(TestData.Packages.weekly.product.id)
    )

-    model.awaitPurchaseSelectedPackage(activity)
+    model.purchaseSelectedPackage(activity)
+    purchaseCompleted.first { it }

    coVerify { myAppPurchaseLogic.performPurchase(any(), any()) }

    verifyOrder {
        listener.onPurchaseStarted(any())
        listener.onPurchaseCompleted(customerInfo, any())
    }

    assertThat(model.actionInProgress.value).isFalse
    assertThat(dismissInvoked).isTrue
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
JayShortway: They are internal indeed because the class is internal, but they are "public" in the sense of this class' api. 

I'd really prefer to not add any code specifically for tests, regardless of visibility. A test should verify actual behavior. When adding code specifically for tests, there's a risk that actual behavior and tested behavior deviates. Maybe not now, but that risk grows after more time passes and more changes are made to this class. For instance, we need to remember to copy any change in `purchaseSelectedPackage()` to `awaitPurchaseSelectedPackage()`. That's very error prone imo. 

I don't think we'd need to revert entirely: 
1. The tests can be adjusted to not call the `await*()` variants, as suggested below.
2. `awaitPurchaseSelectedPackage()` can then be removed entirely from both the interface and the class.
3. `awaitRestorePurchases()` can then be removed from the interface, and be made private in the class.

Happy to discuss more of course!

JayShortway: Also, I see your point that `await*()` can be cleaner, I think there's an argument to be made for that. However, then I'd argue that the _fire-and-forget_ methods should be removed (`purchaseSelectedPackage()` and `restorePurchases()`). 
  
But this is a more risky change since we'd need to get a CoroutineScope in Compose ([`rememberCoroutineScope()`](https://developer.android.com/develop/ui/compose/side-effects#remembercoroutinescope)), which has a different life cycle from a `viewModelScope`. So we need to verify that the behavior is still the same. For instance, we need to make sure that that scope doesn't get cancelled when the Activity pauses (e.g. when it is overlaid by the purchase flow).

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
vegaro: That's very interesting, I was actually wondering how this test would look like without the await

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -145,47 +157,79 @@ internal class PaywallViewModelImpl(
...
+                            "is PurchasesAreCompletedBy.MY_APP"
+                    }
+                    val customerInfo = purchases.awaitCustomerInfo()
+                    when (val result = customRestoreHandler(customerInfo)) {
+                        is MyAppRestoreResult.Success -> {
+                            listener?.onRestoreCompleted(customerInfo)
+                        }
+                        is MyAppRestoreResult.Error -> {
+                            val purchasesException = PurchasesException(result.error)
+                            throw purchasesException
```
vegaro, line 194: should this call `listener?.onPurchaseError(result.error)` instead of throwing?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
vegaro: Sorry I missed this conversation but wanted to chime in as well by expressing my opinion, which I believe is the same as yours.

I'd vote for not using coroutines as well and have a regular callbacks because it gives more flexibility, and if someone wants to use coroutines themselves because their purchasing code uses coroutines they still can by wrapping it in a coroutine. If we make it suspend, we are forcing everyone to use coroutines, and some people might not be using those, specially if we take into account that launchBillingFlow doesn't have a suspended version in the billing-ktx library.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
vegaro: > Having said that, the SDK should already give us a lot of the information and track those purchases

Wouldn't the developer need to call `syncPurchases` for these to be synced? At least that's what a regular implementation looks like when completing purchases on your own.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
vegaro: Fwiw for Google, you still need to call `syncPurchases` for us to catch the new purchase, for Amazon, you are required to call `syncAmazonPurchase` which is slightly different because `syncPurchases` doesn't work when using our SDK side by side your own implementation, due to the Amazon SDK not being able to handle two purchase listeners at the same time.

tonidero: > Fwiw for Google, you still need to call syncPurchases for us to catch the new purchase

Hmm wouldn't we get it in the listener ourselves too without calling `syncPurchases`?

For Amazon, yeah they need to call the corresponding sync method.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: I believe this error was caught and then the catching code called the listener, but since we are no longer calling the listener, this code is gone so I will resolve.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -200,32 +241,88 @@ internal class PaywallViewModelImpl(
...
+                _actionError.value = e.error
+            }
+        }
+
+        finishAction()
+    }
+
+    private fun validateState() {
+        if (purchases.purchasesAreCompletedBy == PurchasesAreCompletedBy.MY_APP && options.myAppPurchaseLogic == null) {
+            error(
```
tonidero, line 321: This would crash... I'd avoid crashing and just display some error message in the paywall by setting the `_state` to the error state, for example. Not sure what behaviour we have in iOS though...

tonidero (commented):
This looks great! The only thing I would like to change is to avoid crashing when using `MY_APP` but not setting the logic handlers.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -28,20 +30,69 @@ import com.revenuecat.paywallstester.SamplePaywalls
...
 import com.revenuecat.purchases.ui.revenuecatui.fonts.CustomFontProvider
 import com.revenuecat.purchases.ui.revenuecatui.fonts.FontProvider
 
+class TesterAppPurchaseLogicSuspend : MyAppPurchaseLogic {
+
+    override suspend fun performPurchase(
+        activity: Activity,
+        rcPackage: com.revenuecat.purchases.Package,
+    ): MyAppPurchaseResult {
+        Log.d("RevenueCatUI", "Custom purchase code in performPurchase was called.")
```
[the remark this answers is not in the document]
jamesrb1: Done!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -28,20 +30,69 @@ import com.revenuecat.paywallstester.SamplePaywalls
...
+@Suppress("LongMethod")
 @Composable
 fun PaywallsScreen(
     samplePaywallsLoader: SamplePaywallsLoader = SamplePaywallsLoader(),
 ) {
     var displayPaywallState by remember { mutableStateOf<DisplayPaywallState>(DisplayPaywallState.None) }
 
+    val useCallbackPurchaseLogic = true
+
+    val myAppPurchaseLogic = remember {
```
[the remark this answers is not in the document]
jamesrb1: Done!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -28,20 +30,69 @@ import com.revenuecat.paywallstester.SamplePaywalls
...
 import com.revenuecat.purchases.ui.revenuecatui.fonts.FontProvider
 
+class TesterAppPurchaseLogicSuspend : MyAppPurchaseLogic {
+
+    override suspend fun performPurchase(
+        activity: Activity,
+        rcPackage: com.revenuecat.purchases.Package,
+    ): MyAppPurchaseResult {
+        Log.d("RevenueCatUI", "Custom purchase code in performPurchase was called.")
+        return MyAppPurchaseResult.Success
```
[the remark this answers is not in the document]
jamesrb1: Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ * Abstract class extending `MyAppPurchaseLogic`, providing methods for handling in-app purchases and restorations
+ * with completion callbacks rather than co-routines.
+ *
+ * If you prefer to implement custom purchase and restore logic with coroutines, please implement
+ * `MyAppPurchaseLogic` directly.
+ */
+abstract class MyAppPurchaseLogicCompletion : MyAppPurchaseLogic {
```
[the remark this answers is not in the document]
jamesrb1: Yes! Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -200,32 +241,88 @@ internal class PaywallViewModelImpl(
...
+                        is MyAppPurchaseResult.Cancellation -> {
+                            trackPaywallCancel()
+                        }
+                        is MyAppPurchaseResult.Error -> {
+                            result.errorDetails?.let { _actionError.value = it }
+                        }
+                    }
+                }
+                PurchasesAreCompletedBy.REVENUECAT -> {
+                    listener?.onPurchaseStarted(packageToPurchase)
```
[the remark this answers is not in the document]
jamesrb1: Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
jamesrb1: On iOS we add the red error banner if it's run in debug, and crash in release, the thought being that it is very bad to release anything where you think people can make purchases, but can't. However maybe that's too harsh? I updated Android so it always shows a dialog (pictured below) in debug or release. 

Untitled 39

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Interface for handling in-app purchases and restorations directly by the application rather than by RevenueCat.
+ * These suspend methods are called by a RevenueCat Paywall in order to execute you app's custom purchase/restore code.
+ * These functions are only called when `Purchases.purchasesAreCompletedBy` is set to `MY_APP`.
+ *
+ * If you prefer to implement custom purchase and restore logic with completion handlers, please implement
+ * `MyAppPurchaseLogicCompletion`.
+ */
+interface MyAppPurchaseLogic {
```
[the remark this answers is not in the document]
jamesrb1: Done:

- [PaywallOptionsAPI](https://github.com/RevenueCat/purchases-android/pull/1777/files#diff-166e012ec0c3830df9dec811c1672e004bcb92cb20f2375d91c7cc835c995e76)
- [MyAppPurchaseLogicAPI](https://github.com/RevenueCat/purchases-android/pull/1777/files#diff-d7eab110e8cf59e434e043c014ff11b1e95e4ba93b890570fbbc6d6235634391)

jamesrb1:
Thanks for the review @tonidero, changes made, ready for re-review. Thanks!

tonidero (approved):
LGTM! Great job!!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
JayShortway: What do you think about renaming `MyAppPurchaseLogic` to just `PurchaseLogic`? 

That way we keep the option open to ourselves to put our own logic in an internal implementation of `PurchaseLogic`, e.g. called `RevenueCatPurchaseLogic`. I think doing this refactor is out of scope of this PR, but the suggested rename would leave that door open.

JayShortway (approved):
Nicely done! 🙌 Added 1 question to an existing thread, here: https://github.com/RevenueCat/purchases-android/pull/1777#discussion_r1723242746

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
jamesrb1: I'm not sure, it was named `MyAppPurchaseLogic` in reference to PurchasesAreCompletedBy.MY_APP for which it is needed. If we need to implement something internally in the future, perhaps we can live with the "confusion" of having the base interface call "My App", because I think it does help customers connect the dots.

jamesrb1: Actually, I'll change it.

jamesrb1: Done here: https://github.com/RevenueCat/purchases-android/pull/1777/commits/d6683f77cfb6d3e135adf4fbe0cd85067deabee6

Also renamed the MyApp purchase/restore result types. Because `PurchaseResult` is already taken, it is `PurchaseLogicPurchaseResult` and `PurchaseLogicRestoreResult`.

jamesrb1:
I've did a rename on Joop's suggestion here: https://github.com/RevenueCat/purchases-android/pull/1777/commits/d6683f77cfb6d3e135adf4fbe0cd85067deabee6

jamesrb1:
Moved the suggested rename, plus result type consolidation, to a follow on PR: https://github.com/RevenueCat/purchases-android/pull/1814

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialogOptions.kt
```diff
@@ -82,6 +86,10 @@ data class PaywallDialogOptions internal constructor(
             this.listener = listener
         }
 
+        fun setMyAppPurchaseLogic(myAppPurchaseLogic: MyAppPurchaseLogic?) = apply {
```
tonidero, line 89: I noticed we don't have docs for a few of these... However, it would be good to add some docstring here explaining what this does (same for the `PaywallOptions` one, and that it will be ignored unless SDK is using `PurchasesAreCompletedBy.MyApp`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/MyAppPurchaseLogic.kt
JayShortway: You do make a good point haha. It's not a blocker for me in any way.

PR #1779, opened 2024-07-10
[AUTOMATIC] Release/8.2.1

RCGitBot:

### Bugfixes
* Attempt to fix ANRs by moving some tasks during configure to background (#1772) via Cesar de la Vega (@vegaro)
* Use trimmed API key (#1768) via Toni Rico (@tonidero)
### Other Changes
* Remove leftovers `ExperimentalPreviewRevenueCatUIPurchasesAPI` (#1778) via Cesar de la Vega (@vegaro)
* Update migration docs adding back support for `DEFERRED` upgrades (#1774) via Toni Rico (@tonidero)
* Add 7.12.0 to CHANGELOG (#1769) via Toni Rico (@tonidero)


.circleci/config.yml
```diff
@@ -115,6 +115,7 @@ commands:
             gcloud firebase test android run --type instrumentation \
               --app <<parameters.app_apk_path>> \
               --test <<parameters.test_apk_path>> \
+              --device model=panther,version=33,locale=en,orientation=portrait \
```
tonidero, line 118: Tests started failing on the Pixel5,30 we were using by default before... I tried this and they started passing again, maybe gcloud had some issues with the device 🤷‍♂️ . For now,  I think this is ok.

tonidero (approved):

.circleci/config.yml
JayShortway: If we're testing on 1 api level I think a recent one makes sense!

JayShortway (approved):

PR #1780, opened 2024-07-11
Test integration tests failures

tonidero:
### Description
Testing integration tests failures


tonidero:
After running integration tests in several situations, we believe the failures to be related to the device type and not to the changes in the latest release #1779 (tested reverting with and without changes in #1772 and both are flaky). Closing this.

PR #1783, opened 2024-07-18
Updates Gradle (8.9), AGP (8.5.1) and maven.publish (0.28.0)

JayShortway:
As the title says. 

Gradle was at 8.0 and needed to at least be 8.1. for maven.publish 0.28.0, so I updated to the latest.
There was no requirement to update AGP, but from experience it is good to keep that somewhat in sync with Gradle, so went with the latest here too.


tonidero (approved):
Changes LGTM!

PR #1785, opened 2024-07-18
Add Emerge snapshot testing

rbro112:
### Checklist

### Motivation
Adds Emerge snapshots to purchases UI, as RevenueCat has expressed interest and we figured we'd make the process easier 😄 .

### Description
Adds the basic integration necessary for getting up and running with Emerge snapshots:

- Adds the Emerge gradle plugin to the `testpurchasesuiandroidcompatibility` module.
  - Worth noting that Emerge requires the plugin to be set on an `application` module vs a library module. @tonidero had mentioned that the Previews we want to snapshot are in `:ui:revenuecatui`. Given that this application module depends directly on the `:ui:revenuecatui` module and snapshots use a debug variant, all code should be present from that module, giving Emerge all `@Previews` automatically.
- Adds what I believe is the necessary fastlane/CircleCI setup to invoke snapshots, but my experience here is a bit more limited.
- Updates Compose BOM from `2023.05.01` to `2024.06.00`, as Emerge snapshots rely on compose foundation, which can't be resolved without an update. Happy to help test this!

For the best usage, we recommend running snapshots on every commit - all PR commits and main commits. Emerge will run & snapshot all `@Preview` annotated functions, storing the generated snapshots based on the `sha` of that upload. 

If the commit has a valid `baseSha` set and we have snapshots for that `baseSha`, we'll automatically compare the snapshots and let you know if there's any visual changes directly on the PR. See our [snapshot docs](https://docs.emergetools.com/docs/snapshot-testing) for all functionality!

Remaining tasks for RevenueCat to implement to complete this:
- [ ] Connect the Emerge GitHub app ([docs](https://docs.emergetools.com/docs/github)). This allows us to post status checks on PRs, letting you know when a visual change has occurred.
- [ ] Add the `EMERGE_API_TOKEN` as a GitHub secret
- [ ] Wire up CircleCI variables to see the `sha` & `baseSha` in the Emerge gradle plugin config block. These are what we use for [diffing snapshots](https://docs.emergetools.com/docs/continuous-integration).
- [ ] Validate the fastlane/CircleCI setup I added is correct and what is intended.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -648,6 +648,7 @@ workflows:
       - run-backend-integration-tests
       - verify-revenuecatui-snapshots
       - run-revenuecatui-ui-tests
+      - emerge_purchases_ui_snapshot_tests
       - verify_debug_view_snapshot_tests: *release-branches
       - integration-tests-build: *release-branches
       - purchases-integration-tests-build: *release-branches
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -513,6 +513,21 @@ jobs:
             bundle exec fastlane android verify_debug_view_snapshot_tests
       - android/save-build-cache
 
+  emerge_purchases_ui_snapshot_tests:
+    description: "Emerge purchases ui snapshot tests"
+    <<: *android-executor
+    steps:
+      - checkout
+      - install-sdkman
+      - revenuecat/install-gem-unix-dependencies:
+          cache-version: v1
+      - android/restore-build-cache
+      - run:
+          name: Build and run purchases ui snapshot tests
+          command: |
+            bundle exec fastlane android emerge_purchases_ui_snapshot_tests
+      - android/save-build-cache
+
   run-firebase-tests:
     description: "Run integration tests for Android in Firebase. Variant latestDependencies"
     executor: gcp-cli/google
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -40,6 +40,8 @@ commonmark = "0.21.0"
 activity-compose = "1.7.2"
 fragment = "1.6.1"
 hamcrest = "1.3"
+emergeGradlePlugin = "3.1.2"
+emergeSnapshots = "1.1.3"
 
 [plugins]
 android-application = { id = "com.android.application", version.ref = "agp" }
@@ -48,6 +50,7 @@ androidx-navigation-safeargs = { id = "androidx.navigation.safeargs.kotlin", ver
 dependencyGraph = { id = "com.savvasdalkitsis.module-dependency-graph", version.ref = "dependencyGraph" }
 detekt = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt" }
 dokka = { id ="org.jetbrains.dokka", version.ref = "dokka"}
+emerge = { id = "com.emergetools.android", version.ref = "emergeGradlePlugin" }
 kotlin-android = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
 kotlin-parcelize = { id = "org.jetbrains.kotlin.plugin.parcelize", version.ref = "kotlin" }
 kotlin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlin" }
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -512,6 +512,21 @@ jobs:
             bundle exec fastlane android verify_debug_view_snapshot_tests
       - android/save-build-cache
 
+  emerge_purchases_ui_snapshot_tests:
```
tonidero, line 515: We will need to add this job to one of the workflows below. Probably the `build-test-deploy` workflow, so it runs on all commits.

gradle/libs.versions.toml
```diff
@@ -114,7 +119,7 @@ tink = { module = "com.google.crypto.tink:tink-android", version.ref = "tink" }
 coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
 
 kotlin-bom = "org.jetbrains.kotlin:kotlin-bom:1.8.0"
-compose-bom = "androidx.compose:compose-bom:2023.05.01"
+compose-bom = "androidx.compose:compose-bom:2024.06.00"
```
tonidero, line 122: We actually had to upgrade to `2024.08.00` recently due to an unrelated bug, so this change probably isn't needed anymore/might cause a conflict.

test-apps/testpurchasesuiandroidcompatibility/build.gradle.kts
```diff
@@ -48,10 +49,29 @@ android {
     }
 }
 
+emerge {
+    // TODO: RevenueCat to set from CircleCi variables
+    apiToken.set(System.getenv("EMERGE_API_TOKEN"))
+
+    vcs {
+        // TODO: RevenueCat to set from CircleCi variables
+        sha.set("")
```
tonidero, line 58: I think it might be good to get this from environment variables too. CircleCI sets this as `CIRCLE_SHA1`, but we can pass a more generic env like `COMMIT_SHA1` or something like that (or default to empty if none found). Wdyt?

test-apps/testpurchasesuiandroidcompatibility/build.gradle.kts
```diff
@@ -48,10 +49,29 @@ android {
...
+        // TODO: RevenueCat to set from CircleCi variables
+        // Should skip setting for main branch uploads
+        baseSha.set("")
```
tonidero, line 61: Just to check, this will be the sha for the latest commit in `main` right? I don't think CircleCI gives us this https://circleci.com/docs/variables/? Though we can calculate it through a script of course

tonidero (commented):
Really sorry for the delay! This is looking great! Just some small questions and I think we can get it merged

test-apps/testpurchasesuiandroidcompatibility/build.gradle.kts
rbro112: That is correct, needs to be the commit in the default branch (i.e. main) that the PR is based on. Seems we might be able to get it from GitHub data in CircleCI.

Proposition here: let's add this in a future PR. Most snapshots integrations are best as two steps anyway:
1. Add uploads from all main branch commits. This way we can ensure everything's running properly and we get snapshots for all base builds.
2. In follow up PR, we add support for PR commits. That way diffs we can test that diffs are working properly.

So this PR will be step 1 above, and you or I can follow with step 2 where we add the baseSha just for PR commits. Sound good?

test-apps/testpurchasesuiandroidcompatibility/build.gradle.kts
rbro112: Just needs to match GitHub for proper status check posting. Updated to use this variable!

gradle/libs.versions.toml
rbro112: Resolved!

.circleci/config.yml
rbro112: Added!

test-apps/testpurchasesuiandroidcompatibility/build.gradle.kts
tonidero: Makes a lot of sense thank you!

tonidero (approved):
This looks great! Will get this merged so we will start using this soon. Thank you!

PR #1791, opened 2024-07-25
[Paywalls] Fix `RemoteImage` flicker on activity changes

tonidero:
### Description
We were seeing some flickering in the paywall images when the activity was paused/started. This seems to have been caused because the loader was recreated every time. The loader only actually uses a reference to the application context, so we can just change it to avoid that recreation and store a reference to the image loader.


joshdholtz (approved):
:shipit: 💪

PR #1795, opened 2024-07-26
Diagnostics: Remove unused anonymizer

tonidero:
### Description
The `Anonymizer` was unused and we don't plan to track any sort of PII with this feature, so removing this code for now.


JayShortway (commented):
`PurchasesConfiguration.diagnosticsEnabled()` has the following line in the docs: 
> This information will be anonymized so it can't be traced back to the end-user.

Should we remove or reword that? We could say something like:
> No personal identifiable information will be collected.

Other than that it looks good to me!

JayShortway (approved):
LGTM!

PR #1796, opened 2024-07-29
Adds the Detekt IntelliJ plugin settings to version control.

JayShortway:
As the title says. The [Detekt IntelliJ plugin](https://plugins.jetbrains.com/plugin/10761-detekt) requires some manual setup. If every team member has to do that individually, there's a higher chance of discrepancies and errors.

### DIFF
--- a/.idea/detekt.xml
+++ b/.idea/detekt.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="DetektPluginSettings">
+    <option name="configurationFiles">
+      <list>
+        <option value="$PROJECT_DIR$/config/detekt/detekt.yml" />
+      </list>
+    </option>
+    <option name="enableDetekt" value="true" />
+  </component>
+</project>
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ vendor/
 
 # Snapshots repo folder
 purchases-android-snapshots
+
+# Make sure we share the same Detekt IntelliJ plugin settings
+!.idea/detekt.xml


tonidero (approved):
🙌 Love it!

.idea/detekt.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="DetektPluginSettings">
+    <option name="configurationFiles">
+      <list>
+        <option value="$PROJECT_DIR$/config/detekt/detekt.yml" />
```
tonidero, line 6: One question, will this pickup the baseline file as well? Or maybe we should just remove it...

JayShortway: It won't, but it can. It's a separate setting. 

However, I deliberately did not include it, as I thought it would be nice to keep being reminded of our "debt" in the IDE. If we add it, all those warnings will be ignored, and I'm afraid we might completely forget about them. Let me know if you disagree though!

tonidero: Well, my only concern is having different configuration in the terminal vs in the IDE... but it's probably fine yeah. In my mind, when we touch the files with those errors, we will probably modify them... Which I think is fine 👍

vegaro: I have been wondering if we need the baseline afterall, I think the ignores should be inline the code, it's easy to miss a warning is ignored if it's in the baseline file

vegaro (approved):

.idea/detekt.xml
JayShortway: Wait, @vegaro's comment was not showing for me when I replied yesterday. So my reply is to @tonidero haha. Apologies. 
  
My thinking was that having the warnings in the baseline still show up in the IDE, is that it incentivizes us to fix them. 

However with 2 people on the fence about this approach, I'll add the baseline to the IntelliJ plugin settings. PR coming up.

JayShortway: For reference: https://github.com/RevenueCat/purchases-android/pull/1798

PR #1805, opened 2024-08-07
ISSUE-1765: Make suspend coroutine wrappers cancellable

### LINKED ISSUE #1765: Possible bug ?
https://github.com/RevenueCat/purchases-android/blob/36ddd195619d39653d40c877d56ac1cefc3a0e92/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt#L24

it should be suspendCancellableCoroutine instead of suspendCoroutine

IlyaPavlovskii:
### Checklist

### Motivation
Stuck with the same problem mentioned in [the issue 1765](https://github.com/RevenueCat/purchases-android/issues/1765)

Resolves #1765

### Description

As it was mention in the description: when we cancel a coroutine using a job.cancel the suspendCoroutine doesn't cancel the revenueCat call.. but wrapping the normal non suspend function for get current customer in my own suspendCancellableCoroutine cancells the coroutine.

Checked on own integration - coroutines cancelled immediately after `job.cancel()` call, instead of prev solution, when await function might execute some code postponed, event it was cancelled


JayShortway:
Hi, thanks for taking the effort to raise this PR! 

I replied in #1765. Let's discuss it there for now.

PR #1806, opened 2024-08-08
Support to round+truncate prices displayed on paywalls for backend-specified countries

jamesrb1:
## Round and truncate prices displayed on paywalls for storefronts specified by the backend

Some countries have cents as part of their currency, but it is rare to see them written in prices. This PR makes it possible to round/hide the cents component.

Prices are only truncated if the price ends in .00.

The steps to do this are:

### 1. Get list of storefronts for which rounding should occur. 
- Get list of prices of storefront countries for which rounding applies. There are different storefront codes for Google (Android) and Apple, and we only save the Google ones, in `PaywallData` as `zeroDecimalPlaceCountries`. 

### 2. Determine if this user is in a "rounding" country.
- If the country of the storefront that the device is connected to is in the list returned by the backend, we set a new property, `VariableProcessor.PackageContext.showZeroDecimalPlacePrices` to `true`. `VariableProcessor.PackageContext` is used by `VariableProcessor` to translate `{{ price_variables }}` into the correct strings.

### 3. Round prices if needed while processing variables
- When the `VariableProcessor` is asked for a localized price string (`localizedPrice(...)`, `localizedPricePerWeek(...)` etc., it returns a rounded price if rounding conditions are met.

The server is currently returning "PH", "KZ", "TW", "MX", "TH" as zero decimal countries. If your google account is not in one of these countries, you can turn temporarily turn rounding "on" by going to `PackageConfigurationFactory.kt` and  replacing

```kotlin
                paywallData.zeroDecimalPlaceCountries.contains(
                    Purchases.sharedInstance.storefrontCountryCode,
                )
```
with
```kotlin
                paywallData.zeroDecimalPlaceCountries.contains(
                    "TW",
                )
```

Note that prices are formatted according to the locale of the RC paywall. For example, if the device's language setting is Canadian English (en_CA), and attached the Canadian Play store (CA), but there is no RC paywall localized to English (Canada), then the paywall that we show falls back to the en_US paywall, and all prices will be formatted according to en_US rules, meaning a price will show as e.g. "CA$7.99". If there is an en_CA paywall, then prices are formatted according to en_CA rules, e.g. "$7.99".

This is the Android version of https://github.com/RevenueCat/purchases-ios/pull/4132


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -76,6 +76,9 @@ class Purchases internal constructor(
     val appUserID: String
         @Synchronized get() = purchasesOrchestrator.appUserID
 
+    val storefrontCountryCode: String?
+        @Synchronized get() = purchasesOrchestrator.storefrontCountryCode
```
JayShortway, line 80: "Storefront" is quite an Apple-focused term which I haven't really seen being used in relation to Google Play / Android, although it does cover the meaning well. 

If we want to make this more generic we could opt for `storeCountryCode`? (On both platforms I mean.) What do you think?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -76,6 +76,9 @@ class Purchases internal constructor(
     val appUserID: String
         @Synchronized get() = purchasesOrchestrator.appUserID
 
+    val storefrontCountryCode: String?
```
tonidero, line 79: Need to add these to the API tests?

JayShortway (commented):
Looking good! Just a few minor comments.

JayShortway:
`StoreProduct` and `PricingPhase` have `.pricePerPeriod()` functions too. Should they exhibit the same rounding behavior, or is this purely Paywall presentation logic?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -7,11 +7,37 @@ import com.revenuecat.purchases.utils.serializers.OptionalURLSerializer
...
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.jsonArray
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
 import java.net.URL
 import java.util.Locale
 
+object GoogleListSerializer : KSerializer<List<String>> {
```
tonidero, line 25: This should be made `internal` or `private`. Also, might be good to move to a different file in the `serializers` folder?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -44,6 +70,10 @@ data class PaywallData(
 
     @SerialName("localized_strings_by_tier") internal val localizationByTier:
     Map<String, Map<String, LocalizedConfiguration>> = emptyMap(),
+
+    @SerialName("zero_decimal_place_countries")
+    @Serializable(with = GoogleListSerializer::class)
```
tonidero, line 75: Hmm I guess this is to dereference the `google` sublevel without having to represent that in the data object right? IMO It would have been better if the backend provided the appropriate list instead of having to do this now and using more data... But we can go with this for now.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
```diff
@@ -44,6 +70,10 @@ data class PaywallData(
...
+    val zeroDecimalPlaceCountries: List<String> = emptyList(),
```
tonidero, line 76: Also need to add this to API tests

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -222,11 +223,20 @@ internal object PackageConfigurationFactory {
                 it.product.pricePerMonth(),
                 mostExpensivePricePerMonth,
             )
+
+            val shouldRound = if (Purchases.isConfigured) {
```
tonidero, line 227: Hmm I might be missing something but is this check needed? I would assume that if we have `paywallData` at all, the SDK should be configured?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -177,3 +222,25 @@ private fun Period.normalizedMonths(): Period {
...
     }
 }
+
+// Price extensions
+
+@SuppressWarnings("MagicNumber")
+private fun Price.endsIn99or00Cents(): Boolean {
+    val normalPrice = amountMicros / MICRO_MULTIPLIER
+    val roundedCents = (normalPrice * 100).toInt() % 100
+    return roundedCents == 99 || roundedCents == 0
```
tonidero, line 232: I will say, I'm a bit scared about rounding this number when it's `99` cents... This basically means we're not providing the real price by the store... I think it would be ok if it was a floating number precision error (which shouldn't happen I think?), but not a whole cent. Wdyt?

tonidero (commented):
Mostly some API tests missing and some questions but looking good!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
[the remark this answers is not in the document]
jamesrb1: Ya, it's used to decode `google` only, and then flatten that directly onto `zeroDecimalPlaceCountries`, so we access that directly rather than going `zeroDecimalPlaceCountries.google` as it would be if it reflected the json.

I'm a bit hesitant to diverge from the json structure, however in this case it seems like a safe bet and makes the API nicer. Although thinking more now, it's possible that other Android stores could use different storefront codes...? If that were the case we'd have to choose the correct one to decode and expose via `zeroDecimalPlaceCountries`. I don't _think_ we'd ever need to have more than one set?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -222,11 +223,20 @@ internal object PackageConfigurationFactory {
...
+                paywallData.zeroDecimalPlaceCountries.contains(
+                    Purchases.sharedInstance.storefrontCountryCode,
+                )
```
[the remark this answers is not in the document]
jamesrb1: Yes, good point.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
jamesrb1: Ya, this value should be passed in from elsewhere. The issue was with a number of tests that hit his method without configuring Purchases.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
jamesrb1: storefront is used by Apple, but it is also generic term that can be used to refer to the user-facing aspect of an online store, which we've adopted into the Android SDK already.

```kotlin
interface StorefrontProvider {
    fun getStorefront(): String?
}
```

```kotlin
    /**
     * Obtain store country code in ISO 3166-1-alpha-2 standard format.
     * Null if there has been an error.
     */
    abstract fun getStorefront(
        onSuccess: (String) -> Unit,
        onError: PurchasesErrorCallback,
    )
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
jamesrb1: Ya, will raise on Slack.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
jamesrb1: Done and done!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
jamesrb1: I don't see any explicit API testers for PaywallData, but the sample paywalls in SamplePaywalls.kt serves this purpose. Is there somewhere else this should be added?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
jamesrb1: Done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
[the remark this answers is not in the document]
jamesrb1: Done! https://github.com/RevenueCat/purchases-android/pull/1806/commits/5912b9f9aaa4263ea8fda236209505f24b70a652

jamesrb1:
> `StoreProduct` and `PricingPhase` have `.pricePerPeriod()` functions too. Should they exhibit the same rounding behavior, or is this purely Paywall presentation logic?

AFAIK this rounding is only needed for paywall presentation. I see you added the code for `PricingPhase` - do you know if these values are _ever_ presented to the user?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
jamesrb1: Actually, can a developer set a price on Google Play to $10.00, or do you have to choose $9.99? Because I assume they do want to round that to $10, which would be more than floating-point precision rounding. I can check this tomorrow (my @revenuecat.com needs an invitation to Google Play), but I think what we do here largely depends on how precisely developers can set pricing.

tonidero: Google does allow you to set any prices you want, in any currency (within some bounds). So yeah, I think I would only allow rounding for amounts without cents.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: Yea I'm totally fine with keeping it! It's a word and it has a meaning and it applies here 😄

JayShortway:
> > `StoreProduct` and `PricingPhase` have `.pricePerPeriod()` functions too. Should they exhibit the same rounding behavior, or is this purely Paywall presentation logic?
> 
> AFAIK this rounding is only needed for paywall presentation. I see you added the code for `PricingPhase` - do you know if these values are ever presented to the user?

(Really missing GitLab's comment threads ☝️) 

Alright! They're not presented to the user, unless the app developer does it in their own UI. They're purely convenience functions (but I do imagine their main purpose is to show in UI). 

cc @tonidero for verification 🙏

tonidero:
> They're not presented to the user, unless the app developer does it in their own UI. They're purely convenience functions (but I do imagine their main purpose is to show in UI).

Yeah, that's true. It's part of the public API, in case a developer wants to use them. They can also write this logic themselves for sure, so this would be a mostly convenience change. However, note that this is configured from the paywalls, so maybe it makes more sense to restrain this logic to paywalls and just use what comes from the store otherwise? Happy to discuss this more over Slack if needed.

JayShortway:
> so maybe it makes more sense to restrain this logic to paywalls and just use what comes from the store otherwise? 

Yea I think that makes a lot of sense!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
jamesrb1: Done!

jamesrb1:
I believe I've addressed all concerns raised.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallData.kt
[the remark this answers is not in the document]
JayShortway: Yea we could define the `ZeroDecimalPlaceCountries` object and omit the `apple` field (so it only has `google`). That should work because we have set `ignoreUnknownKeys = true` afaik. Then you'd indeed need to do `zeroDecimalPlaceCountries.google` but you avoid the need for a custom serializer. 

I'm fine either way. Not a blocker.

JayShortway (approved):
LGTM! 

To fix `PurchasesAPI.java`: 
```diff
- final String storefrontCountryCode = purchases.storefrontCountryCode();
+ final String storefrontCountryCode = purchases.getStorefrontCountryCode();
```

Java loves its getters and setters. So a `val` in Kotlin becomes a getter. A `var` becomes a pair of getter and setter.

jamesrb1:
> To fix PurchasesAPI.java:

THANK YOU!!!

PR #1814, opened 2024-08-20
Rename and result type consolidation

jamesrb1:
- Renames `MyAppPurchaseLogic` to `PurchaseLogic`
- Consolidates `MyAppPurchaseResult` and `MyAppRestoreResult` into the single type `PurchaseLogicResult`.

If the user cancels a restore, it is ignored, because this was probably the result of a user action and doesn't necessitate any sort of additional information. (The app implementer could also implement action if required.)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
```diff
@@ -104,39 +104,23 @@ abstract class MyAppPurchaseLogicWithCallback : MyAppPurchaseLogic {
 /**
  * Represents the result of a purchase attempt made by custom app-based code (not RevenueCat).
  */
-sealed interface MyAppPurchaseResult {
+sealed interface PurchaseLogicResult {
     /**
-     * Indicates a successful purchase.
+     * Indicates a successful purchase or restore.
      *
```
tonidero, line 110: Just noticed there is an extra line here.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialogOptions.kt
```diff
@@ -86,8 +86,8 @@ data class PaywallDialogOptions internal constructor(
             this.listener = listener
         }
 
-        fun setMyAppPurchaseLogic(myAppPurchaseLogic: MyAppPurchaseLogic?) = apply {
-            this.myAppPurchaseLogic = myAppPurchaseLogic
+        fun AppPurchaseLogic(purchaseLogic: PurchaseLogic?) = apply {
```
tonidero, line 89: I think this is supposed to be
```suggestion
        fun setPurchaseLogic(purchaseLogic: PurchaseLogic?) = apply {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -79,8 +79,8 @@ data class PaywallOptions internal constructor(
             this.listener = listener
         }
 
-        fun setMyAppPurchaseLogic(myAppPurchaseLogic: MyAppPurchaseLogic?) = apply {
-            this.myAppPurchaseLogic = myAppPurchaseLogic
+        fun setPurchaseLogic(purchaseLogic: PurchaseLogic?) = apply {
```
tonidero, line 82: I honestly don't fully love this naming... Feels a bit too generic for what it does from an API standpoint... I kinda prefer what we had before `setMyAppPurchaseLogic` or maybe `setCustomPurchaseLogic`. Wdyt?

tonidero (commented):
I think this is ok, just that rename for the `setter` option

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
JayShortway: I agree that the function name could be too generic. `setCustomPurchaseLogic` could work 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialogOptions.kt
JayShortway: @tonidero This would then have the same comment as https://github.com/RevenueCat/purchases-android/pull/1814/files#r1724558113 I assume?

JayShortway (commented):
Looks good! No additional comments besides Toni's.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialogOptions.kt
tonidero: Yes that’s correct! That comment applies here too

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
vegaro: I think `setCustomPurchaseLogic` could infer it's always going to set it to `MyApp`, as in set it to use my custom logic. I think I prefer `setMyAppPurchaseLogic`

vegaro: I agree as well.  Removing the `custom` makes it look like it is mandatory to set I think. Also, why is the `purchaseLogic` parameter optional? Why would someone set it to null?

jamesrb1: Thanks! I've changed the setter to `setCustomPurchaseLogic`.

As for the nullable parameter, I was following the pattern for assigning objects established above it:

```kotlin
        fun setOffering(offering: Offering?) = apply {
            this.offering = offering
        }

        fun setFontProvider(fontProvider: FontProvider?) = apply {
            this.fontProvider = fontProvider
        }

        fun setListener(listener: PaywallListener?) = apply {
            this.listener = listener
        }
```

If they aren't nullable, calling code that currently looks like this:

```kotlin
        PaywallDialogOptions.Builder()
            .setDismissRequest(onDismiss)
            .setOffering(currentState.offering)
            .setFontProvider(currentState.fontProvider)
            .setCustomPurchaseLogic(currentState.purchaseLogic)
            .build()
```

gets a little messier:

```kotlin
        PaywallDialogOptions.Builder().apply {
            setDismissRequest(onDismiss)
            currentState.offering?.let { setOffering(it) }
            currentState.fontProvider?.let { setFontProvider(it) }
            currentState.purchaseLogic?.let { setCustomPurchaseLogic(it) }
        }.build()
```

joshdholtz: Nullable looks fine to me since it matches the existing methods for consistency 🤷‍♂️

joshdholtz (approved):
This looks good to me! 👍

PR #1817, opened 2024-08-22
Update `compose-bom` to latest version

tonidero:
### Description
We are getting some issues with Template7 of paywalls in RN and flutter (it works fine on native): 
- https://community.revenuecat.com/sdks-51/revenuecat-paywall-component-causing-crash-on-android-nosuchmethoderror-in-animatedcontent-method-4956
- https://github.com/RevenueCat/purchases-flutter/issues/1145

From looking around at the issue, some people mentioned a library incompatibility, so I updated the compose versions we're using and that seems to solve the issue. After investigating, seems that the hybrids are updating to a newer version of compose that is not compatible with the version we were using in the native. We were also using an experimental API (which is not experimental in the latest version)

I did a run through our UI and didn't notice any issues, but might be good for someone else to double-check


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
```diff
@@ -368,7 +366,7 @@ private fun Features(
     val localization = selectedLocalizationForTier(selectedTier)
 
     AnimatedContent(
-        targetState = selectedTier,
+        targetState = localization,
```
tonidero, line 369: Changed this, since a linter was complaining if we weren't using the state within the `AnimatedContent` block.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
```diff
@@ -357,7 +356,6 @@ private fun selectedLocalizationForTier(tier: TemplateConfiguration.TierInfo): P
     return tier.defaultPackage.localization
 }
 
-@OptIn(ExperimentalAnimationApi::class)
```
tonidero, line 360: Not experimental anymore

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template7.kt
JayShortway: That's good! Leads to subtle bugs if you don't use the state.

JayShortway (approved):
LGTM!

PR #1818, opened 2024-08-22
[AUTOMATIC] Release/8.5.1

RCGitBot:

### Dependency Updates
* Update `compose-bom` to 2024.08.00 (#1817) via Toni Rico (@tonidero)


JayShortway (approved):

CHANGELOG.latest.md
```diff
@@ -1,5 +1,2 @@
-### New Features
-* [Paywalls] Add default locale (#1809) via Josh Holtz (@joshdholtz)
-* [Paywalls] Support to round+truncate prices displayed on paywalls for backend-specified countries (#1806) via James Borthwick (@jamesrb1)
-### Other Changes
-* Publish paywall tester on each release (#1803) via Toni Rico (@tonidero)
+### Dependency Updates
+* Update `compose-bom` to latest version (#1817) via Toni Rico (@tonidero)
```
vegaro, line 2: Might be a good idea to indicate the version here

vegaro: Also, maybe we should at least make it a minor?

tonidero: Yes to the version. As for making it a minor... Following semver, this isn't really a new feature, so it should be a patch? And I also don't think people would look that much closer a minor vs a patch IMO. Wdyt?

vegaro: We have determined the version number before based on dependencies updates, and since we are changing dependencies by a minor number, I thought this could be a minor. That was my reasoning.

tonidero: Ok that's fine yeah, will just regenerate the release PR then!

PR #1821, opened 2024-08-22
Converts MagicWeather's Gradle files to Kotlin

JayShortway:
As the title says.


examples/MagicWeather/app/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+    alias(libs.plugins.kotlin.android)
+}
+
+android {
+    val compileVersion: Int by rootProject.extra
+    compileSdk = compileVersion
+
+    defaultConfig {
+        applicationId = "com.revenuecat.purchases_sample"
+        minSdk = 26
```
tonidero, line 12: Should we also move the minSdk to the root gradle.kts? I was also thinking about the `applicationId`... But maybe it's ok to leave here.

tonidero (approved):
LGTM!

examples/MagicWeather/app/build.gradle.kts
JayShortway: Hmm now that you bring it up, I'm actually more inclined to move `compileVersion` to the `:app` module's build file haha. That's where it is used, and we're not sharing these variables with any other modules within MagicWeather. What do you think?

vegaro (approved):

PR #1822, opened 2024-08-22
Adds Detekt rules for Compose. Updates detekt to 1.23.6.

JayShortway:
## Motivation
Compose has many footguns that are easy to forget. (Think state restoration, remembering, modifier parameters, viewmodel forwarding.) It's unrealistic to expect everyone to always remember all of these. These Detekt rules help to avoid these issues a lot, in my experience. Static checks also ease the pressure on code reviews to spot these issues.  

You can find a description of the rules [here](https://mrmans0n.github.io/compose-rules/rules/). They were originally developed at [Twitter](https://github.com/twitter/compose-rules) (1.0), and are currently maintained by [the engineer who started the project](https://x.com/mrmans0n/status/1507390768796909571). Slack is maintaining [the Lint version](https://slackhq.github.io/compose-lints/) of these same rules. 

**Note:** the Detekt IntelliJ plugin doesn't show violations of these rules (running via Gradle / CI does). For that we'd need to point it to a 66 Mb plugin jar file, which is probably a bad idea to add to the repo. We can do something fancy with Gradle to download it automatically, but that's less of a quick win than just adding the rules. So I figured we can do that later.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,7 @@ plugins {
 
 dependencies {
     detektPlugins libs.detekt.formatting
+    detektPlugins libs.detekt.compose
 }
 
 allprojects {
--- a/config/detekt/detekt-baseline.xml
+++ b/config/detekt/detekt-baseline.xml
@@ -2,9 +2,20 @@
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
+    <ID>ComposableNaming:PaywallAPI.kt$PaywallAPI$check</ID>
+    <ID>ComposableNaming:PaywallAPI.kt$PaywallAPI$checkDialog</ID>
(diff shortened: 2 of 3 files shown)


jamesrb1 (approved):
Excellent

tonidero (approved):
Haven't used this set of rules, but I think they make sense! I'm afraid of having the smells in the baseline will make us forget it... But I think at least it enables us to fix things moving forward 👍

JayShortway:
@tonidero I was debating that in my mind as well. In the end I figured that if we want to fix the smells before adding these rules, it'd be a while before we reap the benefits of them. Maybe we add tickets or something, to tackle 1 offended rule at a time from the baseline?

tonidero:
I think that makes sense yeah. I wouldn't consider it a priority, but we can dedicate some time if we have some downtime

PR #1825, opened 2024-08-28
[Diagnostics] Send backend_error_code in http request failures if available

tonidero:
### Description
This adds a new property to the `http_request_performed` diagnostics event including information of any backend provided error codes if any.

- [x] Hold until backend support has been deployed.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.common.networking
 
 import com.revenuecat.purchases.VerificationResult
+import com.revenuecat.purchases.common.isSuccessful
 import org.json.JSONObject
 import java.util.Date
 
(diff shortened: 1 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -51,6 +52,15 @@ internal data class HTTPResult(
 
     val body: JSONObject = payload.takeIf { it.isNotBlank() }?.let { JSONObject(it) } ?: JSONObject()
 
+    val backendErrorCode: Int? = if (!isSuccessful()) body.optInt("code").takeIf { it > 0 } else null
```
jamesrb1, line 55: What would a zero or negative code mean, or is this just theoretically impossible? Apparent AWS' load balancer [can return 000](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#response-code-000) before it sends a GOAWAY and closes the connection 🤨.

jamesrb1 (approved):
Looks good to me, just one question!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
tonidero: These are usually the codes sent by our own servers, which start on `7XXX`. I believe those kind of errors shouldn't return a body with a `code` field like this, so we should be safe from what I can tell. Thanks for checking!

PR #1829, opened 2024-09-04
Add `$kochavaDeviceId` attribution convencience method

tonidero:
### Description
This adds a new API `setKochavaDeviceId` that allows to set the kochava device ID required for the Kochava integration. 

Keeping on hold until backend support is added

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -634,6 +634,16 @@ class Purchases internal constructor(
         purchasesOrchestrator.setCleverTapID(cleverTapID)
     }
 
+    /**
+     * Subscriber attribute associated with the Kochava Device ID for the user
+     * Recommended for the RevenueCat Kochava integration
+     *
+     * @param kochavaDeviceID null or an empty string will delete the subscriber attribute.
+     */
+    fun setKochavaDeviceID(kochavaDeviceID: String?) {
+        purchasesOrchestrator.setKochavaDeviceID(kochavaDeviceID)
+    }
+
     // endregion
     // region Campaign parameters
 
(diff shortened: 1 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/subscriberattributes/SpecialSubscriberAttributes.kt
```diff
@@ -67,6 +68,7 @@ internal sealed class SubscriberAttributeKey(val backendKey: String) {
         object Facebook : AttributionIds(ReservedSubscriberAttribute.FB_ANON_ID)
         object Mparticle : AttributionIds(ReservedSubscriberAttribute.MPARTICLE_ID)
         object CleverTap : AttributionIds(ReservedSubscriberAttribute.CLEVER_TAP_ID)
+        object Kochava : AttributionIds(ReservedSubscriberAttribute.KOCHAVA_DEVICE_ID)
```
tonidero, line 71: I thought about moving this to the `DeviceIdentifiers` but seems to fit better here

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -634,6 +634,16 @@ class Purchases internal constructor(
         purchasesOrchestrator.setCleverTapID(cleverTapID)
     }
 
+    /**
+     * Subscriber attribute associated with the Kochava Device ID for the user
+     * Recommended for the RevenueCat Kochava integration
+     *
+     * @param kochavaDeviceID null or an empty string will delete the subscriber attribute.
+     */
+    fun setKochavaDeviceID(kochavaDeviceID: String?) {
```
tonidero, line 643: I thought about using `setKochavaID`, but this seems to better reflect the data.

JayShortway: Yup, more explicit is more better! They seem to have [a few different types of IDs](https://www.kochava.com/glossary/), so this avoids all confusion.

purchases/src/main/kotlin/com/revenuecat/purchases/common/subscriberattributes/SpecialSubscriberAttributes.kt
JayShortway: Yea I agree, it's mainly an attribution thing. Without Kochava, this ID doesn't exist.

JayShortway (approved):
LGTM!

vegaro (approved):

tonidero:
Changing labels since we are temporarily reverting this change in #1838. Will add the label on the PR bringing this back.

PR #1831, opened 2024-09-05
[EXTERNAL] Add Emerge snapshot testing (#1785) by @rbro112

tonidero:
### Checklist

### Motivation
Adds Emerge snapshots to purchases UI, as RevenueCat has expressed interest and we figured we'd make the process easier 😄 .

### Description
Adds the basic integration necessary for getting up and running with Emerge snapshots:

- Adds the Emerge gradle plugin to the `testpurchasesuiandroidcompatibility` module.
- Worth noting that Emerge requires the plugin to be set on an `application` module vs a library module. @tonidero had mentioned that the Previews we want to snapshot are in `:ui:revenuecatui`. Given that this application module depends directly on the `:ui:revenuecatui` module and snapshots use a debug variant, all code should be present from that module, giving Emerge all `@Previews` automatically.
- Adds what I believe is the necessary fastlane/CircleCI setup to invoke snapshots, but my experience here is a bit more limited.
- Updates Compose BOM from `2023.05.01` to `2024.06.00`, as Emerge snapshots rely on compose foundation, which can't be resolved without an update. Happy to help test this!

For the best usage, we recommend running snapshots on every commit - all PR commits and main commits. Emerge will run & snapshot all `@Preview` annotated functions, storing the generated snapshots based on the `sha` of that upload.

If the commit has a valid `baseSha` set and we have snapshots for that `baseSha`, we'll automatically compare the snapshots and let you know if there's any visual changes directly on the PR. See our [snapshot docs](https://docs.emergetools.com/docs/snapshot-testing) for all functionality!

Remaining tasks to implement to complete this:
- [ ] Wire up CircleCI variables to see the `sha` & `baseSha` in the Emerge gradle plugin config block. These are what we use for [diffing snapshots](https://docs.emergetools.com/docs/continuous-integration). Will do this in follow up PRs

Contributed by @rbro112 in #1785

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -513,6 +513,21 @@ jobs:
             bundle exec fastlane android verify_debug_view_snapshot_tests
       - android/save-build-cache
 
+  emerge_purchases_ui_snapshot_tests:
+    description: "Emerge purchases ui snapshot tests"
+    <<: *android-executor
+    steps:
+      - checkout
+      - install-sdkman
+      - revenuecat/install-gem-unix-dependencies:
+          cache-version: v1
+      - android/restore-build-cache
+      - run:
+          name: Build and run purchases ui snapshot tests
+          command: |
+            bundle exec fastlane android emerge_purchases_ui_snapshot_tests
+      - android/save-build-cache
+
   run-firebase-tests:
     description: "Run integration tests for Android in Firebase. Variant latestDependencies"
     executor: gcp-cli/google
@@ -633,6 +648,7 @@ workflows:
       - run-backend-integration-tests
       - verify-revenuecatui-snapshots
       - run-revenuecatui-ui-tests
+      - emerge_purchases_ui_snapshot_tests
       - verify_debug_view_snapshot_tests: *release-branches
       - integration-tests-build: *release-branches
       - purchases-integration-tests-build: *release-branches
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ plugins {
     alias libs.plugins.kover apply false
     alias libs.plugins.detekt
     alias libs.plugins.dependencyGraph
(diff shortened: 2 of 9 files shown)


tonidero:
Detekt + failing snapshot tests should be fixed in #1832

tonidero:
Hmm is there a way to disable the size analysis check from Emerge @rbro112? Mostly trying to keep the tidy green checks for now, until we get the baseSha working :)

rbro112:
@tonidero correct - although we'll need the baseSha set for comparisons (size & snapshots) to work on PRs which we can add in a follow on PR!

Merging this now looks all good from everything I can tell! Next step would be to merge/get snapshots running on the main branch to build a base. Then once we get the baseSha set in an upcoming PR, the status checks (both size & snapshots) will diff properly which is where both size & snapshots gets really powerful!

.circleci/config.yml
```diff
@@ -513,6 +513,21 @@ jobs:
...
+    steps:
+      - checkout
+      - install-sdkman
+      - revenuecat/install-gem-unix-dependencies:
+          cache-version: v1
+      - android/restore-build-cache
+      - run:
+          name: Build and run purchases ui snapshot tests
+          command: |
+            bundle exec fastlane android emerge_purchases_ui_snapshot_tests
```
JayShortway, line 528: I was gonna comment on the naming of the CircleCI job and Fastlane lane, as the Gradle module is called `revenuecatui`, and we're calling it "purchases ui" here, but I see  that the existing test app already has "purchasesui" in the name, so that inconsistency is not new haha.

JayShortway (approved):
So nice! 🙌

.circleci/config.yml
tonidero: Yeah that's a good comment... I think we can fix separately to make it consistent 👍

PR #1832, opened 2024-09-05
[Paywalls] Fix previews emerge snapshots

tonidero:
### Description
This fixes the broken previews we had so snapshots work as expected in the emerge snapshots PR #1831

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
@@ -6,14 +6,9 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.wrapContentHeight
-import androidx.compose.foundation.rememberScrollState
-import androidx.compose.foundation.verticalScroll
-import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.tooling.preview.Preview
 
 /**
  * Composable offering a minified screen Paywall UI configured from the RevenueCat dashboard.
@@ -61,34 +56,4 @@ fun PaywallFooter(
             paywallComposable()
         }
     }
-}
-
-@Suppress("MagicNumber")
-@Preview(showBackground = true)
-@Composable
-private fun PaywallFooterPreview() {
-    PaywallFooter(
-        options = PaywallOptions.Builder(dismissRequest = {}).build(),
-    ) {
-        Column(
-            modifier = Modifier
-                .fillMaxWidth()
-                .verticalScroll(rememberScrollState())
-                .padding(it),
-        ) {
-            // TODO-PAYWALLS: Implement an actual sample paywall
-            for (i in 1..50) {
-                Text(text = "Main content $i")
-            }
-        }
-    }
-}
-
-@Suppress("MagicNumber")
-@Preview(showBackground = true)
-@Composable
-private fun PaywallFooterNoContentPreview() {
-    PaywallFooter(
-        options = PaywallOptions.Builder(dismissRequest = {}).build(),
-    )
 }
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
```diff
@@ -62,33 +62,3 @@ fun PaywallFooter(
         }
     }
 }
-
-@Suppress("MagicNumber")
-@Preview(showBackground = true)
-@Composable
-private fun PaywallFooterPreview() {
```
tonidero, line 69: These seem to have been broken for a while... Basically we're not mocking the view model, so it can't work on the previews. We do have footer previews for all templates (and the default paywall now) so I think we should be mostly covered...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -70,15 +71,25 @@ internal fun Markdown(
...
+            fontWeight = fontWeight,
+            textAlign = textAlign,
+        )
+    } else {
+        val root = parser.parse(text) as Document
 
-    val density = LocalDensity.current
-    val paragraphPadding = with(density) { style.lineHeight.toDp() }
+        val density = LocalDensity.current
+        val paragraphPadding = with(density) { style.lineHeight.toDp() }
```
tonidero, line 86: This was failing during previews because it seems it was passing a non-sp unit, which can't be converted to `dp`. I opted to just use a normal text composable for previews, but we can dedicate some more time in the future to figure out what's wrong if needed.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PaywallDataExtensions.kt
```diff
@@ -137,15 +136,39 @@ internal fun DefaultPaywallPreview() {
...
+internal fun DefaultPaywallPreview() {
+    InternalPaywall(
+        options = PaywallOptions.Builder(dismissRequest = {}).build(),
+        viewModel = MockViewModel(offering = getDefaultPreviewOffering()),
+    )
+}
+
+@Preview(showBackground = true, locale = "en-rUS", group = "footer")
+@Composable
+internal fun DefaultPaywallFooterPreview() {
```
tonidero, line 160: Added these previews to have footer versions of the default paywall as well.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -73,7 +74,13 @@ internal fun Markdown(
     val root = parser.parse(text) as Document
 
     val density = LocalDensity.current
-    val paragraphPadding = with(density) { style.lineHeight.toDp() }
+    val paragraphPadding = with(density) {
+        if (style.lineHeight.type == TextUnitType.Sp) {
```
tonidero, line 78: This was causing some snapshots not to work because the unit wasn't Sp. For now, I'm just making the spacing 0 if the unit is not Sp, which didn't seem to happen in production, only on previews.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PaywallDataExtensions.kt
```diff
@@ -137,15 +136,39 @@ internal fun DefaultPaywallPreview() {
...
+@Preview(showBackground = true, locale = "es-rES")
+@Composable
+internal fun DefaultPaywallPreview() {
+    InternalPaywall(
+        options = PaywallOptions.Builder(dismissRequest = {}).build(),
+        viewModel = MockViewModel(offering = getDefaultPreviewOffering()),
+    )
+}
+
+@Preview(showBackground = true, locale = "en-rUS", group = "footer")
```
vegaro, line 158: do we also want to add `es-rES`?

vegaro (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PaywallDataExtensions.kt
tonidero: I thought about it, but we already have one for the full screen version, and I don't think it provides much value to have one for the footer? Lmk if you think otherwise and happy to add it though.

PR #1844, opened 2024-09-16
Add `kochava` integration

tonidero:
### Description
This adds the `kochava` integration.

This brings back the work done in #1829 and picks up the work on #1837 (but renaming it to `$deviceVersion`). Finally, it also adds a new header, `X-Platform-Device` that will contain the model version

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -266,6 +266,7 @@ internal class HTTPClient(
             "X-Platform-Flavor" to appConfig.platformInfo.flavor,
             "X-Platform-Flavor-Version" to appConfig.platformInfo.version,
             "X-Platform-Version" to Build.VERSION.SDK_INT.toString(),
+            "X-Platform-Device" to Build.MODEL,
             "X-Version" to Config.frameworkVersion,
             "X-Client-Locale" to appConfig.languageTag,
             "X-Client-Version" to appConfig.versionName,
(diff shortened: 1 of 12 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -266,6 +266,7 @@ internal class HTTPClient(
             "X-Platform-Flavor" to appConfig.platformInfo.flavor,
             "X-Platform-Flavor-Version" to appConfig.platformInfo.version,
             "X-Platform-Version" to Build.VERSION.SDK_INT.toString(),
+            "X-Platform-Device" to Build.MODEL,
```
JayShortway, line 269: Would adding [`Build.BRAND`](https://developer.android.com/reference/android/os/Build#BRAND) be useful here as well? Otherwise we might only see something like "SM-S921E" without knowing it's a Samsung (Galaxy S24 in this case). 

(Samsung's model numbers are quite recognizable, but others might be less so.)

JayShortway (approved):
Good stuff! Just a single suggestion, not blocking.

david-berliner (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Hmm could be useful yeah... I'm just reticent to send it in the same header, so the backend doesn't need to process the string. iOS wouldn't need a similar header, but we can potentially send it in case it's useful in the future. Will add it as a separate header 👍

tonidero:
We've now tested the integration so merging!

Issue #1842, opened 2024-09-14
Crash in Android When launching the paywall


arshavin2000:
- [ ] I have updated Purchases SDK to the latest version 
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
I am implementing Revenue cat in my Android app. It works perfectly and everything is ok for most of users but lately I am receiving this crash on crashlytics for some users and I was unable to reproduce the crash on my devices

`java.lang.IllegalStateException: Attempting to launch an unregistered ActivityResultLauncher with contract com.revenuecat.purchases.ui.revenuecatui.activity.PaywallContract@bae055e and input PaywallActivityArgs(requiredEntitlementIdentifier=flixunblocker_v5, offeringId=flixunblocker_offer_v5, fonts=null, shouldDisplayDismissButton=true). You must ensure the ActivityResultLauncher is registered before calling launch().`

1. Environment
   1. Platform: Android target SDK 44 
   2. SDK version: 8.4.0
   3. OS version: Mac OS
   4. Android Studio version: Android Studio Giraffe | 2022.3.1 Patch 1
   5. How widespread is the issue. Percentage of devices affected.  28 users were affected by this crash according to my Crashlytics report

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/46020) and notified the team. Thank you for reporting, we're checking this out!

tonidero:
Hi @arshavin2000, thanks for reaching out!

Have you been able to reproduce the issue to provide some retro steps? In case not, could you send a snippet of how with how you're creating and presenting the paywall? This might help debugging the issue. Also, OS version distribution of the crash could be useful.

From that error message, it seems like in some situations you might be trying to use the `PaywallActivityLauncher` without creating it first during the `onCreate` phase of your activity... Check the paywall docs on how to display the paywall in an activity here: https://www.revenuecat.com/docs/tools/paywalls/displaying-paywalls#how-to-display-a-fullscreen-paywall-in-your-app-1

arshavin2000:
[...]

PR #1846, opened 2024-09-18
Fix crash if activity finished while calculating presentation logic

tonidero:
### Description
We got a couple reports of a similar issue:
- https://github.com/RevenueCat/purchases-android/issues/1842
- https://community.revenuecat.com/sdks-51/crash-in-kotlin-4999

Seems that those happened mostly on android TV. While I haven't been able to reproduce normally, I was able to reproduce (with the same stacktrace) by adding an artificial delay between the obtention of the customer info and presenting the paywall when `requiredEntitlementIdentifier` is shown and killing the activity.

This is a possible approach to fixing this.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
@@ -34,6 +34,7 @@ interface PaywallDisplayCallback {
  */
 class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler: PaywallResultHandler) {
     private val activityResultLauncher: ActivityResultLauncher<PaywallActivityArgs>
+
     // We need to know whether the activity is running or finished to avoid launching the paywall
     // after the activity has been destroyed. See https://github.com/RevenueCat/purchases-android/issues/1842.
     // We keep a weak reference to avoid memory leaks.
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.revenuecatui.activity
 
+import android.app.Activity
 import androidx.activity.ComponentActivity
 import androidx.activity.result.ActivityResultCallback
 import androidx.activity.result.ActivityResultCaller
@@ -10,6 +11,7 @@ import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.ui.revenuecatui.fonts.ParcelizableFontProvider
 import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayBlockForEntitlementIdentifier
 import com.revenuecat.purchases.ui.revenuecatui.helpers.shouldDisplayPaywall
+import java.lang.ref.WeakReference
 
 /**
  * Implement this interface to receive the result of the paywall activity.
@@ -31,6 +33,8 @@ interface PaywallDisplayCallback {
  */
 class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler: PaywallResultHandler) {
     private val activityResultLauncher: ActivityResultLauncher<PaywallActivityArgs>
+    private val weakActivity = WeakReference(resultCaller as? Activity)
+    private val weakFragment = WeakReference(resultCaller as? Fragment)
 
     init {
         activityResultLauncher = resultCaller.registerForActivityResult(PaywallContract(), resultHandler)
@@ -105,7 +109,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         val shouldDisplayBlock = shouldDisplayBlockForEntitlementIdentifier(requiredEntitlementIdentifier)
         shouldDisplayPaywall(shouldDisplayBlock) { shouldDisplay ->
             paywallDisplayCallback?.onPaywallDisplayResult(shouldDisplay)
-            if (shouldDisplay) {
+            if (shouldDisplay && !isActivityFinishing()) {
                 activityResultLauncher.launch(
                     PaywallActivityArgs(
                         requiredEntitlementIdentifier = requiredEntitlementIdentifier,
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -31,6 +33,7 @@ interface PaywallDisplayCallback {
  */
 class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler: PaywallResultHandler) {
     private val activityResultLauncher: ActivityResultLauncher<PaywallActivityArgs>
+    private val weakActivity = WeakReference(resultCaller as? Activity)
```
tonidero, line 36: Honestly, I believe this is quite hacky... But I wasn't able to figure out how to obtain a reference to the activity otherwise that would let us check if it was finishing from here... Open to ideas.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -31,6 +33,8 @@ interface PaywallDisplayCallback {
...
+    private val weakFragment = WeakReference(resultCaller as? Fragment)
```
tonidero, line 37: Honestly, I believe this is very hacky... But I wasn't able to figure out how to obtain a reference to the activity otherwise that would let us check if it was finishing from here... 

This should work most times since usually the `ActivityResultCaller` is normally the activity or a fragment 

I was thinking an alternative would be to remove this `PaywallActivityLauncher`, expose the `PaywallActivity` and have the developer manage the presentation/result but I don't really like that. Open to ideas...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -105,7 +109,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         val shouldDisplayBlock = shouldDisplayBlockForEntitlementIdentifier(requiredEntitlementIdentifier)
         shouldDisplayPaywall(shouldDisplayBlock) { shouldDisplay ->
             paywallDisplayCallback?.onPaywallDisplayResult(shouldDisplay)
-            if (shouldDisplay) {
+            if (shouldDisplay && !isActivityFinishing()) {
                 activityResultLauncher.launch(
```
tonidero, line 113: This is what would crash if the original activity is finished by the time it gets here.

JayShortway: Would it be helpful to print a warning, or maybe even error, log to indicate why the Paywall didn't launch?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
JayShortway: I don't think this is _that_ bad to be honest. We have a requirement to check whether the Activity is still alive, but if we keep a hard reference we cause a memory leak. I think `WeakReference` is the only option.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -183,4 +187,10 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
             }
         }
     }
+
+    private fun isActivityFinishing(): Boolean {
+        val activity = weakActivity.get()
+        val fragment = weakFragment.get()
+        return activity?.isFinishing == true || fragment?.activity?.isFinishing == true
+    }
```
JayShortway, line 195: Maybe this should also return true if the Activity is completely dereferenced? That is, if `activity` or `fragment.activity` is null (but not if only the `fragment` is null imo). What do you think?

JayShortway (commented):
I think this is an improvement in safety! Added 2 suggestions.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Yeah absolutely! Fixing this, thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Good suggestion! Will add something

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Ah I remember now... The reason I didn't do that was because a developer could potentially implement their own `ActivityResultCaller` which is not an activity or a fragment (honestly I don't know of any valid use cases, but they could do this), and we would still want to display the paywall there, or it could potentially cause issues in that case... 

I think we should leave as is, and if we don't have an activity to check, just continue as we do right now. Lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/1846/commits/20d9ad3b23d3dc736d6d288e5e76401bcb126fab

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
aboedo: kinda reminds me of this TBH https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L73-L78, I don't think it's terrible as long as its under our control. 

But definitely add comments here so future you doesn't hate present you for doing this and not explaining why

aboedo (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
JayShortway: I agree that this is probably good enough. If we really want to catch the case where the `ActivityResultCaller` is neither an Activity nor a Fragment, we can store that info when we assign the WeakReferences. Not a blocker for me.

JayShortway (approved):

Issue #1849, opened 2024-09-20
Just like on the iOS system, I cannot subscribe to other plans while processing an order.


YiLin-East:
On Android devices, when I click to subscribe to a plan, the "PaywallActivity" does not close immediately after the subscription is successful. During this time, I can subscribe to other plans again. On iOS, this period is not interactive. I hope the interaction experience on Android can be the same as on iOS.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/46268) and notified the team. Thank you for reporting, we're checking this out!

tonidero:
Hi @YiLin-East, thanks for reporting! 

This was indeed a regression in the last couple SDK versions. We weren't disabling interaction while a purchase was in progress. I just worked on a fix in #1850 and should be released in the next SDK version.

Thanks again for reporting!

YiLin-East:
![good](https://github.com/user-attachments/assets/9db54306-e8c4-4048-9375-3dc26edfbb75)

PR #1850, opened 2024-09-20
Fix interaction not disabled during purchases

tonidero:
### Description
This was a regression in #1777, and was reported in #1849 

We weren't marking as an action in progress when performing purchases.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -152,8 +152,12 @@ internal class PaywallViewModelImpl(
             Logger.e("Activity is null, not initiating package purchase")
             return
         }
+        if (verifyNoActionInProgressOrStartAction()) {
+            return
+        }
```
tonidero, line 157: I thought about moving these inside the internal method, but no sense to perform extra operations without need IMO.

JayShortway (approved):
Nice!

aboedo:
thanks for taking care of this!

PR #1853, opened 2024-09-24
Adds some more test cases validating {{ total_price_and_per_month }} for quarterly packages. 

JayShortway:
As the title says. When investigating an issue I wanted to eliminate the possibility that our SDK is somehow processing these incorrectly. Turns out it isn't. Can't hurt to add these tests to avoid regressions I think?


tonidero (approved):
Should be fine yup 👍

PR #1854, opened 2024-09-24
Fix multiple paywalls display issues

tonidero:
### Description
This is an approach to fix DENG-955

If trying to display multiple paywalls at the same time, all paywalls would currently share the same view model. This caused that all of them would be exactly the same, which was not great for some use cases. This adds a key to the view model injection using the paywall options view model hash code. This means we will have one view model for each combination of paywall options parameters.

image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -175,6 +175,7 @@ internal fun getPaywallViewModel(
 ): PaywallViewModel {
     val applicationContext = LocalContext.current.applicationContext
     val viewModel = viewModel<PaywallViewModelImpl>(
+        key = options.dataHash,
         factory = PaywallViewModelFactory(
             applicationContext.toResourceProvider(),
             options,
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
@@ -46,6 +46,9 @@ data class PaywallOptions internal constructor(
         dismissRequest = builder.dismissRequest,
     )
 
+    // This hash is used to determine if the paywall should use a different view model.
+    // Not using hashCode/equals because the listener may change in some rerenders and we don't want to change
+    // the view model in those cases.
     internal val dataHash: String = run {
         var result = offeringSelection.offeringIdentifier.hashCode()
         result = hashMultiplier * result + shouldDisplayDismissButton.hashCode()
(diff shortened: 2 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -175,6 +175,7 @@ internal fun getPaywallViewModel(
 ): PaywallViewModel {
     val applicationContext = LocalContext.current.applicationContext
     val viewModel = viewModel<PaywallViewModelImpl>(
+        key = options.hashCode().toString(),
```
tonidero, line 178: This is the main change. With this we can have multiple view models living at the same time.

But more importantly, this is not new but the view model isn't currently cleared when using composables or views and dismissing the paywall, since the view model owner (the activity) isn't cleared. With the increase in the possible number of view models, this means we can potentially increase our memory usage if the paywall is displayed with multiple different options. This should normally not be a problem, but something to consider...

To solve this, it might be tricky to scope the view model to a screen though, since we don't have access to the navigation graph... Any ideas?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -43,6 +43,26 @@ data class PaywallOptions internal constructor(
         dismissRequest = builder.dismissRequest,
     )
 
+    override fun hashCode(): Int {
+        var result = offeringSelection.offeringIdentifier.hashCode()
+        result = 31 * result + shouldDisplayDismissButton.hashCode()
+        result = 31 * result + mode.hashCode()
```
tonidero, line 49: Right now only using the offeringId+shouldDisplayDismissButton+mode as identifier of the paywall. The others are objects with callbacks which would be much trickier to compare.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -43,6 +43,13 @@ data class PaywallOptions internal constructor(
         dismissRequest = builder.dismissRequest,
     )
 
+    internal val dataHash: String = run {
```
tonidero, line 46: Note that I'm not overriding the `hashCode`/`equals` functions here since we are comparing them (including the listeners) for other methods, like the `updateOptions` method. This was the least disruptive approach that didn't mean recreating the view model for every recreation of the listeners.

aboedo: that makes sense, but let's add a comment so future our future selves understand why we have `dataHash` in the first place

aboedo (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
tonidero: Yes, if using jetpack navigation. So we should be mostly ok in that case. However, it's not guaranteed for devs to be using that library... But yeah, it further constraints this limitation 👍

JayShortway: Yea you're right. It's a good limitation to be aware of!

PR #1859, opened 2024-09-27
[CustomerCenter] Add networking layer and models

tonidero:
### Description
- Adds networking layer and models to `RevenueCat` to support the new CustomerCenter.
- Adds a new `X-Preferred-Locales` header.
- Adds a new `Purchases.getCustomerCenterConfigData` method to access the data from `RevenueCatUI`

### DIFF
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
@@ -15,9 +15,7 @@ import com.revenuecat.purchases.PurchasesConfiguration;
 import com.revenuecat.purchases.PurchasesError;
 import com.revenuecat.purchases.Store;
 import com.revenuecat.purchases.amazon.AmazonConfiguration;
-import com.revenuecat.purchases.customercenter.CustomerCenterConfigData;
 import com.revenuecat.purchases.interfaces.GetAmazonLWAConsentStatusCallback;
-import com.revenuecat.purchases.interfaces.GetCustomerCenterConfigCallback;
 import com.revenuecat.purchases.interfaces.LogInCallback;
 import com.revenuecat.purchases.interfaces.ReceiveCustomerInfoCallback;
 import com.revenuecat.purchases.interfaces.SyncAttributesAndOfferingsCallback;
@@ -78,16 +76,6 @@ final class PurchasesAPI {
             }
         };
 
-        final GetCustomerCenterConfigCallback getCustomerCenterConfigCallback = new GetCustomerCenterConfigCallback() {
-            @Override
-            public void onError(@NonNull PurchasesError error) {
-            }
-
-            @Override
-            public void onSuccess(@NonNull CustomerCenterConfigData customerCenterConfig) {
-            }
-        };
-
         purchases.syncAttributesAndOfferingsIfNeeded(syncAttributesAndOfferingsCallback);
         purchases.syncPurchases();
         purchases.syncPurchases(syncPurchasesCallback);
@@ -97,7 +85,6 @@ final class PurchasesAPI {
         purchases.getCustomerInfo(receiveCustomerInfoListener);
         purchases.getCustomerInfo(CacheFetchPolicy.CACHED_OR_FETCHED, receiveCustomerInfoListener);
         purchases.getAmazonLWAConsentStatus(getAmazonLWAContentStatusCallback);
-        purchases.getCustomerCenterConfigData(getCustomerCenterConfigCallback);
 
         purchases.restorePurchases(receiveCustomerInfoListener);
         purchases.invalidateCustomerInfoCache();
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
@@ -15,7 +15,9 @@ import com.revenuecat.purchases.PurchasesConfiguration;
 import com.revenuecat.purchases.PurchasesError;
 import com.revenuecat.purchases.Store;
 import com.revenuecat.purchases.amazon.AmazonConfiguration;
+import com.revenuecat.purchases.customercenter.CustomerCenterConfigData;
 import com.revenuecat.purchases.interfaces.GetAmazonLWAConsentStatusCallback;
+import com.revenuecat.purchases.interfaces.GetCustomerCenterConfigCallback;
 import com.revenuecat.purchases.interfaces.LogInCallback;
 import com.revenuecat.purchases.interfaces.ReceiveCustomerInfoCallback;
 import com.revenuecat.purchases.interfaces.SyncAttributesAndOfferingsCallback;
@@ -76,6 +78,16 @@ final class PurchasesAPI {
             }
         };
 
+        final GetCustomerCenterConfigCallback getCustomerCenterConfigCallback = new GetCustomerCenterConfigCallback() {
+            @Override
+            public void onError(@NonNull PurchasesError error) {
+            }
+
+            @Override
+            public void onSuccess(@NonNull CustomerCenterConfigData customerCenterConfig) {
+            }
+        };
+
         purchases.syncAttributesAndOfferingsIfNeeded(syncAttributesAndOfferingsCallback);
         purchases.syncPurchases();
         purchases.syncPurchases(syncPurchasesCallback);
@@ -85,6 +97,7 @@ final class PurchasesAPI {
         purchases.getCustomerInfo(receiveCustomerInfoListener);
         purchases.getCustomerInfo(CacheFetchPolicy.CACHED_OR_FETCHED, receiveCustomerInfoListener);
         purchases.getAmazonLWAConsentStatus(getAmazonLWAContentStatusCallback);
+        purchases.getCustomerCenterConfigData(getCustomerCenterConfigCallback);
 
         purchases.restorePurchases(receiveCustomerInfoListener);
         purchases.invalidateCustomerInfoCache();
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
@@ -3,6 +3,7 @@ package com.revenuecat.purchases
 import com.revenuecat.purchases.CacheFetchPolicy.CACHED_OR_FETCHED
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
 import com.revenuecat.purchases.data.LogInResult
+import com.revenuecat.purchases.interfaces.GetCustomerCenterConfigCallback
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
@@ -160,9 +161,14 @@ suspend fun Purchases.getAmazonLWAConsentStatus(): AmazonLWAConsentStatus {
 @Throws(PurchasesException::class)
 suspend fun Purchases.awaitCustomerCenterConfigData(): CustomerCenterConfigData {
     return suspendCoroutine { continuation ->
-        getCustomerCenterConfigDataWith(
-            onError = { continuation.resumeWithException(PurchasesException(it)) },
-            onSuccess = continuation::resume,
-        )
+        getCustomerCenterConfigData(object : GetCustomerCenterConfigCallback {
+            override fun onSuccess(customerCenterConfig: CustomerCenterConfigData) {
+                continuation.resume(customerCenterConfig)
+            }
+
+            override fun onError(error: PurchasesError) {
+                continuation.resumeWithException(PurchasesException(error))
+            }
+        })
     }
 }
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
@@ -159,6 +159,7 @@ suspend fun Purchases.getAmazonLWAConsentStatus(): AmazonLWAConsentStatus {
  */
 @JvmSynthetic
 @Throws(PurchasesException::class)
+@ExperimentalPreviewRevenueCatPurchasesAPI
 suspend fun Purchases.awaitCustomerCenterConfigData(): CustomerCenterConfigData {
     return suspendCoroutine { continuation ->
         getCustomerCenterConfigData(object : GetCustomerCenterConfigCallback {
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases
 
 import com.revenuecat.purchases.CacheFetchPolicy.CACHED_OR_FETCHED
+import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
 import com.revenuecat.purchases.data.LogInResult
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
@@ -103,7 +104,7 @@ suspend fun Purchases.awaitSyncPurchases(): CustomerInfo {
  * This method is rate limited to 5 calls per minute. It will log a warning and return offerings cache when reached.
  *
  * Refer to [the guide](https://www.revenuecat.com/docs/tools/targeting) for more targeting information
- * For more offerings information, see [getOfferings]
+ * For more offerings information, see [Purchases.getOfferings]
  *
  * Coroutine friendly version of [Purchases.syncAttributesAndOfferingsIfNeeded].
  *
@@ -144,6 +145,24 @@ suspend fun Purchases.getAmazonLWAConsentStatus(): AmazonLWAConsentStatus {
         getAmazonLWAConsentStatusWith(
             onSuccess = continuation::resume,
             onError = { continuation.resumeWithException(PurchasesException(it)) },
+        )
+    }
+}
+
+/**
+ * Gets the current user's [CustomerCenterConfigData]. Used by RevenueCatUI to present the customer center.
+ *
+ * @throws [PurchasesException] with the first [PurchasesError] if there's an error getting the customer center
+ * config data.
+ * @returns The [CustomerCenterConfigData] for the current user.
+ */
+@JvmSynthetic
+@Throws(PurchasesException::class)
+suspend fun Purchases.awaitCustomerCenterConfigData(): CustomerCenterConfigData {
+    return suspendCoroutine { continuation ->
+        getCustomerCenterConfigDataWith(
+            onError = { continuation.resumeWithException(PurchasesException(it)) },
+            onSuccess = continuation::resume,
         )
     }
 }
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
@@ -1,7 +1,9 @@
 package com.revenuecat.purchases
 
 import android.app.Activity
+import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
 import com.revenuecat.purchases.interfaces.GetAmazonLWAConsentStatusCallback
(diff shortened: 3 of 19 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    @Serializable
+    class HelpPath(
+        val id: String,
+        val title: String,
+        val type: PathType,
+        @SerialName("promotional_offer") val promotionalOffer: PathDetail.PromotionalOffer? = null,
+        @SerialName("feedback_survey") val feedbackSurvey: PathDetail.FeedbackSurvey? = null,
```
tonidero, line 170: Note that this is a bit different to iOS. In iOS, there is a single `detail` field, which makes sense since you can only have either `promotional_offer` or `feedback_survey`. But that was trickier to deserialize and I believe it's fine to separate these 2 into separate fields. Lmk if you have concerns.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,260 @@
...
+    ) {
+        @Serializable
+        sealed class PathDetail {
+            @Serializable
+            class PromotionalOffer(
+                // WIP: Rename this field name to a new android id
+                @SerialName("ios_offer_id") val androidOfferId: String,
```
tonidero, line 177: Right now I'm reading the `ios_offer_id` but we should read the android one before merging this.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,260 @@
+package com.revenuecat.purchases.customercenter
+
+import com.revenuecat.purchases.paywalls.PaywallColor
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+typealias RCColor = PaywallColor
```
tonidero, line 7: Not a big fun of these typealiases, but it's how we have it in iOS, and it kinda made sense now.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/ScreenMapSerializer.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.revenuecat.purchases.customercenter.CustomerCenterConfigData.Screen.ScreenType
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.builtins.MapSerializer
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.jsonObject
+
+internal object ScreenMapSerializer : KSerializer<Map<ScreenType, Screen>> {
```
tonidero, line 14: Had to write a custom deserializer here for the map of screen types to screen, to make sure if we start receiving unknown screen types in the future, we still can parse the known ones.

purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
```diff
@@ -240,6 +255,16 @@ fun Purchases.getAmazonLWAConsentStatusWith(
     getAmazonLWAConsentStatus(getAmazonLWAConsentStatusListener(onSuccess, onError))
 }
 
+/**
+ * Gets the [CustomerCenterConfigData] for the current user. Used by RevenueCatUI to present the CustomerCenter.
+ */
+fun Purchases.getCustomerCenterConfigDataWith(
```
tonidero, line 261: This is meant to be used internally, so maybe it's not needed to create these listener and async versions of this API... But I thought I might as well add them for consistency. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,260 @@
...
+
+@Serializable
+internal class CustomerCenterRoot(
+    @SerialName("customer_center") val customerCenter: CustomerCenterConfigData,
+)
+
+@Serializable
+class CustomerCenterConfigData(
```
tonidero, line 15: Need to write API tests for all these.

tonidero, line 15: Also note that the constructors here are public, so test data can be created more easily from within the `RevenueCatUI` module.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,260 @@
...
+                        "You can manage your subscription by using the App Store app on an Apple device."
+                    GOOGLE_SUBSCRIPTION_MANAGE ->
+                        "You can manage your subscription by using the Play Store app on an Android device"
+                    AMAZON_SUBSCRIPTION_MANAGE ->
+                        "You can manage your subscription in the Amazon Appstore app on an Amazon device."
+                }
+        }
+
+        fun commonLocalizedString(key: CommonLocalizedString): String {
+            return localizedStrings[key.name.lowercase()] ?: key.defaultValue
```
tonidero, line 160: I didn't fully like this... It means the enum name must always match the key in uppercase. Want to write a test to make sure this is setup correctly.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,468 @@
...
+        }
+
+        override fun equals(other: Any?): Boolean {
```
tonidero, line 165: Had to add all these equals because I'm not using data classes :(, I honestly don't know what's best, since we now need to remember to maintain them...

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
tonidero: Test written in https://github.com/RevenueCat/purchases-android/pull/1859/commits/8cb3f5f814524f0a1e0f8b5b633be516005476dc

tonidero:
I expect the new backend integration test to fail until we've given the project access to the customer center. It worked fine for a different project I tested on.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
JayShortway: Maybe we can just use data classes here? These are not _really_ public API classes. 

Data classes are primarily bad for binary compatibility. Any library depending on RevenueCat should not be calling `Purchases.getCustomerCenterConfigData()` anyway. (If we're worried about that, we could add a note to that function saying it's not guaranteed to be binary compatible.)

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/ScreenMapSerializer.kt
JayShortway: Great foresight! Does the builtin deserializer fail even with `ignoreUnknownKeys`?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,465 @@
...
+    @Serializable(with = ScreenMapSerializer::class) val screens: Map<Screen.ScreenType, Screen>,
+    val appearance: Appearance,
+    val localization: Localization,
+    val support: Support,
+    @SerialName("last_published_app_version") val lastPublishedAppVersion: String? = null,
+) {
+    @Serializable
+    class Localization(
+        val locale: String,
+        @SerialName("localized_strings") val localizedStrings: Map<String, String>,
```
JayShortway, line 27: I guess if we add a default map here containing every enum with its default value, it will be entirely replaced if the response contains a `localized_strings` map, right? Would be great if the maps would be merged, giving the map in the response precedence.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
JayShortway: In this case as well I wouldn't be opposed to having the enum just match the JSON key. (Testing would still be good, but the "rule" is simpler.)

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
JayShortway: For my understanding, in Android we could have both?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,465 @@
...
+            REFUND_REQUEST,
+
+            @SerialName("CHANGE_PLANS")
+            CHANGE_PLANS,
+
+            @SerialName("CANCEL")
+            CANCEL,
+
+            @SerialName("UNKNOWN")
+            UNKNOWN,
```
JayShortway, line 292: Are these uppercase in the JSON? If so, these `@SerialName`s are redundant right?

purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
JayShortway: Or we could make `Purchases.getCustomerCenterConfigData` exactly how we want it (e.g. suspend or function parameters), as we don't need to care about Java compatibility since `RevenueCatUI` is Kotlin.

api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
```diff
@@ -85,6 +97,7 @@ public void onSuccess(@NonNull AmazonLWAConsentStatus contentStatus) {
         purchases.getCustomerInfo(receiveCustomerInfoListener);
         purchases.getCustomerInfo(CacheFetchPolicy.CACHED_OR_FETCHED, receiveCustomerInfoListener);
         purchases.getAmazonLWAConsentStatus(getAmazonLWAContentStatusCallback);
+        purchases.getCustomerCenterConfigData(getCustomerCenterConfigCallback);
```
JayShortway, line 100: Related to a previous comment, it can be argued that this is not needed, as it isn't intended to be a public API.

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionBackendIntegrationTest.kt
```diff
@@ -198,4 +200,34 @@ internal class ProductionBackendIntegrationTest: BaseBackendIntegrationTest() {
...
+                },
+                onErrorHandler = { error ->
+                    fail("Expected success. Got $error")
+                }
+            )
+        }
+        if (customerCenterConfigData == null) {
+            error("Expected customer center config data")
+        }
+        customerCenterConfigData?.let { customerCenterConfigData->
```
JayShortway, line 222: There is a lint warning here about shadowing the `customerCenterConfigData` name.

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetCustomerCenterConfigTest.kt
```diff
@@ -0,0 +1,315 @@
...
+    private val mockBaseURL = URL("http://mock-api-test.revenuecat.com/")
+    private val MOCK_RESPONSE_FILENAME = "get_customer_center_config_success.json"
+
+    private lateinit var appConfig: AppConfig
+    private lateinit var httpClient: HTTPClient
+
+    private lateinit var backend: Backend
+    private lateinit var asyncBackend: Backend
+
+    private val expectedCustomerCenterConfigData = CustomerCenterConfigData(
```
JayShortway, line 48: I have never used it, but [Kotest's property test framework](https://kotest.io/docs/proptest/property-based-testing.html) can supposedly help with generating these test fixtures. (Not for this PR of course, just something we could look into at some point.)

JayShortway (commented):
Nice, I think this is close! I just had some minor comments and suggestions.

purchases/src/defaults/kotlin/com/revenuecat/purchases/listenerConversions.kt
tonidero: Hmm right... FWIW, another reason to make these public is in case devs want to use the customer center config data themselves (same with paywall data within `Offering`)... But yeah, this seems like a less likely case for this to be useful... Will remove those and we can always add those later if needed 👍

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
tonidero: Yeah... I think I might do that since maintaining all these equals is a bit annoying 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -580,6 +595,64 @@ internal class Backend(
...
+                    postFieldsToSign = null,
+                    backendHelper.authenticationHeaders,
+                )
+            }
+
+            override fun onError(error: PurchasesError) {
+                synchronized(this@Backend) {
+                    customerCenterCallbacks.remove(path)
+                }?.forEach { (_, onErrorHandler) ->
+                    onErrorHandler(error)
```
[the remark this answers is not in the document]
tonidero: I believe most of the `onError` callbacks are already logged inside the [AsyncCall](https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt#L47). But yeah, it's a bit mix and match currently... I think we need to refactor the `Backend` class at some point since it's getting pretty big and there is quite a bit of duplication, but not for this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -0,0 +1,465 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+typealias RCColor = PaywallColor
+
+private const val HASH_CODE_MULTIPLIER = 31
+
+@Serializable
+internal class CustomerCenterRoot(
+    @SerialName("customer_center") val customerCenter: CustomerCenterConfigData,
```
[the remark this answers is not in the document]
tonidero: Hmm since these classes are still public, I would try to keep convention and use kotlin syntax... I think it would look weird in the usage as well... I would prefer to have an intermediate "network model" layer and map those to different models to expose them to the rest of the SDK, but then I feel we would be adding a lot of overhead IMO... Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/ScreenMapSerializer.kt
tonidero: Right, if there was a new screen, like `TEST`, without this code it would fail with `com.revenuecat.purchases.customercenter.CustomerCenterConfigData.Screen.ScreenType does not contain element with name 'TEST' at path $.customer_center.screens`

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
tonidero: >  it will be entirely replaced if the response contains a localized_strings map, right?

That's correct.

> Would be great if the maps would be merged, giving the map in the response precedence.

Yeah that would be good... Maybe we can write a custom serializer to do that...

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
tonidero: Not right now since the flow wouldn't make much sense if both are present... Having said that, we're not currently constraining this in the backend, so it's possible to have both.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
tonidero: Actually, after thinking about it, I think it might be this way to be able to access strings that are not part of the enum? I don't see a reason why we should allow that, but it's something that would be limited if we make this map a map of `CommonLocalizedString` to `String`... I think I might leave it as is to keep the same API as in iOS. cc @vegaro

vegaro: hmmm, that's a good point, at the end of the day these are just common strings for now by definition. We could at some point return more arbitrary strings that are arbitrary, and don't have a default. I can't think of a usage for that now but might be better to keep that flexibility. The cool thing about making it `CommonLocalizedString` to `String` is that the map will only contain strings that are used by that version of the SDK.

I think it's fine to just do what we do in iOS already.

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/ScreenMapSerializer.kt
JayShortway: Oh right, because of the enum. Makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
JayShortway: Alright, yea I think these models should reflect the backend as closely as possible, so this is fine with me.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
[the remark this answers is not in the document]
JayShortway: Alright! Yes a quick win could be to have a (private) subclass of the `AsyncCall` that can be reused within the `Backend` class, to decrease some of the repetition. But indeed, not for this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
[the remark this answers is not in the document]
JayShortway: >I would prefer to have an intermediate "network model" layer and map those to different models to expose them to the rest of the SDK

Oh right, yes, since we are using these network models directly, it's indeed good to keep the naming idiomatic Kotlin. I agree that mapping them to different models is probably not worth it.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
JayShortway: I'm fine with doing what we do in iOS, yes. Let's not write a custom serializer just for this. Was just musing 😇

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -159,6 +159,7 @@ suspend fun Purchases.getAmazonLWAConsentStatus(): AmazonLWAConsentStatus {
  */
 @JvmSynthetic
 @Throws(PurchasesException::class)
+@ExperimentalPreviewRevenueCatPurchasesAPI
```
vegaro, line 162: ❤️

PR #1863, opened 2024-10-01
[CustomerCenter] Adds SubscriptionDetailsView

JayShortway:
## In this PR

image

- A new `SubscriptionDetailsView` ☝️, modelled after the iOS namesake. (The black border is Android Studio's preview UI.)
- Slightly tweaked the `TopLevelPropertyNaming` Detekt rule to allow CamelCase constants, following the Compose guidelines. I modified [Detekt's recommendation](https://detekt.dev/docs/introduction/compose#toplevelpropertynaming-for-compose) to also allow SCREAMING_SNAKE_CASE, as we use that in other parts.

## Not in this PR
- Any logic to display the new `SubscriptionDetailsView`.

I was thinking we can update the design as we go, when we have a better idea of what this is going to look like, but let me know what you think!

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/CalendarMonth.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/CalendarMonth.kt
@@ -1,0 +1,133 @@
+@file:Suppress("MagicNumber", "ObjectPropertyName")
+
+package com.revenuecat.purchases.ui.revenuecatui.icons
+
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.PathFillType
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.StrokeCap
+import androidx.compose.ui.graphics.StrokeJoin
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.graphics.vector.path
+import androidx.compose.ui.unit.dp
+
+/**
+ * Exported from [Compose Icons](https://composeicons.com/icons/material-symbols/outlined/calendar_month).
+ */
+internal val CalendarMonth: ImageVector
+    get() {
+        if (_Calendar_month != null) {
+            return _Calendar_month!!
+        }
+        _Calendar_month = ImageVector.Builder(
+            name = "CalendarMonth",
+            defaultWidth = 24.dp,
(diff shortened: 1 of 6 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/CalendarMonth.kt
```diff
@@ -0,0 +1,133 @@
...
+import androidx.compose.ui.graphics.PathFillType
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.StrokeCap
+import androidx.compose.ui.graphics.StrokeJoin
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.graphics.vector.path
+import androidx.compose.ui.unit.dp
+
+/**
+ * Exported from [Compose Icons](https://composeicons.com/icons/material-symbols/outlined/calendar_month).
```
JayShortway, line 15: I decided against including the `androidx.compose.material:material-icons-extended` library, as that one is really large and we're not minifying it properly. Let me know what you think.

tonidero: Hmm I think we would be minifying it? As in, we are only not minifying our SDKs, but not our dependencies I believe... But yeah, if we only need these images, it might be ok to just have them here.

tonidero (approved):
Makes sense to me! As you said, we can iterate on that later if needed.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/CalendarMonth.kt
JayShortway: Ah you're right! No strong opinion on this. I guess a tiny advantage of not including the library is that we don't increase the size of developers' debug builds.

tonidero: Yeah let's do this for now and we can always revisit later if needed.

JayShortway: Yea, agreed!

PR #1866, opened 2024-10-04
Avoid recreating purchases same config

tonidero:
### Description
This came from a conversation in Slack. I think we can limit the amount of multiple configuration issues if we don't reconfigure the SDK with the same parameters that were originally passed. I don't think there should be any use cases for that in any case? Keeping on hold to get feedback


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -36,7 +36,7 @@ open class PurchasesConfiguration(builder: Builder) {
     val pendingTransactionsForPrepaidPlansEnabled: Boolean
 
     init {
-        this.context = builder.context
+        this.context = builder.context.applicationContext
```
tonidero, line 39: I noticed we could potentially be leaking an activity if the developer gave us the activity context and we hold a reference to this object. So I changed it here so we only hold the application context, which is the one developers should normally give us.

tonidero:
Closing in favor of #1872

PR #1868, opened 2024-10-04
[CustomerCenter] Add `CustomerCenterBottomSheet` 

tonidero:
### Description
This adds a new composable `CustomerCenterBottomSheet` as an entry point for the customer center. This basically wraps the `CustomerCenter` composable into a bottom sheet, for convenience.
Additionally, it has some commented out code to display this bottom sheet in the paywall tester app. Need to make public the composables and uncomment to test it.

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.material3.Button
+import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.OutlinedTextField
 import androidx.compose.material3.Surface
@@ -28,12 +29,15 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.Dialog
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.revenuecat.purchases.ui.debugview.DebugRevenueCatBottomSheet
+import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
 @Composable
 fun AppInfoScreen(viewModel: AppInfoScreenViewModel = viewModel<AppInfoScreenViewModelImpl>()) {
     var isDebugBottomSheetVisible by remember { mutableStateOf(false) }
+    var isCustomerCenterBottomSheetVisible by remember { mutableStateOf(false) }
     var showLogInDialog by remember { mutableStateOf(false) }
 
     Column(
@@ -52,6 +56,9 @@ fun AppInfoScreen(viewModel: AppInfoScreenViewModel = viewModel<AppInfoScreenVie
         Button(onClick = { isDebugBottomSheetVisible = true }) {
             Text(text = "Show debug view")
         }
+        Button(onClick = { isCustomerCenterBottomSheetVisible = true }) {
+            Text(text = "Show customer center bottom sheet")
+        }
     }
 
     if (showLogInDialog) {
@@ -64,6 +71,16 @@ fun AppInfoScreen(viewModel: AppInfoScreenViewModel = viewModel<AppInfoScreenVie
         isVisible = isDebugBottomSheetVisible,
         onDismissCallback = { isDebugBottomSheetVisible = false },
     )
+
+    // CustomerCenter WIP: Uncomment once access to CustomerCenter is public.
+//    if (isCustomerCenterBottomSheetVisible) {
+//        val customerCenterSheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)
+//
+//        CustomerCenterBottomSheet(
+//            onDismissRequest = { isCustomerCenterBottomSheetVisible = false },
+//            sheetState = customerCenterSheetState,
+//        )
+//    }
 }
 
 @Composable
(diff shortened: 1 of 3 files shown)


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
```diff
@@ -64,6 +71,16 @@ fun AppInfoScreen(viewModel: AppInfoScreenViewModel = viewModel<AppInfoScreenVie
         isVisible = isDebugBottomSheetVisible,
         onDismissCallback = { isDebugBottomSheetVisible = false },
     )
+
+    // CustomerCenter WIP: Uncomment once access to CustomerCenter is public.
```
tonidero, line 75: I need to keep this commented out since I'm not making public the composables in the revenuecatui module yet. I'm planning to create a PR just making those public and this available so we can cherry-pick as we did in iOS. Lmk if you have a better idea!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterBottomSheet.kt
```diff
@@ -0,0 +1,62 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import com.revenuecat.purchases.ui.revenuecatui.ExperimentalPreviewRevenueCatUIPurchasesAPI
+
+/**
+ * Composable offering a bottom sheet Customer Center UI configured from the RevenueCat dashboard.
+ */
+@ExperimentalPreviewRevenueCatUIPurchasesAPI
+@ExperimentalMaterial3Api
```
tonidero, line 18: I'm propagating the required annotations to the developer, since we're using google's experimental APIs (even though it's the recommended API 😅)

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
JayShortway: Short term I don't have a better idea. In the medium term, maybe we can do something with variants, e.g. create a customer center variant that has a different source set. Or maybe we can see if we can replicate the compiler flag approach in iOS 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterBottomSheet.kt
JayShortway: Hmm, this means we're risking binary incompatibility if the Material3 version we expect is different from the one we get at runtime. We basically need to always update to the latest Material3 version (until this is stable), and thereby force our consumers to do the same. 

Do we need the bottomsheet composable? It is fairly trivial for people to wrap the `CustomerCenterView` in a `ModalBottomSheet`. We could even provide sample code in our docs.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterBottomSheet.kt
```diff
@@ -0,0 +1,62 @@
...
+) {
+    CustomerCenterBottomSheetScaffold(
+        onDismissRequest,
+        sheetState,
+    ) {
+        InternalCustomerCenter()
+    }
+}
+
+private const val CUSTOMER_CENTER_BOTTOM_SHEET_MAX_HEIGHT_PERCENTAGE = 0.9f
```
JayShortway, line 33: Nit, but this could be slightly more concise as it is private:
```diff
- private const val CUSTOMER_CENTER_BOTTOM_SHEET_MAX_HEIGHT_PERCENTAGE = 0.9f
+ private const val BOTTOM_SHEET_MAX_HEIGHT_PERCENTAGE = 0.9f
```

JayShortway (commented):
Looks good! Only concern is the risk of binary incompatibility vs the value provided.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
tonidero: Hmm yeah, I don't think there is a way to have a compiler flag in Android afaik? As for a variant, it could work but I think it's a lot of overhead IMO for this?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterBottomSheet.kt
tonidero: Well we don't *need* it... So it would be:
```
if (isCustomerCenterBottomSheetVisible) {
    val customerCenterSheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)

    CustomerCenterBottomSheet(
        onDismissRequest = { isCustomerCenterBottomSheetVisible = false },
        sheetState = customerCenterSheetState,
    )
}
```
vs 
```
if (isCustomerCenterBottomSheetVisible) {
    val customerCenterSheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true)

    ModalBottomSheet(
        onDismissRequest = { isCustomerCenterBottomSheetVisible = false } ,
        modifier = Modifier.fillMaxHeight(...), // Optional, can customize how it looks
        sheetState = sheetState,
    ) {
        CustomerCenter(modifier = Modifier.fillMaxSize()) // Or any other modifiers...
    }
}
```

So a custom composable would give us more control of the display and I think it's slightly nicer (I'm still a bit reticent to leave total control to the developer 😅) but I see what you mean, and I don't have a solution... so yeah I think we should hold on this and see what devs think. As you said, we can try to document this. 

cc @vegaro @aboedo in case you have other thoughts.

tonidero:
Will close this and add it to paywall tester using the `CustomerCenter` composable instead. Thanks for the feedback @JayShortway!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterBottomSheet.kt
JayShortway: Yea it would definitely be a nice addition! I'm just not sure if that weighs up against the binary incompatibility risk. We can also decide to add it as soon as the `ModalBottomSheet` API becomes stable?

tonidero: Yup that sounds like a good idea 👍

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
JayShortway: Yea that's probably a lot of overhead indeed. 😕

PR #1869, opened 2024-10-04
[CustomerCenter] Fix help path deserializing when unknown type

tonidero:
### Description
While testing one thing I noticed we were not handling unknown help path types. This basically caused the whole screen to not be deserialized, since we catch deserialization issues at that level, but we can do better than that and just not deserialize the unknown path type.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/HelpPathsSerializer.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/HelpPathsSerializer.kt
@@ -2,11 +2,8 @@ package com.revenuecat.purchases.customercenter
 
 import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
 import com.revenuecat.purchases.common.debugLog
-import com.revenuecat.purchases.customercenter.CustomerCenterConfigData.Screen
-import com.revenuecat.purchases.customercenter.CustomerCenterConfigData.Screen.ScreenType
 import kotlinx.serialization.KSerializer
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/HelpPathsSerializer.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.serialization.builtins.MapSerializer
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.jsonArray
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal object HelpPathsSerializer : KSerializer<List<CustomerCenterConfigData.HelpPath>> {
+    override val descriptor: SerialDescriptor = MapSerializer(ScreenType.serializer(), Screen.serializer()).descriptor
```
JayShortway, line 18: Should this be `ListSerializer`?

JayShortway (approved):
Ah yes, this is a frequent annoyance with collections + kotlinx.serialization. Thanks for taking care of it, much better!

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/HelpPathsSerializer.kt
tonidero: 🤦 Yes... thanks!

PR #1871, opened 2024-10-07
Fix crash on PaywallActivity if app restarted

tonidero:
### Description
If the app is killed by the system while the `PaywallActivity` is showing, it was crashing when trying to display the paywall. This fixes that so instead, we just finish the activity with an error.

This situation could happen when coming back to the app from the recent apps menu of the device after it was killed.


tonidero:
Back to draft as we compare the approach with the one in #1872

tonidero:
Closing in favor of #1872

PR #1872, opened 2024-10-07
Save SDK configuration on paywall activity so it can reconfigure automatically

tonidero:
### Description
In here we store the latest known state for `Purchases` so we can reconfigure when displaying the `PaywallActivity` and the SDK is not configured. This may happen if the app is killed while displaying the `PaywallActivity`, the developer is not configuring in the application but on an activity and the user returns through the app through the recent apps menu.

Additionally, it includes the changes from #1866 so we avoid reconfiguring the SDK if the configuration parameters are the same.

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint
 import android.app.Activity
 import android.content.Context
 import androidx.annotation.VisibleForTesting
+import com.revenuecat.purchases.common.Config
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.PlatformInfo
 import com.revenuecat.purchases.common.infoLog
@@ -882,7 +883,12 @@ class Purchases internal constructor(
             configuration: PurchasesConfiguration,
         ): Purchases {
             if (isConfigured) {
-                infoLog(ConfigureStrings.INSTANCE_ALREADY_EXISTS)
+                if (backingFieldSharedInstance?.purchasesOrchestrator?.configuration == configuration) {
+                    infoLog(ConfigureStrings.INSTANCE_ALREADY_EXISTS_WITH_SAME_CONFIG)
+                    return sharedInstance
+                } else {
+                    infoLog(ConfigureStrings.INSTANCE_ALREADY_EXISTS)
+                }
             }
             return PurchasesFactory().createPurchases(
                 configuration,
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -41,8 +41,8 @@ class Purchases internal constructor(
     /**
      * The configuration that was used to start Purchases.
      */
-    val configuration: PurchasesConfiguration
-        get() = purchasesOrchestrator.getLatestConfiguration()
+    val latestConfiguration: PurchasesConfiguration
+        get() = purchasesOrchestrator.latestConfiguration
 
     /**
      * Default to TRUE, set this to FALSE if you are consuming and acknowledging transactions
@@ -889,7 +889,7 @@ class Purchases internal constructor(
             configuration: PurchasesConfiguration,
         ): Purchases {
             if (isConfigured) {
-                if (backingFieldSharedInstance?.purchasesOrchestrator?.configuration == configuration) {
+                if (backingFieldSharedInstance?.purchasesOrchestrator?.latestConfiguration == configuration) {
                     infoLog(ConfigureStrings.INSTANCE_ALREADY_EXISTS_WITH_SAME_CONFIG)
                     return sharedInstance
                 } else {
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -38,6 +38,11 @@ import java.net.URL
 class Purchases internal constructor(
     @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
 ) : LifecycleDelegate {
+    /**
+     * The configuration that was used to start Purchases.
+     */
+    val configuration: PurchasesConfiguration
+        get() = purchasesOrchestrator.getLatestConfiguration()
 
     /**
      * Default to TRUE, set this to FALSE if you are consuming and acknowledging transactions
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -39,10 +39,10 @@ class Purchases internal constructor(
     @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
 ) : LifecycleDelegate {
     /**
-     * The configuration that was used to start Purchases.
+     * The current configuration parameters of the Purchases SDK.
      */
-    val latestConfiguration: PurchasesConfiguration
-        get() = purchasesOrchestrator.latestConfiguration
+    val currentConfiguration: PurchasesConfiguration
+        get() = purchasesOrchestrator.currentConfiguration
 
     /**
      * Default to TRUE, set this to FALSE if you are consuming and acknowledging transactions
@@ -889,7 +889,7 @@ class Purchases internal constructor(
             configuration: PurchasesConfiguration,
         ): Purchases {
             if (isConfigured) {
-                if (backingFieldSharedInstance?.purchasesOrchestrator?.latestConfiguration == configuration) {
+                if (backingFieldSharedInstance?.purchasesOrchestrator?.currentConfiguration == configuration) {
                     infoLog(ConfigureStrings.INSTANCE_ALREADY_EXISTS_WITH_SAME_CONFIG)
                     return sharedInstance
                 } else {
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint
 import android.app.Activity
 import android.content.Context
 import androidx.annotation.VisibleForTesting
-import com.revenuecat.purchases.common.Config
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.PlatformInfo
 import com.revenuecat.purchases.common.infoLog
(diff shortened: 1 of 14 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -230,6 +230,10 @@ internal class PurchasesOrchestrator(
         }
     }
 
+    fun getLatestConfiguration(): PurchasesConfiguration {
+        return configuration.copy(appUserID = this.appUserID)
```
JayShortway, line 234: We could also make `configuration` always be the latest? Otherwise we might get confused between `configuration` and `getLatestConfiguration()`, and when to use which.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -49,6 +60,28 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
         }
     }
 
+    @Parcelize
+    private data class SdkConfigArgs(
```
JayShortway, line 64: We could use `PurchasesConfiguration` directly if we make it parcelable. We can ignore the `ExecutorService` using `@IgnoredOnParcel`. See [the docs](https://developer.android.com/kotlin/parcelize#skip_properties_from_serialization). 

The upside of using `PurchasesConfiguration` directly is that we don't accidentally add more properties that aren't parcelable, or at least it forces us to think about that.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -49,6 +60,28 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
...
+        val store: Store,
+        val diagnosticsEnabled: Boolean,
+        val verificationMode: EntitlementVerificationMode,
+        val dangerousSettings: DangerousSettings,
+        val pendingTransactionsForPrepaidPlansEnabled: Boolean,
+    ) : Parcelable
+
+    private fun getSdkConfigArgs(savedInstanceState: Bundle): SdkConfigArgs? {
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            savedInstanceState.getParcelable(SDK_CONFIG_EXTRA, SdkConfigArgs::class.java)
```
JayShortway, line 78: I always found the warning in the docs for this one a bit scary: 

> **Warning:** the class that implements `Parcelable` has to be the immediately enclosing class of the runtime type of its CREATOR field (that is, `Class.getEnclosingClass()` has to return the parcelable implementing class), otherwise this method might throw an exception. If the `Parcelable` class does not enclose the CREATOR, use the deprecated `getParcelable(String)` instead.

We should be good, but we can also just use the deprecated variant to avoid it altogether. If we ever change things around in the future, we might run into this exception.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -123,6 +180,26 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
         setResult(RESULT_OK, createResultIntent(PaywallResult.Error(error)))
     }
 
+    private fun configureSdkWithSavedData(savedInstanceState: Bundle) {
+        val sdkConfigArgs = getSdkConfigArgs(savedInstanceState)
+        if (sdkConfigArgs == null) {
+            Logger.e("TEST: Missing SDK configuration arguments")
```
JayShortway, line 186: Good log, but maybe without the "TEST"? 😛

JayShortway (commented):
Pretty cool! Some suggestions, but this should be a nice improvement!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
tonidero: 😅 Yeah my bad, I didn't clean it up :P

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
tonidero: I think we also would need to ignore the `Context` right? And then we would need to get the application context again, which I'm not sure if it would be easy to do from there...

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: yeah good point... can probably clean this up a bit.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
JayShortway: Ah yes, we'd also need to ignore the `Context`! It would be difficult to get the Application Context from `PurchasesConfiguration` indeed. Nevermind haha.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -263,12 +280,10 @@ open class PurchasesConfiguration(builder: Builder) {
     }
 
     override fun hashCode(): Int {
-        var result = context.hashCode()
```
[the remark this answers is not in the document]
JayShortway: I think so too. For the purpose of checking whether the configuration set in `PaywallActivity` is the same as one set in a different Activity, we can ignore the `Context` and `ExecutorService` in the equality check. I don't know if we're doing any other equality checks though, or are intending to do so in the future?

tonidero:
I changed the base to `main` so it also includes the changes in #1866, since I think it's going to be easier to review all together.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -110,6 +111,13 @@ internal class PurchasesOrchestrator(
             purchasesStateCache.purchasesState = value
         }
 
+    val latestConfiguration: PurchasesConfiguration
+        get() = if (initialConfiguration.appUserID == null) {
```
tonidero, line 115: I had to add this condition... Basically, the condition to avoid reconfiguration was never wrong if the developer doesn't set an `appUserId`. In this case, we should just reuse the last configured user id, so it should be safe.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -38,6 +38,11 @@ import java.net.URL
 class Purchases internal constructor(
     @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
 ) : LifecycleDelegate {
+    /**
+     * The configuration that was used to start Purchases.
+     */
+    val latestConfiguration: PurchasesConfiguration
```
JayShortway, line 44: Some naming suggestions: 
- `currentConfiguration`
- or maybe just `configuration`

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -110,6 +111,13 @@ internal class PurchasesOrchestrator(
...
+            initialConfiguration
+        } else {
+            initialConfiguration.copy(appUserID = this.appUserID)
```
JayShortway, line 118: We're now creating new copies every time this is read. Could we keep the up-to-date configuration in a backing `var` maybe? We can update it whenever the `appUserId` changes.

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesConfigureTest.kt
```diff
@@ -69,4 +72,100 @@ internal class PurchasesConfigureTest : BasePurchasesTest() {
         Purchases.configure(builder.build())
         assertThat(Purchases.sharedInstance.store).isEqualTo(Store.PLAY_STORE)
     }
+
+    @Test
+    fun `Calling configure multiple times with same configuration does not create a new instance`() {
```
JayShortway, line 77: Nice tests! 🙌

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -123,6 +176,26 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
...
+                .appUserID(sdkConfigArgs.appUserId)
+                .purchasesAreCompletedBy(sdkConfigArgs.purchasesAreCompletedBy)
+                .showInAppMessagesAutomatically(sdkConfigArgs.showInAppMessagesAutomatically)
+                .store(sdkConfigArgs.store)
+                .diagnosticsEnabled(sdkConfigArgs.diagnosticsEnabled)
+                .entitlementVerificationMode(sdkConfigArgs.verificationMode)
+                .dangerousSettings(sdkConfigArgs.dangerousSettings)
+                .pendingTransactionsForPrepaidPlansEnabled(sdkConfigArgs.pendingTransactionsForPrepaidPlansEnabled)
+                .build(),
+        )
```
JayShortway, line 196: If we ever add a configuration option, we have to remember to add it to `SdkConfigArgs`. Can't really think of a solution though haha (other than some custom Detekt stuff or something).

JayShortway (approved):
Looks good! Just had 2 more minor suggestions.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Hmm I've been thinking whether this would be more work right now to calculate on user change or as it is now, when there is a configuration or the PaywallActivity's `onSaveInstanceState` is called... And I think it might be less work as it is right now? Normally users only configure once, so this won't be called, but there is a bigger chance of logging a different user (though also an uncommon case)... In any case I think we might be trying to early optimizing and I prefer to calculate it on the fly rather than trying to cache the value which might be outdated later... Lmk if you think otherwise!

JayShortway: Yea this is less error prone, and the performance difference is probably negligible. I'm okay with it!

PR #1873, opened 2024-10-08
Adds X-Is-Debug-Build header

JayShortway:
As the title says. Quite a few changed files. Here's what's new: 

- `AppConfig` has a new `isDebugBuild` property.
- `HTTPClient` reads this property and sets the `X-Is-Debug-Build` header accordingly.
- `PurchasesFactory` uses the new `IsDebugBuildProvider` to determine the `isDebugBuild` property when it creates the `AppConfig`. 
- `AppConfigTest` is adjusted.
- There's a new `HTTPClientTest` case. 
- `IsDebugBuildProviderTest` was added.

iOS implementation: https://github.com/RevenueCat/purchases-ios/pull/4364

### DIFF
--- a/purchases/src/androidTest/kotlin/com/revenuecat/purchases/purchasesExtensions.kt
+++ b/purchases/src/androidTest/kotlin/com/revenuecat/purchases/purchasesExtensions.kt
@@ -9,7 +9,7 @@ internal fun Purchases.Companion.configure(
     forceServerErrors: Boolean = false,
     forceSigningErrors: Boolean = false,
 ): Purchases {
-    return PurchasesFactory(isDebugBuild = { false }).createPurchases(
+    return PurchasesFactory(isDebugBuild = { true }).createPurchases(
         configuration,
         platformInfo,
         proxyURL,
--- a/purchases/src/androidTest/kotlin/com/revenuecat/purchases/purchasesExtensions.kt
+++ b/purchases/src/androidTest/kotlin/com/revenuecat/purchases/purchasesExtensions.kt
@@ -9,7 +9,7 @@ internal fun Purchases.Companion.configure(
     forceServerErrors: Boolean = false,
     forceSigningErrors: Boolean = false,
 ): Purchases {
-    return PurchasesFactory().createPurchases(
+    return PurchasesFactory(isDebugBuild = { false }).createPurchases(
         configuration,
         platformInfo,
         proxyURL,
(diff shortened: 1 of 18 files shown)


purchases/src/androidTest/kotlin/com/revenuecat/purchases/purchasesExtensions.kt
```diff
@@ -9,7 +9,7 @@ internal fun Purchases.Companion.configure(
     forceServerErrors: Boolean = false,
     forceSigningErrors: Boolean = false,
 ): Purchases {
-    return PurchasesFactory().createPurchases(
+    return PurchasesFactory(isDebugBuild = { false }).createPurchases(
```
tonidero, line 12: Hmm since this is for tests, I wonder if we should default to `true`...

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -50,6 +51,7 @@ import java.util.concurrent.Executors
 import java.util.concurrent.ThreadFactory
 
 internal class PurchasesFactory(
+    private val isDebugBuild: IsDebugBuildProvider,
```
tonidero, line 54: Hmm by the name `isDebugBuild` I would think this to be a boolean... Maybe we should rename to `debugBuildProvider` and have a `get()` method or something instead?

purchases/src/test/java/com/revenuecat/purchases/utils/IsDebugBuildProviderTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import android.content.Context
+import android.content.pm.ApplicationInfo
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class IsDebugBuildProviderTest {
```
tonidero, line 11: ❤️

tonidero (approved):
Looks great! Just a couple suggestions but not blockers.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
JayShortway: Hmm, I see what you mean, but the reason is that we call it as a function, on line 79 below: 
```kotlin
isDebugBuild()
```  
Imo that's "worth it", because otherwise that line would become something like: 
```kolin
debugBuildProvider.get()
``` 
which I think is less readable. 

What do you think?

tonidero: Yeah makes sense, while there might be some confusion when reading the name of the property, it's more legible after so I'm fine with it 👍

PR #1875, opened 2024-10-09
Android paywall custom package tag

simonmitchell:
### Checklist

### Motivation

This is motivated by a need for us to provide discount offers to our users whilst also using the revenuecat footer UI. We run our offer codes using specifically tagged `SubscriptionOption` like so:

```
rcPackage.product.subscriptionOptions?.withTag(offerCode)?.getOrNull(0)
```

without this change to allow overriding the `SubscriptionOption` the user is purchasing using the footer component, we cannot apply offer codes in our custom paywall implementation.

### Description

A new interface has been added to the ui package called `SubscriptionOptionProvider` with a single function:
```
fun subscriptionOption(rcPackage: Package): SubscriptionOption?
```

This provider can be provided to the library via the paywall options builder.

If provided, the `PaywallViewModel` will call the function, and if the `subscriptionOption` function returns non-null then it will use `awaitPurchase` function that takes a `SubscriptionOption` rather than a `Package`.

These changes were tested locally by running our app from the branch on our fork, and has been in the production version of our Android app for the past few weeks.

Note that I have not written any unit tests, as I am unsure if required here, but would be more than happy to if you deem necessary! 😄


skienzl:
@JayShortway any plans on moving forward with this? 
For example our usecase is that we have multiple discounts at the same time. And we need to create new play store subscriptions for each discount which makes it harder to keep them im sync and also provide the user with winbacks and so.

PR #1882, opened 2024-10-16
[AUTOMATIC] Release/8.9.0

RCGitBot:

### 🔄 Other Changes
* Adds X-Is-Debug-Build header (#1873) via JayShortway (@JayShortway)


tonidero:
Hmm going to hold this release, but lmk if you want to ship this and I can trigger it again!

JayShortway:
My thinking was that it is somewhat useful to ship this header at roughly the same time for Android and iOS, but it's definitely not essential.

tonidero:
Right... Let's release it then 👍

tonidero:
Actually, going to recreate it but making it a patch. I don't think this should be considered a new feature by devs using the SDK.

PR #1886, opened 2024-10-22
Detekt auto-fixes

tonidero:
### Description
Detekt is applying these fixes for me, not sure why it's not applying them remotely... They seemed to make sense so I just applied them for now.


JayShortway (approved):
Thanks for doing this! I've seen these get applied on my end too, so maybe they were missed at some point.

PR #1888, opened 2024-10-24
[CI] Allow publishing snapshot releases manually from branches

tonidero:
### Description
Add new workflow to deploy snapshots from branches manually

Plan to use this to deploy some preview versions of some features, allowing people to test them before actually making a stable release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ orbs:
 parameters:
   action:
     type: enum
-    enum: [default, bump, paywall_tester_release, generate_revenuecatui_snapshots, deploy_snapshot]
+    enum: [default, bump, paywall_tester_release, generate_revenuecatui_snapshots, deploy_snapshot_release]
     default: default
 
 commands:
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -13,7 +13,7 @@
 #Fri Mar 31 10:31:20 PDT 2023
 GROUP=com.revenuecat.purchases
 
-VERSION_NAME=8.9.0-SNAPSHOT
+VERSION_NAME=8.9.0-SNAPSHOT-ANONYMOUS-REDEMPTION
```
JayShortway, line 16: Should this be committed?

.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ orbs:
 parameters:
   action:
     type: enum
-    enum: [default, bump, paywall_tester_release, generate_revenuecatui_snapshots]
+    enum: [default, bump, paywall_tester_release, generate_revenuecatui_snapshots, deploy_snapshot]
```
JayShortway, line 51: Since the term "snapshot" is pretty ambiguous (e.g. snapshot tests, snapshot releases), maybe we can name this action `deploy_snapshot_release`? (Or `publish_snapshot_release`, but I see we're using "deploy" in more places in this file.)

JayShortway (commented):
Nice idea to make this easier! Just a comment on the naming. I have the same comment on the PR title haha, maybe something like: 
> Allow publishing snapshot releases manually from branches

gradle.properties
tonidero: Whoops sorry no 🙏

JayShortway (approved):
LGTM!

PR #1889, opened 2024-10-24
[Experimental] Web purchase redemption

tonidero:
### Description
This adds experimental support to redeeming anonymous web purchases performed through RCBilling. The main API changes here are:
- Adds a new `Purchases.parseAsDeepLink` that parses the intent and returns a deep link if it needs to be handled
- Adds a new `Purchases.sharedInstance.redeemWebPurchase` that uses a parsed deep link to perform the redemption.
- Adds a new `DeepLink` sealed class with the types of deep links
- Adds a new `RedeemWebPurchaseListener` to listen to the result of the redemption.
- Adds a new `RedeemWebPurchaseListener.Result` sealed class with the result of the redemption.

All these are currently experimental APIs


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
```diff
@@ -113,6 +115,27 @@ class OverviewFragment : Fragment(), OfferingCardAdapter.OfferingCardAdapterList
...
     }
 
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    override fun onResume() {
+        super.onResume()
+
+        val activity = requireActivity() as MainActivity
+        val deepLink = activity.rcDeepLink ?: return
+        if (deepLink is Purchases.DeepLink.WebRedemptionLink) {
+            activity.clearDeepLink()
```
tonidero, line 125: Not a fan of the developer having to handle the deep links like this... I think there is a big chance of the developer passing us the same link multiple times (which could still happen with the approach in #1880 but less likely IMO).

gradle.properties
```diff
@@ -13,7 +13,7 @@
 #Fri Mar 31 10:31:20 PDT 2023
 GROUP=com.revenuecat.purchases
 
-VERSION_NAME=8.9.0-SNAPSHOT
+VERSION_NAME=8.9.0-ANONYMOUS-REDEMPTION-SNAPSHOT.01
```
tonidero, line 16: Ignore these. It was modified to push a snapshot with the changes.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/MainActivity.kt
```diff
@@ -1,13 +1,29 @@
...
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.Purchases
 import com.revenuecat.purchases_sample.R
 
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
 class MainActivity : AppCompatActivity() {
 
+    internal var rcDeepLink: Purchases.DeepLink? = null
```
tonidero, line 13: If multiple links are used continuously before redeeming each, the developer is in charge of handling each on their own... For this PoC, I'm just caching the last one used, but again, it's more logic the developer will need to handle.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
```diff
@@ -113,6 +115,27 @@ class OverviewFragment : Fragment(), OfferingCardAdapter.OfferingCardAdapterList
         }
     }
 
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    override fun onResume() {
```
tonidero, line 119: Could have done this on the Activity, but preferred to do it in the fragment to show how it would be to handle at the UI part of the app (whether it's using compose or fragments)

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -181,6 +181,11 @@ class Purchases internal constructor(
     }
     //endregion
 
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    sealed class DeepLink {
+        class WebRedemptionLink internal constructor(internal val redemptionToken: String) : DeepLink()
```
tonidero, line 186: Made everything internal. Devs shouldn't need to build an instance of this on their own, nor access the token directly.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -788,9 +791,26 @@ class Purchases internal constructor(
...
+    sealed class DeepLink {
+        class WebRedemptionLink internal constructor(internal val redemptionToken: String) : DeepLink()
+    }
+
     // region Static
     companion object {
 
+        @ExperimentalPreviewRevenueCatPurchasesAPI
+        @JvmStatic
+        fun parseIntent(intent: Intent?): DeepLink? {
```
JayShortway, line 809: Maybe this function could have some mention of "deep link" in the name so it's easy to find in IDE autocomplete? Some ideas: 
- `parseAsDeepLink()`
- `createDeepLink()`
- `buildDeepLink()`
- `getDeepLink()`

JayShortway, line 809: I think `Intent` can be not-nullable here, unless I'm missing something? From an API perspective, it communicates that `null` is valid input, but I don't think it is? (It will never parse into a `DeepLink`.)

JayShortway: Maybe we can also add an extension function that then calls this function? Something like 
```kotlin
fun Intent.parseAsDeepLink(): DeepLink?
```

There's [`@file:JvmMultifileClass`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-multifile-class/) but it doesn't seem to work great for our case.

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: Good call! What's the reason this is nested in `Purchases`? (Don't really have an opinion, just curious haha.)

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/MainActivity.kt
JayShortway: I see your point, but wouldn't they need to manage the `WebPurchaseRedeemer` objects in the other approach? 

I'm also trying to think how much of an edge case this is. Is it realistic that a user is buying multiple things, clicking on the links for each (e.g. before being logged in in the app), and then starts a login flow and expects that all purchases are then redeemed? Let me know if you have a different scenario in mind, just trying to think this through haha.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
JayShortway: Can we keep track of what is already redeemed? E.g. by storing those tokens/urls on disk or something.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
```diff
@@ -113,6 +115,27 @@ class OverviewFragment : Fragment(), OfferingCardAdapter.OfferingCardAdapterList
...
+            Purchases.sharedInstance.redeemWebPurchase(deepLink) { result ->
+                when (result) {
+                    is RedeemWebResultListener.Result.Success -> {
+                        showToast("Successfully redeemed web purchase. Updating customer info.")
+                    }
+                    is RedeemWebResultListener.Result.Error -> {
+                        showUserError(requireActivity(), result.error)
+                    }
+                }
+            }
```
JayShortway, line 135: I do like the simplicity of this.

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/RedeemWebResultListener.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import com.revenuecat.purchases.CustomerInfo
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.PurchasesError
+
+/**
+ * Interface to handle the redemption of a RevenueCat Web purchase.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+fun interface RedeemWebResultListener {
```
JayShortway, line 11: Naming suggestion: `RedeemWebPurchaseListener`. (Mostly focused on getting the word "purchase" in there.)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -788,9 +791,26 @@ class Purchases internal constructor(
...
     // endregion
 
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    fun redeemWebPurchase(webRedemptionLink: DeepLink.WebRedemptionLink, listener: RedeemWebResultListener) {
+        purchasesOrchestrator.redeemWebPurchase(webRedemptionLink, listener)
+    }
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    sealed class DeepLink {
+        class WebRedemptionLink internal constructor(internal val redemptionToken: String) : DeepLink()
```
JayShortway, line 801: Naming suggestion: `WebPurchaseRedemption`. Also focussed on getting the word "purchase" in there, and I think `Link` can be omitted because it is already nested inside `DeepLink`.

purchases/src/main/kotlin/com/revenuecat/purchases/deeplinks/DeepLinkParser.kt
```diff
@@ -0,0 +1,28 @@
+package com.revenuecat.purchases.deeplinks
+
+import android.net.Uri
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.common.debugLog
+
+internal class DeepLinkParser {
```
JayShortway, line 8: Maybe this can be an object, or will it have instance state later? (Or even a global internal extension function if you like.)

JayShortway (commented):
I do like it! Had some naming suggestions mostly. I also think it speaks to its simplicity that it's about half the additional lines vs the other approach (342 vs 767). 

I think the main decision boils down to how much we want to cater to the edge case you mentioned, and how big of an edge case that is. Let me know if you have a different take!

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Mostly seemed `DeepLink` was too generic as a top-level class 😅 So I wanted to namespace it inside the SDK.

JayShortway: Haha alright makes sense! (Devs can always use import aliasing of course.)

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
tonidero: Yes we could... Though I guess there is also a chance it's a valid use case to redeem, as in, maybe the redemption failed the first time due to a connection error or something and they need to try again. I guess we could only "cache" it if there is a successful redemption in this case, but if we for example add deep links to display the paywall in the future, there would be valid use cases of showing the paywall multiple times.

Not to say it's not ok! This approach does provide much more flexibility to the developer, but also more responsibility :P

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/MainActivity.kt
tonidero: Yeah it's certainly an edge case... And yeah it might be ok to just leave it to the developer. We can always try to provide some code samples if devs start asking for them.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: yeah... I did it for simplicity purposes so devs don't need to check for a null intent... But I agree, it doesn't make sense. Will change it, thanks!

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewFragment.kt
JayShortway: Yea indeed, I meant to cache successful redemptions. If only to alleviate our own backend. Paywall or other non web purchase deeplinks would indeed be excluded from this.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -788,9 +791,41 @@ class Purchases internal constructor(
     }
     // endregion
 
+    /**
+     * Redeem a web purchase using a [DeepLink.WebPurchaseRedemption] object obtained
+     * through [Purchases.parseAsDeepLink].
```
JayShortway, line 796: Great to point people to `parseAsDeepLink` here!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -788,9 +791,41 @@ class Purchases internal constructor(
...
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    fun redeemWebPurchase(webPurchaseRedemption: DeepLink.WebPurchaseRedemption, listener: RedeemWebPurchaseListener) {
+        purchasesOrchestrator.redeemWebPurchase(webPurchaseRedemption, listener)
+    }
+
+    /**
+     * Represents a valid RevenueCat deep link.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    sealed class DeepLink {
```
JayShortway, line 807: Could be a sealed interface, but nabd. (Same in `customEntitlementComputation`.)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -788,9 +791,41 @@ class Purchases internal constructor(
...
     companion object {
 
+        /**
+         * Given an intent, parses the deep link if any and returns a parsed version of it.
+         * Currently supports web redemption links.
+         * @return A parsed version of the deep link or null if it's not a valid RevenueCat deep link.
+         */
+        @ExperimentalPreviewRevenueCatPurchasesAPI
+        @JvmStatic
+        fun parseAsDeepLink(intent: Intent): DeepLink? {
```
JayShortway, line 824: What do you think about providing an extension variant of this, for idiomatic Kotlin?

purchases/src/main/kotlin/com/revenuecat/purchases/deeplinks/WebPurchaseRedemptionHelper.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val identityManager: IdentityManager,
+    private val offlineEntitlementsManager: OfflineEntitlementsManager,
+    private val customerInfoUpdateHandler: CustomerInfoUpdateHandler,
+    private val mainHandler: Handler? = Handler(Looper.getMainLooper()),
+) {
+    fun handleRedeemWebPurchase(
+        deepLink: Purchases.DeepLink.WebPurchaseRedemption,
+        listener: RedeemWebPurchaseListener,
+    ) {
+        debugLog("Starting redeeming web purchase.")
```
JayShortway, line 27: Nit:
```suggestion
        debugLog("Starting web purchase redemption.")
```

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendRedeemWebPurchaseTest.kt
```diff
@@ -0,0 +1,184 @@
...
+import java.net.URL
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration.Companion.seconds
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+@RunWith(AndroidJUnit4::class)
+class BackendRedeemWebPurchaseTest {
```
JayShortway, line 42: 🙌

JayShortway (approved):
Amazing! Great job on this! Also good to see it's extensively tested. Had some minor comments / suggestions, but no blockers. Feel free to ignore at your discretion.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/1889/commits/b55892d6c12da4f3e5ec9c37e841afd5934334c8. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/IntentExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package com.revenuecat.purchases
+
+import android.content.Intent
+
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@JvmSynthetic
```
JayShortway, line 6: Good one!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: Perfect!

PR #1890, opened 2024-10-25
[DO NOT MERGE] Eco 1550 demo app  for Web Redeems

nicfix:
### Checklist


tonidero (commented):
Some comments. As a test app, there is nothing blocking, but I do wonder about whether we should simplify this test app a lot instead of copying over from `MagicWeatherCompose`... Wdyt? Happy to help on that!

tonidero:
@nicfix I created a new separate demo app and merged it in #1908. It's used in the sample video in my last demo in Slack. I felt it better than copying the entire magic weather again for this... I think we can either:
- Port the changes to the actual MagicWeatherCompose, so it serves as a sample and remove the sample app in my PR
- Keep the new sample app separate since it's simpler.

In any case, I think we can probably close this?

PR #1891, opened 2024-10-28
Keeps the org.json package.

JayShortway:
As the title says. This rule will normally have no effect as we don't declare the `org.json:json` dependency, but some apps do (transitively). In that scenario, this rule avoids `NoSuchMethodError`s and `NoSuchFieldError`s. 

**Note:** I gave this the `pr:fix` label, as it will fix the issue reported in [this Zendesk ticket](https://revenuecat.zendesk.com/agent/tickets/47240). Let me know if you think we should give it another label instead.

### DIFF
--- a/purchases/consumer-rules.pro
+++ b/purchases/consumer-rules.pro
@@ -54,3 +54,8 @@
 -keep class kotlinx.serialization.internal.ClassValueParametrizedCache$initClassValue$1 { ** computeValue(java.lang.Class); }
 -keep class kotlinx.serialization.internal.ClassValueCache$initClassValue$1 { ** computeValue(java.lang.Class); }
 # END Keep kotlinx.serialization annotations.
+
+# The org.json package is part of the Android framework, so the classes are always available.
+# However, sometimes it gets added to the app's classpath, either explicitly or transitively. When
+# this happens, it becomes susceptible to be shrunk. The following rule avoids that.
+-keep class org.json.* { *; }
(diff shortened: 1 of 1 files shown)


purchases/consumer-rules.pro
```diff
@@ -53,4 +53,9 @@
...
 -keep class kotlinx.serialization.internal.ClassValueParametrizedCache$initClassValue$1 { ** computeValue(java.lang.Class); }
 -keep class kotlinx.serialization.internal.ClassValueCache$initClassValue$1 { ** computeValue(java.lang.Class); }
-# END Keep kotlinx.serialization annotations.
\ No newline at end of file
+# END Keep kotlinx.serialization annotations.
+
+# The org.json package is part of the Android framework, so the classes are always available.
+# However, sometimes it gets added to the app's classpath, either explicitly or transitively. When
+# this happens, it becomes susceptible to be shrunk. The following rule avoids that.
+-keep class org.json.* { *; }
```
tonidero, line 61: > However, sometimes it gets added to the app's classpath, either explicitly or transitively.

Hmm is this something we're doing? As in, I'm a bit reticent to add keep rules to our SDK unless we know it's our fault...

JayShortway: Ah no it isn't, but some apps might do it. See the linked Zendesk ticket in the PR description. I updated the comment in the ProGuard file to clarify this: https://github.com/RevenueCat/purchases-android/pull/1891/commits/3673777dca8e75b14fd3bf4033aec99e0e3c3eb3. 

If the dependency is not included by the app, this rule does nothing. (`org.json` is provided by the Android OS in that case.) If the dependency is included by the app, this rule makes sure that all classes that we expect are kept. Without this rule, the dependency can be shrunk by R8/ProGuard, causing errors.

tonidero (approved):
Ok I think it should be ok 👍

PR #1893, opened 2024-10-29
Fix application of modifiers in `Markdown` component

tonidero:
### Description
We were passing the modifier for the `Markdown` composable down to the whole composable tree inside that Markdown. We are currently passing some margins to the markdown composable which were basically applied to all individual components of the markdown, so if processing the markdown generated several components, all of them were applied the paddings for example. 

I believe this was not expected so I decided to just not propagate the modifiers inside the `Markdown` composable and just keep it at the root.

| Before | After |
| --- | --- |
| image | image |


tonidero:
This has changed the snapshots a bit as can be seen in Emerge ^... Mostly some margins have moved... I feel these are ok but would like someone to double-check @JayShortway @vegaro

JayShortway:
@tonidero 

Regarding the baseline: that's interesting. Seems like this specific one wasn't caught by Detekt.
There are indeed some similar violations, if you search for `ModifierReused:Markdown.kt` and `ModifierNotUsedAtRoot:Markdown.kt` in `detekt-baseline.xml`. Not a blocker for me, feel free to decide if you want to fix those in this PR or not.

Regarding the Emerge screenshots: my only concern is the [`PurchaseButton`](https://www.emergetools.com/snapshot/25ef19d5-316a-468c-8da8-4f640e85afdc?tab=changed&sort=diff_desc&selected=fa7fea65-d2e0-486e-ad8c-8b1f1a508b80). It seems to be a bit less tall now, making it a smaller touch target. Could decrease conversion + accessibility. 
(The `PurchaseButton` also seems to be the main cause of the changes in e.g. [`Template4`](https://www.emergetools.com/snapshot/25ef19d5-316a-468c-8da8-4f640e85afdc?tab=changed&sort=diff_desc&selected=d8b9510d-31e4-462c-b2a2-91657fee8ee7).)

tonidero:
> my only concern is the [PurchaseButton](https://www.emergetools.com/snapshot/25ef19d5-316a-468c-8da8-4f640e85afdc?tab=changed&sort=diff_desc&selected=fa7fea65-d2e0-486e-ad8c-8b1f1a508b80).

Hmm good call... Going to see if I can minimize the change there. Great catch!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PurchaseButton.kt
```diff
@@ -150,7 +150,7 @@ private fun PurchaseButton(
                         fontWeight = FontWeight.SemiBold,
                         allowLinks = false,
                         modifier = Modifier
-                            .padding(vertical = UIConstant.defaultVerticalSpacing / 3)
+                            .padding(vertical = UIConstant.defaultVerticalSpacing * 2 / 3)
```
tonidero, line 153: This doesn't exactly match back up to the original size but it's pretty close. I think it's difficult to match it exactly since the components within markdown before perform some custom painting that was altering the sizes slightly... Lmk what you think!

JayShortway (approved):
Very nice improvement! Check out [this one](https://www.emergetools.com/snapshot/796de2da-1801-4f2d-bef4-f1cb26d061c8?tab=changed&sort=diff_desc&selected=9f86882e-6c5d-461a-902b-95fb3e3b7a05), the new one is much better haha. Thanks for fixing!

vegaro (approved):

PR #1897, opened 2024-11-04
Add `tenjinAnalyticsInstallationId` setter property

tonidero:
### Description
This adds a new reserved attribute `$tenjinId` and adds a new `setTenjinAnalyticsInstallationID` API to the SDK to set it.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/subscriberattributes/SpecialSubscriberAttributes.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/subscriberattributes/SpecialSubscriberAttributes.kt
@@ -37,7 +37,7 @@ internal enum class ReservedSubscriberAttribute(val value: String) {
      */
     MIXPANEL_DISTINCT_ID("\$mixpanelDistinctId"),
     FIREBASE_APP_INSTANCE_ID("\$firebaseAppInstanceId"),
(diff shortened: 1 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/subscriberattributes/SpecialSubscriberAttributes.kt
```diff
@@ -37,6 +37,7 @@ internal enum class ReservedSubscriberAttribute(val value: String) {
      */
     MIXPANEL_DISTINCT_ID("\$mixpanelDistinctId"),
     FIREBASE_APP_INSTANCE_ID("\$firebaseAppInstanceId"),
+    TENJIN_ANALYTICS_INSTALLATION_ID("\$tenjinId"),
```
tonidero, line 40: Thought about using `$tenjinAnalyticsInstallationId` to be more specific, but the backend is currently expecting this, and I don't think we will need to send anything else anytime soon, so decided to leave it as is.

Carlosedo (approved):
I'd like someone with SDK experience to also take a look, but everything seems fine from what I understand of the code haha

Thank you for taking care of this @tonidero!

MarkVillacampa (approved):

PR #1899, opened 2024-11-04
Add extra error cases for web purchases redemptions

tonidero:
### Description
This adds some new result types to a web purchase redemption:
- InvalidToken
- AlreadyRedeemed
- Expired. This one includes additional information like the obfuscated email where a new token was/will be sent and whether that email was sent.

This falls behind the experimental APIs introduced in #1889 

Additionally, this renames `parseAsDeepLink` to `parseAsWebPurchaseRedemption` and removes the `DeepLink` type. This is a breaking change, but should be ok since the API is marked as experimental and not usable still by any customers.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/WebPurchaseRedemption.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/WebPurchaseRedemption.kt
@@ -1,0 +1,9 @@
+package com.revenuecat.purchases
+
+/**
+ * Represents a web redemption link, that can be redeemed using [Purchases.redeemWebPurchase]
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+class WebPurchaseRedemption internal constructor(
+    internal val redemptionToken: String,
+)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -22,6 +22,7 @@ import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMap
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
 import com.revenuecat.purchases.customercenter.CustomerCenterRoot
+import com.revenuecat.purchases.interfaces.RedeemWebPurchaseListener
 import com.revenuecat.purchases.models.GoogleReplacementMode
 import com.revenuecat.purchases.models.PricingPhase
 import com.revenuecat.purchases.paywalls.events.PaywallEventRequest
@@ -76,7 +77,8 @@ internal typealias ProductEntitlementCallback = Pair<(ProductEntitlementMapping)
 @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
 internal typealias CustomerCenterCallback = Pair<(CustomerCenterConfigData) -> Unit, (PurchasesError) -> Unit>
 
-internal typealias RedeemWebPurchaseCallback = Pair<(CustomerInfo) -> Unit, (PurchasesError) -> Unit>
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal typealias RedeemWebPurchaseCallback = (RedeemWebPurchaseListener.Result) -> Unit
 
 internal enum class PostReceiptErrorHandlingBehavior {
     SHOULD_BE_MARKED_SYNCED,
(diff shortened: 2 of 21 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -760,11 +779,11 @@ internal class Backend(
         }
     }
 
-    private fun <K, S, E> MutableMap<K, MutableList<Pair<S, E>>>.addCallback(
+    private fun <K, F> MutableMap<K, MutableList<F>>.addCallback(
```
tonidero, line 782: I didn't need separate error/success callbacks for this operation so had to change this. I could have moved that transformation outside, but it would mean we would need to be leaking the result body out of the backend class, so I thought this was a bit better...

purchases/src/main/kotlin/com/revenuecat/purchases/common/errors.kt
```diff
@@ -97,6 +100,9 @@ private fun BackendErrorCode.toPurchasesErrorCode(): PurchasesErrorCode {
         BackendErrorCode.BackendInternalServerError,
         -> PurchasesErrorCode.UnexpectedBackendResponseError
         BackendErrorCode.BackendProductIDsMalformed -> PurchasesErrorCode.UnsupportedError
+        BackendErrorCode.BackendInvalidWebRedemptionToken -> PurchasesErrorCode.PurchaseInvalidError
+        BackendErrorCode.BackendWebPurchaseAlreadyRedeemed -> PurchasesErrorCode.ProductAlreadyPurchasedError
+        BackendErrorCode.BackendExpiredWebRedemptionToken -> PurchasesErrorCode.PurchaseInvalidError
```
tonidero, line 105: These shouldn't really be used... But I added it so it fixes the compiler.

purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
```diff
@@ -24,6 +24,24 @@ class PurchasesError(
     override fun toString(): String {
         return "PurchasesError(code=$code, underlyingErrorMessage=$underlyingErrorMessage, message='$message')"
     }
+
+    override fun equals(other: Any?): Boolean {
```
tonidero, line 28: Needed for some tests. I think it shouldn't hurt anything.

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/RedeemWebPurchaseListener.kt
```diff
@@ -13,16 +13,43 @@ fun interface RedeemWebPurchaseListener {
...
+         * Indicates that the redemption token is invalid.
+         */
+        object InvalidToken : Result()
+
+        /**
+         * Indicates that the redemption token has expired. An email with a new redemption token
+         * might be sent if a new one wasn't already sent recently.
+         * The email where it will be sent is indicated by the [obfuscatedEmail].
+         */
+        data class Expired(val obfuscatedEmail: String) : Result()
```
vegaro, line 36: Is obfuscated something like `c*****ro@gmail.com`?

purchases/src/main/kotlin/com/revenuecat/purchases/WebPurchaseRedemption.kt
```diff
@@ -0,0 +1,9 @@
+package com.revenuecat.purchases
+
+/**
+ * Represents a web redemption link, that can be redeemed using [Purchases.redeemWebPurchase]
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+class WebPurchaseRedemption internal constructor(
+    internal val redemptionToken: String,
```
vegaro, line 8: should this just be `token`?

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/RedeemWebPurchaseListener.kt
tonidero: Yeah, for example, `t****@r*****.c**`

purchases/src/main/kotlin/com/revenuecat/purchases/WebPurchaseRedemption.kt
tonidero: Hmm could be... It's internal so we can really change it whenever... I think it might be better to leave it as is so we know the type of token it is, not a strong opinion though

tonidero:
Will merge this. Everything is marked as experimental APIs and functionality hidden behind a feature flag right now.

PR #1900, opened 2024-11-05
[Fix] Consider a network error as not successfully synced for paywall events

MarkVillacampa:
### Checklist

### Motivation

While investigating missing impression events, I found a possible logic issue on Android where we are considering a network error as "successfully synced" and hence deleting any pending events.


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -524,7 +524,7 @@ internal class Backend(
                 synchronized(this@Backend) {
                     paywallEventsCallbacks.remove(paywallEventRequest.cacheKey)
                 }?.forEach { (_, onErrorHandler) ->
-                    onErrorHandler(error, true)
+                    onErrorHandler(error, false)
```
tonidero, line 527: I can't think of any reason why we should consider network errors as synced yeah... Might be worth adding some tests for this but this looks good!

tonidero (approved):

PR #1922, opened 2024-11-19
Ensure the correct error message is shown when failing to open a Uri in paywalls

JayShortway:
Quick follow-up to https://github.com/RevenueCat/purchases-android/pull/1904.

## Motivation
We have 2 error cases when opening Uris:
- "Link is invalid"
- "No browser installed" 

However, if the app includes a more recent version of Compose, `UriHandler` throws the same exception (`IllegalArgumentException`) for both of these cases ([source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/AndroidUriHandler.android.kt;l=36;bpv=1;bpt=0;drc=17632dc3b7a9426fa96f93d4dd1d6f119a24da35;dlc=4d53400eca9f3ac90c3a3f6cffcbc5bf492ec536)), which means we never actually show the "No browser installed" error message. 

This PR fixes that by no longer relying on `UriHandler`, and implementing the (very simple) logic ourselves. I think this is justified, as we cannot know which version of `UriHandler` we get at runtime.


vegaro (approved):

JayShortway:
@vegaro I requested re-review as my last commit changed some stuff around again.

vegaro (approved):

PR #1925, opened 2024-11-20
[Paywalls V2] Adds `SizeConstraints`, `Size`, `Padding` and `Shadow` properties

JayShortway:
As the title says.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Padding.kt
```diff
@@ -0,0 +1,31 @@
+package com.revenuecat.purchases.paywalls.components.properties
+
+import kotlinx.serialization.Serializable
+
+/**
+ * Contains padding values for 4 axes, in dp.
+ */
+@Serializable
+internal data class Padding(
```
vegaro, line 9: I think all these internal should have `@JvmSynthetic` to prevent exposing them into Java

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Padding.kt
JayShortway: Ugh you're right! It gets a red underline in Java, but it compiles fine. 😓 

However, `@JvmSynthetic` cannot be applied to classes or constructors. Not sure if there's anything we can do. (Applying to the file doesn't work.)

```kotlin
@Target(AnnotationTarget.FILE, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER, AnnotationTarget.FIELD)
@Retention(AnnotationRetention.SOURCE)
public actual annotation class JvmSynthetic
```

JayShortway: Will merge for now, as the rest of the stack added way more `internal` types. So if we come up with any solution, I'll apply it to the top of the stack.

PR #1935, opened 2024-11-22
[Paywalls V2] Adds `PaywallComponentsData`

JayShortway:
This PR adds: 
- `PaywallComponentsData`
- `ComponentsConfig`
- `PaywallComponentsConfig`
- `Background`
- `LocalizationData` with a custom serializer
- tests


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/Localization.kt
```diff
@@ -0,0 +1,58 @@
...
+}
+
+private object LocalizationDataSerializer : KSerializer<LocalizationData> {
+    // Documentation says to use either PrimitiveSerialDescriptor or buildClassSerialDescriptor. However, we need a
+    // polymorphic descriptor that's either a primitive (string) or a class. So neither of those options fit the bill.
+    @OptIn(InternalSerializationApi::class, ExperimentalSerializationApi::class)
+    override val descriptor: SerialDescriptor = buildSerialDescriptor(
+        serialName = "LocalizationData",
+        kind = PolymorphicKind.SEALED,
+    )
```
JayShortway, line 44: I get the feeling that kotlinx.serialization doesn't quite "agree" with what we're doing, which is the same JSON key sometimes pointing to a primitive (string) and sometimes to a class (`ThemeImageUrls`). That's why I had to add these annotations to make it work.

joshdholtz: Ahhhh yeaaah, I can see this not working well in some places 😬 It felt surpringly well in Swift but 🤔 I can't see us adding more than just text or images though. Do you think this is still okay to go this way?

joshdholtz (approved):
:shipit:

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/Localization.kt
JayShortway: As discussed, we'll explore changing the schema for this. I'm going to merge this now and will update the SDK separately.

PR #1936, opened 2024-11-22
[Paywalls V2] `LocalizationKey` is an inline value class now.

JayShortway:
Small adjustment to the `LocalizationKey` type, making it an [inline value class](https://kotlinlang.org/docs/inline-classes.html), giving us a bit more type safety without the runtime cost. `LocaleId` is already an inline value class.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/Localization.kt
```diff
@@ -16,7 +16,9 @@ import kotlinx.serialization.encoding.Encoder
 @JvmInline
 internal value class LocaleId(val value: String)
 
-internal typealias LocalizationKey = String
+@Serializable
+@JvmInline
+internal value class LocalizationKey(val value: String)
```
joshdholtz, line 21: I totally know what a `@JvmInline` is... yup yup yup 😅

joshdholtz (approved):
:shipit:

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/Localization.kt
JayShortway: Haha, this instructs the Kotlin compiler to make it a [`value class`](https://openjdk.org/projects/valhalla/) from a Java perspective.

PR #1937, opened 2024-11-22
[Paywalls V2] Moves any non-component file to a new `common` package.

JayShortway:
This only moves stuff around. No code changes at all.

It moves:
- `Background.kt`
- `ComponentOverrides.kt`
- `ComponentsConfig.kt`
- `Localization.kt`
- `PaywallComponentsData.kt`
- and their corresponding tests

to a new `paywalls.components.common` package. This way `paywall.components` only contains actual components.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PartialComponent.kt
```diff
@@ -0,0 +1,6 @@
+package com.revenuecat.purchases.paywalls.components
+
+/**
+ * Marker interface for partial components.
+ */
+internal sealed interface PartialComponent
```
JayShortway, line 6: This used to be in `ComponentOverrides.kt`, but it has to remain in the `paywalls.components` package, where its implementing classes live (in `ImageComponent.kt` and `TextComponent.kt`).

joshdholtz (approved):
:shipit:

PR #1939, opened 2024-11-25
[Paywalls V2] Makes `TextComponent` public

JayShortway:
## Public
This makes the `TextComponent` type, and all types it references, public. This allows us to use these types from `revenuecatui`. Note that they are still marked as `@InternalRevenueCatAPI`, so using them requires explicit opt-in. I still kept constructors internal for now. Will open things up more as we need it.

## Data classes
We decided a while back that we don't want to use `data class`es in our public API, because they increase the API surface substantially (with `copy()` and `componentN()` functions), and are [hard to evolve in a binary-compatible way](https://jakewharton.com/public-api-challenges-in-kotlin/).

However, having property-based `equals()` and `hashCode()` (and `toString()`) functions is still highly desirable. While these can be generated in Android Studio (CMD + N), the extra boilerplate adds substantial cognitive load when reading. 

For this reason, this PR adds a compiler plugin called [Poko](https://github.com/drewhamilton/Poko), which generates these functions for us. It seems to be well maintained. This gives us the benefits of these functions, without the drawbacks of boilerplate. Let me know what you think.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Padding.kt
```diff
@@ -26,8 +28,8 @@ internal data class Padding(
      */
     val trailing: Double,
 ) {
-    companion object {
-        val zero = Padding(0.0, 0.0, 0.0, 0.0)
-        val default = Padding(10.0, 10.0, 20.0, 20.0)
+    internal companion object {
+        internal val zero = Padding(0.0, 0.0, 0.0, 0.0)
+        internal val default = Padding(10.0, 10.0, 20.0, 20.0)
```
vegaro, line 33: do these need the internal as well? I think it will be implicit

vegaro (approved):

joshdholtz (approved):
This looks good to me 🤷‍♂️

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Padding.kt
JayShortway: You used to be able to still call members of a non-`public` companion object from Java, but I think that had to do with `private` companion objects and not `internal` ones. 

However, marking the fields as `internal` is not sufficient either to hide it from Java completely, as `internal` things become public, albeit with [mangled names](https://kotlinlang.org/docs/java-to-kotlin-interop.html#visibility). (This became `getDefault$purchases_defaultsRelease()`.)

I experimented some more and went with your earlier suggestion of using `@JvmSynthetic` (https://github.com/RevenueCat/purchases-android/pull/1939/commits/94003a4528341f4615e2e8619cf267404282da50). This achieves the goal of hiding these fields from Java completely. Thanks for making me double check this!

PR #1942, opened 2024-11-27
[Paywalls V2] Minimizes Java API

JayShortway:
Adds `@JvmSynthetic` to all public and internal Paywall Components fields to minimize the Java API. (Kotlin's `internal` visibility becomes `public` in Java.)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Size.kt
```diff
@@ -29,6 +29,6 @@ sealed interface SizeConstraint {
     @Serializable
     @SerialName("fixed")
     class Fixed internal constructor(
-        val value: UInt,
+        @get:JvmSynthetic val value: UInt,
```
vegaro, line 32: shouldn't the constructor be internal? Isn't that what's internal in kotlin?

vegaro (approved):

PR #1943, opened 2024-11-28
Uses `Sequence` instead of `Stream` to avoid errors due to unavailable Java 8 APIs

JayShortway:
## Motivation
`Stream` is a Java 8 API, which needs [desugaring](https://developer.android.com/studio/write/java8-support) to function correctly across devices and API levels. This can lead to crashes such as https://github.com/RevenueCat/purchases-flutter/issues/1213, which is only surfacing now because we fixed error-swallowing in https://github.com/RevenueCat/purchases-android/pull/1894.

## Changes
Fortunately, Kotlin's `Sequence` is a near drop-in replacement for `Stream`. This PR implements that replacement.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/FileHelper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/FileHelper.kt
@@ -9,7 +9,6 @@ import java.io.File
 import java.io.FileInputStream
 import java.io.FileOutputStream
 import java.io.InputStreamReader
-import java.util.stream.Stream
 
 internal class FileHelper(
     private val applicationContext: Context,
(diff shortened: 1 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/FileHelper.kt
```diff
@@ -34,21 +33,21 @@ internal class FileHelper(
         return file.delete()
     }
 
-    // This is using a lambda with a Stream instead of returning the Stream itself. This is so we keep
-    // the responsibility of closing the bufferedReader to this class. Note that the Stream should
-    // be used synchronously, otherwise the bufferedReader will be closed before the stream is used.
+    // This is using a lambda with a Sequence instead of returning the Sequence itself. This is so we keep
+    // the responsibility of closing the bufferedReader to this class. Note that the Sequence should
+    // be used synchronously, otherwise the bufferedReader will be closed before the Sequence is used.
     @RequiresApi(Build.VERSION_CODES.N)
```
tonidero, line 39: I believe this is probably not needed anymore, as you mentioned over slack.

tonidero (approved):
This is great! Maybe just removing the required version annotations... Tested it and seems to work as expected 💪

purchases/src/main/kotlin/com/revenuecat/purchases/common/FileHelper.kt
JayShortway: Alright, will create a follow-up PR to clean these up if we decide they're not needed.

JayShortway: For reference: https://github.com/RevenueCat/purchases-android/pull/1944

PR #1944, opened 2024-11-28
Removes `@RequiresApi(N)` from `FileHelper` and related classes

JayShortway:
As the title says. Just a bit of cleanup as these annotations are no longer needed. They were probably added because we were using `Stream`, but our collective memory is fuzzy. (We're no longer using `Stream` as of #1943.)


tonidero (approved):
Thanks for the cleanup!

PR #1946, opened 2024-11-28
Increase integration test timeout

tonidero:
### Description
Integration tests have been failing pretty frequently. From some tests, it seems that verifying the purchase token in the backend after a while takes a bit longer, which makes it frequently go over the timeout limit of 5s set for purchasing. This increases the timeout so it has time to finish the purchase on those situations.

### DIFF
--- a/purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
+++ b/purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
@@ -44,7 +44,7 @@ open class BasePurchasesIntegrationTest {
     protected open val initialForceServerErrors: Boolean = false
     protected open val initialForceSigningErrors: Boolean = false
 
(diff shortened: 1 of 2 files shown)


purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
```diff
@@ -44,7 +44,7 @@ open class BasePurchasesIntegrationTest {
     protected open val initialForceServerErrors: Boolean = false
     protected open val initialForceSigningErrors: Boolean = false
 
-    protected val testTimeout = 5.seconds
+    protected val testTimeout = 10.seconds
```
tonidero, line 47: I thought to only increase it for the tests that were actually failing more often, but we have a couple of those in different flavors, so I thought to just increase it across the board. Lmk what you think

JayShortway (approved):

PR #1949, opened 2024-11-29
Modify intent extension API to parse redemption links to mimick iOS

tonidero:
### Description
This modifies the `Intent.parseAsWebPurchaseRedemption()` method to `Intent.asWebPurchaseRedemption()` following the naming of the API in the iOS SDK, so we consolidate them. This is a breaking change but this is an experimental API not open to developers yet so it should be safe.

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/IntentExtensions.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/IntentExtensions.kt
@@ -3,6 +3,7 @@ package com.revenuecat.purchases
 import android.content.Intent
 
 @ExperimentalPreviewRevenueCatPurchasesAPI
-val Intent.asWebPurchaseRedemption: WebPurchaseRedemption?
-    @JvmSynthetic
-    get() = Purchases.parseAsWebPurchaseRedemption(this)
+@JvmSynthetic
+fun Intent.asWebPurchaseRedemption(): WebPurchaseRedemption? {
+    return Purchases.parseAsWebPurchaseRedemption(this)
+}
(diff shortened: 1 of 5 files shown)


purchases/src/defaults/kotlin/com/revenuecat/purchases/IntentExtensions.kt
```diff
@@ -3,7 +3,6 @@ package com.revenuecat.purchases
 import android.content.Intent
 
 @ExperimentalPreviewRevenueCatPurchasesAPI
-@JvmSynthetic
-fun Intent.parseAsWebPurchaseRedemption(): WebPurchaseRedemption? {
-    return Purchases.parseAsWebPurchaseRedemption(this)
-}
+val Intent.asWebPurchaseRedemption: WebPurchaseRedemption?
```
JayShortway, line 6: I like the name! But it does feel to me that this should be a function rather than a property. We are converting the object, so something is "happening", and a `WebPurchaseRedemption` is not really a property of the broad `Intent` type. Just my 2 cents. What do you think?

JayShortway (commented):
Consistency is 🔑! Just 1 comment on function vs property.

purchases/src/defaults/kotlin/com/revenuecat/purchases/IntentExtensions.kt
tonidero: Hmm I agree with you... I'm just doubting since it's a property in iOS... I guess we don't really need to copy it exactly though, so I can make the change 👍

JayShortway: We can make it a function in iOS too 😈

JayShortway (approved):
Nice, thanks for making the change! 🙏

purchases/src/defaults/kotlin/com/revenuecat/purchases/IntentExtensions.kt
JayShortway: Ah yes, let's stay consistent with what's there already. Thanks for checking!

PR #1951, opened 2024-12-02
[Paywalls V2] Add support for markdown in text component

tonidero:
### Description
This adds support for Markdown inside the text component of paywalls V2. Here you can see the text component previews before (top) and after (bottom)
Screenshot 2024-12-02 at 18 32 43

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
@@ -2,10 +2,10 @@
 
 package com.revenuecat.purchases.ui.revenuecatui.components.text
 
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.material3.LocalTextStyle
-import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -26,6 +26,7 @@ import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fi
 import com.revenuecat.purchases.ui.revenuecatui.components.ColorStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.modifier.background
 import com.revenuecat.purchases.ui.revenuecatui.components.modifier.size
+import com.revenuecat.purchases.ui.revenuecatui.composables.Markdown
 import com.revenuecat.purchases.ui.revenuecatui.extensions.conditional
 
 @Composable
@@ -47,7 +48,7 @@ private fun TextComponentView(
     }
 
     if (style.visible) {
-        Text(
+        Markdown(
             text = style.text,
             modifier = modifier
                 .size(style.size, horizontalAlignment = style.horizontalAlignment)
@@ -58,6 +59,7 @@ private fun TextComponentView(
             fontSize = style.fontSize,
             fontWeight = style.fontWeight,
             fontFamily = style.fontFamily,
+            horizontalAlignment = style.horizontalAlignment,
             textAlign = style.textAlign,
             style = textStyle,
         )
@@ -156,6 +158,18 @@ private fun TextComponentView_Preview_Customizations() {
             backgroundColor = ColorScheme(light = ColorInfo.Hex(Color(red = 0xde, green = 0xde, blue = 0xde).toArgb())),
             padding = Padding(top = 10.0, bottom = 10.0, leading = 20.0, trailing = 20.0),
             margin = Padding(top = 20.0, bottom = 20.0, leading = 10.0, trailing = 10.0),
+        ),
+    )
+}
+
+@Preview(name = "Default")
+@Composable
+private fun TextComponentView_Preview_Markdown() {
+    TextComponentView(
+        style = previewTextComponentStyle(
+            text = "Hello, **bold**, *italic* or _italic2_ with ~strikethrough~ and `monospace`. " +
+                "Click [here](https://revenuecat.com)",
+            color = ColorScheme(light = ColorInfo.Hex(Color.Black.toArgb())),
         ),
     )
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
@@ -62,6 +62,11 @@ private val parser = Parser.builder()
     .extensions(listOf(StrikethroughExtension.create()))
     .build()
 
+/**
+ * @param allowLinks If true, links will be decorated and clickable.
+ * @param textFillMaxWidth If true, the text will fill the maximum width available. This was used by paywalls V1 and
+ * left to avoid unintended UI changes.
+ */
 @SuppressWarnings("LongParameterList")
 @Composable
 internal fun Markdown(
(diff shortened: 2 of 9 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -67,9 +69,13 @@ internal fun Markdown(
     modifier: Modifier = Modifier,
     color: Color = Color.Unspecified,
     style: TextStyle = TextStyle.Default,
+    fontSize: TextUnit = TextUnit.Unspecified,
     fontWeight: FontWeight? = null,
+    fontFamily: FontFamily? = null,
+    horizontalAlignment: Alignment.Horizontal = Alignment.Start,
     textAlign: TextAlign? = null,
     allowLinks: Boolean = true,
+    textFillMaxWidth: Boolean = false, // This is to support V1 paywalls
```
tonidero, line 78: This is a bad hack... but I was running into issues... Basically `Markdown` was for some reason making each text use the `fillMaxWidth` modifier. This was probably wrong... When I removed it and I tried to fix it in all the templates I thought that in some cases, the view can potentially be modified with the change... So I wanted to avoid the UI change. The result is this flag to modify the behavior. We can remove this and revisit all the old templates and do what we think is right, but this was the "least risky" change. Lmk what you think.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -84,9 +90,10 @@ internal fun Markdown(
 
     Column(
         verticalArrangement = Arrangement.spacedBy(paragraphPadding),
+        horizontalAlignment = horizontalAlignment,
```
tonidero, line 93: We are also passing the horizontal alignment to the modifier, but I had to pass it here as well. Reason is that, if the content size is set to `Fill`, the horizontal alignment was previously mostly ignored. I think this might be a better place for it. I thought of removing it from the modifier, but left it for now just in case

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -96,12 +103,24 @@ private fun MDDocument(
     document: Document,
     color: Color,
     style: TextStyle,
+    fontSize: TextUnit,
     fontWeight: FontWeight?,
+    fontFamily: FontFamily?,
     textAlign: TextAlign?,
     allowLinks: Boolean,
-    modifier: Modifier = Modifier,
```
tonidero, line 102: Note that I've removed the `modifier` all over the place in here... Reason is that, after adding the new properties, I had to add make the composables use multiple lines, and for some unexplained reason, that caused detekt to find a bunch of new issues related to reusing the same `modifier` instance in multiple views, which is disallowed. Since we weren't using them anyway, I thought to just remove them where I could.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
```diff
@@ -98,6 +98,7 @@ private fun ColumnScope.Template1MainContent(state: PaywallState.Loaded) {
                 fontWeight = FontWeight.Black,
                 textAlign = TextAlign.Center,
                 color = colors.text1,
+                textFillMaxWidth = true,
```
tonidero, line 101: All the old templates will keep this to `true`. For Paywalls V2, the default will be false.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -388,10 +540,14 @@ private fun MarkdownText(
...
         style = style,
+        fontSize = fontSize,
         fontWeight = fontWeight,
+        fontFamily = fontFamily,
         textAlign = textAlign,
         modifier = modifier
-            .fillMaxWidth()
+            .conditional(textFillMaxWidth) {
+                fillMaxWidth()
+            }
```
tonidero, line 550: This is what brought me a headache in this change... Tried to minimize the visual changes with this.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -134,7 +166,17 @@ private fun MDHeading(
         val text = buildAnnotatedString {
             appendMarkdownChildren(heading, color, allowLinks)
         }
-        MarkdownText(text, color, overriddenStyle, fontWeight, textAlign, allowLinks, modifier)
+        MarkdownText(
+            text,
+            color,
+            overriddenStyle,
+            TextUnit.Unspecified,
```
tonidero, line 173: Note that I'm not passing the font size here... This is actually for the headings that we decided not to support for now. In any case, I decided to, instead of hardcode the size to what's given, keep using the one from the app's theme.

tonidero (commented):
I'm quite annoyed with the `textFillMaxWidth` parameter I added, but I thought it was the least risky option... Lmk if you have other thoughts!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
joshdholtz: This seems fine to me 🤷‍♂️

joshdholtz (approved):
This looks good to me! 😁

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
```diff
@@ -160,6 +162,18 @@ private fun TextComponentView_Preview_Customizations() {
...
+@Preview(name = "Default")
+@Composable
+private fun TextComponentView_Preview_Markdown() {
+    TextComponentView(
+        style = previewTextComponentStyle(
+            text = "Hello, **bold**, *italic* or _italic2_ with ~strikethrough~ and `monospace`. " +
+                "Click [here](https://revenuecat.com)",
+            color = ColorScheme(light = ColorInfo.Hex(Color.Black.toArgb())),
+        ),
+    )
```
JayShortway, line 174: 🙌

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
JayShortway: Maybe we can add a doc to the `Markdown` function explaining this parameter? Something like:

```kotlin
/**
 * @param textFillMaxWidth <explanation>
 */
internal fun Markdown(
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
JayShortway: Will it ever be true for Paywalls V2? If so, when?

JayShortway (commented):
Super nice! Just a request for docs for the new parameter.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/templates/Template1.kt
tonidero: Probably not, since I think the correct behavior is to have the markdown just use the required space. I guess it could be used if for some reason we wanted all texts in the markdown to fill the available space... But it's not likely IMO.

JayShortway: Alright, I think so too. Thanks for explaining!

PR #1952, opened 2024-12-03
[Paywalls V2] Adds support for shadows

JayShortway:
The main addition is `WithShadow`, which allows adding shadows using the properties we get from the response (radius, x and y offset). These properties are different from the `elevation` property expected by [the built-in `.shadow()` modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#(androidx.compose.ui.Modifier).shadow(androidx.compose.ui.unit.Dp,androidx.compose.ui.graphics.Shape,kotlin.Boolean,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color)). 

We could have built something similar using a `.drawBehind()` modifier. However our spec technically allows for gradient shadows, and I had a hard time figuring out how to draw a gradient on the canvas. Pragmatism! 

This also adds a new `properties` package and moves the existing `ColorStyle` to it. 

**Note:** [Emerge](https://www.emergetools.com/snapshot/fb33e583-4058-4c44-9217-536d24ce2799?tab=added) seems to ignore the blur when rendering the shadows. Android Studio previews for reference: 

image

(Top-left is intentionally unblurred.)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Shadow.kt
```diff
@@ -5,7 +5,7 @@ import kotlinx.serialization.Serializable
 
 @InternalRevenueCatAPI
 @Serializable
-internal data class Shadow(
+data class Shadow(
```
tonidero, line 8: Should this also use `Poko` instead of being a data class?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/composable/WithShadow.kt
```diff
@@ -0,0 +1,266 @@
...
+    // We cannot use the built-in .shadow() modifier, as it takes very different parameters from what we have in our
+    // ShadowStyle. WithShadow also can't be a modifier itself, because the modifiers we use to draw the shadow
+    // (offset, blur and background) cannot be part of the same modifier chain as those applied to the content. If they
+    // are, they will be combined in ways we don't want.
+    // We might be able to achieve a similar result using a .drawBehind() modifier, but we would need to figure out how
+    // to draw gradients on a canvas.
+    Layout(
+        content = {
+            // Content
+            content()
```
tonidero, line 63: Just to make sure, will `Layout` put the `content` "on top" of the shadow?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/composable/WithShadow.kt
```diff
@@ -0,0 +1,266 @@
...
+
+            // Shadow
+            Box(
+                Modifier
+                    .offset { IntOffset(x = shadow.x.roundToPx(), y = shadow.y.roundToPx()) }
+                    .blur(shadow.radius, edgeTreatment = BlurredEdgeTreatment.Unbounded)
```
tonidero, line 69: Hmm I wonder if blurring could have some performance issues in older devices...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/composable/WithShadow.kt
```diff
@@ -0,0 +1,266 @@
...
+    margin: PaddingValues = PaddingValues(all = 0.dp),
+    content: @Composable () -> Unit,
+) {
+    // We cannot use the built-in .shadow() modifier, as it takes very different parameters from what we have in our
+    // ShadowStyle. WithShadow also can't be a modifier itself, because the modifiers we use to draw the shadow
+    // (offset, blur and background) cannot be part of the same modifier chain as those applied to the content. If they
+    // are, they will be combined in ways we don't want.
+    // We might be able to achieve a similar result using a .drawBehind() modifier, but we would need to figure out how
+    // to draw gradients on a canvas.
+    Layout(
```
tonidero, line 60: Should we skip the whole `Layout` composable if radius is 0?

tonidero (approved):
LGTM! Just some questions

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/composable/WithShadow.kt
JayShortway: Yes, that happens below, where we first place the `shadowPlaceable` and then the `contentPlaceable`.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Shadow.kt
JayShortway: Good catch! https://github.com/RevenueCat/purchases-android/pull/1952/commits/ae8140c9634e7bca722714bf24a16f20204c946f

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/composable/WithShadow.kt
JayShortway: No because you can still have a shadow without blur. It would be a very sharp shadow. See the top-left circle in the PR description.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/composable/WithShadow.kt
JayShortway: Here it is: https://github.com/RevenueCat/purchases-android/pull/1957.

PR #1953, opened 2024-12-03
[Paywalls V2] Adds borders and image backgrounds

JayShortway:
As the title says. Supports gradient borders because why not. Also adds a `ColorScheme.colorsForCurrentTheme` extension in the same spirit as `ThemeImageUrls.urlsForCurrentTheme`.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Background.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Background.kt
@@ -8,7 +8,7 @@ import kotlinx.serialization.Serializable
 
 @InternalRevenueCatAPI
 @Serializable
-internal sealed interface Background {
+sealed interface Background {
     @Serializable
     @SerialName("color")
     data class Color(@get:JvmSynthetic val value: ColorScheme) : Background
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ColorScheme.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ColorScheme.kt
@@ -1,0 +1,11 @@
+@file:JvmSynthetic
+
+package com.revenuecat.purchases.ui.revenuecatui.components.ktx
+
+import androidx.compose.foundation.isSystemInDarkTheme
+import androidx.compose.runtime.Composable
+import com.revenuecat.purchases.paywalls.components.properties.ColorInfo
+import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
+
+internal val ColorScheme.colorsForCurrentTheme: ColorInfo
+    @Composable get() = if (isSystemInDarkTheme()) dark ?: light else light
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
@@ -1,0 +1,134 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.properties
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.requiredSize
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import coil.compose.rememberAsyncImagePainter
+import com.revenuecat.purchases.paywalls.components.common.Background
+import com.revenuecat.purchases.paywalls.components.properties.ColorInfo
+import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.urlsForCurrentTheme
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.background
+
+/**
+ * Ready to use background properties for the current theme.
+ */
+internal sealed interface BackgroundStyle {
+    @JvmInline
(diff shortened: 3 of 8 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
```diff
@@ -0,0 +1,134 @@
...
+                        model = imageUrls.webpLowRes,
+                        error = null,
+                        fallback = null,
+                        contentScale = ContentScale.Crop,
+                    ),
+                    error = null,
+                    fallback = null,
+                    contentScale = ContentScale.Crop,
+                ),
+            )
```
JayShortway, line 51: I'm using the low-res variant as a placeholder for now. We can think about supporting [blurhash](https://blurha.sh/), but ideally the images are cached already before the paywall is shown.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ColorScheme.kt
```diff
@@ -0,0 +1,11 @@
+@file:JvmSynthetic
```
tonidero, line 1: Hmm this is going to be internal right? I think we might not need this? As in, I would probably only add it on files that will contain public APIs but we don't want to be accessible from Java. Wdyt?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
tonidero: Hmm I'm wondering if we should have a bundled placeholder (or background basically 😅. But yeah hopefully this will be enough.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Background.kt
```diff
@@ -8,7 +8,7 @@ import kotlinx.serialization.Serializable
 
 @InternalRevenueCatAPI
 @Serializable
-internal sealed interface Background {
+sealed interface Background {
```
tonidero, line 11: Hmm I'm wondering if we should namespace this one, since it's too generic... I believe we haven't been doing that for many of these, but I think we might be including a lot of API at the top level IMO...

tonidero (approved):
A few comments but nothing really blocking this PR. Looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ColorScheme.kt
JayShortway: Anything internal is public in Java, because Java doesn't have the internal keyword. So that's why I've been adding it to internal APIs as well. The `@file:` scope is [supposedly avoiding the creation of a facade class](https://youtrack.jetbrains.com/issue/KT-41884) (`ColorSchemeKt`) that would otherwise be generated.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Background.kt
JayShortway: Right, but Kotlin has packages so namespacing isn't as needed as in Swift imo. The compiler makes sure we use the right type. Which kind of problems do you foresee?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
JayShortway: Yea indeed, something to think about!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Background.kt
tonidero: Mostly autocomplete related problems... As in, if someone is writing in their code a background related class, this would appear more easily in their autocomplete list I think? 

Naming collisions could also happen if they have a `Background` type and they would need to use import aliases (not that developers should ever need to use this in general so I'm less concerned about this.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ColorScheme.kt
JayShortway: Internal symbols go through name mangling in Java, so something like this:
```kotlin
internal val default: Padding
```

becomes: 

```java
public Padding getDefault$purchases_defaultsRelease()
```

I think we should avoid that as much as possible, even if we're already doing this in some places. (The name can be fixed with `@JvmName`, but why not hide it completely at that point.)

>this reduces readability

I agree, but that's our problem. I'm afraid that we make it the developer's problem if we leak internal symbols to Java, which is aggravated by the name mangling. 

>why isn't the kotlin compiler doing this automatically upon finding any internal class/property? 

I'm assuming because you might want to access internal symbols from Java code in the same module. That would be impossible if everything internal is automatically hidden from Java.

For reference, the documentation on this behavior is [here](https://kotlinlang.org/docs/java-to-kotlin-interop.html#visibility).

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Background.kt
tonidero: Yeah I've recently had this issue with compose's Color vs android graphics color for example 😅... I think we can probably go ahead for now with this, but something to consider

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ColorScheme.kt
tonidero: Yeah, I've added this topic to our weekly so we can discuss it... I'm not sure it's worth the extra overhead to try to hide these APIs from Java...

PR #1954, opened 2024-12-03
[Paywalls V2] Preparation for `StackComponentView`

JayShortway:
This PR adds various preparations for the upcoming `StackComponentView`. The main ones:
- Adds a new `sealed interface ComponentStyle` in a new `style` package, and makes `TextComponentStyle` implement it. This means `TextComponentStyle` (and all upcoming styles) need to be in the same `style` package (because `ComponentStyle` is `sealed`). This will allow `StackComponentView` to iterate over the "children styles", and use an exhaustive `when` to figure out which child composable to call. 
- Makes a few necessary things fake-public in purchases.
- Adds alignment and distribution mapping functions.


tonidero (approved):
LGTM!

PR #1956, opened 2024-12-03
[Paywalls V2] Adds `StackComponentView`

JayShortway:
As the title says. Also adds `StackComponentStyle`.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StackComponentStyle.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.revenuecat.purchases.paywalls.components.properties.Dimension
+import com.revenuecat.purchases.paywalls.components.properties.Size
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.BackgroundStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.BorderStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.ShadowStyle
+
+@Suppress("LongParameterList")
+@Immutable
+internal class StackComponentStyle(
+    @get:JvmSynthetic
```
tonidero, line 16: Mentioned this in another PR but same here. I think we could remove these synthetic annotations for internal code, it would be more readable as well IMO. Wdyt?

tonidero (approved):
Just a comment about the use of synthetics that I also mentioned in another PR, but looks great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StackComponentStyle.kt
```diff
@@ -0,0 +1,38 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.style
```
tonidero, line 1: Hmm thinking about this, should this also go in the `stack` package? Seems like a better place to keep the same types together, and it's how we have the `TextComponentView`/`TextComponentStyle`.

tonidero: Ohh I see we moved the styles in the base PR... I'm wondering if it would make sense to keep them together though... Personally, I think that's better to keep the related files together.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -0,0 +1,269 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyle.Solid
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.ShadowStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.style.StackComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.style.TextComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.text.TextComponentView
+import com.revenuecat.purchases.ui.revenuecatui.extensions.applyIfNotNull
+
+@Suppress("LongMethod")
+@Composable
+internal fun StackComponentView(
```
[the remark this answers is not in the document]
tonidero: So I should read things more thoroughly. I was just checking that `StackComponent` and `TextComponent` are the model holders that come from the main SDK... I think it's not a good practice to name composables as `...View`, but not sure what to name them then... Maybe it's ok to just continue with this for now, since it's not public facing so we can change it later if needed.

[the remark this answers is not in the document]
JayShortway: Yea I agree the names could be better, but it's private indeed so we can always change it later.

It's a bit unfortunate that the iOS SDK used `StackComponent` (etc) for the JSON models. That would indeed have been a better name for the composables, but now I think it would be too confusing to have the same name mean different things. 

Another not-perfect option is to name the composables `Stack` (etc). However that becomes slightly confusing for things like `Text`, `Button` and `Package`. 

We can think about this a bit longer.

PR #1957, opened 2024-12-03
[Paywalls V2] Extends support for blurred shadows to all Android versions

JayShortway:
Alternative title: Removes `WithShadow` in favor of a `.shadow()` modifier. (I figured it out yay 🎉) 

This uses [`BlurMaskFilter`](https://developer.android.com/reference/android/graphics/BlurMaskFilter), which has been available since API level 1, so this `.shadow()` modifier now works as expected across all API levels.


tonidero (approved):
Nice!! Great job on this 💪 !

PR #1958, opened 2024-12-04
Avoid generating documentation for symbols annotated with `@InternalRevenueCatAPI`

JayShortway:
## Motivation
We don't want to generate documentation for anything annotated with `@InternalRevenueCatAPI`, as those symbols are not part of the public API. 

## Changes
This adds a new Dokka plugin, `dokka-hide-internal`, which suppresses documentation for anything annotated with `@InternalRevenueCatAPI`. It is made by following [the plugin tutorial](https://kotlin.github.io/dokka/1.8.10/developer_guide/plugin-development/sample-plugin-tutorial), which covers exactly this use case. All of this can be removed if https://github.com/Kotlin/dokka/issues/1861 gets implemented.


dokka-hide-internal/src/main/kotlin/com/revenuecat/dokka/plugin/hideinternal/HideInternalRevenueCatAPIPlugin.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.jetbrains.dokka.base.transformers.documentables.SuppressedByConditionDocumentableFilterTransformer
+import org.jetbrains.dokka.model.Annotations
+import org.jetbrains.dokka.model.Documentable
+import org.jetbrains.dokka.model.properties.WithExtraProperties
+import org.jetbrains.dokka.plugability.DokkaContext
+import org.jetbrains.dokka.plugability.DokkaPlugin
+import org.jetbrains.dokka.plugability.DokkaPluginApiPreview
+import org.jetbrains.dokka.plugability.PluginApiPreviewAcknowledgement
+
+class HideInternalRevenueCatAPIPlugin : DokkaPlugin() {
```
vegaro, line 13: I guess this has to be public?

vegaro (approved):

dokka-hide-internal/src/main/kotlin/com/revenuecat/dokka/plugin/hideinternal/HideInternalRevenueCatAPIPlugin.kt
JayShortway: I think so yes, but it's not published to Maven (unless I'm overlooking something). It's just used during docs generation, which happens locally ("locally" on CI).

JayShortway: Did some experiments, and it has to be public indeed. It's using Java's ServiceLoader mechanism to discover the plugin class, and it can't find it if it's private. Internal works, but that's probably because ServiceLoaders are a Java thing, and internal in Kotlin = public in Java.

PR #1959, opened 2024-12-04
[Paywalls V2] Adds `ImageComponentView`

tonidero:
### Description
This adds a new component `ImageComponentView` and the corresponding `ImageComponentStyle`. It doesn't include transformations from `ImageComponent` to `ImageComponentStyle`.

- Placeholder to image loading:

https://github.com/user-attachments/assets/ae010171-2ae4-4d82-b84c-68fa9015d2bd

- Previews
image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
@@ -1,0 +1,138 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.image
+
+import android.graphics.Color
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.draw.drawWithCache
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.revenuecat.purchases.paywalls.components.properties.ColorInfo
+import com.revenuecat.purchases.paywalls.components.properties.Size
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.size
+import com.revenuecat.purchases.ui.revenuecatui.components.style.ImageComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.composables.RemoteImage
+import com.revenuecat.purchases.ui.revenuecatui.extensions.applyIfNotNull
+import java.net.URL
+import androidx.compose.ui.graphics.Color as ComposeColor
+
+@Suppress("SpreadOperator", "LongParameterList")
+@JvmSynthetic
+@Composable
+internal fun ImageComponentView(
+    style: ImageComponentStyle,
+    modifier: Modifier = Modifier,
+) {
+    if (style.visible) {
+        Box(modifier = modifier.applyIfNotNull(style.shape) { clip(it) }) {
+            RemoteImage(
+                urlString = style.urlToUse().toString(),
+                modifier = Modifier
+                    .size(style.size)
+                    .applyIfNotNull(style.gradientColors()) { gradientColors ->
+                        drawWithCache {
+                            val gradientBrush: Brush? = when (gradientColors) {
+                                is ColorInfo.Gradient.Radial -> {
+                                    Brush.radialGradient(*gradientColors.points.toColorStops())
+                                }
+                                is ColorInfo.Gradient.Linear -> {
+                                    // TODO-PAYWALLS-V2: Support degrees
+                                    Brush.verticalGradient(*gradientColors.points.toColorStops())
+                                }
+                                else -> null
+                            }
+                            onDrawWithContent {
+                                drawContent()
+                                gradientBrush?.let {
+                                    drawRect(it)
+                                }
+                            }
+                        }
+                    },
+                placeholderUrlString = style.lowResUrlToUse()?.toString(),
+                contentScale = style.contentScale,
+            )
+        }
+    }
+}
+
+private fun List<ColorInfo.Gradient.Point>.toColorStops(): Array<Pair<Float, ComposeColor>> {
+    return map { point ->
+        point.percent to ComposeColor(point.color)
+    }.toTypedArray()
+}
+
+@Preview
+@Composable
+private fun ImageComponentView_Preview_Default() {
+    Box(modifier = Modifier.background(ComposeColor.Red)) {
+        ImageComponentView(
+            style = previewImageComponentStyle(),
+        )
+    }
+}
+
+@Preview
+@Composable
+private fun ImageComponentView_Preview_LinearGradient() {
+    Box(modifier = Modifier.background(androidx.compose.ui.graphics.Color.Red)) {
+        ImageComponentView(
+            style = previewImageComponentStyle(
+                gradientColors = ColorInfo.Gradient.Linear(
+                    degrees = 45f,
+                    points = listOf(
+                        ColorInfo.Gradient.Point(color = Color.parseColor("#88FF0000"), percent = 0f),
+                        ColorInfo.Gradient.Point(color = Color.parseColor("#8800FF00"), percent = 0.5f),
+                        ColorInfo.Gradient.Point(color = Color.parseColor("#880000FF"), percent = 1f),
+                    ),
+                ),
+            ),
+        )
+    }
+}
+
+@Preview
+@Composable
+private fun ImageComponentView_Preview_RadialGradient() {
+    Box(modifier = Modifier.background(androidx.compose.ui.graphics.Color.Red)) {
+        ImageComponentView(
+            style = previewImageComponentStyle(
+                gradientColors = ColorInfo.Gradient.Radial(
+                    points = listOf(
+                        ColorInfo.Gradient.Point(color = Color.parseColor("#88FF0000"), percent = 0f),
+                        ColorInfo.Gradient.Point(color = Color.parseColor("#8800FF00"), percent = 0.5f),
+                        ColorInfo.Gradient.Point(color = Color.parseColor("#880000FF"), percent = 1f),
+                    ),
+                ),
+            ),
+        )
+    }
+}
+
+@Suppress("LongParameterList")
+@Composable
+private fun previewImageComponentStyle(
+    url: URL = URL("https://sample-videos.com/img/Sample-jpg-image-5mb.jpg"),
+    lowResURL: URL? = URL("https://assets.pawwalls.com/954459_1701163461.jpg"),
+    visible: Boolean = true,
+    size: Size = Size(width = SizeConstraint.Fixed(400u), height = SizeConstraint.Fixed(400u)),
+    contentScale: ContentScale = ContentScale.Crop,
+    gradientColors: ColorInfo.Gradient? = null,
+) = ImageComponentStyle(
+    visible = visible,
+    url = url,
+    lowResURL = lowResURL,
+    darkURL = null,
+    darkLowResURL = null,
+    size = size,
+    shape = RoundedCornerShape(20.dp, 20.dp, 20.dp, 20.dp),
+    gradientColors = gradientColors,
+    darkGradientColors = null,
+    contentScale = contentScale,
+)
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
@@ -12,10 +12,13 @@ import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
-import com.revenuecat.purchases.paywalls.components.properties.ColorInfo
+import com.revenuecat.purchases.paywalls.components.properties.ImageUrls
 import com.revenuecat.purchases.paywalls.components.properties.Size
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import com.revenuecat.purchases.paywalls.components.properties.ThemeImageUrls
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.urlsForCurrentTheme
 import com.revenuecat.purchases.ui.revenuecatui.components.modifier.size
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.style.ImageComponentStyle
 import com.revenuecat.purchases.ui.revenuecatui.composables.RemoteImage
 import com.revenuecat.purchases.ui.revenuecatui.extensions.applyIfNotNull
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -0,0 +1,138 @@
...
+import com.revenuecat.purchases.paywalls.components.properties.Size
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.size
+import com.revenuecat.purchases.ui.revenuecatui.components.style.ImageComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.composables.RemoteImage
+import com.revenuecat.purchases.ui.revenuecatui.extensions.applyIfNotNull
+import java.net.URL
+import androidx.compose.ui.graphics.Color as ComposeColor
+
+@Suppress("SpreadOperator", "LongParameterList")
```
tonidero, line 25: I tried to find an alternative to the SpreadOperator. It's not recommended by detekt to convert from a list/array to vararg due to performance concerns since it might copy all elements... But the number of parameter should normally not be too big, so I think it's ok?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -0,0 +1,138 @@
...
+@JvmSynthetic
+@Composable
+internal fun ImageComponentView(
+    style: ImageComponentStyle,
+    modifier: Modifier = Modifier,
+) {
+    if (style.visible) {
+        Box(modifier = modifier.applyIfNotNull(style.shape) { clip(it) }) {
```
tonidero, line 33: Had to apply the clipping in an outer box... If I applied it to the image itself, it didn't work together with the gradient overlays.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -0,0 +1,138 @@
...
+                modifier = Modifier
+                    .size(style.size)
+                    .applyIfNotNull(style.gradientColors()) { gradientColors ->
+                        drawWithCache {
+                            val gradientBrush: Brush? = when (gradientColors) {
+                                is ColorInfo.Gradient.Radial -> {
+                                    Brush.radialGradient(*gradientColors.points.toColorStops())
+                                }
+                                is ColorInfo.Gradient.Linear -> {
+                                    // TODO-PAYWALLS-V2: Support degrees
```
tonidero, line 45: It's a bit tricky to support degrees in a linear gradient... Since in iOS it seems we only support vertical gradients I thought to leave it like this for now and come back to it later.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -80,17 +83,13 @@ private sealed class ImageSource {
...
     source: ImageSource,
+    placeholderSource: ImageSource.Remote?,
     modifier: Modifier = Modifier,
     contentScale: ContentScale,
     contentDescription: String?,
     transformation: Transformation?,
     alpha: Float,
 ) {
-    // Previews don't support images
-    if (isInPreviewMode()) {
```
tonidero, line 90: I noticed we could have some kind of previews by using `placeholder`. I think it should be safe as long as it's `inPreviewMode`, and more useful than the current solution.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ImageComponentStyle.kt
```diff
@@ -0,0 +1,90 @@
...
+    @get:JvmSynthetic
+    val size: Size,
+    @get:JvmSynthetic
+    val shape: Shape?,
+    @get:JvmSynthetic
+    val gradientColors: ColorInfo.Gradient?,
+    @get:JvmSynthetic
+    val darkGradientColors: ColorInfo.Gradient?,
+    @get:JvmSynthetic
+    val contentScale: ContentScale,
```
tonidero, line 35: Note that this might require some transformations from the `ImageComponent` to this `ImageComponentStyle`. For example, this expects a `ContentScale` but in `ImageComponent` we have our own `FitMode`. I believe this translation should be done separately and in advance though.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
JayShortway: If you specify a `ColorStyle` instead of `ColorInfo` in `ImageComponentStyle`, you get the correct brush right away, including degrees support (*). There's already a `ColorScheme.toColorStyle()` transformation method. 

I think that also means you can consolidate `gradientColors` and `darkGradientColors`. (Maybe it can just be named `colorOverlay`, as I believe we support solid colors as well?)

(*): degrees support is the same as on iOS right now. We want to tweak it slightly on both platforms. See [this Slack discussion](https://revenuecat.slack.com/archives/C07H5S8CFL3/p1732716458417749?thread_ts=1732554856.424879&cid=C07H5S8CFL3).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ImageComponentStyle.kt
```diff
@@ -0,0 +1,90 @@
...
+    @get:JvmSynthetic
+    val visible: Boolean,
+    @get:JvmSynthetic
+    val url: URL,
+    @get:JvmSynthetic
+    val lowResURL: URL?,
+    @get:JvmSynthetic
+    val darkURL: URL?,
+    @get:JvmSynthetic
+    val darkLowResURL: URL?,
```
JayShortway, line 25: I think we can consolidate these as well, to just keep the `url` and `lowResURL`. We have `ThemeImageUrls.urlsForCurrentTheme` to make sure they are the correct ones for the current theme.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
tonidero: Ohhh nice! I totally missed that, thanks for the heads up!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ImageComponentStyle.kt
tonidero: Ahhh nice! I also missed that 🤦 Thanks again!

tonidero: I'm just passing the `ThemeImageUrls` which already contain both the `web` and `lowResWebp` urls. Did the change in https://github.com/RevenueCat/purchases-android/pull/1959/commits/680f7c8418868de790c7c43b3d737d564e5aa492

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -0,0 +1,123 @@
...
+private fun previewImageComponentStyle(
+    url: URL = URL("https://sample-videos.com/img/Sample-jpg-image-5mb.jpg"),
+    lowResURL: URL = URL("https://assets.pawwalls.com/954459_1701163461.jpg"),
+    visible: Boolean = true,
+    size: Size = Size(width = SizeConstraint.Fixed(400u), height = SizeConstraint.Fixed(400u)),
+    contentScale: ContentScale = ContentScale.Crop,
+    colorStyle: ColorStyle? = null,
+) = ImageComponentStyle(
+    visible = visible,
+    themeImageUrls = ThemeImageUrls(light = ImageUrls(url, url, lowResURL, 200u, 200u)),
```
tonidero, line 118: I'm not sure if the sizes of the images are affecting anything... Not according to my tests 🤔

JayShortway: We can use the width and height to already give the image view the appropriate size before the image is downloaded. That way the layout won't "jump" when the image arrives. 

Since width and height are part of `ThemeImageUrls`, which is provided by `ImageComponentStyle`, I think we can remove the separate `size` from `ImageComponentStyle`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -145,23 +155,29 @@ private fun AsyncImage(
...
+            rememberAsyncImagePainter(
+                model = it.data,
+                placeholder = if (isInPreviewMode()) painterResource(R.drawable.android) else null,
+                imageLoader = imageLoader,
+                contentScale = contentScale,
+                onError = { errorState ->
+                    Logger.e("Error loading placeholder image", errorState.result.throwable)
+                },
+            )
+        } ?: if (isInPreviewMode()) painterResource(R.drawable.android) else null,
```
JayShortway, line 168: Nice!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ImageComponentStyle.kt
```diff
@@ -0,0 +1,51 @@
...
+    @get:JvmSynthetic
+    val visible: Boolean,
+    @get:JvmSynthetic
+    val urls: ThemeImageUrls,
+    @get:JvmSynthetic
+    val size: Size,
+    @get:JvmSynthetic
+    val shape: Shape?,
+    @get:JvmSynthetic
+    val colorStyle: ColorStyle?,
```
JayShortway, line 23: Mabye this can be named `overlay` to indicate how this will be used?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
JayShortway: I don't think it's being used anymore now, right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
JayShortway: You can change the `drawWithCache` modifier to draw an `Outline` instead of a `Rect`. That way we don't need the `Box`. Might be nice to turn it into a custom modifier (for readability and reusability)? Something like: 

```kotlin
@JvmSynthetic
@Stable
internal fun Modifier.overlay(
    color: ColorStyle,
    shape: Shape = RectangleShape,
): Modifier = this then drawWithCache {
    val outline = shape.createOutline(size, layoutDirection, this)

    onDrawWithContent {
        drawContent()
        when (color) {
            is ColorStyle.Solid -> drawOutline(outline, color.color)
            is ColorStyle.Gradient -> drawOutline(outline, color.brush)
        }
    }
}
```

Which can then be used as: 

```kotlin
.applyIfNotNull(style.colorStyle) { colorStyle ->
    overlay(colorStyle, style.shape ?: RectangleShape)
}
// This is now needed, as the Box is no longer clipping the image itself.
.applyIfNotNull(style.shape) { clip(it) }
```

JayShortway (commented):
Really nice job on this! Just a few minor comments.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
tonidero: Ohhh ok that makes a lot of sense 😅  Thanks!

tonidero: Hmm as I was thinking about this, how would that work if we want the image to "Fill" or "Fit" the screen? In the `ThemeImageUrls` there is no way to pass if we want one of those two right?  I think passing the specific width/height only makes sense if we want a "fixed" size...

JayShortway: Summary of offline conversation:
- We cannot remove the `size` as that indicates the size of the image view, not the image itself. 
- The `ImageUrls` width/height come into play when the image view is set to `Fit`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Size.kt
```diff
@@ -23,32 +24,45 @@ import androidx.compose.ui.Modifier
...
     verticalAlignment: Alignment.Vertical? = null,
+    imageUrls: ImageUrls? = null,
 ): Modifier {
     val widthModifier = when (val width = size.width) {
-        is Fit -> Modifier.wrapContentWidth(align = horizontalAlignment ?: Alignment.CenterHorizontally)
+        is Fit -> if (imageUrls == null) {
+            Modifier.wrapContentWidth(align = horizontalAlignment ?: Alignment.CenterHorizontally)
+        } else {
+            Modifier.width(imageUrls.width.toInt().dp)
+        }
```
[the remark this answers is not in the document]
tonidero: I like that 👍 Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/ThemeImageUrls.kt
```diff
@@ -4,8 +4,10 @@ package com.revenuecat.purchases.ui.revenuecatui.components.ktx
 
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.ReadOnlyComposable
 import com.revenuecat.purchases.paywalls.components.properties.ImageUrls
 import com.revenuecat.purchases.paywalls.components.properties.ThemeImageUrls
 
 internal val ThemeImageUrls.urlsForCurrentTheme: ImageUrls
-    @Composable get() = if (isSystemInDarkTheme()) dark ?: light else light
+    @ReadOnlyComposable @Composable
```
tonidero, line 12: Not important, but I think we can make some composables like this read only composables.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ImageComponentStyle.kt
```diff
@@ -0,0 +1,83 @@
...
+                overlay = overlay,
+                contentScale = contentScale,
+            )
+        }
+    }
+
+    @JvmSynthetic
+    @ReadOnlyComposable
+    @Composable
+    fun adjustedSize(): Size {
```
tonidero, line 63: I moved this to the `ImageComponentStyle` since it felt like a better place to put it than as simple extensions of the `Size` class, since here is where we actually have the data to adjust the size. Not sure if there are other components that would require a similar adjustment, in which case, I think we can extract it to extension functions as suggested. Wdyt?

JayShortway: I like this!

JayShortway (approved):
Super nice!

PR #1962, opened 2024-12-05
WebPurchaseRedemption: Rename `AlreadyRedeemed` result to `PurchaseBelongsToOtherUser`

tonidero:
### Description
We decided to rename this redemption result value to better represent the meaning of the result. Note that the API is marked as experimental and the feature is still not enabled in our backend, so it should be ok to rename it.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -712,7 +712,7 @@ internal class Backend(
                                     callback(RedeemWebPurchaseListener.Result.Expired(obfuscatedEmail))
                                 }
                             }
-                            BackendErrorCode.BackendPurchaseBelongsToDifferentUser.value -> {
+                            BackendErrorCode.BackendPurchaseBelongsToOtherUser.value -> {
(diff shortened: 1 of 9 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -712,8 +712,8 @@ internal class Backend(
                                     callback(RedeemWebPurchaseListener.Result.Expired(obfuscatedEmail))
                                 }
                             }
-                            BackendErrorCode.BackendWebPurchaseAlreadyRedeemed.value -> {
-                                callback(RedeemWebPurchaseListener.Result.AlreadyRedeemed)
+                            BackendErrorCode.BackendPurchaseBelongsToDifferentUser.value -> {
```
vegaro, line 715: There's a mismatch different vs other `BackendPurchaseBelongsToDifferentUser` vs `PurchaseBelongsToOtherUser` can we consolidate?

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Ups good catch, thanks!

antoniobg (approved):

PR #1963, opened 2024-12-05
[Paywalls V2] Add `ButtonComponentView`

tonidero:
### Description
This PR adds `ButtonComponentView` to our components

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
@@ -1,0 +1,9 @@
+package com.revenuecat.purchases.ui.revenuecatui.components
+
+import com.revenuecat.purchases.paywalls.components.ButtonComponent
+
+internal sealed interface PaywallAction {
+    object RestorePurchases : PaywallAction
+    object NavigateBack : PaywallAction
+    data class NavigateTo(val destination: ButtonComponent.Destination) : PaywallAction
+}
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
@@ -1,9 +1,11 @@
 package com.revenuecat.purchases.ui.revenuecatui.components
 
 import com.revenuecat.purchases.paywalls.components.ButtonComponent
+import dev.drewhamilton.poko.Poko
 
 internal sealed interface PaywallAction {
     object RestorePurchases : PaywallAction
     object NavigateBack : PaywallAction
-    data class NavigateTo(val destination: ButtonComponent.Destination) : PaywallAction
+
+    @Poko class NavigateTo(@get:JvmSynthetic val destination: ButtonComponent.Destination) : PaywallAction
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
@@ -1,3 +1,5 @@
+@file:JvmSynthetic
+
 package com.revenuecat.purchases.ui.revenuecatui.components.button
 
 import androidx.compose.foundation.clickable
@@ -43,11 +45,11 @@ internal fun ButtonComponentView(
 @Preview
 @Composable
 private fun ButtonComponentView_Preview_Default() {
-    ButtonComponentView(previewButtonComponentViewStyle())
+    ButtonComponentView(previewButtonComponentStyle())
 }
 
 @Composable
-private fun previewButtonComponentViewStyle(
+private fun previewButtonComponentStyle(
     stackComponentStyle: StackComponentStyle = StackComponentStyle(
         visible = true,
         children = listOf(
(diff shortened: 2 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
```diff
@@ -0,0 +1,9 @@
+package com.revenuecat.purchases.ui.revenuecatui.components
+
+import com.revenuecat.purchases.paywalls.components.ButtonComponent
+
+internal sealed interface PaywallAction {
```
tonidero, line 5: I've been thinking whether to reuse the `ButtonComponent.Action` for now, since right now we will require to map from that to this type....The idea is to have all paywall customer actions here eventually, so we can pass the action back to the view model where the logic would live to update the state. Lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
```diff
@@ -0,0 +1,9 @@
...
+    object RestorePurchases : PaywallAction
+    object NavigateBack : PaywallAction
+    data class NavigateTo(val destination: ButtonComponent.Destination) : PaywallAction
```
tonidero, line 8: Note I did end up reusing `ButtonComponent.Destination` since I felt that was fine.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.components.PaywallAction
+
+@Immutable
+internal class ButtonComponentStyle private constructor(
+    @get:JvmSynthetic
+    val stackComponentStyle: StackComponentStyle,
+    @get:JvmSynthetic
+    val action: PaywallAction,
+    @get:JvmSynthetic
+    val actionHandler: (PaywallAction) -> Unit,
```
tonidero, line 14: I've been wondering as well whether it would make sense to remove the `action` parameter and have a single lambda passed in that would be called when the button is clicked... But I think this provides more flexibility in case we need to do anything in the button itself.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
JayShortway: I think we'll have to build this eventually, as this is going to contain other actions too if I'm not mistaken? Things like "purchase", and "select package". So we might as well have it now already.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -0,0 +1,91 @@
...
+}
+
+@Preview
+@Composable
+private fun ButtonComponentView_Preview_Default() {
+    ButtonComponentView(previewButtonComponentViewStyle())
+}
+
+@Composable
+private fun previewButtonComponentViewStyle(
```
JayShortway, line 50: Extremely minor: 

```diff
- private fun previewButtonComponentViewStyle(
+ private fun previewButtonComponentStyle(
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
JayShortway: I think this is fine! Don't really have an opinion yet I think haha. We can go with this and always change it if we see that the other option would've been better.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
JayShortway: Works for me! 

We could hide the parameter from Java:
```suggestion
    data class NavigateTo(@get:JvmSynthetic val destination: ButtonComponent.Destination) : PaywallAction
```

  
Also, we could think about making it a regular class and using `@Poko` to remove `copy()` and `componentN()`, although I'm not sure if that's necessary as Java users would already get a warning about using `internal` types.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
JayShortway, line 5: Should we mark it with `@InternalRevenueCatAPI` for Java users? Honestly not sure, as Java users would already get a warning about using `internal` types.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -0,0 +1,91 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.button
```
JayShortway, line 1: Just in case, although not sure if needed for Composables: 

```suggestion
@file:JvmSynthetic

package com.revenuecat.purchases.ui.revenuecatui.components.button
```

JayShortway (approved):
Another one down, very nice!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
tonidero: Hmm yeah... IMO, I think it would just be another warning they need to bypass but wouldn't actually remove the API for them... So I think I would probably leave it as is.

JayShortway: Yea makes sense!

PR #1965, opened 2024-12-06
[Paywalls V2] Adds a `StyleFactory`

JayShortway:
## Description
This PR includes all the scaffolding needed to recursively build the tree of `ComponentStyle`s, and to validate that all components are correctly configured (e.g. all localizations are available). Any errors encountered are accumulated, so they can all be reported at once. (As opposed to failing fast, in which case only the first error would be reported.)

Apologies for the size of this PR, but I think separating some of these changes would make it hard to see how they're used. So, take this 👇 map and let me guide you.

#### In support of the main change
- Some classes are further opened up, i.e. made fake-public, because they are needed by `revenuecatui`. Data classes are made `@Poko`s if they're public.
- Documentation is added to `Result` to explain the fact that it is duplicated in `revenuecatui`. 

#### In support of the main change
- As discussed, the `Result` class from `purchases` is copied. A few extensions, mostly inspired by [Arrow](https://arrow-kt.io/), are added. Most importantly:
    - `zipOrAccumulate`: takes multiple `Result`s and either combines their successful values, or accumulates the errors.
    - `mapOrAccumulate`: takes a `List<Result>` and turns it into a `Result<List>` by mapping either its successful values, or accumulating the errors. 
- A new `NonEmptyList` class is added, also inspired by [Arrow](https://arrow-kt.io/). This is useful for error accumulation. If we used a normal `List` type for this, we could still end up in the invalid state of receiving an `Error` result with an empty list of errors. Using a `NonEmptyList`, it is guaranteed we have at least 1 error in this case. 
- Tests are adjusted for the new `Result` type.
- Tests are added for the new extensions.

#### The main change
- There's a new `StyleFactory` that can create `ComponentStyle`s recursively to build a tree. It uses the new `Result` class, the extensions, and the `NonEmptyList` to accumulate any errors found. Its logic is loosely based on the `ViewModelFactory` and individual `*ComponentViewModel`s in iOS.
- Adds `PackageContext`, based on the class of the same name in iOS. 
- Some transformation methods are added, such as `toShape()`, to be able to build the `StackComponentStyle`. 
- `StyleFactoryTests` are added. 
- `VariableContextTests` are added. This is an inner class to `PackageContext` containing the most logic.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -0,0 +1,192 @@
...
+    private companion object {
+        private const val DEFAULT_SPACING = 0f
+        private val DEFAULT_SHAPE = RectangleShape
+    }
+
+    @Composable
+    fun create(component: PaywallComponent): Result<ComponentStyle, NonEmptyList<PaywallValidationError>> =
+        when (component) {
+            is ButtonComponent -> TODO("ButtonComponentStyle is not yet implemented.")
+            is ImageComponent -> TODO("ImageComponentStyle is not yet implemented.")
```
JayShortway, line 64: This is not entirely true anymore, but I'd like to verify that this approach works before we add support for all the other `ComponentStyle`s.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -15,6 +15,7 @@ import com.revenuecat.purchases.ui.revenuecatui.errors.PaywallValidationError
 import com.revenuecat.purchases.ui.revenuecatui.extensions.createDefault
 import com.revenuecat.purchases.ui.revenuecatui.extensions.createDefaultForIdentifiers
 import com.revenuecat.purchases.ui.revenuecatui.extensions.defaultTemplate
+import kotlin.Result
```
JayShortway, line 18: This was needed because the `OfferingToStateMapper` is in the same package as the new `Result` class, but it is using `kotlin.Result`.

ui/revenuecatui/build.gradle
```diff
@@ -86,6 +86,7 @@ dependencies {
     testImplementation libs.coroutines.test
     testImplementation libs.kotlinx.serialization.json
     testImplementation libs.androidx.test.compose
+    testImplementation libs.androidx.test.compose.manifest
```
JayShortway, line 89: Tests [failed on CI due to a manifest merger issue](https://app.circleci.com/pipelines/github/RevenueCat/purchases-android/12328/workflows/1ce073b2-3c74-4e05-bac4-6a1f855f6cbe/jobs/53596/tests).

This was already added as `debugImplementation`, but we run tests for the release variant (as well), e.g. `testDefaultsReleaseUnitTest`, so this makes sure we have it in any test variant, whether debug or release.

tonidero (approved):
I think this all makes sense! Just a couple questions

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Shape.kt
```diff
@@ -0,0 +1,21 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.ktx
+
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.ui.graphics.RectangleShape
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.dp
+import com.revenuecat.purchases.paywalls.components.properties.Shape as RcShape
+
+@JvmSynthetic
```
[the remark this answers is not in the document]
JayShortway: Good catch! https://github.com/RevenueCat/purchases-android/pull/1965/commits/79107d12e468e2589120ccfd4e2fffa322d8b119

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -0,0 +1,192 @@
...
+     * [package][PackageContext.selectedPackage] and [locale].
+     */
+    @Composable
+    private fun rememberProcessedText(
+        originalText: String,
+        packageContext: PackageContext,
+        variables: VariableDataProvider,
+        locale: Locale,
+    ): String {
+        val processedText by remember(packageContext) {
```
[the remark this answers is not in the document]
tonidero: Ahh right makes sense 👍

PR #1967, opened 2024-12-10
[Paywalls V2] Adds image background tests

JayShortway:
## Description

We can't easily have `@Preview`s for remote images (*) and thus don't have snapshot tests for image backgrounds. This PR adds unit tests to verify them instead. It caught a bug too (highlighted below).

--------
(*): we _might_ be able to do something with the `FakeImageLoaderEngine` in previews too, although we need to be careful not to ship this to production. Something we could look into later.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
@@ -37,9 +37,9 @@ internal fun Background.toBackgroundStyle(): BackgroundStyle =
             val imageUrls = value.urlsForCurrentTheme
             BackgroundStyle.Image(
                 painter = rememberAsyncImagePainter(
-                    model = imageUrls.webp,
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
```diff
@@ -37,9 +37,9 @@ internal fun Background.toBackgroundStyle(): BackgroundStyle =
             val imageUrls = value.urlsForCurrentTheme
             BackgroundStyle.Image(
                 painter = rememberAsyncImagePainter(
-                    model = imageUrls.webp,
+                    model = imageUrls.webp.toString(),
                     placeholder = rememberAsyncImagePainter(
-                        model = imageUrls.webpLowRes,
+                        model = imageUrls.webpLowRes.toString(),
```
JayShortway, line 42: This bug was fixed by the tests. It caused the image to not load at all, as we don't have a `Fetcher` that supports the `URL` type.

tonidero (approved):

PR #1976, opened 2024-12-12
Parse web redemption link strings more defensively

tonidero:
### Description
Followup to #1972, this makes sure we handle issues when parsing URLs if any. It should normally not happen though, but better to be defensive.


JayShortway (approved):
Thanks for this! No idea if this will ever catch anything, but better safe than sorry I guess 😄

PR #1981, opened 2024-12-12
[Paywalls V2] Fixes `TextComponentView`'s  font size

JayShortway:
## Description

For some reason, the font size is not properly picked up by the `Markdown` composable, but only when running on a device. This worked fine in previews, but I noticed it when using PaywallsTester. This PR adds 2 Previews, but they already looked fine before the fix. The fix itself is very small, highlighted below.

## Comparison
| Before | After |
|--------|------|
|  |  |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
@@ -66,6 +66,9 @@ private val parser = Parser.builder()
  * @param allowLinks If true, links will be decorated and clickable.
  * @param textFillMaxWidth If true, the text will fill the maximum width available. This was used by paywalls V1 and
  * left to avoid unintended UI changes.
+ * @param applyFontSizeToParagraph If true, the provided [fontSize] will be applied to the annotated string used to
+ * build a Markdown paragraph from the [text]. This was not the case in Paywalls V1, but is needed for Paywalls V2.
+ * (See `TextComponentViewTests` for more info.)
  */
 @SuppressWarnings("LongParameterList")
 @Composable
@@ -80,7 +83,9 @@ internal fun Markdown(
     horizontalAlignment: Alignment.Horizontal = Alignment.Start,
     textAlign: TextAlign? = null,
     allowLinks: Boolean = true,
-    textFillMaxWidth: Boolean = false, // This is to support V1 paywalls
+    // The below parameters are used to avoid unintended changes to V1 paywalls.
+    textFillMaxWidth: Boolean = false,
+    applyFontSizeToParagraph: Boolean = true,
 ) {
     val root = parser.parse(text) as Document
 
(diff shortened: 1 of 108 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -202,7 +202,7 @@ private fun MDParagraph(
         val styledText = buildAnnotatedString {
             pushStyle(
                 style
-                    .copy(fontWeight = fontWeight)
+                    .copy(fontWeight = fontWeight, fontSize = fontSize)
```
JayShortway, line 205: This is the actual fix. Again, this worked fine in previews so that's a bit annoying in terms of avoiding regression.

tonidero: Uhh that's interesting and annoying indeed... I'm guessing this will actually affect existing paywalls 😢... I'm wondering if we should somehow apply this only for paywall components to avoid unintended changes

 Great catch though!!

JayShortway: Yea.. 55 modified snapshots 🥲 I'll see if I can apply it only to V2 indeed, good one!

JayShortway: So I've done some more testing. I can't for the life of me get a regression test to work in the `revenuecatui` project. Everything "just works". However, if I create a new app and integrate Paywalls V2, I can see the issue (same as PaywallsTester). 

It seems to be an interplay between the theme and the Markdown component, because removing either of those fixes the issue. 

@tonidero @vegaro What are your thoughts on having a new module with a single regression test for this? Alternatively we could add that test to the existing `paywall-tester` module. It's a bit odd to have a single test in a separate module, but I'm also not super happy about the prospect of not having a regression test for this.

JayShortway: >Just to check, have you injected a theme like the one in one of these apps into the previewer?

In the previewer didn't work, but wrapping my test in a `MaterialTheme` totally did! Thanks for that! 💡 Pushing soon. 

(Was indeed not gonna delay V2 for the special regression test app, but this one is just a "normal" test.)

JayShortway: Alright: 
- Regression test: https://github.com/RevenueCat/purchases-android/pull/1981/commits/914d9dac057f0d5f1ce5d84c44262b8376872515
- Excluding Paywalls V1 from new behavior: https://github.com/RevenueCat/purchases-android/pull/1981/commits/9abcb6a0ce0edcb90a06cef942c4b57800058c39

tonidero (approved):
LGTM!

PR #1984, opened 2024-12-16
[PaywallsV2] Improve `ButtonComponentView` to handle disabling the button and add `PurchasePackage` action

tonidero:
### Description
This PR does 2 things:
- Makes `ButtonComponentStyle`'s `actionHandler` a suspendable function, and disables the button clickability until previous taps have finished processing.
- Adds `PurchasePackage` to the available `PaywallAction` sealed interface.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
@@ -6,6 +6,7 @@ import dev.drewhamilton.poko.Poko
 internal sealed interface PaywallAction {
     object RestorePurchases : PaywallAction
     object NavigateBack : PaywallAction
+    object PurchasePackage : PaywallAction
 
     @Poko class NavigateTo(@get:JvmSynthetic val destination: ButtonComponent.Destination) : PaywallAction
 }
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
```diff
@@ -6,6 +6,7 @@ import dev.drewhamilton.poko.Poko
 internal sealed interface PaywallAction {
     object RestorePurchases : PaywallAction
     object NavigateBack : PaywallAction
+    object PurchasePackage : PaywallAction
```
tonidero, line 9: We discussed this over Slack but we're not passing the actual package here since the view model should already know what's being purchased.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -30,15 +35,27 @@ import com.revenuecat.purchases.ui.revenuecatui.components.stack.StackComponentV
...
 internal fun ButtonComponentView(
     style: ButtonComponentStyle,
     modifier: Modifier = Modifier,
 ) {
+    val coroutineScope = rememberCoroutineScope()
+    var isClickable by remember { mutableStateOf(true) }
     StackComponentView(
         style.stackComponentStyle,
-        modifier.clickable { style.actionHandler(style.action) },
+        modifier.conditional(isClickable) {
```
tonidero, line 50: This doesn't actually have any visual feedback of it being disabled. We might want to pass this state downwards however, like if we want to have a spinner instead of the actual "StackComponentView" for example... But I thought this was better than nothing for now.

JayShortway (approved):

PR #1988, opened 2024-12-16
[Paywalls V2] Moves more state to `PaywallState`

JayShortway:
## Description
In this PR we start making more use of `PaywallState.Loaded.Components` to drive.. you guessed it.. state! 😄 Currently `TextComponentView` is the most affected by this. 

## Other changes
- Merges `PackageContext` with `PaywallState.Loaded.Components`, and deletes the former.
- Adds `LoadedPaywallComponentsLocaleTests`.
- Adds `localeChangingTest()` for easy testing of behavior under locale changes.
- Renames `VariableContextTests` to `MostExpensivePricePerMonthMicrosTests`.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -48,10 +57,56 @@ internal sealed interface PaywallState {
...
+            private fun Locale.toLocaleId(): LocaleId =
+                LocaleId(toLanguageTag().replace('-', '_'))
+
+            private fun List<Package>.mostExpensivePricePerMonthMicros(): Long? =
+                asSequence()
+                    .map { pkg -> pkg.product }
+                    .mapNotNull { product -> product.pricePerMonth() }
+                    .maxByOrNull { price -> price.amountMicros }
+                    ?.amountMicros
+        }
```
JayShortway, line 109: This is the main change in this PR ☝️

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -48,10 +57,56 @@ internal sealed interface PaywallState {
...
+            initialSelectedPackage: Package? = null,
+        ) : Loaded {
+            private var localeId by mutableStateOf(initialLocaleList.toLocaleId())
+
+            val localizationDictionary by derivedStateOf { data.componentsLocalizations.getValue(localeId) }
+            val locale by derivedStateOf { localeId.toLocale() }
+
+            var isEligibleForIntroOffer by mutableStateOf(initialIsEligibleForIntroOffer)
+                private set
+            var selectedPackage by mutableStateOf<Package?>(initialSelectedPackage)
```
tonidero, line 75: Hmm I wonder if all this data should arrive to the components already as part of the "Style" object... So the VM generates the tree of styles, and regenerates it when the package/locale/dark mode change (would probably need to let it know that it changed somehow), and components just listen to it... Wdyt?

tonidero (approved):
Changes are ok but I'm just wondering about the architecture... We're still on time to change it if needed though, so will approve it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
JayShortway: Ah yes, I was actually moving away from that, for a few reasons: 
- Creating the tree of styles is relatively expensive (depending on the complexity of the paywall), and this would mean we throw it away each time a state change happens. 
- For the same reason, ideally we create the tree as part of the "loading" phase in the ViewModel, which we can't if we have to redo it on state changes. 
- Also, Compose is really good at figuring out selective updates to a tree structure, so my intention is to make use of that. 
- Lastly, it's usually most convenient to listen to (and handle) state changes from within a Composable context (e.g. dark mode) than it is to wire that back up to the ViewModel. 

You can see a first step towards this in the next PR: https://github.com/RevenueCat/purchases-android/pull/1989.

tonidero: Ok I think that makes sense. Let's do it this way then 👍

PR #1990, opened 2024-12-18
[Paywalls V2] Adds a `FLAG_PAYWALL_COMPONENTS` build flag

JayShortway:
## Description
Since the iOS closed beta is coming up, I thought it'd be good to be extra careful about parsing the `paywall_components` JSON, as more people will soon have the ability to create those via the dashboard.

## How to
To enable, add the following line to your `local.properties` file in the root of the repo:
```properties
revenuecat.flag.paywallComponents=true
```

To disable, set it to false, or remove the line entirely.


tonidero (approved):
Good thinking!

PR #1993, opened 2024-12-18
[Paywalls V2] `StackComponentView` handles overrides with `StackComponentState`

JayShortway:
## Description
This PR is analogous to #1989, and the motivation behind it is the same. 

- Introduces `StackComponentState`.
- `StackComponentView` uses `StackComponentState`.
- `StyleFactory` is simplified further.
- Adds `StackComponentViewWindowTests`.
- Adds 2 tests to `StackComponentViewTests`.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -0,0 +1,121 @@
...
+    private var selected by mutableStateOf(initialSelected)
+    private val presentedPartial by derivedStateOf {
+        val windowCondition = ScreenCondition.from(windowSize)
+        val componentState = if (selected) ComponentViewState.SELECTED else ComponentViewState.DEFAULT
+
+        style.overrides?.buildPresentedPartial(windowCondition, isEligibleForIntroOffer, componentState)
+    }
+
+    @get:JvmSynthetic
+    val visible by derivedStateOf { presentedPartial?.partial?.visible ?: true }
```
tonidero, line 79: Should this default to the style visibility? I don't know if this is something controllable from the dashboard but seems like a better default IMO?
```suggestion
    val visible by derivedStateOf { presentedPartial?.partial?.visible ?: style.visible }
```

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
JayShortway: Oh good that you mention it! This made me realize that `*ComponentStyle`s don't need a `visible` property anymore. I removed them from `StackComponentStyle` and `TextComponentStyle` here: https://github.com/RevenueCat/purchases-android/pull/1993/commits/bf27e456c3073c76a8cc18b1f9632b18588f4e19. Will need to do the same once `ImageComponentStyle` gets an accompanying `ImageComponentState`. 

The reason for this is that a component is always visible, except when a `PresentedPartial` override tells it it's not. This is to support scenarios where a developer wants to hide a component for a certain window size, for instance. That's why only the `PresentedPartial` overrides have a `visible` property, and when that's not set the component is visible.

tonidero: 👍 makes sense! Yeah I didn't think it made much sense to control the visibility from the dashboard like that. This makes more sense!

tonidero (approved):

PR #1997, opened 2024-12-19
Fix missing import after PR merge conflict

tonidero:
### Description
Fix issue after merging recent Paywalls PR that was making it fail to compile. Tests passed on the respective branches but we didn't have tests against latest main before merging the latest.


JayShortway (approved):

tonidero:
Emerge size tests failed because build failed on main, so it can't compare it. Merging anyway.

PR #1998, opened 2024-12-19
Build Customer Center from JSON

vegaro:
First PR building the Customer Center off the JSON response. It's not complete yet but I want to open multiple PRs

| Active Product | No Active Product |
|--------|--------|
| Screenshot_20241219_150608 |  Screenshot_20241219_150927 |


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
vegaro, line 1: Updated this file with new stuff we added

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterConfigTestData.kt
vegaro, line 1: Copied this from iOS, there is some unused stuff

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterViewModel.kt
```diff
@@ -35,4 +37,16 @@ internal class CustomerCenterViewModelImpl(
...
         initialValue = CustomerCenterState.Loading,
     )
+
+    private suspend fun loadPurchaseInformation(): PurchaseInformation? {
+        val customerInfo = purchases.awaitCustomerInfo(fetchPolicy = CacheFetchPolicy.FETCH_CURRENT)
+
+        // Customer Center WIP: udpate when we have subscription information in CustomerInfo
+        val activeEntitlement = customerInfo.entitlements.active.isEmpty()
+        if (activeEntitlement) {
+            return CustomerCenterConfigTestData.purchaseInformationMonthlyRenewing
```
vegaro, line 47: hardcoding this for now, will add the logic in a different PR. I need to add the new field to the `CustomerInfo`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/ManageSubscriptionsViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+            screen = screen,
+            purchaseInformation = purchaseInformation,
+        ),
+    )
+
+    val uiState: StateFlow<UiState> = _uiState.asStateFlow()
+
+    suspend fun determineFlow(path: CustomerCenterConfigData.HelpPath) {
+        // Implement the logic to determine the flow for the given path
+        if (path.type == CustomerCenterConfigData.HelpPath.PathType.MISSING_PURCHASE) {
```
vegaro, line 35: This is not the real implementation, just adding restore for now. As the others, this logic will come in a separate PR

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -75,12 +75,31 @@ private fun CustomerCenterError(state: CustomerCenterState.Error) {
...
 @Composable
 private fun CustomerCenterLoaded(state: CustomerCenterState.Success) {
-    // CustomerCenter WIP: Add proper success UI
-    Column {
-        Text("Customer Center config:", fontWeight = FontWeight.Bold, fontSize = 20.sp)
-        Text(state.customerCenterConfigDataString)
+    val configuration = state.customerCenterConfigData
+
+    if (state.purchaseInformation != null) {
+        if (configuration.screens.containsKey(CustomerCenterConfigData.Screen.ScreenType.MANAGEMENT)) {
```
tonidero, line 84: I'm wondering if it should be a parsing error for the json not to contain this and/or the NO_ACTIVE screens instead of having to force unwrap here...

In that case, we should maybe make accesors in the `CustomerCenterConfigData` or convert `screens` to a different type that has ready accessors... 

All of this can be done in different PRs if we want to however, so not a blocker for this PR.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/ManageSubscriptionsViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.PurchaseInformation
+import com.revenuecat.purchases.ui.revenuecatui.data.PurchasesType
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal class ManageSubscriptionsViewModel(
```
tonidero, line 13: Hmm I'm wondering if we should have the VM at the screen level, or at the `CustomerCenter` level... That is, whether this should be `CustomerCenterViewModel`. We normally only display one screen at the same time, so it kinda makes sense to me to have the VM at that level. 

I guess having different VMs for different screens could allow us to separate logic more effectively, but IMO, most of the logic can be shared for all screens, wdyt?

tonidero (commented):
Looking great! My only concern is whether we should have a single VM in charge of the entire CustomerCenter

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/ManageSubscriptionsViewModel.kt
vegaro: I like it! It will simplify the code tbh. Refactoring!!!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: I think an error makes sense tbh. If we don't get those screens it's clearly an error

tonidero: Right, I'm just wondering if we should fail early, at the parsing the response phase... But we can keep this for now.

tonidero (approved):

PR #2000, opened 2024-12-20
[Paywalls V2] `TextComponentState` handles locale changes

JayShortway:
## Description
This PR does only 1 thing, but it cascades into a lot of small changes in a lot of files. The main change is that `TextComponentStyle` now contains the string it needs for all locales, and `TextComponentState` will make sure the right one is emitted, for the current locale.

## Other changes
- Adds 2 tests to `StyleFactoryTests` to check the validation logic.
- Adds 2 tests to `TextComponentViewTests` to check the right string is shown.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentState.kt
```diff
@@ -81,7 +87,12 @@ internal class TextComponentState(
     val visible by derivedStateOf { presentedPartial?.partial?.visible ?: true }
 
     @get:JvmSynthetic
-    val text by derivedStateOf { presentedPartial?.text ?: style.text }
+    val text by derivedStateOf {
+        val localeId = localeProvider().toLocaleId()
+        // TODO Fallback to default locale
+        presentedPartial?.texts?.getValue(localeId)
+            ?: style.texts.getValue(localeId)
```
JayShortway, line 94: This will be in the next PR.

tonidero (approved):

PR #2001, opened 2024-12-20
[Paywalls V2] Localizations are kept in a NonEmptyMap

JayShortway:
## Description
This PR also introduces only 1 thing, which cascades into a lot of small changes in a lot of files. The main change here is that it introduces the concept of a `NonEmptyMap` to keep the localizations in. This means that the components do not need to worry about missing localizations. When the paywall is validated, we know that we have at least 1 localization. This avoids a lot of error scenarios in the components themselves. 

## Other changes
- Adds a test for `NonEmptyMap`.
- Adds some tests for the `mapOrAccumulate` variant with a `NonEmptyMap` receiver.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Localization.kt
```diff
@@ -26,9 +26,6 @@ value class LocaleId(@get:JvmSynthetic val value: String)
 @JvmInline
 value class LocalizationKey(@get:JvmSynthetic val value: String)
 
-@InternalRevenueCatAPI
-typealias LocalizationDictionary = Map<LocalizationKey, LocalizationData>
```
JayShortway, line 30: This is moved to `revenuecatui` so it can be a `NonEmptyMap`.

tonidero (approved):

PR #2002, opened 2024-12-20
[Paywalls V2] Add `PurchaseButtonComponent` support

tonidero:
### Description
Adds support for the `PurchaseButtonComponent`, which maps to a normal `ButtonComponentView` for now. We also pass action and actionhandler from the style factory, which might also change once we move more logic to the VM.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
@@ -1,33 +1,14 @@
 package com.revenuecat.purchases.ui.revenuecatui.components.style
 
-import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
 import com.revenuecat.purchases.ui.revenuecatui.components.PaywallAction
 
 @Immutable
-internal class ButtonComponentStyle private constructor(
+internal class ButtonComponentStyle(
     @get:JvmSynthetic
     val stackComponentStyle: StackComponentStyle,
     @get:JvmSynthetic
     val action: PaywallAction,
     @get:JvmSynthetic
     val actionHandler: suspend (PaywallAction) -> Unit,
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
```diff
@@ -1,33 +1,14 @@
...
 
 @Immutable
-internal class ButtonComponentStyle private constructor(
+internal class ButtonComponentStyle(
     @get:JvmSynthetic
     val stackComponentStyle: StackComponentStyle,
     @get:JvmSynthetic
     val action: PaywallAction,
     @get:JvmSynthetic
     val actionHandler: suspend (PaywallAction) -> Unit,
```
JayShortway, line 13: I think eventually we'll need to move this to `ButtonComponentView` as a parameter, to decouple UI logic from `StyleFactory`. Can be done later as well of course.

JayShortway (approved):
Very nice!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
tonidero: Yeah, once we build the logic to inject this lambda in the view, it can be passed through there directly, without involving the style. For now, I think I will leave it here and can be moved in a separate PR as you said 👍

JayShortway: Sounds good!

JayShortway: For posterity, this is done here: https://github.com/RevenueCat/purchases-android/pull/2016

PR #2004, opened 2024-12-20
[Paywalls V2] Add more image component previews to test parent being smaller than image size

tonidero:
### Description
This has 2 main changes:
- Adds one more preview for `ImageComponentView` that verifies how it looks when the dimension size is `Fit` (which translates to the image size) but the parent is smaller than the image.
- Adds one more parameter to `RemoteImage` and `LocalImage` that allows to set an optional preview resource drawable.


JayShortway (approved):
Nice to get some extra test coverage!

PR #2005, opened 2024-12-21
Support repurchasing of consumables in paywalls

joshdholtz:

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -154,6 +154,8 @@ internal class PaywallViewModelImpl(
             Logger.e("Activity is null, not initiating package purchase")
             return
         }
+
+        // JOSH: One section
         if (verifyNoActionInProgressOrStartAction()) {
             return
         }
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -389,6 +389,8 @@ internal class PaywallViewModelImpl(
                 variableDataProvider = variableDataProvider,
                 activelySubscribedProductIdentifiers = customerInfo.activeSubscriptions,
                 nonSubscriptionProductIdentifiers = customerInfo.nonSubscriptionTransactions
+                    // Filter out consumables so they don't get checked in valiation
+                    .filter { !it.shouldConsume }
```
tonidero, line 393: Hmm I'm honestly wondering if we should just remove the validation for repurchasing any product and just leave it to google to display a "Already purchased product" if the customer is already subscribed/has a non-consumed product...

tonidero (commented):
Looking good, but I do wonder if we should just rely on Google more for this...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
aboedo: +1

joshdholtz: Did we have this conversation already on my prototype of this from months ago? Because this feels awkwardly familiar? 🤔

aboedo:
we're good to close this one right?

tonidero:
> we're good to close this one right?

Possibly... We are now not checking custom packages: https://github.com/RevenueCat/purchases-android/pull/2044. However, we still won't allow subscribing if the package type is a known one... I think we should still consider removing all the checks for all package types entirely, and just rely on Google for this... But in any case, I think this PR can be closed yeah.

PR #2007, opened 2024-12-23
[Paywalls V2] Moves validation logic to the Loading phase

JayShortway:
## Description
This PR moves `StyleFactory.create()` to the Loading phase, and adds some missing validation pieces. Don't be scared by the large diff. 😄 A great deal of that is previews and other non-consequential changes that rippled out from the main changes. 

## Main changes
- All validation logic is now in `OfferingToStateMapper.kt`. When validation fails, we return all the errors plus the existing legacy fallback paywall. The legacy paywall validation logic should behave the same.
- The structure of `PaywallValidationResult` changed somewhat. Its errors are now in a `NonEmptySet`. A Legacy paywall still only has 1 error at most. 
- `PaywallState.Loaded.Components` no longer has the network response (`PaywallComponentsData`). Instead, it only has what it needs, which is now validated. 

## Other changes
- Adds a `NonEmptySet` collection type.
- Adds `PaywallComponentDataValidationTests`.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -46,11 +55,72 @@ internal fun Offering.validatedLegacyPaywall(
...
+                config.stickyFooter?.let { styleFactory.create(it, actionHandler) }.orSuccessfullyNull(),
+            ) { stack, stickyFooter ->
+                PaywallValidationResult.Components(
+                    stack = stack,
+                    stickyFooter = stickyFooter,
+                    background = config.background,
+                    locales = localizations.keys,
+                )
+            }
+        }
```
vegaro, line 122: This chain is not complex, but it's a bit hard to read, it takes some time to fully understand what's going on. I think extracting into some private functions could help

vegaro (approved):
image

I was scared, but as you said, it's not that complex lol

I am just a bit worried about that chain of functions. It took me a lot of focus to fully grasp what's going on

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
JayShortway: Yea I agree. And we'll have to spend that time every time we need to read this code. Private functions is a good idea. I was also thinking of adding some early returns. I'll have a look.

JayShortway: Decided to do this separately to minimize the conflicts, as I ended up making some changes to these lines in follow-up PRs. Have a look here: https://github.com/RevenueCat/purchases-android/pull/2017

PR #2008, opened 2024-12-23
Add Cancel subscriptions support 

vegaro:
Make Cancel Subscription open the play store


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
```diff
@@ -148,6 +148,7 @@ private class SubscriptionInformationProvider : PreviewParameterProvider<Purchas
             expirationDateString = "June 1st, 2024",
             willRenew = true,
             active = true,
+            productId = "basic_monthly",
```
JayShortway, line 151: I just noticed that this is a `PreviewParameterProvider` and was gonna comment that Emerge doesn't support those. However it seems [they do support it since version 1.3.0](https://docs.emergetools.com/docs/android-snapshots-v1#previewparameter-annotated-parameters) of their Snapshots SDK. We're using 1.2.0. Could you bump? 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -122,9 +134,21 @@ internal class CustomerCenterViewModelImpl(
         val customerInfo = purchases.awaitCustomerInfo(fetchPolicy = CacheFetchPolicy.FETCH_CURRENT)
 
         // Customer Center WIP: update when we have subscription information in CustomerInfo
-        val activeEntitlement = customerInfo.entitlements.active.isEmpty()
+        val activeEntitlement = customerInfo.entitlements.active.isNotEmpty()
         if (activeEntitlement) {
-            return CustomerCenterConfigTestData.purchaseInformationMonthlyRenewing
+            val entitlement = customerInfo.entitlements.active.values.first()
```
JayShortway, line 139: We're reading `entitlements.active` twice here, but we can't be sure that it's still not empty by the time we call `first()`. (Maybe we can be sure with the current implementation of `CustomerInfo`, but we never know how it will change in the future.)

We can also do something like this: 

```kotlin
val entitlement = customerInfo.entitlements.active.values.firstOrNull()
if (entitlement != null) {
    // ...
}
```

JayShortway (commented):
Good stuff! 2 minor comments.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
vegaro: done in [`d20b581` (#2008)](https://github.com/RevenueCat/purchases-android/pull/2008/commits/d20b5811de9f56b36cbc2865ae157585d17515a1)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: Sure. This is temporary btw, the logic requires we add `subscriptionsByProductIdentifier` to `CustomerInfo` so I am just doing the dumb logic here. But will change that to prevent weird issues

JayShortway: Ah nice okay, thanks for the context!

JayShortway (approved):
:shipit:

PR #2009, opened 2024-12-23
[Paywalls V2] Add `Badge`'s `overlay` style layout

tonidero:
### Description
This adds the layout for `Badge` to display correctly with the `overlay` layout.

This PR moves the existing Stack layout inside another composable: `MainStackComponent`. And at the top-level, we now decide what layout to use depending on the badge style. In this PR, we are implementing the first one: `overlay`.
image


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -68,58 +79,119 @@ internal fun StackComponentView(
...
+    }
+    Box(modifier = modifier) {
+        // TODO Fix margins when using badges
+        MainStackComponent(style, state, modifier = Modifier.padding(padding))
+        StackComponentView(
+            badgeStack,
+            state,
+            modifier = Modifier
+                .align(alignment.toAlignment())
+                .onGloballyPositioned {
```
tonidero, line 131: I don't really like using this, since emerge/snapshot tests usually don't pick these changes correctly... Not sure if there is a better way to do this, while allowing arbitrary contents inside the badge though...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -68,58 +79,119 @@ internal fun StackComponentView(
...
+        -> PaddingValues(bottom = paddingDp)
+        else -> PaddingValues(0.dp)
+    }
+    Box(modifier = modifier) {
+        // TODO Fix margins when using badges
+        MainStackComponent(style, state, modifier = Modifier.padding(padding))
+        StackComponentView(
+            badgeStack,
+            state,
+            modifier = Modifier
```
tonidero, line 129: I'm wondering if we should automatically add some start/end paddings if the "root" stack has rounded corners (as it is in the designs), or just rely on given margins for the "badge's stack". This would mean we need to add those margins from the backend, on the other hand, it's more flexible, since less logic in the SDK.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -80,58 +91,119 @@ internal fun StackComponentView(
     )
 
     if (stackState.visible) {
-        val backgroundColorStyle = stackState.backgroundColor?.let { rememberColorStyle(scheme = it) }
-        val borderStyle = stackState.border?.let { rememberBorderStyle(border = it) }
-        val shadowStyle = stackState.shadow?.let { rememberShadowStyle(shadow = it) }
+        if (style.badge != null) {
+            when (style.badge.style) {
```
JayShortway, line 95: Just a heads-up since this changed recently: `style.badge.style` should probably be part of `stackState` so it can participate in the overrides mechanism.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -80,58 +91,119 @@ internal fun StackComponentView(
...
-                verticalAlignment = dimension.alignment.toAlignment(),
-                horizontalArrangement = dimension.distribution.toHorizontalArrangement(
-                    spacing = stackState.spacing,
-                ),
-            ) { content() }
+@Composable
+private fun StackWithOverlaidBadge(
+    stackState: StackComponentState,
+    state: PaywallState.Loaded.Components,
+    badgeStack: StackComponentStyle,
```
JayShortway, line 116: Another heads up: I _think_ this composable should turn the `badgeStack` into a `StackComponentState` using `rememberUpdatedStackComponentState()`, just like we do with the main `StackComponentStyle`. Can probably happen internally? Not sure, haven't thought long about it haha.

tonidero: Hmm I might be wrong but I think that would happen inside the `StackComponentView` for the badge right? [Here](https://github.com/RevenueCat/purchases-android/pull/2009/files#diff-e7e668cc1d6e5bf8f8f340aca574089508c20722df12c523871268042b0eee7bR159). Do lmk if I'm misunderstanding something though!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -108,6 +108,9 @@ internal class StackComponentState(
     @get:JvmSynthetic
     val shadow by derivedStateOf { presentedPartial?.partial?.shadow ?: style.shadow }
 
+    @get:JvmSynthetic
+    val badge by derivedStateOf { style.badge }
```
tonidero, line 112: So I went a lot of back and forth on this... I think it's fine even if this doesn't use the `presentedPartial`? The overrides on the inner stack would already be applied in the style itself I believe. The only thing that wouldn't be overriden is the style and alignment, but not sure if it's worth the refactor that would be required to support that... Lmk if I misunderstood something though!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Generally influencing the layout of a composable can be done in a `.layout { }` modifier. That allows us to keep layout logic in the layout phase, and avoids an extra composition. Instead of `.onGloballyPositioned { }`, we can do something like this: 

```kotlin
.layout { measurable, constraints ->
    val placeable = measurable.measure(constraints)
    layout(placeable.width, placeable.height) {
        placeable.placeRelative(
            x = getOverlaidBadgeOffsetX(placeable.width, alignment),
            y = getOverlaidBadgeOffsetY(placeable.height, alignment),
        )
    }
}
```

with:

```kotlin
private fun getOverlaidBadgeOffsetX(width: Int, alignment: TwoDimensionalAlignment) =
    when (alignment) {
        TwoDimensionalAlignment.CENTER,
        TwoDimensionalAlignment.TOP,
        TwoDimensionalAlignment.BOTTOM -> 0
        TwoDimensionalAlignment.LEADING,
        TwoDimensionalAlignment.TOP_LEADING,
        TwoDimensionalAlignment.BOTTOM_LEADING -> (-(width.toFloat() / 2)).roundToInt()
        TwoDimensionalAlignment.TRAILING,
        TwoDimensionalAlignment.TOP_TRAILING,
        TwoDimensionalAlignment.BOTTOM_TRAILING -> (width.toFloat() / 2).roundToInt()
    }

private fun getOverlaidBadgeOffsetY(height: Int, alignment: TwoDimensionalAlignment) =
    when (alignment) {
        TwoDimensionalAlignment.CENTER,
        TwoDimensionalAlignment.LEADING,
        TwoDimensionalAlignment.TRAILING -> 0
        TwoDimensionalAlignment.TOP,
        TwoDimensionalAlignment.TOP_LEADING,
        TwoDimensionalAlignment.TOP_TRAILING -> (-(height.toFloat() / 2)).roundToInt()
        TwoDimensionalAlignment.BOTTOM,
        TwoDimensionalAlignment.BOTTOM_LEADING,
        TwoDimensionalAlignment.BOTTOM_TRAILING -> (height.toFloat() / 2).roundToInt()
    }
```

(We should then also remove the padding from the `MainStackComponent`.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -90,69 +101,146 @@ internal fun StackComponentView(
...
+        -> PaddingValues(bottom = paddingDp)
+        else -> PaddingValues(0.dp)
+    }
+    Box(modifier = modifier) {
+        // TODO Fix margins when using badges
+        MainStackComponent(stackState, state, clickHandler, modifier = Modifier.padding(padding), selected = selected)
+        StackComponentView(
+            badgeStack,
+            state,
+            clickHandler,
```
JayShortway, line 163: Maybe we can create a `Badge` composable here that takes a `TwoDimensionalAlignment` and a  `Badge.Style`, and is implemented as a `StackComponentView` with the right (layout) modifier? That might simplify this file a bit. 

Alternatively, as a first step we could create an `OverlaidBadge` composable that does the same thing, but only implements the `Overlay` style, and just takes a `TwoDimensionalAlignment`.

Just a suggestion. Let me know what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: I think we should do it on the backend, for the flexibility reason you mention + otherwise it becomes very hard to reason about SDK behavior. Also, that would be in line with our "a badge is actually a stack" approach. The backend should provide all properties to be able to render that badge stack correctly imo, just as it does for normal stacks.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Yes I think you're right! Nevermind me 😄

JayShortway (commented):
I like it a lot! Just the 2 comments on the overrides and the layout modifier.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Oh nice one! With this code, it can get a bit out of the parent container like:
image

But I can probably fix that, and it's much better, thanks!

tonidero: Actually it's just making `getOverlaidBadgeOffsetX` be 0 always :P

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
tonidero: I ended up changing it to override the style and alignment in a single field by creating a new instance of the style if needed: https://github.com/RevenueCat/purchases-android/pull/2009/commits/6d0d6eeccaed3bd18ea1459d90b9175ed797c95d. This makes it simpler since we only have a single nullable field instead of 3. Lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Did this in https://github.com/RevenueCat/purchases-android/pull/2009/commits/321571277b394d0efb4d9d0dce12800a00c7c32a

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2009/commits/1129ef4167b3112a75befd83d7904ae514a8d653

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -90,71 +99,165 @@ internal fun StackComponentView(
...
+    badgeStack: StackComponentStyle,
+    alignment: TwoDimensionalAlignment,
+    clickHandler: suspend (PaywallAction) -> Unit,
+    modifier: Modifier = Modifier,
+    selected: Boolean = false,
+) {
+    Box(modifier = modifier) {
+        MainStackComponent(stackState, state, clickHandler, selected = selected)
+        OverlaidBadge(badgeStack, state, alignment, selected = selected)
+    }
```
JayShortway, line 143: Beautiful!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Oh of course! For some reason I thought that was the requirement haha 😅

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
JayShortway: I ... think this is perfect? Nice one!

JayShortway (approved):
Looks amazing! And thanks for making the changes!

PR #2011, opened 2024-12-24
Promotional offers

vegaro:
Please note the CTA button is still just showing "Accept offer", it should have information about the price but that will come in a different PR since the logic is a bit complex.

Screenshot 2025-01-29 at 14 07 17


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -307,6 +358,67 @@ internal class CustomerCenterViewModelImpl(
         }
     }
 
+    override fun clearActionError() {
+        _actionError.value = null
+    }
+
+    override fun loadAndDisplayPromotionalOffer(
```
vegaro, line 365: One thing I haven't found a way of accomplishing is checking if the user is currently subscribed to that particular offer. I don't think we have a way :/

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/ErrorDialog.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import com.revenuecat.purchases.ui.revenuecatui.R
+
+@Composable
+internal fun ErrorDialog(
```
JayShortway, line 14: Nice to extract this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
```diff
@@ -38,3 +40,10 @@ internal data class FeedbackSurveyData(
     val feedbackSurvey: CustomerCenterConfigData.HelpPath.PathDetail.FeedbackSurvey,
     val onOptionSelected: (CustomerCenterConfigData.HelpPath.PathDetail.FeedbackSurvey.Option?) -> Unit,
 )
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal data class PromotionalOfferData(
+    val configuredPromotionalOffer: CustomerCenterConfigData.HelpPath.PathDetail.PromotionalOffer,
+    val subscriptionOption: SubscriptionOption,
+    val originalPath: CustomerCenterConfigData.HelpPath,
```
JayShortway, line 48: In Paywalls we've been marking `val`s of `internal` classes with `@get:JvmSynthetic` to hide the getters from Java.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -88,17 +111,42 @@ internal class CustomerCenterViewModelImpl(
...
+    ) {
+        if (product == null || promotionalOffer == null) {
+            mainPathAction(path, context)
+            return
+        }
+        loadAndDisplayPromotionalOffer(
+            product,
+            promotionalOffer,
+            path,
+        )
```
JayShortway, line 149: Nit, but maybe this should just be inlined where it's called? It's now doing 2 things (has `Or` in the name), so maybe it doesn't need to be a single function?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
JayShortway: Just checking but the `CustomerInfo.activeSubscriptions` id doesn't contain this info right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferView.kt
```diff
@@ -0,0 +1,199 @@
...
+                .padding(bottom = 16.dp),
+        ) {
+            Text("No Thanks")
+        }
+    }
+}
+
+@SuppressLint("DiscouragedApi")
+@Composable
+fun AppIconView(
```
JayShortway, line 101: We have `AppIcon` in `IconImage.kt`. Maybe that can be reused?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/ContextExtensions.kt
```diff
@@ -0,0 +1,14 @@
+package com.revenuecat.purchases.ui.revenuecatui.helpers
+
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
+
+fun Context.findActivity(): Activity {
```
JayShortway, line 7: We also have this in `InternalPaywall.kt`. Probably good to consolidate?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/ContextExtensions.kt
```diff
@@ -0,0 +1,14 @@
...
+    var context = this
+    while (context is ContextWrapper) {
+        if (context is Activity) return context
+        context = context.baseContext
+    }
+    error("RevenueCatDebugView: Could not find activity context from current context.")
```
JayShortway, line 13: Is `RevenueCatDebugView` on purpose? 😄

JayShortway (approved):
Looking good! Just a few suggestions on reusing existing things.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/ErrorDialog.kt
tonidero, line 14: FWIW, we've started marking all these internal functions with `@JvmSynthetic` for Paywalls V2 and the file itself with `@file:JvmSynthetic`, so they are not available in Java. In case we want to do the same for this and everything else moving forward?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -199,6 +218,18 @@ private fun CustomerCenterLoaded(
 ) {
     if (state.feedbackSurveyData != null) {
         FeedbackSurveyView(state.feedbackSurveyData)
+    } else if (state.promotionalOfferData != null) {
+        state.promotionalOfferData.let { promotionalOfferData ->
```
tonidero, line 222: Nitpick, but I kinda prefer to define a variable if we want to have more convenient access to the variable than using let (except if it's nullable) So doing something like: `val promotionalOfferData = state.promotionalOfferData`. It avoids having another indentation

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/actions/CustomerCenterAction.kt
```diff
@@ -3,13 +3,20 @@ package com.revenuecat.purchases.ui.revenuecatui.customercenter.actions
...
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
+import com.revenuecat.purchases.models.StoreProduct
+import com.revenuecat.purchases.models.SubscriptionOption
 
 @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
 internal sealed class CustomerCenterAction {
-    data class PathButtonPressed(val path: CustomerCenterConfigData.HelpPath) : CustomerCenterAction()
+    data class PathButtonPressed(
+        val path: CustomerCenterConfigData.HelpPath,
+        val product: StoreProduct?,
```
tonidero, line 13: Hmm I'm wondering if we should separate this as a separate action for those paths that require a product...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -66,6 +87,8 @@ internal sealed class TransactionDetails(
     ) : TransactionDetails(productIdentifier, store)
 }
 
+private const val RC_CUSTOMER_CENTER_TAG = "rc-customer-center"
```
tonidero, line 90: Hmm nitpick but I'm wondering if we should define this somewhere in the `purchases` module as public, and have it accessible from the UI module, so we only define it once and we don't forget to update it if needed...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
tonidero: Hmm yeah I think we can only know the base plan they are subscribed to, but not the particular offer I believe... I'm not sure how we can do this... We might need Catforms's help and try to store some of this info in the backend... 

Another thing to consider is pending purchases and/or deferred updates, since those might be delayed... We probably need a way to expose all this info from the backend...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferView.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        Button(
+            onClick = {
+                onAccept(promotionalOfferData.subscriptionOption)
+            },
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(bottom = 16.dp),
+        ) {
+            Text("Accept offer")
```
tonidero, line 86: We might leave some `// TODO` comments in here if we're planning to work in this in follow-up PRs? Specially since they aren't localized right now.

tonidero (commented):
Looking good! Left some comments but lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
JayShortway: Public public or `@InternalRevenueCatAPI` public? (Not advocating for one or the other, genuine question.)

tonidero: Well, this should mostly only matter to us, so I think we can make it `@InternalRevenueCatAPI` and leave it like that forever yeah.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
vegaro: right! I will add those

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/ContextExtensions.kt
vegaro: Would you use `CompositionLocalProvider` as in `InternalPaywall`?

```
CompositionLocalProvider(
        LocalActivity provides activity,
    ) {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: I agree, I think I had this as nullable before and then changed it and left this as let

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: Makes sense! I wasn't comfortable with copying this either. Where would you have it public though?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/actions/CustomerCenterAction.kt
vegaro: Hmm that's a good question, it might make things a bit more clear. Let me think, if it requires a lot of refactoring I might do it in another PR

vegaro: ok I took a look, and I think leaving as we have is a bit better. The main reason is that we would have to somehow distinguish the paths here:

```
ManageSubscriptionsButtonsView(screen = screen, onButtonPress = {
                    onAction(CustomerCenterAction.PathButtonPressed(it, purchaseInformation.product))
                })
```

because `purchaseInformation.product` is nullable. Also, any path can in theory trigger a promo offer, which is why we have this product here.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/ContextExtensions.kt
JayShortway: Yea maybe that's nice for discoverability if it's needed in children of `InternalCustomerCenter`!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
tonidero: Hmm I'm thinking inside `SubscriptionOption` or some `Constants` file... Not a strong opinion though.

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: I discovered a Constants file so I added it there as a public subclass

PR #2018, opened 2024-12-30
[Paywalls V2] Fixes various deserialization issues

JayShortway:
## Description
After hooking up the Android SDK to the official V2 templates repo, some deserialization issues were uncovered: 
- `semi_bold` should be `semibold`
- When there's no padding / margin, the backend sends an empty object (`{}`). To deserialize this properly, we have to make all properties of `Padding` optional.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Padding.kt
```diff
@@ -14,19 +14,19 @@ class Padding(
     /**
      * The top padding, in dp.
      */
-    @get:JvmSynthetic val top: Double,
+    @get:JvmSynthetic val top: Double = 0.0,
```
tonidero, line 17: Hmm I wonder if this should be nullable instead of defaulting to 0... Having said that, I think it should be the same, only advantage of it being nullable is that we can skip applying the modifier in the composable entirely if it's null.

tonidero (approved):
Just a comment, but not a blocker at all.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Padding.kt
JayShortway: (Will merge this but feel free to comment ofc. We can make changes in follow-up PRs.)

PR #2021, opened 2024-12-31
[Paywalls V2] Image backgrounds support `fit_mode` and `color_overlay`

JayShortway:
## Description
This PR adds support for `fitMode` and `colorOverlay` to the `.background()` modifier. This is most noticeable in the root background, as can be seen below.

| Before | After |
|-------|------|
|image|image|


tonidero (approved):

tonidero (commented):
Might be good to add some previews/snapshots tests with these modifiers as well, to make sure there are no regressions.

JayShortway:
@tonidero Yea it's a bit hard to preview this one, because the images are loaded over the network and we're not using `RemoteImage` here since this is a modifier. That's why I added tests instead of previews. 

I also have some ideas around using a custom Coil `ImageLoader` specifically for previews, similar to the `FakeImageLoader` used in tests. Something we can look into.

PR #2024, opened 2025-01-02
[Paywalls V2] Fixes `ImageComponentView` size when axes are Fit and Fixed

JayShortway:
## Description
The `ImageComponentView` was incorrect when 1 axis was set to `Fixed` and the other to `Fit`. The `Fit` axes was adjusted to `Fixed` using the original size of the image. This PR changes that to take the `Fixed` size of the other axis into account, and scale accordingly. 

## Example
image = 200x200
width = `Fixed(100)`
height = `Fit`

|Before|After|
|------|-----|
| adjusted height = `Fixed(200)` | adjusted height = `Fixed(100)` |
| before | after |


tonidero (approved):
Ohhh makes sense, great catch! Thanks for fixing!

PR #2028, opened 2025-01-06
Migrates to Coil 3

JayShortway:
## Description
This PR updates Coil to 3.0.4. The maven coordinates have changed wrt Coil 2, which means there should be no binary incompatibility issues. For reference, the official migration docs are [here](https://coil-kt.github.io/coil/upgrading_to_coil3/).
- https://github.com/RevenueCat/purchases-android/commit/c34a6d9f5eb2ae27a57bdb5989f1d72ed3b3955c Migrates the code base to Coil 3.0.4. This is mostly a version bump + changing imports.
- https://github.com/RevenueCat/purchases-android/commit/191d7b956b8610e135d6f5a7db9540f7e96f749d Migrates all previews to make use of the new `PreviewHandler` mechanism. Check out `ImagePreviews.kt`. 

## Motivation
The main motivation is Coil's new `PreviewHandler` mechanism. This gives us great control over the images used in Compose previews. This comes in handy when creating previews for the Paywalls V2 image component, as it should handle various sizes and fit modes correctly. This automatically protects us against regressions in combination with Emerge Snapshots.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PreviewImages.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PreviewImages.kt
@@ -1,0 +1,53 @@
+package com.revenuecat.purchases.ui.revenuecatui.helpers
+
+import android.graphics.drawable.ColorDrawable
+import androidx.annotation.DrawableRes
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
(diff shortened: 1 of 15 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PreviewImages.kt
```diff
@@ -0,0 +1,53 @@
...
+import coil3.compose.LocalAsyncImagePreviewHandler
+import com.revenuecat.purchases.InternalRevenueCatAPI
+
+/**
+ * **For Compose Previews only.**
+ *
+ * Replaces all images in the [content] with the provided [color].
+ */
+@InternalRevenueCatAPI
+@ExperimentalCoilApi
```
JayShortway, line 23: I chose to propagate `@ExperimentalCoilApi` so we think twice if we accidentally try to use this in production code.

PR #2029, opened 2025-01-06
[Paywalls V2] Improves image previews

JayShortway:
## Motivation
The current mechanisms to show remote images in Compose Previews did not allow us to specify the size of the remote image. The replacement image was a solid color or a vector drawable, both of which resize uniformly to the view they're displayed in. However, in the real world, Paywalls V2 will receive images with a fixed size. We needed a way to preview resizing and scaling behavior with these kinds of images.

## Description
This PR adds a new `previewImageLoader` parameter, which can be used to return whatever kind of image we want. This is then used in `ImageComponentView` previews to check its behavior for images with a fixed size. It's slightly hacky, but the upside is that it will only run in the preview environment. The alternative approach involves updating to Coil 3 (https://github.com/RevenueCat/purchases-android/pull/2028), but that bumps Kotlin to 2.0.21. So that's left for a later date. 

This PR also fixes a rendering bug when the width is `Fill` and the height is `Fit`, by introducing an `aspectRatio` modifier. These size constraints are often used for header images. The changes to the preview mechanism allowed us to write a regression "test" (preview) for this.

## Note
This changes some existing previews for `ImageComponentView`, since the preview image is now no longer scaled uniformly.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
@@ -47,13 +47,14 @@ import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
 import com.revenuecat.purchases.paywalls.components.properties.FitMode
 import com.revenuecat.purchases.paywalls.components.properties.ImageUrls
 import com.revenuecat.purchases.paywalls.components.properties.Size
-import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fill
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fit
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fixed
 import com.revenuecat.purchases.paywalls.components.properties.ThemeImageUrls
 import com.revenuecat.purchases.ui.revenuecatui.R
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toContentScale
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.urlsForCurrentTheme
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.aspectRatio
 import com.revenuecat.purchases.ui.revenuecatui.components.modifier.overlay
 import com.revenuecat.purchases.ui.revenuecatui.components.modifier.size
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.rememberColorStyle
@@ -90,6 +91,7 @@ internal fun ImageComponentView(
             urlString = imageState.imageUrls.webp.toString(),
             modifier = modifier
                 .size(imageState.size)
+                .applyIfNotNull(imageState.aspectRatio) { aspectRatio(it) }
                 .applyIfNotNull(overlay) { overlay(it, imageState.shape ?: RectangleShape) }
                 .applyIfNotNull(imageState.shape) { clip(it) },
             placeholderUrlString = imageState.imageUrls.webpLowRes.toString(),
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -229,3 +244,35 @@ private fun Context.getRevenueCatUIImageLoader(readCache: Boolean): ImageLoader
         .memoryCachePolicy(cachePolicy)
         .build()
 }
+
+/**
+ * This is loosely based on [Accompanist's Drawable Painter](https://google.github.io/accompanist/drawablepainter/).
+ * This is not production-quality code and should only be used for Previews. If we ever have a need for this, it's
+ * better to use the Accompanist Drawable Painter library directly.
+ */
+private class DrawablePainter(
```
tonidero, line 253: I'm wondering if it's worth mark it as experimental even if it's private code, so people don't use it without looking at the doc above. Wdyt?

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -236,3 +331,80 @@ private fun previewEmptyState(): PaywallState.Loaded.Components {
...
+        }
+        .build()
+}
+
+private fun previewThemeImageUrls(widthPx: UInt, heightPx: UInt): ThemeImageUrls =
+    ThemeImageUrls(
+        light = ImageUrls(
+            original = URL("https://preview"),
+            webp = URL("https://preview"),
+            webpLowRes = URL("https://preview"),
```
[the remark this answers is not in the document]
JayShortway: I usually make these as invalid as possible to avoid accidentally hitting a live, untrusted endpoint. I guess we can use a domain we own to limit that risk, but we really don't want a network connection here. So maybe that's not worth it? This also screams "fake", which is good in this case I think haha.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -236,3 +331,80 @@ private fun previewEmptyState(): PaywallState.Loaded.Components {
...
+        .components {
+            add { chain ->
+                SuccessResult(
+                    drawable = BitmapDrawable(
+                        chain.request.context.resources,
+                        context.getDrawable(resource)!!.toBitmap(
+                            width = imageUrls.width,
+                            height = imageUrls.height,
+                            // Create a deterministic color from the URL and size.
+                            background = with(imageUrls) { "$original:$width$height".toRgbColor() },
```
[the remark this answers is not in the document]
JayShortway: It's definitely not needed, but the idea was to make it easier to differentiate the images (for humans). A 400x400 image will be a different color than a 400x200 image. This helps to identify these source images as different, because with the right `size` and `fitMode` options they can be rendered to look very similar.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
JayShortway: I like it! Will make the change.

JayShortway: Done here: https://github.com/RevenueCat/purchases-android/pull/2029/commits/a6959c2f879cce305c7de00fab875b6b7b0f3591. I split it out in a separate method so we can `@OptIn` as narrowly as possible. Let me know what you think!

PR #2030, opened 2025-01-07
[Paywalls V2] Handles gradient percentages in the 0..100 range

JayShortway:
## Description
The backend sends gradient percentages in the 0..100 range, but the code assumed it was in the 0..1 range. This PR fixes that. 

The fix is tiny obviously. The rest of the changes are to make sure previews don't change.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/ColorInfo.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/ColorInfo.kt
@@ -57,6 +57,9 @@ sealed interface ColorInfo {
             @Serializable(with = RgbaStringArgbColorIntDeserializer::class)
             @ColorInt
             @get:JvmSynthetic val color: Int,
+            /**
+             * A percentage value in the range 0.0..100.0.
+             */
             @get:JvmSynthetic val percent: Float,
         )
     }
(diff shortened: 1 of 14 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
```diff
@@ -71,7 +76,7 @@ internal fun ColorInfo.toColorStyle(): ColorStyle {
 }
 
 private fun List<ColorInfo.Gradient.Point>.toColorStops(): Array<Pair<Float, Color>> =
-    map { point -> point.percent to Color(point.color) }
+    map { point -> point.percent / PERCENT_SCALE to Color(point.color) }
```
JayShortway, line 79: This is the actual fix.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/ColorInfo.kt
```diff
@@ -57,6 +57,9 @@ sealed interface ColorInfo {
             @Serializable(with = RgbaStringArgbColorIntDeserializer::class)
             @ColorInt
             @get:JvmSynthetic val color: Int,
+            /**
+             * A percentage value in the range 0.0..100.0.
```
tonidero, line 61: Nitpick should this be ints intead of floats?
```suggestion
             * A percentage value in the range 0..100.
```

tonidero (approved):
LGTM!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/ColorInfo.kt
JayShortway: Oh yea the underlying field is a `Float` so that's why, but I think for documentation purposes an int is clearer, especially with all the dots haha. Thanks!

Issue #2006, opened 2024-12-22
`NotSerializableException` when serializing `PurchasesException`


MMP0:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**

`PurchasesException` inherits from `Exception` which inherits from `Throwable` which implements `Serializable`, so it must be “serializable” but it’s not.
When you try to serialize a `PurchasesException`, a `NotSerializableException` is thrown because the underlying `PurchasesError` doesn’t implement `Serializable`.

In the case of my app (which keeps some `Exception`s for error dialog), it crashes when `onSaveInstanceState` is called.

1. Environment
   1. Platform: Android
   2. SDK version: 8.10.6
   3. OS version: 15
   4. Android Studio version:
   5. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
   1. Try to serialize `PurchasesException`.
   
   Expected: `PurchasesException` is serialized.
   Actual: `NotSerializableException` is thrown.
5. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)
   Sample code:
   ```kotlin
   var exception = PurchasesException(PurchasesError(PurchasesErrorCode.UnknownError, "Serialize test."))
   
   @Suppress("USELESS_IS_CHECK")
   println(exception is Serializable)  // <- true of course
   
   var bundle = Bundle().apply {
        putSerializable("exception", exception)
   }
   
   val parcel = Parcel.obtain()
   try {
        // Write
        parcel.writeBundle(bundle)  // <- CRASH!
   
        // Read
        parcel.setDataPosition(0)
        bundle = parcel.readBundle(javaClass.classLoader)!!
   } finally {
        parcel.recycle()
   }
   
   exception = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
        bundle.getSerializable("exception", PurchasesException::class.java)
   } else {
[...]

PR #2031, opened 2025-01-08
Make `PurchasesError` `java.io.Serializable`

tonidero:
### Description
This was requested in #2006

### DIFF
--- a/purchases/src/androidTest/kotlin/com/revenuecat/purchases/ErrorsTest.kt
+++ b/purchases/src/androidTest/kotlin/com/revenuecat/purchases/ErrorsTest.kt
@@ -23,6 +23,8 @@ class ErrorsTest {
         val parcel = Parcel.obtain()
         val readBundle: Bundle?
         try {
+            // Write the bundle to a parcel in order to actually perform the serialization. Without this,
+            // the bundle will not be serialized and the test will be successful while not actually being serializable.
             parcel.writeBundle(bundle)
             parcel.setDataPosition(0)
             readBundle = parcel.readBundle(javaClass.classLoader)!!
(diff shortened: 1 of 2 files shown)


purchases/src/androidTest/kotlin/com/revenuecat/purchases/ErrorsTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        }
+        val parcel = Parcel.obtain()
+        val readBundle: Bundle?
+        try {
+            parcel.writeBundle(bundle)
+            parcel.setDataPosition(0)
+            readBundle = parcel.readBundle(javaClass.classLoader)!!
+        } finally {
+            parcel.recycle()
+        }
```
JayShortway, line 31: I'm curious why we're writing the bundle to a parcel and reading it back. Does it otherwise not actually serialize the values? Would be helpful to have a comment here I think 😄

JayShortway (approved):
Thanks for taking care of this! Just a request for a comment.

purchases/src/androidTest/kotlin/com/revenuecat/purchases/ErrorsTest.kt
tonidero: Pretty much yeah... If I just pass it to the bundle it doesn't actually crash. It's only when I write it to a parcel that it fails.

tonidero: Added comment in: https://github.com/RevenueCat/purchases-android/pull/2031/commits/c29c251024e476feb7ea7c6b36a140147176aef1

JayShortway: Nice, I read about this behavior some time ago. Thanks for adding the comment!

PR #2033, opened 2025-01-08
[Paywalls V2] Fixes click consumption in text buttons

JayShortway:
## Description
`ButtonComponent`s containing a `TextComponent` were not responding to clicks properly.

## Cause
This is caused by the `pointerInput` modifier added to the `Markdown` composable consuming all the clicks.

## Fix
The proper fix is to use a `LinkAnnotation` introduced in Compose UI 1.7.0, which was only a single-month BOM update away. I figured this is worth it, and so that's what this PR implements. 

An alternative fix would be to conditionally add the `pointerInput` modifier only if the markdown text contains links. However, that would still mean that a button with text that is partially a link would only allow clicking the link. The button wouldn't function. An edge case, sure, but not the expected behavior imo.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
```diff
@@ -304,7 +304,7 @@ private fun TextComponentView_Preview_Customizations() {
 private fun TextComponentView_Preview_Markdown() {
     TextComponentView(
         style = previewTextComponentStyle(
-            text = "Hello, **bold**, *italic* or _italic2_ with ~strikethrough~ and `monospace`. " +
+            text = "Hello, **bold**, *italic* or _italic2_ with ~strikethrough~, ~~strikethrough2~~ and `monospace`. " +
```
JayShortway, line 307: This is only tangentially related (it's also markdown) but wanted to add the `~~` strikethrough case to make sure it works (it does!) and didn't know where else to put it. Felt a bit overkill to open a PR for this alone 😅

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallActionTests.kt
```diff
@@ -134,7 +134,7 @@ class PaywallActionTests {
         when (this) {
             is PaywallAction.NavigateBack -> ButtonComponent.Action.NavigateBack
             is PaywallAction.NavigateTo -> ButtonComponent.Action.NavigateTo(destination)
-            is PaywallAction.RestorePurchases -> ButtonComponent.Action.NavigateBack
+            is PaywallAction.RestorePurchases -> ButtonComponent.Action.RestorePurchases
```
JayShortway, line 137: This was breaking a test locally. Not sure how this got in.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
```diff
@@ -116,7 +116,7 @@ internal fun TextComponentView(
             fontSize = textState.fontSize.toTextUnit(),
             fontWeight = textState.fontWeight,
             fontFamily = textState.fontFamily,
-            horizontalAlignment = textState.horizontalAlignment,
+            // horizontalAlignment = textState.horizontalAlignment,
```
tonidero, line 119: Is this intended?

tonidero (approved):
Nice! Looks great! Just a comment about a leftover commented code

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
JayShortway: Absolutely not, thanks for catching this! 🙏 https://github.com/RevenueCat/purchases-android/pull/2033/commits/65cdf4d8c2cc14ea4a5b6175e6dd935d502c06e9

PR #2035, opened 2025-01-08
[Paywalls V2] Adds Custom Tabs to support in-app browser URL destinations

JayShortway:
## Description
As the title says. I think using Custom Tabs is the most pragmatic solution to supporting in-app browsers, and also the most user-friendly. For instance, it shares cookies with their external browser so they remain logged in on websites. (In fact, it _is_ their external browser, just presented in-app.) This is also closest to the `SafariView` approach we use on iOS. We can iterate on this and e.g. add more styling as we go. 

## Discarded approach
An alternative would be to use a `WebView`, but this brings a whole lot of complexities. We'd need to build navigation logic for users to go back and forward on web pages. We'd need to build UI for this navigation, as well as a close button and maybe a refresh button. We'd also need to build a Compose wrapper for `WebView`, as that's a View. (We could use [Accompanist's WebView](https://google.github.io/accompanist/web/), but that's deprecated.) 

We get all of the above for free with Custom Tabs.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -288,19 +290,31 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
...
-                getString(R.string.no_browser_cannot_open_link)
-            } else {
-                getString(R.string.cannot_open_link)
+        ButtonComponent.UrlMethod.IN_APP_BROWSER -> {
+            val intent = CustomTabsIntent.Builder()
+                .build()
+            @Suppress("TooGenericExceptionCaught")
+            try {
+                intent.launchUrl(this, Uri.parse(url))
+            } catch (e: Exception) {
```
tonidero, line 310: I *think* this should never throw a `Throwable` right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -288,19 +290,31 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
...
-        ButtonComponent.UrlMethod.IN_APP_BROWSER -> TODO()
-        ButtonComponent.UrlMethod.EXTERNAL_BROWSER,
-        ButtonComponent.UrlMethod.DEEP_LINK,
-        -> openUriOrElse(url) { exception ->
-            val message = if (exception is ActivityNotFoundException) {
-                getString(R.string.no_browser_cannot_open_link)
-            } else {
-                getString(R.string.cannot_open_link)
+        ButtonComponent.UrlMethod.IN_APP_BROWSER -> {
+            val intent = CustomTabsIntent.Builder()
```
tonidero, line 305: Hmm I foresee some devs might want to be able to customize this down the line... But I think this is ok for now and we can see what devs think about it 👍

tonidero (approved):
LGTM! Nicely done!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
JayShortway: Yea agreed!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
tonidero: I think either works... I'm a bit on the side of avoiding crashes, but I think just catching the known exceptions here makes sense 👍

JayShortway: I did that for now, as our `openUriOrElse` does the same thing. 

Since devs can't provide a custom `Logger` (yet), unknown exceptions in the wild can fly under the radar, which is what I'm afraid of.

Issue #2037, opened 2025-01-09
Serialization issue with `PurchasesError` (related to #2006)


MMP0:
- [ ] I have updated Purchases SDK to the latest version
- [ ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [ ] I have searched the [Community](https://community.revenuecat.com)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ ] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
@tonidero Thank you for fixing #2006 (in #2031)!

https://github.com/RevenueCat/purchases-android/blob/92eeff9d822fbb5dd45ebffdc297a9e76b6c44a3/purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt#L26-L28

The `message` field of `PurchasesError` is annotated with `@IgnoredOnParcel` annotation; for `Serializable`, it may need to be annotated with `@Transient` annotation to exclude it from serialization.

```kotlin
@IgnoredOnParcel
@Transient
val message: String = code.description
```

Personally, I think it would be better to change from field to getter:

```kotlin
val message: String
    get() = code.description
```

1. Environment
   1. Platform: Android
   2. SDK version: e3fce6c81eae62c4193ab6145206a4c6a0540258
   3. OS version: 15
   4. Android Studio version:
   5. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
   1. Serialize `PurchasesError`.
   
   Expected: The `message` field is skipped for serialization.
   Actual: The `message` field is serialized.
5. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)
   Sample code:
   
   ```kotlin
   operator fun ByteArray.contains(subArray: ByteArray): Boolean {
       if (subArray.isEmpty() || subArray.size > this.size) return false

       for (i in 0..this.size - subArray.size) {
           var found = true
           for (j in subArray.indices) {
               if (this[i + j] != subArray[j]) {
                   found = false
                   break
               }
           }
           if (found) return true
       }
       return false
   }
   
   val baos = ByteArrayOutputStream()
   ObjectOutputStream(baos).use {
[...]

PR #2038, opened 2025-01-09
Convert error message property into computed property

tonidero:
### Description
Addressing #2037

This makes it so the message is not part of the serialization, since it's not needed

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
@@ -1,7 +1,6 @@
 package com.revenuecat.purchases
 
 import android.os.Parcelable
-import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import java.io.Serializable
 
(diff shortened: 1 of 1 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
```diff
@@ -24,8 +23,8 @@ class PurchasesError(
     }
 
     // Message explaining the error
-    @IgnoredOnParcel
-    val message: String = code.description
+    val message: String
+        get() = code.description
```
JayShortway, line 27: Just curious, why do we actually not want to parcelize/serialize this?

JayShortway (approved):
LGTM! Just a question.

purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
tonidero: Well, I think it's already part of the `code` itself, so it doesn't provide extra data. As in, it's duplicated data IMO... Lmk if you think we should serialize it for any reason though!

JayShortway: Oh no, you're right. Makes sense! Thanks for explaining.

PR #2039, opened 2025-01-09
[Paywalls V2] Add `Badge`'s `edgeToEdge` `Top`/`Bottom` alignment style layout

tonidero:
### Description
Follow-up to #2009 
This PR adds the `edgeToEdge` top/bottom style for badges.
image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
@@ -105,8 +105,8 @@ internal class StackComponentState(
     val shape by derivedStateOf { presentedPartial?.partial?.shape?.toShape() ?: style.shape }
 
     @get:JvmSynthetic
-    val cornerRadiuses by derivedStateOf {
-        (presentedPartial?.partial?.shape as? Shape.Rectangle)?.corners ?: style.cornerRadiuses
+    val rcShape by derivedStateOf {
+        (presentedPartial?.partial?.shape as? Shape.Rectangle)?.corners ?: style.rcShape
     }
 
     @get:JvmSynthetic
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
@@ -13,7 +13,6 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.unit.dp
 import androidx.window.core.layout.WindowWidthSizeClass
 import com.revenuecat.purchases.Package
-import com.revenuecat.purchases.paywalls.components.properties.Shape
 import com.revenuecat.purchases.ui.revenuecatui.components.ComponentViewState
 import com.revenuecat.purchases.ui.revenuecatui.components.ScreenCondition
 import com.revenuecat.purchases.ui.revenuecatui.components.buildPresentedPartial
@@ -106,9 +105,7 @@ internal class StackComponentState(
     val shape by derivedStateOf { presentedPartial?.partial?.shape?.toShape() ?: style.shape }
 
     @get:JvmSynthetic
-    val rcShape by derivedStateOf {
-        (presentedPartial?.partial?.shape as? Shape.Rectangle)?.corners ?: style.rcShape
-    }
+    val rcShape by derivedStateOf { presentedPartial?.partial?.shape ?: style.rcShape }
 
     @get:JvmSynthetic
     val border by derivedStateOf { presentedPartial?.partial?.border ?: style.border }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
@@ -17,7 +17,6 @@ import com.revenuecat.purchases.ui.revenuecatui.components.ComponentViewState
 import com.revenuecat.purchases.ui.revenuecatui.components.ScreenCondition
 import com.revenuecat.purchases.ui.revenuecatui.components.buildPresentedPartial
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toPaddingValues
-import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toShape
 import com.revenuecat.purchases.ui.revenuecatui.components.style.BadgeStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.style.StackComponentStyle
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallState
@@ -102,10 +101,7 @@ internal class StackComponentState(
     val margin by derivedStateOf { presentedPartial?.partial?.margin?.toPaddingValues() ?: style.margin }
 
     @get:JvmSynthetic
-    val shape by derivedStateOf { presentedPartial?.partial?.shape?.toShape() ?: style.shape }
-
-    @get:JvmSynthetic
-    val rcShape by derivedStateOf { presentedPartial?.partial?.shape ?: style.rcShape }
+    val shape by derivedStateOf { presentedPartial?.partial?.shape ?: style.shape }
 
     @get:JvmSynthetic
     val border by derivedStateOf { presentedPartial?.partial?.border ?: style.border }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
@@ -12,6 +12,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.unit.dp
 import androidx.window.core.layout.WindowWidthSizeClass
+import com.revenuecat.purchases.paywalls.components.properties.Shape
 import com.revenuecat.purchases.ui.revenuecatui.components.ComponentViewState
 import com.revenuecat.purchases.ui.revenuecatui.components.ScreenCondition
 import com.revenuecat.purchases.ui.revenuecatui.components.buildPresentedPartial
@@ -102,6 +103,11 @@ internal class StackComponentState(
 
     @get:JvmSynthetic
     val shape by derivedStateOf { presentedPartial?.partial?.shape?.toShape() ?: style.shape }
+
+    @get:JvmSynthetic
+    val cornerRadiuses by derivedStateOf {
+        (presentedPartial?.partial?.shape as? Shape.Rectangle)?.corners ?: style.cornerRadiuses
+    }
 
     @get:JvmSynthetic
     val border by derivedStateOf { presentedPartial?.partial?.border ?: style.border }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -7,31 +7,28 @@ import android.content.res.Configuration
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.CornerSize
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.derivedStateOf
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.RectangleShape
 import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.layout.Placeable
+import androidx.compose.ui.layout.SubcomposeLayout
 import androidx.compose.ui.layout.layout
-import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.unit.Density
 import androidx.compose.ui.unit.dp
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.paywalls.components.StackComponent
@@ -94,6 +91,7 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.toComponentsPaywallState
 import com.revenuecat.purchases.ui.revenuecatui.helpers.validatePaywallComponentsDataOrNull
 import java.net.URL
 import kotlin.math.roundToInt
+import androidx.compose.ui.geometry.Size as ComposeSize
 
 @Suppress("LongMethod")
 @Composable
@@ -132,7 +130,7 @@ internal fun StackComponentView(
                         stackState,
                         state,
                         badge.stackStyle,
-                        badge.alignment,
+                        badge.alignment.isTop,
                         clickHandler,
                         modifier,
                         selected,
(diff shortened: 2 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -102,6 +103,11 @@ internal class StackComponentState(
     @get:JvmSynthetic
     val shape by derivedStateOf { presentedPartial?.partial?.shape?.toShape() ?: style.shape }
 
+    @get:JvmSynthetic
+    val cornerRadiuses by derivedStateOf {
```
tonidero, line 107: We need access to the corner radiuses from the `StackComponentView`. This is to properly calculate the height of the background. This is only available when using a `Rectangle` shape. We won't currently support other shapes for the badge (If other shape, we will just not use a shape, so a sharp rectangle).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -117,9 +125,19 @@ internal fun StackComponentView(
...
+                        badge.stackStyle,
+                        badge.alignment,
+                        clickHandler,
+                        modifier,
+                        selected,
+                    )
                 }
 
                 Badge.Style.Nested -> {
+                    MainStackComponent(stackState, state, clickHandler, modifier, selected)
```
tonidero, line 140: This still needs to be done in follow up PRs.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -171,6 +189,112 @@ private fun StackWithOverlaidBadge(
     }
 }
 
+@Suppress("LongMethod", "LongParameterList", "CyclomaticComplexMethod")
+@Composable
+private fun StackWithEdgeToEdgeBadge(
```
tonidero, line 194: This got quite big... I was thinking about ways to extract this, but it got tricky... I thought to leave it like this for now and try to clean it up later.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -171,6 +189,112 @@ private fun StackWithOverlaidBadge(
...
+    val badgeHeightDp = with(LocalDensity.current) { badgeHeight.toDp() }
+    val badge = @Composable {
+        StackComponentView(
+            // The background and border is applied to the parent container so we null it out here.
+            // We make the badge use all the available width without increasing the size of the main content.
+            badgeStack.copy(
+                backgroundColor = null,
+                size = Size(width = Fill, height = badgeStack.size.height),
+                border = null,
+                margin = PaddingValues(0.dp),
```
tonidero, line 215: We basically clear the properties that are already applied to the background composable so they are not applied again to the badge itself.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -171,6 +189,112 @@ private fun StackWithOverlaidBadge(
...
+    val backgroundShape = badgeStack.cornerRadiuses?.let { cornerRadiuses ->
+        val filteredCornerRadiuses = when (alignment) {
+            TwoDimensionalAlignment.TOP_LEADING,
+            TwoDimensionalAlignment.TOP,
+            TwoDimensionalAlignment.TOP_TRAILING,
+            -> cornerRadiuses.copy(bottomLeading = 0.0, bottomTrailing = 0.0)
+            TwoDimensionalAlignment.BOTTOM_LEADING,
+            TwoDimensionalAlignment.BOTTOM,
+            TwoDimensionalAlignment.BOTTOM_TRAILING,
+            -> cornerRadiuses.copy(topLeading = 0.0, topTrailing = 0.0)
```
tonidero, line 245: Here I'm clearing some of the corner radiuses for the badge, so it reaches the expected point.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -632,21 +819,27 @@ private fun previewBadge(alignment: TwoDimensionalAlignment): BadgeStyle {
                     overrides = null,
                 ),
             ),
-            dimension = Dimension.Horizontal(alignment = VerticalAlignment.CENTER, distribution = START),
+            dimension = Dimension.Vertical(alignment = HorizontalAlignment.CENTER, distribution = CENTER),
```
tonidero, line 822: Made this change in the preview to properly center the content of the badge. I think the backend should also return the same style, but will test this again once it has been published to the dashboard.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
JayShortway: I don't think this is needed anymore if `onGloballyPositioned { }` is removed, right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -143,6 +166,112 @@ private fun StackWithOverlaidBadge(
...
+            stackState,
+            state,
+            clickHandler,
+            selected = selected,
+        )
+    }
+    val backgroundColorStyle = badgeStack.backgroundColor?.let { rememberColorStyle(scheme = it) }
+    val borderStyle = badgeStack.border?.let { rememberBorderStyle(border = it) }
+    val shadowStyle = badgeStack.shadow?.let { rememberShadowStyle(shadow = it) }
+    val backgroundShape = badgeStack.cornerRadiuses?.let { cornerRadiuses ->
```
JayShortway, line 213: Regarding not supporting pill-shaped stacks: maybe we can concede to using `onGloballyPositioned { }` for this? 😅 We could get the height of the stack, and use half of the height to determine the corner radius for the badge background shape.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: That sounds good to me!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StackComponentStyle.kt
```diff
@@ -32,11 +33,44 @@ internal class StackComponentStyle(
...
     @get:JvmSynthetic
     val shadow: Shadow?,
     @get:JvmSynthetic
     val badge: BadgeStyle?,
     @get:JvmSynthetic
     val overrides: PresentedOverrides<PresentedStackPartial>?,
-) : ComponentStyle
+) : ComponentStyle {
+    @JvmSynthetic
+    internal fun copy(
```
JayShortway, line 47: I would not be terribly opposed to making this a data class, as it's already internal. (Same for `CornerRadiuses`.) However no need to change it imo, as the function is already written haha.

JayShortway (commented):
Nice nice! Had a few comments, but nearly there I think.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -143,6 +166,112 @@ private fun StackWithOverlaidBadge(
...
+                badgeHeightDp + maxOf(it.topLeading, it.topTrailing, it.bottomLeading, it.bottomTrailing).dp
+            } ?: badgeHeightDp
+        }
+    }
+    Box(modifier = modifier.width(IntrinsicSize.Min)) {
+        Box(
+            modifier = Modifier
+                .align(alignment.toAlignment())
+                .fillMaxWidth()
+                .height(backgroundHeight)
```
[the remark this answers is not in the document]
tonidero: Hmm if we do it like this, the background would extend over the whole parent. We would need to make sure the bottom/top (depending on alignment) of the background is clipped the same way as the "main" stack, instead of the badgeStack to make sure it doesn't peek out of the "main" stack?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -143,6 +166,112 @@ private fun StackWithOverlaidBadge(
...
+                    badge()
+                    mainContent()
+                }
+                TwoDimensionalAlignment.BOTTOM_LEADING,
+                TwoDimensionalAlignment.BOTTOM,
+                TwoDimensionalAlignment.BOTTOM_TRAILING,
+                -> {
+                    mainContent()
+                    badge()
+                }
```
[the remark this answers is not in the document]
tonidero: Ohhh I totally missed that... thanks! And yeah, I can work on that, but would like to extract to a separate PR if possible.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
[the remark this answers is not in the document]
tonidero: Hmm I guess this wouldn't work if the main stack is using a "Pill" shape, but it's not like we're supporting that case very well right now either...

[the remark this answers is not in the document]
JayShortway: >Hmm if we do it like this, the background would extend over the whole parent. 
  
Yes true, but it kinda looks like that [on Figma](https://www.figma.com/design/H7zMUjYBh7rQAmVNJelAou/Paywalls?node-id=6417-36028&t=9DT3HOQk4UWfSusK-0) too. Not sure how required that is but we could easily support that using a tiny bit of padding.

>Hmm I guess this wouldn't work if the main stack is using a "Pill" shape, but it's not like we're supporting that case very well right now either...

True, let me see if I can get the `SubcomposeLayout` thing to work.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Seems like we can (linking for posterity): https://github.com/RevenueCat/purchases-android/pull/2046

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -143,6 +159,134 @@ private fun StackWithOverlaidBadge(
...
+        val totalHeight = stackPlaceable.height + badgeHeight
+        val backgroundHeight = badgeHeight + halfStackHeight
+
+        // Subcompose the background
+        val backgroundMeasurable = subcompose("background") {
+            val backgroundColorStyle = badgeStack.backgroundColor?.let { rememberColorStyle(scheme = it) }
+            val borderStyle = badgeStack.border?.let { rememberBorderStyle(border = it) }
+            val shadowStyle = badgeStack.shadow?.let { rememberShadowStyle(shadow = it) }
+            val backgroundShape = when (badgeStack.rcShape) {
+                is Shape.Pill -> null // Pill badges not supported in edgeToEdge
```
tonidero, line 222: I changed a few things in my last commit @JayShortway. Same as in iOS, I removed support for pill badges. I think we can support them later if needed. Then, for the background, I'm now only extending it to half the height of the main stack. This is so it's not visible in the other side of the main stack, only the side where the badge is. 

Happy to chat about these changes tomorrow!

JayShortway: I'm always for simplicity and removing obsolete code, but if we want to support them later, it's beneficial for devs if we support them from the first Paywalls V2 SDK release onwards. It's not like we still need to write the code to support this case, we already have it.

Regarding the background extending to half the height, I'm not sure if we can get away with that. If you look at this example, the teal edge-to-edge badge is visible all the way to the bottom of the stack. 

image

Let me know what you think!

tonidero: Ok brought back pill badge support [here](https://github.com/RevenueCat/purchases-android/pull/2039/commits/661ce4d21b43175349279f75f81e99e4cdc29027) and, as we talked over slack, we are now extending the height of the background up to covering the rounded corners

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -104,6 +104,9 @@ internal class StackComponentState(
     @get:JvmSynthetic
     val shape by derivedStateOf { presentedPartial?.partial?.shape?.toShape() ?: style.shape }
 
+    @get:JvmSynthetic
+    val rcShape by derivedStateOf { presentedPartial?.partial?.shape ?: style.rcShape }
```
JayShortway, line 108: I'd say we can remove `shape` to avoid confusion (and rename `rcShape` to `shape`). Otherwise we'd have 2 shape fields that we need to keep in sync, and it's not automatically clear why we have 2.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -142,6 +160,164 @@ private fun StackWithOverlaidBadge(
...
+/**
+ * @param topBadge Whether the badge should be placed on the top or bottom of the stack.
+ */
+@Suppress("LongMethod", "LongParameterList", "CyclomaticComplexMethod")
+@Composable
+private fun StackWithEdgeToEdgeBadge(
+    stackState: StackComponentState,
+    state: PaywallState.Loaded.Components,
+    badgeStack: StackComponentStyle,
+    topBadge: Boolean,
```
JayShortway, line 172: Minor, maybe the PR title can reflect that this only implements top/bottom alignments.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -142,6 +160,164 @@ private fun StackWithOverlaidBadge(
...
+            is Shape.Rectangle -> {
+                (
+                    (stackState.rcShape as? Shape.Rectangle)?.let { rcShape ->
+                        if (topBadge) {
+                            maxOf(rcShape.corners?.topLeading ?: 0.0, rcShape.corners?.topTrailing ?: 0.0)
+                        } else {
+                            maxOf(rcShape.corners?.bottomLeading ?: 0.0, rcShape.corners?.bottomTrailing ?: 0.0)
+                        }
+                    } ?: 0.0
+                    ).dp.toPx().roundToInt()
```
JayShortway, line 218: Minor, more of an FYI:

```suggestion
                    ).dp.roundToPx()
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -746,22 +1068,26 @@ private fun previewBadge(alignment: TwoDimensionalAlignment): BadgeStyle {
...
-            dimension = Dimension.Horizontal(alignment = VerticalAlignment.CENTER, distribution = START),
+            dimension = Dimension.Vertical(
+                alignment = HorizontalAlignment.CENTER,
+                distribution = FlexDistribution.CENTER,
+            ),
             size = Size(width = Fit, height = Fit),
             spacing = 0.dp,
             backgroundColor = ColorScheme(
                 light = ColorInfo.Hex(Color.Green.toArgb()),
             ),
```
JayShortway, line 1079: Maybe we can give the badge a vertical gradient background? That way we ensure it doesn't break.

JayShortway (approved):
Very nice job! Had a few suggestions here and there, but overall looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
tonidero: That's much cleaner thanks! Done in https://github.com/RevenueCat/purchases-android/pull/2039/commits/281887b75042461d9cd6d3cf0b2bcfdd2a08b686

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Looks like it's diagonal now (`45f`), which also works. My idea with vertical was that we can immediately see if we accidentally change the height of the background, as the vertical gradient would look more (or less) stretched then.

tonidero:
Will merge this but can address further feedback in follow-up PRs!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: I think you're right! We still have to fix the gradient angle so it's also correct for non-squares (different topic), but I think it still works. Good idea!

PR #2041, opened 2025-01-09
[Paywalls V2] Add `Badge`'s `nested` style layout

tonidero:
### Description
Followup to #2009 and #2039.

This adds support for `nested` style badges.

As discussed over Slack, the nested badges won't move the content of the stack, and just be placed on top.
image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -137,7 +137,19 @@ internal fun StackComponentView(
                 }
 
                 Badge.Style.Nested -> {
-                    MainStackComponent(stackState, state, clickHandler, modifier, badge)
+                    val stackModifier = calculateStackModifier(stackState)
+                    Box(modifier = modifier.then(stackModifier)) {
+                        // We avoid applying the same margin, border, shadow and padding twice by
+                        // applying an empty commonModifier
+                        MainStackComponent(stackState, state, clickHandler, Modifier, Modifier)
+                        StackComponentView(
+                            badge.stackStyle,
+                            state,
+                            {},
+                            modifier = Modifier
+                                .align(badge.alignment.toAlignment()),
+                        )
+                    }
                 }
             }
         } else {
@@ -352,25 +364,8 @@ private fun MainStackComponent(
     state: PaywallState.Loaded.Components,
     clickHandler: suspend (PaywallAction) -> Unit,
     modifier: Modifier = Modifier,
-    nestedBadge: BadgeStyle? = null,
+    commonModifier: Modifier = calculateStackModifier(stackState),
 ) {
-    val backgroundColorStyle = stackState.backgroundColor?.let { rememberColorStyle(scheme = it) }
-    val borderStyle = stackState.border?.let { rememberBorderStyle(border = it) }
-    val shadowStyle = stackState.shadow?.let { rememberShadowStyle(shadow = it) }
-    val composeShape by remember(stackState.shape) { derivedStateOf { stackState.shape.toShape() } }
-
-    // Modifier irrespective of dimension.
-    val commonModifier = remember(stackState, backgroundColorStyle, borderStyle, shadowStyle) {
-        Modifier
-            .padding(stackState.margin)
-            .applyIfNotNull(shadowStyle) { shadow(it, composeShape) }
-            .applyIfNotNull(backgroundColorStyle) { background(it, composeShape) }
-            .clip(composeShape)
-            .applyIfNotNull(borderStyle) { border(it, composeShape) }
-            .padding(stackState.padding)
-            .padding(stackState.dimension, stackState.spacing)
-    }
-
     val content: @Composable ((ComponentStyle) -> Modifier) -> Unit = remember(stackState.children) {
         @Composable { modifierProvider ->
             stackState.children.forEach { child ->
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Alignment.kt
```diff
@@ -73,3 +73,31 @@ internal fun TwoDimensionalAlignment.toHorizontalAlignmentOrNull(): Alignment.Ho
         TwoDimensionalAlignment.BOTTOM_LEADING -> Alignment.Start
         TwoDimensionalAlignment.BOTTOM_TRAILING -> Alignment.End
     }
+
+@JvmSynthetic
+internal fun TwoDimensionalAlignment.toVerticalAlignment(): Alignment.Vertical =
```
tonidero, line 78: This is very similar to `toVerticalAlignmentOrNull` but I needed one that assumed a vertical alignment for the "center" types. Same for the horizontal version.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -211,15 +214,9 @@ private fun StackWithEdgeToEdgeBadge(
...
-            TwoDimensionalAlignment.TOP,
-            TwoDimensionalAlignment.TOP_TRAILING,
-            -> cornerRadiuses.copy(bottomLeading = 0.0, bottomTrailing = 0.0)
-            TwoDimensionalAlignment.BOTTOM_LEADING,
-            TwoDimensionalAlignment.BOTTOM,
-            TwoDimensionalAlignment.BOTTOM_TRAILING,
-            -> cornerRadiuses.copy(topLeading = 0.0, topTrailing = 0.0)
+        val filteredCornerRadiuses = when (alignment.toVerticalAlignment()) {
+            Alignment.Top -> cornerRadiuses.copy(bottomLeading = 0.0, bottomTrailing = 0.0)
+            Alignment.Bottom -> cornerRadiuses.copy(topLeading = 0.0, topTrailing = 0.0)
```
tonidero, line 219: These are just some cleanup from just using the respective alignment type.

tonidero (commented):
I will really like to test this again once the option is in the dashboard. For now, on the previews, this seems to be working.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -383,37 +385,54 @@ private fun MainStackComponent(
...
+        Box(modifier = modifier.then(commonModifier)) {
+            stack(Modifier)
+            StackComponentView(
+                nestedBadge.stackStyle,
+                state,
+                {},
+                modifier = Modifier
+                    .align(nestedBadge.alignment.toAlignment()),
+            )
+        }
```
JayShortway, line 435: Could we move this to the `StackComponentView` at the top, inside the `when` checking the `Badge.Style`? 

```kotlin
Badge.Style.Nested -> {
    Box(modifier = modifier) {
        MainStackComponent(stackState, state, clickHandler, modifier)
        StackComponentView(
            badge.stackStyle,
            state,
            {},
            modifier = Modifier
                .align(badge.alignment.toAlignment()),
        )
    }
}
```

That way we don't need to add a `nestedBadge` parameter to the `MainStackComponent`. That parameter is slightly confusing imo, as it doesn't follow the implementation pattern of the other badges.

JayShortway (approved):
Very nice! Just 1 comment on structure.

JayShortway:
Seems that the stack's border is drawn on top of the badge, which is different from [leading/trailing edge-to-edge badges](https://github.com/RevenueCat/purchases-android/pull/2054/files). Is that intentional? Just wanted to double check.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Right the main reason for doing it this way is that we have to apply the `commonModifier` to the parent `Box`, not the `MainStackComponent` in this scenario... I agree it's a bit confusing... I'm trying to think of alternatives

tonidero:
> Seems that the stack's border is drawn on top of the badge, which is different from [leading/trailing edge-to-edge badges](https://github.com/RevenueCat/purchases-android/pull/2054/files). Is that intentional?

Yeah, with nested, it should be "inside" the stack itself, vs the edge-to-edge style which is considered "outside" the badge. If the backend gives us any margin, it would be added within the main stack.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Wdyt of [this approach](https://github.com/RevenueCat/purchases-android/pull/2041/commits/1409d50276ae7a93f84ff6c4fd1223d60974c7ba) @JayShortway ?

JayShortway: Thanks for explaining! 

I think I prefer the `nestedBadge` parameter approach then. 😳 Having 2 modifier parameters is more confusing imo. Sorry for the back and forth! 🙏

tonidero: Np! I was also back and forth on this 😅 Will revert that 👍

PR #2042, opened 2025-01-09
[Paywalls V2] `TextComponentView` uses the correct `Package` for variable values

JayShortway:
## Description
This fixes the bug where `TextComponentView` always uses values from the selected `Package` for its variables, even if it was a child of a `PackageComponentView`. 

## Additionally
Since we're passing the `Package` to the `ComponentStyle` now anyway (to fix the above bug), this can also be used to determine the `selected` state. The existing `selected` variable that many `*ComponentView`s had was obsolete. This PR refactors them to deduce the `selected` state from comparing their `Package` and the currently selected one.


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/pkg/PackageComponentViewTests.kt
```diff
@@ -235,4 +235,89 @@ class PackageComponentViewTests {
             .assertIsDisplayed()
     }
 
+    @Test
+    fun `Should take variable values from the correct package`(): Unit = with(composeTestRule) {
```
JayShortway, line 239: This is what this PR fixes.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
```diff
@@ -408,6 +497,82 @@ class TextComponentViewTests {
         onNodeWithText(expectedTextNlNl).assertIsDisplayed()
     }
 
+    @Test
+    fun `Should update variable values when the selected package changes`(): Unit = with(composeTestRule) {
```
JayShortway, line 501: This already passed, but wanted to make sure.

tonidero (approved):
Looks good! Great job!

PR #2044, opened 2025-01-10
Allow repurchasing custom packages

tonidero:
### Description
Allow repurchasing custom packages, independently of what the user is already subscribed to/purchased. Now that we support non-consumables in Android, as long as the product is marked as non-consumable we won't consume it and Google will stop repurchasing if needed, so this is safe to do now.


aboedo (approved):
change looks good, would be great to follow up with a test so we don't break it in the future

tonidero:
So adding a test for this is a bit trickier than expected... I just tested this manually for now and it works as expected. Will circle back to write a test for this this next week

PR #2045, opened 2025-01-10
Update `agp` to 8.8.0

tonidero:
### Description
After updating to the latest AS, I was getting this update suggested, and we haven't updated in a while.


JayShortway (approved):
Thanks for taking care of this!

PR #2048, opened 2025-01-10
[Paywalls V2] Correctly determines when to show or hide decimals for prices

JayShortway:
## Description
As the title says. This was not hooked up to V2 yet.


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
```diff
@@ -573,6 +577,109 @@ class TextComponentViewTests {
...
+            .assertTextEquals(expectedTextWithDecimals)
+
+        onNodeWithTag("country-nl")
+            .assertIsDisplayed()
+            .onChild()
+            .assertTextEquals(expectedTextWithDecimals)
+
+        onNodeWithTag("country-mx")
+            .assertIsDisplayed()
+            .onChild()
```
JayShortway, line 679: I really dislike this `onChild()` as it makes the test very brittle. If we ever change the implementation of `TextComponentView` without changing the way it looks, this test might still break. However I don't know of a better way currently.

tonidero: Hmm I was thinking about maybe having different prices for each country, and check that the prices labels were exactly the expected ones... but might be trickier... I'm ok with this for now!

tonidero (approved):
LGTM!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
JayShortway: That's a good idea! A quick win was to add a package with a different currency, which I did here: https://github.com/RevenueCat/purchases-android/pull/2048/commits/7e3e60ae8c36c1d36a952c90fde778fc7c8cb16f.

PR #2049, opened 2025-01-13
[Trusted Entitlements] Do not clear CustomerInfo upon enabling Trusted Entitlements

tonidero:
### Description
Until now, when devs enabled Trusted Entitlements, we would clear the cached `CustomerInfo` and customers would need to refetch the latest value from the server, which could potentially cause delays and/or losing access temporarily.

With this PR, we won't be clearing the cached CustomerInfo anymore. Instead, we would return the cached customer info, which will have a `VerificationResult` of `NOT_REQUESTED`. 

- [ ] Update docs

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -21,6 +21,7 @@ import com.revenuecat.purchases.common.diagnostics.DiagnosticsFileHelper
 import com.revenuecat.purchases.common.diagnostics.DiagnosticsHelper
 import com.revenuecat.purchases.common.diagnostics.DiagnosticsSynchronizer
 import com.revenuecat.purchases.common.diagnostics.DiagnosticsTracker
+import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.common.networking.ETagManager
 import com.revenuecat.purchases.common.offerings.OfferingsCache
@@ -113,9 +114,16 @@ internal class PurchasesFactory(
                 warnLog("Diagnostics are only supported on Android N or newer.")
             }
 
-            val signatureVerificationMode = SignatureVerificationMode.fromEntitlementVerificationMode(
-                verificationMode,
-            )
+            val signatureVerificationMode = try {
+                SignatureVerificationMode.fromEntitlementVerificationMode(
+                    verificationMode,
+                )
+            } catch (e: IllegalStateException) {
+                // If we're not able to create the signature verifier, we should disable signature verification
+                // instead of crashing
+                errorLog("Error creating signature verifier: ${e.message}. Disabling signature verification.")
+                SignatureVerificationMode.Disabled
+            }
             val signingManager = SigningManager(signatureVerificationMode, appConfig, apiKey)
 
             val cache = DeviceCache(prefs, apiKey)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
@@ -58,7 +58,7 @@ internal class IdentityManager(
         val cacheEditor = deviceCache.startEditing()
         deviceCache.cacheAppUserID(appUserIDToUse, cacheEditor)
         subscriberAttributesCache.cleanUpSubscriberAttributeCache(appUserIDToUse, cacheEditor)
-        invalidateCustomerInfoAndETagCacheIfNeeded(appUserIDToUse)
+        invalidateETagCacheIfNeeded(appUserIDToUse)
         cacheEditor.apply()
 
         enqueue {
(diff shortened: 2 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -113,9 +114,16 @@ internal class PurchasesFactory(
...
             }
 
-            val signatureVerificationMode = SignatureVerificationMode.fromEntitlementVerificationMode(
-                verificationMode,
-            )
+            val signatureVerificationMode = try {
+                SignatureVerificationMode.fromEntitlementVerificationMode(
+                    verificationMode,
+                )
+            } catch (e: IllegalStateException) {
```
tonidero, line 121: Added this since I wanted to be extra-safe if we're enabling this for more devs.

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -59,7 +58,7 @@ internal class IdentityManager(
         val cacheEditor = deviceCache.startEditing()
         deviceCache.cacheAppUserID(appUserIDToUse, cacheEditor)
         subscriberAttributesCache.cleanUpSubscriberAttributeCache(appUserIDToUse, cacheEditor)
-        invalidateCustomerInfoAndETagCacheIfNeeded(appUserIDToUse, cacheEditor)
+        invalidateETagCacheIfNeeded(appUserIDToUse)
```
tonidero, line 61: Note that I'm still invalidating the ETag cache if the customer info in cache isn't valid, this is because we don't want to reuse an existing response in this case, and we want to receive a full new response

purchases/src/test/java/com/revenuecat/purchases/identity/IdentityManagerTests.kt
```diff
@@ -522,16 +522,13 @@ class IdentityManagerTests {
             true
         )
         identityManager.configure(userId)
-        verify(exactly = 1) {
-            mockDeviceCache.clearCustomerInfoCache(userId, mockEditor)
-        }
```
ajpallares, line 527: Perhaps the same as in iOS should be done here (verify that `clearCustomerInfoCache` is not invoked)?

purchases/src/test/java/com/revenuecat/purchases/identity/IdentityManagerTests.kt
```diff
@@ -540,9 +537,6 @@ class IdentityManagerTests {
...
```
ajpallares, line 545: Same here as above

ajpallares (approved):
I'm not the best one to review this one but it looks good to me. 

Regarding checking in the tests the `clearCustomerInfoCache` is not called: the same point could be if there are any other clear cache calls that should not be executed. I do think making sure that `clearCustomerInfoCache` is not called makes sense because the previous implementation did call it and we don't want it to be executed anymore. But both options are valid, I think.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
JayShortway: Shouldn't this be `IllegalArgumentException`? I see that the constructor of `DefaultSignatureVerifier` calls `Base64.decode()`, which throws `IllegalArgumentException`. Not sure if that is the case you're guarding against here.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -113,9 +114,16 @@ internal class PurchasesFactory(
...
+                // If we're not able to create the signature verifier, we should disable signature verification
+                // instead of crashing
+                errorLog("Error creating signature verifier: ${e.message}. Disabling signature verification.")
+                SignatureVerificationMode.Disabled
```
JayShortway, line 125: 2 separate questions:
- What are possible causes for this failure? 
- What will the `VerificationResult` be if this failed? `NOT_REQUESTED` or `FAILED`?

JayShortway (approved):
Looks good! Just some questions.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Nope, the one I was actually trying to protect against is the constructor of `Ed25519`, in the `Tink` library, which we use for verification: https://github.com/tink-crypto/tink-java/blob/4408a238b73da603e908addc944f9c191549d91c/src/main/java/com/google/crypto/tink/subtle/Ed25519Verify.java#L55

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: > What are possible causes for this failure?

I'm honestly not entirely sure, since it comes from the Tink library we use. I was mostly protecting since I was making sure we didn't cause any unintended behavior.

> What will the VerificationResult be if this failed? NOT_REQUESTED or FAILED?

NOT_REQUESTED

PR #2052, opened 2025-01-13
Add `subscriptionsByProductIdentifier` to `CustomerInfo`

vegaro:
We had more information we were not exposing in the CustomerInfo JSON

This PR adds `CustomerInfo.subscriptionsByProductIdentifier` which has some useful information for devs that don't use the entitlements system

https://github.com/RevenueCat/purchases-ios/pull/4508

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/SubscriptionInfoResponse.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/SubscriptionInfoResponse.kt
@@ -11,26 +11,30 @@ import java.util.Date
 @Serializable
 @SuppressWarnings("LongParameterList")
 internal class SubscriptionInfoResponse(
-    @SerialName("purchase_date") @Serializable(with = ISO8601DateSerializer::class) val purchaseDate: Date,
-    @SerialName("original_purchase_date") @Serializable(with = ISO8601DateSerializer::class) val originalPurchaseDate:
-    Date?,
-    @SerialName("expires_date") @Serializable(with = ISO8601DateSerializer::class) val expiresDate: Date?,
-    @SerialName("store") val store: Store,
-    @SerialName("is_sandbox") val isSandbox: Boolean,
-    @SerialName("unsubscribe_detected_at") @Serializable(with = ISO8601DateSerializer::class) val unsubscribeDetectedAt:
-    Date?,
-    @SerialName(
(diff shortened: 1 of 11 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/utils/EntitlementInfoHelper.kt
```diff
@@ -0,0 +1,23 @@
+package com.revenuecat.purchases.utils
+
+import com.revenuecat.purchases.Store
+import java.util.Date
+
+internal class EntitlementInfoHelper private constructor() {
```
tonidero, line 6: I think you can make this an object
```suggestion
internal object EntitlementInfoHelper {
```
Then, you will need to remove the `companion object` below.

purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/SubscriptionInfoResponse.kt
```diff
@@ -0,0 +1,54 @@
...
+    Date?,
+    @SerialName("expires_date") @Serializable(with = ISO8601DateSerializer::class) val expiresDate: Date?,
+    @SerialName("store") val store: Store,
+    @SerialName("is_sandbox") val isSandbox: Boolean,
+    @SerialName("unsubscribe_detected_at") @Serializable(with = ISO8601DateSerializer::class) val unsubscribeDetectedAt:
+    Date?,
+    @SerialName(
+        "billing_issues_detected_at",
+    ) @Serializable(with = ISO8601DateSerializer::class) val billingIssuesDetectedAt:
+    Date?,
```
tonidero, line 25: Hmm indentation seems weird in some of these?

tonidero (approved):
Just some small things but LGTM!

purchases/src/main/kotlin/com/revenuecat/purchases/utils/EntitlementInfoHelper.kt
vegaro: oh right

purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/SubscriptionInfoResponse.kt
vegaro: right... detekt is doing this. I will see if I can fix it

PR #2053, opened 2025-01-13
[Paywalls V2] Fixes minimum spacing when distribution is `SPACE_BETWEEN`, `SPACE_AROUND` or `SPACE_EVENLY`

JayShortway:
## Description
As the title says. I was planning to do this fully with a custom `Arrangement`. However this is not possible as `Row`/`Column` don't expect a custom `Arrangement` to add any spacing before or after the content. Therefore measurement was off for `SPACE_AROUND` and `SPACE_EVENLY`. To fix this I'm using `padding` to add the required spacing before and after the content.


tonidero (approved):
Ohh right... this is annoying... but can't think of a better solution right now. Looks good!

PR #2054, opened 2025-01-14
[Paywalls V2] Add `edgeToEdge` badge trailing/leading style layout

tonidero:
### Description
This adds the remaining `edgeToEdge` style layouts for the badge, including leading and trailing

image

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
@@ -2,8 +2,8 @@ package com.revenuecat.purchases.paywalls.components.properties
 
 import androidx.annotation.IntRange
 import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.common.errorLog
 import dev.drewhamilton.poko.Poko
-import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
@@ -12,7 +12,7 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 
 @InternalRevenueCatAPI
-@Serializable
+@Serializable(with = CornerRadiusesSerializer::class)
 sealed interface CornerRadiuses {
 
     /**
@@ -104,24 +104,25 @@ sealed interface CornerRadiuses {
     }
 }
 
-@OptIn(InternalRevenueCatAPI::class, ExperimentalSerializationApi::class)
-internal object CornerRadiusesSerializer : KSerializer<CornerRadiuses?> {
-    override val descriptor: SerialDescriptor = CornerRadiuses.Dp.serializer().descriptor
+@OptIn(InternalRevenueCatAPI::class)
+internal object CornerRadiusesSerializer : KSerializer<CornerRadiuses> {
+    private val serializer = CornerRadiuses.Dp.serializer()
+    override val descriptor: SerialDescriptor = serializer.descriptor
 
-    override fun serialize(encoder: Encoder, value: CornerRadiuses?) {
-        when (value) {
-            is CornerRadiuses.Dp -> encoder.encodeSerializableValue(CornerRadiuses.Dp.serializer(), value)
-            is CornerRadiuses.Percentage ->
-                encoder.encodeSerializableValue(CornerRadiuses.Percentage.serializer(), value)
-            null -> encoder.encodeNull()
-        }
+    override fun serialize(encoder: Encoder, value: CornerRadiuses) {
+        // We don't need to serialize CornerRadiuses.
     }
 
     /**
      * For now, the backend will always return corner radiuses in Dp, so this will assume all serializations
      * are done with that serializer.
      */
-    override fun deserialize(decoder: Decoder): CornerRadiuses? {
-        return decoder.decodeNullableSerializableValue(CornerRadiuses.Dp.serializer())
+    override fun deserialize(decoder: Decoder): CornerRadiuses {
+        return try {
+            decoder.decodeSerializableValue(serializer)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+            errorLog("Failed to deserialize CornerRadiuses", e)
+            CornerRadiuses.Dp.zero
+        }
     }
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
@@ -2,7 +2,6 @@ package com.revenuecat.purchases.paywalls.components.properties
 
 import androidx.annotation.IntRange
 import com.revenuecat.purchases.InternalRevenueCatAPI
-import com.revenuecat.purchases.common.errorLog
 import dev.drewhamilton.poko.Poko
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
@@ -118,11 +117,6 @@ internal object CornerRadiusesSerializer : KSerializer<CornerRadiuses> {
      * are done with that serializer.
      */
     override fun deserialize(decoder: Decoder): CornerRadiuses {
-        return try {
-            decoder.decodeSerializableValue(serializer)
-        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
-            errorLog("Failed to deserialize CornerRadiuses", e)
-            CornerRadiuses.Dp.zero
-        }
+        return decoder.decodeSerializableValue(serializer)
     }
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
@@ -1,56 +1,127 @@
 package com.revenuecat.purchases.paywalls.components.properties
 
+import androidx.annotation.IntRange
 import com.revenuecat.purchases.InternalRevenueCatAPI
 import dev.drewhamilton.poko.Poko
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.KSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
 
-/**
- * Contains radius values for 4 corners, in dp.
- */
 @InternalRevenueCatAPI
-@Poko
 @Serializable
-class CornerRadiuses(
-    /**
-     * The top-leading, or top-start, corner radius, in dp.
-     */
-    @get:JvmSynthetic
-    @SerialName("top_leading")
-    val topLeading: Double,
-    /**
-     * The top-trailing, or top-end, corner radius, in dp.
-     */
-    @get:JvmSynthetic
-    @SerialName("top_trailing")
-    val topTrailing: Double,
-    /**
-     * The bottom-leading, or bottom-start, corner radius, in dp.
-     */
-    @get:JvmSynthetic
-    @SerialName("bottom_leading")
-    val bottomLeading: Double,
-    /**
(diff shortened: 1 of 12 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -320,6 +335,62 @@ private fun StackWithEdgeToEdgeBadge(
...
+                    topLeading = stackStateRectangleCorners.topLeading,
+                    topTrailing = 0.0,
+                    bottomLeading = 0.0,
+                    bottomTrailing = stackStateRectangleCorners.bottomTrailing,
+                )
+                else -> CornerRadiuses(all = 0.0)
+            }
+            Shape.Rectangle(corners = adjustedCorners)
+        }
+        Shape.Pill -> badgeStack.shape // If the stack is pill, we just default to the badge shape
```
JayShortway, line 381: I think we should clip the badge here, also going off @bab-s's comment on Slack [here](https://revenuecat.slack.com/archives/C07H5S8CFL3/p1736858263323829?thread_ts=1736849683.989489&cid=C07H5S8CFL3).

JayShortway (commented):
Nice one! Just one comment on clipping when the underlying stack is `Pill`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Good point! Just thinking out loud: maybe we can move the badge "inside" the stack, and then extend its background to the edge of the stack. That way it _looks_ clipped, but its contents will always be visible.

tonidero: Right... I guess this might even happen with non-pill shapes if the stack corner radius is bigger than half height of the badge... 🥴. So yeah I guess we need to separate the content from the background again in these and extend the background as you said... back to the drawing board 😅

JayShortway: Soo many _edge_ cases 🥁 🙁

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -320,6 +334,57 @@ private fun StackWithEdgeToEdgeBadge(
...
+private fun StackWithShortEdgeToEdgeBadge(
+    stackState: StackComponentState,
+    state: PaywallState.Loaded.Components,
+    badgeStack: StackComponentStyle,
+    alignment: TwoDimensionalAlignment,
+    clickHandler: suspend (PaywallAction) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val badgeRectangleCorners = (badgeStack.shape as? Shape.Rectangle)?.corners
+        ?: CornerRadiuses(all = 20.0) // For pill badge shape default to a "medium" rounded corner.
```
tonidero, line 348: I'm not sure about this TBH... But I'm not sure how a pill shaped badge would look like as a leading/trailing badge either, so this seemed like a compromise. Or maybe we should disallow this case?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -320,6 +334,57 @@ private fun StackWithEdgeToEdgeBadge(
...
+                topLeading = badgeRectangleCorners.topLeading,
+                topTrailing = 0.0,
+                bottomLeading = 0.0,
+                bottomTrailing = 0.0,
+            )
+            else -> CornerRadiuses(all = 0.0)
+        },
+    )
+    MainStackComponent(stackState, state, clickHandler, modifier) {
+        StackComponentView(
```
tonidero, line 379: Currently, the badge can be clipped by the corners of the main stack... I think that developers can fix this by adding a padding in the dashboard so it doesn't overlap. We could potentially add some padding manually ourselves, but I think it's a lot of extra logic, and probably less flexible.

This can be a bit of a problem with pill shaped main stacks since the corner is dynamic... But usually an approximate padding should work. 

We can always revisit if needed

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: If it's a pill-shaped badge, shouldn't it be a `50%` corner? We can do that with a Compose `Shape`. Maybe I'm overlooking something!

JayShortway: Hmm, but could we move the `badgeShape` (and `badgeRectangleCorners `) calculation closer to where it's actually used? Eventually it gets transformed to a Compose `Shape` anyway. 

Having a fixed corner radius is "risky", since it can be way too small (if the badge is giant), or way too big (if the badge is tiny).

tonidero: > Hmm, but could we move the badgeShape (and badgeRectangleCorners ) calculation closer to where it's actually used? Eventually it gets transformed to a Compose Shape anyway.

Not easily... What we pass to the badge's `StackComponentView` is our own `Shape`, not compose's Shape, and I don't like having a parameter in `StackComponentView` for something like this...

> Having a fixed corner radius is "risky", since it can be way too small (if the badge is giant), or way too big (if the badge is tiny).

That is indeed true... That's why I was asking in Slack if we should disable this case in the dashboard... But maybe I should just go ahead with the `CornerRadiuses` refactor to accept percentages so we fix this case as well.

tonidero: I ended up doing a biggish refactor: https://github.com/RevenueCat/purchases-android/pull/2054/commits/94c69aeea6efe314c79897d364af823a8dbe9de2... However, I kinda liked that it simplified some of the calculations. Wdyt?

tonidero:
I'm addressing the changes from our last sync up in follow up PRs, I think this one is already big enough, and it covers the cases it's meant for

JayShortway (commented):
Nice one! I like the idea of making `CornerRadiuses` sealed! Had some comments on the serialization part only.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
```diff
@@ -1,56 +1,127 @@
...
+    override val descriptor: SerialDescriptor = CornerRadiuses.Dp.serializer().descriptor
 
-    fun copy(
-        topLeading: Double = this.topLeading,
-        topTrailing: Double = this.topTrailing,
-        bottomLeading: Double = this.bottomLeading,
-        bottomTrailing: Double = this.bottomTrailing,
-    ): CornerRadiuses {
-        return CornerRadiuses(topLeading, topTrailing, bottomLeading, bottomTrailing)
+    override fun serialize(encoder: Encoder, value: CornerRadiuses?) {
```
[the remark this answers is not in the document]
tonidero: Right, I just implemented since it's mandatory for the interface but  I guess we could make it fail...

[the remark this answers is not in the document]
JayShortway: It doesn't need to return anything, so I usually just have an empty method with a comment explaining it's not implemented because it's not needed.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
```diff
@@ -1,56 +1,127 @@
...
-    @get:JvmSynthetic
-    @SerialName("top_leading")
-    val topLeading: Double,
-    /**
-     * The top-trailing, or top-end, corner radius, in dp.
-     */
-    @get:JvmSynthetic
-    @SerialName("top_trailing")
-    val topTrailing: Double,
+sealed interface CornerRadiuses {
```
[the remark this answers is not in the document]
tonidero: Had to go with the second one, since the first one wouldn't work since it needs a discrimantor to decide which serializer to use.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/CornerRadiuses.kt
```diff
@@ -1,56 +1,128 @@
...
+    }
+
+    /**
+     * For now, the backend will always return corner radiuses in Dp, so this will assume all serializations
+     * are done with that serializer.
+     */
+    override fun deserialize(decoder: Decoder): CornerRadiuses {
+        return try {
+            decoder.decodeSerializableValue(serializer)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
```
tonidero, line 123: I don't fully like this... But `ShapeTests` and `MaskShapeTests` were failing to deserialize for me when corner radiuses were provided as "null"...

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/properties/MaskShapeTests.kt
```diff
@@ -4,9 +4,9 @@ import com.revenuecat.purchases.common.OfferingParser
 import org.intellij.lang.annotations.Language
 import org.junit.Test
 import org.junit.runner.RunWith
-import org.junit.runners.Parameterized
+import org.robolectric.ParameterizedRobolectricTestRunner
 
-@RunWith(Parameterized::class)
+@RunWith(ParameterizedRobolectricTestRunner::class)
```
tonidero, line 9: Had to change the runner because I added the log when there is an error deserializing cornerradiuses and that was failing without robolectric.

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/properties/ShapeTests.kt
```diff
@@ -5,6 +5,7 @@ import org.intellij.lang.annotations.Language
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
+import org.robolectric.ParameterizedRobolectricTestRunner
```
JayShortway, line 8: This is unused I think? (Also shouldn't Detekt have caught this? 🤔)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/MaskShape.kt
```diff
@@ -13,7 +13,9 @@ sealed interface MaskShape {
     @Serializable
     @SerialName("rectangle")
     class Rectangle(
-        @get:JvmSynthetic val corners: CornerRadiuses? = null,
+        @get:JvmSynthetic
+        @Serializable
+        val corners: CornerRadiuses? = null,
```
JayShortway, line 18: I just realized we don't need `@Serializable` on the field at all if we're not setting a custom serializer haha. Same for `Shape.Rectangle.corners`.

JayShortway (approved):
LGTM! Thanks for making the changes! 🙏

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/properties/ShapeTests.kt
tonidero: Oh right! And I believe we might not be running detekt on the purchases module test source set 😬

JayShortway: Oh this is the test source set of course! I guess that's fine. We could decide to run a few rules on the tests, such as the no unused imports one haha.

tonidero: yeah I actually think we should run all of them TBH 😅 But we probably would need to do a ton of changes to fix it, so we've been delaying it... Still probably worth doing it even if we have to keep the current as "baseline" to avoid these moving forward.

PR #2055, opened 2025-01-14
[Paywalls V2] Implements `Convex` and `Concave` image masks

JayShortway:
## Description
As the title says. Implementation is the same as [the dashboard](https://github.com/RevenueCat/revenuecat-app/blob/main/src/concepts/projects/paywall-builder/components/previewer/components/image.tsx#L88-L115), except it uses a relative offset (0.2 x height) instead of a fixed one (27 px).  

image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Shape.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Shape.kt
@@ -11,6 +11,12 @@ import androidx.compose.ui.unit.dp
 import com.revenuecat.purchases.paywalls.components.properties.MaskShape
 import com.revenuecat.purchases.paywalls.components.properties.Shape as RcShape
 
+/**
+ * The scale factor to use on the Y axis when drawing a Concave or Convex shape. This will be multiplied by Y to
+ * determine the start and end point of the curve from the bottom of the shape.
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Shape.kt
```diff
@@ -11,6 +11,12 @@ import androidx.compose.ui.unit.dp
 import com.revenuecat.purchases.paywalls.components.properties.MaskShape
 import com.revenuecat.purchases.paywalls.components.properties.Shape as RcShape
 
+/**
+ * The scale factor to use on the Y axis when drawing a Concave or Convex shape. This will be multiplied by Y to
+ * determine the start and end point of the curve from the bottom of the shape.
+ */
+private const val SCALE_Y_OFFSET_CONCAVE_CONVEX = 0.2f
```
JayShortway, line 18: This is, as the Germans say, _Fingerspitzengefühl_.

JayShortway: Coincidentally, [web uses the same value](https://revenuecat.slack.com/archives/C07H5S8CFL3/p1736862736434169?thread_ts=1734967060.588079&cid=C07H5S8CFL3).

PR #2056, opened 2025-01-14
[Paywalls V2] Add new `ImageComponent` properties

tonidero:
### Description
Added the following properties to the `ImageComponent`
  - Padding
  - Margin
  - Border
  - Shadow

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
@@ -14,9 +14,14 @@ import androidx.annotation.DrawableRes
 import androidx.annotation.Px
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.RectangleShape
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -291,10 +340,30 @@ private fun previewImageComponentStyle(
     size: Size,
     fitMode: FitMode,
     overlay: ColorScheme? = null,
+    paddingValues: PaddingValues = PaddingValues(0.dp),
+    marginValues: PaddingValues = PaddingValues(0.dp),
+    border: Border? = Border(
+        width = 2.0,
+        color = ColorScheme(
+            light = ColorInfo.Hex(
+                androidx.compose.ui.graphics.Color.Cyan.toArgb(),
```
JayShortway, line 349: We have this aliased like so:
```suggestion
                ComposeColor.Cyan.toArgb(),
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -291,10 +340,30 @@ private fun previewImageComponentStyle(
...
+            ),
+        ),
+    ),
+    shadow: Shadow? = Shadow(
+        color = ColorScheme(ColorInfo.Hex(androidx.compose.ui.graphics.Color.Black.toArgb())),
```
JayShortway, line 354:
```suggestion
        color = ColorScheme(ColorInfo.Hex(ComposeColor.Black.toArgb())),
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -85,13 +96,27 @@ internal fun ImageComponentView(
...
+        }
+
         RemoteImage(
             urlString = imageState.imageUrls.webp.toString(),
             modifier = modifier
                 .size(imageState.size)
                 .applyIfNotNull(imageState.aspectRatio) { aspectRatio(it) }
-                .applyIfNotNull(overlay) { overlay(it, imageState.shape ?: RectangleShape) }
-                .applyIfNotNull(imageState.shape) { clip(it) },
+                .then(commonModifier),
```
JayShortway, line 119: I think we can just append this modifier chain directly here? There's little need to keep it as a separate `commonModifier` field, like there is in `StackComponentView`.

JayShortway (approved):
Thanks for taking care of this! Just some minor comments and suggestions.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
tonidero: Right, it would be just to remember the modifier... But then again, I don't think we're doing anything special there... So I think it makes sense to move it here

JayShortway (approved):

PR #2059, opened 2025-01-15
[Paywalls V2] Font sizes are integers now.

JayShortway:
## Description
As the title says.

Also aligned the mapping of font size names to integers with iOS, which means some previews will probably change.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TextComponent.kt
```diff
@@ -91,3 +101,41 @@ constructor(
...
+                "heading_xs" -> 16
+                "body_xl" -> 18
+                "body_l" -> 17
+                "body_m" -> 15
+                "body_s" -> 13
+                else -> throw SerializationException("Unknown font size name: $fontSizeString")
+            }
+        } else {
+            element.int
+        }
```
JayShortway, line 139: I wanted to do something like this: 

```kotlin
try {
  decoder.decodeInt()
catch (e: SerializationException) {
  FontSize.serializer().deserialize(decoder).value
}
```

But that kept failing with an `IndexOutOfBoundsException` when deserializing `TextComponent` as `PaywallComponent`. My assumption is that it didn't like taking 2 "decoding" attempts for a single field (in the `catch` branch). 

Since this is temporary code that will not be called for any public paywall (as font sizes are ints for those), I thought the current implementation is sufficient. Let me know if you disagree.

tonidero (approved):
Nice! LGTM!

PR #2061, opened 2025-01-16
Fix `OwnershipType` enum serialization

vegaro:
It looks like the backend is actually returning this in uppercase


purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
```diff
@@ -264,18 +264,18 @@ enum class OwnershipType {
     /**
      * The purchase was made directly by this user.
      */
-    @SerialName("purchased")
+    @SerialName("PURCHASED")
```
tonidero, line 267: Hmm I wonder, would it work without actually adding the `@SerialName` annotations?

tonidero (approved):
Just a possible cleanup comment but looks good!

JayShortway (approved):

PR #2064, opened 2025-01-17
[Paywalls V2] Adds support for fallback components

JayShortway:
## Description
As the title says. Adds a custom serializer that looks for a `fallback` key if the `type` is unknown. 

This is the Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4621.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallComponent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallComponent.kt
@@ -1,8 +1,53 @@
 package com.revenuecat.purchases.paywalls.components
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
+import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.builtins.serializer
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.buildClassSerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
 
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallComponent.kt
```diff
@@ -1,8 +1,53 @@
...
+    override fun serialize(encoder: Encoder, value: PaywallComponent) {
+        // Serialization is not implemented as it is not needed.
+    }
+
+    override fun deserialize(decoder: Decoder): PaywallComponent {
+        val jsonDecoder = decoder as? JsonDecoder
+            ?: throw SerializationException("Can only deserialize PaywallComponent from JSON, got: ${decoder::class}")
+        val json = jsonDecoder.decodeJsonElement().jsonObject
+        return when (val type = json["type"]?.jsonPrimitive?.content) {
+            "button" -> jsonDecoder.json.decodeFromString<ButtonComponent>(json.toString())
```
tonidero, line 39: Should we define the strings here in each individual component? We can have the `PaywallComponent` interface have a "key" property to be implemented by each component, which would decrease the chances of us forgetting to add new components here (even if not make it impossible). We could also flip this `when` and check for each component type "key" matches the one in the json, and if not return. That way we would have greater safety.

tonidero (approved):
I think my comment is not a blocker, in case you want to go ahead and do the change in a separate PR, so :shipit:

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallComponent.kt
JayShortway: I like it! Let me noodle on it for a bit 😄

tonidero: ohhh ok... Yeah I don't see many options either... I think this is ok then, we will just need to remember to update here with any new components.

JayShortway: Yea indeed. We should be adding a deserialization test anyway in that case, which will fail, which will then point us in the direction of the custom serializer.

PR #2067, opened 2025-01-17
Add support to manage non-Google purchases in Customer Center

vegaro:
If the active purchase is not from Google, we just display a button for Managing subscriptions and another one for contacting support. Although this should probably be configurable from the dashboard, this is the same we do in iOS atm.

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
@@ -5,9 +5,10 @@ import android.os.Bundle
 import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.Surface
 import androidx.compose.ui.Modifier
 import com.revenuecat.paywallstester.ui.theme.PaywallTesterAndroidTheme
 import com.revenuecat.purchases.Offering
@@ -22,8 +23,8 @@ class MainActivity : ComponentActivity(), PaywallResultHandler {
         super.onCreate(savedInstanceState)
         paywallActivityLauncher = PaywallActivityLauncher(this, this)
         setContent {
-            PaywallTesterAndroidTheme {
-                Surface(modifier = Modifier.fillMaxSize(), color = MaterialTheme.colorScheme.background) {
+            PaywallTesterAndroidTheme(dynamicColor = false) {
+                Box(modifier = Modifier.fillMaxSize().background(color = MaterialTheme.colorScheme.background)) {
                     PaywallTesterApp()
                 }
             }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.revenuecatui.customercenter
 
+import android.net.Uri
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
@@ -219,6 +220,7 @@ private class SubscriptionInformationProvider : PreviewParameterProvider<Purchas
                 ExpirationOrRenewal.Date.DateString("June 1st, 2024"),
             ),
             store = Store.PLAY_STORE,
+            managementURL = Uri.parse("https://play.google.com/store/account/subscriptions"),
         ),
         PurchaseInformation(
             title = "Basic",
@@ -232,6 +234,7 @@ private class SubscriptionInformationProvider : PreviewParameterProvider<Purchas
                 ExpirationOrRenewal.Date.DateString("June 1st, 2024"),
             ),
             store = Store.PLAY_STORE,
+            managementURL = Uri.parse("https://play.google.com/store/account/subscriptions"),
         ),
         PurchaseInformation(
             title = "Basic",
@@ -245,6 +248,7 @@ private class SubscriptionInformationProvider : PreviewParameterProvider<Purchas
                 ExpirationOrRenewal.Date.DateString("June 1st, 2024"),
             ),
             store = Store.PLAY_STORE,
+            managementURL = Uri.parse("https://play.google.com/store/account/subscriptions"),
         ),
     )
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
@@ -206,6 +206,8 @@ private val PaddingHorizontal = 8.dp
 private val PaddingVertical = 8.dp
 private const val SizeIconDp = 22
 
+private const val MANAGEMENT_URL = "https://play.google.com/store/account/subscriptions"
+
 private class SubscriptionInformationProvider : PreviewParameterProvider<PurchaseInformation> {
     override val values: Sequence<PurchaseInformation> = sequenceOf(
         PurchaseInformation(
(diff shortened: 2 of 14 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
```diff
@@ -220,6 +221,7 @@ private class SubscriptionInformationProvider : PreviewParameterProvider<Purchas
                 ExpirationOrRenewal.Date.DateString("June 1st, 2024"),
             ),
             store = Store.PLAY_STORE,
+            managementURL = Uri.parse("https://play.google.com/store/account/subscriptions"),
```
tonidero, line 224: Should we extract the URL to a constant? (If we don't have it already)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -282,6 +291,16 @@ internal class CustomerCenterViewModelImpl(
         context.startActivity(Intent.createChooser(intent, "Contact Support"))
     }
 
+    @SuppressWarnings("ForbiddenComment")
+    override fun openURL(context: Context, url: Uri) {
+        // TODO: Handle In-App Browser
+        try {
+            context.startActivity(Intent(Intent.ACTION_VIEW, url))
+        } catch (e: ActivityNotFoundException) {
+            Logger.e("Error opening URL", e)
```
tonidero, line 300: Hmm I wonder if we should display a Toast or something like that in this case, so the customer knows something is happening... We already do that in the paywall footer: https://github.com/RevenueCat/purchases-android/blob/main/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Footer.kt#L264

We might want to extract that logic and copy it here.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -191,29 +214,83 @@ private fun NoActiveSubscriptionsViewPreview() {
...
+                colors = ButtonDefaults.buttonColors(containerColor = MaterialTheme.colorScheme.primary),
+            ) {
+                Text(
+                    text = localization.commonLocalizedString(
+                        CustomerCenterConfigData.Localization.CommonLocalizedString.MANAGE_SUBSCRIPTION,
+                    ),
+                    color = if (useOutlinedButton) {
+                        MaterialTheme.colorScheme.primary
+                    } else {
+                        MaterialTheme.colorScheme.onPrimary
```
tonidero, line 268: Ohh I hadn't used the `on...` colors before... I think that should be ok!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -191,29 +214,83 @@ private fun NoActiveSubscriptionsViewPreview() {
...
+                    },
+                )
+            }
+        }
+        support.email?.let {
+            Button(
```
tonidero, line 274: I believe much of this is repeated... Should we extract it to another composable?

vegaro: oh I can actually reuse the other button. I think when I started the PR I couldn't reuse, but now I can

vegaro: oh right, actually I need another one because these have hardcoded values. So I will extract another `Composable`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: updated!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
```diff
@@ -22,8 +23,8 @@ class MainActivity : ComponentActivity(), PaywallResultHandler {
         super.onCreate(savedInstanceState)
         paywallActivityLauncher = PaywallActivityLauncher(this, this)
         setContent {
-            PaywallTesterAndroidTheme {
-                Surface(modifier = Modifier.fillMaxSize(), color = MaterialTheme.colorScheme.background) {
+            PaywallTesterAndroidTheme(dynamicColor = false) {
+                Box(modifier = Modifier.fillMaxSize().background(color = MaterialTheme.colorScheme.background)) {
```
vegaro, line 27: changed this because I was facing some issues and thought this could be the reason, but in any case, I think it's better to use a `Box` here

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
```diff
@@ -22,8 +23,8 @@ class MainActivity : ComponentActivity(), PaywallResultHandler {
         super.onCreate(savedInstanceState)
         paywallActivityLauncher = PaywallActivityLauncher(this, this)
         setContent {
-            PaywallTesterAndroidTheme {
-                Surface(modifier = Modifier.fillMaxSize(), color = MaterialTheme.colorScheme.background) {
+            PaywallTesterAndroidTheme(dynamicColor = false) {
```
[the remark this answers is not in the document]
tonidero: Hmm I'm thinking it might be good to keep paywall tester using the defaults for new apps, so we're more easily able to test any behavior change... I believe colors should normally only change from dark/light mode right? Do you know in what other situations it changed?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -187,52 +208,108 @@ private fun NoActiveSubscriptionsViewPreview() {
...
+                textAlign = TextAlign.Start,
+                style = MaterialTheme.typography.bodyLarge,
+            )
+        },
+    )
+}
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+@Composable
+private fun CustomerCenterButton(
```
tonidero, line 300: I wonder if this composable should be extracted to a separate file so it can be reused in different screens... Fine to do that in future PRs though.

tonidero (approved):
LGTM!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/MainActivity.kt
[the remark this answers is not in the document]
vegaro: With this, it created a theme mimicking the background of the device, ignoring the colors I kept setting in the theme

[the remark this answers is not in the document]
tonidero: Right, so for most people they will have this enabled right? So I think it might make more sense to keep it enabled and fix the issues where they appear... Otherwise we might miss some background not working as expected on developers apps I think?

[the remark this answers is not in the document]
JayShortway: Just to clarify: [dynamic color](https://m3.material.io/styles/color/dynamic/choosing-a-source) takes the theme colors from the device wallpaper. This has nothing to do with the background of the app, nor with dark and light mode. 

>it might be good to keep paywall tester using the defaults for new apps  
  
I agree with this in general! However, I would say it's very difficult for a tester app to look different depending on the wallpaper of the device it's running on. Makes all testing non-deterministic. 

>fix the issues where they appear  

This is not really possible, as the dynamic color feature can make it seem like there is an issue when there is none. As an example, we had a wallpaper that caused the `surface` and `background` colors to be the same in the theme. This made us think that the `surface` color was not applied properly, as it was "invisible". We kept messing with the theme and changing the `surface` color, and printing its value in various places, to no avail. Disabling the dynamic color feature avoids this red herring in the future. 

If this happens to someone's app, that's okay. It's their theme and their choice. But for us, we need to know whether we're applying the theme colors properly. I'd definitely vote for turning it off.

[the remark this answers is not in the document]
tonidero: Ok, that makes sense, let's go ahead with this then 👍

PR #2068, opened 2025-01-17
[Paywalls V2] Parses `UiConfig`

JayShortway:
## Description
Adds `UiConfig` to `Offering`. This is the Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4628. The `UiConfig` is not actually used yet. That will come in follow-up PRs. 

Lots of file and line changes, but not terribly complex.


purchases/src/test/java/com/revenuecat/purchases/UiConfigTests.kt
```diff
@@ -0,0 +1,177 @@
...
+                    "web": {
+                      "type": "google_fonts",
+                      "value": "Gothic"
+                    }
+                  }
+                }
+              },
+              "localizations": {
+                "en_US": {
+                  "monthly": "monthly"
```
tonidero, line 87: I believe it should work... But if we receive unknown localization keys here, it should continue parsing only the known ones correct?

purchases/src/test/java/com/revenuecat/purchases/UiConfigTests.kt
```diff
@@ -0,0 +1,177 @@
...
+                    }
+                  }
+                },
+                "fonts": {
+                  "primary": {
+                    "ios": {
+                      "type": "name",
+                      "value": "SF Pro"
+                    },
+                    "android": {
```
tonidero, line 74: Should we have a "secondary" font that uses `google_fonts` for android? (I see we have one for web, but that would be ignored by the deserializer)

tonidero (approved):
LGTM!

purchases/src/test/java/com/revenuecat/purchases/UiConfigTests.kt
JayShortway: Good idea! https://github.com/RevenueCat/purchases-android/pull/2068/commits/eed558c7df23f1befeb44945a9b86a01cc97499f

purchases/src/test/java/com/revenuecat/purchases/UiConfigTests.kt
JayShortway: Actually this didn't work, good call! I have "handling unknown enums" on my list, but decided to do this one now as it is slightly different because it's "handling unknown enums in a map (of maps)". https://github.com/RevenueCat/purchases-android/pull/2068/commits/e802ad67c389a2c727a2911932b020ab25c58135

.circleci/config.yml
```diff
@@ -180,7 +180,7 @@ jobs:
           command: ./gradlew :test-apps:testpurchasesuiandroidcompatibility:assembleDebug
       - run:
           name: Run Tests
-          command: ./gradlew lint test
+          command: java --version && echo $JAVA_HOME && ./gradlew lint test
```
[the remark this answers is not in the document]
JayShortway: Yea, I'm debugging that failing test. I can't get it to fail locally so wanted to double check the JDK on CI. Even with the same JDK and Gradle command it doesn't fail on my machine. So I'll be debugging some more. 😅 Will revert everything before merging.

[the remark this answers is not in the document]
JayShortway: The fix was to enable the `paywallComponents` build flag on CI. The only thing that's being controlled by that is parsing the `PaywallComponentsData` in the offerings response, which is exactly what that test was verifying. 😅 https://github.com/RevenueCat/purchases-android/pull/2068/commits/0894ed3da41226ae9754aa0732dcf6f11ce8e20e 

I reverted all the debug commits.

tonidero (approved):
Ahmm ka

PR #2069, opened 2025-01-20
[Paywalls V2] Badge: Handle main stack border width correctly in overlay and nested badge styles

tonidero:
### Description
- In Overlay style badges, the badge will now be placed in the middle of the border if any.
- In Nested style badges, we fix an issue where the border width wasn't added as padding, so the border was drawn on top of the stack content. This also means the badge is now placed inside the border.

image

image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -25,6 +25,7 @@ import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.layout.Placeable
 import androidx.compose.ui.layout.SubcomposeLayout
 import androidx.compose.ui.layout.layout
+import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.tooling.preview.PreviewParameterProvider
@@ -169,7 +170,8 @@ private fun StackWithOverlaidBadge(
 ) {
     Box(modifier = modifier) {
         MainStackComponent(stackState, state, clickHandler)
-        OverlaidBadge(badgeStack, state, alignment)
+        val mainStackBorderWidth = with(LocalDensity.current) { stackState.border?.width?.dp?.toPx() }
+        OverlaidBadge(badgeStack, state, alignment, mainStackBorderWidth)
     }
 }
 
@@ -405,6 +407,7 @@ private fun BoxScope.OverlaidBadge(
     badgeStack: StackComponentStyle,
     state: PaywallState.Loaded.Components,
     alignment: TwoDimensionalAlignment,
+    mainStackBorderWidth: Float?,
     modifier: Modifier = Modifier,
 ) {
     StackComponentView(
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -499,7 +502,10 @@ private fun MainStackComponent(
 
     val innerShapeModifier = remember(stackState, borderStyle) {
         Modifier
-            .applyIfNotNull(borderStyle) { border(it, composeShape) }
+            .applyIfNotNull(borderStyle) {
+                border(it, composeShape)
+                    .padding(it.width)
```
tonidero, line 507: This modifies the badge content as well, but I think this is the expected behavior, as in, the border shouldn't be drawn on top of the content, but just be considered extra padding for the content.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -170,7 +171,8 @@ private fun StackWithOverlaidBadge(
 ) {
     Box(modifier = modifier) {
         MainStackComponent(stackState, state, clickHandler)
-        OverlaidBadge(badgeStack, state, alignment)
+        val mainStackBorderWidth = with(LocalDensity.current) { stackState.border?.width?.dp?.toPx() }
```
JayShortway, line 174: Suggestion, I like adding the `Px` (or `Dp`) suffix to variables if the unit is not known from the type. So in this case that would be `mainStackBorderWidthPx`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Taking a nested badge in the upper left corner as an example, how would this work if the upper left corner radius of the badge is smaller than the upper left corner radius of the main stack? Would the border / main stack clip the upper left corner of the badge, or would the badge be "moved" more inside the main stack by additional margin? I'd say that the former is desired.

JayShortway (approved):
LGTM! Just a suggestion and a question.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: You mean something like this:
image
It will get clipped by the main stack border/corner as you can see, but just the inside of the border.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2069/commits/23818f9c9f24a19d649ea2eba0cdeb5b5159a2ec

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Yea that's indeed what I mean, and that's also what I think would be the desired result. Nice! Thanks for confirming!

PR #2070, opened 2025-01-20
[Paywalls V2] Make `edgeToEdge` top/bottom badge extend to entire stack background

tonidero:
### Description
This changes the `edgeToEdge` top and bottom badge extend to the whole parent stack

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -735,7 +735,9 @@ private fun StackComponentView_Preview_EdgeToEdge_Badge(
     Box(
         modifier = Modifier.padding(all = 32.dp),
     ) {
-        val badgeShape = Shape.Pill
+        val badgeShape = Shape.Rectangle(
+            corners = CornerRadiuses.Dp(all = 20.0),
+        )
         StackComponentView(
             style = StackComponentStyle(
                 children = previewChildren(),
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -244,36 +232,64 @@ private fun StackWithLongEdgeToEdgeBadge(
...
                 is CornerRadiuses.Percentage -> {
-                    if (topBadge) {
+                    (badgeStack.shape.toShape() as? RoundedCornerShape)?.let { composeShape ->
                         RoundedCornerShape(
-                            topStartPercent = badgeCornerRadiuses.topLeading,
-                            topEndPercent = badgeCornerRadiuses.topTrailing,
+                            topStart = composeShape.topStart.makeAbsolute(stackPlaceable, LocalDensity.current),
+                            topEnd = composeShape.topEnd.makeAbsolute(stackPlaceable, LocalDensity.current),
+                            bottomEnd = composeShape.bottomEnd.makeAbsolute(stackPlaceable, LocalDensity.current),
+                            bottomStart = composeShape.bottomStart.makeAbsolute(stackPlaceable, LocalDensity.current),
```
JayShortway, line 240: Just checking: isn't this now copying all corner radiuses from the main stack? Shouldn't it only copy the bottom ones for a top badge, and the top ones for a bottom badge? As the badge corners are configurable.

tonidero: Right, this is mostly when the badge shape is pill. In which case it's not configurable I think? And in that case, it looks like:
image

JayShortway: You're right! I was reading the `as? RoundedCornerShape` part and thought this was for all shapes, but I should have read the `is CornerRadiuses.Percentage` part. 😇 Ignore me!

JayShortway (approved):

PR #2071, opened 2025-01-20
[Paywalls V2] Add `IconComponent`

tonidero:
### Description
This adds a new `IconComponent`

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/paywalls/components/IconComponentTests.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/paywalls/components/IconComponentTests.kt
@@ -1,0 +1,388 @@
+package com.revenuecat.purchases.paywalls.components
+
+import com.revenuecat.purchases.common.OfferingParser
+import com.revenuecat.purchases.paywalls.components.properties.Border
+import com.revenuecat.purchases.paywalls.components.properties.ColorInfo
+import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
+import com.revenuecat.purchases.paywalls.components.properties.CornerRadiuses
+import com.revenuecat.purchases.paywalls.components.properties.MaskShape
+import com.revenuecat.purchases.paywalls.components.properties.Padding
+import com.revenuecat.purchases.paywalls.components.properties.Shadow
+import com.revenuecat.purchases.paywalls.components.properties.Size
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import org.intellij.lang.annotations.Language
+import org.junit.Test
+import org.junit.experimental.runners.Enclosed
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+@RunWith(Enclosed::class)
+internal class IconComponentTests {
+
+    @RunWith(Parameterized::class)
+    class DeserializeIconComponentTests(@Suppress("UNUSED_PARAMETER") name: String, private val args: Args) {
+
+        class Args(
+            @Language("json")
+            val json: String,
+            val expected: IconComponent,
+        )
+
+        companion object {
+
+            @Suppress("LongMethod")
+            @JvmStatic
+            @Parameterized.Parameters(name = "{0}")
+            fun parameters(): Collection<*> = listOf(
+                arrayOf(
+                    "all values present",
+                    Args(
+                        json = """
+                        {
+                          "base_url": "https://assets.pawwalls.com",
+                          "icon_name": "Test icon name",
+                          "formats": {
+                            "webp": "test.webp"
+                          },
+                          "color": {
+                            "light": {
+                              "type": "alias",
+                              "value": "primary"
+                            }
+                          },
+                          "padding": {
+                            "bottom": 16,
+                            "leading": 14,
+                            "top": 12,
+                            "trailing": 10
+                          },
+                          "margin": {
+                            "bottom": 10,
+                            "leading": 12,
+                            "top": 14,
+                            "trailing": 16
+                          },
+                          "size": {
+                            "height": {
+                              "type": "fixed",
+                              "value": 200
+                            },
+                            "width": {
+                              "type": "fixed",
+                              "value": 200
+                            }
+                          },
+                          "icon_background": {
+                            "color": {
+                              "light": {
+                                "type": "alias",
+                                "value": "secondary"
+                              }
+                            },
+                            "shape": {
+                              "corners": {
+                                "top_leading": 3,
+                                "top_trailing": 4,
+                                "bottom_leading": 1,
+                                "bottom_trailing": 2
+                              },
+                              "type": "rectangle"
+                            },
+                            "shadow": {
+                              "color": {
+                                "light": {
+                                  "type": "alias",
+                                  "value": "tertiary"
+                                }
+                              },
+                              "radius": 20.1,
+                              "x": 23.6,
+                              "y": 45.2
+                            },
+                            "border": {
+                              "color": {
+                                "light": {
+                                  "type": "alias",
+                                  "value": "secondary"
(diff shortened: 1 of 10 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
```diff
@@ -0,0 +1,130 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.iconcomponent
```
tonidero, line 1: Note that I named the package `iconcomponent`. This is because we're ignoring folders with the name `icon` in gitignore... I could try to find the reason for it, but this was easier for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
```diff
@@ -0,0 +1,209 @@
...
+            .clip(composeShape)
+            .applyIfNotNull(borderStyle) { border(it, composeShape) }
+            .padding(iconState.padding),
+        previewImageLoader = previewImageLoader,
+    )
+}
+
+@Preview
+@Composable
+private fun IconComponentView_Preview() {
```
tonidero, line 98: Still WIP, for some reason it wasn't working and AS didn't let me see the stack trace in the previewer.

JayShortway: Just in case: you can run the preview on a device/emulator using the "run" icon in the gutter. This should allow you to debug much easier, e.g. seeing stack traces and print statements. 
image

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
JayShortway: Haha that is very good to know!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
```diff
@@ -0,0 +1,224 @@
...
+
+    val borderStyle = iconState.border?.let { rememberBorderStyle(border = it) }
+    val shadowStyle = iconState.shadow?.let { rememberShadowStyle(shadow = it) }
+    val composeShape by remember(iconState.shape) { derivedStateOf { iconState.shape ?: RectangleShape } }
+    val backgroundColor = iconState.backgroundColorScheme?.let { rememberColorStyle(it) }
+    val tintColor = iconState.tintColor?.let { rememberColorStyle(it) }
+    val colorFilter by remember(tintColor) {
+        derivedStateOf {
+            // TODO Support gradient tints
+            (tintColor as? ColorStyle.Solid)?.let { ColorFilter.tint(it.color) }
```
JayShortway, line 87: Will this be a follow-up PR? Maybe we should just use the existing `overlay` (or `underlay`) modifier, which already supports gradients. Because I'm not sure if `ColorFilter` supports gradients. (I vaguely remember looking into that a while ago.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
```diff
@@ -0,0 +1,224 @@
...
+    overrides = null,
+)
+
+private fun previewEmptyState(): PaywallState.Loaded.Components {
+    val data = PaywallComponentsData(
+        templateName = "template",
+        assetBaseURL = URL("https://assets.pawwalls.com"),
+        componentsConfig = ComponentsConfig(
+            base = PaywallComponentsConfig(
+                // This would normally contain at least one ImageComponent, but that's not needed for previews.
```
JayShortway, line 180: 😋 
```suggestion
                // This would normally contain at least one IconComponent, but that's not needed for previews.
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -49,6 +50,7 @@ internal fun LocalImage(
     contentDescription: String? = null,
     transformation: Transformation? = null,
     alpha: Float = 1f,
+    colorFilter: ColorFilter? = null,
```
JayShortway, line 53: Just leaving a comment as this is related to the previous comment on the gradient overlay.

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/IconComponentTests.kt
```diff
@@ -0,0 +1,388 @@
...
+        }
+
+        @Test
+        fun `Should properly deserialize IconComponent as PaywallComponent`() {
+            // Arrange, Act
+            val actual = OfferingParser.json.decodeFromString<PaywallComponent>(args.json)
+
+            // Assert
+            assert(actual == args.expected)
+        }
```
JayShortway, line 211: Just a heads-up: the custom `PaywallComponent` serializer probably needs to be updated for this.

JayShortway (approved):
Good stuff! Just wondering if we can implement the tint using the existing `overlay`/`underlay` modifiers.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
tonidero: Yeah `ColorFilter` doesn't support gradients. I did try to check with `overlay`, but it was tinting the whole image, including the transparent background. I played around with the blendmodes but wasn't able to make it work, so I went with the solid color for now... But yeah, I would like to keep looking into this.

JayShortway: Hmm yea, we need to somehow clip the gradient `overlay`. (`underlay` probably would tint everything _except_ the icon.) Wild idea is to use the icon's SVG path and turn that into a clip path with [`Path.addSvg()`](https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/Path#(androidx.compose.ui.graphics.Path).addSvg(kotlin.String)).

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/IconComponentTests.kt
tonidero: Fixed after https://github.com/RevenueCat/purchases-android/pull/2071/commits/a0309537c71ef55db9ca441fe6e57b6450776d57

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
tonidero: We talked over Slack, but will leave this as is for now, we won't support gradients on icons for now.

PR #2072, opened 2025-01-22
Some Material 3 updates to the `ManageSubscriptionsView`

JayShortway:
## Description
As the title says. These are just some suggestions. Feel free to ignore.

- Uses `TextButton`s instead of primary `Button`s, and places them on a `Surface`. (The latter is more of an artistic choice haha.)
- Uses the `titleLarge` style for the screen title, following the [guidelines](https://m3.material.io/styles/typography/applying-type#e9e0cea3-10cb-405d-98a9-cf6a90758967). 
- Updates spacing. 
- Removes the shadow from the top card, as elevation is not often shown with shadows in Material 3. 

| Before | After |
|-------|------|
|  image  | image |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
@@ -4,6 +4,8 @@ import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.calculateEndPadding
+import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.heightIn
@@ -19,6 +21,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.RectangleShape
+import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -220,10 +213,13 @@ private fun ManageSubscriptionButton(
         .fillMaxWidth()
         .padding(vertical = 4.dp)
 
-    val buttonContent: @Composable () -> Unit = {
+    val buttonContent: @Composable (Modifier) -> Unit = { modifier ->
         Text(
             text = path.title,
-            color = if (useOutlinedButton) MaterialTheme.colorScheme.primary else MaterialTheme.colorScheme.onPrimary,
```
vegaro, line 226: so `useOutlinedButton` would not be used anymore. I was using that for the screen of users with no active subscriptions.

Screenshot 2025-01-22 at 14 22 32

I thought the standard would be this https://m3.material.io/foundations/content-design/global-writing/word-choice#715006c7-2637-4734-a6e6-d2dc261af528

hiddevdploeg:
It feels much more like an Android, great job! The only thing I’m wondering about is whether the title shouldn’t be slightly heavier in font weight, or if that’s just my preference.

hiddevdploeg:
CleanShot 2025-01-22 at 14 49 21@2x
Also this alignment is off by just a few points

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
JayShortway: It is still being used, but the text color is no longer dependent on it. This actually made me realize an improvement, which is to handle the text color using `outlinedButtonColors()` / `textButtonColors()`. See here: https://github.com/RevenueCat/purchases-android/pull/2072/commits/aa641cb1d9a960a8018e7306ad1dc12cc7610ae9. The outlined button colors should be unchanged.

JayShortway:
@hiddevdploeg 

Regarding font weight: we're now taking the font weight of the `TextStyle`, which means that users can override it by changing the `titleLarge` style in their app. They've probably already done that, so this ensures that the Customer Center title looks the same as other titles in their app. If we were to manually set the font weight, it's a fixed thing. 

Regarding the alignment: that's a great catch! Should be fixed here: https://github.com/RevenueCat/purchases-android/pull/2072/commits/0dc98c9d9344cb62a42bb3751089032fbe78de98. I updated the screenshot in the PR description.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
vegaro: right right, it's used in line 228 😮‍💨 

perfect

vegaro (approved):

PR #2075, opened 2025-01-22
feat: Add basic customer center events

facumenzella:
### Motivation
This PR adds basic customer center events definition:
- `customer_center_impression`
- `customer_center_survey_option_chosen`

### Description
Inspired in iOS [PR](https://github.com/RevenueCat/purchases-ios/pull/4392/files), and using my really really basic (almost negative) Kotlin expertise.

Let me know what you think 🙏


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -0,0 +1,45 @@
...
+import kotlinx.serialization.Serializable
+import java.util.Date
+import java.util.UUID
+
+/**
+ * Type representing a customer center event and associated data. Meant for RevenueCatUI use.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@Serializable
+data class CustomerCenterEvent(
```
tonidero, line 15: This is good but we've recently decided to start moving away from data classes since they introduced methods that may introduce breaking changes down the line if it changes. Instead, we've been using the `Poko` library to autogenerate some of the methods that might be useful. In this case, the changes are propose are:
```suggestion
@Poko
@Serializable 
class CustomerCenterEvent(
```
It would be the same for the other inner classes.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEventType.kt
```diff
@@ -0,0 +1,23 @@
+package com.revenuecat.purchases.customercenter.events
+
+/**
+ * Types of events that can be tracked in the Customer Center. Meant for RevenueCatUI use.
+ */
+enum class CustomerCenterEventType(val value: String) {
```
tonidero, line 6: Should this also be annotated with the `@ExperimentalPreviewRevenueCatPurchasesAPI` annotation?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEventType.kt
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * The customer center was closed by the user.
+     */
+    SURVEY_OPTION_CHOSEN("customer_center_survey_option_chosen"), // Fixed typo
+}
+
+/**
+ * Display mode for the Customer Center. Meant for RevenueCatUI use.
+ */
+enum class CustomerCenterDisplayMode {
```
tonidero, line 21: Hmm a couple questions here... Are we planning to support other display modes in the short/medium term? If not, maybe we can avoid adding this for now?

In case we do, I wonder if this would be better in a separate file? Seems weird that it's within the event type when it shouldn't really have anything to do with events. Also, same about the `@ExperimentalPreviewRevenueCatPurchasesAPI` annotation

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -0,0 +1,45 @@
...
+        @Serializable(with = UUIDSerializer::class)
+        val id: UUID,
+        @Serializable(with = DateSerializer::class)
+        val date: Date,
+    )
+
+    @Serializable
+    data class Data(
+        @Serializable(with = UUIDSerializer::class)
+        val id: UUID?,
```
tonidero, line 31: Hmm what does this id represent/how does it differ from the one in the `CreationData`? We might want to give it a different name if needed... And depending on the answer to that, is it ok for it to be nullable?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEventType.kt
tonidero, line 6: Hmm also, does this need to be `Serializable`? Maybe we don't need the `value` property and just add `@SerialName` to each enum value so it can be serialized correctly.

tonidero (commented):
Looking good! Left some comments. It might also be good to add some test to test the serialization of the events works correctly.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEventType.kt
facumenzella: We don't. As far as I understand, there's no other display mode in Android. In iOS we do have two. I thought it would be nice to have them listed anyway, just for consistency between platforms; but I can remove it 👌

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
facumenzella: Actually, I am not sure. Just double-checked, and it does not make any sense. Removed it 👍

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -0,0 +1,61 @@
...
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@Poko
+@Serializable
+class CustomerCenterEvent(
+    val creationData: CreationData,
+    val eventData: Data,
+) {
+
+    companion object {
+        val json = Json.Default
```
tonidero, line 25: I think we can keep this internal? Also make it so it warns us if we try to use it anywhere other than for tests?
```suggestion
        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
        internal val json = Json.Default
```

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        fun fromString(string: String): CustomerCenterEvent {
```
tonidero, line 27: This also could be `internal` I think? I'm not sure if it's actually used though...

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -0,0 +1,61 @@
...
+        val timestamp: Date,
+        val darkMode: Boolean,
+        val locale: String,
+        val isSandbox: Boolean,
+        val version: Int = 1,
+        val revisionId: Int = 1,
+        val displayMode: CustomerCenterDisplayMode = CustomerCenterDisplayMode.FULL_SCREEN,
+    )
+
+    override fun toString(): String {
```
tonidero, line 58: Hmm do we need to override this?

tonidero (commented):
LGTM! Just some questions. And thanks for adding the tests!

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
JayShortway, line 25: I think our `Json` instance should be defined in a single place to avoid confusion. Right now we also have one in `OfferingParser`, but it is configured differently. (It sets `ignoreUnknownKeys` to true.) This means the SDK behaves differently depending on which instance of `Json` is used to deserialize. 

I propose to define this once in some central place, and remove the others. Maybe something like: 

```kotlin
internal object JsonParser {
    @get:JvmSynthetic
    val json: Json = Json {
        ignoreUnknownKeys = true
    }
}
```

I can imagine that removing the `OfferingParser.json` one will have some ripple effects, so it's probably better to do that in a separate PR.

JayShortway (commented):
Nice one! Just 1 additional comment.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
facumenzella: I inspired on the one defined in `PaywallStoredEvent`

```kotlin
    companion object {
        val json = Json.Default
        fun fromString(string: String): PaywallStoredEvent {
            return json.decodeFromString(string)
        }
    }
```

Should we remove that one as well? Probably better to do it in a new PR right?

JayShortway: Oh we have three? Yes I'd consolidate them all into one, and indeed best to do it in a separate PR. I expect tiny changes in lots of files 😬

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    @Poko
+    @Serializable
+    @SuppressWarnings("LongParameterList")
+    class Data(
+        val type: CustomerCenterEventType,
+        @Serializable(with = DateSerializer::class)
+        val timestamp: Date,
+        val darkMode: Boolean,
```
tonidero, line 47: Just checking but does it need to be snake_case or camelCase? Currently this will serialize it as `darkMode`. If we want to make it snake_case, we could do `@SerialName("dark_mode")`

tonidero (approved):
LGTM! Just a question

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEventType.kt
facumenzella: I'll keep it for now, but we can come back to this once the whole event structure is set up

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
facumenzella: Just double-checked, we're good

PR #2076, opened 2025-01-22
[Paywalls V2] Implements color aliases for `StackComponent`

JayShortway:
## Description
This is similar to this iOS PR: https://github.com/RevenueCat/purchases-ios/pull/4650. This is not as bad as it looks. Hear me out. 😇 

Some small changes with lots of ripple effects:
- `ColorInfo.Alias`'s value is a `ColorAlias` now, allowing us to get color aliases from the `UiConfig` map with type safety. 
- Adds `ColorStyles`, which is basically to `ColorStyle` what `ColorScheme` is to `ColorInfo`. We need an object containing actual colors for both light and dark, without aliases. 
- Uses `ColorStyles` instead of `ColorScheme` in `StackComponentStyle` and `PresentedStackPartial`. 

A lot of the changed lines are tests:
- `PresentedStackPartialTests`
- `ColorStyleTests`

Follow-up PRs will change all usages of `ColorScheme`s in the other `*ComponentStyle`s to `ColorStyles`.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
@@ -2,7 +2,6 @@ package com.revenuecat.purchases.ui.revenuecatui.components
 
 import com.revenuecat.purchases.ColorAlias
 import com.revenuecat.purchases.paywalls.components.PartialStackComponent
-import com.revenuecat.purchases.paywalls.components.PartialTextComponent
 import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyles
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.toColorStyles
@@ -21,9 +20,9 @@ internal class PresentedStackPartial(
 
     companion object {
         /**
-         * Creates a [LocalizedTextPartial] from the provided [PartialTextComponent] and [LocalizationDictionary]. If
-         * [PartialTextComponent.text] is non null, it should exist in the [LocalizationDictionary]. If it doesn't,
-         * this function will return a failure result.
+         * Creates a [PresentedStackPartial] from the provided [PartialStackComponent] and [using] map. If
+         * [PartialStackComponent.backgroundColor] is non null and contains a color alias, it should exist in the
+         * [using] map. If it doesn't, this function will return a failure result.
          */
         @JvmSynthetic
         operator fun invoke(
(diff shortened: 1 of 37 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
```diff
@@ -1,16 +1,51 @@
...
+    companion object {
+        /**
+         * Creates a [LocalizedTextPartial] from the provided [PartialTextComponent] and [LocalizationDictionary]. If
+         * [PartialTextComponent.text] is non null, it should exist in the [LocalizationDictionary]. If it doesn't,
+         * this function will return a failure result.
+         */
+        @JvmSynthetic
+        operator fun invoke(
+            from: PartialStackComponent,
+            using: Map<ColorAlias, ColorScheme>,
```
tonidero, line 31: Naming seems weird. Should we rename this to `aliases` or `colorAliases` as well?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
```diff
@@ -44,13 +55,25 @@ internal sealed interface ColorStyle {
     value class Gradient(@JvmSynthetic val brush: Brush) : ColorStyle
 }
 
+/**
+ * [ColorStyle]s for light and dark mode.
+ */
+internal data class ColorStyles(
```
tonidero, line 61: Hmm I'm wondering if we should name this following the theme nomenclature in android, something like `DayNightColorStyle`... but then again, that's more verbose so it's probably ok like this.

tonidero (approved):
Looks great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
JayShortway: I'm open to changing the name, but as far as I know Android doesn't use "day" and "night" any more. At least, I'm basing that off the existence of the `isSystemInDarkTheme()` function. We can do `LightDarkColorStyles`? Doesn't roll of the tongue though 😆

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
JayShortway: Agreed! It was based on `LocalizedTextPartial`, which in turn was based on iOS. Changed here: https://github.com/RevenueCat/purchases-android/pull/2076/commits/b1482271e9ac403935addd4d35f7fc003969d501

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
tonidero: Makes sense yeah, i'm fine keeping it as is 👍

PR #2082, opened 2025-01-24
[Paywalls V2] Process `TimelineComponent` overrides and state

tonidero:
### Description
Based on #2047 

This performs the processing for the `TimelineComponent` overrides and its items. It also changes a couple of small things in the network parsing. This does not include the actual view implementation, just an skeleton. The actual implementation will come in a follow-up PR.

PR #2047, opened 2025-01-10
[Paywalls V2] Add `TimelineComponent` network parsing

tonidero:
### Description
This adds the parsing of the `TimelineComponent` to the SDK (it still won't display any UI)

PR #2083, opened 2025-01-24
[Paywalls V2] Add `TimelineComponentView`

tonidero:
### Description
Based on #2047  and #2082 

This adds the `TimelineComponentView` with the actual view for displaying the timeline component. 
image

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -42,6 +42,7 @@ fragment = "1.6.1"
 hamcrest = "1.3"
 emergeGradlePlugin = "4.0.0"
 emergeSnapshots = "1.3.0"
+constraintlayout = "1.1.0"
 
 [plugins]
 android-application = { id = "com.android.application", version.ref = "agp" }
@@ -136,6 +137,7 @@ compose-material3 = { module = "androidx.compose.material3:material3" }
 compose-window-size = { module = "androidx.compose.material3:material3-window-size-class" }
 compose-material3Adaptive = { module = "androidx.compose.material3.adaptive:adaptive" }
 compose-ui-tooling = { module = "androidx.compose.ui:ui-tooling" }
+compose-constraintlayout = { module = "androidx.constraintlayout:constraintlayout-compose", version.ref = "constraintlayout" }
 
 window = { module = "androidx.window:window", version.ref = "window" }
 window-core = { module = "androidx.window:window-core", version.ref = "window" }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedTimelinePartial.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedTimelinePartial.kt
@@ -1,7 +1,17 @@
 package com.revenuecat.purchases.ui.revenuecatui.components
 
+import com.revenuecat.purchases.ColorAlias
 import com.revenuecat.purchases.paywalls.components.PartialTimelineComponent
 import com.revenuecat.purchases.paywalls.components.PartialTimelineComponentItem
+import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toPaddingValues
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.toColorStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.style.TimelineComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.errors.PaywallValidationError
+import com.revenuecat.purchases.ui.revenuecatui.helpers.NonEmptyList
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Result
+import com.revenuecat.purchases.ui.revenuecatui.helpers.map
+import com.revenuecat.purchases.ui.revenuecatui.helpers.orSuccessfullyNull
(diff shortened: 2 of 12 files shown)


gradle/libs.versions.toml
```diff
@@ -42,6 +42,7 @@ fragment = "1.6.1"
 hamcrest = "1.3"
 emergeGradlePlugin = "4.0.0"
 emergeSnapshots = "1.3.0"
+constraintlayout = "1.1.0"
```
tonidero, line 45: Ended up adding the constraint layout library since it simplified laying out the elements in the timeline quite a bit, specially considering we need both the `Title` and `TitleAndDescription` alignments. This seems like it increased the size of the library quite a bit, but if users are minifying their app, most of it should be removed (We probably should change emerge tests to run on a minified build)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PreviewHelpers.kt
```diff
@@ -58,3 +86,98 @@ internal fun previewEmptyState(): PaywallState.Loaded.Components {
         storefrontCountryCode = null,
     )
 }
+
+@Suppress("LongParameterList")
+@JvmSynthetic
+internal fun previewTextComponentStyle(
```
tonidero, line 92: Moving more of these to this file so they can be used across components.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentState.kt
```diff
@@ -79,7 +79,7 @@ internal class TimelineComponentState(
     val visible by derivedStateOf { presentedPartial?.partial?.visible ?: true }
 
     @get:JvmSynthetic
-    val iconSpacing by derivedStateOf { presentedPartial?.partial?.itemSpacing ?: style.itemSpacing }
+    val itemSpacing by derivedStateOf { presentedPartial?.partial?.itemSpacing ?: style.itemSpacing }
```
tonidero, line 82: Use proper name provided by the backend 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -33,36 +57,168 @@ internal fun TimelineComponentView(
...
-    Box(
+    Column(
         modifier = modifier
             .size(timelineState.size)
             .padding(timelineState.margin)
             .padding(timelineState.padding),
     ) {
-        Text("TODO")
+        style.items.map {
+            TimelineItem(
```
tonidero, line 67: I'm going to change this to use `SubcomposeLayout` in order to calculate the height of the connector properly... Right now, it's just the size of the row, which might not match the desired output.

JayShortway: Could you explain what kind of desired output this might not match? Also, I was thinking maybe a `ConstraintLayout` as a parent can offer the desired flexibility? (Not sure, as I don't fully know the issue 😄)

tonidero:
Back to ready for review! In the end I changed it to use a `ConstraintLayout` for the whole Timeline, this made it easier to create constraints for the connector to connect between the icons, which wasn't possible when using constraint layout for each row.

I thought of using `SubscomposeLayout` but then I needed to remove `ConstraintLayout` in order to size and position all elements and the maths were getting annoying, so I went with this approach. Lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedTimelinePartial.kt
```diff
@@ -27,17 +37,44 @@ internal class PresentedTimelinePartial(
...
+                    partial = from,
+                    connectorStyle = from.connector?.let { connector ->
+                        if (colorStyles == null) {
+                            return@let null
+                        }
+                        TimelineComponentStyle.ConnectorStyle(
+                            width = connector.width,
+                            margin = connector.margin.toPaddingValues(),
+                            color = colorStyles,
+                        )
```
JayShortway, line 62: Suggestion for readability: 
```suggestion
                        if (colorStyles != null) {
                            TimelineComponentStyle.ConnectorStyle(
                                 width = connector.width,
                                 margin = connector.margin.toPaddingValues(),
                                 color = colorStyles,
                            )
                        } else null
                        
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -33,36 +60,170 @@ internal fun TimelineComponentView(
         return
     }
 
-    Box(
+    ConstraintLayout(
```
JayShortway, line 63: Haha I had a pending comment that suggested this instead of a `SubcomposeLayout`. Great minds... 🤝

JayShortway (approved):
Very, very nice! 💪 `ConstraintLayout` looks so clean!

PR #2084, opened 2025-01-24
[Paywalls V2] Implements color aliases for borders

JayShortway:
## Description
As the title says. Similar to https://github.com/RevenueCat/purchases-android/pull/2076 but for the `BorderStyle`.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BorderStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BorderStyle.kt
@@ -2,9 +2,16 @@ package com.revenuecat.purchases.ui.revenuecatui.components.properties
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.remember
 import androidx.compose.ui.unit.Dp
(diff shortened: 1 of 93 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BorderStyle.kt
```diff
@@ -24,3 +52,15 @@ internal fun rememberBorderStyle(border: Border): BorderStyle {
         color = colorStyle,
     )
 }
+
+@Composable
+@JvmSynthetic
+internal fun rememberBorderStyle(border: BorderStyles): BorderStyle {
```
tonidero, line 58: Do we still need the remember method that accepts a `Border` after this? I see it's only used for some previews, but we could change to only use this?

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BorderStyle.kt
tonidero: Ohh I see this was done in a follow up PR. Nvm then! :)

PR #2087, opened 2025-01-24
[Paywalls V2] Cleans up after implementing color aliases

JayShortway:
## Description
After implementing color aliases everywhere, there were some functions that could be removed as they should no longer be used. This is done in this PR. This also removes the function that had the `TODO` for supporting color aliases, so that is now resolved.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
```diff
@@ -63,40 +61,28 @@ internal data class ColorStyles(
...
+ * color stops in a [ColorInfo.Gradient] than it is to create a brush to be used in [ColorStyle.Gradient].
+ */
 @JvmSynthetic
-@Composable
-internal fun ColorScheme.toColorStyle(): ColorStyle = colorsForCurrentTheme.toColorStyle()
-
-@JvmSynthetic
-internal fun ColorInfo.toColorStyle(): ColorStyle {
-    return when (this) {
-        is ColorInfo.Alias -> TODO("Color aliases are not yet implemented.")
```
JayShortway, line 84: They are now! 😎

tonidero (approved):

PR #2089, opened 2025-01-27
`PaywallTesterApp` is no longer wrapped in a `Surface`

JayShortway:
## Motivation
While debugging some theming issues in PaywallsTester, I noticed that the entire app composable is wrapped in a `Surface`.  This turned out not to be the cause of the issue, but it made my spidey sense tingle, as `Surface` overrides the `LocalContentColor` of its children. To avoid potential issues, and confusion / red-herring-chasing in the future, I figured we can change this to a `Box` with a `.background()`.


tonidero (approved):
I believe this should be ok 👍

PR #2090, opened 2025-01-28
[Paywalls V2] Adds support for custom fonts

JayShortway:
## Description
Lots of files and lines changed, but the complexity is quite contained. I've added comments to files that only contain "ripple effect changes". A lot of lines in the diff are caused by tests. 

This adds support for custom fonts, both local and Google Fonts. It's the Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4631. 

## FontSpec
It introduces the concept of a `FontSpec`. This is an intermediately-resolved font. Why intermediate? 
1. Because on-device, we can determine if a `FontInfo.Name` is a system font or a local font. This allows us to look up the resources for these local fonts in an optimized way, only once for each font. 
2. Because we cannot fully resolve it in the validation phase, as we don't know what `FontWeight` will apply, as the overrides can influence that. Final resolution happens in `TextComponentState`. 

## How does it work?
1. `TextComponent`s have an optional `FontAlias` field. 
4. `UiConfig` contains a map of `FontAlias` to `FontInfo` (wrapped in a `FontsConfig`). 
5. `OfferingToStateMapper` will convert this map to a map of `FontAlias` to `FontSpec`, and passes it to `StyleFactory`.
6. `StyleFactory` will look up all `FontAlias`es declared in every `TextComponent`, and accumulate errors for any that are not found. The resulting `TextComponentStyle` has a `FontSpec` field.
7. `TextComponentState` resolves the `FontSpec` if it's non-null, and provide the resulting `FontFamily` to the `TextComponentView`.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TextComponent.kt
```diff
@@ -44,7 +45,7 @@ constructor(
     val backgroundColor: ColorScheme? = null,
     @get:JvmSynthetic
     @SerialName("font_name")
-    val fontName: String? = null,
+    val fontName: FontAlias? = null,
```
JayShortway, line 48: Deserializing this as a `FontAlias` directly allows us to look up fonts in `UiConfig.app.fonts` in a type safe way.

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/TextComponentTests.kt
```diff
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.paywalls.components
 
+import com.revenuecat.purchases.FontAlias
```
JayShortway, line 3: This file only contains ripple effect changes caused by changing `TextComponent.fontName` to be a `FontAlias`.

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/common/ComponentOverridesTests.kt
```diff
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.paywalls.components.common
 
+import com.revenuecat.purchases.FontAlias
```
JayShortway, line 3: This file only contains ripple effect changes caused by changing `TextComponent.fontName` to be a `FontAlias`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/TextComponentStyle.kt
```diff
@@ -26,7 +26,7 @@ internal class TextComponentStyle(
     @get:JvmSynthetic
     val fontWeight: FontWeight?,
     @get:JvmSynthetic
-    val fontFamily: FontFamily?,
+    val fontSpec: FontSpec?,
```
JayShortway, line 29: This is needed because we can only actually resolve the `FontFamily` when we know the `FontWeight` that's applied, which is only known in the `TextComponentState`/`TextComponentView`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -891,7 +891,7 @@ private fun StackComponentView_Preview_ZLayer() {
                         ),
                         fontSize = 15,
                         fontWeight = FontWeight.REGULAR.toFontWeight(),
-                        fontFamily = null,
+                        fontSpec = null,
```
JayShortway, line 894: This file only contains ripple effect changes caused by changing `TextComponentStyle.fontFamily` to `fontSpec`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
```diff
@@ -32,6 +32,7 @@ import com.revenuecat.purchases.ui.revenuecatui.components.previewEmptyState
 import com.revenuecat.purchases.ui.revenuecatui.components.previewTextComponentStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.FontSpec
```
JayShortway, line 35: This file only contains ripple effect changes caused by changing `TextComponentStyle.fontFamily` to `fontSpec`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -221,6 +221,7 @@ private fun MDParagraph(
                     .copy(
                         fontWeight = fontWeight,
                         fontSize = if (applyFontSizeToParagraph) fontSize else style.fontSize,
+                        fontFamily = fontFamily,
```
JayShortway, line 224: `FontFamily` was not being applied to the style. Not a priority, but I wouldn't mind replacing this `Markdown` component at some point 😅

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/BuildPresentedPartialTests.kt
```diff
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.ui.revenuecatui.components
 
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.toArgb
+import com.revenuecat.purchases.FontAlias
```
JayShortway, line 5: This file only contains ripple effect changes caused by adding a `fontAliases` map to `StyleFactory`.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LocalizedTextPartialTests.kt
```diff
@@ -3,6 +3,7 @@ package com.revenuecat.purchases.ui.revenuecatui.components
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.toArgb
 import com.revenuecat.purchases.ColorAlias
+import com.revenuecat.purchases.FontAlias
```
JayShortway, line 6: This file contains a bunch of ripple effect changes caused by adding a `fontAliases` map to `StyleFactory`, and by changing `TextComponent.fontName` to be a `FontAlias`.

It also contains some additional tests.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentViewTests.kt
```diff
@@ -75,7 +75,7 @@ class ButtonComponentViewTests {
                             ),
```
JayShortway, line 75: This file only contains ripple effect changes.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/pkg/PackageComponentViewTests.kt
```diff
@@ -39,9 +39,9 @@ import com.revenuecat.purchases.paywalls.components.properties.HorizontalAlignme
 import com.revenuecat.purchases.ui.revenuecatui.components.style.PackageComponentStyle
```
JayShortway, line 39: This file only contains ripple effect changes.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentViewWindowTests.kt
```diff
@@ -138,6 +138,7 @@ internal class StackComponentViewWindowTests {
                 )
```
JayShortway, line 138: This file only contains ripple effect changes.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
```diff
@@ -51,11 +51,11 @@ import com.revenuecat.purchases.ui.revenuecatui.components.style.StyleFactory
 import com.revenuecat.purchases.ui.revenuecatui.components.style.TextComponentStyle
```
JayShortway, line 51: This file only contains ripple effect changes.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
tonidero: Oops, nice catch! And yeah I wouldn't be opposed at all... We kinda copied some functionality... but I would agree the current code is not the best 😅 .

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -0,0 +1,119 @@
...
+    when (info) {
+        is FontInfo.GoogleFonts -> FontSpec.Google(name = info.value)
+        is FontInfo.Name -> when (info.value) {
+            FontFamily.SansSerif.name -> FontSpec.Generic.SansSerif
+            FontFamily.Serif.name -> FontSpec.Generic.Serif
+            FontFamily.Monospace.name -> FontSpec.Generic.Monospace
+            FontFamily.Cursive.name -> FontSpec.Generic.Cursive
+            else -> {
+                @SuppressLint("DiscouragedApi")
+                val fontId = getResourceIdentifier(name = info.value, type = "font")
```
tonidero, line 106: I've been wondering whether it makes sense for the dashboard to have an option to use... But I guess that wouldn't work super well with supporting both iOS + Android... 

As for the optimization, I guess if we find this is a problem we could also try to further optimize this by caching this information across paywall restarts... (Surviving app restarts would be trickier though) but I think it might be ok with this. We might want to do some performance checks on an older device just to make sure though.

tonidero (approved):
This is great! Thanks for tackling this! Quite a few things to handle with this one 🫶

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
JayShortway: >to have an option to use...

Is this missing a word? 😅 Otherwise, could you clarify what you mean? 

Those are indeed good strategies we can use to further optimize! We should definitely do a round of performance checks, but I think that can be done separately.

tonidero: right, sorry, I meant `to have an option to use either a resource or system font...` 🙇. But yeah, I think this is ok for now 👍

PR #2091, opened 2025-01-28
[Paywalls V2] No longer provides the entire `UiConfig` to `StyleFactory`

JayShortway:
## Description
Just a bit of a consistency update after #2090. Just as we do with `fontAliases`, we now only pass the `colorAliases` instead of the entire `UiConfig`. It is not the `StyleFactory`'s job to figure out what it needs from `UiConfig`. 

No actual logic changes in this PR.


tonidero (approved):
Nice cleanup!

PR #2095, opened 2025-01-30
[Paywalls V2] Add `CarouselComponentView`

tonidero:
### Description
This adds the `CarouselComponentView` that actually displays the carousel

Second part of Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4722

https://github.com/user-attachments/assets/9649e370-d207-4b81-9119-7c60dc9cff13

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
@@ -2,33 +2,62 @@
 
 package com.revenuecat.purchases.ui.revenuecatui.components.carousel
 
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.material3.Text
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.pager.HorizontalPager
+import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import com.revenuecat.purchases.paywalls.components.CarouselComponent
+import com.revenuecat.purchases.paywalls.components.properties.Dimension
+import com.revenuecat.purchases.paywalls.components.properties.FlexDistribution
+import com.revenuecat.purchases.paywalls.components.properties.HorizontalAlignment
 import com.revenuecat.purchases.paywalls.components.properties.Shape
 import com.revenuecat.purchases.paywalls.components.properties.Size
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import com.revenuecat.purchases.ui.revenuecatui.components.PaywallAction
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toShape
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.background
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.border
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.shadow
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.size
 import com.revenuecat.purchases.ui.revenuecatui.components.previewEmptyState
+import com.revenuecat.purchases.ui.revenuecatui.components.previewTextComponentStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.BorderStyles
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyle
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyles
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.ShadowStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.forCurrentTheme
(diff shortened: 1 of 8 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -40,16 +73,146 @@ internal fun CarouselComponentView(
         return
     }
 
-    Text("TODO", modifier = modifier)
+    val backgroundColorStyle = carouselState.backgroundColor?.forCurrentTheme
+    val borderStyle = carouselState.border?.let { rememberBorderStyle(border = it) }
+    val shadowStyle = carouselState.shadow?.let { rememberShadowStyle(shadow = it) }
+    val composeShape by remember(carouselState.shape) { derivedStateOf { carouselState.shape.toShape() } }
```
JayShortway, line 79: I don't _think_ we need `derivedStateOf` here, as we're only reading 1 state value (`carouselState.shape`). That said, maybe `carouselState.shape` can already be the right type so we don't have to convert it here?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -40,16 +73,146 @@ internal fun CarouselComponentView(
...
+
+    val pageCount = style.slides.size
+
+    val initialPage = if (carouselState.loop) {
+        var currentPage = Int.MAX_VALUE / 2
+        while (currentPage % pageCount != carouselState.initialSlideIndex) {
+            currentPage++
+        }
+        currentPage
```
JayShortway, line 88: I think a oneliner comment to explain what's happening here would help grok it more easily. 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -40,16 +73,146 @@ internal fun CarouselComponentView(
...
+            userScrollEnabled = style.autoAdvance == null,
+        ) { page ->
+            StackComponentView(
+                style = carouselState.slides[page % pageCount],
+                state = state,
+                clickHandler = clickHandler,
+            )
+        }
+
+        carouselState.pageControl?.let {
```
JayShortway, line 147: How's the spacing between the pager and the indicator defined?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -40,16 +73,146 @@ internal fun CarouselComponentView(
...
+) {
+    Row(
+        modifier = modifier.align(Alignment.CenterHorizontally),
+        horizontalArrangement = Arrangement.Center,
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        repeat(pageCount) { iteration ->
+            Indicator(if (currentPageIndex == iteration) pageControl.active else pageControl.default)
+        }
+    }
```
JayShortway, line 173: I love the simplicity of this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -40,16 +73,146 @@ internal fun CarouselComponentView(
...
+}
+
+@Composable
+private fun Indicator(indicator: CarouselComponentStyle.IndicatorStyles) {
+    Box(
+        modifier = Modifier
+            .padding(indicator.margin)
+            .padding(horizontal = indicator.spacing / 2)
+            .clip(Shape.Pill.toShape())
```
JayShortway, line 182: Micro optimization, as we don't need to use RC shapes here: 
```suggestion
            .clip(CircleShape)
```

JayShortway (approved):
This is such a cool component!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: Hmm right... I copied this over from the `StackComponentView`, but we changed it in there to have our `Shape` type, because we needed access to the original rounded corners size for the badge, but that constraint doesn't exist here since we don't have badges... So we can probably just have the state hold the compose shape directly 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: It's provided by the indicator margins. If the active/default indicators have a 10.dp top padding, it would have a 10.dp spacing. That's also configurable from the dashboard (I'm not sure if the schema would change but that's how I was thinking about it cc @joshdholtz)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: Right, I thought about this, but I wanted to use a pill shape here, in case someone wanted to have pill shaped indicator. Note that in the designs, width and height of the indicators are configurable, both for the active and default state.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2095/commits/3b72a05e33715b077dd1605de874a743d96fdff1

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: I added some changes to the preview so it's clear it might not have a circle shape: https://github.com/RevenueCat/purchases-android/pull/2095/commits/0012886ec677f7366a3b9205a3a8002387d1e15c

PR #2096, opened 2025-01-30
refactor: Introduce EventsManager to track events for different features

facumenzella:
### Motivation
Introduce events for Customer Center

### Description
I am working on introducing events for customer center, and I needed an event manager an realized PaywallEventManager could be generalized to be reused. This introduces EventsManager, with backwards compatibility for previous events


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -320,7 +320,13 @@ internal class PurchasesFactory(
...
                 identityManager,
                 eventsDispatcher,
-                backend,
+                postPaywallEvents = { request, onSuccess, onError ->
+                    backend.postPaywallEvents(
+                        paywallEventRequest = request,
+                        onSuccessHandler = onSuccess,
+                        onErrorHandler = onError,
+                    )
+                },
```
facumenzella, line 329: I wasn't sure about this. Is there a better way to handle it?

facumenzella: I didn't want to refactor the `Backend` class, so this felt easier

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,99 @@
...
+
+internal class EventsManager(
+    private val fileHelper: EventsFileHelper<BackendEvent>,
+    private val identityManager: IdentityManager,
+    private val eventsDispatcher: Dispatcher,
+    private val postEvents: (
+        EventRequest,
+        () -> Unit,
+        (error: PurchasesError, shouldMarkAsSynced: Boolean) -> Unit,
+    ) -> Unit,
```
facumenzella, line 21: This is temporary, till Backend has one function to track everything

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEventsManager.kt
```diff
@@ -1,89 +1,49 @@
...
-import com.revenuecat.purchases.common.debugLog
-import com.revenuecat.purchases.common.errorLog
-import com.revenuecat.purchases.common.verboseLog
+import com.revenuecat.purchases.common.events.EventRequest
+import com.revenuecat.purchases.common.events.EventsManager
 import com.revenuecat.purchases.identity.IdentityManager
 import com.revenuecat.purchases.utils.EventsFileHelper
 
 @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
 internal class PaywallEventsManager(
```
facumenzella, line 12: This one should be removed, once the other implementation works 🙃

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -302,25 +303,42 @@ internal class PurchasesFactory(
...
-            PaywallEventsManager(
-                EventsFileHelper(
+            EventsManager(
+                legacyEventsFileHelper = EventsFileHelper(
                     FileHelper(context),
-                    PaywallEventsManager.PAYWALL_EVENTS_FILE_PATH,
+                    EventsManager.PAYWALL_EVENTS_FILE_PATH,
+                    PaywallStoredEvent::toString,
                     PaywallStoredEvent::fromString,
                 ),
```
facumenzella, line 322: One thing I've thought is that maybe its worth just declaring PaywallsEventManager here instead, since it's already tested.

facumenzella:
@RevenueCat/coresdk one thing I am thinking is that maybe its worth is keeping PaywallsEventManager instead of the legacy FileHelper because it's already tested.

Either that, or just keep the tests from PaywallsManager instead EventsManagerTests

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventRequest.kt
```diff
@@ -1,12 +1,13 @@
-package com.revenuecat.purchases.paywalls.events
+package com.revenuecat.purchases.common.events
 
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 
 @Serializable
-internal data class PaywallEventRequest(
-    val events: List<PaywallBackendEvent>,
+public data class EventRequest internal constructor(
```
vegaro, line 7: does this need to be made public?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterBackendEvent.kt
```diff
@@ -0,0 +1,23 @@
+package com.revenuecat.purchases.customercenter.events
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CustomerCenterBackendEvent(
```
vegaro, line 7: this can be made `internal`

you would need to add `@get:JvmSynthetic ` to each property as well to hide it from Java, since Java doesn't recognize internal. By the way, if you can add that to `PaywallBackendEvent` as well (JvmSynthetic is missing there)

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterBackendEvent.kt
```diff
@@ -0,0 +1,23 @@
...
+    val id: String,
+    val revisionId: Int,
```
vegaro, line 9: for consistency, let's make this one `revisionID` and add the `@SerialName`

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/FeatureEvent.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.revenuecat.purchases.customercenter.events.CustomerCenterEvent
+import com.revenuecat.purchases.paywalls.events.PaywallEvent
+
+interface FeatureEvent {
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    val paywallEvent: PaywallEvent? get() = null
+
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    val customerCenterEvent: CustomerCenterEvent? get() = null
+}
```
vegaro, line 13: This interface is a bit out of the norm. Maybe a sealed class fits here better.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,187 @@
...
+    fun flushEvents() {
+        enqueue {
+            if (flushInProgress) {
+                debugLog("Flush already in progress.")
+                return@enqueue
+            }
+            flushInProgress = true
+
+            if (!legacyFlushDone) {
+                flushLegacyEvents()
```
vegaro, line 101: this is being enqueued, so I would not call `legacyFlushDone` done, since it's actually not guaranteed it's been completed. Maybe `legacyFlushTriggered`?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,187 @@
...
+    private val postEvents: (
+        EventRequest,
+        () -> Unit,
+        (error: PurchasesError, shouldMarkAsSynced: Boolean) -> Unit,
+    ) -> Unit,
+    private val flushCount: Int = 50,
+) {
+
+    companion object {
+        const val PAYWALL_EVENTS_FILE_PATH = "RevenueCat/paywall_event_store/paywall_event_store.jsonl"
```
vegaro, line 33: This is only used in this file now, so it could be `private`

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,187 @@
...
+        const val EVENTS_FILE_PATH_NEW = "RevenueCat/event_store/event_store.jsonl"
+
+        internal val json = Json {
```
vegaro, line 36: I think this could be private?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private fun getLegacyStoredEvents(): List<BackendEvent> {
+        var events: List<PaywallStoredEvent?> = emptyList()
+        legacyEventsFileHelper.readFile { sequence ->
+            events = sequence.take(flushCount).toList()
+        }
+        return events
+            .filterNotNull()
+            .map { BackendEvent.Paywalls(it.toPaywallBackendEvent()) }
```
vegaro, line 181: Since this is returning `BackendEvent.Paywalls`, I would make that clear in the function name. Also you can use `mapNotNull` here and remove the `filterNotNull`

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,187 @@
...
+                    }
+                },
+            )
+        }
+    }
+
+    private fun flushLegacyEvents() {
+        enqueue {
+            val storedLegacyEventsWithNullValues = getLegacyStoredEvents()
+            val storedLegacyEvents = storedLegacyEventsWithNullValues.filterNotNull()
```
vegaro, line 140: `getLegacyStoredEvents` is already returning not null values, so this is not necessary

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@Serializable
+internal sealed class BackendEvent : Event {
+    @Serializable
+    @SerialName("customer_center")
+    data class CustomerCenter(val event: CustomerCenterBackendEvent) : BackendEvent()
+
+    @Serializable
+    @SerialName("paywalls")
+    data class Paywalls(val event: PaywallBackendEvent) : BackendEvent()
```
vegaro, line 19: I find a bit confusing that `BackendEvent.Paywalls` contains a `PaywallBackendEvent`. Maybe it could be `PaywallBackendEventData`?

vegaro (changes requested):

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/FeatureEvent.kt
facumenzella: I wanted to do something similar to a protocol in swift 😅 
I guess that this could be seen like something like `AnyEvent` wrapper. I'll go with a sealed class and rename it

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -0,0 +1,51 @@
...
+        val offeringID: String,
+        @SerialName("paywall_revision")
+        val paywallRevision: Int,
+        val timestamp: Long,
+        @SerialName("display_mode")
+        val displayMode: String,
+        @SerialName("dark_mode")
+        val darkMode: Boolean,
+        @SerialName("locale")
+        val localeIdentifier: String,
```
vegaro, line 49: Looking at iOS, it looks like this should be `locale`? I might be wrong, but it would be good to double check

facumenzella: Yes, it's locale for the event we send to the SDK (see EventsRequest+Paywall.swift)

```swift
    private enum CodingKeys: String, CodingKey {

        case id
        case version
        case type
        case appUserID = "appUserId"
        case sessionID = "sessionId"
        case offeringID = "offeringId"
        case paywallRevision
        case timestamp
        case displayMode
        case darkMode
        case localeIdentifier = "locale"

    }
```

facumenzella: We're good, it's locale in iOS as well

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventRequest.kt
```diff
@@ -0,0 +1,29 @@
+package com.revenuecat.purchases.common.events
+
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.polymorphic
+
+@Serializable
+data class EventRequest internal constructor(
```
vegaro, line 9: I would name this `EventsRequest` since it contains multiple Events

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventRequest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    companion object {
+        val json = Json {
+            serializersModule = SerializersModule {
+                polymorphic(BackendEvent::class) {
+                    subclass(BackendEvent.CustomerCenter::class, BackendEvent.CustomerCenter.serializer())
+                    subclass(BackendEvent.Paywalls::class, BackendEvent.Paywalls.serializer())
+                }
+            }
+            classDiscriminator = "discriminator"
```
vegaro, line 21: I thought in the request there's no discriminator?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -0,0 +1,189 @@
...
+        (error: PurchasesError, shouldMarkAsSynced: Boolean) -> Unit,
+    ) -> Unit,
+) {
+
+    companion object {
+        private const val FLUSH_COUNT = 50
+        private const val PAYWALL_EVENTS_FILE_PATH = "RevenueCat/paywall_event_store/paywall_event_store.jsonl"
+
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        const val EVENTS_FILE_PATH_NEW = "RevenueCat/event_store/event_store.jsonl"
```
vegaro, line 37: You can make it internal and it will be visible by tests

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.revenuecat.purchases.paywalls.events.PaywallEvent
+import com.revenuecat.purchases.utils.Event
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal sealed class BackendStoredEvent : Event {
+    @Serializable
+    @SerialName("customer_center")
+    data class CustomerCenter(val event: CustomerCenterBackendEvent) : BackendStoredEvent()
```
vegaro, line 15: I am very confused with the naming. There are:
- `BackendStoredEvent. CustomerCenter `
- `CustomerCenterBackendEvent`
- `BackendEvent.CustomerCenter`

Also, can't event just be of type `BackendEvent.CustomerCenter` and remove `CustomerCenterBackendEvent` completely? `BackendEvent.CustomerCenter` and `CustomerCenterBackendEvent` look the same

Then you wouldn't need `toBackendEvent` because you're storing the thing you're sending to the backend directly. You still need this sealed class because it adds the discriminator and to be able to know how to deserialise from storage

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
```diff
@@ -54,7 +55,7 @@ import kotlin.random.Random
 @RunWith(AndroidJUnit4::class)
 @Config(manifest = Config.NONE)
 @Suppress("DEPRECATION")
-internal class PurchasesTest : BasePurchasesTest() {
+internal class PurchasesTest  : BasePurchasesTest() {
```
vegaro, line 58:
```suggestion
internal class PurchasesTest : BasePurchasesTest() {
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
facumenzella: My reasoning was to split this in three sections:
1. Feature code
2. Storing code
3. Requests code
---
1. Each feature has it's own events: `PaywallEvent` and `CustomerCenterEvent`
```kotlin
sealed class FeatureEvent {
    data class Paywall(val event: PaywallEvent) : FeatureEvent()
    data class CustomerCenter(val event: CustomerCenterEvent) : FeatureEvent()
}
```

2. Events manager works with `BackendStoredEvent` + `CustomerCenterBackendEvent` + `PaywallBackendEvent`.
```kotlin
@Serializable
internal sealed class BackendStoredEvent : Event {
    @Serializable
    @SerialName("customer_center")
    data class CustomerCenter(val event: CustomerCenterBackendEvent) : BackendStoredEvent()

    @Serializable
    @SerialName("paywalls")
    data class Paywalls(val event: PaywallBackendEvent) : BackendStoredEvent()
}
```

3. Backend works with EventsRequest + BackendEvent

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventRequest.kt
facumenzella: I ended up adding it because tests were failing. If I understood correctly, you must use a `classDiscriminator` when using polymorphic

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -0,0 +1,99 @@
...
+     */
+    @Serializable
+    data class CustomerCenter(
+        val id: String,
+        @SerialName("revision_id")
+        val revisionID: Int,
+        val type: String,
+        @SerialName("app_user_id")
+        val appUserID: String,
+        @SerialName("session_id")
```
vegaro, line 35: note this should be `app_session_id`

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Converts a `BackendStoredEvent` into a `BackendEvent`.
+ *
+ * @receiver The stored backend event to be converted.
+ * @return A `BackendEvent` instance or `null` if conversion is not possible.
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal fun BackendStoredEvent.toBackendEvent(): BackendEvent? {
+    return when (this) {
+        is BackendStoredEvent.Paywalls -> {
```
vegaro, line 43: this can be now `is BackendStoredEvent.Paywalls -> { this.event }` since it's already a `BackendEvent.Paywalls`. You might not even need `toBackendEvent` at all? I would at least rename it to indicate it returns the underlying event and it's not a conversion

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
vegaro: actually since it's enqueued... there's a race condition if `flushEvents` executes again before flushLegacyEvents finishes? A way to solve it is removing the enqueuing in `flushLegacyEvents`. Or actually... shouldn't it go before `flushLegacyEvents` and not after?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/FeatureEvent.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+sealed class FeatureEvent {
+
+    /**
+     * Represents an event related to paywalls.
+     *
+     * @property event The associated `PaywallEvent`.
+     */
+    data class Paywall(val event: PaywallEvent) : FeatureEvent()
```
vegaro, line 20: not a merge stopper but this could be directly the `PaywallEvent`? So you could make the `PaywallEvent` a `FeatureEvent`

vegaro (approved):
I think this should be good to merge. Just some small comments

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
facumenzella: I'll put it before since this is just a fire-and-forget thing.

PR #2098, opened 2025-01-30
[Paywalls V2] Adds new price calculations needed for Variables V2

JayShortway:
## Description
As the title says. These will be used in the next PR. Added API tests for anything that's newly public.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.models
 
 import android.os.Parcelable
+import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.common.errorLog
 import kotlinx.parcelize.Parcelize
 import kotlin.math.roundToInt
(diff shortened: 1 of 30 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
```diff
@@ -60,17 +61,31 @@ data class Period(
...
+                errorLog("Unknown period unit trying to get value in days: $unit")
+                0.0
+            }
+        }
+
     /**
      * The period value in week units. This is an approximated value.
      */
-    internal val valueInWeeks: Double
+    @InternalRevenueCatAPI
```
tonidero, line 80: Hmm I'm wondering if it would make sense to just make it public... But yeah, unless we have some request for it, I think this is the best approach 👍

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
```diff
@@ -60,17 +61,31 @@ data class Period(
...
+    val valueInWeeks: Double
         get() = when (unit) {
             Unit.DAY -> value / PeriodConstants.DAYS_PER_WEEK
             Unit.WEEK -> value.toDouble()
             Unit.MONTH -> value.toDouble() * PeriodConstants.WEEKS_PER_MONTH
             Unit.YEAR -> value * PeriodConstants.WEEKS_PER_YEAR
             Unit.UNKNOWN -> {
-                errorLog("Unknown period unit trying to get value in months: $unit")
+                errorLog("Unknown period unit trying to get value in weeks: $unit")
```
tonidero, line 88: Oops

JayShortway: I figured fixing this word while I'm here can't hurt. 😄

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
JayShortway: Yea that was indeed my exact thought haha. We can quite easily open it up later.

PR #2100, opened 2025-01-30
Publishes PaywallsTester to Internal testing track on every commit to `main`

JayShortway:
## Description
As the title says. It also enables Paywalls V2 for the Closed Testing track, but I think that should be okay. Any other behavior should remain unchanged.


.circleci/config.yml
```diff
@@ -709,6 +718,16 @@ workflows:
     jobs:
       - prepare-next-version: *only-main-branch
 
+  publish-paywall-tester-to-internal-track:
+    when:
+      and:
+        - not:
+            equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
+        - equal: [ "publish-paywall-tester-to-internal", << pipeline.git.branch >> ]
```
JayShortway, line 726: Note to self: change this to "main" before merging.

PR #2101, opened 2025-01-30
[Paywalls V2] Adds deserialization of `TabsComponent`

JayShortway:
## Description
The tabs feature is actually made up of a few different components, but they're all in `TabsComponent.kt`. For reference, check this iOS PR: https://github.com/RevenueCat/purchases-ios/pull/4648.

There's no logic or rendering yet. It's just parsing the network response.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TabsComponent.kt
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    @InternalRevenueCatAPI
+    @Serializable
+    sealed interface TabControl {
+        @InternalRevenueCatAPI
+        @Poko
+        @Serializable
+        @SerialName("buttons")
+        class Buttons(@get:JvmSynthetic val stack: StackComponent) : TabControl
```
tonidero, line 100: Hmm I believe this is like this in iOS as well, but I wonder if this should receive a list of `TabControlButtonComponent` instead... Then again, I guess this is indeed more flexible so 🤷

tonidero (approved):
Looks great!

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/TabsComponentTests.kt
```diff
@@ -0,0 +1,386 @@
...
+                        """.trimIndent(),
+                    expected = StackComponent(
+                        components = listOf(
+                            TabControlComponent,
+                        ),
+                    )
+                ),
+            ),
+            arrayOf(
+                "tabs - all values present",
```
JayShortway, line 151: @joshdholtz Could you double check the schema I have here? (Below this comment 👇)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TabsComponent.kt
JayShortway: Yea, it's a bit confusing, but the `TabControlButtonComponent`s don't need to be a direct child of `Buttons.stack`. They can be as nested as desired. So that's why this has a free-form `StackComponent` instead of a more limiting list of `TabControlButtonComponent`s.

PR #2102, opened 2025-01-30
Customer Center use remote appearance config

vegaro:
For this configuration:

Screenshot 2025-01-30 at 19 40 25

This is how it's built:

| Light | Dark |
|--------|--------|
| Screenshot 2025-01-30 at 19 47 34 | Screenshot 2025-01-30 at 19 41 27 |
| Screenshot 2025-01-30 at 19 47 02 | Screenshot 2025-01-30 at 19 46 19 |
| Screenshot 2025-01-30 at 19 48 01 | Screenshot 2025-01-30 at 19 48 24 |
| Screenshot 2025-01-30 at 19 49 02 | Screenshot 2025-01-30 at 19 48 44 |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
@@ -5,6 +5,7 @@
 package com.revenuecat.purchases.ui.revenuecatui.customercenter
 
 import androidx.activity.compose.BackHandler
+import androidx.compose.foundation.background
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
@@ -37,6 +38,7 @@ import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesErrorCode
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
+import com.revenuecat.purchases.ui.revenuecatui.components.modifier.background
 import com.revenuecat.purchases.ui.revenuecatui.composables.ErrorDialog
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.actions.CustomerCenterAction
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCenterConfigTestData
@@ -143,8 +145,20 @@ private fun InternalCustomerCenter(
         } else {
             appearance.light?.accentColor?.colorInt
         }?.let { Color(it) }
+
+        // Only change background if there is a promotional offer
+        var backgroundColor: Color? = null
+        if (state.promotionalOfferData != null) {
+            backgroundColor = if (isDark) {
+                appearance.dark?.backgroundColor?.colorInt
+            } else {
+                appearance.light?.backgroundColor?.colorInt
+            }?.let { Color(it) }
+        }
+
         colorScheme = MaterialTheme.colorScheme.copy(
             primary = accentColor ?: MaterialTheme.colorScheme.primary,
+            background = backgroundColor ?: MaterialTheme.colorScheme.background,
         )
     } else {
         colorScheme = MaterialTheme.colorScheme
@@ -154,7 +168,8 @@ private fun InternalCustomerCenter(
         colorScheme = colorScheme,
     ) {
         CustomerCenterScaffold(
-            modifier = modifier,
+            modifier = modifier
+                .background(MaterialTheme.colorScheme.background),
             title = title,
             onAction = onAction,
             navigationButtonType =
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -130,25 +134,52 @@ private fun InternalCustomerCenter(
...
+            appearance.getColorForTheme(isDark) { it.backgroundColor }
         } else {
-            CustomerCenterState.NavigationButtonType.CLOSE
-        },
+            null
+        }
+
+        colorScheme = MaterialTheme.colorScheme.copy(
+            primary = accentColor ?: MaterialTheme.colorScheme.primary,
+            background = backgroundColor ?: MaterialTheme.colorScheme.background,
```
tonidero, line 153: Hmm depending on the settings of the app, this might not look great (like in our test app) but I guess they can change their theme colors if needed... This might be trickier on the hybrids though.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferView.kt
```diff
@@ -33,15 +36,22 @@ import com.revenuecat.purchases.models.toRecurrenceMode
 import com.revenuecat.purchases.ui.revenuecatui.composables.AppIcon
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCenterConfigTestData
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.PromotionalOfferData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.getColorForTheme
 
 @JvmSynthetic
 @Composable
 internal fun PromotionalOfferView(
```
tonidero, line 43: Not from this PR but I wonder if this should be called `PromotionalOfferScreen`? FWIW, it's not recommended to call composables "Views" since they are not actually views. We do call them views in Paywall V2, but it's mostly a name collision issue, since the name would collide with the original "data" components (as in "TextComponent" is the data holder for "TextComponentView"). FFTI though, since it's not part of this PR!

tonidero (approved):
I have a small concern about this not being customizable enough, specially for the hybrids... But I'm ok merging this for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -130,25 +134,52 @@ private fun InternalCustomerCenter(
...
-    CustomerCenterScaffold(
-        modifier = modifier,
-        title = title,
-        onAction = onAction,
-        navigationButtonType =
-        if (state is CustomerCenterState.Success) {
-            state.navigationButtonType
+
+    val colorScheme: ColorScheme
+    if (state is CustomerCenterState.Success) {
```
JayShortway, line 139: Nit, but we can assign `colorScheme` directly to the `if`: 
```kotlin
val colorScheme = if (state is CustomerCenterState.Success) {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: hmmm... I think we would have the same issue in iOS right? The problem is the backup if background color is null?

JayShortway (approved):
Nice! Just 1 nit besides Toni's comments.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
tonidero: Well, this background is only applied to the promotional offer screen, but not to the other screens right? I believe the problem would be on those screens. And yeah I think the same issue could apply in iOS, depending on their app's settings. Maybe not something to act on right now, but ideally we make sure with the default colors of a new app, the other screens look ok

vegaro: right... maybe adding a modifier? not sure. I am honestly starting to not like making the colors configurable remotely lol

tonidero: That's where I moved to thinking about using the Paywalls V2 editor for this :P

vegaro: Right, same. This is just a way to cover that need for now, but it will be replaced with paywalls v2 editor at some point

PR #2104, opened 2025-01-31
Prices in accept promotional offer button

vegaro:
### Checklist


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/extensions/SubscriptionOptionExtensions.kt
```diff
@@ -0,0 +1,148 @@
+@file:JvmSynthetic
+@file:OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+
+package com.revenuecat.purchases.ui.revenuecatui.customercenter.extensions
```
tonidero, line 4: Hmm I wonder if it makes sense to move this outside the `customercenter` package, and make it more generic in case we want to use them for paywalls... (I'm honestly not sure if we would like to use these extensions in Paywalls though cc @JayShortway)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PriceExtensions.kt
```diff
@@ -0,0 +1,22 @@
+package com.revenuecat.purchases.ui.revenuecatui.extensions
+
+import com.revenuecat.purchases.models.Period
+import com.revenuecat.purchases.models.Price
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.endsIn00Cents
+import com.revenuecat.purchases.ui.revenuecatui.data.processed.getTruncatedFormatted
+import java.util.Locale
+
+internal fun Price.localized(locale: Locale, showZeroDecimalPlacePrices: Boolean): String {
```
tonidero, line 9: Need to add the `@JvmSynthetic` annotations in this file.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Stubs.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.revenuecat.purchases.models.InstallmentsInfo
+import com.revenuecat.purchases.models.Period
+import com.revenuecat.purchases.models.Price
+import com.revenuecat.purchases.models.PricingPhase
+import com.revenuecat.purchases.models.PurchasingData
+import com.revenuecat.purchases.models.SubscriptionOption
+import com.revenuecat.purchases.models.toRecurrenceMode
+
+@SuppressWarnings("LongParameterList")
+internal fun stubSubscriptionOption(
```
tonidero, line 15: Also `@JvmSynthetic` annotations.

tonidero, line 15: Additionally, what do you think about naming this `previewSubscriptionOption`? And the file could also be something like `PreviewDataFactory` or something like that. I feel that's a better naming than `stub` so we don't confuse this with test code

tonidero (approved):
LGTM! Just some comments and questions

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Stubs.kt
vegaro: I am actually using `stubPricingPhase` in tests (`SubscriptionOptionExtensionsTest`) as well, to avoid duplication. What do you think?

tonidero: Right, I think it makes sense! But I still think it feels weird to add "stubs" to production code, which is why I would like to use the "preview" naming, which can still be used in tests I think.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/extensions/SubscriptionOptionExtensions.kt
JayShortway: Thanks for pinging! They do seem very Customer Center specific. Also, the variables are different from paywalls. So I think it's fine to leave them here for now.

PR #2105, opened 2025-02-03
[Paywalls V2] Add CarouselComponent page indicator animation

tonidero:
### Description
This adds a transition animation to the page indicator when switching pages 

https://github.com/user-attachments/assets/3392510d-74b9-4ac2-9fef-6b5c7c191956

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
@@ -2,6 +2,7 @@
 
 package com.revenuecat.purchases.ui.revenuecatui.components.carousel
 
+import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -160,20 +165,104 @@ private fun ColumnScope.PagerIndicator(
...
+    }
+    val targetSpacing by remember(progress) {
+        derivedStateOf {
+            lerpUnit(
+                pageControl.default.spacing,
+                pageControl.active.spacing,
+                progress,
+            )
+        }
+    }
```
JayShortway, line 238: Do we need `remember(progress)` for all of these? It would mean that whenever the progress changes (which is often), we would create new State instances. Would it not work with a regular `remember`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/SizeConstraintExtensions.kt
```diff
@@ -0,0 +1,15 @@
+@file:JvmSynthetic
+
+package com.revenuecat.purchases.ui.revenuecatui.extensions
+
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+
+@get:JvmSynthetic
+internal val SizeConstraint.dp: Dp?
```
JayShortway, line 10: Just a suggestion, but maybe this could be a bit more explicit like so:
```diff
- internal val SizeConstraint.dp: Dp?
+ internal fun SizeConstraint.dpOrNull(): Dp? 
```

This makes it analogous to what the stdlib does, e.g. with `firstOrNull()` on collections.

JayShortway (approved):
So sleek! 1 comment regarding `remember`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/SizeConstraintExtensions.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2105/commits/ae8e966c6ea59b4b0a7055e23d4ef943317fe6dc

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2105/commits/cb0717d692d460244ba76f282b3590b4f0678d05

vegaro (approved):

PR #2106, opened 2025-02-03
[Paywalls V2] Use original paywall fallback when trying to use Footer modes in a Components paywall

tonidero:
### Description
Until now, if trying to present a components paywall as a footer, it would display the paywall mostly full screen with some indeterminate behavior.

After this, if the developer tries to present a paywall v2 as a footer/condensed footer, we will fallback to the original paywall fallback paywall, which uses template 2 right now.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/strings/PaywallValidationErrorStrings.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/strings/PaywallValidationErrorStrings.kt
@@ -22,4 +22,6 @@ internal object PaywallValidationErrorStrings {
     const val MISSING_COLOR_ALIAS = "Aliased color '%s' does not exist."
     const val ALIASED_COLOR_IS_ALIAS = "Aliased color '%s' has an aliased value '%s', which is not allowed."
     const val MISSING_FONT_ALIAS = "Aliased font '%s' does not exist."
+    const val INVALID_MODE_FOR_COMPONENTS_PAYWALL =
+        "Components paywall does not support footer modes. Falling back to legacy fallback paywall."
 }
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/strings/PaywallValidationErrorStrings.kt
```diff
@@ -22,4 +22,6 @@ internal object PaywallValidationErrorStrings {
     const val MISSING_COLOR_ALIAS = "Aliased color '%s' does not exist."
     const val ALIASED_COLOR_IS_ALIAS = "Aliased color '%s' has an aliased value '%s', which is not allowed."
     const val MISSING_FONT_ALIAS = "Aliased font '%s' does not exist."
+    const val INVALID_MODE_FOR_COMPONENTS_PAYWALL =
+        "Components paywall does not support footer modes. Falling back to legacy fallback paywall."
```
JayShortway, line 26: I don't think we're using the word "components" in public text. Maybe something like this? 
```suggestion
        "Paywalls V2 does not support footer modes. Falling back to legacy fallback paywall."
```

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
```diff
@@ -572,6 +572,66 @@ class PaywallViewModelTest {
...
+
+        // Act
+        val model = create(offering = offering, mode = PaywallMode.FULL_SCREEN)
+
+        // Assert
+        assertThat(model.state.value).isInstanceOf(PaywallState.Loaded.Components::class.java)
+    }
+
+    @Test
+    fun `Should load fallback paywall if using components paywall in footer mode`(): Unit = runBlocking {
```
JayShortway, line 594: 🙌

JayShortway (approved):
Another one down! Just a copy suggestion.

PR #2108, opened 2025-02-04
[Paywalls V2] Support scroll orientation in StackComponent

tonidero:
### Description
Support scroll orientation in StackComponent

https://github.com/user-attachments/assets/7d36d912-42e6-4ec3-ac2a-96aeec2b966b


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/StackComponent.kt
```diff
@@ -49,8 +49,21 @@ class StackComponent(
     @get:JvmSynthetic
     val badge: Badge? = null,
     @get:JvmSynthetic
+    @SerialName("scroll_orientation")
+    val scrollOrientation: ScrollOrientation? = null,
```
tonidero, line 53: I'm not sure yet if this is how we want the schema to look... But adding this for now. Other properties that I was considering adding but elected not to for now: 
- `scrollInitialPadding` (I'm not sure if we want to allow starting the scroll in an intermediate point... Might be better to indicate the initialView instead? But then we need to have some sort of reference to a child component) 
-  `autoScrollSpeed` of some sort
- `loop` in case we want to loop (honestly for these last 2 I believe a carousel would probably work better)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -479,6 +493,9 @@ private fun MainStackComponent(
                         horizontalAlignment = dimension.alignment.toHorizontalAlignmentOrNull(),
                         verticalAlignment = dimension.alignment.toVerticalAlignmentOrNull(),
                     )
+                    .applyIfNotNull(scrollState, stackState.scrollOrientation) { state, orientation ->
+                        scrollable(state, orientation)
```
tonidero, line 497: MIght not want to support this for Zstacks... Thought of adding it just in case, but honestly, this would only be needed in cases of adding a nested stack that goes out of the bounds I guess...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/ModifierExtensions.kt
```diff
@@ -17,3 +17,11 @@ internal fun <T> Modifier.applyIfNotNull(value: T?, modifier: Modifier.(T) -> Mo
         this
     }
 }
+
+internal fun <T, V> Modifier.applyIfNotNull(value: T?, value2: V?, modifier: Modifier.(T, V) -> Modifier): Modifier {
```
tonidero, line 21: I've had this situation of wanting 2 things not to be null for a while... So I decided to add a new extension :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -451,10 +453,19 @@ private fun MainStackComponent(
 
     // Show the right container composable depending on the dimension.
     val stack: @Composable (Modifier) -> Unit = { rootModifier ->
+        val scrollState = if (stackState.scrollOrientation == null) {
+            null
+        } else {
+            rememberScrollState()
+        }
```
JayShortway, line 460: Nit, but we could write this like so: 
```suggestion
        val scrollState = stackState.scrollOrientation?.let { rememberScrollState() }
```

Imo that's more readable, but I realize that's very subjective so not a blocker 😇

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -617,6 +634,58 @@ private fun StackComponentView_Preview_Vertical() {
...
+            state = previewEmptyState(),
+            clickHandler = {},
+        )
+    }
+}
+
+@Suppress("MagicNumber")
+@Preview
+@Composable
+private fun StackComponentView_Preview_Scroll_Vertical() {
```
JayShortway, line 650: Maybe we can include both `VerticalStack` and `VerticalScroll` in this name?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -811,6 +883,58 @@ private fun StackComponentView_Preview_Horizontal() {
...
+            state = previewEmptyState(),
+            clickHandler = { },
+        )
+    }
+}
+
+@Suppress("MagicNumber")
+@Preview
+@Composable
+private fun StackComponentView_Preview_Scroll_Horizontal() {
```
JayShortway, line 899: Same here, maybe this name can include `HorizontalStack` and `HorizontalScroll`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/ModifierExtensions.kt
JayShortway: Nice! 😄

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/StackComponent.kt
JayShortway: cc @joshdholtz 

Just another idea as a variation on `scrollInitialPadding`, we could also do something like a `start`, `center`, `end` enum to control the initial position? Maybe name it `scrollInitialPosition` or `scrollInitialAlignment` or something.

JayShortway (approved):
Great stuff! Faster than the JSON haha 😄

JayShortway (approved):
LGTM!

tonidero:
Will merge this for now, and we can always change the schema in follow-up PRs if needed cc @joshdholtz

PR #2110, opened 2025-02-04
Fixes `X` in promo offer not dismissing the promo

vegaro:
Touching the `X` navigation button in the promotional offer closes the whole Customer Center, this PR modifies that behavior and adds a way to override what the X does. In the case of the promotional offer, it should call `dismissPromotionalOffer` which effectively calls the main action that the user triggered (like Cancel Purchase)

Also fixes another bug related to promotional offers, when the promo offer can't be loaded, it wasn't triggering the main path either, it was not doing anything.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
@@ -32,7 +32,6 @@ internal sealed class CustomerCenterState(
         @get:JvmSynthetic val promotionalOfferData: PromotionalOfferData? = null,
         @get:JvmSynthetic val title: String? = null,
         @get:JvmSynthetic override val navigationButtonType: NavigationButtonType = NavigationButtonType.CLOSE,
-        @get:JvmSynthetic val onNavigationOverride: (() -> Unit)? = null,
     ) : CustomerCenterState(navigationButtonType)
 }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
@@ -32,6 +32,7 @@ internal sealed class CustomerCenterState(
         @get:JvmSynthetic val promotionalOfferData: PromotionalOfferData? = null,
         @get:JvmSynthetic val title: String? = null,
         @get:JvmSynthetic override val navigationButtonType: NavigationButtonType = NavigationButtonType.CLOSE,
+        @get:JvmSynthetic val onNavigationOverride: (() -> Unit)? = null,
     ) : CustomerCenterState(navigationButtonType)
 }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
@@ -66,7 +66,8 @@ internal interface CustomerCenterViewModel {
         product: StoreProduct,
         promotionalOffer: CustomerCenterConfigData.HelpPath.PathDetail.PromotionalOffer,
         originalPath: CustomerCenterConfigData.HelpPath,
-    )
+        context: Context,
+    ): Boolean
 
     suspend fun onAcceptedPromotionalOffer(subscriptionOption: SubscriptionOption, activity: Activity?)
     fun dismissPromotionalOffer(originalPath: CustomerCenterConfigData.HelpPath, context: Context)
@@ -138,6 +139,7 @@ internal class CustomerCenterViewModelImpl(
                             product,
                             it.promotionalOffer!!,
                             path,
+                            context,
                         )
                     } else {
                         mainPathAction(path, context)
@@ -152,6 +154,7 @@ internal class CustomerCenterViewModelImpl(
                 product,
                 path.promotionalOffer!!,
                 path,
+                context,
             )
         } else {
             mainPathAction(path, context)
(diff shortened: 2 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -453,13 +466,15 @@ internal class CustomerCenterViewModelImpl(
...
+                    currentState.onNavigationOverride.invoke()
+                    currentState.copy(onNavigationOverride = null)
+                }
                 currentState is CustomerCenterState.Success &&
                     currentState.navigationButtonType == CustomerCenterState.NavigationButtonType.BACK -> {
                     currentState.resetToMainScreen()
                 }
-                currentState is CustomerCenterState.Success -> {
-                    CustomerCenterState.NotLoaded
-                }
```
vegaro, line 462: This was introduced in #2109 and I think it would be a regression

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -453,14 +466,16 @@ internal class CustomerCenterViewModelImpl(
...
-                else -> currentState
+                else -> CustomerCenterState.NotLoaded
```
vegaro, line 478: This was introduced in #2109 and I think it would be a regression

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -148,11 +153,15 @@ internal class CustomerCenterViewModelImpl(
         }
 
         if (product != null && path.promotionalOffer != null) {
-            loadAndDisplayPromotionalOffer(
+            val loaded = loadAndDisplayPromotionalOffer(
```
vegaro, line 156: Alternatively we could pass a lambda, but I think this works as well?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -407,12 +417,15 @@ internal class CustomerCenterViewModelImpl(
                             originalPath,
                             pricingPhasesDescription,
                         ),
+                        onNavigationOverride = { dismissPromotionalOffer(originalPath, context) },
```
vegaro, line 420: I don't think using the context in the lambda would be problematic?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
```diff
@@ -32,6 +32,7 @@ internal sealed class CustomerCenterState(
         @get:JvmSynthetic val promotionalOfferData: PromotionalOfferData? = null,
         @get:JvmSynthetic val title: String? = null,
         @get:JvmSynthetic override val navigationButtonType: NavigationButtonType = NavigationButtonType.CLOSE,
+        @get:JvmSynthetic val onNavigationOverride: (() -> Unit)? = null,
```
tonidero, line 35: Hmm so, IMO, if we're not using a navigation library, the VM should be in charge of holding the back stack. Then each Screen composable should use the "BackHandler" composable to customize the behavior upon a back button, then have the VM just update the state accordingly... I think having the state hold this is mixing the state with behavior, which should belong to the VM IMO.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: ok haha, I will give it another thought

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -451,16 +466,18 @@ internal class CustomerCenterViewModelImpl(
...
-            when {
-                currentState is CustomerCenterState.Success &&
-                    currentState.navigationButtonType == CustomerCenterState.NavigationButtonType.BACK -> {
-                    currentState.resetToMainScreen()
-                }
-                currentState is CustomerCenterState.Success -> {
-                    CustomerCenterState.NotLoaded
+        if (_state.value is CustomerCenterState.Success && onNavigationOverride != null) {
+            onNavigationOverride?.invoke()
+            onNavigationOverride = null
```
facumenzella, line 471: Is nulling out the variable a needed? I wasn't expecting that

facumenzella: Oh got it. This gets set when the promotional offer is opened, and once you close it, you null it out

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -412,7 +422,9 @@ internal class CustomerCenterViewModelImpl(
                     currentState
                 }
             }
+            return true
```
tonidero, line 425: So here we say we've loaded the promo offer, but it could be that we haven't in case the state is not `Success` right? It might be ok in that case though so 🤷

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: hmm you're right, I think I can return only when it's actually loaded

PR #2111, opened 2025-02-04
[Paywalls V2] Deprecate PaywallFooter in favor of OriginalTemplatePaywallFooter

tonidero:
### Description
This deprecates the existing Paywall Footer APIs with others to make it clearer that they won't work as expected with the new V2 paywalls. It additionally adds some docs about the `displayCloseButton` and `fontProvider` paywall options to indicate they will be ignored with paywalls V2.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
```diff
@@ -13,18 +13,45 @@ import androidx.compose.ui.Modifier
...
  * @param condensed Whether to condense the composable even more.
  * @param mainContent The content to display in the main area of the [Paywall]. We give a [PaddingValues] so you can
  * add that padding to your own content. This padding corresponds to the height of the rounded corner area of
  * the paywall.
  */
+@Deprecated(
+    "Use OriginalTemplatePaywallFooter instead",
+    ReplaceWith(
+        "OriginalTemplatePaywallFooter(options, condensed, mainContent)",
+    ),
```
JayShortway, line 28: Just checking, does this work without setting the [imports](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-replace-with/#1397414482%2FProperties%2F1895087430) property on `ReplaceWith`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
```diff
@@ -13,18 +13,45 @@ import androidx.compose.ui.Modifier
 /**
  * Composable offering a minified screen Paywall UI configured from the RevenueCat dashboard.
  * You can pass in your own Composables to be displayed in the main content area.
+ * Only compatible with original Paywalls. If using V2 paywalls, this will display a fallback paywall instead.
```
JayShortway, line 16: I noticed this was capitalized in `PaywallOptions.kt`.
```suggestion
 * Only compatible with original Paywalls. If using V2 Paywalls, this will display a fallback paywall instead.
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
```diff
@@ -13,18 +13,45 @@ import androidx.compose.ui.Modifier
...
     condensed: Boolean = false,
     mainContent: @Composable ((PaddingValues) -> Unit)? = null,
+) {
+    OriginalTemplatePaywallFooter(options, condensed, mainContent)
+}
+
+/**
+ * Composable offering a minified screen Paywall UI configured from the RevenueCat dashboard.
+ * You can pass in your own Composables to be displayed in the main content area.
+ * Only compatible with original Paywalls. If using V2 paywalls, this will display a fallback paywall instead.
```
JayShortway, line 42:
```suggestion
 * Only compatible with original Paywalls. If using V2 Paywalls, this will display a fallback paywall instead.
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/OriginalTemplatePaywallFooterView.kt
```diff
@@ -13,16 +13,45 @@ import com.revenuecat.purchases.Package
...
- * View that wraps the [PaywallFooter] Composable to display the Paywall Footer through XML layouts and the View system.
+ * View that wraps the [OriginalTemplatePaywallFooter] Composable to display the Paywall Footer
+ * through XML layouts and the View system.
  */
-open class PaywallFooterView : FrameLayout {
+@Deprecated(
+    "Use OriginalTemplatePaywallFooterView instead",
+    ReplaceWith(
+        "OriginalTemplatePaywallFooterView",
+    ),
```
JayShortway, line 30: Same comment regarding the imports property.

JayShortway (approved):
Very nice and thorough!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallFooter.kt
tonidero: From my tests in AS, this will automatically add the import when performed automatically with AS without having to actually add the imports... Maybe it's smart enough to automatically detect the class and perform the import if it's in the same file? I did notice It doesn't remove the old import... But I don't think that works even with the imports option.

JayShortway: Ah cool! Reading [this](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-replace-with/#681885954%2FProperties%2F1895087430): 
>The replacement expression is interpreted in the context of the symbol being used, and can reference members of enclosing classes **etc.**

Maybe "etc." means this works because the replacement is in the same file as the deprecated one? 😅

PR #2113, opened 2025-02-04
[Paywalls V2] Adds `TabsComponentStyle`

JayShortway:
## Description

Adds all the scaffolding to create `TabsComponentStyle`s and related `*ComponentStyle`s. A large part of the diff is tests.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedTabsPartial.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedTabsPartial.kt
@@ -1,0 +1,78 @@
+package com.revenuecat.purchases.ui.revenuecatui.components
+
+import com.revenuecat.purchases.ColorAlias
+import com.revenuecat.purchases.paywalls.components.PartialTabsComponent
+import com.revenuecat.purchases.paywalls.components.properties.ColorScheme
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.BorderStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.ColorStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.ShadowStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.toBorderStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.toColorStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.toShadowStyles
+import com.revenuecat.purchases.ui.revenuecatui.errors.PaywallValidationError
+import com.revenuecat.purchases.ui.revenuecatui.helpers.NonEmptyList
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Result
+import com.revenuecat.purchases.ui.revenuecatui.helpers.orSuccessfullyNull
+import com.revenuecat.purchases.ui.revenuecatui.helpers.zipOrAccumulate
+import dev.drewhamilton.poko.Poko
+
+@Poko
+internal class PresentedTabsPartial(
+    @get:JvmSynthetic val backgroundColorStyles: ColorStyles?,
+    @get:JvmSynthetic val borderStyles: BorderStyles?,
+    @get:JvmSynthetic val shadowStyles: ShadowStyles?,
+    @get:JvmSynthetic val partial: PartialTabsComponent,
+) : PresentedPartial<PresentedTabsPartial> {
+
+    companion object {
+        /**
+         * Creates a [PresentedTabsPartial] from the provided [PartialTabsComponent] and [aliases] map. If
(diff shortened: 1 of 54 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -81,36 +84,39 @@ internal class StyleFactory(
     fun create(
         component: PaywallComponent,
         rcPackage: Package? = null,
+        tabControl: TabControlStyle? = null,
```
tonidero, line 87: Hmm I'm likely missing it but do you mind explaining why we need to pass this here? Seems like we want to pass the tabControl to all the children of a tab recursively correct? Why is that needed? (Will review the follow-up PRs now which might help me understand that)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/TabsComponentStyle.kt
```diff
@@ -0,0 +1,82 @@
...
+) : ComponentStyle {
+
+    @Immutable
+    class Tab(@get:JvmSynthetic val stack: StackComponentStyle)
+}
+
+@Immutable
+internal sealed interface TabControlStyle : ComponentStyle {
+    @Immutable
+    class Buttons(@get:JvmSynthetic val stack: StackComponentStyle) : TabControlStyle {
```
tonidero, line 74: I must say that with the plural name here, I would expect to have a list of `StackComponentStyle` here instead of only one... (I know it comes from the component itself, so not sure if it would be a good idea to change that...)

tonidero (commented):
Still reviewing PRs but have some questions... Will hold until I finish reviewing the follow-up PRs.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
tonidero: Ahh I see, it's for the "selected" state... I do wonder if we need a selected state for tabs TBH... but if that's the spec it makes sense then 👍

JayShortway: Sorry I should have shared [the iOS PR](https://github.com/RevenueCat/purchases-ios/pull/4648), that also helped to make it click for me. 

If you look at the `TabsComponent` example JSON in that PR, this comment specifically: 
>// This is all that is needed for `tab_control`
>// The `control` property from above gets injected here in the view

That's what `tabControl` is for. It's to inject the already-created tab control into whatever view it appears in. To clarify: the `tabControl` element is always _inside_ the tab, but it doesn't have to be a direct child of the tab. It can be deeply nested.

tonidero: Ahhh thanks that helped 👍

tonidero (approved):
Just a couple comments about naming and using a data class... but looks good to me!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedTabsPartial.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.toColorStyles
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.toShadowStyles
+import com.revenuecat.purchases.ui.revenuecatui.errors.PaywallValidationError
+import com.revenuecat.purchases.ui.revenuecatui.helpers.NonEmptyList
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Result
+import com.revenuecat.purchases.ui.revenuecatui.helpers.orSuccessfullyNull
+import com.revenuecat.purchases.ui.revenuecatui.helpers.zipOrAccumulate
+import dev.drewhamilton.poko.Poko
+
+@Poko
```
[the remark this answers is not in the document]
JayShortway: Ah yes we can! Although every `Presented*Partial` is using `@Poko` now, so I rather do them all in 1 go in a consistency push.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/TabsComponentStyle.kt
JayShortway: It's indeed a bit confusing, but in a different way imo 😄 The thing is that this `stack` will contain as many buttons as there are tabs. So that's where the plurality comes from.

PR #2116, opened 2025-02-04
Ignores `RestorePurchasesDialog` previews using `IgnoreEmergeSnapshot`

vegaro:
There's an issue in Emerge Snapshots when previewing a Dialog https://github.com/EmergeTools/emerge-android/issues/136

They've suggested we ignore the snapshot for now. Added the dependency as `compileOnly` so it doesn't get included in the aar

### DIFF
--- a/ui/revenuecatui/build.gradle
+++ b/ui/revenuecatui/build.gradle
@@ -56,7 +56,7 @@ android {
 }
 
 dependencies {
-    compileOnly libs.emerge.snapshots.annotations
+    debugImplementation libs.emerge.snapshots.annotations
 
     implementation project(path: ':purchases')
 
--- a/ui/revenuecatui/build.gradle
+++ b/ui/revenuecatui/build.gradle
@@ -56,6 +56,8 @@ android {
 }
 
 dependencies {
+    compileOnly libs.emerge.snapshots.annotations
+
     implementation project(path: ':purchases')
 
     implementation(platform(libs.kotlin.bom))
--- a/ui/revenuecatui/build.gradle
+++ b/ui/revenuecatui/build.gradle
@@ -56,10 +56,10 @@ android {
 }
 
 dependencies {
-    debugImplementation libs.emerge.snapshots.annotations
 
     implementation project(path: ':purchases')
 
+    compileOnly libs.emerge.snapshots.annotations
     implementation(platform(libs.kotlin.bom))
     implementation libs.androidx.core
     implementation platform(libs.compose.bom)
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/build.gradle
```diff
@@ -56,6 +56,8 @@ android {
 }
 
 dependencies {
+    compileOnly libs.emerge.snapshots.annotations
```
vegaro, line 59: Does this make sense at all?

vegaro: Oh interesting, I stopped at "You'll need to include this as an `implementation` dependency in whatever soruce set you're intending to ignore snapshots." lol<br><br>Yeah, with debug, it won't be able to compile on release

vegaro: It says "Previews are only included in `debug` builds". Maybe it just magically works

vegaro: weird, as soon as I add it

```
Caused by: com.vanniktech.maven.publish.MissingVariantException: Invalid MavenPublish Configuration. Unable to find variant to publish named defaultsRelease. Try setting the 'androidVariantToPublish' property in the mavenPublish extension object to something that matches the variant that ought to be published.
```

tonidero: Ohh that's because we have a variant `defaults`... so we might need to do `defaultsDebugImplementation`?

vegaro: hmm wouldn't `debugImplemenation` cover all variants?

vegaro: I've thought of copying the annotation under the same package to our repo? Is that too crazy?

JayShortway: What's the error when using `compileOnly`? I think if we use `compileOnly` in `revenuecatui`, we should also add it as `debugImplementation` to the `testpurchasesuiandroidcompatibility` app.

vegaro: Doing that leads to

```
Caused by: com.vanniktech.maven.publish.MissingVariantException: Invalid MavenPublish Configuration. Unable to find variant to publish named defaultsRelease. Try setting the 'androidVariantToPublish' property in the mavenPublish extension object to something that matches the variant that ought to be published.
```

vegaro: I think that's just covering the real issue

JayShortway (approved):
Let's go! This will finally turn `main` green again 😄

tonidero (approved):

PR #2117, opened 2025-02-05
feat: Add support for tracking customer center events

facumenzella:
### Description
This is a follow-up of https://github.com/RevenueCat/purchases-android/pull/2096
- Rename `backend.postPaywallEvents` to `backend.postEvents`
- Add missing pieces for `CustomerCenterEvent`
- Add a coupe of tests to validate mixed tracking

### Next
- Track customer center events

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.common.events
 
 import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.customercenter.events.CustomerCenterEvent
 import com.revenuecat.purchases.paywalls.events.PaywallEvent
 import com.revenuecat.purchases.utils.Event
 import kotlinx.serialization.SerialName
@@ -70,6 +71,31 @@ internal fun PaywallEvent.toBackendStoredEvent(appUserID: String): BackendStored
             displayMode = data.displayMode,
             darkMode = data.darkMode,
             localeIdentifier = data.localeIdentifier,
+        ),
+    )
+}
+
+/**
+ * Converts a `PaywallEvent` into a `BackendStoredEvent.Paywalls` instance.
+ *
+ * @receiver The `PaywallEvent` to be converted.
+ * @param appUserID The user ID associated with the event.
+ * @return A `BackendStoredEvent.Paywalls` containing a `BackendEvent.Paywalls`.
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal fun CustomerCenterEvent.toBackendStoredEvent(appUserID: String): BackendStoredEvent {
+    return BackendStoredEvent.CustomerCenter(
+        BackendEvent.CustomerCenter(
+            id = creationData.id.toString(),
+            revisionID = data.revisionID,
+            type = data.type.value,
+            appUserID = appUserID,
+            appSessionID = data.sessionIdentifier.toString(),
+            timestamp = data.timestamp.time,
+            darkMode = data.darkMode,
+            locale = data.locale,
+            isSandbox = data.isSandbox,
+            displayMode = data.displayMode.value,
         ),
     )
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
@@ -8,6 +8,7 @@ import com.revenuecat.purchases.common.FileHelper
 import com.revenuecat.purchases.common.debugLog
 import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.verboseLog
+import com.revenuecat.purchases.customercenter.events.CustomerCenterEvent
 import com.revenuecat.purchases.identity.IdentityManager
 import com.revenuecat.purchases.paywalls.events.PaywallEvent
 import com.revenuecat.purchases.paywalls.events.PaywallStoredEvent
@@ -103,6 +104,7 @@ internal class EventsManager(
 
             val backendEvent = when (event) {
                 is PaywallEvent -> event.toBackendStoredEvent(identityManager.currentAppUserID)
+                is CustomerCenterEvent -> event.toBackendStoredEvent(identityManager.currentAppUserID)
                 else -> null
             }
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
@@ -1,7 +1,6 @@
 package com.revenuecat.purchases.customercenter.events
 
 import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
-import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
 /**
@@ -9,8 +8,7 @@ import kotlinx.serialization.Serializable
  */
 @Serializable
 @ExperimentalPreviewRevenueCatPurchasesAPI
-enum class CustomerCenterDisplayMode {
+enum class CustomerCenterDisplayMode(val value: String) {
 
-    @SerialName("full_screen")
-    FULL_SCREEN,
+    FULL_SCREEN(value = "full_screen"),
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
@@ -15,11 +15,10 @@ import java.util.UUID
  * Type representing a customer center event and associated data. Meant for RevenueCatUI use.
  */
 @ExperimentalPreviewRevenueCatPurchasesAPI
-@Poko
 @Serializable
-class CustomerCenterEvent(
+data class CustomerCenterEvent constructor(
     val creationData: CreationData,
-    val eventData: Data,
+    val data: Data,
 ) : FeatureEvent {
 
     companion object {
@@ -38,18 +37,19 @@ class CustomerCenterEvent(
     )
 
     @ExperimentalPreviewRevenueCatPurchasesAPI
-    @Poko
     @Serializable
     @SuppressWarnings("LongParameterList")
-    class Data(
+    data class Data(
         val type: CustomerCenterEventType,
         @Serializable(with = DateSerializer::class)
         val timestamp: Date,
+        @Serializable(with = UUIDSerializer::class)
+        val sessionIdentifier: UUID,
         val darkMode: Boolean,
         val locale: String,
         val isSandbox: Boolean,
         val version: Int = 1,
-        val revisionId: Int = 1,
+        val revisionID: Int = 1,
         val displayMode: CustomerCenterDisplayMode = CustomerCenterDisplayMode.FULL_SCREEN,
     )
 }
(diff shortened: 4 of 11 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
```diff
@@ -73,3 +74,28 @@ internal fun PaywallEvent.toBackendStoredEvent(appUserID: String): BackendStored
...
+
+/**
+ * Converts a `PaywallEvent` into a `BackendStoredEvent.Paywalls` instance.
+ *
+ * @receiver The `PaywallEvent` to be converted.
+ * @param appUserID The user ID associated with the event.
+ * @return A `BackendStoredEvent.Paywalls` containing a `BackendEvent.Paywalls`.
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal fun CustomerCenterEvent.toBackendStoredEvent(appUserID: String): BackendStoredEvent {
```
tonidero, line 86: We've started adding the `@JvmSynthetic` annotation to internal functions like this so they are not available in Java (internal is kind of ignored in Java :( ). We should add it here, and anywhere else moving forward.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -103,6 +104,7 @@ internal class EventsManager(
 
             val backendEvent = when (event) {
                 is PaywallEvent -> event.toBackendStoredEvent(identityManager.currentAppUserID)
+                is CustomerCenterEvent -> event.toBackendStoredEvent(identityManager.currentAppUserID)
```
tonidero, line 107: Hmm I'm wondering if the `toBackendStoredEvent` could be moved to the `FeatureEvent` interface, so we're forced to implement them and we don't need need to do this separation.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
```diff
@@ -1,16 +1,14 @@
...
-import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
 /**
  * Display mode for the Customer Center. Meant for RevenueCatUI use.
  */
 @Serializable
 @ExperimentalPreviewRevenueCatPurchasesAPI
-enum class CustomerCenterDisplayMode {
+enum class CustomerCenterDisplayMode(val value: String) {
```
tonidero, line 11: Hmm any reason to have it as a property instead of the serialized name?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -38,18 +37,19 @@ class CustomerCenterEvent(
     )
 
     @ExperimentalPreviewRevenueCatPurchasesAPI
-    @Poko
     @Serializable
     @SuppressWarnings("LongParameterList")
-    class Data(
+    data class Data(
```
tonidero, line 42: Same here about the data class vs Poko

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
```diff
@@ -73,3 +74,28 @@ internal fun PaywallEvent.toBackendStoredEvent(appUserID: String): BackendStored
...
+    return BackendStoredEvent.CustomerCenter(
+        BackendEvent.CustomerCenter(
+            id = creationData.id.toString(),
+            revisionID = data.revisionID,
+            type = data.type.value,
+            appUserID = appUserID,
+            appSessionID = data.sessionIdentifier.toString(),
```
vegaro, line 93: For context, `appSessionID` differs from Paywall's `sessionID` in that `appSessionID` is per app session. This is something we added for Customer Center and is an id that gets generated whenever the SDK is configured. The regular Session ID that Paywalls uses is used to track a paywalls session.

purchases/src/test/java/com/revenuecat/purchases/common/events/EventsManagerTest.kt
```diff
@@ -103,14 +123,14 @@ class EventsManagerTest {
     }
 
     @Test
-    fun `tracking events adds them to file`() {
-        eventsManager.track(event)
+    fun `tracking paywall adds them to file`() {
```
tonidero, line 126: Small fix
```suggestion
    fun `tracking paywall events adds them to file`() {
```

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
```diff
@@ -1,16 +1,14 @@
...
 
-    @SerialName("full_screen")
-    FULL_SCREEN,
+    FULL_SCREEN(value = "full_screen"),
```
vegaro, line 13: I think the value is not used at all?

tonidero (commented):
Mostly some questions but looking good!

purchases/src/test/java/com/revenuecat/purchases/common/events/EventsManagerTest.kt
vegaro, line 126:
```suggestion
    fun `tracking paywall event adds them to file`() {
```

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
vegaro: Love it

facumenzella: Thought about that while working on it. I'll apply it in this PR then.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
facumenzella: I saw that `PaywallEventType` was doing the same thing. Is there a clean way to access the `Serialized` value?

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -15,11 +15,10 @@ import java.util.UUID
  * Type representing a customer center event and associated data. Meant for RevenueCatUI use.
  */
 @ExperimentalPreviewRevenueCatPurchasesAPI
-@Poko
 @Serializable
-class CustomerCenterEvent(
+data class CustomerCenterEvent constructor(
```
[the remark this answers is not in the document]
facumenzella: I actually did this to not break a test. In the test `copy` is used to override just a value. Is there a clean way to do it without this?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
facumenzella: I'll update the docs

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
facumenzella: It's related to https://github.com/RevenueCat/purchases-android/pull/2117#discussion_r1943160575

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
facumenzella: One caveat is that we would need to make `BackendStoredEvent` public, which I am not sure we want to. Unless I can make something like a private protocol extension in swift? 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
tonidero: Not really, other than serializing it again... But ideally, our code doesn't need the serialized value except to send it to the backend/store it in the file/retrieve it from the file I guess?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
tonidero: Ohh right, I didn't think about that... I don't think you can add an extension to an interface like that :( 
I guess we could instead create a sealed interface with a private extension and a computed property to do this... But we are trying to avoid those sealed types since they can cause breaking changes if new types are added... 

So I actually think your current solution might be better then, to avoid exposing more API without need... Please ignore my comment 🙇

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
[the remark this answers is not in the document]
tonidero: Ah right, the `copy` constructor is automatically provided with data classes... I believe you can implement your own copy constructor in the test source set then, that just creates a new one by just changing the specific value? We can think about adding it to the main source set if it needs to be used there at some point.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
facumenzella: It's used here
```kotlin
    return BackendStoredEvent.CustomerCenter(
        BackendEvent.CustomerCenter(
            id = creationData.id.toString(),
            revisionID = data.revisionID,
            type = data.type.value,
            appUserID = appUserID,
            appSessionID = data.sessionIdentifier.toString(),
            timestamp = data.timestamp.time,
            darkMode = data.darkMode,
            locale = data.locale,
            isSandbox = data.isSandbox,
            displayMode = data.displayMode.value,
        ),
    )
```

Is there a way to access the serialized value like `.rawValue`? 🤔

facumenzella: I guess I could encode it / decode it instead. Let me try that out

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
facumenzella: Using SerialName now 🚀

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterDisplayMode.kt
facumenzella: Fixed, can you please take another look @tonidero

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
```diff
@@ -38,18 +38,43 @@ class CustomerCenterEvent(
...
         val locale: String,
         val isSandbox: Boolean,
         val version: Int = 1,
-        val revisionId: Int = 1,
+        val revisionID: Int = 1,
         val displayMode: CustomerCenterDisplayMode = CustomerCenterDisplayMode.FULL_SCREEN,
-    )
+    ) {
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        fun copy(
```
tonidero, line 58: We could make this `internal` I think? Since it's only used for tests

tonidero (approved):
Looks great! Just a comment, thanks!

tonidero:
One thing, I would probably not tag this PR with `pr:feat` but `pr:other`, since it doesn't actually add functionality from the perspective of SDK customers (customer center is still private)

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterEvent.kt
facumenzella:
```suggestion
        internal fun copy(
```

PR #2118, opened 2025-02-05
[Paywalls V2] Remove cursive as a generic font

joshdholtz:
### Motivation

Remove `cursive` as a generic font to match with what is possible with other platforms (iOS)

### Description

(Above)

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
@@ -44,7 +44,6 @@ internal sealed interface FontSpec {
         object SansSerif : Generic
         object Serif : Generic
         object Monospace : Generic
-        object Cursive : Generic
     }
 
     data class System(@get:JvmSynthetic val name: String) : FontSpec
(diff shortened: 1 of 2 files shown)


tonidero (approved):
LGTM!

JayShortway (commented):
I'm not entirely sure we should do this. Do we not want to hide this at the dashboard level? If someone wants to use a cursive font on both platforms, are we really going to make them add a cursive `.ttf` to their Android app even if that font is included in the system?

joshdholtz:
> I'm not entirely sure we should do this. Do we not want to hide this at the dashboard level? If someone wants to use a cursive font on both platforms, are we really going to make them add a cursive `.ttf` to their Android app even if that font is included in the system? Open to be convinced otherwise of course!

@JayShortway Your reasoning here makes a lot sense! I _think_ why I would lean towards removing it is for consistency between platforms and specifically the UX around how we would handle cursive on Android but not on iOS. We wouldn't be able to allow cursive as an option in the main font dropdown... or show it but then mention they need to add an ttf to use it on iOS. I think it would overall just lead to a kind of clunky feeling experience that would need additional docs for explaining why 🤷‍♂️ 

Thoughts? 😇

JayShortway:
@joshdholtz I can definitely get on board with consistency! And indeed, I would not show "cursive" in the default font dropdown on the dashboard. 

How would users add custom fonts to the dashboard? Is that 2 free-form text fields, 1 for Android and 1 for iOS (*)? And would there be any accompanying explanation there in the UI, saying something like: "to make this work, now add `.ttf`s with these exact names to your apps"? Or would the UX be completely different?

*: Or I guess a text field for each app within the project?

PR #2120, opened 2025-02-06
[Paywalls V2] New overrides structure

tonidero:
## Motivation

Migrating `overrides` from a structured object to an array of override objects

### Previously

Overrides were a very structure object that had all of the different types of states/conditions that allowed base properties of a component to be overridden. However, this was not flexible when it came to multiple being activated as it was missing for cases when multiples types of conditions were being evaluated.

### Now

Overrides is an array that has an array of `conditions` and `properties`. It will get applied in order from first to last IF all of the `conditions` are true.

This allows the FE/BE to determine the orders of overrides from lowest to higher priority so all of the SDKs/renderers don't need to duplicate that logic.

Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4705


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -0,0 +1,68 @@
...
+        COMPACT,
+        MEDIUM,
+        EXPANDED,
+        INTRO_OFFER,
+        MULTIPLE_INTRO_OFFERS,
+        SELECTED,
+        UNSUPPORTED,
+    }
+
+    private object ConditionSerializer : KSerializer<Condition> {
```
tonidero, line 36: I was a bit confused with this one... The backend right now returns an object with a `type` property but in iOS, we deserialize that directly to the `Condition` enum so I followed the same pattern... I'm not sure if we add more properties to that object, if we would be able to expose it with this API though...

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/common/ComponentOverridesTests.kt
```diff
@@ -9,14 +9,15 @@ import org.junit.Test
...
+import org.robolectric.ParameterizedRobolectricTestRunner
 
 @RunWith(Enclosed::class)
 internal class ComponentOverridesTests {
 
     // This tests deserialization of ComponentOverrides containing PartialTextComponent and PartialImageComponent, just
     // to make sure deserialization of generics works as expected.
 
-    @RunWith(Parameterized::class)
+    @RunWith(ParameterizedRobolectricTestRunner::class)
```
tonidero, line 20: I'm adding a test where a log gets called. It was failing with the junit runner.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -0,0 +1,68 @@
...
+            val composite = decoder.beginStructure(descriptor)
+            var result: Condition? = null
+
+            loop@ while (true) {
+                when (val index = composite.decodeElementIndex(descriptor)) {
+                    CompositeDecoder.DECODE_DONE -> break@loop
+                    0 -> {
+                        val typeValue = composite.decodeStringElement(descriptor, 0)
+                        result = try {
+                            Condition.valueOf(typeValue.uppercase())
```
tonidero, line 55: I didn't fully like this... But not sure of the best way to do this otherwise. Open to suggestions!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ToPresentedOverridesTests.kt
```diff
@@ -58,6 +56,33 @@ internal class ToPresentedOverridesTests(@Suppress("UNUSED_PARAMETER") name: Str
             medium to FontSpec.System("medium"),
             expanded to FontSpec.System("expanded"),
         )
+
+        private val defaultAvailableOverrides = listOf(
```
JayShortway, line 60: Good idea haha!

JayShortway (approved):
Very nice job! 💪 Just a comment on ignoring unknown JSON keys in `condition`, and some suggestions for tests.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedPartial.kt
```diff
@@ -128,77 +72,62 @@ internal fun <T : PartialComponent, P : PresentedPartial<P>> ComponentOverrides<
...
+            ComponentOverride.Condition.EXPANDED,
+            -> {
+                if (!windowSize.applicableConditions.contains(condition)) return false
+            }
+            ComponentOverride.Condition.MULTIPLE_INTRO_OFFERS -> {
+                if (introOfferEligibility != IntroOfferEligibility.MULTIPLE_OFFERS_ELIGIBLE) return false
+            }
+            ComponentOverride.Condition.INTRO_OFFER -> {
+                if (introOfferEligibility == IntroOfferEligibility.INELIGIBLE) return false
+            }
```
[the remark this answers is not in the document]
tonidero: Nope, for the condition processing, the order doesn't really matter. We basically iterate over all the conditions (in whatever order), and make sure they all apply. If not, we will continue with the next override in the list

purchases/src/test/java/com/revenuecat/purchases/paywalls/components/common/ComponentOverridesTests.kt
```diff
@@ -25,65 +26,103 @@ internal class ComponentOverridesTests {
...
         companion object {
 
             @Suppress("LongMethod")
             @JvmStatic
-            @Parameterized.Parameters(name = "{0}")
+            @ParameterizedRobolectricTestRunner.Parameters(name = "{0}")
             fun parameters(): Collection<*> = listOf(
                 arrayOf(
-                    "all values present",
+                    "all conditions present",
```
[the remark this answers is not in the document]
tonidero: Good point! 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedPartial.kt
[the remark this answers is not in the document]
JayShortway: Perfect, thanks for explaining!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
tonidero: Right, I already have tests deserializing all valid values (and one unknown) so we're probably ok 👍

PR #2114, opened 2025-02-04
[Paywalls V2] Adds `TabsComponentView`

JayShortway:
## Description

Adds all code required to render `TabsComponentView` and related Composables. Tests and samples in PaywallsTester will come in the next PR: https://github.com/RevenueCat/purchases-android/pull/2115 (Also see that PR for some visual examples.)

PR #2122, opened 2025-02-06
[Paywalls V2] Correctly selects packages on tabs

JayShortway:
## Description
This PR adds the required logic to:
- select the correct default package, on a per tab basis, and
- update the selected package correctly, when the tab changes.

It also handles various combinations of packages existing both on tabs and "globally" (outside any tabs). 

Now, I don't _love_ the implementation, as it traverses the tree again. However I think we can live with this for now, and look for optimizations later. (For instance, adding this logic to `StyleFactory` risks making that class very complex. So I'd rather do that carefully.)

## Diff guide
Main changes are in `PaywallState` and `OfferingToStateMapper`. There are a lot of tests added to `TabsComponentViewTests`, as this is quite tricky logic imo.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/TabsComponentStyle.kt
```diff
@@ -57,6 +57,8 @@ internal class TabsComponentStyle(
     @get:JvmSynthetic
     val shadow: ShadowStyles?,
     @get:JvmSynthetic
+    val control: TabControlStyle,
```
JayShortway, line 60: I mentioned that this is not needed in https://github.com/RevenueCat/purchases-android/pull/2114#discussion_r1941610937, but turns out it is haha.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -84,21 +85,60 @@ internal sealed interface PaywallState {
...
+            private val packages: AvailablePackages,
             initialLocaleList: LocaleList = LocaleList.current,
-            initialSelectedPackage: Package? = null,
-            initialSelectedTabIndex: Int = 0,
+            initialSelectedTabIndex: Int? = null,
         ) : Loaded {
 
+            data class AvailablePackages(
+                val packagesOutsideTabs: List<Info>,
+                val packagesByTab: Map<Int, List<Info>>,
```
tonidero, line 95: Hmm I'm wondering if we could just use `packagesByTab` and just assume `packagesOutsideTabs` to be "tab 0" in case there are no tabs... but then again, I guess there might be tabs without  packages... So I think this is probably the best 😅  

Just thinking whether there might be a better name than `packagesOutsideTabs`... Maybe `generalPackages` or `singleTierPackages`? Not sure TBH, maybe it's ok as is for now and we can revisit later.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -112,21 +152,36 @@ internal sealed interface PaywallState {
                 }
             }
 
-            var selectedTabIndex by mutableIntStateOf(initialSelectedTabIndex)
-                private set
-
-            val mostExpensivePricePerMonthMicros: Long? = offering.availablePackages.mostExpensivePricePerMonthMicros()
+            val mostExpensivePricePerMonthMicros by derivedStateOf {
```
tonidero, line 155: Hmm looking at this change I'm thinking that with tiered offers, we might want to separate this by tab... but I think that can come in a separate PR.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -370,3 +397,124 @@ private val PaywallComponentsData.defaultLocalization: Map<LocalizationKey, Loca
...
+        is ButtonComponentStyle -> add(parent.stackComponentStyle)
+        is PackageComponentStyle -> add(parent.stackComponentStyle)
+        is StickyFooterComponentStyle -> add(parent.stackComponentStyle)
+        is CarouselComponentStyle -> addAll(parent.slides)
+        is TabControlButtonComponentStyle -> add(parent.stack)
+        is TabControlStyle.Buttons -> add(parent.stack)
+        is TabControlStyle.Toggle -> add(parent.stack)
+        is TabsComponentStyle -> addAll(parent.tabs.map { it.stack })
+        is TimelineComponentStyle ->
+            addAll(parent.items.flatMap { item -> listOfNotNull(item.title, item.description, item.icon) })
```
tonidero, line 510: Hmm the timeline component won't have any packages... But I think this makes sense to keep these utility functions fully functional 👍

tonidero (approved):
Nice work!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
JayShortway: The intent here is to separate by tab haha. The value will change when the tab changes, and it will only take the currently-visible packages (on the current tab) into account. Is there anything I'm overlooking?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
JayShortway: >we could just use packagesByTab and just assume packagesOutsideTabs to be "tab 0" 

Yea I wanted to do that at first, but eventually I didn't because there can be packages on a tab and outside tabs at the same time.

>Just thinking whether there might be a better name than packagesOutsideTabs. 

I had been using "global packages" to refer to these, but I thought `packagesOutsideTabs` was more explicit.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
JayShortway: Yea agreed!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
tonidero: Ah right! Nvm, got confused with the sum of `packagesOutsideTabs` and the selected tab packages but I think this should be ok for the scenarios I'm thinking

PR #2123, opened 2025-02-06
Revamp UI in Customer Center

vegaro:
| Active | No Active |
|--------|--------|
| Screenshot 2025-02-06 at 21 11 37 | Screenshot 2025-02-06 at 21 11 16 |


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
```diff
@@ -281,10 +270,10 @@ private class SubscriptionInformationProvider : PreviewParameterProvider<Purchas
 }
 
 @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
-@Preview(group = "scale = 1", fontScale = 1F)
+@Preview(group = "scale = 1", fontScale = 1F, device = "spec:width=1080px,height=2720px,dpi=440")
 // Unrealistically long device to make the Column fit. Can be removed once Emerge Snapshot Test supports
 // @PreviewParameter.
-@Preview(group = "scale = 2", fontScale = 2F, device = "spec:width=1080px,height=4720px,dpi=440")
+@Preview(group = "scale = 2", fontScale = 2F, device = "spec:width=1080px,height=2720px,dpi=440")
```
vegaro, line 276: This was too long so I made it smaller

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
```diff
@@ -46,67 +48,93 @@ internal fun SubscriptionDetailsView(
...
+    icon: ImageVector? = null,
+    prominentSubtitle: Boolean = true,
+) {
+    val titleTextStyle = TextStyle(
+        fontSize = if (prominentSubtitle) 14.sp else 20.sp,
+        fontWeight = FontWeight.Normal,
+        color = if (prominentSubtitle) {
+            MaterialTheme.colorScheme.onSurfaceVariant
+        } else {
+            MaterialTheme.colorScheme.onSurface
```
vegaro, line 106: I am not super sure about using `onSurface` here, I wanted to use `onBackground` but there's no `onBackgroundVariant`. But maybe we can just use `onBackground` and apply an alpha

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
```diff
@@ -46,67 +48,93 @@ internal fun SubscriptionDetailsView(
...
+        } else {
+            MaterialTheme.colorScheme.onSurfaceVariant
+        },
+    )
+    Row(
+        verticalAlignment = Alignment.CenterVertically,
+        modifier = modifier
+            .defaultMinSize(minHeight = 60.dp)
+            .fillMaxWidth()
+            .padding(horizontal = 16.dp, vertical = 16.dp),
```
vegaro, line 124: I should probably extract these constants

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
...
-            modifier = Modifier.padding(bottom = 8.dp),
-        )
+                .fillMaxWidth()
+                .padding(PADDING_LARGE),
+            horizontalAlignment = Alignment.CenterHorizontally,
+        ) {
+            Icon(
+                imageVector = Icons.Rounded.Warning,
+                contentDescription = null,
+                tint = MaterialTheme.colorScheme.onSurface,
```
vegaro, line 171: Same here, unsure about this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
...
 ) {
-    Column(
-        modifier = modifier
-            .fillMaxWidth()
-            .padding(16.dp),
-        horizontalAlignment = Alignment.CenterHorizontally,
+    Surface(
+        modifier = modifier,
+        shape = MaterialTheme.shapes.medium,
+        color = MaterialTheme.colorScheme.surface,
```
vegaro, line 160: By default on new apps, this is not set. But I think the correct way is using `surface` here

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
...
+    showSectionHeader: Boolean = true,
     onButtonPress: (CustomerCenterConfigData.HelpPath) -> Unit,
-    useOutlinedButton: Boolean = false,
 ) {
+    val isDark = isSystemInDarkTheme()
+    val accentColor = appearance.getColorForTheme(isDark) { it.accentColor } ?: MaterialTheme.colorScheme.primary
+
     Column {
+        Spacer(modifier = Modifier.size(PADDING_LARGE))
+        if (showSectionHeader) {
```
vegaro, line 207: I thought it looked weird to show "Manage" in the NoActive case (there's nothing to manage), so I went with a divider

Screenshot 2025-02-06 at 21 17 55

Maybe we can find a word that's not "Manage" that works for both cases

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -45,82 +48,77 @@ internal fun ManageSubscriptionsView(
...
-            .fillMaxSize()
-            .padding(horizontal = 16.dp),
+        modifier = modifier.fillMaxSize(),
         contentAlignment = Alignment.TopCenter,
     ) {
         Column(
-            modifier = Modifier.fillMaxWidth(),
+            modifier = Modifier
+                .fillMaxWidth()
+                .verticalScroll(rememberScrollState()),
```
vegaro, line 63: I would like to move the title to a top app bar, but step by step

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
...
+                style = MaterialTheme.typography.headlineSmall.copy(fontWeight = FontWeight.Bold),
                 modifier = Modifier.padding(bottom = 8.dp),
             )
+
+            description?.let {
+                Text(
+                    text = it,
+                    style = MaterialTheme.typography.bodyLarge,
+                    textAlign = TextAlign.Center,
+                    modifier = Modifier.padding(vertical = PADDING_LARGE),
```
tonidero, line 186: One thing I was noticing but if we have a description here, we would have a bottom padding of `PADDING_LARGE` which is 16.dp, but if not, the title bottom padding would be 8.dp instead. For consistency, maybe we should apply the overall paddings to the column, and just apply a top padding to the description if it exists?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
tonidero: Maybe from the point of view of the user, it could be like a question as in "What do you want to do?". If we prefer to keep it into a single word, I think your compromise here makes sense and looks good 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -45,82 +48,77 @@ internal fun ManageSubscriptionsView(
...
     onAction: (CustomerCenterAction) -> Unit,
 ) {
     Box(
-        modifier = modifier
-            .fillMaxSize()
-            .padding(horizontal = 16.dp),
+        modifier = modifier.fillMaxSize(),
         contentAlignment = Alignment.TopCenter,
     ) {
         Column(
```
tonidero, line 60: Hmm do we need the outer `Box`? I think we could remove it and just add the `fillMaxSize` here?

tonidero (approved):
left some comments but nothing blocking. Looking great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
tonidero: I'm honestly having problems visualizing the different colors in the theme yeah... It's tricky 😅 I would try to avoid using as many as possible.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
JayShortway: I think we can make this separate previews now. This was a hack because Emerge didn't support `@PreviewParameter`, but it does now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
JayShortway: Hmm, but the `SubscriptionDetailsView` doesn't have a surface anymore right? If so, I wouldn't use `onSurface` colors at all I think.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
JayShortway: >I should probably extract these constants

They are already extracted, in this file. 😄 For instance, there's `PaddingContent` which is `16.dp`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/composables/CustomerCenterButton.kt
```diff
@@ -0,0 +1,51 @@
...
+    onClick: () -> Unit,
+) {
+    Row(
+        verticalAlignment = Alignment.CenterVertically,
+        modifier = modifier
+            .fillMaxWidth()
+            .defaultMinSize(minHeight = 60.dp)
+            .clickable { onClick() }
+            .padding(horizontal = 16.dp, vertical = 16.dp),
+        horizontalArrangement = Arrangement.spacedBy(8.dp),
```
JayShortway, line 34: Is this doing anything for a row with a single child?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/composables/CustomerCenterButton.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+
+@Composable
+@JvmSynthetic
+internal fun CustomerCenterButton(
```
JayShortway, line 22: Maybe this can be something like `SettingsButton`? `CustomerCenterButton` doesn't really explain what this is, or why this is a special button.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
JayShortway: Yea I also like the "no-surface" look here!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
 @Composable
 private fun NoActiveUserManagementView(
```
JayShortway, line 128: I'm not entirely sure about all the white space that exists on this screen, especially since it's all at the bottom of the screen. Have you played with centering the `CompatibilityContentUnavailableView`? (Not a blocker for this PR btw, just bringing it up.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
...
+            localization,
+            appearance,
+            showSectionHeader = false,
+            onButtonPress,
         )
     }
 }
 
 @Composable
 fun CompatibilityContentUnavailableView(
```
JayShortway, line 152: Should this be `private`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
JayShortway: "Resolve"? Not 100% convinced myself haha. Divider looks good too!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -129,106 +127,97 @@ private fun ActiveUserManagementView(
...
+        Spacer(modifier = Modifier.size(PADDING_LARGE))
+        if (showSectionHeader) {
+            Text(
+                localization.commonLocalizedString(CustomerCenterConfigData.Localization.CommonLocalizedString.MANAGE),
+                style = MaterialTheme.typography.bodyMedium,
+                color = accentColor,
+                modifier = Modifier.padding(horizontal = PADDING_LARGE, vertical = 8.dp),
+            )
+        } else {
+            HorizontalDivider()
```
JayShortway, line 215: I saw on a different screenshot that the divider is the full screen width. Maybe it can have the same horizontal padding as the rest of the content on the screen?

JayShortway (approved):
Such a big improvement, really cool! Just a few suggestions here and there, but nothing blocking.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/SubscriptionDetailsView.kt
vegaro: yes, it makes me twitch

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
vegaro: yeah I agree. Let's remove it

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
vegaro: I think I am going to add a divider in the other case as well. I think this word could be problematic and it's hard to make it compatible with the frontend we have for iOS as well, so I am thinking of removing it completely

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
vegaro: yes, I think it looks a bit weird with the current type of buttons we have. And as we discussed using the regular outlined button could make some people request we use the other layout, so for simplicity and consistency and to prevent folks from requesting it I decided to go this way

PR #2124, opened 2025-02-07
feat: Track IMPRESSION and SURVER_OPTION_CHOSEN for CustomerCenter

facumenzella:
### Motivation
Introduce event tracking for CustomerCenter

### Description
This PR starts tracking IMPRESSION and SURVER_OPTION_CHOSEN for CustomerCenter. 

Things to have in mind:
- I've removed serialization from CustomerCenter events because it was not used.
- I've removed `isSandbox` because we can't track that here
- I've moved `appSessionID` to events manager (similar to what's being done in iOS)

### Doubts
- I've added the locale, colorScheme and darkMode refresh methods BUT I wasn't sure if state needs to be updated. It does not seem to be used. Let me know if it should.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -70,6 +76,10 @@ internal fun InternalCustomerCenter(
         }
     }
 
+    LaunchedEffect(Unit) {
+        viewModel.trackImpression()
+    }
```
facumenzella, line 81: Not sure if this is the best way to do it. Let me know if there's another way

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -478,6 +493,47 @@ internal class CustomerCenterViewModelImpl(
...
+            shouldRefresh = true
+        }
+        if (_colorScheme.value != colorScheme) {
+            _colorScheme.value = colorScheme
+            shouldRefresh = true
+        }
+
+        if (shouldRefresh) {
+            _state.update { it }
+        }
```
facumenzella, line 517: I wanted to trigger a refresh, assuming this is how darkMode refresh works.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -478,6 +493,47 @@ internal class CustomerCenterViewModelImpl(
         }
     }
 
+    override fun refreshStateIfLocaleChanged() {
+        val currentLocaleList = getCurrentLocaleList()
+        if (_lastLocaleList.value != currentLocaleList) {
+            _lastLocaleList.value = currentLocaleList
+        }
+    }
```
facumenzella, line 501: Since this is only for tracking purposes, I assumed there's no need to do a state update

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -44,10 +45,12 @@ internal sealed class BackendEvent : Event {
         @SerialName("dark_mode")
         val darkMode: Boolean,
         val locale: String,
-        @SerialName("is_sandbox")
-        val isSandbox: Boolean,
         @SerialName("display_mode")
         val displayMode: CustomerCenterDisplayMode,
+        // only valid for survey option chosen
+        val path: CustomerCenterConfigData.HelpPath.PathType?,
+        val url: String?,
```
vegaro, line 52:
```        
        case surveyOptionID = "surveyOptionId"
        case surveyOptionTitleKey = "surveyOptionTitleKey"
        case additionalContext = "additionalContext"
```

Those are also missing

vegaro (changes requested):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
tonidero: This works most times, but could potentially be called multiple times if the activity is recreated (like in config changes like changing device rotation for example). So what we've done is to cache the data in the view model, since the view model survives activity recreations, and then you can call this method like `trackImpresionIfNeeded`, which will only track if the VM doesn't already have cached data.

I suggest looking at how this looks for paywalls: https://github.com/RevenueCat/purchases-android/blob/a07303d64091c856ee511f013fbe3f4dc280f0ea/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt#L177-L182

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -44,10 +45,12 @@ internal sealed class BackendEvent : Event {
         @SerialName("dark_mode")
         val darkMode: Boolean,
         val locale: String,
-        @SerialName("is_sandbox")
-        val isSandbox: Boolean,
```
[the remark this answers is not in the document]
facumenzella: Checked the backend, it's optional =)

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
tonidero: Just checking but what does `additionalContext` is supposed to contain? Just wondering since I noticed we didn't add it yet.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -27,6 +30,7 @@ import kotlinx.serialization.modules.polymorphic
  * @property postEvents Function for sending events to the backend.
  */
 internal class EventsManager(
+    private val appSessionID: UUID = UUID.randomUUID(),
```
tonidero, line 33: This may potentially be recreated if someone reconfigures the SDK. It's rare, but possible... Maybe it's ok to do it like this anyway since that's a pretty rare case. The alternative being storing this into a static variable.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterImpressionEvent.kt
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Type representing a customer center event and associated data. Meant for RevenueCatUI use.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@Poko
+class CustomerCenterImpressionEvent(
+    val creationData: CreationData = CreationData(),
+    val data: Data,
+) : FeatureEvent {
+    @ExperimentalPreviewRevenueCatPurchasesAPI
```
tonidero, line 18: NABD but this shouldn't be needed in nested classes of already `@ExperimentalPreviewRevenueCatPurchasesAPI`.

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterSurverOptionChosenEvent.kt
```diff
@@ -0,0 +1,54 @@
...
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@Poko
+class CustomerCenterSurverOptionChosenEvent(
+    val creationData: CreationData = CreationData(),
+    val data: Data,
+) : FeatureEvent {
+
+    companion object {
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal val json = Json.Default
```
tonidero, line 24: I believe this was mentioned in another PR, but ideally we reuse the same json serializer/deserializer accross the app. Otherwise we might forget some setting. For example, we might want to set `ignoreUnknownKeys = true` so we don't have any issues with unknown keys...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
tonidero: Right, as long as we don't need to update the state it makes sense... I guess with CustomerCenter we are not updating on locale changes anyway since we need to refetch the backend response so makes sense 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
tonidero: Hmm I believe this updates the state with itself, so it wouldn't actually do any change. We would need to add some property to the state that actually needs to be updated I think? (which might come in a separate PR). Otherwise this is not needed I believe.

tonidero (commented):
Looking good! Just some more comments

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
vegaro: It's for when we add a free text option. It is not required to add it right now since it will be always null, but it's in the event definition.

facumenzella:
Had a quick chat with @tonidero. I'll address the multiple json encoders / decoders in another PR, so we fix it once and for all.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
```diff
@@ -85,19 +88,57 @@ internal fun PaywallEvent.toBackendStoredEvent(appUserID: String): BackendStored
...
+        ),
+    )
+}
+
+/**
+ * Converts a `PaywallEvent` into a `BackendStoredEvent.Paywalls` instance.
+ *
+ * @receiver The `PaywallEvent` to be converted.
+ * @param appUserID The user ID associated with the event.
+ * @return A `BackendStoredEvent.Paywalls` containing a `BackendEvent.Paywalls`.
```
vegaro, line 119: let's update this docs and the ones of the above function

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/events/CustomerCenterSurverOptionChosenEvent.kt
```diff
@@ -0,0 +1,54 @@
...
+import kotlinx.serialization.json.Json
+import java.util.Date
+import java.util.UUID
+
+/**
+ * Type representing a customer center event and associated data. Meant for RevenueCatUI use.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@Poko
+class CustomerCenterSurverOptionChosenEvent(
```
vegaro, line 17:
```suggestion
class CustomerCenterSurveyOptionChosenEvent(
```

PR #2125, opened 2025-02-07
Extend paywall content below sticky footer

tonidero:
### Description
Currently the sticky footer and main paywall content are part of a column, so there is no overlap between them. This makes it so the paywall content can extend below the sticky footer in case the sticky footer has a transparent background.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedPaywallComponents.kt
```diff
@@ -64,23 +74,29 @@ internal fun LoadedPaywallComponents(
...
-            style = style,
-            state = state,
-            onClick = clickHandler,
-            modifier = Modifier
-                .fillMaxWidth()
-                .weight(1f)
-                .verticalScroll(rememberScrollState()),
-        )
+    Box(modifier = modifier.background(background)) {
+        var footerHeight by remember { mutableIntStateOf(0) }
```
tonidero, line 78: I thought about using SubcomposeLayout for this instead but this seemed easier for this use case? wdyt?

tonidero:
The preview changes in emerge are due to the recomposition caused by the `onSizeChange`, but it's actually centered within the same view.

tonidero:
We still want to do this but not for the open beta. Until then, leaving this as draft

PR #2126, opened 2025-02-07
[Paywalls V2] Fix stack clipping issues

tonidero:
### Description
A number of components clip their contents so they don't go out of bounds of the component itself. This is a problem for the `StackComponent` since sometimes we want shadows of children or other elements to extend beyond the bounds of the stack itself. An example of this, is with overlaid badges.

This PR removes the clipping of the stack component itself except when there is a Leading/Trailing EdgeToEdge badge or Nested badge, in which we actually need to clip.

| Before | After |
| ---- | ---- |
| image | image |
| image | image |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -4,6 +4,7 @@
 package com.revenuecat.purchases.ui.revenuecatui.components.stack
 
 import android.content.res.Configuration
+import androidx.compose.foundation.background
 import androidx.compose.foundation.gestures.Orientation
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -536,7 +533,10 @@ private fun MainStackComponent(
             )
         }
     } else if (overlay != null) {
-        Box(modifier = modifier.then(outerShapeModifier)) {
+        Box(modifier = modifier
+            .then(outerShapeModifier)
+            .clip(composeShape)
+        ) {
```
JayShortway, line 539: Might be nice to have a preview (or a few) of cases that are fixed by this PR? That way we can avoid a regression.

JayShortway (approved):
A fix is a fix! 💪 Just a suggestion for previews.

tonidero:
Note that a bunch of snapshots diffs changed with this change, which seem related to removing the clipping changing a few pixels , but the changes are minimal, so I think they are ok.

JayShortway:
I reran the tests. ("Gradle build daemon disappeared unexpectedly")

PR #2128, opened 2025-02-07
Add missing response fields to `CustomerInfo`. Including transaction Prices

vegaro:
This pull request introduces several enhancements and new features to the `SubscriptionInfo` and related classes in the `purchases` module. The most important changes include adding new properties to the `SubscriptionInfo` class, creating a new `PriceResponse` class, and updating the `SubscriptionInfoResponse` class to include additional fields.

### Enhancements to `SubscriptionInfo` and related classes:

* `purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt`:
  - Added new properties `autoResumeDate`, `displayName`, `price`, and `productPlanIdentifier` to the `SubscriptionInfo` class. [[1]](diffhunk://#diff-52e3fe0a40aa5cfc833e7bf697a42177c2cd46b8a28719f632ff217ca331056fR78-R81) [[2]](diffhunk://#diff-52e3fe0a40aa5cfc833e7bf697a42177c2cd46b8a28719f632ff217ca331056fR144-R147)
  - Added a `locale` parameter to the constructor of `SubscriptionInfo`, for building the price

* `purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/PriceResponse.kt`:
  - Created a new `PriceResponse` class to handle price-related information.

* `purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/SubscriptionInfoResponse.kt`:
  - Updated the `SubscriptionInfoResponse` class to include `autoResumeDate`, `displayName`, `price`, and `productPlanIdentifier` fields. [[1]](diffhunk://#diff-e0bbc4a1b6881dd39cc8730db5b5ad8d581ff38a760b624a8203a211cef2d08fR38-R45) [[2]](diffhunk://#diff-e0bbc4a1b6881dd39cc8730db5b5ad8d581ff38a760b624a8203a211cef2d08fR63-R66)

* `purchases/src/main/kotlin/com/revenuecat/purchases/models/Transaction.kt`:
  - Added new properties `displayName`, `isSandbox`, `originalPurchaseDate`, and `price` to the `Transaction` class. [[1]](diffhunk://#diff-af5fd33181c0405ff56deb5891af1ef213ea76fbad00f8af1fad3e33d9aaea7bR28-R31) [[2]](diffhunk://#diff-af5fd33181c0405ff56deb5891af1ef213ea76fbad00f8af1fad3e33d9aaea7bR44-R49)

### Additional updates:

* `purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceExtensions.kt`:
  - Changed the visibility of `MICRO_MULTIPLIER` to `internal`.


purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt
```diff
@@ -73,6 +75,22 @@ class SubscriptionInfo(
...
      */
     val storeTransactionId: String?,
+    /**
+     * Date when the subscription will auto-resume.
+     */
+    val autoResumeDate: Date?,
+    /**
+     * The display name of the subscription.
+     */
+    val displayName: String?,
```
facumenzella, line 85: This is the one set in the dashboard right?

purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/PriceResponse.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.revenuecat.purchases.models.Price
+import com.revenuecat.purchases.utils.MICRO_MULTIPLIER
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import java.text.NumberFormat
+import java.util.Currency
+import java.util.Locale
+
+@Serializable
+internal class PriceResponse(
```
facumenzella, line 14: Shouldn't this be a `data class` instead?

purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt
```diff
@@ -74,6 +76,22 @@ class SubscriptionInfo(
...
+    /**
+     * Paid price for the subscription.
+     */
+    val price: Price?,
+    /**
+     * The identifier of the product plan.
+     */
+    val productPlanIdentifier: String?,
```
tonidero, line 94: To avoid the breaking change, we should add a constructor without all these fields defaulting to null and deprecate it. Alternatively, we could default all of them to `null` and add a `@JvmOverloads`, though I think that would still break for non-named arguments since we have a non-optional parameter after this (`requestDate`, though we could always move all these to after that one I guess)

purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt
tonidero: Right, it would be great to dissambiguate this somehow... Maybe could be `rcDisplayName`? Not sure TBH 😅

purchases/src/main/kotlin/com/revenuecat/purchases/models/Transaction.kt
```diff
@@ -24,9 +29,24 @@ data class Transaction(
     val purchaseDate: Date,
     val storeTransactionId: String?,
     val store: Store,
+    val displayName: String?,
+    val isSandbox: Boolean = false,
+    val originalPurchaseDate: Date?,
+    val price: Price?,
```
tonidero, line 35: Same, some of these could be breaking changes, so we might need to add a constructor without these fields to keep backwards compatibility.

purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt
```diff
@@ -74,6 +76,22 @@ class SubscriptionInfo(
      * The transaction id in the store of the subscription.
      */
     val storeTransactionId: String?,
+    /**
+     * Date when the subscription will auto-resume.
```
tonidero, line 80: Maybe add more details that it will only have a value if the subscription is paused and that it's for Google subs only (I think?)

tonidero (commented):
Mostly just some comments to avoid the breaking changes but looking good!

purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt
vegaro: Maybe we can just make it more clear in the docs?

tonidero: Yeah, rethinking about this, I think it's fine 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/responses/PriceResponse.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.revenuecat.purchases.models.Price
+import com.revenuecat.purchases.utils.MICRO_MULTIPLIER
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import java.text.NumberFormat
+import java.util.Currency
+import java.util.Locale
+
+@Serializable
+internal data class PriceResponse(
```
tonidero, line 14: Hmm I wondered if it would make sense to make this a nested class inside the `SubscriptionInfoResponse` to avoid increasing the top level API surface... But I think it's fine as this as well.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceExtensions.kt
```diff
@@ -7,7 +7,7 @@ import java.text.NumberFormat
 import java.util.Currency
 import java.util.Locale
 
-private const val MICRO_MULTIPLIER = 1_000_000.0
+internal const val MICRO_MULTIPLIER = 1_000_000.0
```
tonidero, line 10: Hmm should we move this to `com.revenuecat.purchases.common.Constants`?

vegaro:
checking the tests now

purchases/src/test/java/com/revenuecat/purchases/utils/Responses.kt
```diff
@@ -106,7 +136,12 @@ object Responses {
                         "period_type": "normal",
                         "ownership_type": "PURCHASED",
                         "store_transaction_id" : "GPA.3394-7009-4518-41945..8",
-                        "expires_date": "${Iso8601Utils.format(threeMonthFreeTrialExpirationDate)}"
+                        "expires_date": "${Iso8601Utils.format(threeMonthFreeTrialExpirationDate)}",
+                        "display_name": "Basic Monthly",
+                        "price": {
+                          "currency": "USD",
+                          "amount": "5.99"
+                        }
```
tonidero, line 144: Should we have some sample response with `auto_resume_date` and `product_plan_identifier`?

tonidero (approved):
Left some comments but nothing really blocking. Thanks for adding this!

tonidero (approved):
Thanks for doing those changes! :shipit:

PR #2129, opened 2025-02-10
[Paywalls V2] Actually ignores the font provider for V2.

JayShortway:
## Motivation
Despite what we say in the updated docs, I noticed that the font provider wasn't ignored at all for V2 paywalls. If there's no custom font set on the dashboard, the font provider still sets the font. This PR corrects that.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -62,12 +62,13 @@ internal fun InternalPaywall(
     BackHandler {
```
JayShortway, line 62: Hiding whitespace in this file helps with reviewing the actual changes.

tonidero (approved):
Oh right, nice catch!

PR #2130, opened 2025-02-10
[Paywalls V2] Avoids a crash if a package is missing

JayShortway:
## Description 
Avoids a crash and returns a validation error instead if a package is missing. Also uses the correct error message 😅

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
@@ -133,12 +133,13 @@ internal class StyleFactory(
         component: PackageComponent,
         tabControl: TabControlStyle?,
     ): Result<PackageComponentStyle, NonEmptyList<PaywallValidationError>> =
-        offering.getPackage(component.packageId)
+        offering.getPackageOrNull(component.packageId)
             .errorIfNull(
                 nonEmptyListOf(
                     PaywallValidationError.MissingPackage(
                         offeringId = offering.identifier,
-                        packageId = component.packageId,
+                        missingPackageId = component.packageId,
+                        allPackageIds = offering.availablePackages.map { it.identifier },
                     ),
                 ),
             ).flatMap { rcPackage ->
@@ -564,4 +565,12 @@ internal class StyleFactory(
             .orSuccessfullyNull()
             // Ensure the default source keyed by the default locale is present in the result.
             .map { nonEmptyMapOf(localizations.entry.key to this, it.orEmpty()) }
+
+    private fun Offering.getPackageOrNull(identifier: String): Package? =
+        @Suppress("SwallowedException")
+        try {
+            getPackage(identifier)
+        } catch (e: NoSuchElementException) {
+            null
+        }
 }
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -564,4 +565,12 @@ internal class StyleFactory(
             .orSuccessfullyNull()
             // Ensure the default source keyed by the default locale is present in the result.
             .map { nonEmptyMapOf(localizations.entry.key to this, it.orEmpty()) }
+
+    private fun Offering.getPackageOrNull(identifier: String): Package? =
+        @Suppress("SwallowedException")
+        try {
+            getPackage(identifier)
+        } catch (e: NoSuchElementException) {
```
tonidero, line 573: I think you can change to use `_` and it won't complain about the `SwallowedException`?
```suggestion
        } catch (_: NoSuchElementException) {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/strings/PaywallValidationErrorStrings.kt
```diff
@@ -20,7 +20,7 @@ internal object PaywallValidationErrorStrings {
     const val ALL_VARIABLE_LOCALIZATIONS_MISSING_FOR_LOCALE =
         "All variable localizations for locale '%s' are missing."
     const val MISSING_PACKAGE =
-        "Offering with id '%s' does not have a package with id '%s'."
+        "Offering with id '%s' does not have a package with id '%s'. It has these packages instead: [%s]."
```
tonidero, line 23: ❤️

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -564,4 +565,12 @@ internal class StyleFactory(
             .orSuccessfullyNull()
             // Ensure the default source keyed by the default locale is present in the result.
             .map { nonEmptyMapOf(localizations.entry.key to this, it.orEmpty()) }
+
+    private fun Offering.getPackageOrNull(identifier: String): Package? =
```
tonidero, line 569: I wonder if we might want to add this to the actual `Offering` class... But this looks good for now 👍

tonidero (approved):
Thanks for fixing this! 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JayShortway: TIL, thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JayShortway: Yea could be nice indeed!

PR #2131, opened 2025-02-10
[Paywalls V2] Support background images in StackComponent, CarouselComponent, TabsComponent

tonidero:
### Description
This adds support for the existing `Background` property to some extra component so we can add images as backgrounds for them.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -22,6 +22,7 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.RectangleShape
 import androidx.compose.ui.graphics.toArgb
+import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.layout.Placeable
 import androidx.compose.ui.layout.SubcomposeLayout
 import androidx.compose.ui.layout.layout
@@ -38,12 +39,14 @@ import com.revenuecat.purchases.paywalls.components.properties.CornerRadiuses
 import com.revenuecat.purchases.paywalls.components.properties.Dimension
 import com.revenuecat.purchases.paywalls.components.properties.FlexDistribution
 import com.revenuecat.purchases.paywalls.components.properties.HorizontalAlignment
+import com.revenuecat.purchases.paywalls.components.properties.ImageUrls
 import com.revenuecat.purchases.paywalls.components.properties.Padding
 import com.revenuecat.purchases.paywalls.components.properties.Shape
 import com.revenuecat.purchases.paywalls.components.properties.Size
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fill
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fit
 import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint.Fixed
+import com.revenuecat.purchases.paywalls.components.properties.ThemeImageUrls
 import com.revenuecat.purchases.paywalls.components.properties.TwoDimensionalAlignment
 import com.revenuecat.purchases.paywalls.components.properties.VerticalAlignment
 import com.revenuecat.purchases.ui.revenuecatui.components.ComponentView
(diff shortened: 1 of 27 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -497,16 +497,16 @@ private fun MainStackComponent(
         }
     }
 
-    val backgroundColorStyle = stackState.backgroundColor?.forCurrentTheme
+    val backgroundStyle = stackState.background?.let { rememberBackgroundStyle(background = it) }
```
tonidero, line 500: When testing the `Background` modifier that accepts a `BackgroundStyle` in these new components, it seems that there are some sizing issues until the image has downloaded from web, causing some weird reloading. I mean to test it more thoroughly in case we can do something to the background modifier... Otherwise, we might need to treat the image background differently in order to avoid the resizing that's currently happening...

tonidero:
While I still need to figure out those resizing issues, I think this is mostly ready for review.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -556,6 +559,16 @@ internal class StyleFactory(
...
             .map { stack -> TabsComponentStyle.Tab(stack) }
 
+    private fun createBackgroundStyles(
+        background: Background?,
+        backgroundColor: ColorScheme?,
+    ): Result<BackgroundStyles?, NonEmptyList<PaywallValidationError>> =
+        background?.toBackgroundStyles(colorAliases)
+            ?: (
+                backgroundColor?.toColorStyles(colorAliases)?.map { color -> BackgroundStyles.Color(color) }
+                    ?: Result.Success(null)
```
JayShortway, line 569: We could use `orSuccessfullyNull()` here I think.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/tabs/TabsComponentView.kt
```diff
@@ -77,7 +78,7 @@ internal fun TabsComponentView(
     )
     if (!tabsState.visible) return
 
-    val backgroundColorStyle = tabsState.backgroundColor?.forCurrentTheme
+    val backgroundColorStyle = tabsState.background?.let { rememberBackgroundStyle(it) }
```
JayShortway, line 81: Suggestion: 
```diff
- val backgroundColorStyle = tabsState.background?.let { rememberBackgroundStyle(it) }
+ val background = tabsState.background?.let { rememberBackgroundStyle(it) }
```

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactoryTests.kt
```diff
@@ -133,6 +135,49 @@ class StyleFactoryTests {
...
+        assertThat(result).isInstanceOf(Result.Success::class.java)
+        val style = (result as Result.Success).value as StackComponentStyle
+        with(style.background as BackgroundStyles.Color) {
+            val colorStyle = color.light as ColorStyle.Solid
+            assertThat(colorStyle.color).isEqualTo(Color.Red)
+        }
+    }
+
+    @Test
+    fun `Should create a StackComponentStyle where background takes preference over backgroundColor`() {
```
JayShortway, line 160: ❤️

JayShortway (approved):
Nice job! Good idea to get those extra tests in too!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: So the issue with the sizing still is happening on the preview, where the stack takes the whole available space. However, it seems it gets fixed by following the workaround in https://github.com/coil-kt/coil/issues/1910. Fixed in https://github.com/RevenueCat/purchases-android/pull/2131/commits/d71b22426700eea873a5dced478a7946cf7c42cb

tonidero: Another update, looks like that change didn't actually fix the issue, there is still resizing noticeable in slow connections. It's not the end of the world since it only happens the first time we're loading the background images, but still something we should fix.

PR #2135, opened 2025-02-11
fix: Hide contact support if there's no email

facumenzella:
### Description
Hide the contact support button if the email is not set in CustomerCenter

Screenshot 2025-02-11 at 12 37 14


tonidero (approved):
LG! Thanks for fixing!

PR #2136, opened 2025-02-11
[Paywalls V2] Make root stack component use all the available space by default

tonidero:
### Description
Currently the backend sends the root stack component as having a width = Fill, height = Fit, which is not fitting correctly with the way the main layout in Android looks like. We believe the root component should take all available space, so for now, this introduces a quick hack that makes sure the paywall uses that space. This might change in the future depending on how we want to support root stack sizes, but this looks how it should in android with this change.


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallComponentValidationTests.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.components.style.StackComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.extensions.validatePaywallComponentsDataOrNull
+import com.revenuecat.purchases.ui.revenuecatui.helpers.UiConfig
+import com.revenuecat.purchases.ui.revenuecatui.helpers.getOrThrow
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.net.URL
+
+@RunWith(AndroidJUnit4::class)
+internal class PaywallComponentValidationTests {
```
JayShortway, line 29: We have an existing `PaywallComponentDataValidationTests` file. Maybe this test should go there?

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallComponentValidationTests.kt
```diff
@@ -0,0 +1,76 @@
...
+            metadata = emptyMap(),
+            availablePackages = emptyList(),
+            paywallComponents = Offering.PaywallComponents(UiConfig(), paywallComponents),
+        )
+    }
+
+    // This tests a temporal hack to make the root component fill the screen. This will be removed once we have a
+    // definite solution for positioning the root component.
+    @Test
+    fun `Should use Fill size in root component even if given Fit`() {
```
JayShortway, line 71: 🙌

PR #2137, opened 2025-02-11
[Paywalls V2] Adds edge-to-edge support

JayShortway:
## Description
This is the Android equivalent of this iOS PR: https://github.com/RevenueCat/purchases-ios/pull/4744.

It follows these rules:  
- For bottom system bars padding:
    - Has a sticky footer? Apply to the sticky footer.
    - No sticky footer? Apply to the root.
- For top system bars padding:
    - Has a hero image? Apply to the hero image's parent, and ignore this padding for the hero image itself.
    - No hero image? Apply to the root.

To see a paywall edge-to-edge in PaywallsTester, click "Display paywall as view in an activity" on the bottom-right tab. 

Here as well, it would be nice to improve this in a future iteration to gather this info in the `StyleFactory` directly.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -518,3 +584,56 @@ private fun ArrayDeque<ComponentStyle>.addChildrenOf(parent: ComponentStyle) {
...
     }
 }
+
+@Suppress("UNCHECKED_CAST")
+private fun <T : ComponentStyle> T.map(
+    transform: (ComponentStyle) -> ComponentStyle,
+): T {
+    return when (val transformed = transform(this)) {
+        is StackComponentStyle -> transformed.copy(
+            children = transformed.children.map { it.map(transform) },
```
tonidero, line 594: Hmm considering this will be recursive, what do you think of calling this method `deepMap` or `deepTransform`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -97,6 +97,12 @@ internal class StackComponentState(
     @get:JvmSynthetic
     val children = style.children
 
+    @get:JvmSynthetic
+    val applyTopWindowInsets = style.applyTopWindowInsets
```
tonidero, line 101: Wdyt of adding an option to paywall tester to display the paywall as an edge to edge paywall? That way we would be able to more easily test this scenario. Right now I think it would be tricky to test... I'm imagining another button when launching the paywall here:
image

Maybe that's tricky and better to do in a separate PR though...

tonidero (approved):
Code looks good to me! It would be great to be able to test it more easily though... Since previews would be trickier, maybe we can make it easier to test with paywall tester?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
JayShortway: Ugh, I could have mentioned this in the PR description.. but "Display paywall as view in an activity" shows it as edge to edge!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
JayShortway: Yea good idea to make it more explicit! I went with `recursiveMap`. Let me know if you hate it!

PR #2138, opened 2025-02-11
fix: Use TextButton instead of outline for Promotionals

facumenzella:
### Description
The OutlineButton does not offer a way of customizing it.
This PR replaces OutlineButton with TextButton, and uses localization instead of a hardcoded string.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferScreen.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferScreen.kt
@@ -90,6 +90,9 @@ internal fun PromotionalOfferScreen(
 
         TextButton(
             onClick = { onDismiss() },
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(bottom = 16.dp),
         ) {
             Text(
                 text = localization.commonLocalizedString(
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferScreen.kt
```diff
@@ -85,17 +87,16 @@ internal fun PromotionalOfferScreen(
         ) {
             Text(promotionalOfferData.localizedPricingPhasesDescription)
         }
-        OutlinedButton(
+
+        TextButton(
             onClick = { onDismiss() },
-            modifier = Modifier
-                .fillMaxWidth()
-                .padding(bottom = 16.dp),
```
tonidero, line 92: Do we need to add some extra padding to the button? Looking at the diff in emerge, seems that it's too close to the bottom IMO?

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PromotionalOfferScreen.kt
facumenzella: You're right, I'll put it back

vegaro (approved):

PR #2140, opened 2025-02-11
feat: Filter CANCEL from help path if lifetime

facumenzella:
### Motivation
CANCEL should not be shown if the product is a lifetime product.

### Description
This PR follows the same logic as Android, and filters out CANCEL if the product is lifetime:
1. Check entitlement
2. Subscription is false
3. Non subscription is true

See https://github.com/RevenueCat/purchases-ios/pull/4755


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
vegaro, line 1: Just noticed this file doesn't have the `JvmSynthetic` annotations. Can you add them in this PR 🙏 ?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -65,6 +65,7 @@ internal fun ManageSubscriptionsView(
         } else {
             NoActiveUserManagementView(
                 screen,
+                null,
```
vegaro, line 68: let's be specific on the parameter here `purchaseInformation =`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -172,11 +179,18 @@ private fun ContentUnavailableView(
...
 private fun ManageSubscriptionsButtonsView(
     screen: CustomerCenterConfigData.Screen,
+    purchaseInformation: PurchaseInformation?,
     onButtonPress: (CustomerCenterConfigData.HelpPath) -> Unit,
 ) {
+    val filteredPaths = if (purchaseInformation?.isLifetime == true) {
+        screen.supportedPaths.filter { it.type != CustomerCenterConfigData.HelpPath.PathType.CANCEL }
+    } else {
+        screen.supportedPaths
+    }
```
vegaro, line 189: hmm, this is the View, it feels like this logic should live in the view model instead

facumenzella: Discussed offline
I'll introduce something else (also good for testing), keeping in mind having only one view model

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -261,3 +270,12 @@ private fun NoActiveSubscriptionsViewNoDescription_Preview() {
...
 }
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal fun PurchaseInformation.isSupportedPath(pathType: PathType): Boolean {
+    return if (isLifetime) {
+        pathType != CustomerCenterConfigData.HelpPath.PathType.CANCEL
+    } else {
+        true
+    }
+}
```
vegaro, line 281: I would have this directly as a function in `PurchaseInformation`? Even there it feels like `PurchaseInformation` should just be holding data and not doing any logic. Maybe we can have `supportedPaths` as a parameter of `ManageSubscriptionsView`? 

It could be part of `CustomerCenterState.Success` and be calculated at `loadCustomerCenter` after loading the `purchaseInformation` in the `viewModel`

That way we don't have to do filtering in the view (`screen.supportedPaths.filter { isSupportedPath(it.type) }`), and the filtering logic stays in the view model which is easier to test. Right now we are not testing that the filtering is actually happening.

facumenzella:
I think the snapshot makes sense @vegaro , right?

PR #2142, opened 2025-02-11
[Paywalls V2] Makes linear gradients consistent with CSS

JayShortway:
## Description
This PR does 2 things:
- Ensures the angle is consistent, even if the area is not square.
- Ensures the interpretation of the angle is consistent with CSS' [`linear-gradient`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient). It's not a 100% 1 to 1, but it's close I think. 

| CSS | purchases-android |
|-----|--------------------|
| image | image |
| image | image |
| image | image |


tonidero:
Looks like a bunch of screenshots changed... Most of them made sense, but we might want to fix some of them (looking at the timeline component gradients for example 😅

JayShortway:
@tonidero Right, good call haha! I adjusted the angle in the most severe cases here: https://github.com/RevenueCat/purchases-android/pull/2142/commits/2c2d2bc72df78ac9997afe9bd2cf43629fd0df6b. That should reduce the diff.

PR #2143, opened 2025-02-11
[Paywalls V2] Predownload paywall low res images

tonidero:
### Description
This prefetches the images for paywalls V2 for the `current` offering. It includes background images, icon images and ImageComponent images.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
@@ -69,6 +69,7 @@ data class Offerings internal constructor(
     )
 }
 
+@OptIn(InternalRevenueCatAPI::class)
 internal fun Offering.withPresentedContext(placementId: String?, targeting: Offerings.Targeting?): Offering {
     val updatedAvailablePackages = this.availablePackages.map {
         val oldContext = it.presentedOfferingContext
@@ -95,5 +96,6 @@ internal fun Offering.withPresentedContext(placementId: String?, targeting: Offe
         metadata = this.metadata,
         availablePackages = updatedAvailablePackages,
         paywall = this.paywall,
+        paywallComponents = this.paywallComponents,
     )
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
@@ -77,7 +77,7 @@ internal class OfferingImagePreDownloader(
         val paywallComponentsConfig = paywallComponents.data.componentsConfig.base
 
         return paywallComponentsConfig.stack.findImageUrisToDownload() +
-            (paywallComponentsConfig.stickyFooter?.stack?.findImageUrisToDownload() ?: emptyList()) +
+            (paywallComponentsConfig.stickyFooter?.stack?.findImageUrisToDownload().orEmpty()) +
             paywallComponentsConfig.background.findImageUrisToDownload()
     }
 
@@ -88,34 +88,34 @@ internal class OfferingImagePreDownloader(
                 it is CarouselComponent ||
                 it is TabsComponent ||
                 it is ImageComponent
-        }.flatMap { component ->
+        }.flatMapTo(mutableSetOf()) { component ->
             when (component) {
                 is StackComponent -> {
-                    component.background.findImageUrisToDownload() + component.overrides.flatMap { override ->
-                        override.properties.background.findImageUrisToDownload()
-                    }.toSet()
+                    component.background.findImageUrisToDownload() + component.overrides.flatMapTo(mutableSetOf()) {
+                        it.properties.background.findImageUrisToDownload()
+                    }
                 }
                 is IconComponent -> {
                     setOf(Uri.parse(component.baseUrl).buildUpon().path(component.formats.webp).build())
                 }
                 is CarouselComponent -> {
-                    component.background.findImageUrisToDownload() + component.overrides.flatMap { override ->
-                        override.properties.background.findImageUrisToDownload()
-                    }.toSet()
+                    component.background.findImageUrisToDownload() + component.overrides.flatMapTo(mutableSetOf()) {
+                        it.properties.background.findImageUrisToDownload()
+                    }
                 }
                 is TabsComponent -> {
-                    component.background.findImageUrisToDownload() + component.overrides.flatMap { override ->
-                        override.properties.background.findImageUrisToDownload()
-                    }.toSet()
+                    component.background.findImageUrisToDownload() + component.overrides.flatMapTo(mutableSetOf()) {
+                        it.properties.background.findImageUrisToDownload()
+                    }
                 }
                 is ImageComponent -> {
-                    component.source.findImageUrisToDownload() + component.overrides.flatMap { override ->
-                        override.properties.source?.findImageUrisToDownload() ?: emptySet()
-                    }.toSet()
+                    component.source.findImageUrisToDownload() + component.overrides.flatMapTo(mutableSetOf()) {
+                        it.properties.source?.findImageUrisToDownload().orEmpty()
+                    }
                 }
                 else -> emptySet()
             }
-        }.toSet()
+        }
     }
 
     private fun Background?.findImageUrisToDownload(): Set<Uri> {
@@ -141,6 +141,7 @@ internal class OfferingImagePreDownloader(
      *
      * Implemented as breadth-first search.
      */
+    @Suppress("CyclomaticComplexMethod")
     private fun PaywallComponent.filter(predicate: (PaywallComponent) -> Boolean): List<PaywallComponent> {
         val matches = mutableListOf<PaywallComponent>()
         val queue = ArrayDeque<PaywallComponent>()
@@ -168,10 +169,12 @@ internal class OfferingImagePreDownloader(
                     }
                     queue.addAll(current.tabs.map { it.stack })
                 }
+                is TimelineComponent -> {
+                    queue.addAll(current.items.flatMap { listOfNotNull(it.title, it.description, it.icon) })
+                }
 
                 is TabControlToggleComponent,
                 is TabControlComponent,
-                is TimelineComponent,
                 is ImageComponent,
                 is IconComponent,
                 is TextComponent,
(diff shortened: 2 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -95,5 +96,6 @@ internal fun Offering.withPresentedContext(placementId: String?, targeting: Offe
         metadata = this.metadata,
         availablePackages = updatedAvailablePackages,
         paywall = this.paywall,
+        paywallComponents = this.paywallComponents,
```
tonidero, line 99: This was causing some issues with the `current` offering not having paywall components.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
```diff
@@ -104,19 +110,28 @@ internal fun rememberBackgroundStyle(background: BackgroundStyles): BackgroundSt
     }
 
 @Composable
-private fun rememberAsyncImagePainter(imageUrls: ImageUrls, contentScale: ContentScale): AsyncImagePainter =
-    rememberAsyncImagePainter(
+private fun rememberAsyncImagePainter(imageUrls: ImageUrls, contentScale: ContentScale): AsyncImagePainter {
+    var useCache by remember { mutableStateOf(true) }
```
tonidero, line 114: Copied the same mechanism we have for `RemoteImage`. We added this because sometimes it seems the cache could get corrupted, so we fallback to loading from network again.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -27,14 +47,140 @@ internal class OfferingImagePreDownloader(
...
+    }
+
+    private fun downloadV1Images(offering: Offering) {
         offering.paywall?.let { paywallData ->
             val imageUris = paywallData.config.images.all.map {
                 Uri.parse(paywallData.assetBaseURL.toString()).buildUpon().path(it).build()
             }
             imageUris.forEach {
-                debugLog("Pre-downloading paywall image: $it")
+                debugLog("Pre-downloading Paywall V1 paywall image: $it")
```
JayShortway, line 60:
```suggestion
                debugLog("Pre-downloading Paywall V1 image: $it")
```

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -27,14 +47,140 @@ internal class OfferingImagePreDownloader(
...
+                coilImageDownloader.downloadImage(it)
+            }
+        }
+    }
+
+    private fun downloadV2Images(offering: Offering) {
+        offering.paywallComponents?.let { paywallComponents ->
+            val imageUrls = findImageUrisToDownload(paywallComponents)
+            imageUrls.forEach {
+                debugLog("Pre-downloading Paywall V2 paywall image: $it")
```
JayShortway, line 70:
```suggestion
                debugLog("Pre-downloading Paywall V2 image: $it")
```

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -27,14 +47,140 @@ internal class OfferingImagePreDownloader(
...
                 coilImageDownloader.downloadImage(it)
             }
         }
     }
+
+    private fun findImageUrisToDownload(paywallComponents: Offering.PaywallComponents): Set<Uri> {
+        val paywallComponentsConfig = paywallComponents.data.componentsConfig.base
+
+        return paywallComponentsConfig.stack.findImageUrisToDownload() +
+            (paywallComponentsConfig.stickyFooter?.stack?.findImageUrisToDownload() ?: emptyList()) +
```
JayShortway, line 80:
```suggestion
            paywallComponentsConfig.stickyFooter?.stack?.findImageUrisToDownload().orEmpty() +
```

(Don't apply on GitHub, probably needs an import 😅)

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -27,14 +47,140 @@ internal class OfferingImagePreDownloader(
...
+                }
+                is TabsComponent -> {
+                    component.background.findImageUrisToDownload() + component.overrides.flatMap { override ->
+                        override.properties.background.findImageUrisToDownload()
+                    }.toSet()
+                }
+                is ImageComponent -> {
+                    component.source.findImageUrisToDownload() + component.overrides.flatMap { override ->
+                        override.properties.source?.findImageUrisToDownload() ?: emptySet()
+                    }.toSet()
```
JayShortway, line 114: We can use `flatMapTo(mutableSetOf())` to avoid creating the intermediate list, so we don't have to call `toSet()`.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -27,14 +47,140 @@ internal class OfferingImagePreDownloader(
...
+                }
+                is TabsComponent -> {
+                    component.background.findImageUrisToDownload() + component.overrides.flatMap { override ->
+                        override.properties.background.findImageUrisToDownload()
+                    }.toSet()
+                }
+                is ImageComponent -> {
+                    component.source.findImageUrisToDownload() + component.overrides.flatMap { override ->
+                        override.properties.source?.findImageUrisToDownload() ?: emptySet()
```
JayShortway, line 113: Here `?: emptySet()` and `orEmpty()` are not very different (no extra parenthesis needed for instance), so feel free to ignore. 
```suggestion
                        override.properties.source?.findImageUrisToDownload().orEmpty()
```
(Also needs an import probably.)

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -27,14 +47,140 @@ internal class OfferingImagePreDownloader(
...
+                    when (val control = current.control) {
+                        is TabsComponent.TabControl.Buttons -> queue.add(control.stack)
+                        is TabsComponent.TabControl.Toggle -> queue.add(control.stack)
+                    }
+                    queue.addAll(current.tabs.map { it.stack })
+                }
+
+                is TabControlToggleComponent,
+                is TabControlComponent,
+                is TimelineComponent,
```
JayShortway, line 174: For completeness, maybe we should add the timeline item's `title`, `description` and `icon` to the queue? (Also, the icon should be predownloaded right?)

purchases/src/test/java/com/revenuecat/purchases/utils/OfferingImagePreDownloaderTest.kt
```diff
@@ -45,33 +93,178 @@ class OfferingImagePreDownloaderTest {
...
+            "https://pawwalls.com/test_sticky_footer_override_2_dark_low_res.webp",
+            "https://pawwalls.com/test_icon_2.webp",
+        )
 
-        verify(exactly = 0) {
-            coilImageDownloader.downloadImage(any())
+        preDownloader.preDownloadOfferingImages(createOfferingWithV2Paywall(
+            paywallComponentsConfig = PaywallComponentsConfig(
+                stack = StackComponent(
+                    components = listOf(
```
JayShortway, line 143: Maybe we can add a `TimelineComponent` here with some icons.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
JayShortway: We don't have to recreate the entire `ImageLoader` though. (The `ImageLoader` is intended to be a singleton, which we're already not doing haha.) We should be able to just update the request, like so: 

```kotlin
var useCache by remember { mutableStateOf(true) }
val cachePolicy by remember { derivedStateOf { if (useCache) CachePolicy.ENABLED else CachePolicy.WRITE_ONLY } }
val request by remember(context) {
    derivedStateOf {
        ImageRequest.Builder(context)
            .data(imageUrls.webp.toString())
            .diskCachePolicy(cachePolicy)
            .memoryCachePolicy(cachePolicy)
            .build()
    }
}
```

We can then pass this `request` as the `model` to `rememberAsyncImagePainter`.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/tabs/TabsComponentViewTests.kt
```diff
@@ -1225,9 +1225,9 @@ class TabsComponentViewTests {
         LocalizationData.Text(value.replace('_', ' '))
 
     /**
-     * Returns all ComponentStyles that satisfy the predicate.
+     * Returns all PaywallComponent that satisfy the predicate.
```
JayShortway, line 1228: Ah good catch!
```suggestion
     * Returns all PaywallComponents that satisfy the predicate.
```

JayShortway (approved):
Let's go! 🚀  2 comments on the timeline icons and the image loader.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
tonidero: Ahh good catch!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
tonidero: Hmm using the ImageRequest won't allow us to change the directory of the cache right? We could remove that to use the default location instead... But we did it this way in order to have a separate cache for revenuecat and avoid any conflicting issues or anything.

tonidero: Talked over Slack, we need an image loader to specify the directory for the cache. However, we indeed should be using a single ImageLoader accross the SDK. Will keep it like this and work on that in future PRs.

PR #2145, opened 2025-02-12
[Paywalls V2] Correctly handles variables for lifetime products

JayShortway:
## Description
As the title says. Before this PR, lifetime and one-time-purchase products were treated equally, which is incorrect.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Localization.kt
```diff
@@ -168,6 +168,9 @@ enum class VariableLocalizationKey {
     @SerialName("annual_short")
     ANNUAL_SHORT,
 
+    @SerialName("lifetime")
+    LIFETIME,
```
JayShortway, line 172: Adding this to the network response is tracked by [MON-891](https://linear.app/revenuecat/issue/MON-891/network-response-should-contain-a-lifetime-variable-localization-key).

tonidero (approved):
LGTM!

PR #2146, opened 2025-02-12
[Paywalls V2] Extract ImageLoader to use a single one

tonidero:
### Description
We should be using a single `ImageLoader` but we're actually creating a bunch. This refactors our code to create a single one and pass that one where needed.


api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
```diff
@@ -42,6 +44,7 @@ import java.util.concurrent.ExecutorService
 
 @Suppress("unused", "UNUSED_VARIABLE", "EmptyFunctionBlock", "DEPRECATION")
 private class PurchasesAPI {
+    @OptIn(InternalRevenueCatAPI::class)
```
ajpallares, line 47: Is this the equivalent to what we have in iOS with `_spi(Internal)`. Just curious :)

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
```diff
@@ -26,25 +26,19 @@ import com.revenuecat.purchases.paywalls.components.common.Background
 import com.revenuecat.purchases.paywalls.components.properties.ThemeImageUrls
 
 internal class OfferingImagePreDownloader(
-    /**
-     * We check for the existance of the paywalls SDK. If so, the Coil SDK should be available to
-     * pre-download the images.
-     */
```
ajpallares, line 32: Should we perhaps keep this clarifying comment?

ajpallares (commented):
Looks good, but I'm not ready to approve these yet 🙏 
A couple of comments though

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingImagePreDownloader.kt
tonidero: Oh whoops true. I moved it and forgot to bring it back. Thanks!

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
tonidero: Yup! You need to explicitly annotate usages of these APIs like this.

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
```diff
@@ -102,6 +105,8 @@ private class PurchasesAPI {
         purchases.redeemWebPurchase(webPurchaseRedemption, redeemWebPurchaseListener)
         val parsedWebPurchaseRedemption: WebPurchaseRedemption? = Purchases.parseAsWebPurchaseRedemption(intent)
         val parsedWebPurchaseRedemption2: WebPurchaseRedemption? = Purchases.parseAsWebPurchaseRedemption("")
+
+        val imageLoader: ImageLoader? = purchases.getImageLoader()
```
JayShortway, line 109: We could argue that this doesn't need to be covered by API tests, as it is internal. However, I don't mind to have it.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -112,6 +114,8 @@ internal class PurchasesOrchestrator(
             offlineEntitlementsManager,
             customerInfoUpdateHandler,
         ),
+    val imageLoader: ImageLoader? = null,
+    val processLifecycleOwnerProvider: () -> LifecycleOwner = { ProcessLifecycleOwner.get() },
```
JayShortway, line 118: Just curious, why was this change needed?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -811,6 +812,11 @@ class Purchases internal constructor(
         purchasesOrchestrator.redeemWebPurchase(webPurchaseRedemption, listener)
     }
 
+    @InternalRevenueCatAPI
+    fun getImageLoader(): ImageLoader? {
+        return purchasesOrchestrator.imageLoader
+    }
```
JayShortway, line 818: This could be a property too, right? That would communicate more clearly (imo) that we're not doing anything special and you're just getting the instance.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/CoilImageDownloader.kt
```diff
@@ -3,35 +3,15 @@ package com.revenuecat.purchases.utils
...
 ) {
-    fun downloadImage(uri: Uri) {
+    fun downloadImage(uri: Uri, loader: ImageLoader) {
         val request = ImageRequest.Builder(applicationContext)
             .data(uri)
             .build()
-        applicationContext.getRevenueCatUIImageLoader().enqueue(request)
+        loader.enqueue(request)
     }
 }
```
JayShortway, line 17: This class is now so small, we could also inline its logic in `OfferingImagePreDownloader`. Or are we reusing this somewhere? What do you think?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/CoilImageLoader.kt
```diff
@@ -26,19 +27,29 @@ internal fun Context.getRevenueCatUIImageLoader(): ImageLoader {
...
-                    MemoryCache.Builder(this)
-                        .build(),
-                )
-                .build()
-            cachedImageLoader = newImageLoader
+            var newImageLoader = if (Purchases.isConfigured) {
+                Purchases.sharedInstance.getImageLoader()
+            } else {
+                null
+            }
```
JayShortway, line 34: Instead of creating it at `configure()` time, it would be nicer to have a static field in `Purchases` that gets set the first time it is accessed, e.g. `Purchases.imageLoader`. We could take inspiration from [`newImageLoader()`](https://github.com/coil-kt/coil/blob/2.7.0/coil-singleton/src/main/java/coil/Coil.kt#L66-L79) in Coil 2.7.0. 

This would avoid this entire scenario, because we'd always be able to get an image loader. That also means we don't need to duplicate the creation logic in both SDKs. What do you think?

JayShortway (commented):
This is great! The garbage collector will thank us 😄 Just a suggestion on making this a static field instead.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Ahh good question 😅 So, it seems that adding the `testImplementation` with the code library, changed our dependency with `androidx.lifecycle`. Before this PR, it had an implementation in Java, where we could mockk the `ProcessLifecycleOwner` using `mockkStatic`. After the new version, it was in kotlin, and our previous implementation didn't work. I thought of modifying the mocks to work, but then I thought we should try to inject it instead of mocking it like we were until now.

As I was thinking about injecting it, in order to avoid any behavior change, I preferred not to get it at the time the `PurchasesOrchestrator` is built (normally during Application's `onCreate`), instead, this is basically lazy loading it when it was being called before. I think it might be fine to get it at the time of the `PurchasesOrchestrator` creation, but not sure. Wdyt? Happy to change it to just store the object instead of lazy loading it.

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
tonidero: Yeah... not a strong opinion, but I thought to add it so at least we're aware of any changes.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/CoilImageDownloader.kt
tonidero: I would like it to keep it separate so we can more easily test the logic for the `OfferingImagePreDownloader`. Do lmk if you feel strongly about it though!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/CoilImageLoader.kt
tonidero: Hmm I'm wondering... I think the accessor in `Purchases` should still be nullable in case the `Coil` dependency isn't added... Though TBH, you would need to add it in order to access the class of that property at all... So maybe I will just make it not nullable and say it will crash if `Coil` isn't present. It's meant for internal usage anyway.

tonidero: Another thing as I was starting this, we do need the `Application` context to create the ImageLoader, which is only given to us at `configure` time. I guess we could lazy load it, but considering we need to prefetch images anyway, I feel it's better the way it is. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/CoilImageDownloader.kt
JayShortway: Good point!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/CoilImageLoader.kt
JayShortway: Hmm, yea my main concern is the duplication of the image loader configuration. I'd love to get rid of that with this effort. I see 2 options: 
- We just use `Purchases.sharedInstance.imageLoader` everywhere, and get rid of the `Context.getRevenueCatUIImageLoader()` function. We won't be showing any UI (and thus images) if we're not configured anyway. 
- Probably cleaner: we create a static `Purchases.getImageLoader(Context)` function, which only creates a new one the first time it is called, and otherwise just returns the same instance every time. (Good point on needing a context!)

tonidero: Hmm yeah I think that can work! I will do the second option then 👍

purchases/build.gradle
```diff
@@ -101,6 +101,7 @@ dependencies {
 
     dokkaPlugin(project(":dokka-hide-internal"))
 
+    testImplementation libs.coil.compose
```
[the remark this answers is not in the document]
JayShortway: It's not the end of the world imo!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1284,10 +1288,33 @@ internal class PurchasesOrchestrator(
...
+                            .maxSizeBytes(maxCacheSizeBytes)
+                            .build()
+                    }
+                    .build()
+                cachedImageLoader = imageLoader
+                imageLoader
+            } else {
+                currentImageLoader
+            }
+        }
```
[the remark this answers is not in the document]
tonidero: Whoops good point, thanks!!

JayShortway (approved):

PR #2147, opened 2025-02-12
[Paywalls V2] Add shadows to badge previews

tonidero:
### Description
This adds some shadows to the existing badge previews, which allow us to visualize some issues that will be fixed in a follow-up PR.


JayShortway (approved):
🙈

PR #2151, opened 2025-02-12
[Paywalls V2] Adjusts the convex/concave offset to 10% of the image height

JayShortway:
## Description
As the title says.


tonidero (approved):
Sounds good! I wonder if we should change this depending on whether the image has bigger width or height... But we can merge this for now.

PR #2152, opened 2025-02-12
[Paywalls V2] Fix issue with sizing of stacks with nested badges

tonidero:
### Description
Fixes an issue with sizing of stacks that have nested badges. The padding wasn't applied in the correct place

| Before | After |
| ----- | ----- |
| image | image |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
@@ -918,7 +918,7 @@ private fun StackComponentView_Preview_Nested_Badge(
                         light = ColorStyle.Solid(Color.Red),
                     ),
                 ),
-                padding = PaddingValues(all = 0.dp),
+                padding = PaddingValues(all = 20.dp),
                 margin = PaddingValues(all = 0.dp),
                 shape = Shape.Rectangle(CornerRadiuses.Dp(all = 20.0)),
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -547,27 +547,37 @@ private fun MainStackComponent(
...
         Modifier
             .applyIfNotNull(borderStyle) {
                 border(it, composeShape)
                     .padding(it.width)
             }
+    }
+
+    val innerShapeModifier = remember(stackState, borderStyle) {
+        Modifier
             .padding(stackState.padding)
```
JayShortway, line 560: Just curious, where was the padding applied before this PR?

JayShortway (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: Before this PR, for nested badges, the padding was applied to the box containing the stack + the nested badge. After this PR, the padding is applied as margin of the stack inside. This means the padding now doesn't affect the nested badge.

JayShortway: Makes sense, thanks for explaining! 🙏

PR #2154, opened 2025-02-13
[Paywalls V2] Change stack distribution system to use spacers instead of custom arrangements

tonidero:
### Description
This is an attempt to fix some issues with the arrangement in stacks not distributing space correctly between items. Specifically in SPACE_AROUND and SPACE_EVENLY, we are not positioning the items correctly.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
```diff
@@ -23,15 +19,11 @@ internal fun FlexDistribution.toHorizontalArrangement(spacing: Dp): Arrangement.
...
         FlexDistribution.CENTER -> Arrangement.spacedBy(spacing, Alignment.CenterHorizontally)
+        // For these, we modify the composable tree instead by adding composables with weight(1f) where appropriate.
         FlexDistribution.SPACE_BETWEEN,
-        // Arrangement.Horizontal.spacing is interpreted by Row as the spacing between items when measuring its
-        // size. Therefore we cannot rely on Arrangement.Horizontal alone to arrange items as SPACE_AROUND and
-        // SPACE_EVENLY with a minimum spacing, as those have spacing before and after all items too. This
-        // spacing is ignored by Row when measuring, causing children to be cut off.
-        // We'll need to add appropriately-sized Spacers manually.
         FlexDistribution.SPACE_AROUND,
         FlexDistribution.SPACE_EVENLY,
```
JayShortway, line 25: I don't think `SPACE_EVENLY` is rendering correctly now. It should have equal spacing before, between, and after the items, but the spacing between the items is larger than before and after. See [Emerge](https://www.emergetools.com/snapshot/aafefefe-b157-4d5c-9ea9-36e83d9e673b?tab=changed&selected=b72a8d31-662a-4441-9abe-8d296f7042c9): 
image

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
```diff
@@ -23,15 +19,11 @@ internal fun FlexDistribution.toHorizontalArrangement(spacing: Dp): Arrangement.
...
         FlexDistribution.END -> Arrangement.spacedBy(spacing, Alignment.End)
         FlexDistribution.CENTER -> Arrangement.spacedBy(spacing, Alignment.CenterHorizontally)
+        // For these, we modify the composable tree instead by adding composables with weight(1f) where appropriate.
         FlexDistribution.SPACE_BETWEEN,
-        // Arrangement.Horizontal.spacing is interpreted by Row as the spacing between items when measuring its
-        // size. Therefore we cannot rely on Arrangement.Horizontal alone to arrange items as SPACE_AROUND and
-        // SPACE_EVENLY with a minimum spacing, as those have spacing before and after all items too. This
-        // spacing is ignored by Row when measuring, causing children to be cut off.
-        // We'll need to add appropriately-sized Spacers manually.
         FlexDistribution.SPACE_AROUND,
```
JayShortway, line 24: `SPACE_AROUND` also doesn't seem correct. It should have twice as much space between the items as before and after. It's more easily visible in [the vertical one](https://www.emergetools.com/snapshot/aafefefe-b157-4d5c-9ea9-36e83d9e673b?tab=changed&selected=eea63b59-7055-4d03-9a98-e8d2531f7466). The space between the items is slightly larger than before/after, but not twice as much. 
image

JayShortway (commented):
In the right direction! ➡️

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
tonidero: Note that the preview actually has some spacing, which only gets applied in between the items. So I think this makes sense? If I remove that spacing, the space is equally distributed as expected.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
tonidero: Hmm isn't that also about the spacing? As in, right now the `spacing` is applied between the items, but not in the top/end of the items, as I consider that item-spacing. I think in the previous snapshot, the spacing made it to look like it was double but it's just a bit of spacing? Lmk if I'm wrong though!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
JayShortway: Oh that could be! Maybe we can have 2 previews then, with and without spacing? Because at a glance this looks incorrect, but it's good to verify both scenarios.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
JayShortway: Maybe we can have 2 previews here as well, with and without spacing? That helps verify both scenarios.

JayShortway: Hmm, `SPACE_AROUND` is still not entirely correct I think. [This one](https://www.emergetools.com/snapshot/ce59cce4-60dd-47c3-958c-dde8b931a0f2?tab=added&selected=57bd94ae-6811-4910-ad17-333da6ebe6c1) is without spacing: 
image
The space between items should be double the space before and after the items.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Distribution.kt
JayShortway: Checking Emerge, `SPACE_EVENLY` looks good!

JayShortway (approved):
Great job! 💪

JayShortway (approved):
Let's go!

PR #2155, opened 2025-02-13
Explicitly sets the `defaults` flavor as the default.

JayShortway:
## Description
As the title says. This makes Android Studio recognize this as the default too, which should reduce errors caused by mismatched variants. (For instance, previews in `revenuecatui` break if its selected variant doesn't match `purchases`.)

image


tonidero (approved):
❤️ Thanks for doing this!! 🙇

vegaro (approved):
Sweeeeeeet

PR #2157, opened 2025-02-13
[Paywalls V2] Fixes empty stacks not showing up.

JayShortway:
## Description
This was a though one. The cause of this is twofold:
1. `Column` and `Row` don't draw anything without children. Fixed by checking for this case and using a `Box` instead.
2. A `margin` would eat away the available space, making the divider invisible even with problem 1 fixed. This is fixed by adjusting the `size` for the `margin`. 

Fixes [MON-917](https://linear.app/revenuecat/issue/MON-917).

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
@@ -2,6 +2,9 @@
 
 package com.revenuecat.purchases.ui.revenuecatui.components.stack
 
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.calculateEndPadding
+import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.material3.adaptive.currentWindowAdaptiveInfo
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Stable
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -498,46 +501,56 @@ private fun MainStackComponent(
...
-                    }
+                    .size(stackState.size)
                     .then(rootModifier),
-                verticalArrangement = dimension.distribution.toVerticalArrangement(
-                    spacing = stackState.spacing,
-                ),
-                horizontalAlignment = dimension.alignment.toAlignment(),
-            ) { content { child -> if (child.size.height == Fill) Modifier.weight(1f) else Modifier } }
+            )
+        } else {
```
JayShortway, line 512: Hiding whitespace helps to show that the `when` below is unchanged.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -149,7 +162,39 @@ internal class StackComponentState(
...
     ) {
         if (windowSize != null) this.windowSize = windowSize
+        if (layoutDirection != null) this.layoutDirection = layoutDirection
+    }
+
+    /**
+     * Since margin is not a thing in Compose, and we apply it as extra padding outside the border and background, we
+     * need to make sure the Composable's size is large enough to contain the margin.
+     */
+    private fun Size.adjustForMargin(margin: PaddingValues, layoutDirection: LayoutDirection): Size {
```
tonidero, line 175: So we adapt `Fixed` sizes to include the margin if any... I guess that's making some sizes differently (as we can see in the modified snapshots)... but maybe it's ok to consider it like that?

tonidero (approved):
I think this makes sense. I do wonder if we might not want a divider component instead... But I think we can go with this for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
JayShortway: Yea I thought we might wanna consider it okay. The modified snapshots were fixed at 200, but were actually showing as 168 (margin of 16 on both sides). I _think_ what we have now is a valid interpretation.

PR #2158, opened 2025-02-13
[Paywalls V2] Apply stack margins to overlay badges

tonidero:
### Description
We were not applying the main stack's margin to the `Overlaid` badges. This means that the badges could look like they were outside the main stack. I think it makes sense that we apply the main stack margins to the badge as well.


JayShortway (approved):
Nice and simple!

PR #2159, opened 2025-02-13
[Paywalls V2] Disables the click handler for the selected package

JayShortway:
## Description
This has the desired effect of not being able to "select" a singular package on a paywall, without having to traverse the tree. It also avoids [the potentially confusing case of trying to reselect an already selected package](https://revenuecat.slack.com/archives/C07H5S8CFL3/p1739371997220059).


tonidero (approved):
Oh smart! That makes a lot of sense 👍

PR #2161, opened 2025-02-13
[Paywalls V2] Refactor to use a single ImageLoader singleton in RevenueCatUI

tonidero:
### Description
This is an attempt to fix an issue where some icon images where failing to load. Our theory is that the ImageLoader was getting deallocated, since it seemed the connection was getting closed. I haven't been able to reproduce though, so not sure if this will fix it. A couple things to consider:
- We are only using a single ImageLoader when fetching the images, we are still using multiple image loaders to predownload the images. Plan to address this later
- I noticed for images when cache is disabled, we're not setting the loader which in effect, would fallback to the "default" image loader. We might want to fix that at some point as well.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.revenuecatui.components.properties
 
+import android.content.Context
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.requiredSize
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
```diff
@@ -111,13 +114,16 @@ internal fun rememberBackgroundStyle(background: BackgroundStyles): BackgroundSt
 
 @Composable
 private fun rememberAsyncImagePainter(imageUrls: ImageUrls, contentScale: ContentScale): AsyncImagePainter {
-    var useCache by remember { mutableStateOf(true) }
-    val imageLoader = LocalContext.current.applicationContext.getRevenueCatUIImageLoader(readCache = useCache)
+    var cachePolicy by remember { mutableStateOf(CachePolicy.ENABLED) }
+    val context = LocalContext.current
+    val imageLoader = remember(context) {
+        context.applicationContext.getRevenueCatUIImageLoader()
+    }
```
JayShortway, line 121: This makes a lot of sense to me!

JayShortway (approved):
This is an improvement regardless of whether it fixes the bug haha. Although I failed to reproduce it with this branch sooo 🤞

PR #2162, opened 2025-02-13
[Paywalls V2] Fixes the top system bar padding being applied to all children in a vertical container

JayShortway:
## Description
As the title says. This caused the items to be spaced further apart than desired.


tonidero (approved):

tonidero (commented):
Makes sense :shipit:

PR #2164, opened 2025-02-14
[Paywalls V2] Fix wrong spacers in Stack when some children are Fill

tonidero:
### Description
With distribution `SPACE_AROUND` or `SPACE_EVENLY` we were still adding some spacers to the sides with `weight(1f)` even if some children had size `Fill`. This makes it so we don't add any of those when any children has `Fix` in the main axis of the stack.


JayShortway (approved):
Beautiful!

PR #2166, opened 2025-02-14
[Paywalls V2] Gradients Reloaded: Makes linear gradients match CSS even more closely

JayShortway:
## Description
A follow-up to https://github.com/RevenueCat/purchases-android/pull/2142, as there were some scenarios still incorrect. This adds a bunch of previews too, and I verified it against our template 014 button:

| purchases-android | dashboard |
|------------------|-----------|
|   |   |


tonidero (approved):
That's some interesting math 🔢. Looks good and nice previews too!

PR #2171, opened 2025-02-14
Fix back button dismissing the Customer Center

vegaro:
Pressing the back button was dismissing the whole customer center


tonidero (approved):
Ahh makes sense. Might be worth adding some tests at some point, but looks good for now!

PR #2172, opened 2025-02-14
[Paywalls V2] Adds `visible` property to all components

JayShortway:
## Description
As the title says. This is the Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4781. 

Closes [MON-893](https://linear.app/revenuecat/issue/MON-893/support-visible-in-android-sdk-at-component-level).


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentState.kt
```diff
@@ -89,7 +89,7 @@ internal class CarouselComponentState(
     }
 
     @get:JvmSynthetic
-    val visible by derivedStateOf { presentedPartial?.partial?.visible ?: true }
+    val visible by derivedStateOf { presentedPartial?.partial?.visible ?: style.visible }
```
tonidero, line 92: I was thinking, for the carousel pages, should we filter out those that are not visible?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
```diff
@@ -32,5 +32,6 @@ internal data class ButtonComponentStyle(
         }
     }
 
+    override val visible: Boolean = stackComponentStyle.visible
```
tonidero, line 35: Hmm I see... So the button component itself won't have a visibility property... Makes sense 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentState.kt
```diff
@@ -86,7 +86,7 @@ internal class TimelineComponentState(
     }
 
     @get:JvmSynthetic
-    val visible by derivedStateOf { presentedPartial?.partial?.visible ?: true }
+    val visible by derivedStateOf { presentedPartial?.partial?.visible ?: style.visible }
```
tonidero, line 89: Same thing here in the timeline about maybe filtering out the items that are not visible... Though not sure how this would draw entirely though...

tonidero (approved):
I think we need to fix the carousel pages not visible, not sure about the timeline though... Other than that, looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentState.kt
JayShortway: I think it's a good idea, but it's hard to implement properly right now as the `StackComponentView` (pages) are not stateless. We could do something like this, to calculate the state and then throw it away (because `StackComponentView` expects a `StackComponentStyle` instead of a `StackComponentState`), but I'm not sure if we should. 

```kotlin
val pages by derivedStateOf {
    style.pages.mapNotNull { pageStyle ->
        StackComponentState(
            initialWindowSize = windowSize,
            initialLayoutDirection = layoutDirection,
            style = pageStyle,
            selectedPackageProvider = selectedPackageProvider,
            selectedTabIndexProvider = selectedTabIndexProvider,
        ).takeIf { it.visible }
            ?.let { pageStyle }
    }
}
```

iOS also doesn't support this if I'm not wrong.

JayShortway: Yea I agree. I also think that the page indicator will be off. What do you think about doing this in a follow-up iteration? I think making `*ComponentView`s stateless will be a bit of a refactor.

tonidero: Yeah, I'm ok merging this for now 👍

PR #2173, opened 2025-02-14
Fix promotional offer not being displayed

vegaro:
We were not returning the updated state when loading the promotional offer, so it was effectively never been set in the state


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -454,13 +456,12 @@ internal class CustomerCenterViewModelImpl(
...
                         ),
                     )
-                    return true
                 } else {
                     currentState
                 }
             }
         }
-        return false
+        return loaded
```
vegaro, line 464: I think I can refactor this to check if there's a `promotionalOfferData` in the state, but I would need to do more tests

tonidero (approved):
:shipit:

PR #2176, opened 2025-02-17
Fix issue with release minification with Emerge's annotation 

tonidero:
### Description
This was actually causing compilation issues in release with minification enabled.

PR #2177, opened 2025-02-17
Enable paywall tester minification

tonidero:
### Description
We had minification disabled in paywall tester, which allowed us to miss the issue fixed in #2176. This enables it so CI picks up future issues with our UI module, which aren't caught by purchase tester. It also adds a job to be run as part of the tests building the paywall tester app.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,6 +129,29 @@ commands:
       - store_test_results:
           path: ~/gsutil/
 
+  build-paywall-tester:
+    steps:
+      - android/accept-licenses
+      - android/restore-gradle-cache
+      - android/restore-build-cache
+      - run:
+          name: Prepare Keystore
+          working_directory: examples/paywall-tester
+          command: echo $PAYWALL_TESTER_RELEASE_KEYSTORE | base64 -d > keystore
+      - run:
+          name: Replace API_KEY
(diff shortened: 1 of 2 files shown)


tonidero:
We already build purchase tester, but that doesn't include the revenuecatui module, this should help avoiding this issue in the future.

.circleci/config.yml
```diff
@@ -654,6 +667,7 @@ workflows:
       - test
       - detekt
       - assemble-purchase-tester
+      - assemble-paywall-tester-release
```
JayShortway, line 670: In a follow-up iteration it would be nice to share the Gradle products between jobs, since I believe it will now build twice, as part of `assemble-paywall-tester-release` and `publish-paywall-tester-release` below.

JayShortway (approved):
Thanks for doing this! 🙏

.circleci/config.yml
tonidero: Right certainly! I guess that will only happen on releases, where the publish job runs... But yeah we should do that 👍

JayShortway: It's published on every commit to `main` now 😄

tonidero: Ohh true, I forgot 😅

tonidero: Though in that case, I don't know if we can share anything since it's part of a different workflow... Maybe we can, not sure.

PR #2181, opened 2025-02-17
Creates a `ShowCustomerCenter ` contract for displaying as an Activity

vegaro:
### CustomerCenterActivity
Created a simple activity that:
- Hosts the Customer Center composable
- Handles dismissal through the result of the Activity

### ShowCustomerCenter
Created a simple contract for dismissals of the Customer Center
- Provides a `launch()` method that creates and launches the Customer Center

### Usage

```kotlin
// In Application class for analytics
class MyApplication : Application() {
    override fun onCreate() {
        super.onCreate()
        
        Purchases.sharedInstance.setCustomerCenterEventHandler(object : CustomerCenterEventHandler {
            override fun onRestoreStarted() {
                Analytics.track("customer_center_restore_started")
            }
        })
    }
}

// In Activity for UI
class MyActivity : ComponentActivity() {
    private val customerCenter = registerForActivityResult(ShowCustomerCenter()) { 
        // Handle dismissal through activity result
    }
    
    fun showCustomerCenter() {
        customerCenter.launch()
    }
}
```


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class CustomerCenterLauncher(
+    context: Context,
+    eventHandler: CustomerCenterEventHandler
+) {
+    private val weakContext = WeakReference(context)
+
+    init {
+        CustomerCenterEventBus.setEventHandler(eventHandler)
```
JayShortway, line 21: Can we not set the `eventHandler` at `launch()` time? That would make it feel less disconnected imo. At `launch()` time the dev has the most _context_ about why they're launching the Customer Center, which might influence what they want to do in their listener.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
```diff
@@ -0,0 +1,46 @@
...
+                    CustomerCenterEventBus.onDismissed()
+                    finish()
+                }
+            )
+        }
+    }
+}
+
+internal object CustomerCenterEventBus {
+    private var weakHandler = WeakReference<CustomerCenterEventHandler>(null)
```
JayShortway, line 35: I'm leaning towards keeping a strong reference, and writing cautionary docs about memory leaks. The downside of the weak reference is that all devs (even those not leaking anything) will need to keep their own reference to the `CustomerCenterEventHandler`, because otherwise it will be garbage-collected. This is unintuitive imo.

JayShortway (commented):
I like the `CustomerCenterEventBus`! But I'm challenging the need for a `CustomerCenterLauncher` 😈

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
```diff
@@ -0,0 +1,53 @@
...
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
+import android.content.Intent
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+import java.lang.ref.WeakReference
+
+interface CustomerCenterEventHandler {
+    fun onDismissed()
+}
```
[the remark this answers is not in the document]
vegaro: yes! `restoreStarted`, etc. They are not part of the Android API yet, but they will go here

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
```diff
@@ -0,0 +1,53 @@
...
+
+class CustomerCenterLauncher(
+    context: Context,
+    eventHandler: CustomerCenterEventHandler
+) {
```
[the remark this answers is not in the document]
tonidero: Hmm so FWIW, for paywalls, we make the activity internal, so developers don't really have a chance to interact with the activity directly, only with the API we provide. Otherwise, some developers might try to launch the activity with an intent, without passing the required parameters or something like that. I would probably try to keep that same pattern here.

As for whether to call this a Launcher or not... I don't have a strong opinion but for the sake of convention with paywalls, maybe it makes sense to keep it a launcher, even if it's not using an actual ActivityResultLauncher?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
```diff
@@ -0,0 +1,46 @@
...
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.ui.Modifier
+import java.lang.ref.WeakReference
+
+class CustomerCenterActivity : ComponentActivity() {
```
tonidero, line 12: As discussed in another comment, I would make this internal.

tonidero: Also, we need to update the manifest with this activity to be able to use it I think?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
+import android.content.Intent
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+import java.lang.ref.WeakReference
+
+interface CustomerCenterEventHandler {
+    fun onDismissed()
```
tonidero, line 11: Just a note that we might want to have default implementations if we add further methods in this interface, otherwise it could be considered a breaking change as well I think. (We should confirm how this would look like for Java though, not sure if this would work well there...)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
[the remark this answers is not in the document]
vegaro: one thing I have been thinking is that do we really need to expose this in purchases-android? Or can it be part of phc only? That's the first question, I asume we should have it in purchases-android for folks not using Compose, but wanted to get your thoughts

The other thing is that the event handler is optional, so it's ok if folks just create the Activity using the intent

tonidero (commented):
Looking good! just left some comments.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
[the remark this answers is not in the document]
tonidero: > do we really need to expose this in purchases-android? Or can it be part of phc only?

As you said, I think it might be good to have it in purchases-android to have a more usable Java API 

> The other thing is that the event handler is optional, so it's ok if folks just create the Activity using the intent

That's ok, but I still think it's better that we keep the contract with the activity private in case we need to add a mandatory parameter in the future.

> We could opt for a simple object instead of a class though, as CustomerCenterLauncher doesn't need to keep state, right?

I think that makes sense yeah!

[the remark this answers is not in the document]
vegaro: sounds great! on it!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class CustomerCenterLauncher(
+    context: Context,
+    eventHandler: CustomerCenterEventHandler
```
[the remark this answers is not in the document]
vegaro: hmm if I do that, then I don't need the EventBus right? 

I could just expose a public getter and call it?

```                  Purchases.sharedInstance.getCustomerCenterEventHandler()?.onDismissed()
```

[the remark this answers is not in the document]
tonidero: True yeah... We could mark it as an `@InternalRevenueCatAPI` but yeah,no need for the event bus in this approach.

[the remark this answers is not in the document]
vegaro: right... that makes me think if the `onDismiss` shouldn't be in the `CustomerCenterEventHandler` and be an activity result?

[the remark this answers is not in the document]
tonidero: I think that could work yeah!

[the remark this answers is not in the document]
tonidero: That should also handle activity configuration changes I think? So yeah, I think that makes more sense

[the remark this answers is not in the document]
JayShortway: Good point on letting the dev know at call site that the CC was dismissed. If we're going to use the Activity Result API, I'd propose we make a [custom contract](https://developer.android.com/training/basics/intents/result#custom), and let users use that using the standard `registerForActivityResult` Android APIs. For instance, something like:

```kotlin
val customerCenter = registerForActivityResult(ShowCustomerCenter())
// ...
customerCenter.launch()
```

What do you think?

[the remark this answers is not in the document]
tonidero: Hmm right... IMO the ideal API is one that can get a completion handler directly in the call site. So something like `CustomerCenterLauncher.launch(onCustomerCenterDismissed = { ... })`. However, it's true that it would be very difficult to abstract the API like that when using activities... (This is part of what the `PaywallActivityLauncher` was trying to achieve, unsuccessfully).

In order to use the activity result APIs we've been using for paywalls, devs still need to register this in their activity `onCreate`. Doesn't make for a too nice API IMO... but maybe it's ok seeing as it's standard and as long as we have alternatives for Compose and possibly a `CustomerCenterView` to have an option to not use an activity...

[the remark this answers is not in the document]
tonidero: But yeah, I think the custom contract you mentioned makes more sense than the `PaywallActivityLauncher` IMO!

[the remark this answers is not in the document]
vegaro: jsut to summarize the API. After this discussion, this is how that looks like (I think)

```kotlin
// In Application class for analytics
class MyApplication : Application() {
    override fun onCreate() {
        super.onCreate()
        
        Purchases.sharedInstance.setCustomerCenterEventHandler(object : CustomerCenterEventHandler {
            override fun onRestoreStarted() {
                Analytics.track("customer_center_restore_started")
            }
        })
    }
}

// In Activity for UI
class MyActivity : ComponentActivity() {
    private val customerCenter = registerForActivityResult(ShowCustomerCenter()) { 
        // Handle dismissal through activity result
    }
    
    fun showCustomerCenter() {
        customerCenter.launch()
    }
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterLauncher.kt
vegaro: I think they require Android 24? Our minSdk is 21

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.ui.Modifier
+
+internal class CustomerCenterActivity : ComponentActivity() {
```
tonidero, line 11: I guess some devs might want to display the activity in full screen/immersive mode... But I think that can be added later.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
```diff
@@ -0,0 +1,31 @@
...
+        internal fun createIntent(context: Context): Intent {
+            return Intent(context, CustomerCenterActivity::class.java)
+        }
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContent {
+            CustomerCenter(
```
tonidero, line 22: Once we add the listener in the `Purchases` class, I guess we will need to reference it here right? (Probably left for future PRs)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/ShowCustomerCenter.kt
```diff
@@ -0,0 +1,33 @@
...
+ *     fun showCustomerCenter() {
+ *         customerCenter.launch()
+ *     }
+ * }
+ * ```
+ */
+class ShowCustomerCenter : ActivityResultContract<Unit, Unit>() {
+    override fun createIntent(context: Context, input: Unit): Intent {
+        return CustomerCenterActivity.createIntent(context).apply {
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
tonidero, line 26: I'm always scared with these flags in Android TBH 😅 but `NEW_TASK` seems like it should work!

tonidero (approved):
This makes sense to me!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
JayShortway: On that topic, should we default to edge-to-edge? 🤔 It's the more modern way, and the default in API 35. 

Maybe we can make it configurable via the args for API 34 and below? We'd need to call [`enableEdgeToEdge()`](https://developer.android.com/reference/androidx/activity/ComponentActivity#(androidx.activity.ComponentActivity).enableEdgeToEdge(androidx.activity.SystemBarStyle,androidx.activity.SystemBarStyle)) in `onCreate()` before `super.onCreate()`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/ShowCustomerCenter.kt
```diff
@@ -0,0 +1,33 @@
...
+ * class MyActivity : ComponentActivity() {
+ *     private val customerCenter = registerForActivityResult(ShowCustomerCenter()) {
+ *         // Handle the dismissal
+ *     }
+ *
+ *     fun showCustomerCenter() {
+ *         customerCenter.launch()
+ *     }
+ * }
+ * ```
```
JayShortway, line 21: 🙌

JayShortway (approved):
I like it!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
tonidero: I think making it configurable is the way to go... But might be ok to rely on the default for now and add the option later (this applies to paywalls as well)

JayShortway: Yea would be really nice if we could support this. It's been a big push from Google for a while now to drive apps to use edge-to-edge.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
vegaro: yes! the listener is empty atm since the `CustomerCenter` composable doesn't accept the callbacks

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
vegaro: I am having an issue here I think:

Screenshot 2025-02-21 at 16 09 22

see it's below the app's top app bar?

vegaro: ok, I needed to set the theme to 

```
android:theme="@android:style/Theme.NoTitleBar"
```

tonidero: I think it makes sense for the customer center yeah!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/ShowCustomerCenter.kt
vegaro: https://github.com/RevenueCat/purchases-android/pull/2190 had to revert this lol

When opening the CustomerCenterActivity for result. I get a RESULT_CANCEL right away, even before the activity `onCreate`

PR #2182, opened 2025-02-19
[Paywalls V2] Allow manual scrolling if auto-advance is enabled

tonidero:
### Description
This enables manual scrolling in carousels if auto-advance is enabled, same as in iOS.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
@@ -58,6 +58,7 @@ import com.revenuecat.purchases.ui.revenuecatui.components.style.CarouselCompone
 import com.revenuecat.purchases.ui.revenuecatui.components.style.StackComponentStyle
 import com.revenuecat.purchases.ui.revenuecatui.data.PaywallState
 import com.revenuecat.purchases.ui.revenuecatui.extensions.applyIfNotNull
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -267,6 +252,36 @@ private fun Indicator(
...
+            try {
+                pagerState.animateScrollToPage(
+                    page = nextPage,
+                    animationSpec = tween(
+                        autoAdvance.msTransitionTime,
+                    ),
+                )
+            } catch (_: CancellationException) {
+                // Do nothing, so we continue scrolling on the next loop
+            }
```
JayShortway, line 280: Is my understanding correct that `animateScrollToPage` gets canceled when the user scrolls manually? And if we don't catch that, the carousel will no longer auto-advance, even if the manual scroll has stopped?

JayShortway (approved):
Great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
tonidero: Pretty much yeah, if the user catches the scroll as it's happening, this throws a cancellation exception, which would cancel the whole coroutine inside the `LaunchEffect` unless we catch it here.

PR #2184, opened 2025-02-19
[Paywalls V2] Adds template previews

JayShortway:
## Description
Still a draft because I wanna see how this looks on Emerge.


gradle/libs.versions.toml
```diff
@@ -40,8 +40,8 @@ commonmark = "0.21.0"
 activity-compose = "1.9.3"
 fragment = "1.6.1"
 hamcrest = "1.3"
-emergeGradlePlugin = "4.0.0"
-emergeSnapshots = "1.3.0"
+emergeGradlePlugin = "4.0.9"
+emergeSnapshots = "1.3.2"
```
JayShortway, line 44: This was done on advice from Emerge, but didn't end up fixing the issues.

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -0,0 +1,177 @@
...
+ *  device/emulator, and the PROJECT_DIR when the preview is shown in Android Studio.
+ *
+ *  We could place our files in the res folder and read them in a more straightforward way. However that means we need
+ *  a Context to read them, which we don't have access to in a PreviewParameterProvider. It's beneficial to read the
+ *  offerings JSON file directly in the PreviewParameterProvider, so we can parse the offering IDs and avoid hardcoding
+ *  them.
+ */
+private fun getResourceStream(fileName: String): InputStream =
+    object {}.javaClass.getResource("/$fileName")?.openStream()
+        ?: File(BuildConfig.PROJECT_DIR).resolve("src/debug/resources/$fileName").inputStream()
```
tonidero, line 112: I was wondering if it would be worth adding this to the `testpurchasesuiandroidcompatibility` to avoid including all this into the `revenuecatui` module... But then again, being in the debug source set, which we don't push, it should be ok I think.

ui/revenuecatui/src/debug/resources/offerings_paywalls_v2_templates.json
tonidero, line 1: That's a big json 😅

ui/revenuecatui/build.gradle
```diff
@@ -36,6 +42,7 @@ android {
     }
 
     buildFeatures {
+        buildConfig true
```
tonidero, line 45: I guess this might increase build time so very slightly in release without need, but then again, it shouldn't be noticeable at all, I'm totally ok with it

tonidero (approved):
Many thanks for doing this! It's great to have these 💪

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -0,0 +1,177 @@
...
+    val validationResult = offering.validatePaywallComponentsDataOrNullForPreviews()?.getOrThrow()!!
+    val state = offering.toComponentsPaywallState(
+        validationResult = validationResult,
+        activelySubscribedProductIds = emptySet(),
+        purchasedNonSubscriptionProductIds = emptySet(),
+        storefrontCountryCode = "US",
+        dateProvider = { Date() },
+    )
+
+    LoadedPaywallComponents(
```
tonidero, line 95: Should we wrap this into a full width/height box with a white background or something like that? To better visualize how it would look like on a device.

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
JayShortway: Yea I thought about that, but I don't think we'll be able to preview these in Android Studio right? As the JSON file won't be found if the previews themselves still live in `revenuecatui`. And we can't move the previews to `testpurchasesuiandroidcompatibility`, because we're previewing the internal `LoadedPaywallComponents` composable.

ui/revenuecatui/build.gradle
JayShortway: Hmm yea, I researched a bit but I don't think it's possible to only enable this in debug builds. I think it's negligible indeed, especially since there are no `buildCofigField`s defined for release builds.

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
JayShortway: Emerge doesn't really support this 🙁 I tried it in 2 ways now: 
- Setting a device spec in the `@Preview` annotation.
- Adding a `requiredSize` modifier to the root.

The device spec is ignored, and the `requiredSize` results in things being cut off at the top and bottom: 



Regarding the use of a parent composable, I'm not sure which alignment to pick then, and I'm afraid whichever alignment we pick might influence our results.

JayShortway: 2 more developments on this front: 
- I added a `WindowAndDisplayMetrics` composable to the `debug` source set to get some insights into the Emerge Snapshots rendering environment. Turns out it renders at 411 x 683 dp, while Android Studio's default rendering is at 392 x 850 dp, which is quite a bit taller.
- I tried to use [`DeviceConfigurationOverride`](https://developer.android.com/training/testing/different-screens/tools#deviceconfigurationoverride) to force the dimensions, but somehow Android Studio fails to render in that case. I think this could be a nice future improvement though.

PR #2189, opened 2025-02-21
chore: Add backend integration test for events

facumenzella:
### Motivation
This week we realized some events were not being sent to the backend. Even more, when we tried to add it we realized they errored out.

### Description
This PR adds a basic test to just make sure nothing's broken Ideally, the list of events should be generated somehow, but for now this is enough IMO.


tonidero (approved):
Nice!!

PR #2192, opened 2025-02-24
[Paywalls V2] Adds CI job to update template previews

JayShortway:
This is part 2 of 3:
1. [[#2184](https://github.com/RevenueCat/purchases-android/pull/2184)] Adds the initial `offerings_paywalls_v2_templates.json` file, and the previews themselves.
2. [This one] Adding a Fastlane lane and CircleCI job that updates the `offerings_paywalls_v2_templates.json` file.
3. [Coming soon] Triggering the CircleCI job from the [AutoSyncPaywallTemplatesJob](https://github.com/RevenueCat/khepri/blob/main/khepri/services/paywalls/jobs/auto_sync_paywall_templates.py) in khepri.
------------
## Description
- Adds a new Fastlane lane `update_paywall_templates`, which calls `get_offerings` on the Paywalls V2 Template Repo, saves it to a file, and opens a PR .
- Adds a new CircleCI job and workflow that runs the above lane, triggered by a new `update_paywall_templates` action.


fastlane/Fastfile
```diff
@@ -594,6 +594,45 @@ DESC
     )
   end
 
+  desc "Updates offerings_paywalls_v2_templates.json file and creates/updates PR"
+  lane :update_paywall_templates do |options|
```
tonidero, line 598: I wonder if we could move this entire lane to our fastlane plugin... Since it would be pretty much the same in iOS I imagine once we implement it (other than the file name and maybe the API key?).

tonidero (approved):
Great!! I left a comment, but I think it can be addressed later if we want to get this merged in.

fastlane/Fastfile
JayShortway: Ah yes, that's a great idea! I'll wait to see if we do indeed end up using the same flow in iOS, but if so we should definitely move it to the plugin.

PR #2195, opened 2025-02-25
[Paywalls V2] Adds docs on ignored arguments for Paywalls V2 in more places.

JayShortway:
## Motivation
We've seen some [questions](https://community.revenuecat.com/general-questions-7/close-exit-button-not-visible-even-after-using-shoulddisplaydismissbutton-true-in-kotlin-5953) around `shouldDisplayDismissButton` not working for Paywalls v2. Seems like we missed a few spots when updating the docs. This PR does that + uses lowercase 'v' (#marketing).


tonidero (approved):
Ahh right... too many APIs 😓 Thank you!!

markwitt1:
Whats the reason this decision was made? We used this parameter to only make paywalls dismissible on staging environments

JayShortway:
Hi @markwitt1, thanks for sharing your use case! Since v2 Paywalls are much more dynamic than v1 Paywalls, a boolean parameter that places an "arbitrary" dismiss button is likely to look off.  

How are you displaying the paywall? If you're using the `Paywall` composable, you could overlay your own dismiss button on staging environments, e.g. using a `Box`.

PR #2197, opened 2025-02-26
Adds X-Kotlin-Version header.

JayShortway:
## Motivation
This will allow us to make informed decisions on when to update the Kotlin language version, both in purchases-android and purchases-kmp.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -280,6 +280,7 @@ internal class HTTPClient(
             "X-Custom-Entitlements-Computation" to if (appConfig.customEntitlementComputation) "true" else null,
             "X-Storefront" to storefrontProvider.getStorefront(),
             "X-Is-Debug-Build" to appConfig.isDebugBuild.toString(),
+            "X-Kotlin-Version" to KotlinVersion.CURRENT.toString(),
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -280,6 +280,7 @@ internal class HTTPClient(
             "X-Custom-Entitlements-Computation" to if (appConfig.customEntitlementComputation) "true" else null,
             "X-Storefront" to storefrontProvider.getStorefront(),
             "X-Is-Debug-Build" to appConfig.isDebugBuild.toString(),
+            "X-Kotlin-Version" to KotlinVersion.CURRENT.toString(),
```
tonidero, line 283: Will this use the app's kotlin version and not our SDK's? Just making sure

tonidero (approved):
Nice!! This will be great to know 💪. Just a question but looks good!

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
JayShortway: The former, yes. This is a static field in the Kotlin stdlib. It is not generated at build time or anything like that. So this value represents the stdlib's value that ends up being included in the app. (At least that's how I understand it.)

PR #2199, opened 2025-02-26
Create `CustomerCenterListener`

vegaro:
This will allow developers to listen to events that occur inside the Customer Center


vegaro:
@RevenueCat/coresdk can you please take a look at this draft, I would like to make sure I get the public stuff right before continuing implementing.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -104,6 +105,17 @@ class Purchases internal constructor(
             purchasesOrchestrator.updatedCustomerInfoListener = value
         }
 
+    /**
+     * The listener is responsible for handling actions that happened in the Customer Center.
+     * Make sure [removeCustomerCenterListener] is called when the listener needs to be destroyed.
+     */
+    var customerCenterListener: CustomerCenterListener?
```
tonidero, line 112: I wonder if we would want to support multiple listeners... But then again, this is the API we have for the `CustomerInfo`, so it might make more sense to be consistent 👍

purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterListener.java
```diff
@@ -0,0 +1,31 @@
+package com.revenuecat.purchases.customercenter;
+
+import com.revenuecat.purchases.CustomerInfo;
+import com.revenuecat.purchases.PurchasesError;
+
+// This file is a Java file because doing default implementations in Kotlin interfaces doesn't seem to work well in
+// Java apps and gives a compilation error. This is a workaround to avoid that error.
```
tonidero, line 7: Hmm I wonder if we can use some of the kotlin annotations to make it work in Kotlin... @JayShortway in case you have any ideas.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterOptions.kt
```diff
@@ -0,0 +1,42 @@
+package com.revenuecat.purchases.ui.revenuecatui.customercenter
+
+import com.revenuecat.purchases.customercenter.CustomerCenterListener
+
+class CustomerCenterOptions internal constructor(
+    val listener: CustomerCenterListener?,
+) {
+
+    constructor(builder: Builder) : this(
```
tonidero, line 9: Hmm should the constructor here be `internal`? We should only allow building an instance of the options through the builder `build` method right?

tonidero (commented):
I wonder if there is anything we can do to make the interface in kotlin... Probably not the end of the world though. Looking good!

purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterListener.java
JayShortway: The `@JvmDefault` annotation is deprecated 😄, in favor of the `-Xjvm-default=all` compiler option. @vegaro Did you try adding that option? I think that should do the trick. Here's [an official JetBrains blog post](https://blog.jetbrains.com/kotlin/2020/07/kotlin-1-4-m3-generating-default-methods-in-interfaces/), and here's [the documentation](https://kotlinlang.org/docs/java-to-kotlin-interop.html#default-methods-in-interfaces).

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway, line 112: In terms of API design, I think I would either have 2 functions (`setCustomerCenterListener` and `removeCustomerCenterListener`), or a single, nullable property. Maybe 2 functions is more common?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterOptions.kt
```diff
@@ -0,0 +1,42 @@
+package com.revenuecat.purchases.ui.revenuecatui.customercenter
+
+import com.revenuecat.purchases.customercenter.CustomerCenterListener
+
+class CustomerCenterOptions internal constructor(
+    val listener: CustomerCenterListener?,
```
JayShortway, line 6: This could also be `internal` right?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
vegaro: yeah, I followed this to mimic the updated customer info listener, but I prefer the nullable property one. I think I did the updated customer info listener that way because I wanted to make Java look good, but it's 2025 now...

JayShortway: Yea agreed! I'd say we can get rid of `removeCustomerCenterListener()` then. Java users can call `setCustomerCenterListener(null)` and Kotlin users would do `customerCenterListener = null`.

purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterListener.java
vegaro: It looks like it works! [`e518837` (#2199)](https://github.com/RevenueCat/purchases-android/pull/2199/commits/e518837a67f80d1edcaf09266af220cb7fb7f98f)

JayShortway: Sweet!

tonidero (approved):
Looks great!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -105,6 +106,29 @@ class Purchases internal constructor(
...
+     *
+     * @note To use the Customer Center functionality, you need to include the RevenueCat UI SDK
+     * by adding the 'purchases-ui' dependency to your app's build.gradle file.
+     */
+    var customerCenterListener: CustomerCenterListener?
+        @Synchronized get() = purchasesOrchestrator.customerCenterListener
+
+        @Synchronized set(value) {
+            purchasesOrchestrator.customerCenterListener = value
+        }
```
JayShortway, line 130:
```suggestion
    var customerCenterListener: CustomerCenterListener? by purchasesOrchestrator::customerCenterListener
```

I think we can do without the `@Synchronized` annotations here, since the `purchasesOrchestrator::customerCenterListener` field already has those.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -642,4 +656,30 @@ internal class CustomerCenterViewModelImpl(
...
+    }
+
+    private fun notifyListenersForManageSubscription() {
+        listener?.onManageSubscriptionRequested()
+        Purchases.sharedInstance.customerCenterListener?.onManageSubscriptionRequested()
+    }
+
+    private fun notifyListenersForFeedbackSurveyCompleted(feedbackSurveyOptionId: String) {
+        listener?.onFeedbackSurveyCompleted(feedbackSurveyOptionId)
+        Purchases.sharedInstance.customerCenterListener?.onFeedbackSurveyCompleted(feedbackSurveyOptionId)
```
JayShortway, line 683: Not for this PR, but would be great if we can receive `Purchases.sharedInstance` via the constructor, like `PaywallViewModelImpl` does with `PurchasesType`. Would help a lot with testability.

JayShortway (approved):
Awesome! Really glad how this turned out! Just a suggestion.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
vegaro: oh good catch

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: we do have the `purchases: PurchasesType` in the constructor, I will fix it in this PR it should be simple

PR #2200, opened 2025-02-26
Avoids triggering "unscheduled" workflows when triggering workflows via the CircleCI API

JayShortway:
## Description 
We have a bunch of workflows that check if the `pipeline.trigger_source` is **not** `scheduled_pipeline`. These will all be triggered when triggering workflows from the API, however, which is not what we want. This PR adds an additional check to make sure that the `action` is `default`. 

## Motivation
We've added a CI workflow that updates the Paywalls v2 template previews in [#2192](https://github.com/RevenueCat/purchases-android/pull/2192), which will be triggered from khepri. We don't want to trigger all the other workflows together with that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -645,8 +645,10 @@ workflows:
   version: 2
   danger:
     when:
-      not:
-        equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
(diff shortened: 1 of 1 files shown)


ajpallares (approved):
Makes sense!

PR #2201, opened 2025-02-28
Adds a `LocalPreviewImageLoader` `CompositionLocal`.

JayShortway:
## Description
This replaces the `previewImageLoader` parameter in `RemoteImage` with a `LocalPreviewImageLoader` `CompositionLocal`. This is similar to [how previews are handled in Coil 3](https://coil-kt.github.io/coil/compose/#previews).

## Motivation
This will be very helpful in rendering images in our template previews. Without this change, we would have to propagate the `previewImageLoader` parameter through the entire component tree, adding it to every `ComponentView`. This change avoids that, because we can just wrap our previewed composable in `ProvidePreviewImageLoader()`.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -72,6 +72,7 @@ import com.revenuecat.purchases.ui.revenuecatui.components.style.ImageComponentS
 import com.revenuecat.purchases.ui.revenuecatui.composables.RemoteImage
```
JayShortway, line 72: I highly recommend reviewing this file with whitespace hidden on GitHub.

tonidero (approved):
I love this approach! 🫶 Thanks for doing this!

PR #2202, opened 2025-02-28
Fixes CustomerCenter state not refreshing when reopening

vegaro:
We noticed that the customer center wasn't reloading when reopening it, only after switching to another tab in the Paywalls Tester.

I moved the `loadCustomerCenter` to `LaunchedEffect` next to when we track the impression to fix it


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -70,13 +70,8 @@ internal fun InternalCustomerCenter(
...
     val context = LocalContext.current
 
-    if (state is CustomerCenterState.NotLoaded) {
-        coroutineScope.launch {
-            viewModel.loadCustomerCenter()
-        }
-    }
-
     LaunchedEffect(Unit) {
+        viewModel.loadCustomerCenter()
```
facumenzella, line 74: I see that `loadCustomerCenter` does not cancel previous calls, or at least doesn't seem to. 
Is it worth checking something like ... ?

```kotlin
    override suspend fun loadCustomerCenter() {
        if (_state.value is CustomerCenterState.Loading) {.  <--- early exit if its already loading
            return
        }
        if (_state.value !is CustomerCenterState.Loading) {
            _state.value = CustomerCenterState.Loading
        }
        try {
            val customerCenterConfigData = purchases.awaitCustomerCenterConfigData()
            val purchaseInformation = loadPurchaseInformation(dateFormatter, locale)
            _state.value = CustomerCenterState.Success(
                customerCenterConfigData,
                purchaseInformation,
                supportedPathsForManagementScreen = customerCenterConfigData.getManagementScreen()?.let {
                    supportedPaths(purchaseInformation, it)
                },
            )
            _state.value = CustomerCenterState.Error(e.error)
        }
    }
```

facumenzella (approved):
Just one question

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: hmmm that's a good point. Actually we shouldn't be loading the initial state here. Let me rethink it

facumenzella: What about a `refrehsIfNeeded` kind of thing?

vegaro:
I found the root cause. It was a regression from https://github.com/RevenueCat/purchases-android/pull/2171

vegaro:
I would love to add a test for this, but will do in a separate PR, it might be tricky

vegaro:
Added a test

PR #2203, opened 2025-02-28
Reload Customer Center after a successful restore

vegaro:
Reload the Customer Center after a successful restore

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/dialogs/RestorePurchasesDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/dialogs/RestorePurchasesDialog.kt
@@ -26,14 +26,16 @@ import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCent
 internal fun RestorePurchasesDialog(
     state: RestorePurchasesState,
     localization: CustomerCenterConfigData.Localization,
-    onDismiss: () -> Unit,
+    onDismiss: (restored: Boolean) -> Unit,
     onRestore: () -> Unit,
     onContactSupport: (() -> Unit)?,
 ) {
     when (state) {
         RestorePurchasesState.INITIAL -> InitialStateDialog(
             localization = localization,
-            onDismiss = onDismiss,
+            onDismiss = {
+                onDismiss(false)
+            },
             onRestore = onRestore,
         )
         RestorePurchasesState.PURCHASES_RECOVERED -> PurchasesRecoveredDialog(
@@ -42,7 +44,9 @@ internal fun RestorePurchasesDialog(
         )
         RestorePurchasesState.PURCHASES_NOT_FOUND -> PurchasesNotFoundDialog(
             localization,
-            onDismiss = onDismiss,
+            onDismiss = {
+                onDismiss(false)
+            },
             onContactSupport = onContactSupport,
         )
         RestorePurchasesState.RESTORING -> RestoringDialog(
@@ -129,10 +133,10 @@ private fun RestoringDialog(
 @Composable
 private fun PurchasesRecoveredDialog(
     localization: CustomerCenterConfigData.Localization,
-    onDismiss: () -> Unit,
+    onDismiss: (recovered: Boolean) -> Unit,
 ) {
     AlertDialog(
-        onDismissRequest = onDismiss,
+        onDismissRequest = { onDismiss(true) },
         title = {
             Text(
                 localization.commonLocalizedString(
@@ -150,10 +154,10 @@ private fun PurchasesRecoveredDialog(
             )
         },
         confirmButton = {
-            Button(onClick = onDismiss) {
+            Button(onClick = { onDismiss(true) }) {
                 Text(
                     localization.commonLocalizedString(
-                        key = CustomerCenterConfigData.Localization.CommonLocalizedString.CANCEL,
+                        key = CustomerCenterConfigData.Localization.CommonLocalizedString.DISMISS,
                     ),
                 )
             }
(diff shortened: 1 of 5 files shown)


pantelis-rc (approved):

vegaro:
I've noticed an issue here, since I am updating the state twice. going back to draft

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -235,15 +235,20 @@ internal class CustomerCenterViewModelImpl(
...
-                currentState.copy(
-                    showRestoreDialog = false,
-                    restorePurchasesState = RestorePurchasesState.INITIAL,
-                )
-            } else {
-                currentState
+    override suspend fun dismissRestoreDialog(restored: Boolean) {
+        if (restored) {
+            // Reload customer center data to refresh the UI with the latest information
+            loadCustomerCenter()
```
tonidero, line 241: Hmm does it make sense to just do this always after dismissing the restore dialog, even if there were no purchases? I don't think it would be a huge deal and would simplify the code but not sure if there are reasons for it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/dialogs/RestorePurchasesDialog.kt
```diff
@@ -129,10 +133,10 @@ private fun RestoringDialog(
 @Composable
 private fun PurchasesRecoveredDialog(
     localization: CustomerCenterConfigData.Localization,
-    onDismiss: () -> Unit,
+    onDismiss: (recovered: Boolean) -> Unit,
```
tonidero, line 136: This brought to mind some conversations we've had recently for paywalls, about whether to display our own restore dialogs in paywalls or leave it to the developer. While there are some cases we might want to display them from our side, currently we're letting the developer handle these dialogs on their side so they can customize any way they like. I think CustomerCenter doesn't require that much customization so it makes sense to keep it like this I think...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/dialogs/RestorePurchasesDialog.kt
```diff
@@ -26,14 +26,16 @@ import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCent
...
     when (state) {
         RestorePurchasesState.INITIAL -> InitialStateDialog(
             localization = localization,
-            onDismiss = onDismiss,
+            onDismiss = {
+                onDismiss(false)
+            },
             onRestore = onRestore,
         )
         RestorePurchasesState.PURCHASES_RECOVERED -> PurchasesRecoveredDialog(
```
tonidero, line 41: Not from this PR but I was looking at the logic behind knowing whether purchases were recovered. It seems we just check if there are any active subs or any non subs at all... I do wonder if we can consider that purchases were recovered in that case, since those might be purchases from other stores, and nothing was recovered from the current store for example... 

I guess to get this properly, we would need some sort of diff between the customer info before/after the restore, but that's a bit trickier... Nothing to do in this PR, but something to consider about the behavior here.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterViewModelTests.kt
```diff
@@ -206,4 +194,192 @@ class CustomerCenterViewModelTests {
...
+            type = Screen.ScreenType.MANAGEMENT,
+            title = "Management",
+            subtitle = null,
+            paths = listOf(missingPurchasePath)
+        )
+        
+        every { configData.getManagementScreen() } returns managementScreen
+        
+        // Create a spy on the real ViewModel so we can verify loadCustomerCenter is called
+        val model = spyk(
```
tonidero, line 234: Not a huge fan of spies 🕵️  since I think it makes the tests do more than it should... But I think it's ok to test some end to end behaviors, like we are mostly doing here. 👍

tonidero (approved):
Looking good! left some comments but I wouldn't block the PR. Lmk what you think though!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterViewModelTests.kt
vegaro: right... I hesitated but it was making very easy to test what I wanted

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: you know what? I like it. I wasn't a fan of this restored boolean but didn't find an alternative. If someone is opening the restore dialog anyways it would be good to reload just in case.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterViewModelTests.kt
vegaro: I think I can actually remove it lol, let me check

vegaro: removed it!

PR #2204, opened 2025-02-28
chore: Delete key from customer center survey event

facumenzella:
### Motivation
This is the counterpart of https://github.com/RevenueCat/purchases-ios/pull/4837

### Description
`survey_option_title_key` is ignored in the backend, so there's no need to have it here. 
This PR removes it and adds a test that verifies backwards compatibility with stored events. However, I think we should move to an implementation similar to iOS, where the StoredEvents contains an already encoded version of the event. That would remove any need to test backwards compatibility.


tonidero (approved):
Looks good!

vegaro (approved):

PR #2208, opened 2025-03-03
[Paywalls V2] Some more template previews optimizations

JayShortway:
## Description
Previously, we'd read every Offering into memory (one at a time), and would discard Offerings we'd already seen. That puts pressure on the garbage collector. In this PR, we only instantiate an Offering if we're actually going to return it. 

We do that by first identifying the indices at which each Offering appears in the JSON, and then only reading the characters between those indices when we need a particular Offering. This is an attempt to decrease the rendering errors on Emerge Snapshots.


JayShortway:
My theory for why emerge sees 56 added snapshots, is as follows. 
- The **first** time the snapshots were generated for the commit currently on `main`, a bunch errored and it only saw 206 snapshots. 
- Then I retried the snapshots on `main `, and the **second** time it did see 262 snapshots. 

However, it seems that it is still comparing against those 206 snapshots from the first time. 


image

tonidero (approved):

PR #2209, opened 2025-03-03
[Paywalls V2] Ignores template previews for now.

JayShortway:
## Motivation
`PaywallComponentsTemplate_Preview` has been identified as a problematic preview that crashes the Emerge emulator. Apparently we're running into this emulator crash: https://issuetracker.google.com/issues/398784711. 

I'm ignoring it to keep our shipping velocity up. I will explore rendering them outside of Emerge and then uploading them as an alternative next.

### DIFF
--- a/ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
+++ b/ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
@@ -4,6 +4,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+import com.emergetools.snapshots.annotations.IgnoreEmergeSnapshot
 import com.revenuecat.purchases.Offering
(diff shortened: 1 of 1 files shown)


facumenzella (approved):

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -132,6 +133,7 @@ private class OfferingProvider : PreviewParameterProvider<Offering> {
     }
 }
 
+@IgnoreEmergeSnapshot
```
tonidero, line 136: Looking at the issue in the description, would it make sense to run the emerge job in a macos machine temporarily if that fixes the crash? (they say it didn't happen in aarch64 macos)

tonidero (approved):
Just something to maybe try... But I'm ok merging this for now if we prefer

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
JayShortway: I don't think that will help, as my understanding is that only the upload happens on the CircleCI machine. The actual rendering happens on Emerge infrastructure, in emulators.

PR #2210, opened 2025-03-03
Virtual Currency Balances

fire-at-will:
### Motivation
This PR allows developers to check the balances of their Virtual Currencies through the SDK's `CustomerInfo` object.

### Description
This PR allows developers to access the balances of their Virtual Currencies like so:
```kotlin
val coinBalance: Long? = customerInfo.virtualCurrencies["COIN"]?.balance
```

### New APIs
It accomplishes this with a few API additions:
```kotlin
data class CustomerInfo(
   // ...
   val virtualCurrencies: Map<String, VirtualCurrencyInfo>,
)

@Parcelize
data class VirtualCurrencyInfo internal constructor(
    val balance: Long,
) : Parcelable
```

The `balance` property is a `Long` because `Long`s represent signed 64 bit integers, which is what is used in the iOS SDK and backend.

### Testing
- Adds the new APIs to the Kotlin & Java API Testers
- Adds unit tests for parsing the virtual currency balances from JSON into the new objects on CustomerInfo
- Manually tested in the purchase tester app


purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlinx.parcelize.IgnoredOnParcel
+import kotlinx.parcelize.Parcelize
+import kotlinx.parcelize.TypeParceler
+import org.json.JSONObject
+
+@Parcelize
+@TypeParceler<JSONObject, JSONObjectParceler>()
+data class VirtualCurrencyInfo(
+    val balance: Long,
+    private val jsonObject: JSONObject,
```
fire-at-will, line 15: I'm not super familiar with the pattern here of including the `jsonObject` in the data class but saw it in several other model data classes. Is it appropriate to keep that here?

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
```diff
@@ -0,0 +1,40 @@
...
+        return balance == other.balance
+    }
+
+    override fun hashCode(): Int {
+        return balance.hashCode()
+    }
+
+    companion object {
+        fun fromJson(json: JSONObject): VirtualCurrencyInfo {
+            val balance = json.getLong("amount")
```
fire-at-will, line 36: The value returned by the server may change in the future to "balance" in the future - we'll update that in a follow-up PR if/when it happens.

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.json.JSONObject
+
+/**
+ * A class representing information about a virtual currency in the app.
+ *
+ * @property balance The current balance of the virtual currency.
+ */
+@Parcelize
+data class VirtualCurrencyInfo internal constructor(
+    val balance: Long,
```
tonidero, line 14: I'm wondering if it would be worth adding the currency id/name here as well... It's available as the key of the map in `CustomerInfo` but might be more developer friendly to duplicate it here I think. Can be done in follow-up PRs too!

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlinx.parcelize.Parcelize
+import org.json.JSONObject
+
+/**
+ * A class representing information about a virtual currency in the app.
+ *
+ * @property balance The current balance of the virtual currency.
+ */
+@Parcelize
+data class VirtualCurrencyInfo internal constructor(
```
tonidero, line 13: We have been trying to avoid data classes, since they introduce some methods automatically and it's difficult to keep compatibility as an SDK. I think we should use a normal `class` and annotate it with `@Poko` which would automatically generate the equals + hash code methods.

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
```diff
@@ -0,0 +1,34 @@
...
+    val balance: Long,
+) : Parcelable {
+    override fun equals(other: Any?): Boolean {
```
tonidero, line 16: I think this + hashcode shouldn't be needed whether we're using a data class or using a `@Poko` annotated class.

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
```diff
@@ -0,0 +1,34 @@
...
+        other as VirtualCurrencyInfo
+        return balance == other.balance
+    }
+
+    override fun hashCode(): Int {
+        return balance.hashCode()
+    }
+
+    companion object {
+        fun fromJson(json: JSONObject): VirtualCurrencyInfo {
```
tonidero, line 29: I think this should be `internal`.

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
tonidero, line 13: Do we want to mark this with `@ExperimentalPreviewRevenueCatPurchasesAPI` for now? (Just checking)

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
```diff
@@ -50,6 +51,7 @@ data class CustomerInfo(
     val originalAppUserId: String,
     val managementURL: Uri?,
     val originalPurchaseDate: Date?,
+    val virtualCurrencies: Map<String, VirtualCurrencyInfo>,
```
tonidero, line 54: Same here about marking this with `@ExperimentalPreviewRevenueCatPurchasesAPI`

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
tonidero, line 14: Just to check, we are not planning on supporting decimal numbers for the currencies correct?

tonidero (commented):
Mostly some questions but looking great!

JayShortway (commented):
Good stuff! No real comments besides Toni's.

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
fire-at-will: Love this, changed from a `data` class to `@Poko` in 8a335cba4be2bc2aa8b05e948e00774f114469ec.

For my information, do we have docs or discussions anywhere talking about why we want to avoid data classes? I'm curious 😄

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
fire-at-will: Good catch - made internal in f2efb610f2584c3dc391d1e23437f796cf50af9c

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
JayShortway: Not sure about internal docs, but I always like this blog post on the topic:
[_Public API challenges in Kotlin_ - Jake Wharton](https://jakewharton.com/public-api-challenges-in-kotlin/)

fire-at-will: I think this makes sense! I've added `@ExperimentalPreviewRevenueCatPurchasesAPI` in 1be37ca848c21a586e1d63a960e791be39bdda68.

This required me to opt-in to it in a few places internally, are there any downsides to that? You can see all of them in the commit (1be37ca848c21a586e1d63a960e791be39bdda68). If we don't like it, I can revert the commit.

tonidero: Yup, that post is pretty good to explain the problems with data classes. I would reference that 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/CustomerInfoFactory.kt
```diff
@@ -78,6 +80,17 @@ internal object CustomerInfoFactory {
...
+        val virtualCurrenciesObject = subscriber.optJSONObject(
+            CustomerInfoResponseJsonKeys.VIRTUAL_CURRENCIES,
+        ) ?: JSONObject()
+
+        val virtualCurrencies = buildMap<String, VirtualCurrencyInfo> {
+            virtualCurrenciesObject.keys().forEach { currencyId ->
+                val currencyJson = virtualCurrenciesObject.getJSONObject(currencyId)
+                put(currencyId, VirtualCurrencyInfo.fromJson(currencyJson))
+            }
+        }
```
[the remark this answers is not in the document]
fire-at-will: I may not be using this properly, but I'm finding that `.toMap()` isn't casting the value to a `VirtualCurrencyInfo` object properly. This code causes a crash since it isn't sure how to cast a `JSONObject` to a `VirtualCurrencyInfo`:

```kotlin
val virtualCurrencies: Map<String, VirtualCurrencyInfo> = virtualCurrenciesObject.toMap(deep = true)
val virtualCurrencyInfo: VirtualCurrencyInfo? = virtualCurrencies["WIL"]
println(virtualCurrencyInfo?.balance)
```

```
java.lang.ClassCastException: org.json.JSONObject cannot be cast to com.revenuecat.purchases.VirtualCurrencyInfo                       	
at com.revenuecat.purchases.common.CustomerInfoFactory.buildCustomerInfo(CustomerInfoFactory.kt:91)
// ...
```

Do we have a standard way of handling that, like providing a way to cast `JSONObject` to `VirtualCurrencyInfo`? I'm not seeing a lot of other similar usages for the `.toMap()` function

purchases/src/main/kotlin/com/revenuecat/purchases/VirtualCurrencyInfo.kt
JayShortway: Came across official Kotlin docs on the same topic: 
[_Avoid using data classes in your API_](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api)

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
```diff
@@ -50,6 +52,9 @@ data class CustomerInfo(
     val originalAppUserId: String,
     val managementURL: Uri?,
     val originalPurchaseDate: Date?,
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    val virtualCurrencies: Map<String, VirtualCurrencyInfo>,
```
tonidero, line 57: One of the problems with data classes is that this modifies the constructor + copy constructor for `CustomerInfo`, in case someone is using it... We could try to limit the breaking change by adding a deprecated constructor without the new param

api-tester/src/main/java/com/revenuecat/apitester/java/VirtualCurrencyInfoAPI.java
```diff
@@ -0,0 +1,12 @@
+package com.revenuecat.apitester.java;
+
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI;
+import com.revenuecat.purchases.VirtualCurrencyInfo;
+
+final class VirtualCurrencyInfoAPI {
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    static void check(final VirtualCurrencyInfo virtualCurrencyInfo) {
+        final long balance = virtualCurrencyInfo.getBalance();
+        final Long balance2 = virtualCurrencyInfo.getBalance();
```
tonidero, line 10: I'm not sure if it's really needed to have both the object and the primitive here... But I guess it doesn't hurt 👍

tonidero (approved):
I think this look great! Great job on this!

api-tester/src/main/java/com/revenuecat/apitester/java/VirtualCurrencyInfoAPI.java
fire-at-will: Yeah, I'm not sure if it's completely necessary either, but since both are possible, I figured I'd add it 🤷‍♂️

mvichos (commented):
FYI: [This](https://github.com/RevenueCat/khepri/pull/12142) has been merged. Some fields in the backend response have been renamed.

JayShortway (approved):
Looking good! 💪 Avoiding the breaking constructor in `CustomerInfo` might be nice, as Toni suggested. Then we at least remain source compatible.

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
fire-at-will: Great idea @tonidero, I've added a deprecated constructor without the new param in 112656360778ada938db03c71e9aa25147910a7f 👍

PR #2213, opened 2025-03-04
Adds support for switching between 2 API keys to PaywallsTester

JayShortway:
This adds the ability to switch between 2 projects (API keys) to PaywallsTester. 

## Motivation
To be able to effectively test new paywall templates on Android. See [MON-970](https://linear.app/revenuecat/issue/MON-970) for more context and details.

## Description
I wanted to do this in the most pragmatic way possible without it being hacky, since we might be building other ways of previewing paywalls in the near future. Since the goal is to switch between 2 specific projects, I chose to support exactly that. Other options would have been to allow logging in with your account like RC Mobile, or adding your own API key in a text field, but that would have meant adding persistence. Let me know if you disagree or have other ideas! 🙏 

- `Constants.kt` now has `GOOGLE_API_KEY_A` and `GOOGLE_API_KEY_B`. The app uses `A` by default. During development you can leave `B` empty, as it is not used unless you attempt to switch API keys in the UI. 
- There's a new "Switch API key" button on the App Info screen.
- The published version of PaywallsTester will include the "Paywalls V2 Alpha" (as before) and "Official Paywalls V2 Template Repo" keys. 
- The `$PAYWALLS_V2_TEMPLATE_REPO_API_KEY` environment variable already exists on Circle CI for our `update-template-previews` update job. 

Closes [MON-970](https://linear.app/revenuecat/issue/MON-970).

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
@@ -28,6 +28,7 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.Dialog
 import androidx.lifecycle.viewmodel.compose.viewModel
+import com.revenuecat.paywallstester.Constants
 import com.revenuecat.paywallstester.ui.screens.main.appinfo.AppInfoScreenViewModel.UiState
 import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.ui.debugview.DebugRevenueCatBottomSheet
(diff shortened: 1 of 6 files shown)


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
```diff
@@ -124,17 +146,73 @@ private fun LoginDialog(viewModel: AppInfoScreenViewModel, onDismissed: () -> Un
...
+                        Text("CANCEL")
+                    }
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun ApiKeyButton(label: String, apiKey: String, onClick: (String) -> Unit) {
```
tonidero, line 183: One alternative would be to just default to the "embedded" API key, and then add a text view to introduce a new API key to switch to that (with a reset button I guess). Might provide more flexibility while still working out of the box, but we should probably persist that API key as you said... Which honestly, I think it would be fine to do in SharedPreferences or something like that... 

Having said that, I don't have a strong opinion, so I'm ok with your solution! Just a bit weird to have an `A` and `B` API key without a name 😅

tonidero (approved):
TBH, I wonder if it might be better to just persist the last API key used. Mostly since I think the `A`/`B` API keys distinction are very easy to forget... But then again I don't want to block testing so I'm ok merging this for now.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
JayShortway: Yea I agree, good points. I'll merge this for now to unblock testing, but I'll think about what we can do to make this less confusing.

PR #2221, opened 2025-03-05
Fixes inconsistent package ordering on tiered paywalls (template 7)

JayShortway:
## Bug
The packages on template 7 were ordered inconsistently.

## Cause
The packages were ordered based on the order of the packages in the offering, not on the order of the packages on the paywall.

## Fix
The fix is to use the correct ordering. It's not exactly big-O optimized, but the 'n' should be very small here.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackageConfigurationFactory.kt
```diff
@@ -279,7 +279,7 @@ internal object PackageConfigurationFactory {
         storefrontCountryCode: String?,
         zeroDecimalPlaceCountries: List<String>,
     ): List<TemplateConfiguration.PackageInfo> {
-        val filtered = from.filter { filter.contains(it.identifier) }
+        val filtered = filter.mapNotNull { from.find { pkg -> pkg.identifier == it } }
```
joshdholtz, line 282: So obvious but not so obvious 🫠 

Thanks for finding this!

joshdholtz (approved):
:shipit: :shipit: :shipit:

ajpallares (approved):

PR #2228, opened 2025-03-06
[Paywalls V2] Fix CarouselComponent peek

tonidero:
### Description
Fix CarouselComponent peek, which didn't account for page spacing, making it not look right in some situations.


JayShortway (approved):
Nice and quick!

PR #2229, opened 2025-03-06
fix: Remove API dependency with Coil

tonidero:
### Description
We don't want to have a dependency on our API in the Coil library, which can cause issues with some mocking libraries this hides the dependency.

### DIFF
--- a/integration-tests/proguard-rules.pro
+++ b/integration-tests/proguard-rules.pro
@@ -20,4 +20,3 @@
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
 -keep class com.revenuecat.purchases.** { *; }
--dontwarn coil.**
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
@@ -16,7 +16,6 @@ import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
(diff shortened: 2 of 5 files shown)


integration-tests/proguard-rules.pro
```diff
@@ -20,4 +20,3 @@
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
 -keep class com.revenuecat.purchases.** { *; }
--dontwarn coil.**
```
tonidero, line 23: Tested and this isn't needed anymore

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/BackgroundStyle.kt
```diff
@@ -117,7 +118,7 @@ private fun rememberAsyncImagePainter(imageUrls: ImageUrls, contentScale: Conten
     var cachePolicy by remember { mutableStateOf(CachePolicy.ENABLED) }
     val context = LocalContext.current
     val imageLoader = remember(context) {
-        Purchases.getImageLoader(context.applicationContext)
+        Purchases.getImageLoader(context.applicationContext) as ImageLoader
```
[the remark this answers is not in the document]
JayShortway: Maybe we can add an `internal` (really internal) extension in `revenuecatui` that provides the `ImageLoader` with the correct type already? That way we avoid having to cast this at all the call sites. I'm thinking something like this: 

```kotlin
internal fun Purchases.Companion.getImageLoaderTyped(Context): ImageLoader
```

integration-tests/proguard-rules.pro
JayShortway: Awesome!

JayShortway (approved):
Nice one! Just the suggestion to limit the times we need to cast and give ourselves a slightly nicer api for this.

JayShortway (approved):

PR #2230, opened 2025-03-07
[Paywalls V2] Fixes locales without region and `X-Preferred-Locales`

JayShortway:
## Bug
A paywall with a non-default locale containing only a language would fall back to the default locale, even if the device prefers the non-default one.

## Causes
This has 2 causes:
- The SDK didn't properly match against locales without a region, such as `ja`. The device would have `ja-JP` configured, and it would not pick `ja` from the paywall configuration. 
- The `X-Preferred-Locales` header was sent with locales containing dashes, but khepri expects underscores. This meant that the SDK never received any non-default locales. 

## Fix
Respectively:
- The SDK now checks the "language only" variant of the locale as well.
- The SDK sends the `X-Preferred-Locales` header with underscores.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -270,7 +270,7 @@ internal class HTTPClient(
             "X-Platform-Device" to Build.MODEL,
             "X-Platform-Brand" to Build.BRAND,
             "X-Version" to Config.frameworkVersion,
-            "X-Preferred-Locales" to localeProvider.currentLocalesLanguageTags,
+            "X-Preferred-Locales" to localeProvider.currentLocalesLanguageTags.replace(oldChar = '-', newChar = '_'),
             "X-Client-Locale" to appConfig.languageTag,
             "X-Client-Version" to appConfig.versionName,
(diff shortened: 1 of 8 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -270,7 +270,7 @@ internal class HTTPClient(
             "X-Platform-Device" to Build.MODEL,
             "X-Platform-Brand" to Build.BRAND,
             "X-Version" to Config.frameworkVersion,
-            "X-Preferred-Locales" to localeProvider.currentLocalesLanguageTags,
+            "X-Preferred-Locales" to localeProvider.currentLocalesLanguageTags.replace(oldChar = '-', newChar = '_'),
```
tonidero, line 273: Hmm I wonder if we could do this processing in the backend TBH... That way we would fix existing SDKs.

JayShortway: Yes, [I agree](https://revenuecat.slack.com/archives/C07H5S8CFL3/p1741349946699619?thread_ts=1741297015.998839&cid=C07H5S8CFL3)! I figured we can do both in any case.

PR #2214, opened 2025-03-04
[Paywalls V2] Renders template previews in a fixed resolution using Paparazzi

JayShortway:
## Description
This PR is the next step towards cross-platform paywall rendering validation. It adds a `PaywallComponentsTemplatePreviewRecorder` which renders our existing `PaywallComponentsTemplate_Preview` using Paparazzi. This is done for 2 reasons:
- It allows us to pick a resolution and density, to be able to match the other platforms.
- It gives us access to the screenshot files themselves, for further processing. 

Follow-up PRs will collect the screenshot files and push them to a separate repo. 

**Bonus:** this gives us a bit more coverage in the Android project too, as these fixed-resolution screenshots are also uploaded to [Emerge](https://www.emergetools.com/snapshot/compare/aa166da4-6534-434b-8215-5c033d08622f?tab=added).

PR #2232, opened 2025-03-07
Uploads Paparazzi screenshots to Emerge.

JayShortway:
This is a transformation of https://github.com/RevenueCat/purchases-android/pull/2214. 

## Description
We have quite a few Paparazzi tests from pre-Emerge. This PR uploads them to Emerge to verify them. They are still rendered by Paparazzi. 

**Note:** The Emerge comment complains that the base build is not found. I'm assuming that is referring to the base of the Paparazzi snapshots. My idea is to merge this, see if it starts working in subsequent PRs, and revert if not. Emerge is processing the snapshots fine:
- [Previews](https://www.emergetools.com/snapshot/0a8f5017-02cc-4a7c-a21f-628b496c81fd?tab=unchanged)
- [Paparazzi](https://www.emergetools.com/snapshot/1e98ca9f-6bff-4635-bdb8-29efb6c27755)

## Motivation
Accepting new screenshots as the golden variant is 1 click on Emerge, while previously it took multiple commands and an extra PR.


.circleci/config.yml
```diff
@@ -155,6 +155,28 @@ commands:
       - android/save-gradle-cache
       - android/save-build-cache
 
+  install-ruby:
```
tonidero, line 158: Might be worth using the ruby orb for this: https://circleci.com/developer/orbs/orb/circleci/ruby#commands-install

.circleci/config.yml
```diff
@@ -517,21 +539,6 @@ jobs:
       - store_test_results:
           path: ~/gsutil/
 
-  verify_debug_view_snapshot_tests:
```
tonidero, line 520: 🔥

tonidero (approved):
Sooo much easier with this 🙌 Thank you so much for doing this! Left a comment that I think could clean that up a bit, but not a blocker.

JayShortway:
Bypassing rules only for Emerge Snapshot Test and verify-revenuecatui-snapshots, as those are expected to fail in this PR. Will update the required checks now.

PR #2234, opened 2025-03-10
[Paywalls V2] Handles potential locale id inconsistency between strings and variables

JayShortway:
This was reported here: https://github.com/RevenueCat/react-native-purchases/issues/1178#issuecomment-2710805526.

## Bug
Some paywalls would fail to render due to missing variable localizations: 
>All variable localizations for locale 'de_DE' are missing.

## Cause
This was caused by the fact that the network response would sometimes contain inconsistent locale ids in `components_localizations` (`de_DE`) and `ui_config.localizations` (`de`). The Android SDK would pick `de_DE` and would be unable to find the variable localizations in `ui_config.localizations`. 

## Fix
This is fixed by having the SDK also check the "language only" locale id if the "full" locale id is not found.


joshdholtz (approved):
:shipit: :shipit: :shipit:

PR #2236, opened 2025-03-11
Fix background painting outside composable

tonidero:
### Description
There was an issue where the background painter could actually paint outside the composable itself. This fixes that by making clipping happen before painting.
| Before | After |
| --- | --- |
| image | image |


JayShortway (approved):
Ah this is great, thanks! I see you checked with shadows too, awesome 🙌

PR #2238, opened 2025-03-11
[Paywalls V2] Reorganizes stack arrangement logic

JayShortway:
## Description
This PR adds `HorizontalStack` and `VerticalStack`, which are used by `StackComponentView`. They handle the item arrangement logic, adding spacers where needed. There should be no change in functionality. Just moving stuff around.

## Motivation
`StackComponentView.kt` is a huge file. Moving self-contained pieces of logic out of there improves maintainability.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/HorizontalStack.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/HorizontalStack.kt
@@ -13,16 +13,18 @@ import androidx.compose.ui.unit.Dp
 import com.revenuecat.purchases.paywalls.components.properties.Dimension
 import com.revenuecat.purchases.paywalls.components.properties.FlexDistribution
 import com.revenuecat.purchases.paywalls.components.properties.Size
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/HorizontalStack.kt
```diff
@@ -0,0 +1,108 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.stack
```
tonidero, line 1: Need to add `@file:JvmSynthetic` in this and the `VerticalStack` files?

tonidero (approved):
Nice cleanup! This is a great first step 🙌 

I would really like to extract the badge code somehow, but there are so many things to consider that it's quite tricky and haven't thought about how to do it...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/HorizontalStack.kt
JayShortway: Great catch! https://github.com/RevenueCat/purchases-android/pull/2238/commits/a9ed0cc5647bc4f362dab57f4239062fd44a0bae

PR #2241, opened 2025-03-11
Add Virtual Currency Balances

fire-at-will:
This PR brings the virtual currency balances functionality into `main`. We had originally planned to keep this functionality in a beta branch, but loading the SDK into another app from a branch without a full release is pretty difficult to do since Gradle loads dependencies from maven repositories instead of Git repos.

Instead, this PR brings the virtual currency balance code into `main` to be released under the `@ExperimentalPreviewRevenueCatPurchasesAPI` annotation, which will require developers to explicitly opt-in to using the experimental feature. Since the APIs are marked with `@ExperimentalPreviewRevenueCatPurchasesAPI`, we can make breaking changes in the future without needing to make a major release.

This code has already been reviewed in https://github.com/RevenueCat/purchases-android/pull/2210. That PR was merging the code into the `virtual-currency-dev` branch, this PR merges it into `main`.

### Other Approaches Considered
We looked into distributing the virtual currencies beta with Jitlab, but Jitlab is unable to build and distribute branches of our SDK since it can't access the keystore file (build error logs [here](https://jitpack.io/com/github/revenuecat/purchases-android/8.14.1/build.log)).


purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
```diff
@@ -50,9 +52,38 @@ data class CustomerInfo(
...
     val managementURL: Uri?,
     val originalPurchaseDate: Date?,
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    val virtualCurrencies: Map<String, VirtualCurrencyInfo>,
     private val jsonObject: JSONObject,
 ) : Parcelable, RawDataContainer<JSONObject> {
 
+    @Deprecated("Use constructor with virtualCurrencies parameter")
+    constructor(
```
fire-at-will, line 62: This deprecates the old constructor to keep source compatibility

purchases/src/main/kotlin/com/revenuecat/purchases/common/CustomerInfoFactory.kt
```diff
@@ -78,6 +82,17 @@ internal object CustomerInfoFactory {
...
+        val virtualCurrenciesObject = subscriber.optJSONObject(
+            CustomerInfoResponseJsonKeys.VIRTUAL_CURRENCIES,
+        ) ?: JSONObject()
+
+        val virtualCurrencies = buildMap<String, VirtualCurrencyInfo> {
+            virtualCurrenciesObject.keys().forEach { currencyId ->
+                val currencyJson = virtualCurrenciesObject.getJSONObject(currencyId)
+                put(currencyId, VirtualCurrencyInfo.fromJson(currencyJson))
+            }
+        }
```
tonidero, line 94: I'm wondering, since this is still experimental, should we be defensive here and catch errors parsing this data? In case we want to change the backend response for any reason while it's experimental.

tonidero (commented):
Just a possible improvement to avoid breaking changes in case we need to change anything, but looks good otherwise!

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
fire-at-will: Since the new constructor requires `@ExperimentalPreviewRevenueCatPurchasesAPI`, maybe it'd be a good idea to not deprecate this one yet, keep it as an alternate constructor, and then deprecate it if/when the constructor with the virtual currencies has the experimental annotation removed

purchases/src/main/kotlin/com/revenuecat/purchases/common/CustomerInfoFactory.kt
fire-at-will: Works for me! I'll modify it to default to an empty map if any parsing errors occur

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
```diff
@@ -53,6 +53,39 @@ data class CustomerInfo(
...
+        entitlements: EntitlementInfos,
+        allExpirationDatesByProduct: Map<String, Date?>,
+        allPurchaseDatesByProduct: Map<String, Date?>,
+        requestDate: Date,
+        schemaVersion: Int,
+        firstSeen: Date,
+        originalAppUserId: String,
+        managementURL: Uri?,
+        originalPurchaseDate: Date?,
+        virtualCurrencies: Map<String, VirtualCurrencyInfo>,
```
fire-at-will, line 72: We place `virtualCurrencies` in its own new @OptIn constructor to preserve the existing constructor. We can move it to the primary constructor and deprecate the existing one once virtual currencies are no longer opt-in

fire-at-will:
Closing for now in favor of using Jitpack

PR #2245, opened 2025-03-12
[Paywalls V2] Considers script when matching locales

JayShortway:
## Description
We were not taking the script into account when matching locales. This is particularly important for Traditional and Simplified Chinese. This meant we would not properly fall back to the correct locale for these languages if there's no exact region match. This PR fixes that. 

This is the Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/4870.


purchases/src/test/java/com/revenuecat/purchases/paywalls/components/common/LocaleIdTests.kt
```diff
@@ -1,66 +0,0 @@
-package com.revenuecat.purchases.paywalls.components.common
-
-import org.assertj.core.api.Assertions.assertThat
-import org.junit.Test
-import org.junit.runner.RunWith
-import org.junit.runners.Parameterized
-
-@RunWith(Parameterized::class)
-internal class LocaleIdTests(
```
JayShortway, line 9: I deleted this test because I decided to no longer have parsed fields (`language`, `region`, `script`) as part of the `LocaleId` value class for performance reasons. They were parsed every time they were read, and they were read multiple times when trying to find a locale match. 

This logic is now in `Set<LocaleId>.getBestMatch()` in `Localization.kt`. It is tested by `PaywallComponentDataValidationTests` and `PaywallStateLoadedComponentsLocaleTests`.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Localization.kt
```diff
@@ -24,17 +26,57 @@ import kotlinx.serialization.encoding.Encoder
...
+ * Scripts inferred from the region.
+ */
+private val scriptByRegion = mapOf(
+    "CN" to "Hans",
+    "SG" to "Hans",
+    "MY" to "Hans",
+    "TW" to "Hant",
+    "HK" to "Hant",
+    "MO" to "Hant",
+)
```
JayShortway, line 56: I did not find a way around this, as Java's `Locale` won't return a script if the original language tag doesn't contain one. This seems to be different from iOS. 

We need this to be able to e.g. return `zh_Hant` when `zh_TW` is requested.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Localization.kt
```diff
@@ -16,25 +16,54 @@ import kotlinx.serialization.descriptors.SerialDescriptor
...
 
+/**
+ * Returns the best match to [localeId] in this set, or null if no match is found.
+ */
 @InternalRevenueCatAPI
-fun LocaleId.languageOnly(): LocaleId =
-    LocaleId(language)
+fun Set<LocaleId>.getBestMatch(localeId: LocaleId): LocaleId? {
```
ajpallares, line 49: Probably not a big deal since we're probably dealing with small sets, but would it be more appropriate to do early returns to avoid some computations? 
I totally get that perhaps the code wouldn't be so clean.

I'm thinking that probably in most cases, `contains(localeId)` returns `true`, so all the operations would be pointtless.

ajpallares (approved):
I think it looks good! I only have a small comment

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Localization.kt
JayShortway: I like it! Doing an early return for the exact match is a nice way to squeeze some extra perf without reducing readability. https://github.com/RevenueCat/purchases-android/pull/2245/commits/9112df4dc3dbd4ee91aaec22ae4876fa13edeb5f

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/Localization.kt
```diff
@@ -16,25 +16,57 @@ import kotlinx.serialization.descriptors.SerialDescriptor
...
```
tonidero, line 49: I'm wondering if all this logic belongs here or in the UI module... But I think it might be ok to keep here for now...

Also, this is something we might also have to do for CustomerCenter... though in that case, it might be in the backend. cc @vegaro

tonidero (approved):
This is complicated 😅  but I think it makes sense!

PR #2247, opened 2025-03-12
[Paywalls V2] Refactors `StyleFactory`

JayShortway:
## Description
The `ComponentStyle` tree needs some state in certain branches: `rcPackage`, `tabControl` and `tabIndex`. This PR changes how this state is kept. Instead of passing it recursively through the tree as function arguments, its kept in a new `StyleFactoryScope` receiver.

## Motivation
- This makes parsing the component tree more scaleable. Adding more and more function parameters becomes unwieldy, especially since they need to be added to multiple functions even if only a single component needs the new parameter. 
- This is the first step which will allow us to avoid parsing the tree multiple times, e.g. to find the packages and tabs. We will be able to keep track of the packages and tabs directly in the first pass.


tonidero (approved):
Interesting approach. I like it!

PR #2252, opened 2025-03-13
[Paywalls V2] Avoids re-parsing the component tree to find packages and tabs

JayShortway:
## Description
This is a continuation of https://github.com/RevenueCat/purchases-android/pull/2247. It uses the newly introduced `StyleFactoryScope` to keep track of the packages in the component tree, and whether they are in a tab or not, in a single pass.

## Motivation
Reducing the number of times we parse the component tree improves performance. It also allows us to remove a lot of relatively complex recursive parsing code.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -178,22 +177,16 @@ internal fun Offering.validatePaywallComponentsDataOrNull(
...
+        val stickyFooter = stickyFooterResult?.componentStyle
         // This is a temporary hack to make the root component fill the screen. This will be removed once we have a
         // definite solution for positioning the root component.
         val rootComponent = (backendRootComponent as? StackComponentStyle)
             ?.takeIf { it.size.height == SizeConstraint.Fit }
             ?.copy(size = Size(width = SizeConstraint.Fill, height = SizeConstraint.Fill))
             ?: backendRootComponent
 
-        val packagesOutsideTabs = rootComponent.findPackagesOutsideTabs() + stickyFooter?.findPackagesOutsideTabs().orEmpty()
-        val tabsComponent = rootComponent.findTabsComponentStyle() ?: stickyFooter?.findTabsComponentStyle()
```
JayShortway, line 190: Removing these 2 lines is the main goal of this PR.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -86,37 +87,105 @@ internal class StyleFactory(
     }
 
     private data class StyleFactoryScope(
-        var rcPackage: Package? = null,
+        var packageInfo: AvailablePackages.Info? = null,
         var tabControl: TabControlStyle? = null,
+        var tabControlIndex: Int? = null,
         var tabIndex: Int? = null,
```
tonidero, line 93: Hmm could you explain a bit more of the difference between `tabControlIndex` and `tabIndex`? Feels like they should always be the same?

tonidero (commented):
Looking good but I have a question about the new properties... Might be missing something, but I think it can be confusing to have both `tabIndex` and `tabControlIndex`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JayShortway: Good point, added here: https://github.com/RevenueCat/purchases-android/pull/2252/commits/72f2ad6b71ed16da70b8832a4276a5870fe53dd9. Let me know if that's understandable! 

>Feels like they should always be the same?
Keep in mind btw that these indices are not about what's currently selected. We're not showing anything yet at this point, just building the tree. These indices indicate which branch of the tree is currently being built.

tonidero: Ahhh right 🤦‍♂️ yeah makes sense thanks for those comments!

tonidero (approved):
That explained a lot, thank you! Looks great!

PR #2253, opened 2025-03-13
[Paywalls V2] Timeline component fixes

tonidero:
### Description
We discovered a bunch of issues with the timeline component positioning. Specifically:
- The description wasn't positioned correctly when it was multiline
- If the icons in different items had different sizes, the timeline was completely misaligned.
- The second item could overlay the first item if the icon/text was too big. This PR improves that by aligning either the title or the icon (depending on the icon alignment) to the bottom of the previous row.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -82,39 +92,47 @@ internal fun TimelineComponentView(
...
-                        if (isLastItem) {
-                            bottom.linkTo(parent.bottom)
-                        }
                     },
             )
 
-            IconComponentView(
-                style = item.icon,
-                state = state,
+            Box(
```
tonidero, line 98: Had to wrap the icon into a box so we can size the box with the size of the biggest icon, then the icon can just be center aligned.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -82,39 +92,47 @@ internal fun TimelineComponentView(
...
                 modifier = Modifier.constrainAs(iconRef) {
                     when (timelineState.iconAlignment) {
                         TimelineComponent.IconAlignment.Title -> {
-                            top.linkTo(titleRef.top)
-                            bottom.linkTo(titleRef.bottom)
+                            top.linkTo(currentPreviousItem ?: parent.top)
```
tonidero, line 102: This is part of the fixes I applied to make sure the rows didn't overlap each other.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -82,39 +92,47 @@ internal fun TimelineComponentView(
...
+                        TimelineComponent.IconAlignment.Title -> {
+                            centerVerticallyTo(iconRef)
+                        }
+                        TimelineComponent.IconAlignment.TitleAndDescription -> {
+                            top.linkTo(currentPreviousItem ?: parent.top)
+                        }
+                    }
+                    start.linkTo(iconEndBarrier)
+                    end.linkTo(parent.end)
+                    width = Dimension.fillToConstraints
```
tonidero, line 135: Without this, the text was increasing more than the size of the parent when the text was longer.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -159,9 +181,9 @@ internal fun TimelineComponentView(
                             width = Dimension.value(item.connector?.width?.dp ?: 0.dp)
                             top.linkTo(currentIconRef.top)
                             if (isLastItem) {
-                                bottom.linkTo(parent.bottom)
+                                bottom.linkTo(parent.bottom, margin = offsets.second)
                             } else {
-                                bottom.linkTo(nextIconRef!!.top)
```
tonidero, line 164: This was wrong... I was relaying on the size of each icon being the same and the offset solving this solution, but if we support icons of different sizes, this doesn't work well.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -67,11 +72,16 @@ internal fun TimelineComponentView(
     ) {
         val itemBarriers = mutableListOf<HorizontalAnchor>()
         val iconRefs = mutableListOf<ConstrainedLayoutReference>()
-        for ((index, item) in timelineState.items.withIndex()) {
-            val isLastItem = index == timelineState.items.size - 1
+        val biggestIconWidth: Dp? = remember(timelineState.items) {
+            timelineState.items
+                .maxByOrNull { it.icon.size.width.dpOrNull() ?: 0.dp }
+                ?.icon?.size?.width?.dpOrNull()
+        }
```
JayShortway, line 79: - Since `timelineState.items` is a `State`, we can use `derivedStateOf` to make `biggestIconWidth` a `State` too. Could save some recompositions. 
- I think `maxOfOrNull` works here? Saves the final re-read of the `width` of the max item. 

(Don't apply cause you'll need imports)
```suggestion
        val biggestIconWidth: Dp? by remember {
            derivedStateOf {
                timelineState.items
                    .maxOfOrNull { it.icon.size.width.dpOrNull() ?: 0.dp }
            }
        }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
```diff
@@ -159,9 +181,9 @@ internal fun TimelineComponentView(
                             width = Dimension.value(item.connector?.width?.dp ?: 0.dp)
                             top.linkTo(currentIconRef.top)
                             if (isLastItem) {
-                                bottom.linkTo(parent.bottom)
+                                bottom.linkTo(parent.bottom, margin = offsets.second)
                             } else {
-                                bottom.linkTo(nextIconRef!!.top)
+                                bottom.linkTo(nextIconRef!!.bottom, margin = nextItemIconHalfSize + offsets.second)
```
JayShortway, line 186: I don't mind the double bang, but maybe a comment explaining why we know it won't be null here would be helpful. (I realise this didn't change in this PR.)

JayShortway (approved):
Nice polish! Just 2 suggestions.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentView.kt
tonidero: Oh you're right, thank you!!

JayShortway (approved):

PR #2266, opened 2025-03-18
[Paywalls V2] Fixes badges not being overriden

JayShortway:
## Description

Addresses [MON-927](https://linear.app/revenuecat/issue/MON-927) for Android. iOS PR is here: https://github.com/RevenueCat/purchases-ios/pull/4900.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.ui.revenuecatui.components
 
 import com.revenuecat.purchases.ColorAlias
 import com.revenuecat.purchases.paywalls.components.PartialStackComponent
+import com.revenuecat.purchases.paywalls.components.StackComponent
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
```diff
@@ -36,6 +41,9 @@ internal class PresentedStackPartial(
         operator fun invoke(
             from: PartialStackComponent,
             aliases: Map<ColorAlias, ColorScheme>,
+            createStackComponentStyle: (
```
tonidero, line 44: Hmm since this is only used for badges, what do you think of a rename here  to something like `badgeStyleFactory` or `createBadgeStackComponentStyle`?

tonidero (approved):
Just a naming suggestion, but this looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedStackPartial.kt
JayShortway: Yes, good idea! https://github.com/RevenueCat/purchases-android/pull/2266/commits/2785fc73b4f7cb05f2fafb9e66ef26499a7b2342

PR #2267, opened 2025-03-18
[Diagnostics] Tracks BillingClient connected/disconnected events.

JayShortway:
## Description
As the title says.


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -599,6 +599,11 @@ internal class BillingWrapper(
 
     @Suppress("LongMethod")
     override fun onBillingSetupFinished(billingResult: BillingResult) {
+        diagnosticsTrackerIfEnabled?.trackGoogleServiceConnected(
+            responseCode = billingResult.responseCode,
+            debugMessage = billingResult.debugMessage,
+            pendingRequestCount = serviceRequests.size,
```
JayShortway, line 605: `.size` is O(n) because `serviceRequests` is a queue. While `n` is probably small, we can also decide not to track this. If we do want to track this, here are some alternative I can come up with: 
- Only check whether this is non-zero, using `peek()`.
- Count up to a maximum number (e.g. 10). 

@RevenueCat/coresdk What do you think?

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsEntryName.kt
```diff
@@ -8,6 +8,8 @@ internal enum class DiagnosticsEntryName {
     GOOGLE_QUERY_PRODUCT_DETAILS_REQUEST,
     GOOGLE_QUERY_PURCHASES_REQUEST,
     GOOGLE_QUERY_PURCHASE_HISTORY_REQUEST,
+    GOOGLE_SERVICE_CONNECTED,
```
tonidero, line 11: Made a comment in the doc about maybe renaming this to `GOOGLE_SERVICE_CONNECTION_ATTEMPT` or something similar. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Hmm right, I don't think that `n` would get too big here... But maybe having a maximum would make sense yeah. Probably a bigger number would be ok IMO (maybe 100?).

tonidero (commented):
Just a comment about a possible rename to this new event but looks good otherwise!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
JayShortway: Alright, capped at 100 here: https://github.com/RevenueCat/purchases-android/pull/2267/commits/23aa91345d98e283b5827d974d0a856b99828549

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsEntryName.kt
JayShortway: - Renamed them as discussed, in this commit: https://github.com/RevenueCat/purchases-android/pull/2267/commits/2b04954d430b7a0840fe7789a62711115abcafd7
- Added `GOOGLE_BILLING_START_CONNECTION` here: https://github.com/RevenueCat/purchases-android/pull/2267/commits/0d5df86af35be76aa20371eef5cc7c117e45e716

tonidero (approved):
Nice! Looks great :shipit:

PR #2268, opened 2025-03-18
[Diagnostics] Add `id` and `app_session_id` to events

tonidero:
### Description
This adds some new properties to diagnostics so we're able to better track events and what happens within the same session.

It also adds a small refactor to `DiagnosticsTracker` track events to avoid repetition to pass the app session id.


JayShortway (approved):
Makes sense!

ajpallares (approved):
Looks good!

PR #2269, opened 2025-03-18
[Diagnostics] Add play_store_version and play_services_version to all google events

tonidero:
### Description
This adds `play_store_version` and `play_services_version` to all google events.


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -44,6 +46,15 @@ internal class DiagnosticsTracker(
         const val BILLING_DEBUG_MESSAGE = "billing_debug_message"
     }
 
+    private val commonProperties = if (appConfig.store == Store.PLAY_STORE) {
+        mapOf(
+            "play_store_version" to appConfig.playStoreVersionName,
+            "play_services_version" to appConfig.playServicesVersionName,
```
tonidero, line 52: Originally, I was planning to also add the `billing_client_version` as was in the doc, but I don't think we can get the version at runtime that I could figure out... We could hardcode it somewhere, but then, it basically means we already have the billing client version from the sdk version and we can obtain this info in the backend/DW.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -44,6 +46,15 @@ internal class DiagnosticsTracker(
...
+        ).filterNotNullValues()
+    } else {
+        null
```
ajpallares, line 55: Nit: since we're falling back to `emptyMap()` anyway, why not have it here so that it's cleaner where it's used?
```suggestion
        emptyMap()
```

ajpallares (approved):
Looks good! Just a small nit that you can totally ignore 😄

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
JayShortway: Just mentioning that we could quite easily create a `BuildConfig` field set to `libs.versions.billing.get()`. That doesn't require any more hardcoding than we currently do (i.e. specifying the version in `libs.versions.toml`). But you're right that we can also infer this info from the SDK version.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -243,12 +254,13 @@ internal class DiagnosticsTracker(
...
                 name = eventName,
-                properties = properties,
+                properties = (commonProperties ?: emptyMap()) + properties,
                 appSessionID = appSessionID,
             ),
         )
     }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     fun trackEvent(diagnosticsEntry: DiagnosticsEntry) {
```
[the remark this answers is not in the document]
tonidero: Unfortunately, it is currently used in some tests... For example: https://github.com/RevenueCat/purchases-android/blob/fb33fa4f539209fc8ea2de88008f7dd42a96dcc3/purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerFunctionalTest.kt#L70. 

We could totally refactor this to avoid making this internal, but didn't want to add further refactors in an unrelated PR. I just added this annotation because I noticed this was only used externally in tests and seemed easy to confuse with the other new `trackEvent` method. 

Can do this change in a follow-up PR though.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
tonidero: Hmm that's a good point... But still seems like duplicated info... If we could figure out if the version didn't match at runtime (like for example if there was a bump because a dev was using a newer version of the billing client separately). But I think this way we can just have this information in the DW and even join if we need a view with this information.

JayShortway: Yea you're right. Btw the billing library leaves a `billing.properties` file in the final APK with the version that's being used. We might be able to read that? 👀 
image

Something we could look into.

JayShortway (approved):

PR #2270, opened 2025-03-18
Add `onManagementOptionSelected` to `CustomerCenterListener`

vegaro:
Equivalent of https://github.com/RevenueCat/purchases-ios/pull/4872/files adding a listener that gets triggered as soon as an option is selected

### DIFF
--- a/purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterManagementOption.kt
+++ b/purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterManagementOption.kt
@@ -3,9 +3,9 @@ package com.revenuecat.purchases.customercenter
 import android.net.Uri
 
 /**
- * Sealed interface representing different customer center management options.
+ * Interface representing different customer center management options.
  */
-sealed interface CustomerCenterManagementOption {
+interface CustomerCenterManagementOption {
     /**
      * Action to cancel the current operation
      */
--- a/purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterManagementOption.kt
+++ b/purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterManagementOption.kt
@@ -1,0 +1,24 @@
+package com.revenuecat.purchases.customercenter
+
+import android.net.Uri
+
+/**
+ * Sealed interface representing different customer center management options.
+ */
+sealed interface CustomerCenterManagementOption {
+    /**
+     * Action to cancel the current operation
+     */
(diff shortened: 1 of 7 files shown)


purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterManagementOption.kt
```diff
@@ -0,0 +1,24 @@
+package com.revenuecat.purchases.customercenter
+
+import android.net.Uri
+
+/**
+ * Sealed interface representing different customer center management options.
+ */
+sealed interface CustomerCenterManagementOption {
```
JayShortway, line 8: Just making sure that using `sealed` is a conscious decision? What are we going to do when adding an `*Option`? The alternative would be to drop the `sealed`:
```suggestion
interface CustomerCenterManagementOption {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -683,4 +685,25 @@ internal class CustomerCenterViewModelImpl(
...
+                path.url?.let {
+                    CustomerCenterManagementOption.CustomUrl(Uri.parse(it))
+                }
+
+            else -> null
+        }
+        action?.let {
+            listener?.onManagementOptionSelected(it)
+            purchases.customerCenterListener?.onManagementOptionSelected(it)
+        }
```
JayShortway, line 707: Nitpick, but since we're not returning anything from the `let`, just checking for null might be more readable? 

```kotlin
if (action != null) {
    // `action` is smart-casted to non-null here
    listener?.onManagementOptionSelected(action)
    purchases.customerCenterListener?.onManagementOptionSelected(action)
}
```

JayShortway (commented):
Looking good! Just making sure `sealed` is what we want.

purchases/src/main/java/com/revenuecat/purchases/customercenter/CustomerCenterManagementOption.kt
vegaro: oh right... I added it so it could be used in `when` but yeah, same problems as with the enums when adding a new one. Good point

JayShortway: A regular `interface` can also be used in a `when`, but the `else` is mandatory in that case.

PR #2271, opened 2025-03-19
[Paywalls V2] Ignores missing font alias if it's blank.

JayShortway:
## Bug
A customer reported their paywall failed to render with the following validation error in the logs: 
> Aliased font ‘’ does not exist.

## Cause
The real cause seems to be on the dashboard, as their paywall was able to be published without any Font family selected: 
image

## Fix
While this is ideally fixed on the dashboard, this PR introduces a behavior change in the SDK that will treat this situation as "no font selected". The SDK now ignores blank font aliases. This is a bit more defensive and less strict, but still valid behavior imo.


tonidero (approved):
:shipit:

PR #2276, opened 2025-03-20
[Diagnostics] add `is_retry` to `http_request_performed` event

ajpallares:
### Checklist

### Description
Adds the `is_retry` parameter to the `http_request_performed` event


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -126,7 +126,13 @@ internal class HTTPClient(
             callResult = performCall(baseURL, endpoint, body, postFieldsToSign, requestHeaders, refreshETag)
             callSuccessful = true
         } finally {
-            trackHttpRequestPerformedIfNeeded(endpoint, requestStartTime, callSuccessful, callResult)
+            trackHttpRequestPerformedIfNeeded(
+                endpoint,
+                requestStartTime,
+                callSuccessful,
+                callResult,
+                isRetry = refreshETag,
```
ajpallares, line 134: I''m not 100% sure that this is what we intend for the `is_retry` parameter.

I think that in iOS the logic for retrying requests is different and requests are retried upon receiving a 429 status code unless the `"Is-Retryable"` header from the server is set to `false`.

In Android, I think a retry happens when the server returns a 304 (`NOT_MODIFIED`) but we can't get the information from the cache.

Of course, I could be wrong about this, but that's what I understood from reading the code. 

If I'm right, should we do something to unify the two behaviors? @RevenueCat/coresdk

ajpallares: Thank you for the explanation @tonidero!

I've created the Linear ticket for Android 👍 

> the backend will only return a 304 if there is a cached value with that specific etag... so it's likely mostly unused, but seems like it might still be good to have.

It's not clear to me, should I create a ticket to also implement the existing Android logic for a 304 in iOS, or as it's basically unused, it's not worth it?

PR #2277, opened 2025-03-20
[Paywalls V2] Avoids adding spurious digits to per-period prices

JayShortway:
## Bug
A customer reported that the value for `{{ product.price_per_month }}` in JPY incorrectly has 2 decimals, e.g. "¥20.00". Other amounts did correctly have 0 decimals. 

## Cause
We have [hardcoded a few countries in the backend](https://github.com/RevenueCat/khepri/pull/9770) for which to drop decimals. However, these countries are only hardcoded to override store behavior. Japan is not part of this list, as the Play Store correctly provides formatted prices without decimals. The problem occurs when we perform calculations on these prices, e.g. to calculate a per-period price. We were not taking this situation into account, and added the spurious digits.

## Fix
The fix is to set the number of digits we want before formatting the amount.


tonidero (approved):
Nice fix!

PR #2278, opened 2025-03-20
Downgrades Emerge Snapshots sdk to 1.3.0.

JayShortway:
Drafty.


ajpallares (approved):
Let's try this. Thanks!

PR #2280, opened 2025-03-21
[Diagnostics] add `GET_OFFERINGS_STARTED` and `GET_OFFERINGS_RESULT` events

ajpallares:
This PR adds the events:

`GET_OFFERINGS_STARTED`. Will be sent at the start of a `getOfferings` API call.
`GET_OFFERINGS_RESULT`. Will be sent when returning the result of a `getOfferings` API call.


purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
```diff
@@ -28,19 +37,54 @@ internal class OfferingsManager(
...
+        val startTime = dateProvider.now
+        val onErrorWithTracking: (PurchasesError, DiagnosticsTracker.CacheStatus) -> Unit = { error, cacheStatus ->
+            trackGetOfferingsResultIfNeeded(startTime, cacheStatus, error, null, null)
+            onError?.invoke(error)
+        }
+        val onSuccessWithTracking: (OfferingsResultData, DiagnosticsTracker.CacheStatus) -> Unit =
+            { result, cacheStatus ->
+                trackGetOfferingsResultIfNeeded(
+                    startTime,
+                    DiagnosticsTracker.CacheStatus.NOT_CHECKED,
```
tonidero, line 50: I think this should be 
```suggestion
                    cacheStatus,
```

tonidero (commented):
Looking good! Just a fix

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
ajpallares: Right! Thanks for catching! It was a last-minute refactor to please `detekt` and I totally missed it. Thanks again! 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
```diff
@@ -42,7 +41,18 @@ internal class OfferingsFactory(
                     productIds = allRequestedProductIdentifiers,
                     onCompleted = { productsById ->
                         try {
-                            logMissingProducts(allRequestedProductIdentifiers, productsById)
+                            val notFoundProductIds = allRequestedProductIdentifiers
+                                .filterNot { productsById.containsKey(it) }
+                                .toSet()
+                            notFoundProductIds
+                                .takeIf { it.isNotEmpty() }
+                                ?.let { missingProducts ->
```
tonidero, line 49: Hmm should this be? 
```suggestion
                                ?.also { missingProducts ->
```

I'm wondering if `notFoundProductIds` will have a non-null value with the `let` (maybe it's working not sure)

tonidero (approved):
Just a question to make sure of something, but I believe it's working!

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
tonidero: Ah right, sorry I misread it, this should be fine yeah!

PR #2283, opened 2025-03-24
[Diagnostics] add GET_PRODUCTS_STARTED and GET_PRODUCTS_RESULT events

ajpallares:
This PR adds the events:

`GET_PRODUCTS_STARTED`. Will be sent at the start of a `getProducts` API call.
`GET_PRODUCTS_RESULT`. Will be sent when returning the result of a `getProducts` API call.

Note that the deprecated `getSubscriptionSkus` and `getNonSubscriptionSkus` API calls will also trigger these events, as they currently use the same internal implementation.


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -883,17 +902,21 @@ internal class PurchasesOrchestrator(
...
                             typesRemaining,
                             collectedStoreProducts + storeProducts,
+                            nonNullStartTime,
                             callback,
                         )
                     }
                 },
                 onError = {
                     dispatch {
+                        trackGetProductsResult(nonNullStartTime, productIds, productIds, it)
```
ajpallares, line 912: In this case of error, we're sending all the `productIds` in the `notFoundProductIds` parameter of the diagnostics event to make the error case equal to the implementation in the iOS SDK (see [PurchasesOrchestrator.swift - Lines 325 tot 332](https://github.com/RevenueCat/purchases-ios/blob/203e32b45e5ac9489afe8dc58a4d5089f25cdebd/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L325-L332))

```swift
            let notFoundProductIds = productIdentifiersSet.subtracting(
                products.map { $0.map(\.productIdentifier) }.value.map { Set($0) } ?? []
            )
            let error = products.error
            self.trackProductsResultIfNeeded(requestedProductIds: productIdentifiersSet,
                                             notFoundProductIds: notFoundProductIds,
                                             error: error,
                                             startTime: startTime)
```

(where `notFoundProductIds` equals `productIdentifiersSet` minus the empty set `[]`, i.e. equals `productIdentifiersSet`).

purchases/src/test/java/com/revenuecat/purchases/PurchasesCommonTest.kt
```diff
@@ -1338,6 +1348,141 @@ internal class PurchasesCommonTest: BasePurchasesTest() {
         verify(exactly = 1) { mockDiagnosticsSynchronizer.syncDiagnosticsFileIfNeeded() }
     }
 
+    // endregion
```
tonidero, line 1351: Something that I started doing at some point was to add the same naming in the `endregion` as the start region, so it's easier to note when things start.
```suggestion
    // endregion Diagnostics sync
```
Not a big problem though, most code isn't actually doing this 😅

tonidero (approved):
Looks great!

purchases/src/test/java/com/revenuecat/purchases/PurchasesCommonTest.kt
ajpallares: Oh, that's great to know! I'll do that, for sure. Thanks!

PR #2289, opened 2025-03-24
[Diagnostics] add GET_CUSTOMER_INFO_STARTED and GET_CUSTOMER_INFO_RESULT events

ajpallares:
This PR adds the events:

`GET_CUSTOMER_INFO_STARTED`. Will be sent at the start of a `getCustomerInfo` API call.
`GET_CUSTOMER_INFO_RESULT`. Will be sent when returning the result of a `getCustomerInfo` API call.


purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfoHelper.kt
```diff
@@ -3,21 +3,48 @@ package com.revenuecat.purchases
...
+import kotlin.time.Duration
 
+/**
+ * Wrapper for [ReceiveCustomerInfoCallback] to hold extra information for diagnostics.
+ */
+internal class ReceiveCustomerInfoFullCallback(
+    private val onCustomerInfoReceived: (CustomerInfo, Boolean?) -> Unit,
+    private val onError: (PurchasesError, Boolean?) -> Unit,
+) : ReceiveCustomerInfoCallback {
+    var hadUnsyncedPurchases: Boolean? = null
```
tonidero, line 28: Hmm this pattern of a callback with state seems weird to me... I think I would prefer to keep this state samewhere else... Maybe we can make the `CustomerInfoHelper` return a different data type that contains all the data we need, and we just map that into the `ReceiveCustomerInfoCallback` in the `PurchasesOrchestrator`. This would allow us to remove the `ReceiveCustomerInfoFullCallback` as well. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -556,18 +556,20 @@ internal class PurchasesOrchestrator(
     fun getCustomerInfo(
         callback: ReceiveCustomerInfoCallback,
     ) {
-        getCustomerInfo(CacheFetchPolicy.default(), callback)
+        getCustomerInfo(CacheFetchPolicy.default(), false, callback)
```
tonidero, line 586: For readability, might be good to specify the parameter name here
```suggestion
        getCustomerInfo(CacheFetchPolicy.default(), trackDiagnostics = false, callback)
```

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -556,18 +556,20 @@ internal class PurchasesOrchestrator(
     fun getCustomerInfo(
```
tonidero, line 583: Hmm just noticed this one is only called internally. Might be worth making this private?
```suggestion
    private fun getCustomerInfo(
```

purchases/src/main/kotlin/com/revenuecat/purchases/PostPendingTransactionsHelper.kt
```diff
@@ -23,12 +23,12 @@ internal class PostPendingTransactionsHelper(
 
     fun syncPendingPurchaseQueue(
         allowSharingPlayStoreAccount: Boolean,
-        onError: ((PurchasesError) -> Unit)? = null,
-        onSuccess: ((CustomerInfo?) -> Unit)? = null,
+        onError: ((PurchasesError, Boolean?) -> Unit)? = null,
+        onSuccess: ((CustomerInfo?, Boolean?) -> Unit)? = null,
```
tonidero, line 27: Is it needed to propagate whether there were unsynced purchases from here? I believe that if we call the `onSuccess` with `null`, that meant there were no unsynced purchases right? It might be good to make it more explicit, but don't know if we want the extra complexity for that... Might be worth just documenting the behavior as part of this method.

tonidero (commented):
Looking good! I'm not sure about the new callback with state though... I would try to propagate the data with a different data object if we need to instead. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/PostPendingTransactionsHelper.kt
ajpallares: > I believe that if we call the `onSuccess` with `null`, that meant there were no unsynced purchases right?

Not really. In this case there's a differentiation between
`true` --> there were unsynced purchases
`false` --> there were no unsynced purchases
`null` --> it was not checked whether or not there were unsynced purchases

> Is it needed to propagate whether there were unsynced purchases from here?

I really think so. This `syncPendingPurchaseQueue` method is where this information is fetched and used. So in the callbacks of this method we need to know not only the result (PurchasesError or CustomerInfo) but also whether there were unsynced purchases.

ajpallares:
> I'm not sure about the new callback with state though... I would try to propagate the data with a different data object if we need to instead.

Yes, I see your point. Tbh, I was not very convinced about the callback wrapper either, but it was the best solution I could find without making bigger changes to the methods. Perhaps I need to use the `Result` `sealed class` to make all this cleaner. Initially, I didn't take this approach because it seems too _swifty_ to me, but it would probably be a better approach.

Let me rethink this all. Thanks so much for the detailed review @tonidero!

tonidero (approved):
This makes sense to me! Much more explicit. Thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfoHelper.kt
```diff
@@ -26,66 +47,124 @@ internal class CustomerInfoHelper(
...
+        val startTime = dateProvider.now
+
+        val callbackWithDiagnostics: ((CustomerInfoDataResult) -> Unit)? = if (callback != null || trackDiagnostics) {
+            {
+                    customerInfoDataResult ->
+                trackGetCustomerInfoResultIfNeeded(trackDiagnostics, startTime, customerInfoDataResult, fetchPolicy)
+
+                when (customerInfoDataResult.result) {
+                    is Result.Success -> {
+                        callback?.onReceived(customerInfoDataResult.result.value)
```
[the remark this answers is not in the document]
ajpallares: Yep, makes total sense. This way we avoid the `when` when `callback` is `null`. 
Done in https://github.com/RevenueCat/purchases-android/commit/284b5a667642fc7a442d511596d91d56f00e087f

PR #2291, opened 2025-03-25
[Diagnostics] add `PURCHASE_STARTED` and `PURCHASE_RESULT` events

ajpallares:
This PR adds the events:

`PURCHASE_STARTED`. Will be sent at the start of a `purchase` API call.
`PURCHASE_RESULT`. Will be sent when returning the result of a `purchase` API call.


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1128,13 +1158,36 @@ internal class PurchasesOrchestrator(
...
-            purchaseCallback.dispatch(
-                PurchasesError(
-                    PurchasesErrorCode.PurchaseNotAllowedError,
-                    PurchaseStrings.UPGRADING_INVALID_TYPE,
-                ).also { errorLog(it) },
-            )
+            PurchasesError(
+                PurchasesErrorCode.PurchaseNotAllowedError,
+                PurchaseStrings.UPGRADING_INVALID_TYPE,
+            ).also { errorLog(it) }.also { callbackWithDiagnostics.dispatch(it) }
```
ajpallares, line 1190: Doing this to avoid the linter's error due to `LongMethod`. I needed one less line 😄

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1333,7 +1386,32 @@ internal class PurchasesOrchestrator(
         )
     }
 
-    // endregion
+    private fun trackPurchaseStarted(productId: String, productType: ProductType) {
```
tonidero, line 1389: Not a huge deal but should we  add the `IfNeeded` suffix to these functions? I see that wasn't added to the functions above, but we've been adding it in other places... I think it might be worth it but NABD.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1128,13 +1158,36 @@ internal class PurchasesOrchestrator(
         isPersonalizedPrice: Boolean?,
         purchaseCallback: PurchaseCallback,
     ) {
+        trackPurchaseStarted(purchasingData.productId, purchasingData.productType)
+        val startTime = dateProvider.now
+
+        val callbackWithDiagnostics = object : PurchaseCallback {
```
tonidero, line 1164: We could potentially do a small optimization that if `diagnosticsTrackerIfEnabled` is null, we can just fallback to the `purchaseCallback` to avoid the extra callback. Not a huge deal though.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -529,3 +567,10 @@ internal class DiagnosticsTracker(
         diagnosticsDispatcher.enqueue(command = command)
     }
 }
+
+val ProductType.diagnosticsName: String
```
tonidero, line 571: this should be `private` right?
```suggestion
private val ProductType.diagnosticsName: String
```

tonidero (approved):
Amazing, thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: Yeah, I see. It makes sense. I'll do it!

Related: can you think of any way to avoid the duplicated code (this `callbackWithDiagnostics` and the above `listenerWithDiagnostics` are the same). I could think of some solutions but I think they would be overcomplicated to what they achieve, tbh 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: Done in https://github.com/RevenueCat/purchases-android/pull/2291/commits/4816e6619d3f1f112590beefb5c963ee860ad86a

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: Right! I was thinking of more complicated solutions for some reason... 🤦
Well, I'm glad I asked 😅 
Thank you!

PR #2296, opened 2025-03-26
Wrap Billing client crash with ours with more information

tonidero:
### Description
Billing client 7.1.x introduced a crash in some older android devices due to it not finding a class:
```
FATAL EXCEPTION: main
Process: com.revenuecat.purchases_sample, PID: 20072
java.lang.NoClassDefFoundError: com.android.billingclient.api.zzce
	at com.android.billingclient.api.BillingFlowParams$Builder.build(com.android.billingclient:billing@@7.1.1:3)
```
This gets fixed downgrading to Billing client 7.0.0 or enabling core library desugaring (preferred option): https://developer.android.com/studio/write/java8-support#groovy 

Since this crash is difficult to understand, we are wrapping the relevant errors with our own. This is how it looks like now:
```
FATAL EXCEPTION: main
Process: com.revenuecat.purchases_sample, PID: 344
com.revenuecat.purchases. NoCoreLibraryDesugaringException: Error building BillingFlowParams for subscriptions due to an issue in Google's Billing Client library. Please check TODO-LINK to fix this issue.
	at com.revenuecat.purchases.google.BillingWrapper.buildSubscriptionPurchaseParams(BillingWrapper.kt:959)
	at com.revenuecat.purchases.google.BillingWrapper.buildPurchaseParams(BillingWrapper.kt:893)
	at com.revenuecat.purchases.google.BillingWrapper.access$buildPurchaseParams(BillingWrapper.kt:80)
	at com.revenuecat.purchases.google.BillingWrapper$makePurchaseAsync$2.invoke(BillingWrapper.kt:296)
	at com.revenuecat.purchases.google.BillingWrapper$makePurchaseAsync$2.invoke(BillingWrapper.kt:295)
	at com.revenuecat.purchases.google.BillingWrapper.executePendingRequests$lambda$3$lambda$2$lambda$1(BillingWrapper.kt:144)
	at com.revenuecat.purchases.google.BillingWrapper.$r8$lambda$g3n3vaI7ScqYgOQ9PPJyLRTbbfM(BillingWrapper.kt)
	at com.revenuecat.purchases.google.BillingWrapper$$ExternalSyntheticLambda2.run(D8$$SyntheticClass:0)
	at android.os.Handler.handleCallback(Handler.java:739)
	at android.os.Handler.dispatchMessage(Handler.java:95)
	at android.os.Looper.loop(Looper.java:148)
	at android.app.ActivityThread.main(ActivityThread.java:5619)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:853)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:737)
Caused by: java.lang.NoClassDefFoundError: com.android.billingclient.api.zzce
	at com.android.billingclient.api.BillingFlowParams$Builder.build(com.android.billingclient:billing@@7.1.1:3)
	at com.revenuecat.purchases.google.BillingWrapper.buildSubscriptionPurchaseParams(BillingWrapper.kt:955)
```


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -903,17 +904,26 @@ internal class BillingWrapper(
...
+                            setIsOfferPersonalized(it)
+                        }
                     }
-                }
-                .build(),
-        )
+                    .build(),
+            )
+        } catch (e: NoClassDefFoundError) {
+            // TODO: Add link
```
tonidero, line 920: Still need to add the links here

tonidero (commented):
Still need to add the links but this is ready for review

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesRuntimeException.kt
```diff
@@ -0,0 +1,3 @@
+package com.revenuecat.purchases
+
+class PurchasesRuntimeException(message: String, cause: Throwable) : RuntimeException(message, cause)
```
JayShortway, line 3: Just an idea: should we make this name even more specific? Something like `NoCoreLibraryDesugaringException` or something? We could then have the error message be a constant, and avoid having to define it where we throw this. FFTI

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -903,17 +904,26 @@ internal class BillingWrapper(
...
-        return Result.Success(
-            BillingFlowParams.newBuilder()
-                .setProductDetailsParamsList(listOf(productDetailsParamsList))
-                .setObfuscatedAccountId(appUserID.sha256())
-                .apply {
-                    isPersonalizedPrice?.let {
-                        setIsOfferPersonalized(it)
+        try {
+            return Result.Success(
+                BillingFlowParams.newBuilder()
```
JayShortway, line 909: Could we try calling `BillingFlowParams.newBuilder().build()` in `Purchases.configure()`, and log an error if that fails? Some developers might be forwarding logged errors as non-fatal exceptions to their crash reporting tools, and this would allow them to catch the issue earlier (and get way more reports).

JayShortway (commented):
Looks good, thanks for taking care of this! Just added 2 suggestions. Will hold of approval until the links are added, to avoid accidentally merging this. No blockers from my end.

ajpallares (commented):
Thanks for doing this. I think this is the best we can do. I agree with @JayShortway's comment about logging the issue at configuration time, if possible. I'm guessing that just building a `BillingFlowParams` instance shouldn't have other consequences.

ajpallares:
(sorry I deleted a @tonidero's comment by mistake 🙏)

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: But yeah, I think it's probably ok. Will add that.

tonidero: Hmm so actually, I don't think this is easily doable... We need to pass in some product details parameter to the builder, otherwise it fails with a different exception: `java.lang.IllegalArgumentException: Details of the products must be provided.`. This means we need to fetch the product details from a product from Google before doing this.

We could delay this until the first time we obtain a product from Google I guess... But I feel the point of this log becomes less important. Will hold on this for now but lmk if you have other thoughts!

JayShortway: Ah okay, yea that changes things. Just checking: does that `IllegalArgumentException` also precede the `NoClassDefFoundError` on affected devices?

tonidero: That's correct yeah. I tested on the reproducible device.

JayShortway: Thanks for confirming! My only concern is that this crashes when purchasing, which means that seeing this crash in the wild means (very likely) lost sales. Right now developers do not have any chance of catching this error before someone tries to purchase. If we log at product fetching time, there's at least a way to catch it sooner.

tonidero: Well, considering that they will likely only see this as non-fatals (since it's unlikely they will have one of the affected devices), they will likely hit the issue in any case even if they work on a fix before... But yeah I understand your thoughts, I will try to add a log first time we fetch products from Google.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -114,4 +119,37 @@ internal class QueryProductDetailsUseCase(
...
     }
+
+    // This error was introduced in Google's Billing Client 7.1.0 and 7.1.1. Logging an error to notify the developer,
+    // as soon as possible that purchases will fail unless core library desugaring is enabled.
+    private companion object {
+        val hasLoggedBillingFlowParamsError = AtomicBoolean(false)
+    }
+
+    @Suppress("NestedBlockDepth")
+    private fun logErrorIfIssueBuildingBillingParams(productDetails: List<ProductDetails>) {
```
tonidero, line 130: I'm not really a fan of this code... But I think we shouldn't be doing this for every product fetch either... I would kinda prefer not to include it in the final solution... Wdyt @RevenueCat/coresdk?

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -114,4 +119,37 @@ internal class QueryProductDetailsUseCase(
...
+            if (hasLoggedBillingFlowParamsError.getAndSet(true)) {
+                return
+            }
+            val productDetailsParams = try {
+                val offerToken = it.subscriptionOfferDetails?.firstOrNull()?.offerToken
+                ProductDetailsParams.newBuilder()
+                    .setProductDetails(it)
+                    .apply { if (offerToken != null) setOfferToken(offerToken) }
+                    .build()
+            } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
```
tonidero, line 141: Added this just in case there is something I've missed... I've tested this with subscriptions and one time products, but I don't like how this has a bunch of validations internally that we need to acomodate or crash... This catch is to make sure we don't accidentally break.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -114,4 +119,37 @@ internal class QueryProductDetailsUseCase(
...
+                null
+            }
+
+            if (productDetailsParams == null) return
+
+            try {
+                BillingFlowParams.newBuilder().setProductDetailsParamsList(listOf(productDetailsParams)).build()
+            } catch (e: NoClassDefFoundError) {
+                errorLog(NO_CORE_LIBRARY_DESUGARING_ERROR_MESSAGE, e)
+            }
```
ajpallares, line 152: In addition to `NoClassDefFoundError`, shouldn't we also catch here the more generic error just in case? I don't know if it makes sense, but to make sure that we never throw an error from this safety check method (i.e. same reason as above)

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
ajpallares: I agree with the flag to only do this check once. However, I'm not sure I understand your comment. 

> I would kinda prefer not to include it in the final solution...

Do you mean not to include this whole `logErrorIfIssueBuildingBillingParams` check or not to include the `hasLoggedBillingFlowParamsError` flag part?

ajpallares (commented):
It looks good to me. Just a couple of comments. Mainly, perhaps we should make sure that no exception escapes this method.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
tonidero: I'm doubting the whole `logErrorIfIssueBuildingBillingParams`. I think the small advantage we can get from logging slightly before crashing (when fetching the products instead of when purchasing), does not warrant this extra complexity for all users of the SDK IMO.

JayShortway: Thanks for exploring this path Toni! I agree that it is not as clean as I would have hoped haha. It could be useful for everyone, as it's a check of "can we even purchase" at product-fetching time. However I wouldn't mind if you want to delete this code. I think it's 50/50 for me haha.

ajpallares: Right! Thanks so much for exploring this Toni 🙏

I see your point. The amount of people affected by this should be very small and it scares me a bit that this whole check would run for everyone...

I know that it's not the same, but could we do something alternative with diagnostics? Like "catch the exception, store the specific diagnostics event, and letting the crash happen afterwards" (which would be synced in the next app launch, I guess). I know that it would be a whole different solution, as we're the ones receiving diagnostics instead, but at least we could be more aware of the actual magnitude of this issue... Idk, just throwing the idea, but FFTI.

tonidero: The main problem with diagnostics is that it's opt-in, so it would be difficult to understand the magnitude of the issue in any case, considering only a minority of apps have it enabled it. I guess we could add a tracking event to try to track with our existing apps, but then devs might take a while to update to the latest and possibly enable diagnostics for us to detect it :(

ajpallares: Yeah... makes sense :(

tonidero:
Ready for another pass @RevenueCat/coresdk

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
JayShortway: Works for me!

ajpallares: Ok! Thanks!

ajpallares (approved):
Thanks for the fix and for the iterations! 🙌

PR #2298, opened 2025-03-27
feat: Change default copy for web_subscription_manage #4921

facumenzella:
### Motivation
The copy is confusing for web billing.


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -250,8 +250,8 @@ data class CustomerCenterConfigData(
                     YOU_HAVE_PROMO -> "You've been granted a subscription that doesn’t renew"
                     YOU_HAVE_LIFETIME -> "Your active lifetime subscription"
                     WEB_SUBSCRIPTION_MANAGE ->
-                        "You have an active subscription that was created on the web. " +
-                            "You can manage your subscription by visiting your account."
+                        "You have an active subscription that was purchased on the web. " +
+                            "You can manage your subscription using the button below."
```
vegaro, line 254: Sweet. This makes more sense. We should also split the logic for Web Billing and Stripe. 

Also, maybe we should remove the `subscription` in case the purchase was a lifetime purchase?

vegaro (approved):

tonidero:
Going to change the labels here to `pr:other`. This doesn't feel like a feat to our customers IMO.

PR #2302, opened 2025-03-27
[Paywalls V2] Template previews have images now

JayShortway:
Draft.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,6 +215,7 @@ jobs:
     <<: *android-executor
     shell: /bin/bash --login -o pipefail
     steps:
+      - run: which -a git-lfs
       - checkout
       - install-sdkman
       - android/accept-licenses
@@ -640,6 +641,7 @@ jobs:
     description: "Run RevenueCatUI UI tests for Android in CircleCI"
     <<: *android-machine-emulator
     steps:
+      - run: which -a git-lfs
       - checkout
       - install-sdkman
       - android/create-avd:
@@ -657,13 +659,14 @@ jobs:
     docker:
       - image: cimg/ruby:3.2.0
     steps:
+      - run: which -a git-lfs
       - checkout
       - revenuecat/install-gem-unix-dependencies:
           cache-version: v1
       - revenuecat/setup-git-credentials
-      - run:
-          name: Update paywall templates
-          command: bundle exec fastlane android update_paywall_templates
+#      - run:
+#          name: Update paywall templates
+#          command: bundle exec fastlane android update_paywall_templates
 
 workflows:
   version: 2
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -176,6 +176,36 @@ commands:
             echo 'rvm use << parameters.version >>' >> $BASH_ENV
             source $BASH_ENV
             ruby -v
+  
+  checkout-with-lfs:
```
tonidero, line 180: Might be good to move this to [our orb](https://github.com/RevenueCat/sdks-circleci-orb)

fastlane/Fastfile
```diff
@@ -585,9 +585,14 @@ DESC
   desc "Updates offerings_paywalls_v2_templates.json file and creates/updates PR"
   lane :update_paywall_templates do |options|
     branch_name = "pw2-update-templates-json"
-    templates_file = "../ui/revenuecatui/src/debug/resources/offerings_paywalls_v2_templates.json"
+    templates_dir = "../ui/revenuecatui/src/debug/resources/paywall-templates"
+    templates_file = File.join(templates_dir, "offerings_paywalls_v2_templates.json")
     api_key = ENV["PAYWALLS_V2_TEMPLATE_REPO_API_KEY"] || UI.user_error!("Missing PAYWALLS_V2_TEMPLATE_REPO_API_KEY environment variable")
 
+    # Recreate our output directory so we start fresh.
```
tonidero, line 592: Just curious, is this needed? Or a just in case thing?

fastlane/Fastfile
```diff
@@ -673,6 +684,96 @@ DESC
     end
   end
 
+  private_lane :download_paywall_assets do |options|
```
tonidero, line 687: This seems like it should also belong to our plugin to share with iOS, but we can extract later.

.circleci/config.yml
JayShortway: Yea that's a good idea! I prefer to extract it later though, as it hasn't really been put through its paces yet. I've tested it locally, but would like to see the template update PR use this properly. If it's all working correctly, I can add it to the orb. Let me know what you think!

fastlane/Fastfile
JayShortway: This is done to avoid leaving stale images in the repo if no template uses that image anymore. For instance, if a template is updated to use a different icon, we don't want to keep the old icon in the repo. This job downloads everything anyway, so in terms of bandwidth it doesn't really matter if we clear our output directory or not.

fastlane/Fastfile
JayShortway: Yea definitely! Will extract all of this once we know what the iOS solution will look like. This particular action is somewhat opinionated right now about the folder structure, for instance, because that's what the Compose Preview will look for. So we gotta see if that works for iOS too or if we need to make some adjustments.

.circleci/config.yml
JayShortway: For posterity: https://github.com/RevenueCat/sdks-circleci-orb/pull/11.

PR #2303, opened 2025-03-28
Update AGP to 8.9.1

tonidero:
### Description
This updates AGP to the latest version + the gradle wrapper.


JayShortway (approved):

tonidero:
Ughh, seems that the databinding library was updated to kotlin 2.1.0 and that's not compatible with our Kotlin 1.8.0... Will need to research if I can downgrade that version somehow, or just remove usage of the databinding library (it works fine with paywall tester). Will hold this for now.

PR #2304, opened 2025-03-28
[Diagnostics] Sync diagnostics if file reaches lower limit

tonidero:
### Description
With all the new events we've been adding, we can potentially reach the file size limit much more easily, specially considering we only sync once per app open. This changes the logic, so we also sync if, after tracking an event, the file reaches a lower file size than the maximum. This should help us receiving events more frequently and avoid losing those events.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
@@ -2,12 +2,15 @@ package com.revenuecat.purchases.common.diagnostics
 
 import android.content.Context
 import android.content.SharedPreferences
+import android.os.Build
 import androidx.annotation.VisibleForTesting
 import com.revenuecat.purchases.common.Backend
 import com.revenuecat.purchases.common.Dispatcher
+import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.verboseLog
 import org.json.JSONObject
 import java.io.IOException
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * This class is in charge of syncing all previously tracked diagnostics. All operations will be executed
@@ -22,7 +25,7 @@ internal class DiagnosticsSynchronizer(
     private val diagnosticsTracker: DiagnosticsTracker,
     private val backend: Backend,
     private val diagnosticsDispatcher: Dispatcher,
-) {
+) : DiagnosticsEventTrackerListener {
     companion object {
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
         const val MAX_NUMBER_POST_RETRIES = 3
@@ -37,13 +40,20 @@ internal class DiagnosticsSynchronizer(
             )
     }
 
+    val isSyncing = AtomicBoolean(false)
+
     fun syncDiagnosticsFileIfNeeded() {
         enqueue {
             try {
+                if (isSyncing.getAndSet(true)) {
+                    verboseLog("Already syncing diagnostics file.")
+                    return@enqueue
+                }
                 val diagnosticsList = getEventsToSync()
                 val diagnosticsCount = diagnosticsList.size
                 if (diagnosticsCount == 0) {
                     verboseLog("No diagnostics to sync.")
+                    isSyncing.set(false)
                     return@enqueue
                 }
                 backend.postDiagnostics(
@@ -52,6 +62,7 @@ internal class DiagnosticsSynchronizer(
                         verboseLog("Synced diagnostics file successfully.")
                         diagnosticsHelper.clearConsecutiveNumberOfErrors()
                         diagnosticsFileHelper.clear(diagnosticsCount)
+                        isSyncing.set(false)
                     },
                     onErrorHandler = { error, shouldRetry ->
                         if (shouldRetry) {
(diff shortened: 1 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
```diff
@@ -75,6 +79,8 @@ internal class DiagnosticsTracker(
         emptyMap()
     }
 
+    var listener: DiagnosticsEventTrackerListener? = null
```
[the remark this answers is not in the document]
ajpallares: I think this is OK. However, would this not cause memory leak due to a strong reference cycle? In iOS, it would, so this one should be `weak`.

ajpallares (approved):
Looks great! Just a question that probably does not apply to Kotlin (you know, me coming from Swift...)

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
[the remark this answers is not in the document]
tonidero: Right, that's not a problem in Android. As long as an object (or a collection of them referencing each other) is not held in the object tree from a top-level object (in this case the static purchases instance), it will be released by the GC.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
```diff
@@ -85,6 +97,25 @@ internal class DiagnosticsSynchronizer(
...
+    private fun syncDiagnosticsFileIfBigEnough() {
+        enqueue {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                if (diagnosticsFileHelper.isDiagnosticsFileBigEnoughToSync()) {
+                    verboseLog("Diagnostics file is big enough to sync. Syncing it.")
+                    syncDiagnosticsFileIfNeeded()
+                }
+            } else {
+                // This should never happen since we create this class only if diagnostics is supported
+                errorLog("Diagnostics only supported in Android 24+")
```
JayShortway, line 118: Unrelated, but what is the reason for this?

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
```diff
@@ -37,13 +40,20 @@ internal class DiagnosticsSynchronizer(
...
 
+    val isSyncing = AtomicBoolean(false)
+
     fun syncDiagnosticsFileIfNeeded() {
         enqueue {
             try {
+                if (isSyncing.getAndSet(true)) {
+                    verboseLog("Already syncing diagnostics file.")
+                    return@enqueue
+                }
```
JayShortway, line 51: What happens if an event is tracked while we're syncing?

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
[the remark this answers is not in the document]
JayShortway: Can this be a constructor parameter?

JayShortway (approved):
Makes sense to sync more often! Questions only haha

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt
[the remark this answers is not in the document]
tonidero: Not really... problem is that `DiagnosticsSyncronizer` already depends on the `DiagnosticsTracker` at constructor time, so we can't make `DiagnosticsTracker` depend on the `DiagnosticsSynchronizer at constructor time as well
since it's a circular dependency at constructor time. Lmk if you have other ideas!

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
tonidero: If it's already syncing, it would still be added to the diagnostics file, but not initiate another sync. When the sync in progress finishes, it should clear only the synced events leaving any new events in the file. If after syncing, it's still bigger than the 200 kb size we've set, it would sync on the next event tracked (or upon reconfiguration)

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
tonidero: Hmm right the `DiagnosticsFileHelper` requires Android 24+ but looking at it, I think this is not needed anymore and we might have just forgotten to remove it from here: https://github.com/RevenueCat/purchases-android/pull/1944... Could be good to remove this, but that should be a separate PR.

purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
JayShortway: That's awesome, thanks for explaining!

PR #2307, opened 2025-03-31
Use coil-base instead of coil-compose for purchases package

skydoves:
### Motivation

Use `coil-base` dependency instead of `coil-compose` for the `purchases` package.

This isn't an issue on the client side at all. However, since it involves UI-related APIs (e.g, Jetpack Compose) at compile time, this work helps prevent potential mistakes on the SDK developer side

image

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -143,6 +143,7 @@ compose-constraintlayout = { module = "androidx.constraintlayout:constraintlayou
 window = { module = "androidx.window:window", version.ref = "window" }
 window-core = { module = "androidx.window:window-core", version.ref = "window" }
 
+coil-base = { module = "io.coil-kt:coil-base", version.ref = "coil" }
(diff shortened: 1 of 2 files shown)


JayShortway (approved):
Great find, thank you! 🫶 

(I think the integration tests are being fixed by @tonidero.)

tonidero (approved):
Right this makes sense! And yeah, fixing those tests [here](https://github.com/RevenueCat/purchases-android/pull/2308) (Still working in some other related tests that are failing but that should fix this issue)

tonidero:
That PR has been merged! So feel free to rebase/merge main and merge this!

skydoves:
Thank you so much for the quick review, @JayShortway, @tonidero! 💚 

By the way, I don't have permission to push to this repository yet, if someone else could grant the permission, that would be great.

aboedo:
@skydoves I think you should have access now, lmk if that's not the case! Thanks for doing this!

PR #2308, opened 2025-03-31
Fix load shedder backend integration tests

tonidero:
### Description
These broke with some changes to the load shedder integration tests project. The proper integration tests are still broken, but this fixes the immediate tests so we don't block things for everyone


JayShortway (approved):
Thanks for fixing!

PR #2309, opened 2025-03-31
Attempts to have Emerge Snapshots handle merge queue branches.

JayShortway:
## Description
GitHub's merge queue feature runs CI on intermediate branches with names looking like this: 
> gh-readonly-queue/main/pr-2305-de127b96b159da7def5cef15f51af329369eac92

Such a CI run is not a "PR" run, which means that the `CIRCLE_PULL_REQUEST` environment variable is not set. Fortunately we can extract the PR number from the branch name, which is implemented in this PR.


tonidero (approved):
I think this makes sense. That's some deep github logic there 😅. Let's try it! 🙏

PR #2310, opened 2025-03-31
Remove redundant kotlin stdlib dependencies

skydoves:
### Motivation

Starting from [Kotlin 1.4](https://kotlinlang.org/docs/reference/whatsnew14.html#dependency-on-the-standard-library-added-by-default), the dependency on the standard library added by default via `kotlin-android (org.jetbrains.kotlin.android) plugin`, so we can remove them:

> You no longer need to declare a dependency on the stdlib library in any Kotlin Gradle project, including a multiplatform one. The dependency is added by default. 
>
> The automatically added standard library will be the same version of the Kotlin Gradle plugin, since they have the same versioning.
>
> For platform-specific source sets, the corresponding platform-specific variant of the library is used, while a common standard library is added to the rest. The Kotlin Gradle plugin will select the appropriate JVM standard library depending on the kotlinOptions.jvmTarget compiler option of your Gradle build script.

### DIFF
--- a/examples/purchase-tester/build.gradle
+++ b/examples/purchase-tester/build.gradle
@@ -62,7 +62,6 @@ dependencies {
     implementation project(path: ':purchases')
     implementation project(path: ':feature:amazon')
 
-    implementation libs.kotlin.stdlib
     implementation libs.androidx.appcompat
     implementation libs.androidx.constraintlayout
     implementation libs.androidx.recyclerview
--- a/examples/MagicWeather/app/build.gradle.kts
+++ b/examples/MagicWeather/app/build.gradle.kts
@@ -53,7 +53,6 @@ android {
 dependencies {
     implementation(libs.revenuecat)
     implementation(libs.revenuecat.amazon)
-    implementation(libs.kotlin.stdlib)
     implementation(libs.androidx.core)
     implementation(libs.androidx.appcompat)
     implementation(libs.material)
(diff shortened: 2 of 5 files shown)


examples/purchase-tester/build.gradle
```diff
@@ -62,7 +62,6 @@ dependencies {
     implementation project(path: ':purchases')
     implementation project(path: ':feature:amazon')
 
-    implementation libs.kotlin.stdlib
```
JayShortway, line 65: We can probably also remove it from `MagicWeather/app/build.gradle.kts`, and then from the `libs.versions.toml` file as well?

JayShortway (approved):
Yes, perfect! Thanks for cleaning this up!

examples/purchase-tester/build.gradle
JayShortway: (This comment has nothing to do with the line it's attached to.)

skydoves: That's a great suggestion! 

Done: [d16a196](https://github.com/RevenueCat/purchases-android/pull/2310/commits/d16a196b1e7df00ab465d12814851e72294c3b4c)

JayShortway (approved):

examples/purchase-tester/build.gradle
JayShortway: Nice, thank you!

tonidero (approved):
Great! Thanks! 🫶 Merging this since I believe you still don't have access

PR #2311, opened 2025-03-31
[Paywalls V2] Renders images in template previews

JayShortway:
## Description
Adjusts the template previews so images are rendered.
- Adds `PaywallTemplateImageLoader`.
- Moves all image preview loading utilities to `helpers/ImagePreviews.kt`.
- Adjusts `rememberAsyncImagePainter` in `BackgroundStyle.kt` to support image previews, just like `RemoteImage` already does.


ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -133,6 +143,15 @@ private class OfferingProvider : PreviewParameterProvider<Offering> {
...
 
+/**
+ * If you're not seeing images in these previews and you want to, do the following.
+ * 1. Install [git-lfs](https://git-lfs.com/) using `brew install git-lfs` or
+ * `port install git-lfs`.
+ * 2. Run `git lfs install`.
+ * 3. Run `git lfs pull`.
+ *
+ * You'll only need to perform these steps once.
+ */
```
JayShortway, line 154: Would be great if someone without `git-lfs` on their system could verify these steps! 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/ImagePreviews.kt
```diff
@@ -0,0 +1,75 @@
...
+internal fun ProvidePreviewImageLoader(imageLoader: ImageLoader, content: @Composable () -> Unit): Unit =
+    CompositionLocalProvider(
+        LocalPreviewImageLoader provides imageLoader,
+        content,
+    )
+
+@JvmSynthetic
+@OptIn(ExperimentalPreviewRevenueCatUIPurchasesAPI::class)
+internal fun ImageLoader.getPreviewPlaceholderBlocking(imageRequest: ImageRequest): Painter =
+    when (val result = runBlocking { execute(imageRequest) }) {
```
tonidero, line 36: We should be careful about calling this in production... I guess the name already indicates that, but I wonder if we should split this between the debug/release variants to make sure this code doesn't make it into a release SDK/we don't use it accidentally... Not a blocker though, just a possible improvement I was thinking.

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
tonidero: Hmm so trying to `git pull origin main` now, I get an error: 
```
git-lfs filter-process: git-lfs: command not found
fatal: the remote end hung up unexpectedly
```
Seems that's a hard dependency to use the repo now :(

After I run `brew install git-lfs`, it works, and I'm able to see the templates in AS, didn't need to do anything else.

tonidero (approved):
This is very nice! I left a possible improvement to try to move some of the code away from the main source set... But I think it's looking good!

Also not from this PR, but might be good to figure out if it's possible to work without having `git-lfs` installed... Ideally it's not a hard dependency

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
```diff
@@ -185,14 +174,14 @@ private fun AsyncImage(
...
-                placeholder = if (isInPreviewMode()) imageLoader.getPreviewPlaceholder(imageRequest) else null,
+                placeholder = if (isInPreviewMode()) imageLoader.getPreviewPlaceholderBlocking(imageRequest) else null,
                 imageLoader = imageLoader,
                 contentScale = contentScale,
                 onError = { errorState ->
                     Logger.e("Error loading placeholder image", errorState.result.throwable)
                 },
             )
-        } ?: if (isInPreviewMode()) imageLoader.getPreviewPlaceholder(imageRequest) else null,
+        } ?: if (isInPreviewMode()) imageLoader.getPreviewPlaceholderBlocking(imageRequest) else null,
```
skydoves, line 184: What about adding a parameter `placeholder: Painter? = null` for the `AsyncImage` composable and hoisting the placeholder to the call site? We can keep the source of the truth for the placeholder condition.

```kotlin
val imageLoader = ..
val imageRequest = ..

val placeholder = if (isInPreviewMode()) imageLoader.getPreviewPlaceholderBlocking(imageRequest) else null

if (cachePolicy == CachePolicy.ENABLED) {
        AsyncImage(
                 ..,    
                 placeholder = placeholder,
        )
} else {
        AsyncImage(
                 ..,    
                 placeholder = placeholder,
        )
)
```

skydoves (commented):
This is nice work and looks good to me! I left a trivial comment that may improve the `placeholder` parameter for `AsyncImage`.

skydoves (approved):

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
JayShortway: Updated the base branch of this PR to `pw2-template-previews-submodule` which has migrated away from Git LFS and towards a submodule ([paywall-preview-resources](https://github.com/RevenueCat/paywall-preview-resources)). 

I uninstalled `git-lfs` from my machine, and I can check out this branch (`pw2-template-previews-images-2`) without issues. Could you double check on your end? 🙏

tonidero: Interesting... I uninstalled git-lfs and I can checkout the branch, but I got this message:
`This repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the 'post-checkout' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks').`
But yeah, I think this at least works!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/ImagePreviews.kt
JayShortway: Good idea! Done here: https://github.com/RevenueCat/purchases-android/pull/2311/commits/26bf446b35cd6864267da3c33ee794af13420d86

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/RemoteImage.kt
JayShortway: I like it! Implemented in https://github.com/RevenueCat/purchases-android/pull/2311/commits/0758bef92145876946936131b2c311b900430b36.

skydoves (approved):

PR #2313, opened 2025-04-01
Migrate purchases Gradle module to KTS

skydoves:
### Motivation

Migrate purchases Gradle module (groovy) to KTS (kotlin) format. Planning to gradually migrate to other modules as well.


purchases/build.gradle.kts
```diff
@@ -0,0 +1,197 @@
...
+                    uri("https://github.com/revenuecat/purchases-android/blob/main/public/src/main/java").toURL(),
+                )
+                remoteLineSuffix.set("#L")
+            }
+        }
+    }
+}
+
+// Remove afterEvaluate
+// after https://github.com/Kotlin/kotlinx-kover/issues/362 is fixed
```
tonidero, line 192: Looks like this was pretty outdated 😅 We should try to fix this at some point as well.

tonidero (approved):
Nice work!! This was long overdue. We can definitely take this step by step.

JayShortway (approved):
Thank you for taking care of this! I've been doing them one-by-one but it's been a while since I've had time to do another. This is so much better!

vegaro (approved):

PR #2314, opened 2025-04-01
[Paywalls V2] Make `radialGradient` use bigger size for radius calculation

tonidero:
### Description
In the dashboard + iOS, when using a radial gradient, we use the bigger size of the view we apply the gradient to. In Android however, by default, we apply it to the smaller side. This makes Android look like iOS + Dashboard

| Before | After |
| --- | --- |
| image | image |


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
```diff
@@ -249,25 +250,55 @@ private class RadialGradient(
...
+        }
 
-    override fun createShader(size: Size): Shader = brush.createShader(size)
+        val effectiveCenter = if (centerOffset == Offset.Unspecified) {
+            size.center
+        } else {
+            centerOffset
+        }
+
+        return RadialGradientShader(
```
tonidero, line 279: I believe it's fine to delay creating the shader itself in the `createShader` method 😅. Lmk if you have any concerns.

JayShortway: Wasn't that already happening? As in, is there a timing difference between the new implementation and what `brush.createShader(size)` does? 

In any case, I think this is fine!

JayShortway (approved):
Ah great! Thanks for fixing this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/ColorStyle.kt
JayShortway: Oh got it, thanks! Yea should be fine.

PR #2315, opened 2025-04-01
[Paywalls V2] Removes Git LFS and adds submodule for paywall preview resources

JayShortway:
Draft.


.circleci/config.yml
```diff
@@ -177,36 +177,17 @@ commands:
...
-      # If our restore_cache step restored the LFS files, this won't do anything.
-      - run:
-          name: Pull LFS files if needed
-          command: git lfs pull
-      - save_cache:
-          key: v1-git-lfs-cache-key-{{ checksum ".git-lfs-object-ids" }}
-          paths:
-            - .git/lfs
+            git submodule init << parameters.path >>
+            git submodule update << parameters.path >>
```
tonidero, line 190: Hmm I wonder... would it make sense to do `git submodule init` without a path? So it initializes any submodules? (same for updating). I don't know if there is any situation where we need to be precise here...

tonidero (approved):
Nice, this makes sense to me!

PR #2321, opened 2025-04-02
Add `X-Is-Backgrounded` header to all requests

tonidero:
### Description
Adds a header to determine whether the request was performed in the background or foreground. This can be useful for debugging in some situations.


purchases/src/main/kotlin/com/revenuecat/purchases/common/AppConfig.kt
```diff
@@ -34,6 +35,13 @@ internal class AppConfig(
     var forceSigningErrors: Boolean = forceSigningErrors
         get() = runningTests && field
 
+    private val _isAppBackgrounded: AtomicBoolean = AtomicBoolean(true)
+    var isAppBackgrounded: Boolean
+        get() = _isAppBackgrounded.get()
+        set(value) {
+            _isAppBackgrounded.set(value)
+        }
```
tonidero, line 43: TBH I doubted whether this belonged to `AppConfig`... it doesn't seem like a total fit and that class didn't have much mutable state until now. But I thought about holding this into a separate cache location and pass that cache to the `HTTPClient` (and wherever it's needed in the future), and it seemed like a lot of wiring... This felt like the lesser evil. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/AppConfig.kt
```diff
@@ -34,6 +35,13 @@ internal class AppConfig(
     var forceSigningErrors: Boolean = forceSigningErrors
         get() = runningTests && field
 
+    private val _isAppBackgrounded: AtomicBoolean = AtomicBoolean(true)
```
ajpallares, line 38: Just to confirm. Is the initial value of `true` correct? In other words, does the system calls `onAppForegrounded()` when launching the app?

ajpallares (approved):
Thanks for adding this! Looks great! Just one question to double check

purchases/src/main/kotlin/com/revenuecat/purchases/common/AppConfig.kt
ajpallares: Makes sense. I was sure you had thought about this, but wanted to double check. Thanks for the explanation!

purchases/src/main/kotlin/com/revenuecat/purchases/common/AppConfig.kt
JayShortway: Hmm yea I agree with your point that it doesn't really seem to fit in a class named `AppConfig`. Maybe we should split some stuff out of there and call it `AppState`? Or maybe we can just rename `AppConfig` to `AppState`? Just some ideas. I think it's fine to keep as is for this PR.

JayShortway (approved):
Nice!

PR #2325, opened 2025-04-03
fix: Check eligible for promo offers

facumenzella:
### Motivation
I was adding a log and noticed `eligible` was not being used.

### Description
Add eligibility check, and dedicated test to make sure it doesn't break again
Add a log when the user is not eligible for the promo offer, like https://github.com/RevenueCat/purchases-ios/pull/4949/files


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -435,6 +436,16 @@ internal class CustomerCenterViewModelImpl(
         promotionalOffer: CustomerCenterConfigData.HelpPath.PathDetail.PromotionalOffer,
         originalPath: CustomerCenterConfigData.HelpPath,
     ): Boolean {
+        if (!promotionalOffer.eligible) {
+            Log.d(
```
tonidero, line 440: Hmm I wonder if this should be a verbose log, since this would be a common case after all... NABD though

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterViewModelTests.kt
```diff
@@ -93,6 +94,49 @@ class CustomerCenterViewModelTests {
         clearAllMocks()
     }
 
+    @Test
+    fun `loadAndDisplayPromotionalOffer returns false when offer is not eligible`() = runTest {
```
tonidero, line 98: 🙌

tonidero (approved):
This makes sense to me! :shipit:

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
facumenzella: I went through as debug in iOS, but I can revisit if it's a problem in the future. I thought about debug because our [guidelines](https://docs.revenuecat.com/docs/debugging) suggest `.debug` as a log level for debugging 🤓

PR #2327, opened 2025-04-04
Migrate amazon & debugview modules to KTS

skydoves:
### Motivation

- Migrate `amazon` & `debugview` modules to KTS
- Bump kover to 0.7.2 and remove the `afterEvaluate` tasks. (refer to https://github.com/Kotlin/kotlinx-kover/issues/362)


library.gradle
```diff
@@ -25,12 +25,6 @@ android {
     }
 }
 
-project.afterEvaluate {
-    // Remove afterEvaluate
-    // after https://github.com/Kotlin/kotlinx-kover/issues/362 is fixed
```
tonidero, line 30: 🙌

tonidero (approved):
:shipit:

JayShortway (approved):
Let's go! 🙌

skydoves:
@tonidero @JayShortway Do you have any ideas for passing the test workflow? 😄

api-tester/build.gradle
```diff
@@ -14,6 +14,7 @@ android {
         }
         customEntitlementComputation {
             dimension "apis"
+            matchingFallbacks = ["defaults"]
```
JayShortway, line 17: @skydoves This fixes it locally for me. 🤞 The way I understand the error is that `:feature:amazon`, `:ui:revenuecatui` and `:ui:debugview` don't declare the `customEntitlementsComputation` flavor, and so `api-tester`'s `customEntitlementsComputation` flavor cannot find the appropriate flavor of those dependencies. This `matchingFallbacks` option tells it to fall back to the `defaults` flavor of the dependency if it doesn't have a ``customEntitlementsComputation` flavor. (cc @tonidero)

skydoves: Got it! Thanks so much for jumping in and fixing it directly, @JayShortway! 🙏

tonidero: Oh nice! So we do use this in these modules:
```
    defaultConfig {
        missingDimensionStrategy("apis", "defaults")
    }
```
But seems this is not picking that up... In any case, I think this is fine 👍

JayShortway: Ah right! Should `"apis"` be included in `matchingFallbacks` you think?

tonidero: Hmm well I think `apis` is the "dimension", vs `defaults/customEntitlementComputation` which are the possible values of that dimension. I think `matchingFallbacks` requires the value of the dimension, if I understand correctly? So I think it should be enough as it is... Might be worth confirming it would still fail API tests for the custom entitlement computation flavor though, just in case

JayShortway:
Ugh, those Amazon tests are now failing with 
> Expected stackmap frame at this location.

I don't know what causes it, but I've seen it reported before: https://github.com/RevenueCat/purchases-kmp/issues/271

I think it's related to the Amazon dependency and JDK version, but I don't know why it's suddenly an issue. Maybe updating the dependency from 3.0.5 to 3.0.7 helps? (Although there's nothing mentioned in the [changelog](https://developer.amazon.com/docs/appstore-sdk/release-notes.html).) 

Otherwise we can try setting `-noverify` in the `:feature:amazon` module, as suggested [here](https://github.com/powermock/powermock/issues/693#issuecomment-337695757).

JayShortway:
@skydoves @tonidero This fixes Amazon tests for me locally: https://github.com/RevenueCat/purchases-android/pull/2327/commits/296e40e71588083e6a4a56b9e1c9c5bb8e19cded.

api-tester/build.gradle
JayShortway, line 1: This file is empty. I think it can be removed?

ui/debugview/build.gradle.kts
JayShortway, line 1: `ui/debugview/build.gradle` (the Groovy one) should be deleted right?

library.gradle
JayShortway: [Kover is now failing.](https://app.circleci.com/pipelines/github/RevenueCat/purchases-android/16215/workflows/006f3071-f59e-46c0-84b6-192a29f3a5aa/jobs/86407) Maybe this was needed after all?

api-tester/build.gradle
skydoves: Nice catch! I just cleaned up and updated all the outdated changes 😄

skydoves:
@JayShortway @tonidero Eventually it has passed! 🚀 Thank you so much for your support and patience!

JayShortway:
@skydoves Awesome, thanks for persisting! 😄 Glad to get this in.

PR #2328, opened 2025-04-04
Migrate settings.gradle to KTS and add the dependency resolutions

skydoves:
### Motivation

Migrate settings.gradle to KTS and add the dependency resolutions.

> The includeGroupByRegex function in Gradle's settings.gradle.kts file allows you to filter artifacts from a repository based on regular expressions applied to their group IDs. This is useful for selectively including or excluding dependencies from a repository, improving build performance by reducing the amount of metadata Gradle needs to process. For example, you might use it to only include Android-related libraries from Google's Maven repository.
>
> By specifying a regular expression, you can target specific groups of artifacts. This is more flexible than simply including or excluding entire repositories, as it allows for fine-grained control over which dependencies are used in your project. The function is typically used within the content block of a repository definition in the pluginManagement section of the settings.gradle.kts file.

- includeGroupByRegex("com\\.google\\.firebase(\\..*|)"): Google Billing library basically depends on Firebase.
- includeGroupByRegex("com\\.google\\.prefab"), includeGroupByRegex("com\\.google\\.testing\\.platform"): Some Jetpack Android libraries depend on this.
- includeGroupByRegex("com\\.google\\.crypto\\..*"): Tink library.


settings.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+        google {
+            content {
+                includeGroupByRegex("androidx\\..*")
+                includeGroupByRegex("com\\.android(\\..*|)")
+                includeGroupByRegex("com\\.google\\.android\\..*")
+                includeGroupByRegex("com\\.google\\.crypto\\..*")
+                includeGroupByRegex("com\\.google\\.firebase(\\..*|)")
+                includeGroupByRegex("com\\.google\\.gms(\\..*|)")
+                includeGroupByRegex("com\\.google\\.prefab")
+                includeGroupByRegex("com\\.google\\.testing\\.platform")
```
tonidero, line 13: Hmm I do wonder if it's worth adding this... Don't think we earn much and in case google adds some other dependency here that we need to include we would need to update this... But then again, a win is a win, and adding dependencies is probably not going to happen often, so I don't think it's a blocker 👍

tonidero (approved):
Looking good! Just a question

settings.gradle.kts
skydoves: Thank you for reviewing this, Toni! This dependency filtering feature also helps prevent potential vulnerabilities caused by resolving incorrect artifacts since Gradle retrieves dependencies based on the order of declared repositories, so it may be more trustworthy for our customers. Now it's the default configuration as you can see it will be added automatically when you create a new Android project 😄

settings.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+        // fetch plugins from gradle plugin portal (https://plugins.gradle.org)
+        gradlePluginPortal()
+
+        // fallback for the rest of the dependencies
+        mavenCentral()
+    }
+}
+
+dependencyResolutionManagement {
+    repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
```
JayShortway, line 29: YES

settings.gradle.kts
JayShortway: Yea I think the supply-chain vulnerability is a good reason to include this. Gradle Syncs should also be slightly faster, as it won't try to e.g. get Coil from the `google` repo (referring to `dependencyResolutionManagement` below). 

I see Toni's point about potentially having to keep this up to date manually, but I'm assuming that any issues would be compile-time, right?

JayShortway (approved):
Much better! Thanks for taking care of this!

settings.gradle.kts
tonidero: Ohh I actually didn't know that, make sense, thank you for explaining! This makes sense then 👍 

And yeah, if there are any issues, I would assume them to be at compile-time so I don't think it would be a bit issue at all. Let's ship this! 💪

PR #2329, opened 2025-04-04
[Paywalls V2] Norwegian Bokmål and Norwegian Nynorsk fall back to Norwegian.

JayShortway:
## Bug
When the device is set to `nb` or `nn`, the paywall should pick the `no` localization. This was not happening.


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallStateLoadedComponentsLocaleTests.kt
```diff
@@ -269,6 +269,55 @@ internal class PaywallStateLoadedComponentsLocaleTests(
...
+                ),
+            ),
+            arrayOf(
+                "no-NO -> no",
+                Args(
+                    paywallLocales = nonEmptyListOf("en_US", "no"),
+                    deviceLocales = nonEmptyListOf("no-NO", "en-US"),
+                    expected = "no",
+                ),
+            ),
```
JayShortway, line 287: This part has an accompanying [soundtrack](https://youtu.be/qM5W7Xn7FiA?t=57).

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallStateLoadedComponentsLocaleTests.kt
```diff
@@ -269,6 +269,55 @@ internal class PaywallStateLoadedComponentsLocaleTests(
                     expected = "zh-Hant",
                 ),
             ),
+            arrayOf(
+                "no-NO -> no-NO",
+                Args(
+                    paywallLocales = nonEmptyListOf("en_US", "no_NO"),
```
tonidero, line 275: Probably not possible but just in case, can the paywall locale also have a `nb` or `nn` locale? If so, might be worth adding some test cases for that. We might need to implement a circular fallback system in that case 😅

tonidero (approved):
Just a question, but since I think it's probably not allowed, we should be good. Thanks for fixing!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallStateLoadedComponentsLocaleTests.kt
JayShortway: Yea, it's indeed not possible right now. I went to implement the fallback anyway just in case, but I don't think we should. `no` is a fallback for `nb` and `nn`, but I don't think the same is true the other way around. If a device is set to `no`, but the paywall has `nb` (hypothetically), I'm not sure we should be serving the `nb` localized paywall.

PR #2332, opened 2025-04-04
Use int for VC Balance

fire-at-will:
### Motivation
This PR is the Android equivalent for https://github.com/RevenueCat/purchases-ios/pull/4952. We recently decided to make the maximum balance for virtual currency balances be bounded by a signed 32-bit integer, so using a `Long` here in `purchases-android` is no longer necessary.

### Description
This PR changes the VirtualCurrencyInfo's `balance` field from a `Long` to an `Int`. Kotlin Ints are signed 32-bit integers ([source](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-int/)), so this is fine. On the Java side of things, the `int` primitive is also a signed 32-bit int ([source](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html)), and the Integer class is also a 32-bit signed integer by transitivity since it wraps the `int` primitive ([source](https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html)).

This is technically a breaking change but that's okay since the Virtual Currency features are still in a beta state in the SDKs.


tonidero (approved):
:shipit:

PR #2335, opened 2025-04-07
[Diagnostics] Removes Android 7 requirement

JayShortway:
## Description
We still had a few leftover checks for Android 7, which are obsolete since #1944. This PR removes them, so Diagnostics can be used from Android 5 and up.


tonidero (approved):
Nice work!! Thanks for doing this 🙏  Should we update the docs for `PurchasesConfiguration` as well, to indicate it's also supported there now?

JayShortway:
@tonidero Thanks for pointing me to that doc! Removed the requirement here: https://github.com/RevenueCat/purchases-android/pull/2335/commits/854350d822d508f26080d8a5c4b0e8d8ef593a37.

PR #2339, opened 2025-04-08
Introduce purchases-bom package

skydoves:
### Motivation

Introduce the `purchases-bom` package to ensure consistent dependency management across multiple modules.

I noticed we were using version [0.28.0](https://github.com/vanniktech/gradle-maven-publish-plugin/releases/tag/0.28.0) of the Vanniktech plugin, which is built against Kotlin 1.9.20. This caused compatibility issues with our current Kotlin Gradle Plugin version 1.8.22, particularly when configuring the kotlin("jvm") plugin. To resolve this, I downgraded the plugin to [0.27.0](https://github.com/vanniktech/gradle-maven-publish-plugin/releases/tag/0.27.0), which is compatible with Kotlin 1.8.22 that we're currently using.


gradle/libs.versions.toml
```diff
@@ -38,7 +38,7 @@ kover = "0.7.0"
 legacyCoreUi = "1.0.0"
 lifecycle = "2.5.0"
 material = "1.6.0"
-mavenPublish = "0.28.0"
+mavenPublish = "0.27.0"
```
JayShortway, line 41: Good call, thanks!

JayShortway (approved):
Nice to give people the option to use this indeed!

PR #2345, opened 2025-04-09
Adds `EnumDeserializerWithDefault` and `SealedDeserializerWithDefault`

JayShortway:
This is part 1/3:
1. 👉  #2345 
2. #2346 
3. #2347
---------------
## Description
Adds 2 abstract serializers that will be used in the follow-up PRs to deserialize enums and sealed types with a default case.


purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/SealedDeserializerWithDefault.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+
+/**
+ * Deserializer for sealed classes with a default value.
+ */
+internal abstract class SealedDeserializerWithDefault<T : Any>(
+    private val serialName: String,
+    private val serializerByType: Map<String, () -> KSerializer<out T>>,
+    private val defaultValue: () -> T,
```
ajpallares, line 20: Is an optimization to only lazily compute the `defaultValue` when needed?

ajpallares (commented):
I think it makes sense, but I'll leave approval to someone who knows Kotlin better

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/EnumDeserializerWithDefault.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    constructor(
+        defaultValue: T,
+        typeForValue: (T) -> String = { value -> value.name.lowercase() },
+    ) : this(
+        valuesByType = defaultValue::class.java.enumConstants.associateBy(typeForValue),
+        defaultValue = defaultValue,
+    )
+
+    private val enumName = defaultValue.javaClass.simpleName
```
tonidero, line 26: I guess if/when we fix minification in our SDKs, we should remember to keep the names of these enums... But we should be ok for now 👍

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/SealedDeserializerWithDefault.kt
tonidero: I guess that a sealed class default might have properties that also need to be initialized... I think this way we allow that?

tonidero (approved):
Nice! I haven't checked the usage yet, but these make sense to me!

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/SealedDeserializerWithDefault.kt
JayShortway: Yea, it's very likely premature, but also practically free. I figured this abstract class doesn't know what the default value of its implementation will be, and this avoids any computation as well as object instantiation to when it is needed. Otherwise all default values will be instantiated right away.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/EnumDeserializerWithDefault.kt
JayShortway: Actually, it's only used in the `SerialDescriptor`, so I think that should be fine. Still good to double check when we minify.

PR #2346, opened 2025-04-09
[Paywalls v2] Adds a default case to most enums and sealed types

JayShortway:
This is part 2/3:
1. #2345 
2. 👉  #2346 
3. #2347
---------------
## Description
This uses the `EnumDeserializerWithDefault` and `SealedDeserializerWithDefault` to add default cases to all enums and sealed types, except `Background` and `ButtonComponent`. Those 2 will be done in the next PR, as they have some cascading effects.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Alignment.kt
```diff
@@ -1,62 +1,56 @@
...
-
-    @SerialName("bottom_leading")
     BOTTOM_LEADING,
-
-    @SerialName("bottom_trailing")
     BOTTOM_TRAILING,
 }
+
+@OptIn(InternalRevenueCatAPI::class)
+internal object HorizontalAlignmentDeserializer : EnumDeserializerWithDefault<HorizontalAlignment>(
```
tonidero, line 44: Hmm do we need to make all these `@JvmSynthetic`?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/MaskShape.kt
```diff
@@ -1,31 +1,40 @@
...
 }
+
+@OptIn(InternalRevenueCatAPI::class)
+internal object MaskShapeDeserializer : SealedDeserializerWithDefault<MaskShape>(
+    serialName = "MaskShape",
+    serializerByType = mapOf(
+        "rectangle" to { MaskShape.Rectangle.serializer() },
+        "concave" to { MaskShape.Concave.serializer() },
+        "convex" to { MaskShape.Convex.serializer() },
+        "circle" to { MaskShape.Circle.serializer() },
```
tonidero, line 37: I'm wondering if there is a way for us to not forget to add a new value here 🤔. I guess the tests you added would catch these, so it's probably ok 👍

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CarouselComponent.kt
```diff
@@ -171,3 +170,8 @@ class PartialCarouselComponent(
     @SerialName("auto_advance")
     val autoAdvance: CarouselComponent.AutoAdvancePages? = null,
 ) : PartialComponent
+
+@OptIn(InternalRevenueCatAPI::class)
+internal object CarouselPageControlPositionDeserializer : EnumDeserializerWithDefault<PageControl.Position>(
```
tonidero, line 175: i'm wondering, can these serializers be private?

tonidero (approved):
Just some questions, but looks great!!

skydoves (approved):
Overall looks good to me! 👍

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/MaskShape.kt
JayShortway: Yea agreed. So far I don't think there's any other option other than tests, because sealed types don't have a `values()` function or `entries` property to list all cases. So it's impossible (?) to build a mapping with an exhaustive `when`.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/Alignment.kt
JayShortway: `@JvmSynthetic` is not applicable to classes unfortunately.

PR #2348, opened 2025-04-09
Adjusts precision of some Emerge Snapshots

JayShortway:
## Description
This PR updates the Emerge Snapshots SDK to 1.4.1, the Gradle plugin to 4.3.1 and uses the new `@EmergeSnapshotConfig` annotation to adjust the precision of some flaky snapshots.


ui/revenuecatui/consumer-rules.pro
```diff
@@ -1 +1,2 @@
 -dontwarn com.emergetools.snapshots.annotations.IgnoreEmergeSnapshot
+-dontwarn com.emergetools.snapshots.annotations.EmergeSnapshotConfig
```
JayShortway, line 2: @vegaro @tonidero Do you remember whether this will break the release train?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
```diff
@@ -193,6 +194,7 @@ private class PreviewParametersProvider : PreviewParameterProvider<PreviewParame
     )
 }
 
+@EmergeSnapshotConfig(precision = 0.95f)
```
ajpallares, line 197: Total nit but I'm still curious 😄: why 0.95 here while 0.99 in the others?

ajpallares (commented):
Looks good to me, but since I don't know the answer to whether the change could break the release train... I'll leave approval to someone who does know

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentView.kt
ajpallares: Ohh ok! 😄  Thanks for clarifying!

ui/revenuecatui/consumer-rules.pro
JayShortway: I remember some similar issues when we added the `IgnoreEmergeSnapshot` annotation 🤔 Maybe I'm mistaken.

tonidero: I believe those problems were solved by adding this rule if I remember correctly. So I think this should be good

tonidero (approved):
Looks good!

ui/revenuecatui/consumer-rules.pro
JayShortway: Alright, thanks both for confirming! If 3 people think it's fine, it's _probably_ fine 😄

PR #2350, opened 2025-04-10
[Paywalls v2] Adds the unsupported Background type to the error message

JayShortway:
## Description
As the title says.


ajpallares (approved):
Makes sense! Pretty useful 👍

PR #2351, opened 2025-04-10
[Paywalls v2] Uses a fixed date for template previews to avoid daily changes.

JayShortway:
## Motivation
Since we now have a few templates with date variables, we should use a fixed date to avoid seeing changed snapshots daily.


facumenzella (approved):

tonidero (approved):
🙌

PR #2358, opened 2025-04-15
Add Purchases Coroutine extensions for Result

skydoves:
### Motivation

Add Purchases Coroutine extensions with receivers. While I was building [Cat Paywall Compose](https://github.com/RevenueCat/cat-paywall-compose/), I felt this type of coroutine extension might be useful and a bit more Kotlin way to handle success/errors at some points.

```kotlin
Purchases.awaitOfferings(
  onSuccess = { .. },
  onError = { .. }
)
```


purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -30,6 +30,32 @@ suspend fun Purchases.awaitOfferings(): Offerings {
...
+ * Coroutine friendly version of [Purchases.getOfferings].
+ *
+ * @throws [PurchasesException] with a [PurchasesError] if there's an error retrieving the offerings.
+ * @return The [Offerings] available to this user.
+ */
+@JvmSynthetic
+suspend fun Purchases.awaitOfferings(
+    onSuccess: suspend (Offerings) -> Unit,
+    onError: suspend (PurchasesException) -> Unit,
+) {
```
JayShortway, line 50: So this suspends, but also returns the result in a callback? Doesn't that defeat the purpose of suspending? 

If the goal is to have suspending functions that don't throw, I definitely agree. 😄 In that case, we might do something similar as we do in purchases-kmp? We have these extensions: 
- [kotlin.Result](https://github.com/RevenueCat/purchases-kmp/blob/f2510ece66c4201f82af1455bef4c869472a8d78/result/src/commonMain/kotlin/com/revenuecat/purchases/kmp/result/Result.kt)
- [arrow.core.Either](https://github.com/RevenueCat/purchases-kmp/blob/f2510ece66c4201f82af1455bef4c869472a8d78/either/src/commonMain/kotlin/com/revenuecat/purchases/kmp/either/Either.kt)

skydoves: Exactly, that’s what I was aiming for. The goal was to create suspending functions that don’t throw, allowing us to handle results in a monadic way. I took a look at how it's done in `purchases-kmp`, and it looks like a good approach. Let me create a dedicated `result` module to follow a similar way from the `purchases-kmp`.

JayShortway: Nice! I like that pattern 😄 

Do we need a dedicated `purchases-result` module though? I had doubts on that in purchases-kmp as well haha. I shipped it as a separate module to avoid extra dependencies, but I realized later that kotlin.Result is not an extra dependency at all. 😅 @RevenueCat/coresdk what do you think? Extra module or same module?

tonidero: If we're not adding new dependencies, I would vote for same module. Less complexity :). Might be some confusion with the existing function I guess though, not sure if it would need to be on a separate package.

skydoves: Absolutely agree with you all. The `kotlin.Result` is part of the Kotlin standard library, so there’s no need to introduce an additional library to support it. Fixed on: [6ded6b5](https://github.com/RevenueCat/purchases-android/pull/2358/commits/6ded6b581e0766a4a1f6e49bacb85b8f79bb7ee1)

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -30,6 +30,27 @@ suspend fun Purchases.awaitOfferings(): Offerings {
...
+ *
+ * Offerings will be fetched and cached on instantiation so that, by the time they are needed,
+ * your prices are loaded for your purchase flow. Time is money.
+ *
+ * Coroutine friendly version of [Purchases.getOfferings].
+ *
+ * @return The [Result] of [Offerings] available to this user.
+ */
+@JvmSynthetic
+suspend fun Purchases.awaitOfferingsResult(): Result<Offerings> =
```
tonidero, line 46: Can be done separately, but should we do the same thing for the `defaults` flavor methods: https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -30,6 +30,27 @@ suspend fun Purchases.awaitOfferings(): Offerings {
...
+ * Fetch the configured offerings for this users. Offerings allows you to configure your in-app
+ * products via RevenueCat and greatly simplifies management. See
+ * [the guide](https://docs.revenuecat.com/offerings) for more info.
+ *
+ * Offerings will be fetched and cached on instantiation so that, by the time they are needed,
+ * your prices are loaded for your purchase flow. Time is money.
+ *
+ * Coroutine friendly version of [Purchases.getOfferings].
+ *
+ * @return The [Result] of [Offerings] available to this user.
```
JayShortway, line 43: Since `Result`'s error case is the broadly-typed `Exception`, maybe we can document what kind of exception (`PurchasesException`) devs can expect.

JayShortway (commented):
Love it! No real comments besides Toni's.

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
skydoves: That completely makes sense! I just added clearer comments for the result containing the `Exception` case.

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
skydoves: Oh, I missed that! Let me work on this separately after merging this. Thank you for the details! 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -64,6 +85,38 @@ suspend fun Purchases.awaitPurchase(purchaseParams: PurchaseParams): PurchaseRes
...
+ *   - Filters out offers with "rc-ignore-offer" and "rc-customer-center" tag
+ *   - Uses [SubscriptionOption] with the longest free trial or cheapest first phase
+ *   - Falls back to use base plan
+ *
+ * @params [purchaseParams] The parameters configuring the purchase. See [PurchaseParams.Builder] for options.
+ * @return The [Result] of [PurchaseResult] contains [StoreTransaction] for this purchase and the updated,
+ * [CustomerInfo] for this user.
+ */
+@JvmSynthetic
+suspend fun Purchases.awaitPurchaseResult(purchaseParams: PurchaseParams): Result<PurchaseResult> {
```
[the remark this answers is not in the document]
skydoves: This is interesting. I just added calls to the new APIs in the `api-tester` module. I was actually planning to introduce the binary compatibility validator afterward 😄  Let me check if there are still any issues related to build flavors after done this.

JayShortway (approved):
Great stuff! 🙌  
  
(Maybe we should change the PR title now to indicate it adds functions returning `Result`, for our changelog.)

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -120,3 +206,33 @@ suspend fun Purchases.awaitRestore(): CustomerInfo {
...
+ * `appUserID` of the existing user. Going forward, either `appUserID` will be able to reference
+ * the same user.
+ *
+ * You shouldn't use this method if you have your own account system. In that case
+ * "restoration" is provided by your app passing the same `appUserId` used to purchase originally.
+ *
+ * Coroutine friendly version of [Purchases.restorePurchases].
+ *
+ * @return The [Result] of [CustomerInfo] with the restored purchases if the execution succeeds,
+ * or a [Result] containing [PurchasesTransactionException] if it fails.
```
tonidero, line 224: Hmm should this be
```suggestion
 * or a [Result] containing [PurchasesException] if it fails.
```

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -94,6 +149,37 @@ suspend fun Purchases.awaitGetProducts(
...
+/**
+ * Gets the StoreProduct(s) for the given list of product ids of type [type], or for all types if no type is specified.
+ *
+ * Coroutine friendly version of [Purchases.getProducts].
+ *
+ * @param [productIds] List of productIds
+ * @param [type] A product type to filter by
+ *
+ * @return A [Result] of list of [StoreProduct] with the products that have been able to be fetched from the
+ * store successfully, or a [Result] containing  [PurchasesTransactionException] if it fails.
```
tonidero, line 161: Same here
```suggestion
 * store successfully, or a [Result] containing  [PurchasesException] if it fails.
```

purchases/src/main/kotlin/com/revenuecat/purchases/CoroutinesExtensionsCommon.kt
```diff
@@ -64,6 +86,39 @@ suspend fun Purchases.awaitPurchase(purchaseParams: PurchaseParams): PurchaseRes
...
+ * be purchased.
+ * [StoreProduct.defaultOption] is selected via the following logic:
+ *   - Filters out offers with "rc-ignore-offer" and "rc-customer-center" tag
+ *   - Uses [SubscriptionOption] with the longest free trial or cheapest first phase
+ *   - Falls back to use base plan
+ *
+ * @params [purchaseParams] The parameters configuring the purchase. See [PurchaseParams.Builder] for options.
+ * @return The [Result] of [PurchaseResult] contains [StoreTransaction] for this purchase and the updated,
+ * [CustomerInfo] for this user if the execution succeeds, or a [Result] containing
+ * [PurchasesTransactionException] if it fails.
```
tonidero, line 103: And here
```suggestion
 * [PurchasesException] if it fails.
```

tonidero (approved):
Just some docs updates but looking good!

PR #2361, opened 2025-04-15
Add async storefront country code apis

tonidero:
### Description
We already had a `Purchases.storefrontCountryCode` property. This basically was null until the storefront property was returned. 

This PR adds some new APIs to obtain the `storefrontCountryCode` asynchronously, which is the real nature of this data

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetStorefrontCallback.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetStorefrontCallback.kt
@@ -1,0 +1,25 @@
+//  Purchases
+//
+//  Copyright © 2025 RevenueCat, Inc. All rights reserved.
+//
+package com.revenuecat.purchases.interfaces
+
+import com.revenuecat.purchases.CustomerInfo
+import com.revenuecat.purchases.PurchasesError
+
+/**
+ * Interface to be implemented when calling [Purchases.getStorefrontCountryCode]
+ */
+interface GetStorefrontCallback {
+    /**
+     * Will be called after the call has completed.
+     * @param customerInfo [CustomerInfo] class sent back when the call has completed
+     */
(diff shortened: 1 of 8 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetStorefrontCallback.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.revenuecat.purchases.CustomerInfo
+import com.revenuecat.purchases.PurchasesError
+
+/**
+ * Interface to be implemented when calling [Purchases.getStorefrontCountryCode]
+ */
+interface GetStorefrontCallback {
+    /**
+     * Will be called after the call has completed.
+     * @param customerInfo [CustomerInfo] class sent back when the call has completed
```
ajpallares, line 16: Wrong documentation for this method
```suggestion
     * @param storefrontCountryCode [String] containing the Store country code in ISO-3166-1 alpha2
```

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
ajpallares, line 1: Should we add a test for the error case?

ajpallares, line 1: Should we add a test for the error case? (FFTI)

ajpallares (approved):
Thank you! Only a couple of minor things, but looks great!

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetStorefrontCallback.kt
tonidero: Oops, thank you!!

PR #2368, opened 2025-04-21
Use fallback API hosts when receiving server down response

ajpallares:
### Checklist

### Description
This PR implements a retry mechanism using fallback URL hosts for supported requests that fail with a server down error

### Notes
* Only some requests support a fallback API host. Currently only getCustomerInfo and getProductEntitlementMapping support it.
* `AppConfig` now has a `fallbackBaseURLs` containing the list of hosts, in order of preference, to use for supported requests.
* When a request's response is a server error (5XX), the same request is immediately retried with the next host of the array.
* The logic is implemented per-request: that is, if a request fails with a server-down error for host 1 and then falls back and succeeds for host 2, then all future requests will use again host 1 initially. No state is held between different requests.
* No fallback retries happen when using a proxy URL

iOS PR counterpart --> https://github.com/RevenueCat/purchases-ios/pull/4970


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -136,7 +138,36 @@ internal class HTTPClient(
...
+                body,
+                postFieldsToSign,
+                requestHeaders,
+                refreshETag = true,
+                fallbackBaseURLs,
+                fallbackURLIndex,
+            )
+        } else if (RCHTTPStatusCodes.isServerError(callResult.responseCode) &&
+            endpoint.supportsFallbackBaseURLs &&
+            fallbackURLIndex in fallbackBaseURLs.indices
```
JayShortway, line 153: This checks if `fallbackURLIndex` is valid, but we're using `fallbackURLIndex + 1` (on L169 below).

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
```diff
@@ -682,4 +684,195 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
     }
 
     // endregion
+
+    // region Fallback API host
```
JayShortway, line 688: Could we add a test where the fallback server returns 500 as well?

JayShortway (commented):
Nice job! Just 2 minor comments.

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
ajpallares: That's a good point, yes. To make sure that no new requests happen and that the 500 error is returned. Thanks for the suggestion!

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
JayShortway: Oh yes, you're getting the url at L156 already. Makes sense! Thanks for explaining!

JayShortway, line 153: Perf nit, but a comparison is probably faster than a "contains": 
```suggestion
            fallbackURLIndex <= fallbackBaseURLs.lastIndex
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -516,6 +521,7 @@ internal class Backend(
                     body,
                     postFieldsToSign = null,
                     backendHelper.authenticationHeaders,
+                    fallbackBaseURLs = appConfig.fallbackBaseURLs,
```
tonidero, line 524: Just thinking to myself but should we even pass these fallback URLs for non-supported endpoints? I guess we are handling that separately though so it might be ok to do it like this so we don't forget on new endpoints, and handle in the `Endpoint` as you're doing :)

tonidero (approved):
Great job!! No comments from me!

JayShortway (approved):
Awesome, thanks for adding the test!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
ajpallares: Yeah, I had this same dilemma about this... honestly, I don't know what the best option is. That said, I wanted to avoid duplicating the logic and, most importantly, as you say, I didn't want to forget adding this on new endpoints. 

My first approach was to have the `fallbackBaseURLs` parameter default to `appConfig.fallbackBaseURLs`, but that led to some issues (in unit tests, IIRC).

PR #2369, opened 2025-04-21
[Diagnostics] add host parameter to http_request_performed event

ajpallares:
### Checklist

### Description
Adds a new `host` parameter to the existing `http_request_performed` diagnostics event, which can be useful with the new logic for retry some requests using fallback hosts when receiving a server-down response


tonidero (approved):
:shipit:

PR #2374, opened 2025-04-23
[Paywalls v2] Adds some more tests asserting top window insets behavior.

JayShortway:
## Description
I wrote these tests to investigate a customer's bug report regarding the top window insets. Turned out that was not the cause, but these tests are useful regardless.


tonidero (approved):
:shipit:

PR #2378, opened 2025-04-25
Update pre-commit hooks rule for checking metalava

skydoves:
### Motivation

Add a GitHub pre-commit hook to optionally run Detekt and Metalava checks before each commit. While not mandatory, it can be a convenient way to catch issues early. To use it, simply copy the provided `pre-commit` file into your `.git/hooks` directory and that's all.

If you face the following error: `The '.git/hooks/pre-commit' hook was ignored because it's not set as executable`

You can resolve it by running the following command: `chmod ug+x .git/hooks/*`


hooks/pre-commit
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
JayShortway, line 1: We already have a pre-commit hook haha 😄 It's in `scripts/pre-commit.sh`, and you can install it using: 
```shell
bundle exec fastlane android setup_dev
```

hooks/pre-commit
```diff
@@ -0,0 +1,39 @@
...
+  exit $EXIT_CODE
+fi
+
+
+./gradlew metalavaCheckCompatibilityDefaultsRelease -q
+EXIT_CODE=$?
+if [ $EXIT_CODE -ne 0 ]; then
+  echo "❌ metalava for defaults flavor failed, running it for you..."
+
+  ./gradlew metalavaCheckCompatibilityDefaultsRelease -q
```
JayShortway, line 21: Anecdotally, how long does this task take roughly on your machine? If it's too long it might add friction, but if it's a few seconds it should be fine. It's a balancing act haha.

JayShortway (commented):
Great minds think alike 🧠 Depending on how long metalava takes, we can add it to the existing pre-commit hook!

hooks/pre-commit
skydoves: Oooh, we already have one! Let me merge this into the exist file 😄

hooks/pre-commit
skydoves: I think it really depends on. If I run it when right after pulling from the main branch, it takes ~45 seconds, but if I run it in normal cases, it takes under 5 seconds. It might be a worthy tradeoff!

scripts/pre-commit.sh
```diff
@@ -12,4 +13,28 @@ if [ $EXIT_CODE -ne 0 ]; then
...
\ No newline at end of file
+rm $OUTPUT
+
+echo "Running metalava (defaults) check..."
+./gradlew metalavaCheckCompatibilityDefaultsRelease -q
+EXIT_CODE=$?
+if [ $EXIT_CODE -ne 0 ]; then
+  echo "❌ metalava for defaults flavor failed, running it for you..."
+
+  ./gradlew metalavaCheckCompatibilityDefaultsRelease -q
```
JayShortway, line 24: This runs the "check" task again, right? Or am I reading it wrong? (Same on line 36 below.)

scripts/pre-commit.sh
```diff
@@ -12,4 +13,28 @@ if [ $EXIT_CODE -ne 0 ]; then
...
+
+  echo "API dump done, please check the results and then try your commit again!"
```
JayShortway, line 26: Maybe we should mention that the updated dump should be included in the commit? Something like:
```suggestion
  echo "API dump done. Please check the updated API dump and add it to your commit. Then try again!"
```

(Same on line 38 below.)

hooks/pre-commit
JayShortway: Alright, let's give it a go! 

(Gradle's remote build cache could help reduce this time for the entire team, but we have nothing set up for that yet.)

JayShortway (commented):
Thanks for making the changes! Just a question and a suggestion.

scripts/pre-commit.sh
skydoves: Nice catch! 😄  It has been fixed to generate binaries: [4cd0922](https://github.com/RevenueCat/purchases-android/pull/2378/commits/4cd0922c33f874f83964ba01482c2f6e2a9c49b3)

scripts/pre-commit.sh
skydoves: This sounds much clearer. Thank you for your suggestion!

JayShortway (approved):
Very nice! (Would be good to update the PR title so it's correct in our changelog, since it only adds metalava now.)

PR #2393, opened 2025-05-07
Use SideEffect for refreshing vm states inside InternalPaywall

skydoves:
### Motivation

Since we're ensuring the effect runs after a successful recomposition, it's ideal for triggering side-effects that rely on the final, stable state of the UI. This approach makes running such operations safer during the recomposition phase.


tonidero (approved):
Ohh right! I think this makes a lot of sense, thanks!!

JayShortway (approved):
Nice improvement! 

(I reran the danger job.)

PR #2318, opened 2025-04-02
Clean up libs.versions.toml

skydoves:
### Description

Clean up the `libs.versions.toml` file for consistency and improve readability.

- Versions have nothing changed.
- re-ordered versions & dependencies alphabetically.
- Hard-coded versions are moved to be referenced.

The result:

```toml
[versions]
agp = "8.8.0"
activityCompose = "1.9.3"
# Note that version 18.0.1 adds the AD_ID permission to the SDK.
# We should not add that permission ourselves, so keeping it at 17.0.1 for now.
adsIdentifier = "17.0.1"
amazon = "3.0.5"
androidxCore = "1.8.0"
androidxNavigation = "2.5.3"
annotation = "1.3.0"
appcompat = "1.4.1"
assertJ = "3.22.0"
browser = "1.8.0"
billing = "7.1.1"
composeBom = "2024.09.00"
coil = "2.4.0"
coreTesting = "2.0.0"
commonmark = "0.21.0"
constraintlayout = "1.1.0"
coroutines = "1.6.4"
datastorePreference = "1.0.0"
dependencyGraph = "0.9"
detekt = "1.23.6"
detektRulesCompose = "0.4.10"
dokka = "1.8.10"
emergeGradlePlugin = "4.2.0"
emergeSnapshots = "1.3.0"
espresso = "3.4.0"
fragment = "1.6.1"
hamcrest = "1.3"
recyclerview = "1.2.1"
roboelectric = "4.11.1"
kotlin = "1.8.22"
kotlinBom = "1.8.0"
# Can't update until we use more recent kotlin. 1.6.0 uses Kotlin 1.9.0
kotlinxSerializationJSON = "1.5.1"
kover = "0.7.0"
legacyCoreUi = "1.0.0"
lifecycle = "2.5.0"
material = "1.6.0"
mavenPublish = "0.28.0"
mockk = "1.13.11"
mockwebserver = "4.2.0"
multidex = "2.0.1"
paparrazzi = "1.3.1"
poko = "0.13.1"
testJUnit = "1.1.3"
testJUnitParams = "1.1.1"
testLibraries = "1.4.0"
tink = "1.8.0"
viewmodelCompose = "2.4.0"
window = "1.1.0"

[libraries]
activity-compose = { module = "androidx.activity:activity-compose", version.ref = "activityCompose" }
androidx-appcompat = { module = "androidx.appcompat:appcompat", version.ref = "appcompat" }
[...]

PR #2396, opened 2025-05-14
Fix downgrade of constraint layout version

vegaro:
It looks like this change https://github.com/RevenueCat/purchases-android/pull/2318/files#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfR59 downgraded the version of `androidx.constraintlayout` to `1.1.0`

I am suspicious it could be causing https://github.com/RevenueCat/react-native-purchases/issues/1266


skydoves (approved):
Oh no, I've missed this! This completely makes sense.

PR #2398, opened 2025-05-20
[Paywalls V2] Support stroke color and width in Carousel indicator

tonidero:
### Description
This adds support for stroke color and width for the Carousel indicators.

Equivalent of https://github.com/RevenueCat/purchases-ios/pull/5045

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CarouselComponent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CarouselComponent.kt
@@ -77,7 +77,17 @@ class CarouselComponent(
         @get:JvmSynthetic
         @SerialName("ms_transition_time")
         val msTransitionTime: Int,
-    )
+        @get:JvmSynthetic
+        @SerialName("transition_type")
+        val transitionType: TransitionType?,
+    ) {
+        @Serializable(with = CarouselTransitionTypeDeserializer::class)
+        enum class TransitionType {
+            // SerialNames are handled by the CarouselTransitionTypeDeserializer.
+            FADE,
+            SLIDE,
+        }
+    }
 
     @Poko
     @Serializable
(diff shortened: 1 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CarouselComponent.kt
```diff
@@ -77,7 +77,17 @@ class CarouselComponent(
         @get:JvmSynthetic
         @SerialName("ms_transition_time")
         val msTransitionTime: Int,
-    )
+        @get:JvmSynthetic
+        @SerialName("transition_type")
+        val transitionType: TransitionType?,
+    ) {
+        @Serializable(with = CarouselTransitionTypeDeserializer::class)
+        enum class TransitionType {
```
tonidero, line 85: I started on this before deciding that would be more complex and worth its own PR, but I left it here for now since it was easier and we would need to do this anyway in a followup PR.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentView.kt
```diff
@@ -243,12 +255,28 @@ private fun Indicator(
...
+
     Box(
         modifier = Modifier
             .padding(horizontal = pageControl.spacing / 2)
             .clip(Shape.Pill.toShape())
             .background(color)
-            .size(width = width, height = height),
+            .size(width = width, height = height)
+            .conditional(shouldApplyStroke) {
+                border(width = strokeWidth, color = strokeColor, shape = Shape.Pill.toShape())
```
tonidero, line 278: This border applies on top of the indicator itself, without changing its size. This differs from the border parameter in the stacks or other components, where the border can actually modify the size of the whole component. Not sure if that's the expected behavior... Thoughts?

joshdholtz (approved):
This looks good to me!

PR #2401, opened 2025-05-27
[Paywalls v2] The `ButtonComponentView`'s progress indicator now takes margin into account

JayShortway:
## Bug
The progress indicator shown on top of `ButtonComponentView` was not taking the margin into account, causing the indicator to show outside of the button's bounds. 

## Fix
The margin is now taken into account for both sizing and positioning.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentState.kt
@@ -20,6 +20,7 @@ import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.ui.revenuecatui.components.ComponentViewState
 import com.revenuecat.purchases.ui.revenuecatui.components.ScreenCondition
 import com.revenuecat.purchases.ui.revenuecatui.components.buildPresentedPartial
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -124,15 +148,17 @@ internal fun ButtonComponentView(
...
                 width = totalWidth,
                 height = totalHeight,
             ) {
                 stack.placeRelative(x = 0, y = 0)
                 // Center the progress indicator.
                 progress.placeRelative(
-                    x = ((totalWidth / 2f) - (progress.width / 2f)).roundToInt(),
-                    y = ((totalHeight / 2f) - (progress.height / 2f)).roundToInt(),
+                    x = marginStartPx.toInt() + ((stackWidthMinusMargin / 2f) - (progress.width / 2f)).roundToInt(),
+                    y = marginTopPx.toInt() + ((stackHeightMinusMargin / 2f) - (progress.height / 2f)).roundToInt(),
```
tonidero, line 161: Should we add some previews for the different scenarios here? Not sure if it's easy to take snapshot tests for the progress indicator though (considering it's animating 😅)...

tonidero (approved):
I think this makes sense! Great fix! Just a comment, but happy to merge this as is

PR #2402, opened 2025-05-27
Correctly translates variables when a fuzzy match is required.

JayShortway:
## Bug
In the offerings response, variables could be keyed by just the language (`de`) while the rest of the strings are keyed by language and region (`de_DE`). 

## Fix
The fix is to apply fuzzy locale matching when looking for the right variables to use. 

I also had to make some other changes to properly set up this scenario in a test.

Fixes https://github.com/RevenueCat/purchases-kmp/issues/386


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentState.kt
```diff
@@ -118,7 +119,9 @@ internal class TextComponentState(
 
     @get:JvmSynthetic
     val localizedVariableKeys by derivedStateOf {
-        style.variableLocalizations.run { getOrDefault(localeId, entry.value) }
+        // We use getBestMatch here, because the localeId in `texts` might be different from the one in
+        // `variableLocalizations`. For instance, `texts` might have `de_DE` while `variableLocalizations` has `de`.
+        style.variableLocalizations.run { getBestMatch(localeId) ?: entry.value }
```
JayShortway, line 124: This is the actual fix.

tonidero (approved):
🐛 🥾

PR #2403, opened 2025-05-27

RCGitBot:


tonidero:
Hmm seems like those screenshots have removed their safe area? Is that correct @JayShortway? I see it was changed on the templates, so maybe it's just these snapshots don't have safe areas. Correct?

JayShortway:
@tonidero Yea our screenshots do not know the concept of safe area. (Not possible with Compose Previews afaik.) Our templates used to have additional fixed padding to account for the safe area, but that was removed now that we have safe area support on the dashboard. That's what the screenshot changes are showing (removal of fixed padding). Do you think something is wrong?

PR #2404, opened 2025-05-28
Fix `Transaction` `isSandbox` parsing

tonidero:
### Description
The parsing of the `is_sandbox` field for non subscription transactions was wrong. This fixes that.


ajpallares (approved):
Good find! Thanks for fixing!

PR #2408, opened 2025-05-29
[Paywalls V2] Support missing packages in paywalls

tonidero:
### Description
This makes it so we render the paywall even if there is a missing product from the play store. It will only fail and fallback to the fallback paywall in case no packages are found.


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/FakeData.kt
```diff
@@ -0,0 +1,24 @@
...
+//        packageType = PackageType.CUSTOM,
+//        product = mockk(),
+//        presentedOfferingContext = mockk(),
+//    )
+//
+//    val packageComponent = PackageComponent(
+//        packageId = FAKE_PACKAGE_ID,
+//        isSelectedByDefault = false,
+//        stack = StackComponent(components = emptyList())
+//    )
```
joshdholtz, line 23: Do we want to delete this? 🤔

joshdholtz (approved):
Looks good to me! One comment about commented out code but otherwise :shipit:

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/FakeData.kt
tonidero: Oops thanks!

PR #2409, opened 2025-05-30
[Paywalls V2] DownloadableFonts 1: Move FontSpec calculation to Offering fetching in preparation of supporting downloadable fonts.

tonidero:
### Description
First PR in order to support downloadable fonts.

Since we want to download and cache fonts that are not found locally we need to:
- Find if the font is embedded in the app
- If not, download the fonts

In case it's needed, we want to download the fonts when fetching offerings to when we fetch offerings, so the fonts are ready when displaying the paywall if needed. In order to do this, we need to move the calculation of whether the font is embedded to this point as well.

This PR doesn't add any functionality, but does have a behavior change. We will now calculate whether fonts are embedded beforehand, which might add some load time to offerings. Actual download and caching of fonts will happen in follow-up PRs.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
```diff
@@ -0,0 +1,90 @@
...
+ * It also allows us to defer resolving the actual font to the UI layer, as only at that time do we know the exact
+ * override that's being used. We need to know this, because we need to know the [FontWeight] for which to resolve the
+ * font.
+ */
+@InternalRevenueCatAPI
+sealed interface FontSpec {
+    data class Resource(@get:JvmSynthetic val id: Int) : FontSpec
+    data class Asset(@get:JvmSynthetic val path: String) : FontSpec
+    data class Google(@get:JvmSynthetic val name: String) : FontSpec
+    sealed interface Generic : FontSpec {
```
vegaro, line 29: What's the difference between a Generic font and a System font?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/FontSpecProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.paywalls.components.properties.FontSpec
+import com.revenuecat.purchases.paywalls.components.properties.determineFontSpecs
+
+@OptIn(InternalRevenueCatAPI::class)
+internal class FontSpecProvider internal constructor(
+    private val context: Context,
+) {
+    private val fontSpecs: MutableMap<FontAlias, FontSpec> = mutableMapOf()
```
vegaro, line 14: this is not thread safe right?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -902,6 +905,11 @@ internal class PurchasesOrchestrator(
     //endregion
     //endregion
 
+    @InternalRevenueCatAPI
+    fun getCachedFontSpecs(): Map<FontAlias, FontSpec> {
```
ajpallares, line 909: I believe this one only needs to be `internal` and not `@InternalRevenueCatAPI`, as it's not being used by RevenueCatUI

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
```diff
@@ -0,0 +1,90 @@
...
+    // Create a map of FontAliases to FontSpecs.
+    return mapValues { (_, fontsConfig) -> configToSpec.getValue(fontsConfig) }
+}
+
+@OptIn(InternalRevenueCatAPI::class)
+private fun Context.determineFontSpec(info: FontInfo): FontSpec =
+    when (info) {
+        is FontInfo.GoogleFonts -> FontSpec.Google(name = info.value)
+        is FontInfo.Name -> when (info.value) {
+            "sans-serif" -> FontSpec.Generic.SansSerif
```
vegaro: Would it be worth it to extract these strings?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
```diff
@@ -0,0 +1,90 @@
...
+import android.annotation.SuppressLint
+import android.content.Context
+import com.revenuecat.purchases.FontAlias
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.UiConfig.AppConfig.FontsConfig
+import com.revenuecat.purchases.UiConfig.AppConfig.FontsConfig.FontInfo
+import com.revenuecat.purchases.common.debugLog
+
+/**
+ * A `FontSpec` is a more detailed version of [FontInfo]. A [FontInfo.Name] can be resolved to a generic font
```
vegaro, line 14: Reading this docs, I am not really understanding what a "generic" font is

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/FontSpecProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+    fun loadFonts(offering: Offering) {
+        val fonts = offering.paywallComponents?.uiConfig?.app?.fonts?.takeIf { it.isNotEmpty() } ?: return
+        val fontsToCalculate = fonts.filter { !fontSpecs.contains(it.key) }
+        if (fontsToCalculate.isNotEmpty()) {
+            fontSpecs.putAll(fontsToCalculate.determineFontSpecs(context))
+            // WIP: Trigger download of Fonts that need to be downloaded.
+        }
+    }
+
+    fun getFontSpecMap() = fontSpecs
```
vegaro: Would it make sense to return an inmutable copy of `fontSpecs`?

ajpallares (commented):
Took a look. Only one small comment for now 👌

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Hmm you're right! Only the one in `Purchases` needs to be @InternalRevenueCatAPI. Doesn't even need to be internal since the class is already internal. Good catch!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
tonidero: They are pretty similar IMO. Maybe they could be joined. I think the main point here was to have `Serif`, `SansSerif` and `Monospace` have a higher priority than any other "System" fonts. cc @JayShortway in case you have other thoughts.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
tonidero: Left a comment above. I believe it's mostly about the order of preference for those fonts... But otherwise, it's basically a system-installed font I think.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/FontSpecProvider.kt
tonidero: 🤦 Yup, should do that.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/FontSpecProvider.kt
tonidero: And yup, you're right... This should normally only be called in a single thread, but yeah, we should try to make it thread safe. Great catch!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Actually this is needed because `FontAlias` and `FontSpec` are annotated with this. So we need to either opt in or propagate the annotation. We could opt in here, but considering this is close to the external API, I thought to just propagate the OptIn. Happy to change though, I don't have a strong opinion

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
```diff
@@ -0,0 +1,93 @@
...
+
+@OptIn(InternalRevenueCatAPI::class)
+private fun Context.determineFontSpec(info: FontInfo): FontSpec =
+    when (info) {
+        is FontInfo.GoogleFonts -> FontSpec.Google(name = info.value)
+        is FontInfo.Name -> when (info.value) {
+            SANS_SERIF_FONT_NAME -> FontSpec.Generic.SansSerif
+            SERIF_FONT_NAME -> FontSpec.Generic.Serif
+            MONOSPACE_FONT_NAME -> FontSpec.Generic.Monospace
+            else -> getResourceIdentifier(name = info.value, type = "font")
```
vegaro, line 64: I think extracting some private functions from this code would help understand what's doing

tonidero: Right... we already have some private functions here... but I guess we could organize it better to avoid this pretty complex switch case. I'm ok to refactor this, but since this PR is just moving existing code, I might leave it as its is on this PR.

tonidero:
Closing in favor of #2414

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/FontSpec.kt
JayShortway: Yea, every device is guaranteed to have the generic fonts (serif, sans-serif and monospace), but some devices might have others. System font is basically a superset of generic fonts.

PR #2413, opened 2025-06-02
Use new resource format for cross validation screenshots and fix lint

joshdholtz:
⚠️ I did not run this yet... 

### Motivation

Use new paywall resources format to have multiple sets of paywalls to screenshot

### Description

Looking more dynamically for screenshots


ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -27,52 +27,46 @@ import java.io.InputStream
...
-                .current
-                ?.takeUnless { it.paywallComponents == null }
+        .flatMap { (_, file) ->
+            val uiConfig = JSONObject(file.inputStream().readUiConfig())
+            val packagesArray = JSONObject(getResourceStream("packages.json").readBytes().decodeToString())
+                .getJSONArray("packages")
+            val indices = file.inputStream().indexOfferings()
+
+            indices.mapNotNull { (start, end) ->
+                val offeringJsonString = file.inputStream().readOfferingAt(start, end)
```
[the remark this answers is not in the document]
JayShortway: We are consciously not doing this, as the stream does not stay open long enough when this is rendered in the Emerge Snapshot environment.

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -183,11 +177,15 @@ private fun PaywallTemplateImageLoader(
...
-                // Create the resourcePath by dropping the TLD, reversing the host, and appending the path.
-                val resourcePath = DIR_TEMPLATES + "/" +
-                    url.host.split('.').dropLast(1).reversed().joinToString("/") +
-                    url.path
-                val bitmap = BitmapFactory.decodeStream(getResourceStream(resourcePath))
+                val relativePath = url.path.removePrefix("/")
+
+                // Look for a matching file in all subdirectories
+                val matchingFile = getResourcesBaseDir().walkTopDown()
+                    .firstOrNull { it.isFile && it.name == relativePath.substringAfterLast("/") }
```
JayShortway: This seems very unoptimized. The previous approach was based on the fact that we could be certain of a file's location given its URL. Just walking the file tree every time is a bit wasteful. The Emerge Snapshots environment is also very sensitive to performance problems, so that's why it caught my eye.

JayShortway: Actually it seems that Emerge does trip over the changes in this PR. It only registers removed screenshots, but that is unexpected. There are also some errors. My hunch is that we need to optimize for memory, as I had to do in the previous iteration of this.

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -199,6 +197,21 @@ private fun PaywallTemplateImageLoader(
         .build()
 }
 
+private fun getResourcesBaseDir(): File {
+    return File(BuildConfig.PROJECT_DIR)
+        .resolve("../../upstream/paywall-preview-resources/resources")
+}
```
JayShortway: This only works in Android Studio. When the screenshot is rendered on the actual Android OS (device or emulator, as Emerge does), it won't be able to find the resources. We need to check the class loader instead. Check the implementation of `getResourceStream()` for an example.

JayShortway (commented):
Thanks for doing this! Just some perf issues that we should optimize.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -77,15 +77,25 @@ internal fun Map<FontAlias, FontSpec>.getFontSpec(
...
+internal fun Result<FontSpec, PaywallValidationError>.recoverFromFontAliasError(): Result<FontSpec?, PaywallValidationError> =
     flatMapError { error ->
         if (error is PaywallValidationError.MissingFontAlias && error.alias.value.isBlank()) {
+            // Treating this as a dashboard error and just ignoring it. No need to log.
+            Result.Success(null)
+        } else if (error is PaywallValidationError.MissingFontAlias) {
+            Logger.e(
+                "Font named '${error.alias}' was not found in the font config. Try re-adding it in the Paywall editor.",
+            )
             Result.Success(null)
```
JayShortway, line 99: @tonidero FYI, I don't _think_ this should conflict with your stuff right?

tonidero: Oh it's a different part of the code so I don't think so either. Thanks for checking!

ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt
```diff
@@ -156,35 +166,50 @@ internal class OfferingProvider : PreviewParameterProvider<Offering> {
...
-
-    ProvidePreviewImageLoader(PaywallTemplateImageLoader(LocalContext.current)) {
-        LoadedPaywallComponents(
-            state = state,
-            clickHandler = { },
-        )
+    val offering = paywall.offering
+    val parentFolder = paywall.parentFolder
+    // validatePaywallComponentsDataOrNullForPreviews should only return null if the Offering has no paywallComponents,
+    // but we filter those out in the PaywallResourcesProvider.
```
tonidero, line 174: I think this is fine for previews 👍

tonidero (approved):
This looks great!!

PR #2414, opened 2025-06-03
[Paywalls V2] Downloadable fonts

tonidero:
### Description
This adds support for downloadable fonts in Android. These will be predownloaded when fetching offerings.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -110,29 +114,87 @@ internal fun FontSpec.resolve(
...
+        null
     }
+}
+
+private fun ResourceProvider.getDownloadedFontSpec(fontsConfig: FontsConfig): FontSpec.Downloaded? {
+    val webUrl = fontsConfig.web?.value?.takeIf { it.isNotEmpty() }
+    val webHash = fontsConfig.web?.hash?.takeIf { it.isNotEmpty() }
+    if (webUrl != null && webHash == null) {
+        Logger.w(
+            "Font `$webUrl` does not have a validation hash. " +
```
ajpallares: Because the url is invisible in the dashboard, I would include the Android's font name in the log message, to make it easier for them to identify which font they should re-upload
```suggestion
            "Font `${fontsConfig.android.value}` does not have a validation hash. " +
```

tonidero: Good catch!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -110,29 +114,87 @@ internal fun FontSpec.resolve(
...
+    } else {
+        null
     }
+}
+
+private fun ResourceProvider.getDownloadedFontSpec(fontsConfig: FontsConfig): FontSpec.Downloaded? {
+    val webUrl = fontsConfig.web?.value?.takeIf { it.isNotEmpty() }
+    val webHash = fontsConfig.web?.hash?.takeIf { it.isNotEmpty() }
+    if (webUrl != null && webHash == null) {
+        Logger.w(
```
ajpallares: Do we want a warning or an error here?

tonidero: Hmm TBH, I feel this is more of a warning... since this could potentially be called for older fonts that haven't been uploaded yet, so they don't have a `web` value. I could potentially refactor this so we only log if `web` is not null though, which might be cleaner. In that case, I could consider making it an error yeah.

ajpallares: Ohh right! I didn't realize this was being logged also for when `web` is null. I agree with the approach of only logging if `web` is not null (that's the approach we took in iOS), because in that case, it is an error

tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/2414/commits/0f831ebde88f7cbc85af94d0329bd9832afbc5c5

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+    private val ongoingDownloads = mutableSetOf<String>()
+
+    init {
+        if (!cacheDir.exists()) {
+            if (!cacheDir.mkdirs()) {
+                errorLog("Unable to create cache directory for remote fonts: ${cacheDir.absolutePath}")
+            }
+        } else if (!cacheDir.isDirectory) {
+            errorLog("Remote fonts cache path exists but is not a directory: ${cacheDir.absolutePath}")
+        }
```
JayShortway: Maybe we can leave these checks until we're on the IO dispatcher, in `getCachedFontFileOrStartDownload()`? Checking if a file or directory exists is technically disk access, which we shouldn't be doing in the constructor (nor on the main thread).

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+import java.security.MessageDigest
+
+internal class RemoteFontLoader(
+    private val context: Context,
+    private val cacheDir: File = File(context.cacheDir, "rc_paywall_fonts"),
+    private val ioScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
+    private val urlConnectionFactory: UrlConnectionFactory = DefaultUrlConnectionFactory(),
+) {
+
+    private val ongoingDownloads = mutableSetOf<String>()
```
JayShortway: We never seem to be adding anything to this set?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+            return cachedFile
+        }
+
+        ioScope.launch {
+            synchronized(this) {
+                if (ongoingDownloads.contains(urlHash)) {
+                    verboseLog("Font download already in progress for $url")
+                    return@launch
+                }
+                ongoingDownloads.remove(urlHash)
```
JayShortway: Does this line have any effect? If `ongoingDownloads` contains `urlHash`, we have already returned from the function at this point.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+
+            try {
+                performDownloadAndCache(
+                    url = url,
+                    expectedMd5 = expectedMd5,
+                    urlHash = urlHash,
+                    extension = extension,
+                )
+            } catch (@Suppress("TooGenericExceptionCaught") t: Throwable) {
+                errorLog("Error downloading remote font from $url: ${t.message}")
```
JayShortway: Maybe we should also print the stack trace of the `Throwable`? Can help (us) debugging.

```kotlin
t.printStackTrace()
```

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+                errorLog("Unable to create cache directory for remote fonts: ${cacheDir.absolutePath}")
+            }
+        } else if (!cacheDir.isDirectory) {
+            errorLog("Remote fonts cache path exists but is not a directory: ${cacheDir.absolutePath}")
+        }
+    }
+
+    fun getCachedFontFileOrStartDownload(url: String, expectedMd5: String): File? {
+        val urlHash = md5Hex(url.toByteArray(Charsets.UTF_8))
+        val extension = url.substringAfterLast('.')
```
JayShortway: We should probably add the `missingDelimiterValue = ""` to `substringAfterLast`. Otherwise, if the `url` doesn't contain a period, `extension` will be set to the full `url`.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+import java.io.File
+import java.io.FileOutputStream
+import java.io.IOException
+import java.io.InputStream
+import java.net.HttpURLConnection
+import java.security.MessageDigest
+
+internal class RemoteFontLoader(
```
JayShortway: Maybe this class can be called something like `RemoteFontDownloader` or just `FontDownloader`? I think "Downloader" is less ambiguous than "Loader", as the latter can mean many things (e.g. loading into memory).

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+    private val context: Context,
+    private val cacheDir: File = File(context.cacheDir, "rc_paywall_fonts"),
```
JayShortway: I think this is fine. Just wondering if we considered placing it in the "regular" app directory instead? The cache directory could be emptied by the user in the app info screen. Not sure if that matters. Maybe it's even desirable.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+        return null
+    }
+
+    @Throws(IOException::class)
+    private fun performDownloadAndCache(
+        url: String,
+        expectedMd5: String,
+        urlHash: String,
+        extension: String,
+    ): File {
```
JayShortway: Nit, maybe this can return a `Result` instead of throwing? FFTI

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+        val cachedFile = File(cacheDir, "$urlHash.$extension")
+
+        val tempFile = File.createTempFile("rc_paywall_font_download_", ".$extension", cacheDir)
+        try {
+            downloadToFile(url, tempFile)
+
+            val actualMd5 = md5Hex(tempFile.readBytes())
+            if (!actualMd5.equals(expectedMd5, ignoreCase = true)) {
+                tempFile.delete()
+                errorLog("Font download MD5 mismatch for $url: expected=$expectedMd5, actual=$actualMd5")
```
JayShortway: Maybe this is slightly more informative:
```suggestion
                errorLog("Downloaded font file is corrupt for $url. expected=$expectedMd5, actual=$actualMd5")
```

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+                throw IOException("MD5 mismatch for $url: expected=$expectedMd5, actual=$actualMd5")
+            }
+
+            if (!tempFile.renameTo(cachedFile)) {
+                tempFile.copyTo(cachedFile, overwrite = true)
+                tempFile.delete()
+            }
+            debugLog("Font downloaded successfully and cached: $url")
```
JayShortway: I think "and cached" follows from the fact it was downloaded: 
```suggestion
            debugLog("Font downloaded successfully from $url")
```

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -0,0 +1,146 @@
...
+            while (true) {
+                bytesRead = input.read(buffer)
+                if (bytesRead < 0) break
+                out.write(buffer, 0, bytesRead)
+            }
+        }
+    }
+
+    private fun md5Hex(bytes: ByteArray): String {
+        val md = MessageDigest.getInstance("MD5")
```
JayShortway: Might be premature optimization, but maybe we can reuse the `md` instance? E.g. by making it a lazy member field?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/UrlConnectionFactory.kt
```diff
@@ -0,0 +1,40 @@
+package com.revenuecat.purchases.utils
+
+import java.io.InputStream
+import java.net.HttpURLConnection
+import java.net.URL
+
+internal interface UrlConnectionFactory {
```
JayShortway, line 7: Interesting! Is the plan to use this more widely, e.g. also in our `HTTPClient`?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -291,12 +293,22 @@ internal class PurchasesFactory(
...
             )
 
+            val fontLoader = RemoteFontLoader(
+                context = context,
+            )
+
+            val offeringFontPreDownloader = OfferingFontPreDownloader(
+                context = application,
+                remoteFontLoader = fontLoader,
+            )
```
JayShortway: Why is the `RemoteFontLoader ` getting `context` and `OfferingFontPreDownloader` getting `application`? I think both should be getting `contextForStorage` instead, as both do storage-related things.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
```diff
@@ -0,0 +1,86 @@
...
+    fun preDownloadOfferingFontsIfNeeded(offerings: Offerings) {
+        // Getting the first offering's paywall components to check for fonts.
+        // All offerings are expected to have the same fonts.
+        val fontInfosToDownload = offerings.all.values.firstOrNull()?.paywallComponents?.uiConfig?.app?.fonts?.values
+            ?.filterNot { isBundled(it.android) }
+            ?.filter {
+                if (it.web != null && it.web.hash == null) {
+                    errorLog(
+                        "Font ${it.android.value} does not have a validation hash. Skipping download. " +
+                            "Pleases try to re-upload the font in the RevenueCat dashboard.",
```
JayShortway: ~~error~~ help message, great stuff!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
```diff
@@ -0,0 +1,86 @@
...
+import java.net.URL
+
+@OptIn(InternalRevenueCatAPI::class)
+internal class OfferingFontPreDownloader(
+    private val context: Context,
+    private val remoteFontLoader: RemoteFontLoader,
+) {
+
+    private val assetsFontsDir = "fonts"
+    private val genericFonts = setOf(
```
JayShortway, line 23: Maybe we can add a comment here: 
```suggestion
    // GenericFontFamily names as defined by Compose. Restated here because we don't include any Compose dependencies.
    private val genericFonts = setOf(
```

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
```diff
@@ -47,22 +47,37 @@ class UiConfig(
...
+                // This should not be null, but we keep it nullable for backwards compatibility and handle
+                // nullability through code later, ignoring web fonts that do not have a hash.
+                @get:JvmSynthetic
+                val hash: String? = null,
+            )
+
             @InternalRevenueCatAPI
             @Serializable
             sealed interface FontInfo {
+                val value: String
```
JayShortway: This is kinda "dangerous" as the `value` means something completely different depending on the concrete type, so reading it without knowing the type doesn't really tell you much. Not a big deal, but ideally we avoid this imo.

purchases/src/test/java/com/revenuecat/purchases/UiConfigTests.kt
```diff
@@ -75,9 +75,13 @@ internal class UiConfigTests {
                       "value": "Roboto"
                     },
                     "web": {
-                      "type": "google_fonts",
-                      "value": "Gothic"
-                    }
+                      "type": "name",
```
JayShortway: `WebFontInfo` doesn't have a `type`, right? (Occurs in more places in this test.)

purchases/src/test/java/com/revenuecat/purchases/paywalls/RemoteFontLoaderTest.kt
```diff
@@ -0,0 +1,139 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.io.ByteArrayInputStream
+import java.io.File
+import java.net.HttpURLConnection
+import java.security.MessageDigest
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(AndroidJUnit4::class)
+class RemoteFontLoaderTest {
```
JayShortway: Nice job making use of Coroutine's test utilities!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -110,29 +114,88 @@ internal fun FontSpec.resolve(
...
+    }
+
+    return FontSpec.System(name = fontsConfig.android.value).also {
+        Logger.d(
+            "Could not find a font resource named `${fontsConfig.android.value}`. Assuming it's an OEM system font. " +
+                "If it isn't, make sure the font exists in the `res/font` folder. See for more info: " +
+                "https://developer.android.com/develop/ui/views/text-and-emoji/fonts-in-xml",
+        )
+    }
+}
```
JayShortway: Slightly more idiomatic could be: 

```kotlin
private fun ResourceProvider.determineFontSpec(fontsConfig: FontsConfig): FontSpec {
    return getBundledFontSpec(fontsConfig)
        ?: getDownloadedFontSpec(fontsConfig)
        ?: FontSpec.System(name = fontsConfig.android.value).also {
            Logger.d(
                "Could not find a font resource named `${fontsConfig.android.value}`. Assuming it's an OEM system " +
                    "font. If it isn't, make sure the font exists in the `res/font` folder. See for more info: " +
                    "https://developer.android.com/develop/ui/views/text-and-emoji/fonts-in-xml",
            )
        }
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -110,29 +114,88 @@ internal fun FontSpec.resolve(
...
+                    ?: getAssetFontPath(name = androidInfoValue)
+                        ?.let { path -> FontSpec.Asset(path = path) }
+            }
         }
+    } else {
+        null
     }
+}
+
+private fun ResourceProvider.getDownloadedFontSpec(fontsConfig: FontsConfig): FontSpec.Downloaded? {
```
JayShortway: Great stuff!

JayShortway (commented):
Some comments here and there, but I think the overall approach looks good! Great job!!

purchases/src/test/java/com/revenuecat/purchases/UiConfigTests.kt
tonidero: It's not parsed, but the backend also returns it, which is why I added it... but yeah it would be ignored right now.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/UrlConnectionFactory.kt
tonidero: TBH, I saw these around and I liked it since it helps with testability. I think it would be good to reuse everywhere in the future yeah, though we probably would need to improve it to support POST and other request types. For now, this allowed us to test most of the downloading logic.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: I think that was a complete miss on my side, yeah we should do that, thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
tonidero: Well we use this folder to predownload images... and I think it's fine even if the user clears the cache, it would just cause the fonts/images to be downloaded again. Shouldn't be an issue, unless the app doesn't restart... but I think the app should restart in that scenario?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
tonidero: Note this also loads the file if found in cache, so it's not only downloading. I think it could be fine to use `FontLoader` though... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
tonidero: Hmm I don't fully like that it would print even if the developer hasn't set debug logs... But I guess for errors it would normally be enabled anyway... Will add it for now.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
tonidero: I think I messed this up at some point 🥴 Thanks for catching this! Will fix and add tests

tonidero: Right, this should be adding to the `ongoingDownloads`, which fixes this and the previous comment.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
tonidero: hmm we actually don't need to return anything I guess, since we continue without waiting for the result here... I might just do that instead.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
JayShortway: Yea you're right. I'd suggest `FontProvider`, but that's already a thing. So let's go with `FontLoader` for now! 

(Maybe drop `Remote`, or change to `Web`? I like `Remote` more than `Web`, but we're using `Web` in other places to refer to these already.)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
JayShortway: We can also add an overload to `errorLog` that takes a `Throwable`, and will print the stack trace?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
JayShortway: Yea I think so too!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
```diff
@@ -74,13 +81,23 @@ internal class RemoteFontLoader(
         return null
     }
 
+    private fun initializeIfNeeded() {
```
JayShortway: How about `ensureFoldersExist()`?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/RemoteFontLoader.kt
tonidero: We already have one actually 🤦

tonidero:
Still need to add support for `TextComponent.font_weight_int` but I will do that in a separate PR.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/DownloadedFontFamily.kt
```diff
@@ -0,0 +1,32 @@
...
+    val file: File,
+)
+
+@InternalRevenueCatAPI
+data class DownloadedFontFamily(
+    val family: String,
+) {
+    private val _fonts: MutableList<DownloadedFont> = mutableListOf()
+    val fonts: List<DownloadedFont>
+        get() = _fonts.toList()
```
JayShortway: Instead of making `DownloadedFontFamily` mutable, which can be(come) hard to reason about, we can keep it immutable, and just create a new copy if we have more fonts to add to the family. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -0,0 +1,260 @@
...
+
+        ioScope.launch {
+            ensureFoldersExist()
+
+            val urlHash = md5Hex(url.toByteArray(Charsets.UTF_8))
+            val extension = url.substringAfterLast('.', missingDelimiterValue = "")
+            val cachedFile = File(cacheDir, "$urlHash.$extension")
+
+            synchronized(this) {
+                val fontInfosListeningToHash = fontInfosListeningToSameHashUrl[urlHash]
```
JayShortway: I find the word "listening" a bit hard to understand in both of these variable names. I don't think I fully understand it, but would either of these suggestions work?
- `fontInfosForHash`
- `fontInfosWithHash`

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -0,0 +1,260 @@
...
+            return null
+        }
+        if (fontInfo.weight == null) {
+            errorLog("Font weight is null for ${fontInfo.value}.")
+            return null
+        }
+        if (fontInfo.style == null) {
+            errorLog("Font style is empty for ${fontInfo.value}.")
+            return null
+        }
```
JayShortway: Should we add a troubleshooting step to these error messages, such as trying to reupload? Might save us some support tickets.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
```diff
@@ -0,0 +1,80 @@
...
+    fun preDownloadOfferingFontsIfNeeded(offerings: Offerings) {
+        // Getting the first offering's paywall components to check for fonts.
+        // All offerings are expected to have the same fonts.
+        val fontInfosToDownload = offerings.all.values.firstOrNull()?.paywallComponents?.uiConfig?.app?.fonts?.values
+            ?.map { it.android }
+            ?.filterIsInstance<FontInfo.Name>()
+            ?.filter {
+                it.url?.isNotEmpty() == true &&
+                    it.hash?.isNotEmpty() == true &&
+                    it.family?.isNotEmpty() == true &&
```
JayShortway: Should we check for `isNotBlank()` instead?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -110,29 +114,67 @@ internal fun FontSpec.resolve(
...
+            FontSpec.System(name = androidFontInfoName.value).also {
+                Logger.d(
+                    "Could not find a font resource named `${androidFontInfoName.value}`. " +
+                        "Assuming it's an OEM system font. If it isn't, make sure the font exists in the " +
+                        "`res/font` folder. See for more info: " +
+                        "https://developer.android.com/develop/ui/views/text-and-emoji/fonts-in-xml",
+                )
+            }
+        }
+}
```
JayShortway: I think we should avoid this returning null, so every `FontsConfig` gets turned into a `FontSpec` like before. Otherwise we might have some `FontAlias`es without a `FontSpec` in the resulting map. Suggestion: 

```kotlin
private fun ResourceProvider.determineFontSpec(fontsConfig: FontsConfig): FontSpec {
    return when (val fontInfo = fontsConfig.android) {
        is FontInfo.GoogleFonts -> FontSpec.Google(name = fontInfo.value)
        is FontInfo.Name -> getBundledFontSpec(fontInfo)
            ?: getDownloadedFontSpec(fontInfo)
            ?: FontSpec.System(name = fontInfo.value).also {
                Logger.d(
                    "Could not find a font resource named `${fontInfo.value}`. " +
                        "Assuming it's an OEM system font. If it isn't, make sure the font exists in the " +
                        "`res/font` folder. See for more info: " +
                        "https://developer.android.com/develop/ui/views/text-and-emoji/fonts-in-xml",
                )
            }
    }
}
```

Note that I'm passing `FontInfo.Name` to `getBundledFontSpec` and `getDownloadedFontSpec`.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/DownloadedFontFamily.kt
tonidero: Hmm I will need to refactor the caches a bit, but I agree that would be better. Will do that, thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
tonidero: Much cleaner, thanks for the suggestion!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/DownloadedFontFamily.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2414/commits/b78bc5804031b1bba7a17af03d8f9af3b85a5c4f

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/DownloadedFontFamily.kt
```diff
@@ -0,0 +1,32 @@
...
+    val weight: Int,
+    val style: FontStyle,
+    val file: File,
+)
+
+@InternalRevenueCatAPI
+data class DownloadedFontFamily(
+    val family: String,
+) {
+    private val _fonts: MutableList<DownloadedFont> = mutableListOf()
```
vegaro: this is not thread safe, I see you are synchronizing in `addFileToCache` so I believe the current code is ok, but it could bite us in the future

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -0,0 +1,283 @@
...
+            )
+            return null
+        }
+        if (fontInfo.style == null) {
+            errorLog(
+                "Font style is empty for ${fontInfo.value}." +
+                    "Please try to re-upload your font in the RevenueCat dashboard.",
+            )
+            return null
+        }
```
vegaro: This feels to me like it could live directly as an extension of `FontInfo`? Or maybe a `validate()` function

JayShortway (approved):
Awesome! Great job on this!! 💪

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
```diff
@@ -0,0 +1,80 @@
...
+        // All offerings are expected to have the same fonts.
+        val fontInfosToDownload = offerings.all.values.firstOrNull()?.paywallComponents?.uiConfig?.app?.fonts?.values
+            ?.map { it.android }
+            ?.filterIsInstance<FontInfo.Name>()
+            ?.filter {
+                it.url?.isNotBlank() == true &&
+                    it.hash?.isNotBlank() == true &&
+                    it.family?.isNotBlank() == true &&
+                    it.weight != null &&
+                    it.style != null &&
```
vegaro: Same here, this could be using the `validate()` function I suggested

vegaro (commented):
just some small comments, overall it looks good

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/DownloadedFontFamily.kt
tonidero: Ended up removing the mutability here, so I think we are ok now. But yeah you were right :)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
tonidero: Changed to a new common function here: https://github.com/RevenueCat/purchases-android/pull/2414/commits/5e1ca3e443fa18379d606a66c8f90049a2aefd01

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/fonts/DownloadableFontInfo.kt
```diff
@@ -0,0 +1,74 @@
...
+    val style: FontStyle,
+)
+
+@Suppress("ReturnCount")
+@JvmSynthetic
+@OptIn(InternalRevenueCatAPI::class)
+internal fun FontInfo.Name.toDownloadableFontInfo(): Result<DownloadableFontInfo> {
+    if (url.isNullOrBlank()) {
+        return Result.failure(
+            Exception(
```
tonidero: This was a bit generic, but considering this is only used for logging (not part of the public API), and that it was more difficult to use our `PurchasesException` since that requires a `PurchasesError`, I thought it was ok like this... Lmk if you think otherwise.

JayShortway: We could also use our own `com.revenuecat.purchases.utils.Result` here, which can use any type as an error allowing us to create specific errors. Not a blocker at all for me though, just an idea.

tonidero: Good point! I guess the only issue is that we are already using Kotlin's Result in FontLoader and we would be importing both... but I think that's ok. I do prefer not to create exceptions without need. Thanks!

PR #2415, opened 2025-06-03
Fix display of wrong product information when not using entitlements

vegaro:
Fixes Customer Center’s base plan detection when the user has an active subscription but no entitlements.

We got a report of a customer getting the wrong product information, it was displaying annual's product information instead of monthly's (the actually subscribed product plan)


tonidero (approved):
Makes sense! :shipit:

PR #2419, opened 2025-06-05
[Paywalls] Add `font_weight_int` to `TextComponent`

tonidero:
### Description
This adds support for the new way to specify font weight in TextComponent.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TextComponent.kt
```diff
@@ -52,6 +52,9 @@ constructor(
     @SerialName("font_weight")
     val fontWeight: FontWeight = REGULAR,
     @get:JvmSynthetic
+    @SerialName("font_weight_int")
+    val fontWeightInt: Int? = null,
```
JayShortway, line 56: Should we add a test to assert that `StyleFactory` picks `fontWeightInt` over `fontWeight`? And maybe the same for `TextComponentState` as it has the same logic for the `partial`.

JayShortway (approved):
Looks good! Just a suggestion for a test or 2.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TextComponent.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/2419/commits/858e4d0b5517a2a24a30512d50e764a82c837e41

PR #2433, opened 2025-06-11
Remove assertion on strings in Customer Center test

vegaro:
This assertion was failing every time a new string was added to the `CustomerCenterConfigData`, meaning an update on the fixture and the expected object was required.

I get this test is useful if we are getting the JSON from the backend but being it a fixture, it's just testing that the fixture is correct.


purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetCustomerCenterConfigTest.kt
```diff
@@ -182,54 +182,6 @@ class BackendGetCustomerCenterConfigTest {
             locale = "en",
             localizedStrings = mapOf(
                 "amazon_subscription_manage" to "You can manage your subscription in the Amazon Appstore app on an Amazon device.",
```
JayShortway, line 184: Curious why not all strings have been removed?

vegaro: Oh I just let a couple so the structure is somewhat similar to the reality, keeping all was a bit overkill that's why I removed most of them. I can just keep one?

JayShortway: Oh no it doesn't matter to me! Was just pure curiosity haha. Makes sense to make it resemble reality.

PR #2436, opened 2025-06-11
[Paywalls] Buttons can now open sheets

JayShortway:
## Description
Adds full support for buttons opening sheets.
- Adds a `SimpleBottomSheetScaffold` to implement it all. We cannot use Compose's built-in `BottomSheetScaffold` or `ModalBottomSheet`, as those are experimental APIs. The `SimpleBottomSheetScaffold` is completely decoupled from paywalls, so it can be used for other things if needed.
- Sheets are dismissed by `NavigateBack` buttons, the system back button, and by tapping outside it.
- The paywall content is optionally blurred (if configured on the dashboard). This is only supported on Android 12 and up, as Compose only supports this on 12 and up. Below 12, we show a translucent scrim. There are some less-performant ways we can blur content on older Android versions, which we can look into (see [Haze](https://github.com/chrisbanes/haze)), but for now this is fine imo. #abs

https://github.com/user-attachments/assets/2a1559cf-f0f7-4e93-9e21-12a1b346cb6e


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/SimpleBottomSheetScaffold.kt
```diff
@@ -0,0 +1,178 @@
...
+            enter = fadeIn(animationSpec = blurAnimationSpec()),
+            exit = fadeOut(animationSpec = blurAnimationSpec()),
+        ) {
+            Box(
+                modifier = Modifier
+                    .matchParentSize()
+                    // We don't want the entire screen to show a click indication.
+                    .clickable(
+                        interactionSource = remember { MutableInteractionSource() },
+                        indication = null,
```
vegaro, line 123: cool!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/SimpleBottomSheetScaffold.kt
```diff
@@ -0,0 +1,178 @@
...
+                    text = "Hello from my bottom sheet",
+                )
+                Text(
+                    text = "It's beautiful, isn't it?",
+                )
+            }
+        }
+    }
+
+    SimpleBottomSheetScaffold(
```
vegaro, line 160: Is the bottom sheet intended to wrap the content? because I see it's wrapping it's content in the preview (instead of doing full width)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -477,32 +477,59 @@ internal class StyleFactory(
...
+                destination.method,
+            )
+            is ButtonComponent.Destination.Terms -> buttonComponentStyleUrlDestination(
+                destination.urlLid,
+                destination.method,
+            )
+            is ButtonComponent.Destination.Url -> buttonComponentStyleUrlDestination(
+                destination.urlLid,
+                destination.method,
+            )
```
vegaro, line 514: maybe these threecan be grouped togehter?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -477,32 +477,59 @@ internal class StyleFactory(
...
+            is ButtonComponent.Destination.Sheet ->
+                zipOrAccumulate(
```
vegaro, line 516: nice

vegaro (approved):
This was super cool

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/SimpleBottomSheetScaffold.kt
vegaro: I see there's a `Size` that's part of the bottom sheet

JayShortway: This is a great callout! I agree that the sheet should be full width if `size` is null. Done here: https://github.com/RevenueCat/purchases-android/pull/2436/commits/b27caab7183cd531f786bd1e5538ebc28a12940a. Also rearranged and added some previews, so this behavior is tested.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JayShortway: Unfortunately not, because `destination` is a different type in each of these cases. It's smart-casting here in the `when`. (After this call they all become the same `Url` destination type.)

PR #2437, opened 2025-06-12
Introduce Baseline Profiles for increasing SDK startup performance

skydoves:
> **Note**: You may notice changes in line sizes due to optimizations introduced in this update.

[Baseline Profiles](https://developer.android.com/topic/performance/baselineprofiles/overview) help improve code execution speed—typically by around 30%—right from the first launch. They achieve this by skipping the interpretation and [just-in-time (JIT)](https://developer.android.com/about/versions/nougat/android-7.0#jit_aot) compilation phases for frequently used code paths.

By bundling a Baseline Profile with your app or library, the [Android Runtime (ART)](https://source.android.com/docs/core/runtime) can apply Ahead-of-Time (AOT) compilation to optimize these paths during installation. This results in a faster, smoother experience for every user, even on their first launch or after an update.

In this PR, I’ve added a new `baselineProfile` module targeting the `PaywallTester` sample app. This module pre-compiles relevant parts of the RevenueCat SDK. Using the filtering feature, only the relevant packages are picked up by the `purchases` and `ui` modules.

### Benchmark Results on Pixel 8a (API 34)

**With Baseline Profiles Applied**

```
timeToInitialDisplayMs   min 162.3,   median 188.7,   max 207.2
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```

**Without Baseline Profiles**

```
timeToInitialDisplayMs   min 174.8,   median 197.2,   max 230.0
Traces: Iteration 0 1 2 3 4 5 6 7 8 9
```


baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/BaselineProfileGenerator.kt
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * After you run the generator, you can verify the improvements running the [StartupBenchmarks] benchmark.
+ *
+ * When using this class to generate a baseline profile, only API 33+ or rooted API 28+ are supported.
+ *
+ * The minimum required version of androidx.benchmark to generate a baseline profile is 1.2.0.
+ **/
+@RunWith(AndroidJUnit4::class)
+@LargeTest
+class BaselineProfileGenerator {
```
JayShortway, line 37: We should probably rerun this on CI right? Maybe for every PR? That way the baseline profile gets updated when needed.

gradle/wrapper/gradle-wrapper.jar
JayShortway, line 1: Why was this modified? (Same with the `gradlew` scripts.)

purchases/src/main/baseline-prof.txt
JayShortway, line 1: In terms of additional SDK size, it's going to be this `baseline-prof.txt` file + the one in `revenuecatui` right? That's about 320kb in total. And the performance improvement is about 23 ms at most, looking at the benchmark results, right? That's more than a frame, especially on high refresh rate displays, so that's definitely nice. I think we should discuss with the broader team though whether we think that tradeoff (320 kb for 23 ms) is worth it. (cc @RevenueCat/coresdk)

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/paywalls/PaywallsScreen.kt
```diff
@@ -136,6 +147,7 @@ fun PaywallsScreen(
                 )
                 if (offering.paywallComponents == null) {
                     ButtonWithEmoji(
+                        modifier = Modifier.testTag("full_screen_footer"),
```
JayShortway:
```suggestion
                        modifier = Modifier.testTag("footer_button"),
```

baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/PaywallTesterScenario.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.test.uiautomator.BySelector
+import androidx.test.uiautomator.UiDevice
+import androidx.test.uiautomator.UiObject2
+import androidx.test.uiautomator.Until
+
+fun MacrobenchmarkScope.explorePaywallScreen() = device.apply {
+    wait(Until.hasObject(By.res("paywall_screen")), TIMEOUT)
+    waitForIdle()
+
+    waitAndFindObject(By.res("full_screen_button"), TIMEOUT).click()
```
JayShortway, line 14: Is this opening the first paywall in the list? That's not a "v2" paywall I think. Maybe (future PR) we can make a specific paywall with some common components (image, text, button, purchase button, package), and use that one to generate baseline profiles. That way the components code gets optimized too.

baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/BaselineProfileGenerator.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @get:Rule
+    val baselineProfileRule = BaselineProfileRule()
+
+    @Test
+    fun generate() {
+        baselineProfileRule.collect(
+            packageName = Constants.PACKAGE_NAME,
```
JayShortway, line 45: Can we not get this from the `Context`?

.idea/codeStyles/Project.xml
```diff
@@ -2,6 +2,41 @@
   <code_scheme name="Project" version="173">
     <option name="RIGHT_MARGIN" value="120" />
     <option name="SOFT_MARGINS" value="120" />
+    <JavaCodeStyleSettings>
```
JayShortway: Is this needed? Seems to be an auto-generated change haha.

examples/paywall-tester/src/main/generated/baselineProfiles/baseline-prof.txt
JayShortway, line 1: Is there a way to avoid this file from being generated? We don't really need baseline profiles for the paywall tester app, and it might get annoying if we keep getting diffs in this file.

examples/paywall-tester/src/main/generated/baselineProfiles/startup-prof.txt
JayShortway, line 1: Same for this one, is there a way to avoid this file from being generated?

JayShortway (commented):
Really cool to explore optimizations like this! 🚀 Let's see what the rest of the team thinks about the tradeoffs.

baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/BaselineProfileGenerator.kt
skydoves: We should include this in CI, and we can use the managed device feature in this PR. I’ll handle it in a follow-up PR.

gradle/wrapper/gradle-wrapper.jar
skydoves: Opps, it should be rollback, I was updating AGP for testing, and it was contained 😅

purchases/src/main/baseline-prof.txt
skydoves: That sounds fair. But if you open the APK/AAB file using Android Studio's Analyze APK menu, and access the `assets/dexopt/ directory`. Inside, you'll encounter the `baseline.prof` file. Upon checking its properties, you'll notice that the file size typically falls within the range of 10 to 30 KB (the entire huge 40k lines of baseline profiles) - which means, when you package your Baseline Profiles into an APK/AAB file, a substantial portion of the profile information is optimized, considerably reducing the compiled size. So, we don't need concerned about the file size increases in general.

So I assume the baseline profiles used in our `purchases` and `ui` modules should be quite small—probably less than ~15 KB?

Screenshot 2025-06-12 at 9 58 11 PM

baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/PaywallTesterScenario.kt
skydoves: Sounds like a plan! I’ll follow up on that scenario once the v2 paywall is ready in the sample project 😄

baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/BaselineProfileGenerator.kt
skydoves: Not sure which package name we should be targeting from the Baseline Profile side, might need to clarify that?

https://github.com/android/nowinandroid/blob/ba1a463498b6ac586708c334395af16ac035eb2b/benchmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/baselineprofile/ForYouBaselineProfile.kt#L36

examples/paywall-tester/src/main/generated/baselineProfiles/baseline-prof.txt
skydoves: We can't avoid generating these baseline profiles since we ultimately need to compile all user scenarios from the demo app and filter them into our library modules. How about just adding them to `.gitignore` instead?

baselineprofile/src/main/kotlin/com/revenuecat/baselineprofile/BaselineProfileGenerator.kt
JayShortway: I wonder if this would work? 

```kotlin
InstrumentationRegistry.getInstrumentation().targetContext.packageName
```

I can imagine it doesn't work. Depends a bit on how the `BaselineProfileGenerator` is connected to the Paywall Tester app.

purchases/src/main/baseline-prof.txt
JayShortway: Ah thanks for sharing that context! That's much better. We actually have Emerge Size Analysis for questions exactly like this, but for some reason I can't find the `baseline.prof` file in that [analysis](https://www.emergetools.com/build/791713c9-94b4-4bd2-b50d-d05c99c9038e). I've asked them about this. Would be great to get a conclusive number from Emerge.

examples/paywall-tester/src/main/generated/baselineProfiles/baseline-prof.txt
JayShortway: Yes, I think that works too! Would be good to add a comment to the `.gitignore` explaining why we don't care about these files.

tonidero: Right, I think adding to `.gitignore` makes the most sense to me!

purchases/src/main/baseline-prof.txt
tonidero: Oh nice, thanks for analyzing this! Considering ~15kb vs ~20ms, I would say it's worth it yeah.

tonidero, line 1: I wonder, I think we don't really care about changes to this one so maybe we can also gitignore it? I guess we need to build it on CI first before making a new release, and ideally also make sure to update it locally, to make sure we test everything with the latest version of this.

tonidero (commented):
This is cool!! Thanks for doing this! I'm just wondering if we can gitignore most of these files and just generate them when needed, so we avoid having to keep these files up to date. Other than that, I think the tradeoffs exposed make sense!

purchases/src/main/baseline-prof.txt
JayShortway: Got word back from Emerge, seems like we should upload a release build instead of a debug one to be able to see `baseline.prof`. 

We're currently uploading a single test app to Emerge: `TestPurchasesUiAndroidCompatibility`. That's uploaded as a debug build and I _think_ that is so Compose previews are included. Maybe we should upload a release build too? Would be nice to get actual SDK size numbers.

JayShortway: For these files it might be nice to see which classes and functions are actually being optimized, just to keep an eye on it? (E.g. we add a new call in `configure()`, we should see it being optimized in this file.)

To me this is a similar file as the API dump files.

skydoves: I've added the Baseline Profiles to `.gitignore`. Since our goal is to generate and filter Baseline Profiles specifically for our SDKs, not for how the test app invokes methods, so this approach helps us avoid unnecessary diffs every time profiles are generated. I'm not entirely sure if it's a bug, but it seems that Java signatures occasionally change on compile, even when there are no actual code changes, which causes unexpected updates to the Baseline Profiles from the test app.

tonidero: > To me this is a similar file as the API dump files.

Right, that's what makes me think it might not be worth reviewing the changes in this file. As in, we already have the API dump files. This would be just for optimizations... And not sure if it's worth it the extra changes we need to maintain + review.

> I'm not entirely sure if it's a bug, but it seems that Java signatures occasionally change on compile

Would this happen as well on these files from the libraries? just to make sure it won't cause changes inadvertently.

Having said all this, I'm ok if we want to see how this works and revisit if it gets noisy.

JayShortway: Yea that would be my suggested approach, to keep an eye on how this works and revisit later. Since we're not too familiar with it, I'm afraid this will otherwise be treated as an opaque box, not knowing if it's actually working properly or not.

skydoves: Agreed! The method signatures in Baseline Profiles can often change unpredictably, especially depending on how the user journey is designed in the test application scenario. That’s why I believe tracking changes in Baseline Profiles isn’t particularly meaningful. Personally, I usually regenerate them right before releasing my open-source libraries. And in most cases, that’s been sufficient since the profiles don't directly affect clients. Even if we contain a bit of lines that have been changed or removed method signatures, I do believe it impacts close to zero. (If the changes are not like hundreds of lines)

tonidero: Should we also add this and the baseline profile in the `ui` module to `.gitignore` then? If we did keep them, I think we should add some sort of test to make sure it's updated before each release, bug I think it's probably not worth it.

skydoves: We can simply maintain Baseline Profiles for each library module. In general, people don’t track changes to Baseline Profiles, especially in regular application projects since they don’t represent public API surfaces and are not intended to be versioned or monitored in the same way.

JayShortway (approved):
Yesss, another great improvement! 😄 

(I approved the Emerge changes. We have one flaky snapshot that we should make less strict.)

PR #2438, opened 2025-06-12
`getOfferings()` returns fresh Offerings and Products when the locale has changed

JayShortway:
## Motivation
The `OfferingsCache` would return cached Offerings (including products with localized descriptions) even if the locale has changes since caching. This prevents developers from showing the correct description in this scenario. 

## Changes introduced
`OfferingsCache` now remembers the locales at time of caching, and verifies they are unchanged when checking for staleness. Developers will receive up-to-date Offerings (including products with correctly-localized descriptions) when they call `getOfferings()` after the locale has changed.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsCache.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsCache.kt
@@ -3,6 +3,8 @@ package com.revenuecat.purchases.common.offerings
 import com.revenuecat.purchases.Offerings
 import com.revenuecat.purchases.common.DateProvider
 import com.revenuecat.purchases.common.DefaultDateProvider
+import com.revenuecat.purchases.common.DefaultLocaleProvider
+import com.revenuecat.purchases.common.LocaleProvider
 import com.revenuecat.purchases.common.caching.DeviceCache
 import com.revenuecat.purchases.common.caching.InMemoryCachedObject
 import com.revenuecat.purchases.common.caching.isCacheStale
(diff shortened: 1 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsCache.kt
```diff
@@ -14,19 +16,24 @@ internal class OfferingsCache(
     private val offeringsCachedObject: InMemoryCachedObject<Offerings> = InMemoryCachedObject(
         dateProvider = dateProvider,
     ),
+    private val localeProvider: LocaleProvider = DefaultLocaleProvider(),
 ) {
 
+    private var cachedLanguageTags: String? = null
```
JayShortway, line 22: Wasn't sure whether to put this in `OfferingsCache` or `InMemoryCachedObject`. Chose the former because the latter is more generic (although only used for Offerings). Let me know what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsCache.kt
```diff
@@ -36,13 +43,16 @@ internal class OfferingsCache(
...
+        offeringsCachedObject.lastUpdatedAt.isCacheStale(appInBackground, dateProvider) ||
+            // Locale-based staleness
+            cachedLanguageTags != localeProvider.currentLocalesLanguageTags
 
     @Synchronized
-    fun clearOfferingsCacheTimestamp() {
+    fun forceCacheStale() {
         offeringsCachedObject.clearCacheTimestamp()
+        cachedLanguageTags = null
     }
```
JayShortway, line 56: Renamed this one because it does more now.

PR #2430, opened 2025-06-09
Redo Customer Center subscription card and action buttons

vegaro:
| Active | Cancelled | Free Trial |
|--------|--------|--------|
| Screenshot 2025-06-09 at 21 05 00 | Screenshot 2025-06-09 at 21 05 00 | Screenshot 2025-06-09 at 21 05 00 |

PR #2439, opened 2025-06-12
Clean up paths logic in Customer Center

vegaro:
Split from work on https://github.com/RevenueCat/purchases-android/pull/2430 to make it easier to review

Cleans up `ManageSubscriptionsView` and unifies the `ManageSubscriptionsButtonsView` for both Play Store and non-play store purchases.

It adds the restore purchases functionality to non play store purchases. Before it was just showing managementURL and contact support. This also fixes promotionals, which were possibly showing without any action if no support email was configured (they don't have a management URL).

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterConfigTestData.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterConfigTestData.kt
@@ -76,6 +76,12 @@ internal object CustomerCenterConfigTestData {
                                 ),
                             ),
                         ),
+                        CustomerCenterConfigData.HelpPath(
+                            id = "5",
+                            title = "FAQ",
+                            type = CustomerCenterConfigData.HelpPath.PathType.CUSTOM_URL,
+                            url = "https://www.revenuecat.com/docs/customer-center-faq",
+                        ),
                     ),
                 ),
                 CustomerCenterConfigData.Screen.ScreenType.NO_ACTIVE to CustomerCenterConfigData.Screen(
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -166,17 +167,42 @@ internal class CustomerCenterViewModelImpl(
...
         }
     }
 
+    private fun handleCancelPath(context: Context) {
+        val currentState = _state.value as? CustomerCenterState.Success ?: return
+        val purchaseInfo = currentState.purchaseInformation
+
+        when {
+            purchaseInfo?.product != null -> startGoogleProductCancellation(context, purchaseInfo.product.id)
+            purchaseInfo?.managementURL != null -> startManagementUrlCancellation(context, purchaseInfo.managementURL)
```
facumenzella: Don't you need to check the store? (or until we display the list we're only displaying google products?)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -281,13 +296,34 @@ internal class CustomerCenterViewModelImpl(
...
+            HelpPath.PathType.CUSTOM_URL,
+            -> true
+            HelpPath.PathType.CANCEL ->
+                purchaseInformation?.store == Store.PLAY_STORE || purchaseInformation?.managementURL != null
+            HelpPath.PathType.REFUND_REQUEST,
+            HelpPath.PathType.CHANGE_PLANS,
+            HelpPath.PathType.UNKNOWN,
+            -> false
+        }
     }
```
facumenzella, line 327: Don't we need to add the store check for play store?

vegaro: talked offline but not needed, it's just not available for any platform

PR #2440, opened 2025-06-13
Bump AGP to 8.10.1 and Gradle to 8.14.2

skydoves:
### Description

Bump AGP to 8.10.1 and Gradle to 8.14.2.


gradle/libs.versions.toml
JayShortway, line 1: (This comment has nothing to do with this file 😄)

There seem to be some Kotlin version issues on CI, but I'm not sure what's happening at first glance. Especially confusing since _some_ CI jobs work fine. Do the fastlane/gradle commands (in the failing CI jobs) work locally for you?

JayShortway (commented):
Nice! Thanks for doing this!

gradle/libs.versions.toml
vegaro: Hmm yeah that's interesting not sure why either. That looks like the same issue when trying to add a dependency that's compiled with another version of kotlin

tonidero: Hmm interesting... I actually tried upgrading AGP to 8.9.1 [here](https://github.com/RevenueCat/purchases-android/pull/2303#issuecomment-2761058739) but had issues because of the databinding library... but this seems like a separate error. We should figure this out in case we need to do anything for the next major once BC8 comes out.

tonidero (approved):
Nice! Seems to be working now 🙌. Not sure if we want to take the chance to upgrade AGP to the latest... But otherwise, I think we can merge this!

JayShortway:
@tonidero Let's merge this! We can merge the Renovate one right after: https://github.com/RevenueCat/purchases-android/pull/2472.

PR #2444, opened 2025-06-13

RCGitBot:


JayShortway:
The 122 modified snapshots are not actually changed. They just changed order. A supporting data point is that the Paparazzi ones haven't changed (these are identified by Offering ID instead of order). I added sorting so it shouldn't happen again.

PR #2447, opened 2025-06-16
[Paywalls] Margin no longer decreases icon size

JayShortway:
## Bug
A customer reported an invisible icon.

## Cause
Turned out we incorrectly applied the margin in `IconComponentView`. The margin would subtract from the size. In this case, the icon was 32dp wide, and horizontal margin was 16dp, resulting in an invisible (32-16*2) icon. 

## Fix
This PR applies the margin correctly, by wrapping in a `Box` which is sized according to the icon size + margin.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
@@ -66,7 +66,7 @@ internal fun IconComponentView(
 
     Box(
         modifier = modifier
-            .size(iconState.marginAdjustedSize)
+            .size(iconState.sizePlusMargin)
             .padding(iconState.margin),
     ) {
         RemoteImage(
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentState.kt
@@ -30,7 +30,7 @@ import com.revenuecat.purchases.paywalls.components.properties.ThemeImageUrls
 import com.revenuecat.purchases.ui.revenuecatui.components.ComponentViewState
 import com.revenuecat.purchases.ui.revenuecatui.components.ScreenCondition
 import com.revenuecat.purchases.ui.revenuecatui.components.buildPresentedPartial
-import com.revenuecat.purchases.ui.revenuecatui.components.ktx.adjustForMargin
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.addMargin
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toContentScale
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toLocaleId
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toPaddingValues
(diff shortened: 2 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Size.kt
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+private fun SizeConstraint.addMargin(
+    margin: UInt,
+): SizeConstraint = when (this) {
+    is Fixed -> Fixed(value + margin)
+    is Fill,
+    is Fit,
+    -> this
+}
```
JayShortway, line 29: Moved out of `ImageComponentState` unchanged.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PaddingValues.kt
```diff
@@ -0,0 +1,17 @@
...
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.LayoutDirection
+
+@JvmSynthetic
+internal fun PaddingValues.calculateHorizontalPadding(layoutDirection: LayoutDirection): Dp =
+    calculateStartPadding(layoutDirection) + calculateEndPadding(layoutDirection)
+
+@JvmSynthetic
+internal fun PaddingValues.calculateVerticalPadding(): Dp =
+    calculateTopPadding() + calculateBottomPadding()
```
JayShortway, line 17: Moved out of `ImageComponentState` unchanged.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentState.kt
```diff
@@ -196,13 +193,13 @@ internal class ImageComponentState(
     val margin by derivedStateOf { presentedPartial?.partial?.margin?.toPaddingValues() ?: style.margin }
 
     @get:JvmSynthetic
-    val marginAdjustedSize: Size by derivedStateOf {
-        size.adjustForMargin(margin, layoutDirection)
+    val sizePlusMargin: Size by derivedStateOf {
+        size.addMargin(margin, layoutDirection)
```
JayShortway, line 197: Renamed these, as I noticed I was getting confused myself of what "margin adjusted" actually means.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
```diff
@@ -62,18 +64,23 @@ internal fun IconComponentView(
...
-            .padding(iconState.margin)
-            .applyIfNotNull(shadowStyle) { shadow(it, composeShape) }
-            .applyIfNotNull(backgroundColor) { background(it, composeShape) }
-            .clip(composeShape)
-            .applyIfNotNull(borderStyle) { border(it, composeShape) }
-            .padding(iconState.padding),
-        colorFilter = colorFilter,
-    )
+            .size(iconState.sizePlusMargin)
+            .padding(iconState.margin),
```
tonidero, line 70: Should shadow and border also be applied in the container? not sure, but seems that's what we do in the image component view...

tonidero (approved):
LGTM!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentView.kt
JayShortway: Doesn't look to be needed. Icon previews have borders and they look correct. In the `ImageComponentView` we do that because of the `onDrawWithContent` implementation of the `overlay` modifier, but `IconComponentView` doesn't support overlays.

PR #2456, opened 2025-06-19
[EXTERNAL] Improve logger performance with message lambdas

### LINKED ISSUE #2391: Some devices are encountering out-of-memory exceptions
- [ ] I have updated Purchases SDK to the latest version
- [ ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [ ] I have searched the [Community](https://community.revenuecat.com)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ ] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
We noticed from Firebase Crashlytics that after we updated Revenucat from 8.4.0 to 8.14.2, some of our users are encountering out-of-memory exceptions. We tried to update the SDK multiple times until we reached 8.16.0, but still the same issue.

1. Environment
   1. Platform: Android
   2. SDK version: 8.16.0
   3. OS version: Android 9 to 16
   4. Android Studio version: Android Studio Meerkat | 2024.3.1 Patch 2
   5. How widespread is the issue:  0.0916% (485 users out of 529462 per the last 30 days)
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
```
          Fatal Exception: java.lang.OutOfMemoryError: Failed to allocate a 14321672 byte allocation with 5153472 free bytes and 5032KB until OOM, target footprint 536870912, growth limit 536870912
       at java.util.Arrays.copyOf(Arrays.java:3585)
       at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:177)
       at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:543)
       at java.lang.StringBuilder.append(StringBuilder.java:186)
       at java.lang.StringBuilder.append(StringBuilder.java:179)
       at java.util.AbstractMap.toString(AbstractMap.java:557)
       at java.lang.String.valueOf(String.java:4102)
       at java.lang.StringBuilder.append(StringBuilder.java:179)
       at com.revenuecat.purchases.Offerings.toString(Offerings.kt:23)
       at java.util.Formatter$FormatSpecifier.printString(Formatter.java:3216)
       at java.util.Formatter$FormatSpecifier.print(Formatter.java:3093)
       at java.util.Formatter.format(Formatter.java:2704)
       at java.util.Formatter.format(Formatter.java:2640)
       at java.lang.String.format(String.java:4027)
       at com.revenuecat.purchases.common.offerings.OfferingsFactory$createOfferings$1.invoke(OfferingsFactory.kt:66)
       at com.revenuecat.purchases.common.offerings.OfferingsFactory$createOfferings$1.invoke(OfferingsFactory.kt:40)
       at com.revenuecat.purchases.common.offerings.OfferingsFactory$getStoreProductsById$1.invoke$lambda$1(OfferingsFactory.kt:153)
       at com.revenuecat.purchases.common.Dispatcher.enqueue$lambda$2$lambda$1(Dispatcher.kt:67)
       at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:487)
       at java.util.concurrent.FutureTask.run(FutureTask.java:264)
       at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:307)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
       at java.lang.Thread.run(Thread.java:1012)
```
4. Steps to reproduce, with a description of expected vs. actual behavior: No clear way to reproduce the issue.
5. More info:
- In terms of affected devices: 25% of them are from Honor, 18% from Samsung, 11% from Oneplus, 11% from Motorola, and 4% from other devices.

landarskiy:
### Checklist
- [X] unit tests

### Motivation
This change helps eliminate hidden performance issues in debug logging, such as:
* Large or unnecessary `toString()` serialization
* `String.format()` overhead or even ANRs on low-end devices

Resolves #2391 

### Description
Logger method signatures now accept a `messageBuilder` lambda instead of a direct `String`. This avoids unnecessary string construction when logging is disabled.

Key files:
* `com.revenuecat.purchases.common.logUtils`
* `com.revenuecat.purchases.common.logWrapper`

All other changes reflect usage updates across the codebase.

You can find detailed explanation [here](https://medium.com/proandroiddev/hidden-cost-of-logging-in-android-1fdd7cd3149d)


purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
```diff
@@ -63,18 +62,17 @@ internal class OfferingsFactory(
                                     ),
                                 )
                             } else {
-                                verboseLog(OfferingStrings.CREATED_OFFERINGS.format(offerings.all.size))
+                                verboseLog { OfferingStrings.CREATED_OFFERINGS.format(offerings) }
```
JayShortway: I think we should still avoid logging the entire `offerings` object, as this would still throw an OOM if logging is enabled.

landarskiy: Good point! As far as I understand from the code, in release mode the default log level is INFO. In this case, the message is created for a lower level - VERBOSE.

This means that unless a developer manually lowers the log level to VERBOSE for their release builds (which I really hope they don't), the message won't be calculated at all.

So in this specific case, I don’t think the SDK need to sacrifice log details. The issue is very specific and only happens in production and most likely on very low-end devices, where this expression won’t be evaluated by default unless a developer changes the log level.

If the log level were INFO or higher, the risk would still exist — and then it would be important to avoid logging serialized objects.

The original problem wasn’t that info logs were enabled, but that the expression was being calculated no matter if logging was turned on or not.

What do you think?

JayShortway: That's true. However, verbose logging is functionality offered by the SDK, which should not cause crashes in any circumstance. We can log more details of the `offerings` if needed, like the current offering ID, but we shouldn't be printing the entire object.

landarskiy: Makes sense, I agree. I’ve reverted to the original version with the count. It’s better, as you suggested, to add more details only when needed.

landarskiy:
I also marked the logger functions as `inline` (since they now have a lambda arguments), which gave an additional performance boost - up to 3X in all cases (with and without actual logging).

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
JayShortway: Thank you!

JayShortway (approved):
Thanks a lot for doing this! I'll get this merged.

PR #2458, opened 2025-06-20
Redo navigation in Customer Center

vegaro:


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,34 @@
+package com.revenuecat.purchases.ui.revenuecatui.customercenter.navigation
+
+import androidx.compose.runtime.Immutable
+
+@Immutable
+internal data class CustomerCenterNavigationState(
+    val destinationStack: List<CustomerCenterDestination> = listOf(CustomerCenterDestination.Main),
```
JayShortway: Just a suggestion to use the name `backStack`, as that is more common imo.

JayShortway: Instead of a `List`, it should be more performant to use an actual `Stack`:
```kotlin
val backStack = Stack<CustomerCenterDestination>().apply { push(CustomerCenterDestination.Main) }
```

Or use a `Deque`, as recommended by `Stack`'s own docs:
```kotlin
val backStack: Deque<CustomerCenterDestination> = 
    ArrayDeque<CustomerCenterDestination>().apply { push(CustomerCenterDestination.Main) }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -250,46 +283,110 @@ private fun CustomerCenterError(state: CustomerCenterState.Error) {
 private fun CustomerCenterLoaded(
     state: CustomerCenterState.Success,
     onAction: (CustomerCenterAction) -> Unit,
+    navigationViewModel: CustomerCenterNavigationViewModel,
```
JayShortway: It is generally not recommended to use a ViewModel down into the hierarchy. (See [here](https://developer.android.com/topic/libraries/architecture/viewmodel#best-practices) for instance.) Maybe we don't even need `CustomerCenterNavigationViewModel`? It seems to be a thin wrapper over `CustomerCenterNavigationState`, maybe that's all we need? And then the `CustomerCenterViewModel` can handle the remaining navigation logic.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -324,13 +421,81 @@ private fun MainScreen(
...
+}
+
+private fun isBackwardTransition(
+    from: CustomerCenterDestination,
+    to: CustomerCenterDestination,
+    navigationStack: List<CustomerCenterDestination>,
+): Boolean {
+    // Find positions of from and to in the current stack
+    val fromIndex = navigationStack.indexOf(from)
+    val toIndex = navigationStack.indexOf(to)
```
JayShortway: `indexOf` is not ideal, but probably fine as the stack will always be very small (< 10) right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -250,46 +283,110 @@ private fun CustomerCenterError(state: CustomerCenterState.Error) {
...
+                    is NavigationAction.ReplaceCurrentDestination -> navigationViewModel.replaceCurrentDestination(
+                        action.destination,
+                    )
                 }
             },
         )
-    } else {
-        val configuration = state.customerCenterConfigData
-        MainScreen(state, configuration, onAction)
     }
```
JayShortway: Could you explain why this (`updateNavigationFromState()`) is necessary? In a setup like this I would expect `navigationState.currentDestination` to directly drive an `AnimatedContent`, which is what `CustomerCenterNavigationHost` is already doing.

vegaro: hmm, this is actually the part that I am less convinced about, because it feels like I shouldn't be doing that navigation logic in the UI.

I think it should live in the view model and have the UI react to that. I will give that a thought. Thank you!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: yeah at this point is 2 max

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,68 @@
...
+            return true
+        }
+
+        // For other cases, use the stack positions
+        val stackList = backStack.toList()
+        // 0 will be the top of the backStack
+        val fromIndex = stackList.indexOf(from)
+        val toIndex = stackList.indexOf(to)
+
+        // If either destination is not in the stack, assume forward transition
```
tonidero: Hmm should we log something if the `from` destination is not in the stack? Seems like it would be an unexpected state

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterDestination.kt
```diff
@@ -0,0 +1,31 @@
...
+import androidx.compose.runtime.Immutable
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.FeedbackSurveyData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.PromotionalOfferData
+
+internal enum class CustomerCenterAnimationType {
+    SLIDE_HORIZONTAL,
+}
+
+@Immutable
+internal sealed class CustomerCenterDestination {
```
tonidero, line 12: Would it make sense to add the title property to the destination itself instead of the `CustomerCenterState`? Could make sense, so we remember to add it on the destination itself... However, it might be less flexible if we want to make it more dynamic, so not a strong opinion here.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -251,45 +265,87 @@ private fun CustomerCenterLoaded(
...
-    } else if (state.restorePurchasesState != null) {
-        RestorePurchasesDialog(
-            state = state.restorePurchasesState,
-            localization = state.customerCenterConfigData.localization,
-            onDismiss = { onAction(CustomerCenterAction.DismissRestoreDialog) },
-            onRestore = { onAction(CustomerCenterAction.PerformRestore) },
-            onContactSupport = state.customerCenterConfigData.support.email?.let { email ->
-                {
-                    onAction(CustomerCenterAction.ContactSupport(email))
+            label = "CustomerCenterScreens",
```
tonidero: Just curious, is this needed for something in particular?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,68 @@
+package com.revenuecat.purchases.ui.revenuecatui.customercenter.navigation
+
+import androidx.compose.runtime.Immutable
+import java.util.ArrayDeque
+import java.util.Deque
+
+@Immutable
+internal data class CustomerCenterNavigationState(
```
tonidero, line 9: Not for this PR! I mentioned it over Slack but I wonder if we might want to make this more generic so it can be reused by paywalls or any other UI in the future... For example, I think it might be cool to share any logic for displaying a sheet here, which we recently added in Paywalls... Having said all that, I'm ok making it focus on Customer center for now, since that's the only thing it's going to use it. We can always revisit this point in the future.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,68 @@
...
+        push(CustomerCenterDestination.Main)
+    },
+) {
+    val currentDestination: CustomerCenterDestination
+        get() = backStack.peek() ?: CustomerCenterDestination.Main
+
+    val canNavigateBack: Boolean
+        get() = backStack.size > 1
+
+    fun push(destination: CustomerCenterDestination): CustomerCenterNavigationState {
```
tonidero, line 22: Do we want to add some tests for all this logic? Might also be good to add UI tests for an actual navigation I guess if possible...

tonidero (commented):
This looks good to me! I think I would only make sure to add some tests for the logic at least... For UI tests, it would be great to add if it's not going to take too long.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            this
+        }
+    }
+
+    fun popToMain(): CustomerCenterNavigationState {
+        val newStack = ArrayDeque<CustomerCenterDestination>()
+        newStack.push(CustomerCenterDestination.Main)
+        return copy(backStack = newStack)
+    }
```
facumenzella: This is not really popping, is it? If it's a "pop", I would expect the same instance of "Main" to be reused, rather than creating a new one. Is that possible?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    fun replace(destination: CustomerCenterDestination): CustomerCenterNavigationState {
+        val newStack = ArrayDeque(backStack)
+        if (newStack.isNotEmpty()) {
+            newStack.pop()
+        }
```
facumenzella: this assumes only one level of navigation right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
JayShortway: That label shows up in the Animation Preview. See below on the timeline: 
image

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -251,45 +265,87 @@ private fun CustomerCenterLoaded(
...
-                onAction(CustomerCenterAction.DismissPromotionalOffer(promotionalOfferData.originalPath))
+    CustomerCenterScreenHost(
+        currentDestination = state.currentDestination,
+        customerCenterState = state,
+        onAction = onAction,
+    )
+}
+
+@Composable
+private fun CustomerCenterScreenHost(
```
JayShortway: I think `CustomerCenterNavHost` would be more descriptive.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -251,45 +265,87 @@ private fun CustomerCenterLoaded(
...
+    currentDestination: CustomerCenterDestination,
+    customerCenterState: CustomerCenterState.Success,
+    onAction: (CustomerCenterAction) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Box(modifier = modifier) {
+        AnimatedContent(
```
JayShortway: Why is this wrapped in a `Box`? Should be possible without it I think.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -251,45 +265,87 @@ private fun CustomerCenterLoaded(
...
+            targetState = currentDestination,
+            transitionSpec = {
+                CustomerCenterAnimations.getTransitionForNavigation(
+                    from = initialState,
+                    to = targetState,
+                    navigationState = customerCenterState.navigationState,
+                )
             },
```
JayShortway: Just an idea: in the future we could allow devs to pass `forwardTransition` and `backwardTransition` to the `CustomerCenter`, if we get requests to modify the animations. (Or we could allow these to be remotely configurable.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
JayShortway: That's true. We could traverse the stack from the top and look for the first instance of `Main`. That implementation would also allow this function to be more generic, e.g. `popTo(CustomerCenterDestination)`.

JayShortway (commented):
I think this is looking great! Just a few minor comments.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
tonidero: Ahh nice TIL! Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: I added it becasue I was having some issues with overlaying content, but I think those got fixed other way soI will try to remove this and the clipping

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: yes! I had the exact same thought

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterDestination.kt
vegaro: I think that makes sense yes. I am doing some stuff with the title in https://github.com/RevenueCat/purchases-android/pull/2459/ that could be simplified by doing that

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
vegaro: I agree. I will focus on the Customer Center for now and revisit it because I am worried I would have to redo a lot of things. Making sure I am not breaking any flow is a bit cumbersome and I don't want to delay the project more

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
vegaro: Yes! Good point. Will add some tests

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
vegaro: This `replace` function  is some leftovers. I will remove

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterDestination.kt
tonidero: One thing to consider, we might need to localize the title... So we just need to make sure we support that.

vegaro: Yes! they come localized from the backend

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -535,19 +554,36 @@ internal class CustomerCenterViewModelImpl(
     override fun onNavigationButtonPressed(context: Context, onDismiss: () -> Unit) {
         val currentState = _state.value
         // Handle special case for promotional offers first
-        if (currentState is CustomerCenterState.Success && currentState.promotionalOfferData != null) {
-            dismissPromotionalOffer(context, currentState.promotionalOfferData.originalPath)
+        if (currentState is CustomerCenterState.Success &&
+            currentState.currentDestination is CustomerCenterDestination.PromotionalOffer
```
facumenzella, line 558: Reading this more closely, would it make sense to rename `currentDestination` to `destination`? Feels too verbose

vegaro: yeah I agree, the current doesn't add anything

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -157,23 +158,35 @@ internal class CustomerCenterViewModelImpl(
...
             return
         }
         viewModelScope.launch {
-            handlePromotionalOffer(context, purchaseInformation?.product, path.promotionalOffer, path)
+            val promotionalOfferDisplayed =
+                handlePromotionalOffer(context, purchaseInformation?.product, path.promotionalOffer, path)
+            if (!promotionalOfferDisplayed) {
+                mainPathAction(path, context)
+            }
         }
```
facumenzella, line 190: nit: After a promo offer we're going back to the detail page (if there's more than one subscription). I understand this might not be that easy here, but just to have it in mind

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+import java.util.ArrayDeque
+import java.util.Deque
+
+@Immutable
+internal data class CustomerCenterNavigationState(
+    private val mainTitle: String?,
+    val backStack: Deque<CustomerCenterDestination> = ArrayDeque<CustomerCenterDestination>().apply {
+        // Setting this to null allows the Main destination to be displayed without a title until we have a top bar
+        push(CustomerCenterDestination.Main(null))
```
facumenzella:
```suggestion
        push(CustomerCenterDestination.Main(title = null))
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterNavigationState.kt
```diff
@@ -0,0 +1,76 @@
...
+    },
+) {
+    val currentDestination: CustomerCenterDestination
+        get() = backStack.peek() ?: CustomerCenterDestination.Main(title = null)
```
facumenzella, line 17: nit: what about `backDestination`? I know I said `destination` before hehe, but this is really the top of the backStack, so unless I am missing something it's "where to go if back is pressed", right?

tonidero: FWIW, `back stack` is a pretty common Android term, and it's what this is... So I like it as it is 😇 . Also, I think this includes the "current" destination in the stack right? So it's the stack that gets popped every time the back button is pressed I guess 😅

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
tonidero: I kinda like having `current`... I guess it's kinda implicit, but better to be explicit so there is less confussion... I think just `destination` could actually be a bit more confusing... Not a strong opinion though.

vegaro: hmmm I see your point. Right thisis the peek of the queue, so its actually current... maybe screen? currentScreen? and rename `destination` to `screen`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: do you mean in iOS?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
tonidero: I guess destination is the most commonly used since destinations could include places other than screens... Not sure if we ever would like to have other destinations though, like sheets or dialogs... But probably not? Or they can be implemented in a different way I guess... 

I wouldn't be opposed. Wdyt @JayShortway ?

vegaro: Right, a destination could be its own whole thing with its own screens

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
facumenzella: Yes, in iOS we go back to the detail if there's more than one (not to the list)

vegaro: Ok. I will see if that's easy to do in the other PR

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
JayShortway: I'd have a slight preference for `destination` over `screen`, because indeed it could be something other than a screen. I don't mind the `current`, but can live without it as well.

vegaro: I am just going to leave it as is then, I feel it's minor

tonidero (approved):
I think this looks good! Nice to have some proper navigation in place :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -251,45 +263,83 @@ private fun CustomerCenterLoaded(
...
+        when (destination) {
+            is CustomerCenterDestination.Main -> {
+                MainScreenContent(
+                    state = customerCenterState,
+                    onAction = onAction,
+                )
+            }
+
+            is CustomerCenterDestination.FeedbackSurvey -> {
+                FeedbackSurveyView(destination.data)
```
facumenzella, line 301: ultra-nit: We have `MainScreenContent`, `FeedbackSurveyView` and `PromotionalOfferScreen`. Maybe we can unify the criteria and call all of them `Content` ? Definitely can be done in another PR

facumenzella (approved):
Awesome work @vegaro

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: Makes sense, I think `Content` is good

Will do in another PR since making that change will introduce conflicts in other PRs that are based off this one

PR #2459, opened 2025-06-20
Use Material TopAppBar in Customer Center

vegaro:
Replaced the title in the main screen with a `LargeTopAppBar` that feels more native and will animate when scrolling up

https://github.com/user-attachments/assets/b1dc2f38-843e-4689-9bc4-7d5770fb49b9


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -200,49 +184,160 @@ private fun InternalCustomerCenter(
...
+                CustomerCenterNavigationIcon(
+                    navigationButtonType = scaffoldConfig.navigationButtonType,
+                    onAction = onAction,
                 )
-            }
-        }
-        mainContent()
+            },
+            colors = TopAppBarDefaults.topAppBarColors(
+                containerColor = MaterialTheme.colorScheme.background,
```
tonidero: I see in this one we're not setting `scrolledContainerColor`. Just making sure that's intended.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -200,49 +184,160 @@ private fun InternalCustomerCenter(
...
+@Composable
+private fun CustomerCenterTopBar(
+    scaffoldConfig: CustomerCenterScaffoldConfig,
+    scrollBehavior: TopAppBarScrollBehavior?,
+    onAction: (CustomerCenterAction) -> Unit,
+) {
+    if (scaffoldConfig.shouldUseLargeTopBar) {
+        LargeTopAppBar(
+            title = {
+                Text(text = scaffoldConfig.title ?: "")
```
tonidero: I guess here we need to default to `""` so it still leaves the space/padding correct? (mostly seeing as it's different than the small app bar version)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -200,49 +184,160 @@ private fun InternalCustomerCenter(
...
+    val title: String?,
+    val navigationButtonType: CustomerCenterState.NavigationButtonType,
+    val shouldUseLargeTopBar: Boolean,
+)
+
+private fun createScaffoldState(state: CustomerCenterState): ScaffoldConfigData {
+    if (state is CustomerCenterState.Success) {
+        val title = state.navigationState.currentDestination.title
+        val navigationButtonType = state.navigationButtonType
+        // Only use large top bar for Main screen with active subscriptions
```
tonidero: Quick question but are we checking if the main screen has active subscriptions here? I see the check for `title != null` but that doesn't apparently seem related to whether there are subscriptions or not...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -200,49 +184,160 @@ private fun InternalCustomerCenter(
     }
 }
 
+@Composable
+private fun createColorScheme(state: CustomerCenterState) = if (state is CustomerCenterState.Success) {
```
JayShortway: This could probably be wrapped in `remember(state) { }`? Would be a minor perf improvement.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -200,49 +184,160 @@ private fun InternalCustomerCenter(
...
+    MaterialTheme.colorScheme
+}
+
+private data class ScaffoldConfigData(
+    val title: String?,
+    val navigationButtonType: CustomerCenterState.NavigationButtonType,
+    val shouldUseLargeTopBar: Boolean,
+)
+
+private fun createScaffoldState(state: CustomerCenterState): ScaffoldConfigData {
```
JayShortway, line 227: This one as well I think, can probably be wrapped in `remember(state) { }`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -200,49 +184,160 @@ private fun InternalCustomerCenter(
...
+    } else {
+        null
+    }
+
+    Scaffold(
+        modifier = if (scrollBehavior != null) {
+            modifier.nestedScroll(scrollBehavior.nestedScrollConnection)
+        } else {
+            modifier
+        },
```
JayShortway: I _think_ we have `applyIfNotNull`, or something along those lines?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/ManageSubscriptionsView.kt
```diff
@@ -75,9 +71,9 @@ internal fun ManageSubscriptionsView(
 }
 
 @SuppressWarnings("LongParameterList")
+@Suppress("UnusedParameter")
```
JayShortway: Which parameter is unused?

JayShortway (approved):
Love it! Just some minor stuff.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: good catch!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: I actually just tested this and it looks like it still leaves the space if the Text composable is not there, so I will do the same in both

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: that's a confusing comment I made lol. `Main` is the screen with active subscriptions, that's what I meant. I will remove it. The logic is correct

PR #2462, opened 2025-06-23
fix potential infinite loop in paywalls

ajpallares:
### Motivation
Potential fix for https://github.com/RevenueCat/purchases-flutter/issues/1346 where `OriginalTemplatePaywallFooterView` can become unresponsive in Flutter due to an infinite loop.

### Description
Uses `dataHash` to compare two `PaywallOptions` instances instead of using equals, as documented in https://github.com/RevenueCat/purchases-android/blob/180650266d560006fb961d18d742fd910a2e5c0c/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt#L49-L52

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
@@ -488,7 +488,7 @@ class PaywallViewModelTest {
     }
 
     @Test
-    fun `updateState does not update if paywallsOptions identity doesn't change`() {
+    fun `updateState does not update if paywallsOptions identity does not change`() {
         val options1 = PaywallOptions.Builder(dismissRequest = { dismissInvoked = true })
             .setListener(listener)
             .build()


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -110,7 +110,7 @@ internal class PaywallViewModelImpl(
     }
 
     fun updateOptions(options: PaywallOptions) {
-        if (this.options != options) {
+        if (this.options.dataHash != options.dataHash) {
```
JayShortway: If the `listener` is the only thing that has changed in the `options`, this change will prevent the ViewModel from getting a reference to that new `listener`, which would be incorrect I think.

(`PaywallOptions` contains the `listener`, but the `listener` is excluded from `dataHash`.)

JayShortway: I'm curious what pointed in this direction? Did you have log messages are something like that? Maybe the issue is higher up, e.g. what if the listener incorrectly keeps getting updated?

ajpallares: Ohh, that's actually a great point I hadn't considered. Too naive from me 😞 
But perhaps the solution could go in this vein. Like, always setting `this.options = options`, but only calling `updateState()` if `options.dataHash` actually changed.

> I'm curious what pointed in this direction? Did you have log messages are something like that?

Exactly. I added many logs and stumbled upon this options comparison. Then I saw the comments on `dataHash` and it made sense.

> Maybe the issue is higher up, e.g. what if the listener incorrectly keeps getting updated?

You're right that we should consider this. However, I still believe that, based on the `dataHash` comment, we should avoid doing equals comparisons directly on `options`. WDYT?

ajpallares: I've done this
> always setting `this.options = options`, but only calling `updateState()` if `options.dataHash` actually changed.

in https://github.com/RevenueCat/purchases-android/pull/2462/commits/6712b2af51ad37ddb269624b3f0616e9cd9f8ad4, at least to prevent the potential issue with the `listener`

JayShortway: >However, I still believe that, based on the dataHash comment, we should avoid doing equals comparisons directly on options. WDYT? 
  
I tend to agree. It's a bit murky because `PaywallOptions` is doing 2 jobs: it holds data and logic. 

@tonidero When implementing `dataHash`, you [mentioned](https://github.com/RevenueCat/purchases-android/pull/1854#discussion_r1773647924) you're not overriding `hashCode()` because of this `updateOptions()` function. If we're making this change, we might as well override `hashCode()` right? What do you think?

tonidero: > always setting this.options = options, but only calling updateState() if options.dataHash actually changed.

Hmm I don't remember if there is any state that depended on the listener, we should just make sure nothing needs to be recreated when the listener changes to make sure.

> not overriding hashCode() because of this updateOptions() function

Actually, it's mostly because it acts as the key of the viewmodel factory: https://github.com/RevenueCat/purchases-android/blob/180650266d560006fb961d18d742fd910a2e5c0c/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt#L215

> If we're making this change, we might as well override hashCode() right?

Hmm what would be the benefit? We still need to differentiate between "Anything changed in the paywall options" (the equals method can give us this) vs "Anything changed other than the listeners/fontprovider" (the dataHash calculation gives us this).

I do wonder what could be changing on the options to cause this infinite loop though... As in are they changing the listener every time? (haven't had time to look at the github issue)

ajpallares: > Hmm I don't remember if there is any state that depended on the listener, we should just make sure nothing needs to be recreated when the listener changes to make sure.

I took a look at the `updateState()` method and there doesn't seem to be anything else that depends on the listener changing. The issue I see is that even if that's true now, that could potentially change in the future and it would be difficult to prevent.

> I do wonder what could be changing on the options to cause this infinite loop though... As in are they changing the listener every time? (haven't had time to look at the github issue)

Yeah, I'll take a closer look and will try to detect what's the exact cause of this change. Something must be changing because the two `options` are not equal. It also happens in a very specific scenario (which is also easy to reproduce FWIW). My (totally inexpert) two cents is that this is something inherent to how flutter manages screens, because I was not able to reproduce the issue with a similar Sample done directly in Android native.

Will come back to this with (hopefully) more detailed information. 

Thank you both!

JayShortway: >We still need to differentiate between "Anything changed in the paywall options" [...] vs "Anything changed other than the listeners/fontprovider"
  
My thought was we might not need the former anymore, due to the changes in this PR. If that's correct, then overriding `hashCode()` is less confusing I think. Not a blocker at all, just an idea.

tonidero: I think we still need the former to create the view model. Otherwise we would be creating new models with changes to listener/fontProvider which might happen in the hybrids I think

JayShortway: If I understand correctly we need the latter to create the ViewModel haha

tonidero: 🤦‍♂️sorry you’re right on that. And yeah I think you are right then! We would just need a hashCode that checks for everything except the listener and fontProvider. That would indeed clean this up!

ajpallares: Okay then. On it!

ajpallares: Done in https://github.com/RevenueCat/purchases-android/pull/2462/commits/560c98b3ae767cbc860381e68533f85fd4826b1d and https://github.com/RevenueCat/purchases-android/pull/2462/commits/3a969f462fa1d8d7eb17e1906b995ed8a3758e97
Thank you both for all your input!

tonidero (approved):
I think this looks great! Thanks for doing this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -110,8 +110,11 @@ internal class PaywallViewModelImpl(
     }
 
     fun updateOptions(options: PaywallOptions) {
-        if (this.options != options) {
-            this.options = options
+        val needsUpdateState = this.options.hashCode() != options.hashCode()
```
JayShortway, line 113: If I'm not wrong, the regular equals check would work now that we have overridden the `hashCode()`. Did you happen to try this? 
```suggestion
        val needsUpdateState = this.options != options
```

ajpallares: Yep. I did in https://github.com/RevenueCat/purchases-android/commit/560c98b3ae767cbc860381e68533f85fd4826b1d and... it didn't work. That's why I'm doing the `hashCode()` comparison in the end

JayShortway: I see! Turns out `equals()` in a data class doesn't even use `hashCode()`. Not sure why I assumed that.

Here's the decompiled Java implementation of `equals()`, if you're interested haha:
```java
public boolean equals(@Nullable Object other) {
   if (this == other) {
      return true;
   } else if (!(other instanceof PaywallOptions)) {
      return false;
   } else {
      PaywallOptions var2 = (PaywallOptions)other;
      if (!Intrinsics.areEqual(this.offeringSelection, var2.offeringSelection)) {
         return false;
      } else if (this.shouldDisplayDismissButton != var2.shouldDisplayDismissButton) {
         return false;
      } else if (!Intrinsics.areEqual(this.fontProvider, var2.fontProvider)) {
         return false;
      } else if (!Intrinsics.areEqual(this.listener, var2.listener)) {
         return false;
      } else if (!Intrinsics.areEqual(this.purchaseLogic, var2.purchaseLogic)) {
         return false;
      } else if (this.mode != var2.mode) {
         return false;
      } else {
         return Intrinsics.areEqual(this.dismissRequest, var2.dismissRequest);
      }
   }
}
```

JayShortway (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
ajpallares: Ohh nice to know! I also assumed the same. Thanks for the investigation! 🤩

PR #2463, opened 2025-06-24
Expose CustomerInfo StateFlow

vegaro:
## Summary
- expose CustomerInfo as a `StateFlow`
- keep flow in `CustomerInfoUpdateHandler`
- surface the flow through `PurchasesOrchestrator` and `Purchases`


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -112,6 +113,13 @@ class Purchases internal constructor(
             purchasesOrchestrator.updatedCustomerInfoListener = value
         }
 
+    /**
+     * A [StateFlow] that emits [CustomerInfo] updates.
+     * The first value is emitted from cache when available.
+     */
+    val customerInfo: StateFlow<CustomerInfo?>
+        get() = purchasesOrchestrator.customerInfoStateFlow
```
JayShortway, line 121: I was thinking we have these options: 
- `StateFlow<CustomerInfo?>`
- `Flow<CustomerInfo>`

Honestly not sure what I prefer. The former is suboptimal from an API standpoint because we _always_ have a CustomerInfo after the first fetch, however consumers will need to check for `null` on every emission. (Nothing guarantees that only the first value will ever be `null`.)

The second is suboptimal because it doesn't allow synchronous access.

This is not really a comment, more rambling haha.

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfoUpdateHandler.kt
```diff
@@ -36,12 +39,21 @@ internal class CustomerInfoUpdateHandler(
 
     private var lastSentCustomerInfo: CustomerInfo? = null
 
+    private val _customerInfoStateFlow = MutableStateFlow<CustomerInfo?>(null)
+
+    init {
+        _customerInfoStateFlow.value = getCachedCustomerInfo(identityManager.currentAppUserID)
+    }
```
JayShortway, line 46: I think this should work?
```suggestion
    private val _customerInfoStateFlow = MutableStateFlow<CustomerInfo?>(
        getCachedCustomerInfo(identityManager.currentAppUserID)
    )
```

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfoUpdateHandler.kt
```diff
@@ -36,12 +39,21 @@ internal class CustomerInfoUpdateHandler(
...
+    val customerInfoStateFlow: StateFlow<CustomerInfo?> = _customerInfoStateFlow.asStateFlow()
+
     fun cacheAndNotifyListeners(customerInfo: CustomerInfo) {
         deviceCache.cacheCustomerInfo(identityManager.currentAppUserID, customerInfo)
         notifyListeners(customerInfo)
     }
 
     fun notifyListeners(customerInfo: CustomerInfo) {
+        _customerInfoStateFlow.value = customerInfo
         synchronized(this@CustomerInfoUpdateHandler) { updatedCustomerInfoListener to lastSentCustomerInfo }
```
JayShortway, line 57: Oh right, I was thinking we need to implement this as a listener that updates a flow, but just directly updating a flow from here works as well of course.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -112,6 +113,13 @@ class Purchases internal constructor(
             purchasesOrchestrator.updatedCustomerInfoListener = value
         }
 
+    /**
+     * A [StateFlow] that emits [CustomerInfo] updates.
+     * The first value is emitted from cache when available.
+     */
+    val customerInfo: StateFlow<CustomerInfo?>
```
JayShortway, line 120: Thinking of Java... should this be an extension instead, like our suspend functions?

JayShortway (commented):
This is nice! Just a few minor things.

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -55,6 +56,13 @@ class Purchases internal constructor(
             purchasesOrchestrator.updatedCustomerInfoListener = value
         }
 
+    /**
+     * A [StateFlow] that emits [CustomerInfo] updates.
+     * The first value is emitted from cache when available.
+     */
+    val customerInfo: StateFlow<CustomerInfo?>
```
tonidero: I don't think we should expose this in customEntitlementComputation mode. We don't really expose the getCustomerInfo anywhere here right?

vegaro: oopsies

PR #2466, opened 2025-06-25
Break VCs Out of CustomerInfo: API & Caching

fire-at-will:
### Description
This PR contains the first major part of breaking virtual currencies out of CustomerInfo in purchases-android. Specifically, it creates the new public-facing APIs and implements the caching layer.

### Changes Introduced
- Removes virtual currencies from CustomerInfo
- Updates the VirtualCurrency objects to match what we have on iOS:
   - `VirtualCurrencyInfo` is renamed `VirtualCurrency`
   - Created a `VirtualCurrencies` object
- Implements the skeleton of the `VirtualCurrencyManager`
- Implements the caching layer to cache virtual currencies
- Adds new APIs to Purchases:
   - `cachedVirtualCurrencies` computed property
   - `invalidateVirtualCurrenciesCache()
   - `getVirtualCurrencies()`

### Testing Done
- Included unit tests for new object deserialization, caching, and for the new logic in `VirtualCurrencyManager `
- Included API tests for the new objects and APIs

### Next Steps
In a follow up PR, we will implement fetching virtual currencies from the network, add more unit tests for the `VirtualCurrencyManager`, and add backend integration tests.

To keep the `virtual-currency-dev` branch clean, this PR will be merged into a `fetch-virtual-currencies-new-endpoint` branch, which is a clone of the `virtual-currency-dev` branch. We'll make future PRs go into the `fetch-virtual-currencies-new-endpoint` branch, and when the functionality of the new APIs is complete, we'll merge `fetch-virtual-currencies-new-endpoint` into `virtual-currency-dev`.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManagerTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManagerTest.kt
@@ -77,6 +77,9 @@ class VirtualCurrencyManagerTest {
         every {
             mockDeviceCache.isVirtualCurrenciesCacheStale(any(), any())
         } returns true
+        every {
+            mockDeviceCache.cacheVirtualCurrencies(any(), any())
+        } returns Unit
 
         val mockIdentityManager = mockk<IdentityManager>()
         every {
@@ -104,10 +107,15 @@ class VirtualCurrencyManagerTest {
 
         verify(exactly = 1) {
             mockDeviceCache.isVirtualCurrenciesCacheStale(appUserID = appUserID, appInBackground = false)
-            // TODO: Verify that a network call was made
 
-            // TODO: Verify that the returned VCs match the ones returned from the network call
-//            mockCallback.onReceived(this@VirtualCurrencyManagerTest.virtualCurrencies)
+            // TODO: Ensure that the VCs we cache come from the mocked backend
+//            mockDeviceCache.cacheVirtualCurrencies(
+//                appUserID = appUserID,
+//                virtualCurrencies = this@VirtualCurrencyManagerTest.virtualCurrencies
+//            )
+
+            // TODO: Ensure that we send the VCs from the network call to the callback
+            mockCallback.onReceived(VirtualCurrencies(all = emptyMap()))
         }
         
         verify(exactly = 0) {


purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -227,6 +240,99 @@ internal open class DeviceCache(
...
+                } catch (e: JSONException) {
+                    null
+                }
+            }
+    }
+
+    @Synchronized
+    fun cacheVirtualCurrencies(appUserID: String, virtualCurrencies: VirtualCurrencies) {
+        val virtualCurrenciesData = JSONObject().apply {
+            put("virtual_currencies", virtualCurrencies.rawData)
```
fire-at-will: Here, we wrap the virtual currencies in a JSON object, keyed by `virtual_currencies` to mimic the response from the backend so that we can reuse the decoding function used in the network layer

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -33,6 +34,7 @@ import java.net.URL
 @Suppress("TooManyFunctions")
 class Purchases internal constructor(
     @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
+    @get:JvmSynthetic internal val virtualCurrencyManager: VirtualCurrencyManager,
```
fire-at-will: I'm honestly not super excited about adding `virtualCurrencyManager` to the CEC Purchases object since virtual currencies aren't available in CEC mode, but given that the `virtualCurrencyManager` won't be used here and this is an internal constructor, I suppose it's okay?

I'm super open to suggestions on any ways to avoid adding this to the constructor here if you have them!

purchases/src/test/java/com/revenuecat/purchases/common/DeviceCacheTest.kt
```diff
@@ -681,6 +691,112 @@ class DeviceCacheTest {
...
+    @Test
+    fun `given a valid VirtualCurrencies, the JSON is parsed correctly`() {
+        mockString(cache.virtualCurrenciesCacheKey(appUserID), Responses.validFullVirtualCurrenciesResponse)
+        val vcs = cache.getCachedVirtualCurrencies(appUserID)
+        assertThat(vcs).`as`("vcs is not null").isNotNull
+        assertThat(vcs?.all?.size).isEqualTo(2)
+    }
+
+    @Test
+    fun `given a invalid VirtualCurrencies, the information is null`() {
```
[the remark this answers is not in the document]
fire-at-will: Improved the test's name here: c40f3b41de2b95470345e7a3cb0da3c25f9690bf

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt
```diff
@@ -116,6 +125,9 @@ private class PurchasesAPI {
         purchases.redeemWebPurchase(webPurchaseRedemption, redeemWebPurchaseListener)
         val parsedWebPurchaseRedemption: WebPurchaseRedemption? = Purchases.parseAsWebPurchaseRedemption(intent)
         val parsedWebPurchaseRedemption2: WebPurchaseRedemption? = Purchases.parseAsWebPurchaseRedemption("")
+
+        purchases.getVirtualCurrencies(callback = getVirtualCurrenciesCallback)
+        val cachedVirtualCurrencies: VirtualCurrencies? = purchases.cachedVirtualCurrencies
```
ajpallares:
```suggestion
        purchases.getVirtualCurrencies(callback = getVirtualCurrenciesCallback)
        purchases.invalidateVirtualCurrenciesCache()
        val cachedVirtualCurrencies: VirtualCurrencies? = purchases.cachedVirtualCurrencies
```

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
ajpallares: I don't know if this would be valid, but coming from the iOS world, I was thinking of creating an `init` where the `virtualCurrencyManager` parameter wouldn't get set. 

It would look something like this I think
```kotlin
class Purchases internal constructor(
    @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
) : LifecycleDelegate {

    @Suppress("UNUSED_PARAMETER")
    internal constructor(purchasesOrchestrator: PurchasesOrchestrator, virtualCurrencyManager: VirtualCurrencyManager):
        this(purchasesOrchestrator)
```

Maybe this would be slightly better than keeping the reference?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -462,6 +466,40 @@ class Purchases internal constructor(
         purchasesOrchestrator.getCustomerInfo(fetchPolicy, true, callback)
     }
 
+    /**
+     * Fetches the virtual currencies for the current subscriber.
+     *
+     * @param callback A listener called when the virtual currencies are available.
+     */
+    fun getVirtualCurrencies(
```
ajpallares, line 472: In iOS, this API is
```swift
func virtualCurrencies(completion: )
```
https://github.com/RevenueCat/purchases-ios/blob/a75fe7fa0e5c124eab87d4c4e80a8d9d38316f86/Sources/Purchasing/Purchases/PurchasesType.swift#L990-L993
Unless the difference is due to naming conventions, shouldn't these be named the same if possible?

Seeing other APIs in iOS, I'd say the iOS one should be renamed to `getVirtualCurrencies`

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -462,6 +466,40 @@ class Purchases internal constructor(
...
+     * or if you increased the balance from your backend using the server APIs.
+     *
+     * For more info, see our [virtual currency docs](https://www.revenuecat.com/docs/offerings/virtual-currency)
+     */
+    fun invalidateVirtualCurrenciesCache() {
+        virtualCurrencyManager.invalidateVirtualCurrenciesCache()
+    }
+
+    /**
+     * The currently cached ``VirtualCurrencies`` if one is available.
```
ajpallares: Looking below, I think the double inverted quotes are not the way to do this in Kotlin (too iOS-y)
```suggestion
     * The currently cached [VirtualCurrencies] if one is available.
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
ajpallares: Would it make sense to use the `VirtualCurrenciesResponseJsonKeys.VIRTUAL_CURRENCIES` constant here? It would make it more explicit. Perhaps even adding an explanatory comment could be helpful

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyFactory.kt
```diff
@@ -0,0 +1,39 @@
...
+    val code: String = getString(VirtualCurrencyJsonKeys.CODE)
+    val serverDescription: String? = optString(VirtualCurrencyJsonKeys.SERVER_DESCRIPTION).takeIf { it.isNotBlank() }
+
+    return VirtualCurrency(
+        balance = balance,
+        name = name,
+        code = code,
+        serverDescription = serverDescription,
+    )
+}
```
ajpallares: This feels like it's too manual for a straightforward parsing of property. I assume there's some reason not to use the equivalent to iOS' `Decodable`?

ajpallares (approved):
Looks great! I have some suggestions and small comments. Great job!

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyFactory.kt
fire-at-will: Good idea! I've streamlined the JSON encoding/decoding quite a bit by switching the coding of `VirtualCurrencies` and `VirtualCurrency` to using `@Serializable`. This both simplifies the decoding/encoding logic, but also removes the need to wrap cached virtual currencies in a `virtual_currencies` JSON object, as discussed [here](https://github.com/RevenueCat/purchases-android/pull/2466#discussion_r2185795056).

I made these changes in this commit: faf8a093325cdd92f7bde742ea3f2fd5af1d3c9a

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
fire-at-will: Update from a discussion in Slack: We've renamed the iOS API to `getVirtualCurrencies(completion:)` 👍

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
fire-at-will: Great idea!! I've implemented that here: b2448efd3833c8860ea65a86504cc4437bad497d

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -46,6 +49,7 @@ import java.net.URL
  */
 class Purchases internal constructor(
     @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
+    @get:JvmSynthetic internal val virtualCurrencyManager: VirtualCurrencyManager,
```
tonidero: Hmm could we move this dependency to the `PurchasesOrchestrator`? I would consider `Purchases` just a facade, without any real logic, that just calls `PurchasesOrchestrator` as needed. Then, in CEC we just don't add the API methods calling `PurchasesOrchestrator`.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -462,6 +466,40 @@ class Purchases internal constructor(
...
+        virtualCurrencyManager.invalidateVirtualCurrenciesCache()
+    }
+
+    /**
+     * The currently cached [VirtualCurrencies] if one is available.
+     * This is synchronous, and therefore useful for contexts where an app needs a [VirtualCurrencies]
+     * right away without waiting for a callback.
+     *
+     * This allows initializing state to ensure that UI can be loaded from the very first frame.
+     */
```
tonidero: Would be worth adding a note that this will remain empty until virtual currencies have been fetched at least once using `getVirtualCurrencies`?

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -227,6 +240,100 @@ internal open class DeviceCache(
...
+
+    @Synchronized
+    fun getCachedVirtualCurrencies(appUserID: String): VirtualCurrencies? {
+        return preferences.getString(virtualCurrenciesCacheKey(appUserID), null)
+            ?.let { json ->
+                try {
+                    val cachedJSONObject = JSONObject(json)
+                    return VirtualCurrenciesFactory.buildVirtualCurrencies(body = cachedJSONObject)
+                } catch (e: JSONException) {
+                    null
```
tonidero, line 256: Hmm would it be worth adding a log here? (not sure in what situations this may happen.

tonidero (commented):
This looks great to me! Only real comment is to change the `VirtualCurrencyManager` to be a dependency of the `PurchasesOrchestrator` instead of `Purchases`.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
fire-at-will: This works for me - I'll also create a PR to do this in the iOS SDK as well so the architectures are kept in sync with each other.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
fire-at-will: Definitely! I'll add in all of logs together in a follow-up PR to ensure they're done consistently and to keep this PR's size down.

The `JSONException` is thrown when the string you're trying to parse isn't valid JSON, like if you try to parse `asdf`. This _shouldn't_ happen, but better safe than sorry 😄

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
MarkVillacampa: I think on iOS we've been a little less thorough in keeping everything inside PurchasesOrchestrator, and we do have some logic in the Purchases class 😅

https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Purchases/Purchases.swift#L254-L281

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencies.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlinx.parcelize.Parcelize
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * This class contains all the virtual currencies associated to the user.
+ *
+ * @property all - All of the virtual currencies associated to the user.
+ */
+@Poko
```
MarkVillacampa, line 14: I'm not super familiar with the Poko library but I'm reading it synthesises equals and hashCode methods, but we're manually implementing both here, is it necessary? will they interfere?

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrency.kt
```diff
@@ -0,0 +1,45 @@
...
+            serverDescription == other.serverDescription
+    }
+
+    override fun hashCode(): Int {
+        var result = balance.hashCode()
+        result = 31 * result + name.hashCode()
+        result = 31 * result + code.hashCode()
+        result = 31 * result + (serverDescription?.hashCode() ?: 0)
+        return result
+    }
```
MarkVillacampa: same comment here about the Poko library

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
```diff
@@ -0,0 +1,94 @@
...
+            return
+        }
+
+        fetchVirtualCurrenciesFromBackend(
+            appUserID = appUserID,
+            isAppBackgrounded = isAppBackgrounded,
+            callback = callback,
+        )
+
+        // TODO: Cache the VCs from the network
```
MarkVillacampa: I know the fetch from network method is not implemented, but it seems like we could already call `cacheVirtualCurrencies` here? is there something else missing?

purchases/src/test/java/com/revenuecat/purchases/virtualcurrencies/VirtualCurrenciesTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.json.JSONObject
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class VirtualCurrenciesTest {
```
MarkVillacampa, line 13: Maybe i missed them but could we add some equality tests for the newly added Parcelable classes?

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencies.kt
tonidero: Ohh I missed that but you're right! We shouldn't need to write our own. We can write it if we want to overwrite it, so ours would have preference, but no need to write these from what I can read I think.

purchases/src/test/java/com/revenuecat/purchases/virtualcurrencies/VirtualCurrenciesTest.kt
fire-at-will: Added them here! f13837425813880424dfdd524606086c2894b237

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
fire-at-will: Moved `VirtualCurrencyManager` from `Purchases` to `PurchasesOrchestrator` here! 5db3c4db5847bbbd45cee42f222249805e08c8e3

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
fire-at-will: Updated in b23325e5079a152024c7cc42c288253447dcfb89!

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencies.kt
fire-at-will: Yeah, I think we can rely on Poko's, I don't see any reason to write custom ones here. I'll remove them!

fire-at-will: Removed them here! 3e1dc504ca6505d0b1fcad2e77149771d74344db

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrency.kt
fire-at-will: Removed in 3e1dc504ca6505d0b1fcad2e77149771d74344db

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
fire-at-will: All that's missing is the network call that gives us the value to provide to `cacheVirtualCurrencies()`. I was going to add it here, but this gave a compilation error since `virtualCurrenciesFromNetwork` isn't initialized:

```kotlin
val virtualCurrenciesFromNetwork: VirtualCurrencies // TODO: Fetch from network
        cacheVirtualCurrencies(
            virtualCurrencies = virtualCurrenciesFromNetwork,
            appUserID = appUserID
        )
```

fire-at-will: Added this here: 9ac9904e813e51cc88ebfba1ed615e76008771cb

PR #2467, opened 2025-06-26
Add GitHub Issue Action Ack

joshdholtz:
### Motivation

Get notified of important issue

### Description

Add workflow

### DIFF
--- a/.github/workflows/issue-notifications.yml
+++ b/.github/workflows/issue-notifications.yml
@@ -1,51 +1,19 @@
 name: Issue Notifications
 
 on:
-  workflow_dispatch: # Manual trigger for testing
+  workflow_dispatch:
   issues:
     types: [opened, edited]
   issue_comment:
     types: [created]
   schedule:
-    - cron: "0 * * * *" # Every hour
+    - cron: "0 * * * *"
 
 jobs:
   notify:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Issue Notification Action
-        uses: joshdholtz/github-action-issue-ack@v15
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
-        with:
-          slack_webhook_url: ${{ secrets.ACK_SLACK_WEBHOOK_URL }}
-
-          # Keywords to raise alert
-          title_keywords: ${{ secrets.ACK_ALERT_KEYWORDS }}
-
-          # Excluded labels to silence notifications
-          excluded_labels: "acknowledged"
-
-          # Reaction and comment thresholds to raise alert
-          reaction_threshold: ${{ secrets.ACK_REACTION_THRESHOLD }}
-          comment_threshold: ${{ secrets.ACK_COMMENT_THRESHOLD }}
-
-          # Batch checking configuration
-          check_all_open_issues: "true"
-          max_issues_to_check: "100"
-          issue_state: "open"
-
-          # Custom message for testing
-          message_template: |
-            *{title}*
-            Created: {created_ago}
-            Reactions: {reactions} | Comments: {comments}
-            {url}
-
-          # Custom prefixes for testing
-          new_issue_prefix: "🚨 New issue created on *RevenueCat/purchases-android*!"
-          threshold_prefix: "📈 High-engagement issue on *RevenueCat/purchases-android*!"
-
-          # Enable both notification types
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-notifications.yml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_dispatch: # Manual trigger for testing
+  issues:
+    types: [opened, edited]
+  issue_comment:
+    types: [created]
+  schedule:
+    - cron: "0 * * * *" # Every hour
+
+jobs:
+  notify:
```
tonidero, line 13: Hmm I wonder if it would make sense to extract this to a shared repo... Seems it's pretty much duplicated in each PR right?

.github/workflows/issue-notifications.yml
```diff
@@ -0,0 +1,51 @@
+name: Issue Notifications
+
+on:
+  workflow_dispatch: # Manual trigger for testing
+  issues:
+    types: [opened, edited]
+  issue_comment:
+    types: [created]
+  schedule:
+    - cron: "0 * * * *" # Every hour
```
tonidero: I wonder... should we do this every day? Just so it doesn't get too noisy... But then again, I'm ok trying this for now!

.github/workflows/issue-notifications.yml
JayShortway: Yea, seems like it's possible to reuse a workflow: https://docs.github.com/en/actions/sharing-automations/reusing-workflows

.github/workflows/issue-notifications.yml
ajpallares: I agree that every hour could be too much

.github/workflows/issue-notifications.yml
```diff
@@ -0,0 +1,51 @@
...
+          # Reaction and comment thresholds to raise alert
+          reaction_threshold: ${{ secrets.ACK_REACTION_THRESHOLD }}
+          comment_threshold: ${{ secrets.ACK_COMMENT_THRESHOLD }}
+
+          # Batch checking configuration
+          check_all_open_issues: "true"
+          max_issues_to_check: "100"
+          issue_state: "open"
+
+          # Custom message for testing
```
vegaro: the "for testing" is confusing me

joshdholtz: Ha! I copied this from my test project 😛 I will fix

.github/workflows/issue-notifications.yml
joshdholtz: OMG, I didn't know reusable workflows were a thing... ON IT

.github/workflows/issue-notifications.yml
```diff
@@ -0,0 +1,19 @@
+name: Issue Notifications
+
+on:
+  workflow_dispatch:
+  issues:
+    types: [opened, edited]
+  issue_comment:
+    types: [created]
+  schedule:
+    - cron: "0 * * * *"
```
tonidero, line 10: Still wondering if every hour might be too much... But we can go with this and check how noisy it is.

joshdholtz: The point is to find the important crash / high engagement issues as soon as possible soooo… 😊 <br><br>If too noisey, I would probably suggest we change keywords and/or reaction and comment thresholds 🤷‍♂️

.github/workflows/issue-notifications.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  notify:
+    uses: revenuecat/sdk-github-workflows/.github/workflows/issue-notifications.yml@v1
```
tonidero, line 14: Do we need to pass the repo name as a parameter? Or does it pick it up automatically? Honestly now sure where `repo_name` is coming from here: https://github.com/RevenueCat/sdk-github-workflows/blob/main/.github/workflows/issue-notifications.yml

joshdholtz: Its a variable used in the messaging templates here - https://github.com/joshdholtz/github-action-issue-ack

Screenshot 2025-06-26 at 7 40 16 AM

tonidero: Ahh perfect! Then we should be good, thanks!

PR #2468, opened 2025-06-26
Add `managementURL` to `SubscriptionInfo`

vegaro:
Added `managementURL` to individually manage subscriptions. This offers better support for cross platform subscriptions


facumenzella (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt
```diff
@@ -160,6 +165,59 @@ class SubscriptionInfo(
         displayName = null,
         price = null,
         productPlanIdentifier = null,
+        managementURL = null,
+        requestDate = requestDate,
+    )
+
+    @Deprecated(
```
vegaro, line 172: @facumenzella I didn't realize I was breaking the public API so I added this

facumenzella: Why is the initializer even public? 🤔

vegaro: For testing for example

PR #2469, opened 2025-06-26
Purchases the right package when a purchase button is inside a package component

JayShortway:
## Description
We're now correctly telling the ViewModel which package to purchase.


tonidero (approved):
I do think this should fix it. Looks great. Thanks for the fix!!

PR #2471, opened 2025-06-27
Issue when loading promotional offers when products are missing base plans

vegaro:
There's an issue when loading promotional offers for target products that miss a base plan

For example this configuration is valid if the product is old and configured without base plans in the dashboard:

Screenshot 2025-06-27 at 12 05 36

The SDK will fail to load the promotional becasue it looks for the `product.id`, which in that case would look like `annual_freetrial:p1y`. On the other hand `product.googleProduct?.productId` has only the product id `annual_freetrial` so we should be looking for that instead if we can't find a promotional matching the base plan.

There was also another issue when loading the target product, since we were picking the first one returned, which is not ideal. For example, if the product "subscription" had "monthly" and "annual" base plans. And the target product was just "subscription", `Purchases.getProducts` without an specific base plan returns two products, one for each base plan. Which could load an unexpected product. I changed the logic to load the base plan that's equivalent to the currently active subscription.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
@@ -665,12 +665,13 @@ internal class CustomerCenterViewModelImpl(
         promotionalOffer: CustomerCenterConfigData.HelpPath.PathDetail.PromotionalOffer,
         product: StoreProduct,
     ): SubscriptionOption? {
+        val googleProduct = product.googleProduct
         val crossProductPromotion: CrossProductPromotion? =
             promotionalOffer.crossProductPromotions[product.id]
                 // Check for cross-product promotions first (product.id includes base plan ex: "sub:p1m")
                 // But it's possible the product is not configured with base plan in RevenueCat
                 // so we also check for the Google product ID (which does not include base plan ex: "sub")
-                ?: promotionalOffer.crossProductPromotions[product.googleProduct?.productId ?: ""]
+                ?: googleProduct?.let { promotionalOffer.crossProductPromotions[it.productId] }
                 ?: promotionalOffer.productMapping[product.id]?.let {
                     CrossProductPromotion(
                         storeOfferIdentifier = it,
@@ -687,12 +688,12 @@ internal class CustomerCenterViewModelImpl(
 
         val targetProduct: StoreProduct? = when {
             crossProductPromotion.targetProductId == product.id -> product
-            product.googleProduct?.basePlanId != null ->
-                // Passing oringinal product's base plan ID to find the target product
+            googleProduct?.basePlanId != null ->
+                // Passing original product's base plan ID to find the target product
                 // in case the target product is missing the base plan ID in the dashboard
                 // which is common for old products. Purchases.getProducts would return all products
                 // with the same product ID but different base plan IDs. That way we can find the most relevant product.
-                findTargetProduct(crossProductPromotion.targetProductId, product.googleProduct!!.basePlanId!!)
+                findTargetProduct(crossProductPromotion.targetProductId, googleProduct.basePlanId!!)
             else -> null
         }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
@@ -666,6 +666,10 @@ internal class CustomerCenterViewModelImpl(
     ): SubscriptionOption? {
         val crossProductPromotion: CrossProductPromotion? =
             promotionalOffer.crossProductPromotions[product.id]
+                // Check for cross-product promotions first (product.id includes base plan ex: "sub:p1m")
+                // But it's possible the product is not configured with base plan in RevenueCat
+                // so we also check for the Google product ID (which does not include base plan ex: "sub")
+                ?: promotionalOffer.crossProductPromotions[product.googleProduct?.productId ?: ""]
                 ?: promotionalOffer.productMapping[product.id]?.let {
                     CrossProductPromotion(
                         storeOfferIdentifier = it,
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PurchasesType.kt
```diff
@@ -71,9 +71,13 @@ internal class PurchasesImpl(private val purchases: Purchases = Purchases.shared
...
-        return products.firstOrNull {
-            it.googleProduct?.basePlanId == basePlan
-        } ?: products.firstOrNull()
+        // Get products returns one product per base plan ID if it is a subscription product.
+        // If the product is an in-app product, it will return a single product.
+        return if (basePlan == null) {
+            products.firstOrNull()
+        } else {
+            products.firstOrNull { it.googleProduct?.basePlanId == basePlan }
+        }
```
tonidero, line 80: Hmm can you explain a bit more of why this change is needed? If I understand correctly, I think the behavior remains exactly the same as before. In what cases would this change compared to the previous implementation?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -666,6 +667,10 @@ internal class CustomerCenterViewModelImpl(
     ): SubscriptionOption? {
         val crossProductPromotion: CrossProductPromotion? =
             promotionalOffer.crossProductPromotions[product.id]
+                // Check for cross-product promotions first (product.id includes base plan ex: "sub:p1m")
+                // But it's possible the product is not configured with base plan in RevenueCat
+                // so we also check for the Google product ID (which does not include base plan ex: "sub")
+                ?: promotionalOffer.crossProductPromotions[product.googleProduct?.productId ?: ""]
```
tonidero: Nitpick but how about
```suggestion
                ?: product.googleProduct?.let { promotionalOffer.crossProductPromotions[it] }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -680,10 +685,15 @@ internal class CustomerCenterViewModelImpl(
...
         }
 
-        val targetProduct: StoreProduct? = if (crossProductPromotion.targetProductId == product.id) {
-            product
-        } else {
-            findTargetProduct(crossProductPromotion.targetProductId)
+        val targetProduct: StoreProduct? = when {
+            crossProductPromotion.targetProductId == product.id -> product
+            product.googleProduct?.basePlanId != null ->
+                // Passing oringinal product's base plan ID to find the target product
```
tonidero:
```suggestion
                // Passing original product's base plan ID to find the target product
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -680,10 +685,15 @@ internal class CustomerCenterViewModelImpl(
...
+                // in case the target product is missing the base plan ID in the dashboard
+                // which is common for old products. Purchases.getProducts would return all products
+                // with the same product ID but different base plan IDs. That way we can find the most relevant product.
+                findTargetProduct(crossProductPromotion.targetProductId, product.googleProduct!!.basePlanId!!)
```
tonidero: Hmm could be good extracting `product.googleProduct?.basePlanId` to a variable before the `when` statement, so we can do the `when` over that variable and avoid needing the `!!` here.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -702,10 +712,11 @@ internal class CustomerCenterViewModelImpl(
 
     private suspend fun findTargetProduct(
         targetProductId: String,
+        sourceBasePlan: String,
```
tonidero, line 761: I'm finding the logic for promotional offers here is complex enough to warrant separating to a separate class, which can be better tested. Wdyt? Can be done in a separate PR of course.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PurchasesType.kt
vegaro: Yeah I did not change the behavior, I made this change to just to be more explicit in the code about what happens when base plan is null. The fix is passing the `sourceBasePlan` to this function when looking for a target product.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: yeah good idea. I still need the second `!!` becasue "`googleProduct.basePlanId` is a public API property declared in different module"

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: yeah, any thoughts on naming?

tonidero: Maybe a mix of them, but some that come to mind:

- `PromotionalOfferHelper`
- `PromotionalOfferProvider`
- `EligibleOfferCalculator`

Also thought about namespacing them with a `CustomerCenter` prefix... But I think it's not needed if they are in a customercenter package.

vegaro: Sounds good. I took notes to do this after wrapping the other PRs, otherwise there will be conflicts

PR #2473, opened 2025-06-30
Use subscription's managementURL instead of CustomerInfo's in Customer Center

vegaro:
We should be using the managementURL of the displayed subcription not the CustomerInfo one. We needed #2468 for this

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
@@ -42,7 +42,6 @@ internal data class PurchaseInformation(
         entitlementInfo: EntitlementInfo? = null,
         subscribedProduct: StoreProduct? = null,
         transaction: TransactionDetails,
-        managementURL: Uri?,
         dateFormatter: DateFormatter = DefaultDateFormatter(),
         locale: Locale,
     ) : this(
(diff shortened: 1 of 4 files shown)


vegaro:
considering this a fix since it might actually fix some scenarios

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
```diff
@@ -56,7 +55,7 @@ internal data class PurchaseInformation(
             subscribedProduct?.let { PriceDetails.Paid(it.price.formatted) } ?: PriceDetails.Unknown
         },
         isSubscription = transaction is TransactionDetails.Subscription && transaction.store != Store.PROMOTIONAL,
-        managementURL = managementURL,
+        managementURL = (transaction as? TransactionDetails.Subscription)?.managementURL,
```
tonidero, line 58: FFTI but should we fallback to the CustomerInfo's managementUrl? I guess that might only apply for non-subscriptions? In which case, it might be ok to just leave it as null, so FFTI.

tonidero (approved):
:shipit:

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
vegaro: yes, we actually hide the "Cancel" button for non-subscriptions so it will be ignored

PR #2476, opened 2025-07-01
Migrates publishing to Maven Central Portal.

JayShortway:
## Description
Migrates publishing from OSSRH to Central Portal. 

To publish to Central Portal, we need to update the `maven.publish` plugin. Doing so gives us the following error in the `:bom` module:  
>“You need Kotlin version 1.9.20 or newer”. 

I tried to work around it by setting the `languageVersion` and `apiVersion` to 1.8 (our previous Kotlin version), but that only works at the source level. Binaries compiled with 1.9 cannot be read by the 1.7 compiler (our current minimum version). 

As discussed on Slack, we're temporarily disabling publishing the `:bom`, until we either update our Kotlin version or migrate `:bom` publishing to use a different plugin.


gradle/libs.versions.toml
```diff
@@ -31,14 +31,16 @@ fragment = "1.6.1"
 hamcrest = "1.3"
 recyclerview = "1.2.1"
 roboelectric = "4.11.1"
-kotlin = "1.8.22"
+kotlinCompiler = "1.9.20"
+kotlinLanguage = "1.8"
+kotlinApi = "1.8"
```
JayShortway, line 36: `KotlinVersion.fromString()` doesn't support patch versions.

PR #2477, opened 2025-07-01
BC8 migration

tonidero:
### Description
Migrates to BC8: https://developer.android.com/google/play/billing/release-notes#8-0-0


purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
```diff
@@ -242,19 +242,19 @@ internal class AmazonBilling(
...
+    override fun findPurchaseInActivePurchases(
         appUserID: String,
         productType: RevenueCatProductType,
         productId: String,
         onCompletion: (StoreTransaction) -> Unit,
         onError: (PurchasesError) -> Unit,
     ) {
         log(LogIntent.DEBUG) { RestoreStrings.QUERYING_PURCHASE_WITH_TYPE.format(productId, productType.name) }
-        queryAllPurchases(
+        queryPurchases(
```
tonidero, line 253: This is a change in behavior in Amazon... but I think this is fine, since this method was only used for upgrades/downgrades which is not really supported in Amazon. I just changed the implementation to make it more sense with the new method name.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -146,4 +161,14 @@ internal class QueryProductDetailsUseCase(
...
     }
+
+    private fun convertUnfetchedProductStatusCodeToString(statusCode: Int): String {
+        return when (statusCode) {
+            StatusCode.UNKNOWN -> "UNKNOWN"
+            StatusCode.PRODUCT_NOT_FOUND -> "PRODUCT_NOT_FOUND"
+            StatusCode.INVALID_PRODUCT_ID_FORMAT -> "INVALID_PRODUCT_ID_FORMAT"
+            StatusCode.NO_ELIGIBLE_OFFER -> "NO_ELIGIBLE_OFFER"
+            else -> "UNKNOWN_STATUS_CODE: $statusCode"
+        }
```
tonidero, line 141: This was interesting, for now I'm just logging to console, but in the future it might be useful to track this information with diagnostics or something else and expose it in the dashboard in case there are issues.

purchases/src/main/kotlin/com/revenuecat/purchases/common/BillingAbstract.kt
```diff
@@ -44,6 +44,9 @@ internal abstract class BillingAbstract(
         endConnection()
     }
 
+    // Google Play Billing Library 8.0 removed APIs to query historical purchases.
+    // This method will from now on only return active purchases in Google Play.
+    // However, in Amazon, it will still return historical purchases.
```
tonidero: TBH, I wasn't sure whether to just remove this method entirely... but I wanted to keep Amazon having the same functionality as long as it's possible and since this is abstracting the Store, I preferred to keep it. Lmk if you think otherwise.

purchases/src/main/kotlin/com/revenuecat/purchases/common/BillingAbstract.kt
```diff
@@ -65,7 +68,7 @@ internal abstract class BillingAbstract(
     )
 
     @SuppressWarnings("LongParameterList")
-    abstract fun findPurchaseInPurchaseHistory(
+    abstract fun findPurchaseInActivePurchases(
```
tonidero, line 73: This was only being used for upgrades/downgrades. I'm not sure why were were looking in the purchase history, when upgrades/downgrades are only possible with active subs AFAIK. Now we only check for active purchases.

purchases/src/main/kotlin/com/revenuecat/purchases/common/purchaseHistoryRecordExtensions.kt
```diff
@@ -1,25 +0,0 @@
...
-
-internal fun PurchaseHistoryRecord.toHumanReadableDescription() =
-    "skus: ${
-        this.skus.joinToString(prefix = "[", postfix = "]")
-    }, purchaseTime: ${this.purchaseTime}, purchaseToken: ${this.purchaseToken}"
-
-/**
- * **Important**: until multi-line subscriptions are released,
- * we assume that there will be only a single sku in the purchase.
- * https://android-developers.googleblog.com/2021/05/whats-new-in-google-play-2021.html
```
JayShortway, line 15: Well well well 😅

gradle/libs.versions.toml
```diff
@@ -12,7 +12,7 @@ annotation = "1.3.0"
 appcompat = "1.4.1"
 assertJ = "3.22.0"
 browser = "1.8.0"
-billing = "7.1.1"
+billing = "8.0.0"
```
JayShortway, line 15: Did you happen to check whether they're still using Java 8 APIs?

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
JayShortway: This might be a fun interview problem 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/BillingAbstract.kt
JayShortway: Might be good to make this actual documentation of the method, so it's surfaced by the IDE? 
```kotlin
    /**
     * Google Play Billing Library 8.0 removed APIs to query historical purchases.
     * This method will from now on only return active purchases in Google Play.
     * However, in Amazon, it will still return historical purchases.
     */
```

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
JayShortway: Is there any way to test this?

tonidero: Well, as you can see [here](https://github.com/RevenueCat/purchases-android/blob/80dd89ae891f8702be4a39696748dbd2a57e1dcd/purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt#L297), we early exit from the purchase if the developer gives us any replacement information (we actually should improve that, since it would actually keep the call forever without returning an error 😬)

gradle/libs.versions.toml
tonidero: I didn't but good call! I will test this out in a bit.

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
JayShortway: Ouch, yea that's something we should improve. Can be a separate PR though.

gradle/libs.versions.toml
tonidero: I just checked to the same case that was causing issues in the device we tested on before. It's very easy to reproduce in 7.1.1, but doesn't have any issues in 8.0.0! So I removed all the code we added here: https://github.com/RevenueCat/purchases-android/pull/2477/commits/2f55a5deff661d60df788e0cb102e0292f728be5

JayShortway: OMG that is amazing! Thanks for checking! We can also update our docs when this is released.

purchases/src/main/kotlin/com/revenuecat/purchases/common/purchaseHistoryRecordExtensions.kt
vegaro: 4 years later...

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -330,47 +326,16 @@ internal class BillingWrapper(
...
-            onReceivePurchaseHistory,
-            onReceivePurchaseHistoryError,
-            ::withConnectedClient,
-            ::executeRequestOnUIThread,
-        ).run()
-    }
-
     override fun queryAllPurchases(
         appUserID: String,
         onReceivePurchaseHistory: (List<StoreTransaction>) -> Unit,
```
vegaro, line 331: Maybe these parameters names should be updated.

I wonder if queryAllPurchases can just be `queryPurchases` now as well

tonidero: Right, I left `queryAllPurchases` as it is... The thing is that Amazon can still query for all products, so I think restore/sync should still do that in Amazon. In that case, we need to differentiate between both... (I guess we could also unify both methods and add a parameter, but I kept it separate for now to minimize the changes and left a comment as seen [here](https://github.com/RevenueCat/purchases-android/pull/2477#discussion_r2178128842)

purchases/src/main/kotlin/com/revenuecat/purchases/common/purchaseHistoryRecordExtensions.kt
JayShortway: 😂

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -898,21 +862,17 @@ internal class BillingWrapper(
...
+                .setObfuscatedAccountId(appUserID.sha256())
+                .apply {
+                    isPersonalizedPrice?.let {
+                        setIsOfferPersonalized(it)
                     }
-                    .build(),
-            )
-        } catch (e: NoClassDefFoundError) {
-            throw NoCoreLibraryDesugaringException(e)
-        }
```
vegaro, line 915: did they finally fix this?

purchases/src/main/kotlin/com/revenuecat/purchases/google/errors.kt
```diff
@@ -12,6 +12,16 @@ internal fun @receiver:BillingClient.BillingResponseCode Int.getBillingResponseC
...
 
+@Suppress("MaximumLineLength", "MaxLineLength")
+internal fun @receiver:BillingClient.OnPurchasesUpdatedSubResponseCode Int.getOnPurchasesUpdatedSubResponseCodeName(): String {
+    val allPossibleOnPurchasesUpdatedSubResponseCodes =
+        BillingClient.OnPurchasesUpdatedSubResponseCode::class.java.declaredFields
+    return allPossibleOnPurchasesUpdatedSubResponseCodes
+        .firstOrNull { it.getInt(it) == this }
+        ?.name
+        ?: "$this"
+}
```
vegaro, line 35: I assume `getOnPurchasesUpdatedSubResponseCodeName` will barely get called but since it's using reflection, you could maybe create a static map that this function can access when called

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -64,19 +61,36 @@ internal class QueryProductDetailsUseCase(
...
         log(LogIntent.DEBUG) {
             OfferingStrings.FETCHING_PRODUCTS_FINISHED.format(useCaseParams.productIds.joinToString())
         }
         log(LogIntent.PURCHASE) {
-            OfferingStrings.RETRIEVED_PRODUCTS.format(received.joinToString { it.toString() })
+            OfferingStrings.RETRIEVED_PRODUCTS.format(received.productDetailsList.joinToString { it.toString() })
         }
-        logErrorIfIssueBuildingBillingParams(received)
-        received.takeUnless { it.isEmpty() }?.forEach {
+        received.unfetchedProductList.takeIf { it.isNotEmpty() }?.let {
```
vegaro, line 71: oh this `unfetchedProductList` is nice

We could have somthing similar in `Offerings`

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Yup! I was able to test on the device we used to test the issues. Very easy to repro in main. No issues with BC8 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
tonidero: Yeah, we might want to expose this somehow... But we should probably agree on what API to use. For now, I'm just logging the reason for it to be unfetched, but I also think it might be good to track it with diagnostics, and even give this information in the dashboard.

purchases/src/main/kotlin/com/revenuecat/purchases/google/errors.kt
tonidero: Well, it will get called on every purchase attempt... And yeah, might be better to create a map... I just followed what we do for the responseCode itself, but I think creating a map ourselves feels better yeah... Will do the change, and I might change that other mapping we do while I'm at it.

MarkVillacampa (approved):

tonidero:
Removing breaking label since we're making a new release without this for now, so automations are able to continue and make a normal release before we ship this. Reverted in #2501

PR #2483, opened 2025-07-03
Fixes building sample apps with SNAPSHOT dependencies

JayShortway:
## Description
The SNAPSHOT repo URL has changed, now that we migrated to Central Portal publishing. I also removed it from plugin repos, as we don't use SNAPSHOT plugins, and removed `gradlePluginPortal()` from depencency repos, as no dependencies are published there. 

This should fix the following failing CI jobs: 
- [assemble-magic-weather-compose-sample-app
](https://app.circleci.com/pipelines/github/RevenueCat/purchases-android/15853/workflows/c8bccc8d-259f-4eed-8b60-9128a79683ea/jobs/82702)
- [assemble-custom-entitlement-computation-sample-app
](https://app.circleci.com/pipelines/github/RevenueCat/purchases-android/15853/workflows/c8bccc8d-259f-4eed-8b60-9128a79683ea/jobs/82703)

### DIFF
--- a/examples/CustomEntitlementComputationSample/settings.gradle.kts
+++ b/examples/CustomEntitlementComputationSample/settings.gradle.kts
@@ -2,7 +2,6 @@ pluginManagement {
     repositories {
         google()
         mavenCentral()
-        maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
         gradlePluginPortal()
     }
 }
@@ -11,7 +10,9 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
-        maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
+        maven(url = "https://central.sonatype.com/repository/maven-snapshots/") {
+            content { includeGroup("com.revenuecat.purchases") }
+        }
     }
 }
 rootProject.name = "CustomEntitlementComputationSample"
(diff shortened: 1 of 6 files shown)


examples/CustomEntitlementComputationSample/settings.gradle.kts
```diff
@@ -11,7 +10,9 @@ dependencyResolutionManagement {
     repositories {
         google()
         mavenCentral()
-        maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
+        maven(url = "https://central.sonatype.com/repository/maven-snapshots/") {
```
tonidero, line 13: Hmm I do wonder if we need the snapshots repos in these apps... But I guess I'm ok leaving them in case we want to make use of one of our own snapshots... (I would probably point to mavenLocal in that case 😅 )

JayShortway: True, but our automations are currently setting the dependency versions to snapshots in these apps, e.g. see [libs.versions.toml](https://github.com/RevenueCat/purchases-android/blob/fix-snapshot-repo-url/examples/CustomEntitlementComputationSample/gradle/libs.versions.toml#L5) of the CustomEntitlementComputationSample.

tonidero: Ahh you're right 🤦 Then yeah, we need it. Thanks for letting me know!

PR #2485, opened 2025-07-04
Improves button progress indicator size calculation.

JayShortway:
## Description
No longer makes the progress indicator as big as the button itself. Dynamically calculates some padding, and sets a maximum size. 

With artificially long loading time:

https://github.com/user-attachments/assets/a68da256-02e2-4c90-8379-e49cde937237

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
@@ -23,6 +23,8 @@ import androidx.compose.ui.layout.Layout
 import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.coerceIn
 import androidx.compose.ui.unit.dp
 import com.revenuecat.purchases.paywalls.components.properties.CornerRadiuses
 import com.revenuecat.purchases.paywalls.components.properties.Dimension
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -165,6 +171,33 @@ internal fun ButtonComponentView(
...
+        a = stackWidthPx - stackMarginStartPx - stackMarginEndPx,
+        b = stackHeightPx - stackMarginTopPx - stackMarginBottomPx,
+    ).toDp()
+
+    val progressMarginDp = when {
+        minDimensionDp >= 32.dp -> 16.dp
+        minDimensionDp >= 24.dp -> minDimensionDp - 16.dp
+        minDimensionDp >= 16.dp -> 8.dp
+        minDimensionDp >= 8.dp -> minDimensionDp - 8.dp
+        else -> 0.dp
```
tonidero, line 193: hopefully devs won't have buttons as small, but better to handle it yeah 😅

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -165,6 +171,33 @@ internal fun ButtonComponentView(
...
+    }
+
+    return (minDimensionDp - progressMarginDp)
+        .coerceIn(0.dp, 48.dp)
```
tonidero: Hmm so I guess it will be 48.dp when the minDimension is 64.dp or more right? I wonder if we should have smaller higher end... a 48.dp loading spinner seems huge 😅

tonidero (approved):
Just one question, but not really a blocker, thanks for fixing this!! 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
JayShortway: Ha yes, maybe that's a bit much indeed. All of the Material 3 docs are _expressive_ now, which have different loading indicators, but it suggests `38.dp`. So I made it that. 

Seems to match the loading indicator on the Google Play purchase sheet quite closely:

https://github.com/user-attachments/assets/aaac90ff-3048-4b35-b9f2-eb924ad874bc

JayShortway: Done here: https://github.com/RevenueCat/purchases-android/pull/2485/commits/36078246a1ba31b89a73795c95829eb3e3358a80.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
JayShortway: Haha yea, maybe really small icon buttons? 😅

PR #2487, opened 2025-07-04
Resubscribe if cancelled

vegaro:
If the subscription is already cancelled, it should say "Resubscribe" instead of "Cancel Subscription"

It should also not open any feedback survey or promotional offer


CLAUDE.md
```diff
@@ -0,0 +1,188 @@
...
+# Run instrumentation tests for specific modules
+./gradlew :purchases:connectedDefaultsDebugAndroidTest
+./gradlew :ui:revenuecatui:connectedDefaultsDebugAndroidTest
+
+# Run integration tests (requires device/emulator)
+./gradlew :integration-tests:connectedDebugAndroidTest
+```
+
+### Code Quality and Analysis
+```bash
```
JayShortway:
```suggestion
```bash
# Run lint (static code analysis)
./gradlew lint
```

CLAUDE.md
```diff
@@ -0,0 +1,188 @@
...
+├── paywalls/               # Paywall data and components
+├── customercenter/         # Customer center functionality
+├── models/                 # Data models and DTOs
+├── interfaces/             # Public API contracts
+└── utils/                  # Utility functions
+```
+
+### Product Flavors
+- **`defaults`** - Standard Google Play implementation
+- **`customEntitlementComputation`** - Custom entitlement computation variant
```
JayShortway:
```suggestion
### Product Flavors
The `purchases` module has 2 flavors:
- **`defaults`** - Standard Google Play implementation
- **`customEntitlementComputation`** - Custom entitlement computation variant
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -368,10 +372,37 @@ internal class CustomerCenterViewModelImpl(
...
+            .let { filteredPaths ->
+                transformPathsOnSubscriptionState(filteredPaths, selectedPurchaseInformation, localization)
+            }
+    }
+
+    private fun transformPathsOnSubscriptionState(
+        paths: List<HelpPath>,
+        selectedPurchaseInformation: PurchaseInformation?,
+        localization: CustomerCenterConfigData.Localization,
+    ): List<HelpPath> {
```
JayShortway: Minor, but we can make `paths` the receiver, so we don't need the `let` above:
```kotlin
    private fun List<HelpPath>.transformPathsOnSubscriptionState(
        selectedPurchaseInformation: PurchaseInformation?,
    ): List<HelpPath> {
```

(You can press Option+Enter with your caret in the `paths` parameter, and click "Convert parameter to receiver" to do it automatically.)

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterViewModelTests.kt
```diff
@@ -122,7 +121,7 @@ class CustomerCenterViewModelTests {
     }
 
     @Test
-    fun `loadAndDisplayPromotionalOffer returns false when offer is not eligible`() = runTest {
+    fun `loadAndDisplayPromotionalOffer returns false when offer is not eligible`(): Unit = runBlocking {
```
JayShortway, line 124: 🎉

JayShortway (approved):
Nice, looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -368,10 +372,37 @@ internal class CustomerCenterViewModelImpl(
...
+        return paths.map { path ->
+            // For cancelled subscriptions, show "Resubscribe" instead of "Cancel"
+            if (path.type == CustomerCenterConfigData.HelpPath.PathType.CANCEL &&
```
tonidero: Just a question... does it matter if there is a cancellation survey in this case? (Just checking we verify the subscription is not already cancelled before displaying the survey, which wouldn't make sense 😅).

CLAUDE.md
```diff
@@ -0,0 +1,188 @@
...
+# Run backend integration tests
+bundle exec fastlane run_backend_integration_tests
+```
+
+## Project Architecture
+
+### Module Structure
+This is a multi-module Android project with clear separation of concerns:
+
+- **`:purchases`** - Core SDK module containing main API, business logic, networking, billing abstractions
```
tonidero: I wonder if we would like to add more info about specific packages in this module... But can always be added later, I think this makes sense!

vegaro: oh, I actually had another branch for this CLAUDE.md file, I wasn't intending to add it to this PR. I will extract it

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
vegaro: yeah good catch! that's actually why I am setting the feedbackSurvey and promotionalOffer to null in https://github.com/RevenueCat/purchases-android/pull/2487/files#diff-869560546e1bbd7f1c37b444e38d8537d0d734b488ae4a1c9465978324542aa3R399-R400

PR #2492, opened 2025-07-08
Remove logic to avoid repurchasing already subscribed products

tonidero:
### Description
As the title says. Now, we will rely on Google instead to handle resubscribing. The default behavior by Google is:
- If trying to purchase the same base plan that the user is already subscribed, display a message indicating so, with a link to manage subscriptions
- If trying to purchase a different base plan for a product the user is already subscribed to, perform an upgrade/downgrade.
- If trying to purchase a product the user has an active but cancelled subscription for, resubscribe.


ajpallares (approved):
Nice one! I love this kind of PRs with mostly red diffs 😄 Much simpler now!

PR #2493, opened 2025-07-08
Update to kotlin 2.0.21 while keeping language compatibility

tonidero:
### Description
This updates Kotlin in our project to 2.0.21 while keeping the language level and api level in our libraries to 1.8, to avoid causing breaking changes

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -31,7 +31,7 @@ fragment = "1.6.1"
 hamcrest = "1.3"
 recyclerview = "1.2.1"
 roboelectric = "4.11.1"
-kotlin = "1.8.22"
+kotlin = "2.0.21"
 # Can't update until we use more recent kotlin. 1.6.0 uses Kotlin 1.9.0
 kotlinxSerializationJSON = "1.5.1"
 kover = "0.7.0"
@@ -44,7 +44,7 @@ mockk = "1.13.11"
 mockwebserver = "4.2.0"
 multidex = "2.0.1"
 paparrazzi = "1.3.1"
-poko = "0.13.1"
+poko = "0.17.2"
 testJUnit = "1.1.3"
 testJUnitParams = "1.1.1"
 testLibraries = "1.4.0"
@@ -152,6 +152,7 @@ androidx-profileinstaller = { group = "androidx.profileinstaller", name = "profi
 android-application = { id = "com.android.application", version.ref = "agp" }
 android-library = { id = "com.android.library", version.ref = "agp" }
 androidx-navigation-safeargs = { id = "androidx.navigation.safeargs.kotlin", version.ref = "androidxNavigation" }
+compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
 dependencyGraph = { id = "com.savvasdalkitsis.module-dependency-graph", version.ref = "dependencyGraph" }
 detekt = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt" }
 dokka = { id ="org.jetbrains.dokka", version.ref = "dokka"}
--- a/library.gradle
+++ b/library.gradle
@@ -22,6 +22,8 @@ android {
     }
     kotlinOptions {
         jvmTarget = '1.8'
+        languageVersion = '1.8'
+        apiVersion = '1.8'
     }
 }
 
--- a/ui/debugview/src/main/kotlin/com/revenuecat/purchases/ui/debugview/models/TestModels.kt
+++ b/ui/debugview/src/main/kotlin/com/revenuecat/purchases/ui/debugview/models/TestModels.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.debugview.models
 
+import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.PackageType
@@ -15,6 +16,7 @@ import com.revenuecat.purchases.models.StoreProduct
 import com.revenuecat.purchases.models.SubscriptionOption
 import com.revenuecat.purchases.models.SubscriptionOptions
 
+@OptIn(InternalRevenueCatAPI::class)
 @Suppress("MagicNumber")
 internal val testOffering: Offering
     get() {
--- a/api-tester/build.gradle.kts
+++ b/api-tester/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
     alias(libs.plugins.android.application)
     alias(libs.plugins.kotlin.android)
+    alias(libs.plugins.compose.compiler)
 }
 
 apply(from = "$rootDir/base-application.gradle")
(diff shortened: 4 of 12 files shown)


gradle/libs.versions.toml
```diff
@@ -31,7 +31,7 @@ fragment = "1.6.1"
 hamcrest = "1.3"
 recyclerview = "1.2.1"
 roboelectric = "4.11.1"
-kotlin = "1.8.22"
+kotlin = "2.0.21"
 # Can't update until we use more recent kotlin. 1.6.0 uses Kotlin 1.9.0
 kotlinxSerializationJSON = "1.5.1"
```
vegaro, line 37: we'll be finally able to update this 💃

library.gradle
```diff
@@ -22,6 +22,8 @@ android {
     }
     kotlinOptions {
         jvmTarget = '1.8'
+        languageVersion = '1.8'
+        apiVersion = '1.8'
```
JayShortway: Maybe we can add these to `libs.versions.toml`?

ui/debugview/src/main/kotlin/com/revenuecat/purchases/ui/debugview/models/TestModels.kt
```diff
@@ -15,6 +16,7 @@ import com.revenuecat.purchases.models.StoreProduct
 import com.revenuecat.purchases.models.SubscriptionOption
 import com.revenuecat.purchases.models.SubscriptionOptions
 
+@OptIn(InternalRevenueCatAPI::class)
```
ajpallares: I'm curious, why is this needed now after the update to kotlin 2.0.21?

tonidero: Good callout! It's due to the Offering constructor we're using here and in the API tests. The constructor for `Offering` looks like:
```
data class Offering
@JvmOverloads
constructor(
    val identifier: String,
    val serverDescription: String,
    val metadata: Map<String, Any>,
    val availablePackages: List<Package>,
    val paywall: PaywallData? = null,
    @InternalRevenueCatAPI
    val paywallComponents: PaywallComponents? = null,
) { .... }
```

Note that the `paywallComponents` parameter is marked is of type `PaywallComponents`, which is marked experimental. Seems that now devs using this constructor now need to opt in as well since they are accessing that type, and the OptIn we have in the constructor only affects the type itself...

I originally thought it might be fine, but I do agree it's a small breaking change, since now you need to opt in to use the offering constructor... So I will fix it so at least you can use the constructor without `PaywallComponents` without having to OptIn. Thanks for the callout!

tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/2493/commits/d8d8be4f8d0365ff71f4d3f4d23bb7456ad91107

ajpallares: Makes sense. Thank you for the explanation!

gradle/libs.versions.toml
```diff
@@ -31,7 +31,8 @@ fragment = "1.6.1"
 hamcrest = "1.3"
 recyclerview = "1.2.1"
 roboelectric = "4.11.1"
-kotlin = "1.8.22"
+kotlin = "2.0.21"
+kotlinMinSupported = "1.8"
```
JayShortway: This is not really the minimum supported version. I would probably name this closer to what we use it for, and call it `kotlinLanguage` or something.

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
```diff
@@ -28,8 +28,25 @@ constructor(
     val availablePackages: List<Package>,
     val paywall: PaywallData? = null,
     @InternalRevenueCatAPI
-    val paywallComponents: PaywallComponents? = null,
+    val paywallComponents: PaywallComponents?,
```
JayShortway, line 31: Why did this change?

ui/revenuecatui/api.txt
```diff
@@ -66,7 +66,7 @@ package com.revenuecat.purchases.ui.revenuecatui {
     method public default void onRestoreStarted();
   }
 
-  public final class PaywallOptions {
+  @androidx.compose.runtime.Immutable public final class PaywallOptions {
```
JayShortway, line 69: I got these too. How's that possible? 🤔

JayShortway (approved):
Let's go! 🚀 Minor things only.

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
tonidero: Right, I did it to avoid a conflicting declaration between both constructors. In order to make it non conflicting, the other constructor will be used when no `paywallComponents` parameter is used. However, if one is passed (with a value or null), the main constructor will be used.

gradle/libs.versions.toml
tonidero: I like that!

ui/revenuecatui/api.txt
tonidero: I was indeed confused about this... I would have expected it to be annotated from before this PR, so no idea why it was added with this PR...

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
JayShortway: Ah got it, thanks!

ui/revenuecatui/api.txt
tonidero: Maybe we're not failing tests on changes on the UI library, only on the main SDK?

JayShortway: I'm thinking along those lines, but still not sure what's going on. I've tried this on `main`: 

```shell
./gradlew :ui:revenuecatui:metalavaCheckCompatibilityDefaultsRelease
```
☝️ checking doesn't fail

👇 generating shows new changes
```shell
./gradlew :ui:revenuecatui:metalavaGenerateSignatureDefaultsRelease
```

JayShortway: and then checking again (with the newly generated API dump), doesn't fail either..

tonidero: Just tried removing an entire class from the api.txt file and it doesn't fail either... So it doesn't seem to be working for some reason :/

skydoves: Hmm, that's truly interesting 🤔  Maybe its checking function doesn't detect annotations?

JayShortway: Maybe we should not use the checking task, and just generate new API dump files and using git to verify there's no diff? (All of this goes beyond the scope of this PR of course.)

tonidero: Yeah, I don't think it's due to the automations, since it doesn't seem to fail either after I modify the api file and remove an entire class... And yeah, generating new API files and checking git should also work, though a bit more cumbersome 😅

ajpallares (approved):
Very nice! 🚀

PR #2496, opened 2025-07-08
Fetch VCs from Network

fire-at-will:
### Motivation
This PR is a follow-up PR to https://github.com/RevenueCat/purchases-android/pull/2466 that supports fetching virtual currencies from the backend in the `VirtualCurrencyManager.getVirtualCurrencies()` function.

### Description
This PR:
- Fetches Virtual Currencies from the network in `VirtualCurrencyManager.getVirtualCurrencies()` when the VC cache is empty or stale
- Adds log messages for the virtual currencies features. The log messages match those on iOS and are logged at the same time.
- Makes the errors thrown by `VirtualCurrenciesFactory` more explicit
- Removes the unused `VirtualCurrencyFactory` class and its tests
- Adds buttons to the Purchase Tester app to support manually testing:
   - The `Purchases.getVirtualCurrencies` function
   - Fetching the cached virtual currencies from Purchases.cachedVirtualCurrencies
   - Invalidating the VC cache using Purchases.invalidateVirtualCurrenciesCache()

Here's a screenshot of the updated Purchase Tester screen:
Screenshot 2025-07-10 at 2 56 56 PM

### Testing
This PR also adds unit tests for:
- Ensuring the DeviceCache appropriately handles the new exceptions declared in `VirtualCurrenciesFactory`
- Ensuring that the backend VC request is executed properly
- Updates a few VirtualCurrencyManager tests to reflect the new behavior

I've also manually tested the following scenarios through the purchase tester app:
- `Purchases.shared.getVirtualCurrencies()` fetches VCs from the network when the cache is empty.
- `Purchases.shared.getVirtualCurrencies()` returns cached VCs when non-stale virtual currencies are cached
- `Purchases.shared.getVirtualCurrencies()` fetches VCs from the backend when the cached VCs are present on the device but are stale
-  `Purchases.shared.cachedVirtualCurrencies` returns `null` when the cache is empty.
- `Purchases.shared.cachedVirtualCurrencies` returns virtual currencies immediately after fetching virtual currencies from the network
-`Purchases.shared.cachedVirtualCurrencies` returns cached VCs even if they are stale (useful if the app is offline)

### Next Steps
In a future PR, we'll add backend integration tests.

To keep the virtual-currency-dev branch clean, this PR will be merged into a fetch-virtual-currencies-new-endpoint branch. We'll make future PRs go into the fetch-virtual-currencies-new-endpoint branch, and when the functionality of the new APIs is complete, we'll merge fetch-virtual-currencies-new-endpoint into virtual-currency-dev.


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -742,6 +750,72 @@ internal class Backend(
...
+                    endpoint,
+                    body = null,
+                    postFieldsToSign = null,
+                    backendHelper.authenticationHeaders,
+                    fallbackBaseURLs = appConfig.fallbackBaseURLs,
+                )
+            }
+
+            override fun onError(error: PurchasesError) {
+                synchronized(this@Backend) {
```
harrytmthy, line 776: Nit: consider logging the `PurchasesError` in `onError()` here for consistency with the explicit logging in the ViewModel and to ensure errors during VC fetch are visible in logs

fire-at-will: Hey @harrytmthy, thanks for the review! I've added logging in 8cdde1d691a3cdbba45961d0a2c5f54799ed2dde. Network errors are now logged in the `VirtualCurrencyManager` class 👍

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewViewModel.kt
```diff
@@ -103,9 +106,41 @@ class OverviewViewModel(private val interactionHandler: OverviewInteractionHandl
         interactionHandler.syncAttributes()
     }
 
+    fun onFetchVCsClicked() {
+        viewModelScope.launch {
+            val virtualCurrencies: VirtualCurrencies = Purchases.sharedInstance.awaitGetVirtualCurrencies()
```
tonidero, line 113: Are we planning on using this for something other than logs? Could be good to add a section to visualize the virtual currency to the purchase tester app I think.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -133,6 +137,10 @@ internal class Backend(
     @get:Synchronized @set:Synchronized
     @Volatile var redeemWebPurchaseCallbacks = mutableMapOf<String, MutableList<RedeemWebPurchaseCallback>>()
 
+    @get:Synchronized @set:Synchronized
+    @Volatile var getVirtualCurrenciesCallbacks =
```
tonidero: Maybe `virtualCurrenciesCallbacks`?

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
```diff
@@ -85,16 +100,46 @@ internal class VirtualCurrencyManager(
...
+    ) {
+        backend.getVirtualCurrencies(
+            appUserID = appUserID,
+            appInBackground = isAppBackgrounded,
+            onSuccess = { callback.onReceived(it) },
+            onError = { callback.onError(it) },
+        )
+    }
+
+    private fun fetchedVirtualCurrenciesFromBackendCallback(
```
tonidero: Hmm wdyt of something like `handleVirtualCurrenciesRequestResult`?

Then, we could rename `originalCallback` to `completion`.

purchases/src/test/java/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManagerTest.kt
```diff
@@ -91,6 +92,14 @@ class VirtualCurrencyManagerTest {
             mockAppConfig.isAppBackgrounded
         } returns false
 
+        val mockBackend = mockk<Backend>()
+        every {
+            mockBackend.getVirtualCurrencies(any(), any(), any(), any())
+        } answers {
+            val onSuccess = arg<(VirtualCurrencies) -> Unit>(2)
```
tonidero, line 101: Should we add some tests to also verify the behavior on an error?

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
```diff
@@ -85,16 +100,46 @@ internal class VirtualCurrencyManager(
...
+        override fun onReceived(virtualCurrencies: VirtualCurrencies) {
+            log(LogIntent.RC_SUCCESS) {
+                VirtualCurrencyStrings.VIRTUAL_CURRENCIES_UPDATED_FROM_NETWORK
+            }
+
+            cacheVirtualCurrencies(virtualCurrencies, appUserID)
+            originalCallback.onReceived(virtualCurrencies)
+        }
+        override fun onError(error: PurchasesError) {
+            log(LogIntent.RC_SUCCESS) {
```
tonidero: Hmm is this logIntent expected on an error?

tonidero (commented):
Looking good! Just some comments

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -742,6 +750,72 @@ internal class Backend(
...
+                            val virtualCurrencies = VirtualCurrenciesFactory.buildVirtualCurrencies(
+                                httpResult = result,
+                            )
+                            onSuccessHandler(virtualCurrencies)
+                        } catch (e: JSONException) {
+                            onError(e.toPurchasesError().also { errorLog(it) })
+                        } catch (e: SerializationException) {
+                            onError(e.toPurchasesError().also { errorLog(it) })
+                        } catch (e: IllegalArgumentException) {
+                            onError(e.toPurchasesError().also { errorLog(it) })
```
ajpallares: Important
```suggestion
                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
                        } catch (e: SerializationException) {
                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
                        } catch (e: IllegalArgumentException) {
                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
```

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
```diff
@@ -85,16 +100,46 @@ internal class VirtualCurrencyManager(
...
+                VirtualCurrencyStrings.VIRTUAL_CURRENCIES_UPDATED_FROM_NETWORK_ERROR.format(error)
+            }
+            originalCallback.onError(error)
```
ajpallares: Not saying we should, but I wonder: should we clear the cache at this point?
FWIW, following the code, this point would only be reached if the cache is stale

ajpallares (changes requested):
Looks good to me in general, but there's a small back in the onError callbacks of the get virtual currencies request. I'm requesting changes to make sure we don't miss that before merging. 
Great job btw! 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Ohh totally missed that, great catch!! Yeah we need to fix that before shipping this and possibly add some tests.

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
tonidero: No, I think we still want to allow devs to access stale data by using the `cachedVirtualCurrencies` synchronous accessor.

ajpallares: Makes sense. Thank you for the clarification!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
fire-at-will: Renamed in 078ca5d03ddb87febd7a01dc4fa639ddee7fc0e2 👍

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
fire-at-will: I love this - that naming is much clearer. Renamed in df8db163839392b69952a5f69b1a69448bcb485a 👍

purchases/src/test/java/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManagerTest.kt
fire-at-will: Great idea - I've added a test to ensure that the VCManager forwards the PurchasesError from the backend to the callback here: b1b0117f3defd39d76c55b96fe15d50dafa39d50 👍

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
fire-at-will: Definitely not - I should remove the copy/paste functionality from my computer 🤦‍♂️

Fixed in 1f1b8b25156fe0ab1e31a51cd523a0a9c7bcf9a8

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OverviewViewModel.kt
fire-at-will: @tonidero I've updated the purchase tester app to show the VCs on the overview screen here: b028b29b4e03923e13dd8ac301ce732a601d485e

Here's what it looks like:
Screenshot 2025-07-10 at 2 56 56 PM

fire-at-will:
@tonidero @ajpallares I've applied the fix that @ajpallares suggested - great find! We already had some tests ([here](https://github.com/RevenueCat/purchases-android/blob/633f3c38b2fbde47e81c15feafa639b9c102c0e3/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendTest.kt#L2645) that ensured that the callbacks were triggered when the JSON deserialization fails, but they were passing despite the bug because the function's callback was still being called 😅

I've tried to find a way to ensure that the `Backend.getVirtualCurrencies()` function calls the `Backend.virtualCurrenciesCallbacks` callbacks, but I've honestly haven't been able to figure out how to do that since it's private and there's no way to control what callbacks get added there from outside the function. I'm open to ideas if y'all have any!

tonidero (approved):
This looks great to me! :shipit:

ajpallares (approved):
:shipit:

PR #2498, opened 2025-07-09
Removes data classes from public API

JayShortway:
## Description
Codex started this, but I did most of the work sooooo. 

- Removes data classes from public API, and replaces with `@Poko`. 
- Manually wrote an internal `copy()` function if we were using it ourselves. 
- Adds detekt rule to guard against using data classes in the public API.

The diff is rather large, but most of it is very low in complexity.


gradle/libs.versions.toml
```diff
@@ -21,7 +21,7 @@ constraintlayout = "1.1.0"
 coroutines = "1.6.4"
 datastorePreference = "1.0.0"
 dependencyGraph = "0.9"
-detekt = "1.23.6"
+detekt = "1.23.8"
```
JayShortway, line 24: This is built against Kotlin 2.0.21. 

https://detekt.dev/changelog#1238---2025-02-20

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsRequest.kt
```diff
@@ -1,14 +1,16 @@
...
 
 /**
  * A request object for handling events.
  *
  * @property events A list of [BackendEvent] objects representing the events in the request.
  */
+@Poko
 @Serializable
-data class EventsRequest internal constructor(
+internal class EventsRequest internal constructor(
```
JayShortway, line 13: Note: made this `internal`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -66,6 +68,39 @@ data class PaywallOptions internal constructor(
...
+        return when {
+            this.offeringSelection != other.offeringSelection -> false
+            this.shouldDisplayDismissButton != other.shouldDisplayDismissButton -> false
+            this.fontProvider != other.fontProvider -> false
+            this.listener != other.listener -> false
+            this.purchaseLogic != other.purchaseLogic -> false
+            this.mode != other.mode -> false
+            else -> this.dismissRequest == other.dismissRequest
+        }
+    }
```
JayShortway, line 84: Detekt made me implement this, because we overrode `hashCode()` in https://github.com/RevenueCat/purchases-android/pull/2462. I'm not sure why it started failing now, as it happened even before I bumped the detekt version.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessor.kt
```diff
@@ -10,7 +10,7 @@ internal object VariableProcessor {
     /**
      * Information necessary for computing variables.
      */
-    data class PackageContext(
+    internal data class PackageContext(
```
JayShortway, line 13: Made this `internal` instead of removing `data`. The parent `VariableProcessor` is already `internal`, but detekt still failed.

config/detekt/detekt.yml
```diff
@@ -22,8 +22,20 @@ style:
 complexity:
   LongMethod:
     ignoreAnnotated: [ 'Preview' ]
+  LongParameterList:
+    ignoreAnnotated: [ 'Poko' ]
```
JayShortway, line 26: `LongParameterList` seems to automatically ignore data classes, so I made it ignore `@Poko`-annotated classes too.

config/detekt/detekt.yml
```diff
@@ -22,8 +22,20 @@ style:
...
   TooManyFunctions:
     ignoreAnnotatedFunctions: [ 'Preview' ]
+libraries:
+  active: true
+  ForbiddenPublicDataClass:
+    excludes: [ '**/examples/**' ]
+    ignoreAnnotated: [ 'InternalRevenueCatAPI' ]
+  LibraryEntitiesShouldNotBePublic:
+    active: false
```
JayShortway, line 35: This rule basically flags everything that is public. The idea is that the public classes are in a dedicated `api` or `public` package (or the non-public ones are in an `internal` package). You could then exclude that package from this rule. We don't have our code organized like that, so for now I disabled the rule.

ui/revenuecatui/api.txt
```diff
@@ -66,13 +58,8 @@ package com.revenuecat.purchases.ui.revenuecatui {
     method public default void onRestoreStarted();
   }
 
-  public final class PaywallOptions {
+  @androidx.compose.runtime.Immutable @dev.drewhamilton.poko.Poko public final class PaywallOptions {
```
JayShortway, line 61: Not sure why some of the `@Immutable` ones weren't picked up earlier.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
tonidero: Hmm maybe it didn't require you to implement it even if overriding `hashCode` since it was a data class before, that already had an equals implementation?

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/Offering.kt
```diff
@@ -0,0 +1,22 @@
+package com.revenuecat.purchases.ui.revenuecatui.extensions
+
+import com.revenuecat.purchases.Offering
+import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.paywalls.PaywallData
+
+@Suppress("LongParameterList")
+internal fun Offering.copy(
```
tonidero, line 8: Hmm I see we have some copy methods in the `purchases` module, with a `@InternalRevenueCat` annotation... Would it be worth doing the same for these ones? Not sure if they might be useful there at some point... In any case, not a blocker, I'm also ok with this.

tonidero (approved):
Amazing getting rid of all these!!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
JayShortway: Ah yes, that must be it!

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/Offering.kt
JayShortway: I'm a bit in the middle on this haha. It can't really hurt, but on the other hand it would add it to our "public" API (makes it more visible in auto complete and stuff, even with required opt-in). We can also move it when we end up needing it?

tonidero: Yup! I'm fine with that 👍

PR #2505, opened 2025-07-09
Add VC integration tests

fire-at-will:
### Description
Introduces virtual currency integration tests to the VC branch. These tests test the SDK's ability to fetch virtual currencies from the backend, along with the VC caching capabilities.

Right now, this PR is going into the `vc-network-call` branch, but once https://github.com/RevenueCat/purchases-android/pull/2496 is approved & merged, we will change this PR's base branch to `fetch-virtual-currencies-new-endpoint`.


purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionVirtualCurrenciesIntegrationTest.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private fun validateVirtualCurrenciesObject(
+        virtualCurrencies: VirtualCurrencies,
+        testVCBalance: Int,
+        testVC2Balance: Int,
+        testVC3Balance: Int = 0,
+    ) {
+        assert(virtualCurrencies.all.count() == 3)
+
+        val testCurrency = virtualCurrencies["TEST"]
```
tonidero: Nitpick but I wonder if we should create an entire expected VirtualCurrency object and compare it against each of these... That way we would notice if new fields are added...

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
```diff
@@ -205,6 +208,170 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
...
+                    },
+                )
+            },
+        )
+
+        lock.await(testTimeout.inWholeSeconds, TimeUnit.SECONDS)
+        assertThat(lock.count).isZero
+    }
+
+    private fun validateAllZeroBalanceVirtualCurrenciesObject(virtualCurrencies: VirtualCurrencies?) {
```
tonidero: Nitpick but I think naming can be simplified like this considering the context? 
```suggestion
    private fun validateAllZeroBalances(virtualCurrencies: VirtualCurrencies?) {
```
Same for the method above

tonidero (approved):
This looks good to me! Just a couple of nitpicks

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
fire-at-will: Great idea! Modified the names here 3ec729ca08d11052896c49f2665cd41cc76cdc9f 👍

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionVirtualCurrenciesIntegrationTest.kt
fire-at-will: Love this! Implemented this here: 1f255d51e7e932432083d0598efe5874bde2679d

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
```diff
@@ -205,6 +209,172 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
...
+    ) {
+        assert(virtualCurrencies!!.all.count() == 3)
+
+        val expectedTestVirtualCurrency = VirtualCurrency(
+            code = "TEST",
+            name = "Test Currency",
+            balance = testVCBalance,
+            serverDescription = "This is a test currency",
+        )
+        assert(virtualCurrencies["TEST"] == expectedTestVirtualCurrency)
```
tonidero: Another nitpick. I believe this works! But we've been using `AssertJ` assertions for these since they are a bit more readable I think... NABD though!
```suggestion
        assertThat(virtualCurrencies["TEST"]).isEqualTo(expectedTestVirtualCurrency)
```

tonidero (approved):
Looking good!

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
fire-at-will: Migrated to `assertThat` across the tests here! c8035cadefe0acfcd5b9f54dbeed9a09c93ccef2

PR #2507, opened 2025-07-10
Marks `PaywallData` and `PaywallColor` as `InternalRevenueCatAPI`.

JayShortway:
## Description
Since these are network models, they were never intended to be part of the public API. This makes `PaywallData` equivalent to its `PaywallComponents` successor. 

**Note:** this PR targets the Kotlin 2.0.21 branch.


ajpallares (approved):
Nice catch! Great opportunity for a public API cleanup 👍

tonidero (approved):
Nice!! Yeah, nice to do this now 💪

PR #2508, opened 2025-07-10
Add V9 migration guide

tonidero:
### Description
This adds a small migration guide for SDK V9

### DIFF
--- a/migrations/v9-MIGRATION.md
+++ b/migrations/v9-MIGRATION.md
@@ -1,0 +1,23 @@
+# V89 API Migration Guide
+
+This latest release updates the SDK to use BillingClient 8. This version removed APIs to query for expired subscriptions and consumed one-time products, aside from other improvements. You can check the full list of changes here: https://developer.android.com/google/play/billing/release-notes#8-0-0
+
+Additionally, we've also updated Kotlin to 2.0.21 but kept language compatibility which means we now require Kotlin 1.8.0+. If you were using an older version of Kotlin, you will need to update it.
+
+Regarding API changes, we've also removed data classes from our public APIs. This means that for classes that were previously data classes, the `copy` constructor and `componentN` functions have been removed (unpacking). Equals/HashCode functions should still work as before.
+
+### BC8: No expired subscriptions or consumed one-time products
+
+With BC8, we can't query for expired subscriptions or consumed one-time products. This means that, for users migrating from a different implementation of the Play Billing Library, the SDK will not be able to send purchase information from these purchases. We can still ingest historical data from these purchases through a backend historical import. See [docs](https://www.revenuecat.com/docs/migrating-to-revenuecat/migrating-existing-subscriptions). This doesn't affect developers that have started using RevenueCat since day one.
+
+### Bumped minimum Kotlin version
+
+RevenueCat SDK v9 bumps Kotlin to 2.0.21, with a minimum Kotlin version of 1.8.0.
+
+### Using the SDK with your own IAP code (previously Observer Mode)
+
+Using the SDK with your own IAP code is still supported in v9. Other than updating the SDK version, there are no changes required. Just make sure the version of the billing client is also version 8.0.0+.
+
+### Reporting undocumented issues:
+
+Feel free to file an issue! [New RevenueCat Issue](https://github.com/RevenueCat/purchases-android/issues/new/).


migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,23 @@
+# V89 API Migration Guide
```
ajpallares: Ups
```suggestion
# V9 API Migration Guide
```

migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+This latest release updates the SDK to use BillingClient 8. This version removed APIs to query for expired subscriptions and consumed one-time products, aside from other improvements. You can check the full list of changes here: https://developer.android.com/google/play/billing/release-notes#8-0-0
```
JayShortway: `BillingClient` is not the official name anymore afaik, just the class name haha.

```suggestion
This latest release updates the SDK to use Google Play Billing Library 8. This version removed APIs to query for expired subscriptions and consumed one-time products, aside from other improvements. You can check the full list of changes here: https://developer.android.com/google/play/billing/release-notes#8-0-0
```

ajpallares: > This latest release updates

I would change to

> This release updates

or to 

> The V9 release updates

Now it's the latest, but eventually it won't be

ajpallares: And also, to make it extremely clear that it's not our SDK changing this, I would change
> This version removed APIs

to 

> This version of BillingClient removed APIs

migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+Additionally, we've also updated Kotlin to 2.0.21 but kept language compatibility which means we now require Kotlin 1.8.0+. If you were using an older version of Kotlin, you will need to update it.
```
JayShortway: Not entirely sure of this 1 to 1 relationship, so I'd say let's just state the facts we know are true haha 😅 

```suggestion
Additionally, we've also updated Kotlin to 2.0.21, and our new minimum version is Kotlin 1.8.0+. If you were using an older version of Kotlin, you will need to update it.
```

migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+Regarding API changes, we've also removed data classes from our public APIs. This means that for classes that were previously data classes, the `copy` constructor and `componentN` functions have been removed (unpacking). Equals/HashCode functions should still work as before.
```
JayShortway:
```suggestion
Regarding API changes, we've also removed data classes from our public APIs. This means that for classes that were previously data classes, the `copy` function and `componentN` functions (destructuring declarations) have been removed. `equals` and `hashCode` functions still work as before.
```

ajpallares: > the `copy` constructor and `componentN` functions

Did you mean this?

> the `copy` constructor and `component` functions

ajpallares (approved):
Thanks for updating this! Looks good, only some small corrections

migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+### BC8: No expired subscriptions or consumed one-time products
+
+With BC8, we can't query for expired subscriptions or consumed one-time products. This means that, for users migrating from a different implementation of the Play Billing Library, the SDK will not be able to send purchase information from these purchases. We can still ingest historical data from these purchases through a backend historical import. See [docs](https://www.revenuecat.com/docs/migrating-to-revenuecat/migrating-existing-subscriptions). This doesn't affect developers that have started using RevenueCat since day one.
```
JayShortway: Should this be part of the migration guide? In other words, is this relevant for people migrating from v8 to v9? My understanding is that this is only relevant for people integrating RevenueCat for the first time.

migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Bumped minimum Kotlin version
+
+RevenueCat SDK v9 bumps Kotlin to 2.0.21, with a minimum Kotlin version of 1.8.0.
+
+### Using the SDK with your own IAP code (previously Observer Mode)
+
+Using the SDK with your own IAP code is still supported in v9. Other than updating the SDK version, there are no changes required. Just make sure the version of the billing client is also version 8.0.0+.
```
JayShortway:
```suggestion
Using the SDK with your own IAP code is still supported in v9. Other than updating the SDK version, there are no changes required. Just make sure the version of the Play Billing Library is also version 8.0.0+.
```

migrations/v9-MIGRATION.md
tonidero: I was actually saying `componentN`, for the "unpacking" functions data classes generate... But it may be confusing yeah, I will simplify!

JayShortway: I'd vote to keep calling it `componentN`. JetBrains does the same in their [docs](https://kotlinlang.org/docs/destructuring-declarations.html): 
>The `componentN()` functions need to be marked with the operator keyword to allow using them in a destructuring declaration.

@ajpallares for context, it generates a `component1()`, `component2()` etc. function for every parameter of a data class. That's where the `N` comes from.

ajpallares: Ohh then it was my lack of knowledge! If it's clear for Android devs, I don't think you should change it

migrations/v9-MIGRATION.md
tonidero: Could be for people that have recently migrated to RC I think 🤷

migrations/v9-MIGRATION.md
tonidero: Applied @JayShortway's comment below :)

migrations/v9-MIGRATION.md
JayShortway: Oh true, in that case let me add a suggestion haha.

JayShortway: Tried to make it a bit more obvious that this doesn't apply to most people. Lmk what you think! (Also correct anything that isn't correct haha.)

```suggestion
### Play Billing Library 8: No expired subscriptions or consumed one-time products

**Note:** the following is only relevant if you recently integrated RevenueCat, and do not (yet) have all your transactions imported. 
  
Play Billing Library 8 removed functionality to query expired subscriptions or consumed one-time products. This means that, for users migrating from a non-RevenueCat implementation of the Play Billing Library, the SDK will not be able to send purchase information from these purchases. We can still ingest historical data from these purchases through a backend historical import. See [docs](https://www.revenuecat.com/docs/migrating-to-revenuecat/migrating-existing-subscriptions). This doesn't affect developers that have all transactions in RevenueCat, which is true for the vast majority.
```

tonidero: I like that 🙌 Thanks for the suggestion!

migrations/v9-MIGRATION.md
```diff
@@ -0,0 +1,25 @@
...
+
+This release updates the SDK to use Google Play Billing Library 8. This version of the Billing Library removed APIs to query for expired subscriptions and consumed one-time products, aside from other improvements. You can check the full list of changes here: https://developer.android.com/google/play/billing/release-notes#8-0-0
+
+Additionally, we've also updated Kotlin to 2.0.21 and our new minimum version is Kotlin 1.8.0+. If you were using an older version of Kotlin, you will need to update it.
+
+Regarding API changes, we've also removed data classes from our public APIs. This means that for classes that were previously data classes, the `copy` function and `componentN` functions (destructuring declarations) have been removed. `equals` and `hashCode` functions still work as before.
+
+### Play Billing Library 8: No expired subscriptions or consumed one-time products
+
+**Note:** the following is only relevant if you recently integrated RevenueCat, and do not (yet) have all your transactions imported.
```
aboedo, line 11: I'm late to the game here, but what is "recently" in this context? this migration guide will live for a long time. Maybe we can tie it to a real-life date, even if arbitrary?

JayShortway: Hmm right, "recently" is in relation to the act of upgrading to v9. So if you're upgrading to v9 soon after integrating RevenueCat for the first time, and not all transactions are imported (yet). Maybe something like this is already better? 

```suggestion
**Note:** the following is only relevant if you recently integrated RevenueCat before upgrading to v9, and do not (yet) have all your transactions imported. 
```

tonidero: Right, I don't think we can tie it to a specific date unfortunately, though it will become less relevant over time. I think @JayShortway's suggestion makes sense! Will make a PR with it.

PR #2509, opened 2025-07-10
Fix `Switch` component previews

tonidero:
### Description
This fixes the previews for the `Switch` component so we can preview the checked state. They weren't working since they tried to use an unused property `defaultValue`, which is removed in this PR. The component uses the selected tab index to know whether the switch should be checked or not.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PreviewHelpers.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PreviewHelpers.kt
@@ -69,6 +69,7 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.PaywallResourceProvider
 import com.revenuecat.purchases.ui.revenuecatui.helpers.PaywallValidationResult
 import com.revenuecat.purchases.ui.revenuecatui.helpers.Result
 import com.revenuecat.purchases.ui.revenuecatui.helpers.getOrThrow
+import com.revenuecat.purchases.ui.revenuecatui.helpers.map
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PreviewHelpers.kt
```diff
@@ -111,7 +112,13 @@ internal fun previewEmptyState(): PaywallState.Loaded.Components {
             data = data,
         ),
     )
-    val validated = offering.validatePaywallComponentsDataOrNullForPreviews()?.getOrThrow()!!
+    val validated = offering.validatePaywallComponentsDataOrNullForPreviews()?.map {
+        if (initialSelectedTabIndex != null) {
+            it.copy(initialSelectedTabIndex = initialSelectedTabIndex)
```
JayShortway, line 117: Do we still have a `copy` function after https://github.com/RevenueCat/purchases-android/pull/2498? (I am merging that one when CI is done.) Maybe you can target that branch to see if it still works haha.

JayShortway (approved):
Thanks for taking care of this! Just 1 question.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PreviewHelpers.kt
tonidero: We do! Note that this is from the `PaywallValidationResult.Components` which is an internal data class, so we didn't remove the data class.

PR #2510, opened 2025-07-10
Update CustomEntitlementComputation sample app kotlin version

tonidero:
### Description
This updates the CustomEntitlementComputation sample app to kotlin 1.8.22. This is needed after dropping support for 1.7 in #2493


examples/CustomEntitlementComputationSample/gradle/libs.versions.toml
```diff
@@ -1,9 +1,7 @@
 [versions]
 agp = "8.1.3"
-androidxNavigation = "2.5.3"
```
tonidero, line 3: All these were unused

JayShortway (approved):
Thanks for taking care of this!

PR #2514, opened 2025-07-11
Improve v9 migration docs

tonidero:
### Description
As suggested in https://github.com/RevenueCat/purchases-android/pull/2508#discussion_r2199911803


ajpallares (approved):
I think it makes sense 👍

JayShortway (approved):

PR #2519, opened 2025-07-11
Virtual Currency Support

fire-at-will:
### Description
Android counterpart to https://github.com/RevenueCat/purchases-ios/pull/5108.

This PR introduces virtual currencies to the Android SDK. It's a large PR, but all parts of it have been reviewed individually before here:
- https://github.com/RevenueCat/purchases-android/pull/2466
- https://github.com/RevenueCat/purchases-android/pull/2496
- https://github.com/RevenueCat/purchases-android/pull/2505

It includes the following:

#### New Top-Level Functions
These functions allow you to fetch the virtual currencies for the current subscriber. The `invalidateVirtualCurrenciesCache` function allows developers to clear the cached virtual currencies, which will then force the refreshing of the virtual currencies from the backend the next time `getVirtualCurrencies()` is called.

```kotlin
class Purchases {
   fun getVirtualCurrencies(
        callback: GetVirtualCurrenciesCallback,
    )

   fun invalidateVirtualCurrenciesCache()

   val cachedVirtualCurrencies: VirtualCurrencies?
}
```

#### New Objects
```kotlin
class VirtualCurrencies internal constructor(
    val all: Map<String, VirtualCurrency>,
) : Parcelable {
    operator fun get(code: String): VirtualCurrency? = all[code]
}

class VirtualCurrency internal constructor(
    val balance: Int,
    val name: String,
    val code: String,
    val serverDescription: String? = null,
)
```

#### Example Usage
Kotlin:
```swift
Purchases.sharedInstance.invalidateVirtualCurrenciesCache()

Purchases.sharedInstance.getVirtualCurrencies(
    object: GetVirtualCurrenciesCallback {
        override fun onReceived(virtualCurrencies: VirtualCurrencies) {}
        override fun onError(error: PurchasesError) {}
    }
)

Purchases.sharedInstance.getVirtualCurrenciesWith(
        onError = { _: PurchasesError -> },
        onSuccess = { _: VirtualCurrencies -> },
)

val getVirtualCurrenciesResult: VirtualCurrencies = Purchases.sharedInstance.awaitGetVirtualCurrencies()

val cachedVirtualCurrencies = Purchases.sharedInstance.cachedVirtualCurrencies
```

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/utils/Responses.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/utils/Responses.kt
@@ -322,31 +322,6 @@ object Responses {
           "virtual_currencies": {}
         }
     """.removeJSONFormatting()
-
-    val validVirtualCurrencyResponseWithDescription = """
-        {
-          "balance": 1,
-          "code": "COIN",
-          "description": "It's a coin",
-          "name": "Coin"
-        }
-    """.removeJSONFormatting()
-
-    val validVirtualCurrencyResponseWithoutDescription = """
-        {
-          "balance": 0,
-          "code": "RC_COIN",
-          "name": "RC Coin"
-        }
-    """.removeJSONFormatting()
-
-    val validVirtualCurrencyResponseWithNegativeBalance = """
-        {
-          "balance": -1,
-          "code": "NEGATIVE",
-          "name": "Negative"
-        }
-    """.removeJSONFormatting()
 }
 
 private fun String.removeJSONFormatting(): String = JSONObject(this).toString()


purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -227,6 +242,113 @@ internal open class DeviceCache(
...
+    fun setVirtualCurrenciesCacheTimestamp(appUserID: String, date: Date) {
+        preferences.edit().putLong(virtualCurrenciesLastUpdatedCacheKey(appUserID), date.time).apply()
+    }
+
+    @Synchronized
+    private fun getVirtualCurrenciesCacheLastUpdated(appUserID: String): Date {
+        return Date(preferences.getLong(virtualCurrenciesLastUpdatedCacheKey(appUserID), 0))
+    }
+
+    private fun SharedPreferences.Editor.clearVirtualCurrenciesCacheTimestamp(
```
harrytmthy, line 325: Nit: Since `clearVirtualCurrenciesCache` and `clearVirtualCurrenciesCacheTimestamp` are always called together, consider merging them into a single function like:

```kotlin
fun SharedPreferences.Editor.clearVirtualCurrenciesCacheAndTimestamp(appUserID: String): SharedPreferences.Editor
```

This would reduce call-site verbosity, clarify intent, and prevent potential future inconsistencies where one might be called without the other.

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencies.kt
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * This class contains all the virtual currencies associated to the user.
+ *
+ * @property all - All of the virtual currencies associated to the user.
+ */
+@Poko
+@Parcelize
+@Serializable
+class VirtualCurrencies internal constructor(
```
tonidero, line 17: I'm wondering if we want to make these constructors public so it's easier to test them... NABD though, we can probably make it public later if needed.

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
```diff
@@ -0,0 +1,145 @@
...
+
+@Suppress("UnusedPrivateProperty")
+internal class VirtualCurrencyManager(
+    private val identityManager: IdentityManager,
+    private val deviceCache: DeviceCache,
+    private val backend: Backend,
+    private val appConfig: AppConfig,
+) {
+
+    // region Public functions
```
tonidero: Is this a leftover? Since I don't see a `// endregion Public functions`. Probably not needed in this case I think...

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -95,6 +108,8 @@ internal open class DeviceCache(
             .clearCustomerInfo()
             .clearAppUserID()
             .clearCustomerInfoCacheTimestamp(appUserID)
+            .clearVirtualCurrenciesCacheTimestamp(appUserID)
```
tonidero, line 111: I think this is not needed since the timestamp is already cleared as part of `clearVirtualCurrenciesCache`?

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
tonidero: I think it might be good to keep as a separate private method to avoid bloating the `clearVirtualCurrenciesCache` method but I do agree we should probably only use `clearVirtualCurrenciesCache` generally. 

There may be cases where we want to clear the timestamp and not the data, for example, if we want to force a stale cache. However, we're not doing that right now, so I do think this makes sense (we can always modify in the future if needed).

tonidero (approved):
Looking good! Just some small comments but nothing blocking!

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencies.kt
fire-at-will: Yeah, we can make it public later if we'd like! I made it internal because I was following the pattern we used for `WebPurchaseRedemption`, which uses an internal constructor.

purchases/src/main/kotlin/com/revenuecat/purchases/virtualcurrencies/VirtualCurrencyManager.kt
fire-at-will: Removed in  ae12c0b8c3c94e601c91f8027aa21356eec5d27b 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
fire-at-will: I think this is still needed. It's a little confusing because there are a few `clearVirtualCurrenciesCache` functions defined in this class 😅
- Two defined on line 295 that `DeviceCache` consumers call. This one does clear the VC cache timestamp
- One defined on line 339: `private fun SharedPreferences.Editor.clearVirtualCurrenciesCache(appUserID: String): SharedPreferences.Editor`. This one does not clear the VC cache timestamp.

The `clearVirtualCurrenciesCacheTimestamp` function called here on line 111 calls the `private fun SharedPreferences.Editor.clearVirtualCurrenciesCache(appUserID: String): SharedPreferences.Editor` function, which doesn't clear the timestamp.

Let me know if I might be missing something!

tonidero: Ohhh right! sorry I got confused. I think that makes sense!

tonidero (approved):
🚢

PR #2520, opened 2025-07-14
Make PaywallActivity not consume insets in Android 16+

tonidero:
### Description
This makes it so we don't consume insets in paywalls in Android 16+. Starting on Android 16+, by default apps will go edge to edge. We already consume insets correctly within the paywall as needed, but the activity was actually consuming those insets before that. 

A better fix might be to add an option to enable/disable this consumption, while keeping the current behavior in Android < 16 but changing it for android 16+. This is a short term fix for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,8 @@ aliases:
 
   android-executor: &android-executor
     executor:
-      name: android/android-docker
-      resource-class: xlarge
+      name: android/android_docker
+      resource_class: xlarge
       tag: 2024.04.1
     environment:
       JVM_OPTS: -Xmx6g
@@ -34,13 +34,13 @@ aliases:
 
   android-machine-emulator: &android-machine-emulator
     executor:
-      name: android/android-machine
-      resource-class: xlarge
-      tag: 2024.04.1
+      name: android/android_machine
+      resource_class: xlarge
+      tag: 2024.11.1
 
 version: 2.1
 orbs:
-  android: circleci/android@2.5.0
+  android: circleci/android@3.1.0
   gcp-cli: circleci/gcp-cli@3.3.1
   revenuecat: revenuecat/sdks-common-config@3.0.0
   codecov: codecov/codecov@3.2.4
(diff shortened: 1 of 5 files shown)


JayShortway (approved):
Nice and pragmatic!

.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ aliases:
     executor:
       name: android/android_docker
       resource_class: xlarge
-      tag: 2024.04.1
+      tag: 2025.04.1
```
tonidero, line 30: I missed updating the tag in #2521

build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    extra["compileVersion"] = 34
+    extra["compileVersion"] = 35
```
tonidero: I had this locally and it's actually needed for this to compile... I extracted this change to a separate PR: #2525

gradle/libs.versions.toml
```diff
@@ -30,7 +30,7 @@ espresso = "3.4.0"
 fragment = "1.6.1"
 hamcrest = "1.3"
 recyclerview = "1.2.1"
-roboelectric = "4.11.1"
+roboelectric = "4.15.1"
```
tonidero: Bumped robolectric to support API 35, also part of #2525

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/BlurTransformation.kt
```diff
@@ -56,15 +57,15 @@ internal fun Bitmap.blur(context: Context, radius: Float, scaleDown: Boolean = t
     script.setInput(input)
     script.forEach(output)
 
-    val blurredBitmap = Bitmap.createBitmap(bitmap.width, bitmap.height, config)
+    val blurredBitmap = config?.let { createBitmap(bitmap.width, bitmap.height, it) }
```
tonidero: This is also part of #2525

PR #2521, opened 2025-07-14
Update CircleCI android orb

tonidero:
### Description
This updates the android orb and the tag we use for our jobs. Which will allow us to have access to the latest android SDKs and fix tests in #2520

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,8 @@ aliases:
 
   android-executor: &android-executor
     executor:
-      name: android/android-docker
-      resource-class: xlarge
+      name: android/android_docker
+      resource_class: xlarge
       tag: 2024.04.1
     environment:
       JVM_OPTS: -Xmx6g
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -73,11 +73,11 @@ commands:
 
   android-dependencies:
     steps:
-      - android/restore-gradle-cache
+      - android/restore_gradle_cache
```
JayShortway, line 76: Why does GitHub show all these spurious changed lines? Some encoding difference? 🤔

.circleci/config.yml
```diff
@@ -25,22 +25,22 @@ aliases:
...
-      resource-class: xlarge
-      tag: 2024.04.1
+      name: android/android_machine
+      resource_class: xlarge
+      tag: 2024.11.1
 
 version: 2.1
 orbs:
-  android: circleci/android@2.5.0
+  android: circleci/android@3.1.0
```
JayShortway, line 43: We could do this with Renovate too, but didn't implement it yet as I wasn't sure if it was desired. What do you think?

JayShortway (approved):
Looks good!

.circleci/config.yml
tonidero: These were actually changes in the orb... Not sure why they changed from `-` to `_` in all the orb methods... I *think* it might be the new convention for orb methods.

.circleci/config.yml
tonidero: I think for CircleCI stuff it does make sense! Also low risk changes I think

.circleci/config.yml
JayShortway: Ohh, I didn't notice the underscores haha. That's odd but okay 😅

.circleci/config.yml
JayShortway: Alright, we'll keep it in mind to update at some point.

PR #2523, opened 2025-07-15
Explicit api

JayShortway:
## Description
Draft, because I just wanna check the diff 😬


library.gradle
```diff
@@ -24,6 +24,7 @@ android {
         jvmTarget = '1.8'
         languageVersion = libs.versions.kotlinLanguage.get()
         apiVersion = libs.versions.kotlinLanguage.get()
+        freeCompilerArgs += '-Xexplicit-api=strict'
```
skydoves, line 27: What do you think about starting with [warning mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors) instead? This way, we can avoid compilation errors while still getting warnings to guide the migration. Once everything is addressed, we can switch back to strict mode for full enforcement 😄

JayShortway: Yes, I think that should be the path forward indeed! 
1. Enable warning mode.
2. Open various PRs addressing the warnings.
3. Enable strict mode.

This PR was me charting everything that needed to change haha, using Claude Code. I don't expect to merge this PR directly.

PR #2525, opened 2025-07-15
Increase compile/target SDK version to 35

tonidero:
### Description
Increases compile/target SDK version to 35

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 buildscript {
-    extra["compileVersion"] = 34
+    extra["compileVersion"] = 35
     extra["minVersion"] = 21
 }
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -30,7 +30,7 @@ espresso = "3.4.0"
 fragment = "1.6.1"
 hamcrest = "1.3"
 recyclerview = "1.2.1"
-roboelectric = "4.11.1"
+roboelectric = "4.15.1"
 kotlin = "2.0.21"
 kotlinLanguage = "1.8"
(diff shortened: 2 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/BlurTransformation.kt
```diff
@@ -56,15 +57,15 @@ internal fun Bitmap.blur(context: Context, radius: Float, scaleDown: Boolean = t
     script.setInput(input)
     script.forEach(output)
 
-    val blurredBitmap = Bitmap.createBitmap(bitmap.width, bitmap.height, config)
+    val blurredBitmap = config?.let { createBitmap(bitmap.width, bitmap.height, it) }
```
tonidero, line 60: Looks like it doesn't accept a nullable value anymore... In this case, I'm making it so, if `config` is null, we return the original bitmap without the blurring. I don't fully understand in what situations this might be null though in my tests, it seems to work fine.

JayShortway: From [cs.android.com](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/graphics/java/android/graphics/Bitmap.java;l=1897-1907?q=Bitmap): 
>If the bitmap's internal config is in one of the public formats, return that config, otherwise return null.

I guess that is usually non-null indeed. 🤷

JayShortway (approved):

PR #2527, opened 2025-07-15
Introduces `CompatComposeView` to handle scenarios where the view tree is not set up

JayShortway:
## Description
This opens the door to showing `PaywallView` in a new Window, instead of in an Activity. It should also fix a few crashes we've been seeing in react-native-purchases, when the Paywall is shown modally. 

There's an explainer added to the documentation of `CompatComposeView`.


vegaro:
just tested this as an alternative fix to https://github.com/RevenueCat/react-native-purchases/pull/1319 and it seems to work fine

JayShortway:
@vegaro @tonidero I tested the regular scenario of adding a PaywallView to an Activity (which already worked before), and that still works. Even when rotating the device. If you can think of other scenarios to test that would be great.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
```diff
@@ -0,0 +1,183 @@
...
+ * modern parents such as ComponentActivity and androidx Fragment. But this is not always the case, such as when the
+ * ComposeView is added to a plain Window, or in certain hybrid frameworks. A [CompatComposeView] can handle this
+ * scenario by acting as its own LifecycleOwner, SavedStateRegistryOwner and ViewModelStoreOwner if required.
+ *
+ * Note that, in this scenario, this does imply that the ComposeView's lifecycle is tied to its visibility. It ends
+ * when it is removed from the view tree. There is no concept of keeping state on a back stack. The ComposeView acts as
+ * a modal in this case.
+ */
+@Suppress("TooManyFunctions")
+@InternalRevenueCatAPI
```
tonidero, line 38: Hmm I'm wondering if we should make this `CompatComposeView` a child of `PaywallView` and the others instead of making it a subclass... That way we wouldn't need to expose it. Having said I'm also ok leaving it like this! probably cleaner, even though it forces us to expose this API.

JayShortway: Do you mean an inner class? I'm not sure if that would work either, because `PaywallView` would still need to extend it, which makes it part of the API again. 

If you mean child, it would be an empty view basically, right? It was written with the intention that it's the single root of a screen-level view. If it becomes a child it would be more brittle I think. (E.g. what happens if we accidentally add 2 children, at some point in the future when we've forgotten the details of how this works?)

Let me know!

tonidero: Yeah I meant children. As in, we could have the PaywallView/CustomerCenterView have an instance of this view, which in turn would have a compose view inside... But yeah, I think having it as a subclass makes sense! Just some random thoughts I had 😅

JayShortway: Oh got it, thanks! Yea I think having it as a superclass is simpler for now. We can always change it, as we "provide no guarantees" for `@InternalRevenueCatAPI` 😄

PR #2528, opened 2025-07-15
[POC] Launching a Paywall in a new Window

JayShortway:
## Description
We can launch a Paywall in a new Window instead of in an Activity. This has the following benefits:
- It allows us to set a realtime listener, instead of only getting a single result after the Paywall is dismissed.
- It allows us to show the Paywall from plain Activities. Our PaywallActivity uses an intermediate PaywallFragment, which means that the launching Activity (from the app) needs to be a FragmentActivity. This requirement can be dropped when using a Window. This allows us to support Paywalls in purchases-unity.

Here's a video showing the realtime listener: 

https://github.com/user-attachments/assets/863e88bd-0586-412b-b3ab-c8df76fa7d3a

Still some rough edges:
- The window is showing underneath the popup. Maybe that's just a bug in PaywallTester, but maybe we can fix it from the Paywall side too.
- We still have to implement transition animations.
- We'll have to add edge to edge support.
- We'll need to remove the PaywallView from the Window when the app's Activity gets destroyed.


aboedo:
awesome

PR #2530, opened 2025-07-15
PaywallActivityLauncher: Add `edgeToEdge` parameter to display paywall in full screen

tonidero:
### Description
We currently were not supporting edge to edge in paywalls when using the `launch` and `launchIfNeeded` APIs. This adds a new parameter that allows to display it in edgeToEdge mode, or not.

The default will be true for Android 15+ and false otherwise.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -52,18 +52,22 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
      * will be used. Only available for original template paywalls. Ignored for v2 Paywalls.
      * @param shouldDisplayDismissButton Whether to display the dismiss button in the paywall.
      * Only available for original template paywalls. Ignored for v2 Paywalls.
+     * @param shouldDisplayEdgeToEdge Whether to display the paywall in edge-to-edge mode.
+     * Default is true for Android 16+, false otherwise.
```
tonidero: I was doubting whether it's worth to default to the same across android versions... But I think we shouldn't cause a breaking change in older android versions.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityArgs.kt
```diff
@@ -1,29 +1,34 @@
...
 
+import android.os.Build
 import android.os.Parcelable
 import com.revenuecat.purchases.ui.revenuecatui.fonts.ParcelizableFontProvider
 import com.revenuecat.purchases.ui.revenuecatui.fonts.PaywallFontFamily
 import com.revenuecat.purchases.ui.revenuecatui.fonts.TypographyType
 import kotlinx.parcelize.Parcelize
 
 internal const val DEFAULT_DISPLAY_DISMISS_BUTTON = true
+internal val defaultDisplayEdgeToEdge = Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM
```
tonidero: I was actually thinking about this default... I changed it to also use edge to edge on android 35+, since that's the default starting on that version.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -52,18 +52,22 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
      * Only available for original template paywalls. Ignored for v2 Paywalls.
+     * @param shouldDisplayEdgeToEdge Whether to display the paywall in edge-to-edge mode.
+     * Default is true for Android 15+, false otherwise.
      */
     @JvmOverloads
     fun launch(
         offering: Offering? = null,
         fontProvider: ParcelizableFontProvider? = null,
         shouldDisplayDismissButton: Boolean = DEFAULT_DISPLAY_DISMISS_BUTTON,
+        shouldDisplayEdgeToEdge: Boolean = defaultDisplayEdgeToEdge,
```
JayShortway: Maybe just `edgeToEdge`? Reason for the suggestion is that it's very similarly named to `shouldDisplayDismissButton`, but that one displays some extra element, and the new parameter displays the Paywall itself in a different way. So it's a bit of a false equivalency.

scripts/api-dump.sh
JayShortway, line 1: chmod changes I guess? Was this necessary on your machine? (You can run this script in Android Studio via the run configuration.)

JayShortway (approved):
Very nice to have this!!

scripts/api-dump.sh
tonidero: Yeah had to add extra permissions to be able to run it from terminal (didn't try to run from AS, thanks for the tip!)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/2530/commits/09a73029090f1c248816fa96a33be1f3ed154a63

Issue #2531, opened 2025-07-15
[Bug] Cannot upgrade any subscription on RevenueCat SDK v9.0.0


shawn-lin-013:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
After purchasing the monthly subscription, the upgrade to the yearly subscription fails every time.
The logcat shows that the issue is caused by "Couldn't find existing purchase for SKU."
However, we can find the purchased monthly subscription in both the Querying Purchases result and activeSubscriptions, so we believe the purchased monthly subscription does exist and is valid.

After further investigation, we found that this issue may be caused by the migration to BillingClient v8.0.0, as Google has removed the ability to query purchase history.

Root Cause:
In toMapOfGooglePurchaseWrapper(), the key of the map is purchaseToken.sha1(),
but in findPurchaseInActivePurchases(), the code uses purchasesByProductId[productId] to retrieve the StoreTransaction from the map created by toMapOfGooglePurchaseWrapper() — which leads to a mismatch between the key (purchaseToken.sha1()) and the lookup (productId).

Logcat
```
PurchasesError(code=PurchaseInvalidError, underlyingErrorMessage=Couldn't find existing purchase for SKU: XXXXXX, message='One or more of the arguments provided are invalid.')
```

function toMapOfGooglePurchaseWrapper in QueryPurchasesByTypeUseCase.kt
```kotlin
private fun List<Purchase>.toMapOfGooglePurchaseWrapper(
    @ProductType productType: String,
): Map<String, StoreTransaction> {
    return this.associate { purchase ->
        val hash = purchase.purchaseToken.sha1()
        hash to purchase.toStoreTransaction(productType.toRevenueCatProductType())
    }
}
```

function findPurchaseInActivePurchases in BillingWrapper.kt
```kotlin
val purchasesRecordWrapper = purchasesByProductId[productId]
if (purchasesRecordWrapper != null) {
    onCompletion(purchasesRecordWrapper)
} else {
    val message = PurchaseStrings.NO_EXISTING_PURCHASE.format(productId)
[...]

PR #2532, opened 2025-07-15
Fix upgrades/downgrades after BC8 update

tonidero:
### Description
Upgrades/downgrades were not working after the changes in #2506. This fixes the issue and adds some tests to the relevant code to avoid it from happening again.

This should deal with #2531

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -466,8 +466,10 @@ internal class BillingWrapper(
                     appInBackground,
                     googleProductType,
                 ),
-                { purchasesByProductId ->
-                    val purchasesRecordWrapper = purchasesByProductId[productId]
+                { purchasesByHashedToken ->
+                    val purchasesRecordWrapper = purchasesByHashedToken.values.firstOrNull {
+                        it.productIds.firstOrNull() == productId
+                    }
                     if (purchasesRecordWrapper != null) {
                         onCompletion(purchasesRecordWrapper)
                     } else {
--- a/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
@@ -50,6 +50,7 @@ import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.models.SubscriptionOption
 import com.revenuecat.purchases.models.SubscriptionOptions
 import com.revenuecat.purchases.strings.BillingStrings
+import com.revenuecat.purchases.strings.PurchaseStrings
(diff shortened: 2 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -466,8 +466,10 @@ internal class BillingWrapper(
                     appInBackground,
                     googleProductType,
                 ),
-                { purchasesByProductId ->
-                    val purchasesRecordWrapper = purchasesByProductId[productId]
+                { purchasesByHashedToken ->
+                    val purchasesRecordWrapper = purchasesByHashedToken.values.firstOrNull {
+                        it.productIds.firstOrNull() == productId
+                    }
```
JayShortway, line 472: 2 questions: 
- How big can `purchasesByHashedToken` get? 
- Can multiple active purchases have the same product ID? Maybe not, because for that to happen the IAP has to be consumed, and if it's consumed, it's no longer active?

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -1646,6 +1647,62 @@ class BillingWrapperTest {
         assertThat(receivedError!!.code).isEqualTo(PurchasesErrorCode.StoreProblemError)
     }
 
+    // region findPurchaseInActivePurchases
```
JayShortway, line 1650: ❤️

JayShortway (approved):
Looks good! Just 2 questions.

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -466,8 +466,10 @@ internal class BillingWrapper(
                     appInBackground,
                     googleProductType,
                 ),
-                { purchasesByProductId ->
-                    val purchasesRecordWrapper = purchasesByProductId[productId]
+                { purchasesByHashedToken ->
+                    val purchasesRecordWrapper = purchasesByHashedToken.values.firstOrNull {
```
tonidero, line 470: Note this structure should only have at most one transaction per product Id. Google won't allow multiple active subs for the same product Id.

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: > How big can purchasesByHashedToken get?

Shouldn't be big. It will be a map of every active subscription basically, but don't expect users to have more than 2-3 at most usually?

> Can multiple active purchases have the same product ID? Maybe not, because for that to happen the IAP has to be consumed, and if it's consumed, it's no longer active?

Just answered in a different comment that I hadn't sent, but no, you can only have one active subs for each product, so this should be unique, if any is found.

PR #2538, opened 2025-07-16
Dont run VC tests on load shedder integration tests

fire-at-will:
### Description
We were running virtual currency integration tests both with and without the load shedder. However, the load shedder doesn't currently support virtual currencies, so these tests were failing when run for the load shedder integration test app. This PR updates the integration tests so that the virtual currency integration tests don't run for the load shedder integration tests app.

### Implementation Details
We are able to accomplish this by:
- When we're running integration tests through the integration tests app, fastlane sets the `IS_RUNNING_LOAD_SHEDDER_INTEGRATION_TESTS` string to `true` in `Constants.kt`.
- There's a new function `BasePurchasesIntegrationTest.isRunningLoadShedderIntegrationTests`, that tests can use to determine if they're running in the load shedder integration test app.

### Testing
I manually forced the integration tests to run on this PR, and both the regular and load shedder integration tests passed in Firebase

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -274,7 +274,8 @@ platform :android do
         google_purchase_token: ENV['LOAD_SHEDDER_GOOGLE_PURCHASE_TOKEN'],
         product_id_to_purchase: ENV['LOAD_SHEDDER_PRODUCT_ID_TO_PURCHASE'],
         base_plan_id_to_purchase: ENV['LOAD_SHEDDER_BASE_PLAN_ID_TO_PURCHASE'],
-        active_entitlements_to_verify: ENV['LOAD_SHEDDER_ACTIVE_ENTITLEMENTS_TO_VERIFY'] || ''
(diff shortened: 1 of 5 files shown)


fastlane/Fastfile
```diff
@@ -352,6 +354,13 @@ platform :android do
       allow_empty: true,
       paths_of_files_to_update: [constants_path]
     )
+    if is_load_shedder_integration_tests
+      replace_text_in_files(
+        previous_text: "IS_RUNNING_LOAD_SHEDDER_INTEGRATION_TESTS",
+        new_text: "true",
```
tonidero, line 360: Should we set this to false if not running load shedder tests? I guess the template value converts to false when converting to a Boolean... but seems better to make it more explicit

fire-at-will: I'll add it!

fire-at-will: Added in 98334d02f0e550bb6f9704a488fef456b23e5288 and re-ran the integration tests 👍

PR #2542, opened 2025-07-17
Expose API key validation result to BillingFactory

tonidero:
### Description
This is the first of a series of PRs to support a test store in purchases-android.

In this one, we add support for a new test API key and expose the API key validation result to the billing factory, so we can in future PRs create a billing implementation for this test store.


purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt
```diff
@@ -21,33 +22,40 @@ internal object BillingFactory {
...
-            } catch (e: NoClassDefFoundError) {
-                errorLog(e) { "Make sure purchases-amazon is added as dependency" }
-                throw e
-            }
+        apiKeyValidationResult: APIKeyValidator.ValidationResult,
+    ): BillingAbstract {
+        if (apiKeyValidationResult == APIKeyValidator.ValidationResult.TEST_STORE) {
+            // WIP: Implement a test store billing provider
+            errorLog { "Using test API key is not yet supported. Assuming non test store." }
         }
```
JayShortway, line 30: I'm curious why we need the `ValidationResult` now, instead of just the `Store`?

purchases/src/main/kotlin/com/revenuecat/purchases/strings/ConfigureStrings.kt
```diff
@@ -23,6 +23,10 @@ internal object ConfigureStrings {
     const val GOOGLE_API_KEY_AMAZON_STORE = "Looks like you're using a Google API key but have configured the SDK " +
         "for the Amazon app store.\nEither use an Amazon API key which should look like 'amzn_1a2b3c4d5e6f7h' or " +
         "configure the SDK to use Google.\nSee https://rev.cat/auth for more details."
+    const val TEST_STORE_API_KEY = "Using a Test Store API key.\n" +
+        "The Test Store is for development only. Never use a Test Store API key in production. " +
+        "Test Store purchases are simulated, do not use Google Play or Amazon store, and generate no revenue. " +
+        "Apps submitted with a Test Store API key will be rejected during App Review."
```
JayShortway: 😚👌

JayShortway: I'd say we should also crash in release builds (can be a separate PR), so maybe we should mention that here too?

purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt
tonidero: Well... we could change the store in the `PurchasesFactory` to a new "test store" when the validation result is a test store... However, that would mean changing the `Store` enum which is public. I preferred to keep the changes internal for now... We might want to add a test store to the `Store` enum, but I think that should be added later. Lmk if you think otherwise!

purchases/src/main/kotlin/com/revenuecat/purchases/strings/ConfigureStrings.kt
ajpallares: Is this possible on Android? I thought about this for iOS, but I don't think it's possible to differentiate between debug vs release builds from the SDK 😞

JayShortway: Sure is 😄 Can be read from [`AppConfig.isDebugBuild`](https://github.com/RevenueCat/purchases-android/blob/e2dd48af9f1e5177f5444b987ebd59f480222606/purchases/src/main/kotlin/com/revenuecat/purchases/common/AppConfig.kt#L20), determined by [`DefaultIsDebugBuildProvider`](https://github.com/RevenueCat/purchases-android/blob/e2dd48af9f1e5177f5444b987ebd59f480222606/purchases/src/main/kotlin/com/revenuecat/purchases/utils/IsDebugBuildProvider.kt#L14), implemented in https://github.com/RevenueCat/purchases-android/pull/1873.

purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt
JayShortway: Oh yes, keeping `Store` the same for now makes all the sense. Thanks for the context!

purchases/src/main/kotlin/com/revenuecat/purchases/strings/ConfigureStrings.kt
ajpallares: Oh, thank you for pointing that! I'm seeing the same in iOS [here](https://github.com/RevenueCat/purchases-ios/blob/66f3f082b9fabb63f2b52b55b99bf3109bd37219/Sources/Misc/SystemInfo.swift#L76-L82)
```
    var isDebugBuild: Bool {
#if DEBUG
        return true
#else
        return false
#endif
    }
```

But I _think_ this doesn't work reliably in iOS. For developers integrating the XCFramework, since we build it for distribution (if I'm not wrong), the binary is created without the `DEBUG` compiler flag, i.e., the code inside the `#else` clause is included in the binary. Therefore, even if the developer is testing/debugging (i.e. using the `DEBUG` compiler flag for their project), the SDK's `isDebugBuild` will be returning `false`

tonidero: Added mention in https://github.com/RevenueCat/purchases-android/pull/2542/commits/cc3b9613df2d1c7a136373f907aebde266875aa1

PR #2549, opened 2025-07-18
Parallel test jobs

vegaro:
`build-test-deploy` times down to [17 min](https://app.circleci.com/pipelines/github/RevenueCat/purchases-android/16278/workflows/125652a3-2c59-4841-bb3c-8683d5966b61) from [37 min](https://app.circleci.com/pipelines/github/RevenueCat/purchases-android/16266/workflows/69393456-77d9-458a-b8b8-0a078312eb07)

- split tests in different jobs per flavor
- remove android dependencies download
- prebuild stuff before running tests


.circleci/config.yml
```diff
@@ -233,39 +233,101 @@ jobs:
...
+      - run:
+          name: Run Defaults Release Tests
+          command: ./gradlew testDefaultsReleaseUnitTest --parallel --no-daemon
       - run:
-          name: Kover HTML
+          name: Generate Kover HTML
           command: ./gradlew purchases:koverHtmlReportDefaultsRelease
       - run:
-          name: Kover XML
+          name: Generate Kover XML
```
tonidero, line 282: So we're only using Kover for the defaults release flavor right? I guess this could leave some paths untested even if they are actually tested... but I don't have a better idea right now, so I'm ok with this 😅

tonidero (approved):
❤️ ❤️ ❤️

.circleci/config.yml
vegaro: right! I noticed that the other day as well, that's mainly why I asked about revisiting it. Let's leave it as is since that's what we had yeah

.circleci/config.yml
```diff
@@ -233,39 +233,101 @@ jobs:
...
+      - run:
+          name: Run CEC Release Tests
+          command: ./gradlew testCustomEntitlementComputationReleaseUnitTest --parallel --no-daemon
+      - run:
+          name: Collect JUnit XMLs
+          command: |
+            mkdir -p build/test-results
+            find . -type f -regex ".*/build/test-results/.*xml" -exec cp --parents {} build/test-results/ \;
       - store_test_results:
           path: build/test-results
```
JayShortway, line 332: This job is copied a few times if I'm reading it correctly. Maybe we can extract it to a `command`, which takes a `gradleTask` parameter or something?

.circleci/config.yml
```diff
@@ -233,39 +233,101 @@ jobs:
...
       - run:
-          name: Verify purchases-ui target SDK compatibility (currently 34)
-          command: ./gradlew :test-apps:testpurchasesuiandroidcompatibility:assembleDebug
+          name: Assemble all variants (warms Gradle cache once)
+          command: ./gradlew testClasses --parallel --no-daemon
+      - android/save_gradle_cache
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
```
JayShortway, line 244: How long does this take? We could use [`save-incremental-gradle-build-to-workspace`](https://github.com/RevenueCat/sdks-circleci-orb/blob/main/src/commands/save-incremental-gradle-build-to-workspace.yml) to only persist the relevant directories.

PR #2550, opened 2025-07-18
Fix paywall tester so customer center doesn't close when changing to dark mode

vegaro:
The Customer Center was closing when switching between light and dark mode

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
@@ -50,21 +50,22 @@ fun AppInfoScreen(
     ),
 ) {
     var isDebugBottomSheetVisible by remember { mutableStateOf(false) }
-    var isCustomerCenterVisible by remember { mutableStateOf(false) }
     var showLogInDialog by remember { mutableStateOf(false) }
     var showApiKeyDialog by remember { mutableStateOf(false) }
 
     // Use remember to cache the listener across recompositions
     val customerCenterListener = remember { createCustomerCenterListener() }
 
-    if (isCustomerCenterVisible) {
+    val state by viewModel.state.collectAsState()
+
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -98,7 +98,7 @@ internal interface CustomerCenterViewModel {
 
     // trigger state refresh
     fun refreshStateIfLocaleChanged()
-    fun refreshStateIfColorsChanged(colorScheme: ColorScheme, isDark: Boolean)
+    fun refreshStateIfColorsChanged(currentColorScheme: ColorScheme, isSystemInDarkTheme: Boolean)
```
vegaro, line 101: this was just renames because I found them confusing

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
```diff
@@ -50,21 +50,22 @@ fun AppInfoScreen(
     ),
 ) {
     var isDebugBottomSheetVisible by remember { mutableStateOf(false) }
-    var isCustomerCenterVisible by remember { mutableStateOf(false) }
```
tonidero, line 53: I believe we could have just used `rememberSaveable`... But I'm ok to move the state to the View model :)

tonidero (approved):
Changes look good! I think there might have been an easier fix, but I'm ok keeping it like this!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
vegaro: Oh right! I overcomplicated it lol

Thank you!

PR #2554, opened 2025-07-22
Support test store

tonidero:
### Description
This adds support for a new test store that allows for easier development testing of purchases. It remains behind a build config flag but for now so not adding anything new for now


api-tester/src/main/java/com/revenuecat/apitester/java/EntitlementInfoAPI.java
```diff
@@ -72,6 +72,7 @@ static void store(final Store store) {
             case RC_BILLING:
             case EXTERNAL:
             case PADDLE:
+            case TEST_STORE:
```
tonidero: Added this since I think it makes sense to differentiate on the store... If/when we want to merge this, to avoid the extra API, I can temporarily remove it. Same for the `PurchaseType` enum.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCBillingProductsResponse.kt
```diff
@@ -0,0 +1,39 @@
...
+    val base: RCBillingPhase? = null,
+    val trial: RCBillingPhase? = null,
+    @SerialName("intro_price") val introPrice: RCBillingPhase? = null,
+)
+
+@Serializable
+internal data class RCBillingPhase(
+    val price: RCBillingPrice? = null,
+    @SerialName("period_duration") val periodDuration: String? = null,
+)
```
ajpallares: In iOS I see you also added here the `cycle_count` integer
```swift
struct PricingPhase {
    let periodDuration: String?
    let price: Price?
    let cycleCount: Int
}
```
Did you just miss it or is it not needed?

tonidero: We should probably add it actually and add it to the subscription options that are part of the `TestStoreProduct`. Just need to change the constructor to accept these parameters, which wasn't available right now. Great catch!

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreProductConverter.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    @JvmSynthetic
+    @Suppress("LongMethod")
+    @Throws(PurchasesException::class)
+    fun convertToStoreProduct(productResponse: RCBillingProductResponse): StoreProduct {
+        val defaultPurchaseOptionId = productResponse.defaultPurchaseOptionId
+        val purchaseOptions = productResponse.purchaseOptions
+
+        val purchaseOptionKey = defaultPurchaseOptionId ?: purchaseOptions.keys.first()
+        val purchaseOption = purchaseOptions[purchaseOptionKey]!!
```
ajpallares: Theoretically, it could happen that `purchaseOptionKey` is `null`, couldn't it? That would mean a crash here if I'm not mistaken

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreProductConverter.kt
```diff
@@ -0,0 +1,91 @@
...
+                    currencyCode = priceObj.currency,
+                )
+                if (basePhase.periodDuration != null) {
+                    period = Period.create(basePhase.periodDuration)
+                }
+            } else {
+                throw PurchasesException(
+                    PurchasesError(
+                        PurchasesErrorCode.ProductNotAvailableForPurchaseError,
+                        "Base priceis required for test subscription products",
```
ajpallares:
```suggestion
                        "Base price is required for test subscription products",
```

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreProductConverter.kt
tonidero: Hmm well, it should always have at least a purchase option I believe, so it should be impossible... However, it's true it's probably better to handle this gracefully just in case 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCBillingProductsResponse.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/2554/commits/da774b0ccd992e4764c87a3e38e96e1b8397ffa5

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreProductConverter.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/2554/commits/b23b1ec0fe8a0cc25ecb99213214dce81c8250d7

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreOfferingParser.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.revenuecat.purchases.models.StoreProduct
+import org.json.JSONObject
+
+internal class TestStoreOfferingParser : OfferingParser() {
+    override fun findMatchingProduct(
+        productsById: Map<String, List<StoreProduct>>,
+        packageJson: JSONObject,
+    ): StoreProduct? {
+        val productIdentifier = packageJson.getString("platform_product_identifier")
+        return productsById[productIdentifier]?.first() // For test store, only one product per identifier should exist.
```
ajpallares, line 13: > // For test store, only one product per identifier should exist.

Based on what I've seen in the iOS codebase (the creation of the mapping from productID to `StoreProduct` [here](https://github.com/RevenueCat/purchases-ios/blob/2bd289a9933b71a8cb4ef1ffe2f536ae39cfb40c/Sources/Purchasing/OfferingsManager.swift#L250)), in iOS we already assume that only one product per identifier should exist.

Is this different in Android due to the different stores (Amazon, Play Store)?

tonidero: This is actually because in Google play, we can have several products with the same id, referring to different base plans within the same product. It's a bit confusing but in google, a product acts similarly to a subscription group in iOS, and each base plan would be a product.

This is not going to be the case in the Test Store, so this relationship of id to products should be 1-1, so we should be able to do this.

ajpallares: Ohh right! I understand now. Thanks for clarifying!

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreBillingWrapper.kt
```diff
@@ -0,0 +1,257 @@
...
+    private val mainHandler: Handler,
+    purchasesStateProvider: PurchasesStateProvider,
+    private val backend: Backend,
+    private val dialogHelper: AlertDialogHelper = DefaultAlertDialogHelper(),
+) : BillingAbstract(purchasesStateProvider) {
+
+    @Volatile
+    private var connected = false
+
+    private val purchaseCache = mutableMapOf<String, StoreTransaction>()
```
tonidero: This cache is to try to mimick more closely what the store would have... But as can be seen, this is just an in-memory cache... It should still allow us to test restores for example.

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreBillingWrapper.kt
```diff
@@ -0,0 +1,257 @@
...
+    ) {
+        debugLog { "TestStoreBillingAbstract: showInAppMessagesIfNeeded - no-op for test store" }
+    }
+
+    override fun getStorefront(
+        onSuccess: (String) -> Unit,
+        onError: PurchasesErrorCallback,
+    ) {
+        debugLog { "TestStoreBillingAbstract: getStorefront - returning US by default" }
+        onSuccess("US")
```
tonidero, line 164: We might need a way to mock this from the backend... I expect some people might ask about it. For now, I think it's ok to hardcode it.

purchases/src/main/kotlin/com/revenuecat/purchases/teststore/TestStoreBillingWrapper.kt
```diff
@@ -0,0 +1,257 @@
...
+                )
+            },
+        )
+    }
+
+    private fun completePurchase(
+        product: StoreProduct,
+        presentedOfferingContext: PresentedOfferingContext?,
+    ) {
+        val purchaseToken = UUID.randomUUID().toString()
```
tonidero: Maybe we want to add some specific details in the token like user id or purchase date... For now just generating a random UUID.

tonidero (commented):
Self review

ajpallares (approved):
Looks great! Amazing job! 🙌

PR #2559, opened 2025-07-22
Update design of No Active subscriptions screen in Customer Center

vegaro:
| Old | New |
|--------|--------|
| Screenshot 2025-07-22 at 16 38 15 | Screenshot 2025-07-22 at 16 37 21 |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.ui.revenuecatui.customercenter.views
 
+import android.content.res.Configuration
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
@@ -15,7 +16,6 @@ import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
@@ -27,6 +27,7 @@ import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterUIC
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterUIConstants.ManagementViewHorizontalPadding
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.actions.CustomerCenterAction
 import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCenterConfigTestData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.theme.CustomerCenterPreviewTheme
 
 @Suppress("LongParameterList")
 @Composable
@@ -72,7 +73,8 @@ private fun ContentUnavailableView(
         color = MaterialTheme.colorScheme.surface,
     ) {
         Column(
-            modifier = Modifier.fillMaxWidth()
+            modifier = Modifier
+                .fillMaxWidth()
                 .padding(vertical = 20.dp, horizontal = 16.dp),
             horizontalAlignment = Alignment.CenterHorizontally,
         ) {
@@ -99,13 +101,14 @@ private fun ContentUnavailableView(
     }
 }
 
-@Preview
+@Preview(showBackground = true, uiMode = Configuration.UI_MODE_NIGHT_NO or Configuration.UI_MODE_TYPE_NORMAL)
+@Preview(showBackground = true, uiMode = Configuration.UI_MODE_NIGHT_YES or Configuration.UI_MODE_TYPE_NORMAL)
 @Composable
 private fun NoActiveUserManagementView_Preview() {
     val testData = CustomerCenterConfigTestData.customerCenterData()
     val noActiveScreen =
         testData.screens[CustomerCenterConfigData.Screen.ScreenType.NO_ACTIVE]!!
-    Surface(color = Color.White) {
+    CustomerCenterPreviewTheme {
         NoActiveUserManagementView(
             screenTitle = noActiveScreen.title,
             screenSubtitle = noActiveScreen.subtitle,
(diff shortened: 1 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
```diff
@@ -56,31 +66,53 @@ private fun ContentUnavailableView(
...
-                style = MaterialTheme.typography.bodyLarge,
-                textAlign = TextAlign.Center,
-                modifier = Modifier.padding(top = ContentUnavailableViewPaddingTopDescription),
+                text = title,
+                style = MaterialTheme.typography.titleMedium,
+                modifier = Modifier.padding(vertical = ContentUnavailableViewPaddingTitle),
             )
+
+            description?.let {
+                Text(
```
tonidero, line 100: Small thing but should we add a bottom padding here of `ContentUnavailableViewPaddingTitle`? Mostly thinking about having the bottom padding  here be the same whether there is a description or not.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
```diff
@@ -56,31 +66,53 @@ private fun ContentUnavailableView(
...
-        )
-
-        Text(
-            text = title,
-            style = MaterialTheme.typography.headlineSmall,
-            modifier = Modifier.padding(top = ContentUnavailableViewPaddingTopTitle),
-        )
+        Column(
+            modifier = Modifier.fillMaxWidth()
+                .padding(all = 20.dp),
```
tonidero: Hmm I wonder if the horizontal padding should be smaller... But I'm ok if we think it looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
```diff
@@ -56,31 +66,53 @@ private fun ContentUnavailableView(
...
 ) {
-    Column(
-        modifier = modifier
-            .fillMaxWidth()
-            .padding(ContentUnavailableViewPadding),
-        horizontalAlignment = Alignment.CenterHorizontally,
+    Surface(
+        modifier = modifier,
+        shape = RoundedCornerShape(CustomerCenterConstants.Card.ROUNDED_CORNER_SIZE),
+        color = MaterialTheme.colorScheme.surface,
```
tonidero, line 76: Just to check but I guess that with this we already handle dark mode right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
```diff
@@ -56,31 +66,53 @@ private fun ContentUnavailableView(
...
+                    textAlign = TextAlign.Center,
+                )
+            }
         }
     }
 }
+
+@Preview
+@Composable
+private fun NoActiveUserManagementView_Preview() {
```
tonidero, line 114: I'm thinking it could be nice to add previews for dark mode as well, so we get snapshot tests automatically.

tonidero (approved):
Just some questions but I think it looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
vegaro: yes!

Screenshot 2025-07-22 at 18 10 29

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
vegaro: hmm I tried adding them in [d594cd0](https://github.com/RevenueCat/purchases-android/pull/2559/commits/d594cd03bcc300d6227f7eea2dac39d06aa41e0a) but I don't think it's working 🤔 

Do you know what could be happening?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
vegaro: hmm maybe it would be better to have only a top padding in the title and the description instead. So if there's no description, the padding at the bottom of the surface is 20.dp

vegaro: 049e317cac847d228e53e47f372bdd4055b7e47b

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
tonidero: Hmm weird... seems like the background is dark mode but it's not getting propagated to the view itself... Maybe it has something to do with the surface and the colors there not responding correctly in the previewer? Still I would expect that to happen also in the AS previewer 🤔. Might be worth contacting Emerge if it only reproduces there.

vegaro: It happens in the AS previewer as well...

vegaro: ok, I found out why, the CustomerCenterPreviewTheme didn't have a dark color scheme

tonidero: Ohh makes sense 😅. Thanks for fixing it!

tonidero (approved):
🙌

PR #2560, opened 2025-07-22

RCGitBot:


JayShortway:
@tonidero Hmm, that's annoying. We're [sorting by offering identifier](https://github.com/RevenueCat/purchases-android/blob/d34f8c6443b525d52849eecd2dcc0541b4938579/ui/revenuecatui/src/debug/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TemplatePreviews.kt#L86) to avoid this. 🤔

JayShortway:
@tonidero Oh but of course, if the new one ends up somewhere in the middle, the later ones all get shifted.

tonidero:
Ohhh that makes sense... I think it might be worth finding a different sorting mechanism... But maybe it's ok to merge this for now then.

PR #2564, opened 2025-07-24
Display latest expired subscription if no active subscriptions

vegaro:
Display the latest expired subscription in the customer center, if there are no active subscriptions


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -517,6 +529,13 @@ internal class CustomerCenterViewModelImpl(
         }
     }
 
+    private fun findLatestExpiredSubscription(customerInfo: CustomerInfo): TransactionDetails.Subscription? {
+        return customerInfo.subscriptionsByProductIdentifier.values
+            .filter { !it.isActive }
+            .sortedByDescending { it.expiresDate }
+            .firstOrNull()?.asTransactionDetails()
```
tonidero: Very nitpicky but insted of sorting + getting first, we could try to get the min value by expires date with `minBy { it.expiresDate }` I think? Small optimization since we avoid all the sorting.

tonidero (approved):
Looks good to me! Just a small possible optimization

PR #2565, opened 2025-07-24
Add lifetime badge to customer center

vegaro:
Adds a Lifetime badge on non-subscriptions that are attached to an entitlement. We don't have a way to distinguish between consumables and non-consumables so we are using the "is attached to an entitlement" to assume it's a lifetime purchase:

Screenshot 2025-07-24 at 17 04 29


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PurchaseInformationCardViewTest.kt
```diff
@@ -0,0 +1,247 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.ExpirationOrRenewal
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.PriceDetails
+import io.mockk.mockk
+import io.mockk.verify
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class PurchaseInformationCardViewTest {
```
vegaro, line 15: @tonidero @JayShortway what do you think of this type of test?

tonidero: Hmm I wonder if it would make sense to just make a preview so we get snapshot tests instead? To test UI behavior that's what I personally prefer... I think these tests make more sense for UI behavior, like transitions. Wdyt?

JayShortway: Not opposed to having these tests, but if it's a visual test of static content, a preview could cover more ground (e.g. not only _that_ it's displayed but also _how_ it's displayed).

For behavior testing (e.g. `clickable card triggers callback when clicked`), this makes a ton of sense indeed.

(Small note: `promotional purchase shows free trial badge` doesn't seem to have any assertion.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
```diff
@@ -36,6 +36,11 @@ internal data class PurchaseInformation(
      * until the end of the billing period.
      */
     val isCancelled: Boolean,
+    /**
+     * Indicates whether this is a lifetime purchase.
+     * This is true for promotional lifetime purchases or non-subscription purchases attached to an entitlement.
```
facumenzella, line 41: Shouldn't this be true as well for `non-subscription` that are not attached to an entitlement?

vegaro: They could be consumables. In Android we don't have a way to distinguish between consumables and non-consumables, we only know susbcription or non-subscription. Since non-consumables (lifetimes) are normally attached to entitlements we use that as a proxy

PR #2569, opened 2025-07-28
Add missing dokka-hide-internal tests to CI and fix reported number of tests 

vegaro:
#2549 split tests but it had a difference missing around 3000 tests compared to the previous test logic.

After some debugging I realized we were missing a module's test `dokka-hide-internal`. When adding that I realized that it was reporting 6 tests, but we only have 3 tests in there. That helped me realize we had an issue storing tests, and we were double counting them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,9 +79,9 @@ Regarding API changes, we've also removed data classes from our public APIs. Thi
 
 ### Play Billing Library 8: No expired subscriptions or consumed one-time products
 
-**Note:** the following is only relevant if you recently integrated RevenueCat before upgrading to v9, and do not (yet) have all your transactions imported. 
-  
-Play Billing Library 8 removed functionality to query expired subscriptions or consumed one-time products. This means that, for users migrating from a non-RevenueCat implementation of the Play Billing Library, the SDK will not be able to send purchase information from these purchases. We can still ingest historical data from these purchases through a backend historical import. See [docs](https://www.revenuecat.com/docs/migrating-to-revenuecat/migrating-existing-subscriptions). This doesn't affect developers that have all transactions in RevenueCat, which is true for the vast majority.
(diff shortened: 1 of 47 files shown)


.circleci/config.yml
```diff
@@ -257,10 +257,10 @@ jobs:
       - run:
           name: Collect JUnit XMLs
           command: |
-            mkdir -p build/test-results
-            find . -type f -regex ".*/build/test-results/.*xml" -exec cp --parents {} build/test-results/ \;
+            mkdir -p build/test-results-defaults-debug
+            find . -path ./build/test-results-defaults-debug -prune -o -type f -regex ".*/build/test-results/.*xml" -exec cp --parents {} build/test-results-defaults-debug/ \;
```
vegaro, line 261: This was causing tests to be duplicated. xmls were being stored duplicated like:

```
/home/circleci/project/build/test-results/ui/revenuecatui/build/test-results/testDefaultsDebugUnitTest/TEST-com.revenuecat.purchases.ui.revenuecatui.extensions.PackageExtensionsTest.xml
/home/circleci/project/build/test-results/build/test-results/ui/revenuecatui/build/test-results/testDefaultsDebugUnitTest/TEST-com.revenuecat.purchases.ui.revenuecatui.extensions.PackageExtensionsTest.xml
```

The root cause of the duplication is that the find command is gathering test reports and copying them into a directory, and then on a subsequent pass, it finds the copies it just made and copies them again, creating the nested, duplicated paths.

The fix is to tell the find command to explicitly ignore its own output directory using the `prune`. The command now reads "Search the project. If you encounter the directory named <destination_dir>, do not enter it (-prune). Otherwise (-o), if you find a file that matches our test report pattern, copy it."

I verified that after this change the number of reported tests in CircleCI is half as before. And that running the split tests + the dokka tests, matches the previous test workflow (by half)

PR #2570, opened 2025-07-29
Remove TestStoreProduct purchase check

tonidero:
### Description
Until now we had a check that threw an exception when trying to purchase using a TestStoreProduct. However, as part of supporting the Test store we want to remove this check. The purchase method would still fail later, when checking the purchase type in the Google/Amazon stores.

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -355,13 +355,7 @@ class Purchases internal constructor(
         storeProduct: StoreProduct,
         callback: PurchaseCallback,
     ) {
-        purchasesOrchestrator.startPurchase(
-            activity,
-            storeProduct.purchasingData,
-            null,
-            null,
-            callback,
-        )
+        purchase(PurchaseParams.Builder(activity, storeProduct).build(), callback)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -355,13 +355,7 @@ class Purchases internal constructor(
...
         callback: PurchaseCallback,
     ) {
-        purchasesOrchestrator.startPurchase(
-            activity,
-            storeProduct.purchasingData,
-            null,
-            null,
-            callback,
-        )
+        purchase(PurchaseParams.Builder(activity, storeProduct).build(), callback)
```
tonidero, line 358: This is unrelated... but I saw that we were using methods that could be made internal in purchasesOrchestrator and took the opportunity to refactor it.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -60,17 +59,6 @@ class PurchaseParams(val builder: Builder) {
         constructor(activity: Activity, storeProduct: StoreProduct) :
             this(activity, storeProduct.purchasingData, storeProduct.presentedOfferingContext, storeProduct)
 
-        private fun ensureNoTestProduct(storeProduct: StoreProduct) {
```
tonidero, line 63: This is the main change in this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1264,62 +1264,6 @@ internal class PurchasesOrchestrator(
         }
     }
 
-    fun startDeprecatedProductChange(
```
tonidero, line 1267: This was unused... So took the opportunity to remove it

PR #2573, opened 2025-07-29
Convert WebBilling products into TestStoreProducts

tonidero:
### Description
This is another step in preparation to support the new TestStore. This adds the logic to transform WebBilling products into `TestStoreProduct`. These will be used for purchasing when using a test api key.

The logic here is currently unused, but will be used in follow-up PRs

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
@@ -31,5 +31,4 @@ internal object PriceFactory {
 
         return Price(formatted, amountMicros, currencyCode)
     }
-
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
@@ -1,0 +1,35 @@
+@file:JvmSynthetic
+
+package com.revenuecat.purchases.utils
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.common.SharedConstants.MICRO_MULTIPLIER
+import com.revenuecat.purchases.models.Price
+import java.text.NumberFormat
+import java.util.Currency
+import java.util.Locale
+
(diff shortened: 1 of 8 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
```diff
@@ -0,0 +1,34 @@
...
+package com.revenuecat.purchases.utils
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.common.SharedConstants.MICRO_MULTIPLIER
+import com.revenuecat.purchases.models.Price
+import java.text.NumberFormat
+import java.util.Currency
+import java.util.Locale
+
+internal object PriceFactory {
```
ajpallares, line 12: Nice touch centralizing the logic 👌

purchases/src/test/java/com/revenuecat/purchases/teststore/TestStoreProductConverterTest.kt
```diff
@@ -72,10 +75,10 @@ class TestStoreProductConverterTest {
             )
         )
 
-        val result = TestStoreProductConverter.convertToStoreProduct(productResponse)
+        val result = convertToStoreProduct(productResponse)
```
ajpallares, line 78: Now that the `locale` is a parameter of the `convertToStoreProduct` method, perhaps we should pass the locale here in the test to make them deterministic? (the same in the other tests)

ajpallares (approved):

purchases/src/test/java/com/revenuecat/purchases/teststore/TestStoreProductConverterTest.kt
tonidero: Yup! This is actually a method in the tests that just uses the hardcoded locale for tests, that way we make it deterministic. Lmk if you meant something else!

ajpallares: Ohhh, didn't realize that. Sorry! Then it's perfect. Thanks for clarifying! 🙏

PR #2576, opened 2025-07-31
Don't show product identifier for non-Google purchases in Customer Center

vegaro:
We were displaying the product id for purchases that we don't have the details for. This happens for example for Apple purchases, or for Stripe. 

Screenshot 2025-07-31 at 14 12 32

With this change we will start displaying the entitlement, or if there's no entitlement, "Subscription" or "One time purchase"

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
@@ -49,8 +49,9 @@ internal data class PurchaseInformation(
         transaction: TransactionDetails,
         dateFormatter: DateFormatter = DefaultDateFormatter(),
         locale: Locale,
+        localization: CustomerCenterConfigData.Localization? = null,
     ) : this(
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/PurchaseInformation.kt
```diff
@@ -101,25 +104,56 @@ internal data class PurchaseInformation(
...
+        }
+
+        // In sandbox, we don't know if the price is actually free or not (it's always 0)
+        // So we fall back to the product price.
+        transaction.price?.amountMicros == 0L && !transaction.isSandbox -> {
+            PriceDetails.Free
+        }
+
+        subscribedProduct != null -> {
+            if (subscribedProduct.price.amountMicros == 0L) {
```
facumenzella, line 125: Is this a valid comparison? Or should we check if it's almost zero? In swift you have https://developer.apple.com/documentation/swift/float/iszero

vegaro: It's a `Long` so I believe it's fine

PR #2577, opened 2025-07-31
Adds new tab ID properties to `TabsComponent`.

JayShortway:
## Description
Adds the `tab_id`, `id` and `default_tab_id` properties to the `TabsComponent` model where necessary. These new properties are not used yet. This makes the same changes to the JSON models as these iOS PRs: 
- https://github.com/RevenueCat/purchases-ios/pull/5209
- https://github.com/RevenueCat/purchases-ios/pull/5430


tonidero (approved):
:shipit:

tonidero:
Looks like we need to fix paywall tester

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TabsComponent.kt
```diff
@@ -25,6 +25,8 @@ class TabControlButtonComponent(
     @get:JvmSynthetic
     @SerialName("tab_index")
     val tabIndex: Int,
+    @SerialName("tab_id")
+    val tabId: String,
```
JayShortway, line 31: @joshdholtz Just checking, is it correct that this property will always exist in the response?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/TabsComponent.kt
```diff
@@ -85,12 +87,17 @@ class TabsComponent(
...
+    val defaultTabId: String? = null,
+    @get:JvmSynthetic
     val overrides: List<ComponentOverride<PartialTabsComponent>> = emptyList(),
 ) : PaywallComponent {
     @InternalRevenueCatAPI
     @Poko
     @Serializable
     class Tab(
+        @get:JvmSynthetic
+        val id: String,
```
JayShortway, line 104: @joshdholtz Same question here, is it correct that this property will always exist in the response?

PR #2579, opened 2025-08-01
Expose getStorefront APIs in CustomEntitlementComputation flavor

tonidero:
### Description
While CEC already had the `storefrontCountryCode`, this also adds the APIs `getStorefrontCountryCode` and `awaitStorefrontCountryCode` to the CEC flavor.

### DIFF
--- a/purchases/api-defauts.txt
+++ b/purchases/api-defauts.txt
@@ -27,13 +27,14 @@ package com.revenuecat.purchases {
     method @kotlin.jvm.JvmSynthetic public static suspend Object? awaitPurchaseResult(com.revenuecat.purchases.Purchases, com.revenuecat.purchases.PurchaseParams purchaseParams, kotlin.coroutines.Continuation<? super kotlin.Result<? extends com.revenuecat.purchases.PurchaseResult>>);
     method @kotlin.jvm.JvmSynthetic @kotlin.jvm.Throws(exceptionClasses=PurchasesTransactionException::class) public static suspend Object? awaitRestore(com.revenuecat.purchases.Purchases, kotlin.coroutines.Continuation<? super com.revenuecat.purchases.CustomerInfo>);
     method @kotlin.jvm.JvmSynthetic public static suspend Object? awaitRestoreResult(com.revenuecat.purchases.Purchases, kotlin.coroutines.Continuation<? super kotlin.Result<? extends com.revenuecat.purchases.CustomerInfo>>);
+    method public static suspend Object? awaitStorefrontCountryCode(com.revenuecat.purchases.Purchases, kotlin.coroutines.Continuation<? super java.lang.String>);
   }
 
   public final class CoroutinesExtensionsKt {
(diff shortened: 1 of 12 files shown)


purchases/api-defauts.txt
```diff
@@ -172,8 +174,8 @@ package com.revenuecat.purchases {
...
-    method public static void getStorefrontCountryCodeWith(com.revenuecat.purchases.Purchases, optional kotlin.jvm.functions.Function1<? super com.revenuecat.purchases.PurchasesError,kotlin.Unit> onError, kotlin.jvm.functions.Function1<? super java.lang.String,kotlin.Unit> onSuccess);
     method @Deprecated public static void getSubscriptionSkusWith(com.revenuecat.purchases.Purchases, java.util.List<java.lang.String> skus, optional kotlin.jvm.functions.Function1<? super com.revenuecat.purchases.PurchasesError,kotlin.Unit> onError, kotlin.jvm.functions.Function1<? super java.util.List<? extends com.revenuecat.purchases.models.StoreProduct>,kotlin.Unit> onReceiveSkus);
+    method public static void getVirtualCurrenciesWith(com.revenuecat.purchases.Purchases, optional kotlin.jvm.functions.Function1<? super com.revenuecat.purchases.PurchasesError,kotlin.Unit> onError, kotlin.jvm.functions.Function1<? super com.revenuecat.purchases.virtualcurrencies.VirtualCurrencies,kotlin.Unit> onSuccess);
```
tonidero, line 178: I'm confused why these are not updated nor found in the checks...

JayShortway: Yea it seems like we cannot rely on the check at all. I think we might need to go the git diff route?

tonidero: Yup indeed

purchases/api-defauts.txt
ajpallares, line 1: I'm curious: what are these txt files for?

tonidero: They have the public API for our libraries, both in the defaults flavor and the cec flavor. We're using metalava to generate them. Unfortunately, the check task itself doesn't seem to be working, so we might need to fallback to always dump these api files and compare using git.

JayShortway: These are "dumps" of our public API surface. We have them in other repos too, like [purchases-kmp](https://github.com/RevenueCat/purchases-kmp/blob/main/core/api/core.api). The idea being that any PR would check if the committed text file is still valid. If not, the PR author should regenerate the file, which would add the diff to the PR, allowing us to review it. 

In purchases-android we use Metalava for this (a Google tool), but the "check" task doesn't seem to work consistently. It overlooks changes to the public API, which becomes evident once we regenerate the files and see unrelated changes. (Virtual Currency stuff in this PR for instance.)

JayShortway: (Ha, I should have refreshed before hitting send 😅)

ajpallares: I understand now. Thank you both so much! 🫶 
(two explanations is better than one 😄)

PR #2584, opened 2025-08-01
Fix insets in Paywall Tester

vegaro:
I had noticed wrong padding in the top app bar in the customer center. After some time trying to understand I came to the conclusion that the problem was the nested Scaffold. We have one in the MainScreen of Paywall Tester, and one in the Customer Center. 

It was adding the insets to the top of the top app bar, because they were not being consumed. I thought we had an issue in the Customer Center but it was actually in the paywall tester. Since the outer Scaffold was adding a padding, and not consuming the insets, this is how it was being displayed:

Screenshot 2025-08-01 at 14 42 32

I removed the padding, and added it to each of the screens as needed, and consumed it, so the top bar didn't get unconsumed insets and tried adding them again. This is the final result:

Screenshot 2025-08-01 at 14 41 49

I also noticed the bottom bar was not material 3, and the insets were also not being automatically handled


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/MainScreen.kt
```diff
@@ -69,25 +70,35 @@ private fun MainNavHost(
...
         navController,
         startDestination = Tab.Paywalls.route,
-        modifier = modifier,
     ) {
         composable(Tab.AppInfo.route) {
-            AppInfoScreen()
+            AppInfoScreen(
+                modifier = Modifier
+                    .padding(contentPadding)
+                    .consumeWindowInsets(contentPadding),
```
tonidero: Hmm I'm wondering, would it make more sense not pass the contentPadding as a parameter, but instead just use the [safeContentPadding](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier#(androidx.compose.ui.Modifier).safeContentPadding()) modifier? 

That's supposed to respect unconsumed safe area and also handles consuming the safe area insets by itself.

vegaro: interestingly I tried that and it adds some unexpected padding 🤔 

Screenshot 2025-08-01 at 15 41 31

JayShortway: I think the problem is that the insets are not consumed when they should be. That's why forcefully consuming them with `consumeWindowInsets()` fixes it (assumption).

Small note: instead of adding it to each child of the `NavHost`, I think we can also add it to the `NavHost` itself, or even to the `MainNavHost`.

vegaro: This works though:
```
modifier = Modifier.systemBarsPadding(),
```

vegaro: >Small note: instead of adding it to each child of the NavHost, I think we can also add it to the NavHost itself, or even to the MainNavHost.

Doing that will not make the customer center edge to edge, but it has correct padding

vegaro: > I think the problem is that the insets are not consumed when they should be. 

you mean that the outer `Scaffold` should be consuming the insets, so they are consumed by the time the top app bar in the inner scaffold renders right?

JayShortway: >you mean that the outer Scaffold should be consuming the insets, so they are consumed by the time the top app bar in the inner scaffold renders right?

Yes!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
```diff
@@ -92,13 +57,6 @@ fun AppInfoScreen(
         val currentApiKeyDescription by remember { derivedStateOf { state.apiKeyDescription } }
         Text(text = "Current user ID: $currentUserID")
         Text(text = "Current API key: $currentApiKeyDescription")
-        lastCustomAction?.let { action ->
-            Text(
-                text = "Last custom action: $action",
-                style = MaterialTheme.typography.bodySmall,
-                color = MaterialTheme.colorScheme.primary,
-            )
-        }
```
vegaro, line 101: I asked @facumenzella and he's ok with removing this. We have a Toast already and its mainly for debugging purposes

tonidero (approved):
This all makes sense to me! Thanks for fixing this!

tonidero:
Hmm seems like the screenshots size itself changed for some reason which is super weird since I don't see any related changes...

vegaro:
oh right, I thought things were pushed down, but it's the screenshot size!

vegaro:
I am going to merge since the screenshot change seems unrelated

PR #2585, opened 2025-08-01
Change metalava job to diff dump generated dump files

tonidero:
### Description
The metalava check jobs have been failing for a while, giving success even though there were changes in the api files. This PR changes to not use the check jobs, and instead always dump the api, and check git for changes.

### DIFF
--- a/scripts/api-check.sh
+++ b/scripts/api-check.sh
@@ -6,7 +6,7 @@ exit_code=0
 ./scripts/api-dump.sh || exit_code=$?
 # Check if there are any dirty changes in git
 if ! git diff --quiet; then
-  echo "API dump has changes, please review and commit them."
+  echo "API dump has changes, run the scripts/api-dump.sh script to generate the API dump, review and commit them."
   exit_code=1
 fi
 


scripts/api-check.sh
```diff
@@ -2,7 +2,12 @@
 
 exit_code=0
 
-./gradlew metalavaCheckCompatibilityDefaultsRelease || exit_code=$?
-./gradlew metalavaCheckCompatibilityCustomEntitlementComputationRelease || exit_code=$?
+# Run the api-dump.sh script to generate the API dump
+./scripts/api-dump.sh || exit_code=$?
+# Check if there are any dirty changes in git
+if ! git diff --quiet; then
+  echo "API dump has changes, please review and commit them."
```
ajpallares: Would it be worth mentioning in this message how to generate the dump? something like the comment above
> "Run the api-dump.sh script to generate the API dump"

Perhaps it's too obvious but might be helpful for those of us not so familiar with it. FFTI

ajpallares (approved):
Thanks for fixing!

scripts/api-check.sh
tonidero: I don't think it's a problem to add it :) Good thinking!

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2585/commits/eb420c907a9294243b2b7447435c0eae8ab6f354

PR #2590, opened 2025-08-05
Fail configure call if using test api key in release builds

tonidero:
### Description
Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/5453

This PR introduces a crash at runtime if the app is running on release (on a debuggable process) when configuring the SDK with a Test Store API Key. Debug builds are unaffected.


ajpallares (approved):
Nice one!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -61,6 +61,10 @@ internal class PurchasesFactory(
     private val apiKeyValidator: APIKeyValidator = APIKeyValidator(),
 ) {
 
+    private fun isTestStoreEnabled(): Boolean {
+        return BuildConfig.ENABLE_TEST_STORE
+    }
```
JayShortway: Maybe this can be provided via the constructor, so it can be overridden in tests, and we can already enable the test in this PR? 

```kotlin
internal class PurchasesFactory(
    private val isDebugBuild: IsDebugBuildProvider,
    private val apiKeyValidator: APIKeyValidator = APIKeyValidator(),
    private val isTestStoreEnabled: () -> Boolean = { BuildConfig.ENABLE_TEST_STORE },
)
```

tonidero: I like it! Did it that way in https://github.com/RevenueCat/purchases-android/pull/2590/commits/6c7fd7a33777879ee0acd9303587712ae1dc12c5

JayShortway (approved):

PR #2594, opened 2025-08-07
Add alias users backend call

tonidero:
### Description
This adds the alias endpoint call. Currently unused but will be used in follow-up PRs


ajpallares (approved):
Makes sense!

JayShortway (approved):

PR #2595, opened 2025-08-07
Use `Block store` to backup anonymous user ids across installations

tonidero:
### Description
Based on #2594 

This uses [Block store](https://developer.android.com/identity/block-store) to store the anonymous user id in the cloud when we find it has any purchases. This user ID will be recovered and aliased together with any new anonymous user id when restoring or completing a purchase.

This will be useful to avoid issues when trying to restore purchases with consumed consumables that were meant to never be consumed.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
@@ -129,7 +129,6 @@ constructor(
         }
     }
 
-    // Stores the given userId in the Blockstore and returns the number of bytes stored.
     @Suppress("ReturnCount")
     private suspend fun storeUserIdIfNeeded(
         blockstoreDataMap: Map<String, BlockstoreData>,


examples/purchase-tester/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,7 @@
 
     <application
             android:name="com.revenuecat.purchasetester.MainApplication"
-            android:allowBackup="true"
+            android:allowBackup="false"
```
tonidero, line 9: This disables auto backup, which helps to test block store. Auto backup should also help minimize the spread of this issue in any case, and shouldn't conflict with these changes.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+    private val ioScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
+    private val mainScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main),
+) {
+
+    private companion object {
+        const val BLOCKSTORE_USER_ID_KEY = "revenuecat_blockstore_user_id"
+        const val BLOCKSTORE_MAX_ENTRIES = 16
+    }
+
+    fun storeUserIdIfNeeded(customerInfo: CustomerInfo) {
```
tonidero, line 39: We are only storing if:
- The user id is anonymous
- Has purchases
- No previous user id was stored
- We haven't reached the maximum 16 entries of storage for Block store.

We try to store:
- When obtaining CustomerInfo on app foreground
- When making a purchase

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+                try {
+                    storeBlockstoreUserId(currentUserId)
+                } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                    errorLog { "Failed to store user Id in Block store: $e" }
+                }
+            }
+        }
+    }
+
+    fun recoverAndAliasBlockstoreUserIfNeeded(callback: () -> Unit) {
```
tonidero: We will try to recover the user Id from Block store and alias the users together if:
- Current user is anonymous
- There is a user Id cached in Block store
- That user id is not the same as the current user id.

We try to do this:
- When calling `restorePurchases`
- When completing a purchase

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+                        "Any purchases on previous anonymous user will not be recovered."
+                }
+                callCompletion()
+                return@launch
+            }
+            callCompletion()
+        }
+    }
+
+    fun clearBlockstoreUserIdBackupIfNeeded(callback: () -> Unit) {
```
tonidero: We clear the block store backup when a user logs in. In that case, we shouldn't rely on this system anymore and the developer should just rely on their user ids.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -531,63 +543,67 @@ internal class PurchasesOrchestrator(
...
-                        callback?.onReceived(customerInfo, created)
-                        customerInfoUpdateHandler.notifyListeners(customerInfo)
-                    }
-                    offeringsManager.fetchAndCacheOfferings(newAppUserID, state.appInBackground)
-                },
-                onError = { error ->
-                    dispatch { callback?.onError(error) }
-                },
-            )
+            blockstoreHelper.clearBlockstoreUserIdBackupIfNeeded {
```
JayShortway: What if people do things very wrong and call `logIn` with an anonymous ID? Should we maybe check for that (to be absolutely sure), and alias in that case?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -259,6 +261,15 @@ internal class PurchasesOrchestrator(
                     fetchPolicy = CacheFetchPolicy.FETCH_CURRENT,
                     appInBackground = false,
                     allowSharingPlayStoreAccount = allowSharingPlayStoreAccount,
+                    callback = object : ReceiveCustomerInfoCallback {
+                        override fun onReceived(customerInfo: CustomerInfo) {
+                            blockstoreHelper.storeUserIdIfNeeded(customerInfo)
```
JayShortway, line 266: Maybe we should do this at configure time too, instead of only when the app is foregrounded? That way, we don't care if the app is in the foreground or background, and we would cover more ground, avoiding the limitation that users need to foreground the app for this system to work.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+internal class BlockstoreHelper(
+    applicationContext: Context,
+    private val identityManager: IdentityManager,
+    private val blockstoreClient: BlockstoreClient = Blockstore.getClient(applicationContext),
+    private val ioScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
+    private val mainScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main),
+) {
+
+    private companion object {
+        const val BLOCKSTORE_USER_ID_KEY = "revenuecat_blockstore_user_id"
```
JayShortway: I'd suggest using a fully qualified package name, and the `blockstore` part is redundant imo. Something like: 
```suggestion
        const val BLOCKSTORE_USER_ID_KEY = "com.revenuecat.purchases.app_user_id"
```

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+        const val BLOCKSTORE_MAX_ENTRIES = 16
+    }
+
+    fun storeUserIdIfNeeded(customerInfo: CustomerInfo) {
+        if (!identityManager.currentUserIsAnonymous() || customerInfo.allPurchasedProductIds.isEmpty()) return
+        ioScope.launch {
+            val blockstoreData = try {
+                getBlockstoreData()
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                errorLog { "Failed to retrieve Block store data. Will not store userId. Error: ${e.message}" }
```
JayShortway: Can we also print/log the stack trace? 

(Comment applies to the other catch statements in this file too.)

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+                .addOnSuccessListener { cont.resume(it.blockstoreDataMap) }
+                .addOnFailureListener { cont.resumeWithException(it) }
+        }
+    }
+
+    private fun shouldStoreBlockstoreUserId(blockstoreDataMap: Map<String, BlockstoreData>): Boolean {
+        return blockstoreDataMap[BLOCKSTORE_USER_ID_KEY] == null && blockstoreDataMap.size < BLOCKSTORE_MAX_ENTRIES
+    }
+
+    private suspend fun storeBlockstoreUserId(userId: String): Int {
```
JayShortway: Maybe `shouldStoreBlockstoreUserId` can be part of `storeBlockstoreUserId`, to avoid always having to wrap the latter function in an `if` check.

Also a suggestion to remove `Blockstore` from the function name, as we're already in the `BlockstoreHelper` class. This could be something like `storeUserIdIfNeeded`?

JayShortway: Could we have 1 line of documentation on the return type? 😄

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+    }
+
+    fun recoverAndAliasBlockstoreUserIfNeeded(callback: () -> Unit) {
+        val callCompletion = {
+            mainScope.launch {
+                callback()
+            }
+        }
```
JayShortway: This could be a local `fun` too, but no biggie imo.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
JayShortway: Naming suggestion:

```suggestion
    fun aliasCurrentAndStoredUserIdsIfNeeded(callback: () -> Unit) {
```

I think the "recover" part is beyond the concern of the `BlockstoreHelper`. It just deals with user IDs. It doesn't really care about the side effect of purchases being recovered.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
JayShortway: Should we not alias whatever ID we have in the block store with the new ID? We could have an ID in the Block store with consumed purchases?

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -66,6 +70,32 @@ internal class IdentityManager(
         }
     }
 
+    suspend fun aliasOldUserIdToCurrentOne(
+        oldAppUserID: String,
+    ) {
```
JayShortway: Naming suggestion: 
```suggestion
    suspend fun aliasCurrentUserIdTo(
        oldAppUserID: String,
    ) {
```
  
  
I like how that looks on the call site, but I realize that's subjective so definitely not a blocker.
```kotlin
identityManager.aliasCurrentUserIdTo(oldUserId)
```

purchases/src/main/kotlin/com/revenuecat/purchases/strings/IdentityStrings.kt
```diff
@@ -3,6 +3,7 @@ package com.revenuecat.purchases.strings
 internal object IdentityStrings {
     const val LOGGING_IN = "Logging in from %s -> %s"
     const val LOG_IN_SUCCESSFUL = "Logged in successfully as %s. Created: %s"
+    const val ALIAS_OLD_USER_ID_TO_CURRENT_SUCCESSFUL = "Alias old user ID %s to current user ID %s successful."
```
JayShortway:
```suggestion
    const val ALIAS_OLD_USER_ID_TO_CURRENT_SUCCESSFUL = "Successfully aliased old user ID %s to current user ID %s."
```

JayShortway (commented):
Thanks for taking care of this!! 💪

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: I would say they are seriously holding it wrong 😅 And in that case, this would clear the block store user id, and it would be cached again on the next app foreground/customer info refresh. 

So I guess the case we would miss in this case is if you make a purchase before the logIn, then logIn, then uninstall before the new userId is cached. We would lose the userId that did the purchase.

All things said, I think it's quite an edge case... but then again, I think we should be able to handle it yeah... I will try to do this today!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Thing is I wanted to only cache users after checking if they have any purchases. We only fetch customer info on app foreground (to avoid background requests).

We could change to cache the user in block store even without purchases, but just trying to limit the number of times we need to interact with the block store. It doesn't say anything about rate limits, but I thought it best to limit it. 

I think that would only help for users that had previous purchases before migrating to a version with this, and that don't foreground the app, so I think it's ok to live with that?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Ah wait, logIn didn't create aliases, if the new user id is anonymous I forgot... So yeah, we might need to alias the users ourselves in this case. Still quite an edge case, but better to control it yeah

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
tonidero: But the moment the user calls `logIn`, those users already get aliased (as long as the new user id is not anonymous) so we don't need to alias them again. The developer can just rely in the user Id they passed us, which should already have the purchases of the user before logging in.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Actually, what I'm going to do is to avoid clearing the user id from block store if the new user id is anonymous. It won't give the new user any entitlements, but that's in line with what was happening now I think. If the new anonymous user restores, they should still be able to obtain those purchases.

tonidero: I guess this could still cause some issues, if they logIn with a non-anonymous user AFTER logging in with an anonymous user... TBH, not sure how much we want to handle this edge case though... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
JayShortway: Right, yea it's probably good to be on the safe side wrt rate limits. On the other hand this is still only a few requests per device right? The Block Store API is also designed to be called on app startup if I understand it correctly?

>I think that would only help for users that had previous purchases before migrating to a version with this, and that don't foreground the app

Exactly, so even if the app receives a notification, or has a scheduled job that kicks off, or the app process starts for some other reason, we would store the user ID (assuming they're configuring in `Application.onCreate()`). It would expand the pool of users we capture.

JayShortway (approved):
Thanks for making the changes! 🙏 Just the configure vs onAppForegrounded thing, but not a blocker for me!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Well... For normal devs calling configure in `Application.onCreate`, this would mean potentially calling this quite a bit whenever the app is started even in the background. For example, when there are push notifications, unless the developer Jitters the notifications, it can potentially awake a ton of apps (this happened on my previous company). This could amount to a lot of immediate requests and resulting on being rate limited.

Specially, I think the amount of users we would "earn" by doing this, being those whose app is launched in the background but never open it, seem to be small enough enough IMO to prefer avoiding this risk.

JayShortway: Yea alright, happy to leave it as is!

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,142 @@
...
+                .addOnSuccessListener { cont.resume(it.blockstoreDataMap) }
+                .addOnFailureListener { cont.resumeWithException(it) }
+        }
+    }
+
+    // Stores the given userId in the Blockstore and returns the number of bytes stored.
+    private suspend fun storeUserIdIfNeeded(
+        blockstoreDataMap: Map<String, BlockstoreData>,
+        userId: String,
+    ): Int {
```
vegaro: It looks like the return value is never used or captured

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,142 @@
...
+    fun storeUserIdIfNeeded(customerInfo: CustomerInfo) {
+        if (!identityManager.currentUserIsAnonymous() || customerInfo.allPurchasedProductIds.isEmpty()) return
+        ioScope.launch {
+            val blockstoreData = try {
+                getBlockstoreData()
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                errorLog(e) { "Failed to retrieve Block store data. Will not store userId. Error: ${e.message}" }
+                return@launch
+            }
+            val currentUserId = identityManager.currentAppUserID
```
vegaro: I think there's a race condition and it is possible that when the code reaches here, the user id is not anonymous anymore? I am not sure but I think so

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
tonidero: TBH, I don't think we need it... I might just change it to not return anything. This is the number of bytes stored in Block store if any, but don't have a use case for it right now.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
tonidero: Hmm you're totally right... chances are small, since this happens on app foreground and when completing a purchase successfully... But still we should handle that, great catch!

tonidero: I believe it should be addressed in https://github.com/RevenueCat/purchases-android/pull/2595/commits/1a98c10a91ef691b3466b9149bfa97f23ee79fb8

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
tonidero: Removed in https://github.com/RevenueCat/purchases-android/pull/2595/commits/1a98c10a91ef691b3466b9149bfa97f23ee79fb8

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
JayShortway: >avoid clearing the user id from block store if the new user id is anonymous

I think that makes sense! 

>I guess this could still cause some issues, if they logIn with a non-anonymous user AFTER logging in with an anonymous user...

A restore would recover from this I guess? If so, I'd say that's acceptable.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,161 @@
...
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+internal class BlockstoreHelper
+@OptIn(ExperimentalCoroutinesApi::class)
+constructor(
+    applicationContext: Context,
+    private val identityManager: IdentityManager,
+    private val blockstoreClient: BlockstoreClient = Blockstore.getClient(applicationContext),
+    private val ioScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO.limitedParallelism(1)),
```
JayShortway, line 30: Is the limited parallelism to avoid concurrency issues? If so, I guess we could still technically have 2 instances of `BlockStoreHelper`. Not a huge problem probably.

tonidero: I think we should only be able to have one right? Since it's built once and attached to the purchases orchestrator if I'm correct.

tonidero: And yeah, this is to solve a possible race condition of a user logging in after we start storing the user id in blockstore, and it executing the clearing of blockstore before actually storing the user id, ending up with a user ID stored that has a non-anonymous alias. It's very unlikely but I thought if we execute these operations sequentially, this shouldn't be an issue.

JayShortway: Yea true! But there's nothing enforcing this to be a singleton (we could make the mistake of constructing `BlockStoreHelper()` in some other place), but I think that's totally fine.

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,161 @@
...
+                return@launch
+            }
+            val blockstoreUserId = blockstoreData[BLOCKSTORE_USER_ID_KEY]?.bytes?.let { String(it) }
+            if (blockstoreUserId == null || blockstoreUserId == currentUserId) {
+                callCompletion()
+                return@launch
+            }
+            try {
+                debugLog { "Aliasing Blockstore user ID: $blockstoreUserId with current UserID" }
+                identityManager.aliasCurrentUserIdTo(
```
vegaro, line 88: this made me think that this flow will only work for devs with the default or legacy transfer behavior https://www.revenuecat.com/docs/projects/restore-behavior#transfer-to-new-app-user-id. **Transfer if there are no active subscriptions** won't work because the non-consumable would be active in the previous old customer

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
```diff
@@ -0,0 +1,161 @@
...
+            .setRetrieveAll(true)
+            .build()
+        return suspendCoroutine { cont ->
+            blockstoreClient.retrieveBytes(retrieveRequest)
+                .addOnSuccessListener { cont.resume(it.blockstoreDataMap) }
+                .addOnFailureListener { cont.resumeWithException(it) }
+        }
+    }
+
+    // Stores the given userId in the Blockstore and returns the number of bytes stored.
```
vegaro: remember to update this doc, since it doesn't return anything anymore

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/blockstore/BlockstoreHelper.kt
tonidero: Hmm but this doesn't transfer the purchases but aliases the users together right? I believe that should work? It's true this may create aliases for users that wouldn't be created before, but I think the number should be small in general?

vegaro: ahh right, `/alias` doesn't receive any purchases so it just aliases users. You're correct. I had assumed it behave the same as `/receipts` but it won't since it doesn't get any purchase

PR #2604, opened 2025-08-15
Fixes price formatting discrepancies on Paywalls for `{{ product.price_per_[day|week|month|year] }}`

JayShortway:
## Bug
We've had reports of formatting discrepancies between calculated prices (e.g. monthly price calculated from a yearly subscription) and the price that comes directly from the store (the yearly price of a yearly subscription). This store price is shown on the paywall (`{{ product.price }}`) and on the Play Billing purchase sheet.

## Cause
This is caused by the fact that calculated prices use the device locale to format, and the store uses the storefront country.

## Fix
The fix implemented in this PR is to use the storefront country to format all prices. The difficulty is in the fact that the storefront country is only a country, not a complete locale. It's missing a language. To solve that, we check all available device locales for the storefront country, and pick the best one.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -143,8 +146,41 @@ internal sealed interface PaywallState {
...
+
+                    // We pick the one with the same language as the device if available. If not, we just pick the
+                    // first. If the list is empty, we build our own Locale.
+                    val javaLocale = availableStorefrontCountryLocalesByLanguage[deviceLanguageCode]
+                        ?: availableStorefrontCountryLocalesByLanguage.values.firstOrNull()
+                        ?: Locale.forLanguageTag("$deviceLanguageCode-${storefrontCountryCode.uppercase()}")
+
+                    javaLocale.toComposeLocale()
+                }
+            }
```
JayShortway: This is the fix.

rickvdl: I remember from a past project that there are some exceptions where the locale is not just `language-country`, but may also contain a `-variant`. 

I'm not sure if it's super relevant since the fallback might be good enough, but for example `zh-Hans (Chinese in Simplified script)` would be one of them.

rickvdl:
I'll let someone with a bit more context on this approve the PR :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -143,8 +146,41 @@ internal sealed interface PaywallState {
...
+                    val availableStorefrontCountryLocalesByLanguage: Map<String, Locale> =
+                        availableDeviceLocalesByCountry[storefrontCountryCode.lowercase()]
+                            ?.associateBy { it.language.lowercase() }
+                            ?: emptyMap()
+
+                    // We pick the one with the same language as the device if available. If not, we just pick the
+                    // first. If the list is empty, we build our own Locale.
+                    val javaLocale = availableStorefrontCountryLocalesByLanguage[deviceLanguageCode]
+                        ?: availableStorefrontCountryLocalesByLanguage.values.firstOrNull()
+                        ?: Locale.forLanguageTag("$deviceLanguageCode-${storefrontCountryCode.uppercase()}")
```
tonidero: From my tests on a kotlin playground, if the locale is wrongly formatted here it would still create the locale... what would happen in that case? Just wondering if we should add extra validation here to make sure we build a valid locale or fallback to the device locale instead...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
JayShortway: You're right! I was thinking this is good enough for a fallback, but we can do better. We're now using the entire device locale as a last resort, but only adjusting the country to the storefront one: https://github.com/RevenueCat/purchases-android/pull/2604/commits/feb641ea77b9aa9213e16dc02b3c527b84ecafac.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
JayShortway: Good point! I changed it so we're using the device locale and only adjust the country to the storefront one: https://github.com/RevenueCat/purchases-android/pull/2604/commits/feb641ea77b9aa9213e16dc02b3c527b84ecafac.

PR #2605, opened 2025-08-15
Always uses the provided locale when formatting a localized price for Paywall variables

JayShortway:
**Note:** this PR stacks on top of https://github.com/RevenueCat/purchases-android/pull/2604.

## Description
When working on https://github.com/RevenueCat/purchases-android/pull/2604, I noticed that `Price.localized()` wasn't using the provided locale at all if the decimals didn't need to be truncated, which is counterintuitive. This PR changes that, and also makes `VariableDataProvider` use it consistently.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
@@ -18,6 +18,7 @@ import com.revenuecat.purchases.UiConfig.VariableConfig
 import com.revenuecat.purchases.paywalls.components.common.LocaleId
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.getBestMatch
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toComposeLocale
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toJavaLocale
 import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toLocaleId
 import com.revenuecat.purchases.ui.revenuecatui.components.properties.BackgroundStyles
 import com.revenuecat.purchases.ui.revenuecatui.components.style.ComponentStyle
@@ -172,10 +173,13 @@ internal sealed interface PaywallState {
                             ?: emptyMap()
 
                     // We pick the one with the same language as the device if available. If not, we just pick the
-                    // first. If the list is empty, we build our own Locale.
+                    // first. If the list is empty, we use the device locale with the storefront country.
                     val javaLocale = availableStorefrontCountryLocalesByLanguage[deviceLanguageCode]
                         ?: availableStorefrontCountryLocalesByLanguage.values.firstOrNull()
-                        ?: Locale.forLanguageTag("$deviceLanguageCode-${storefrontCountryCode.uppercase()}")
+                        ?: Locale.Builder()
+                            .setLocale(locale.toJavaLocale())
+                            .setRegion(storefrontCountryCode.uppercase())
+                            .build()
 
                     javaLocale.toComposeLocale()
                 }


purchases/src/main/kotlin/com/revenuecat/purchases/models/TestStoreProduct.kt
```diff
@@ -66,7 +66,7 @@ class TestStoreProduct(
                 billingPeriod = it,
                 recurrenceMode = RecurrenceMode.FINITE_RECURRING,
                 billingCycleCount = 1,
-                price = Price(amountMicros = 0, currencyCode = price.currencyCode, formatted = "Free"),
+                price = Price(amountMicros = 0, currencyCode = price.currencyCode, formatted = "$0.00"),
```
JayShortway, line 69: This is the only thing I'm not entirely sure of. Alternatively we could special-case the first pricing phase doing something like: 

```kotlin
if (price.amountMicros == 0L) price.formatted
else price.localized()
```

That would avoid reformatting the price in case it's free.

Do we know whether the Play Store returns "free" for free trials? If so, my concern is the scenario where the storefront country is wildly different from the device locale. If the former is US and the latter is Spanish, this would result in a Spanish sentence suddenly containing the English word "free". 

If the Play Store doesn't return "free", I propose we don't special-case this scenario and just keep formatting the price ourselves.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PreviewOfferingParser.kt
```diff
@@ -30,7 +30,7 @@ private class PreviewOfferingParser : OfferingParser() {
                 id = "com.revenuecat.lifetime_product",
                 name = "Lifetime",
                 title = "Lifetime (App name)",
-                price = Price(amountMicros = 1_000_000_000, currencyCode = "USD", formatted = "$1,000"),
+                price = Price(amountMicros = 1_000_000_000, currencyCode = "USD", formatted = "$1,000.00"),
```
JayShortway: I think this was an oversight? `en-US` formats prices with 2 decimals.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
```diff
@@ -9,11 +9,11 @@ import java.text.NumberFormat
 import java.util.Currency
 import java.util.Locale
 
-internal object PriceFactory {
+@InternalRevenueCatAPI
+object PriceFactory {
```
JayShortway, line 13: This is now being used from `revenuecatui` (in `Price.localized()`).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -233,25 +206,3 @@ private fun Period.normalizedMonths(): Period {
...
- * For example $3.
- */
-internal fun Price.getTruncatedFormatted(locale: Locale = Locale.getDefault()): String {
-    val numberFormat = NumberFormat.getCurrencyInstance(locale).apply {
-        currency = Currency.getInstance(currencyCode)
-        maximumFractionDigits = 0
-    }
-    val amount = amountMicros / MICRO_MULTIPLIER
-    return numberFormat.format(amount)
-}
```
JayShortway, line 294: These were moved unchanged to `PriceExtensions.kt`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PriceExtensions.kt
```diff
@@ -24,3 +28,23 @@ internal fun Price.localizedPerPeriod(period: Period, locale: Locale, showZeroDe
...
+    val amount = amountMicros / MICRO_MULTIPLIER
+    return numberFormat.format(amount)
+}
+
+@SuppressWarnings("MagicNumber")
+private fun Price.endsIn00Cents(): Boolean {
+    val normalPrice = amountMicros / MICRO_MULTIPLIER
+    val roundedPrice = (normalPrice * 100).roundToLong() / 100.0
+    return roundedPrice * 100 % 100 == 0.0
+}
```
JayShortway, line 50: These were moved unchanged from `VariableDataProvider.kt`.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
```diff
@@ -831,8 +831,9 @@ class TextComponentViewTests {
         val countryWithoutDecimals = "MX"
         val textKey = LocalizationKey("key_selected")
         val textWithPriceVariable = LocalizationData.Text("Price: {{ product.price }}")
-        val expectedTextWithDecimals = "Price: \$ 2.00"
-        val expectedTextWithoutDecimals = "Price: MX\$1"
+        val expectedTextNullStorefront = "Price: $2.00"
+        val expectedTextNlStorefront = "Price: US$ 2,00"
+        val expectedTextMxStorefront = "Price: $1"
```
JayShortway: There are a bunch of changes in expected values in tests in this PR. I think the new values are actually more correct, but please double check. To illustrate: 
- When formatting USD, `NL` would 
    - specify which `$` we're talking about, hence `US$`
    - use a comma as decimal separator
- When formatting MXN, `MX` would
    - omit `MX`, as that's the default (symbol is also `$`)

purchases/src/main/kotlin/com/revenuecat/purchases/models/TestStoreProduct.kt
tonidero: Google does return `Free` here for a free trial, and if I change the device locale to Spanish, it changes to `Gratis`. So I guess in this case, google uses the device locale right?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PreviewOfferingParser.kt
tonidero: Hmm but if it doesn't have any decimals, I think it makes more sense to not add those decimals IMO? We also have some related logic to get the "truncated" version with `getTruncatedFormatted`

PR #2478, opened 2025-07-02
Improve Composable stabilities

skydoves:
Improve Composable stabilities by adding stability annotations to states, functions, and classes.

* **Added `@Immutable`** when all public properties of a class are completely immutable typically for Kotlin `data` classes representing domain models where the state never changes after creation.
* **Added `@Stable`** for mutable types or interfaces with multiple implementations, as long as they provide consistent outputs for the same inputs, commonly used for UI state holders to enable optimized recomposition.

### Differences

Those below composable functions have been "Skippable" now from "Unskippable".

|   Unnamed: 0 | package                                                                           | name                         |
|-------------:|:----------------------------------------------------------------------------------|:-----------------------------|
|            1 | com.revenuecat.purchases.ui.revenuecatui.LoadedPaywall                            | LoadedPaywall                |
|            2 | com.revenuecat.purchases.ui.revenuecatui.TemplatePaywall                          | TemplatePaywall              |
|            9 | com.revenuecat.purchases.ui.revenuecatui.LoadingPaywall                           | LoadingPaywall               |
|           13 | com.revenuecat.purchases.ui.revenuecatui.PaywallDialog                            | PaywallDialog                |
|          199 | com.revenuecat.purchases.ui.revenuecatui.composables.ConsistentPackageContentView | ConsistentPackageContentView |
|          233 | com.revenuecat.purchases.ui.revenuecatui.composables.OfferDetails                 | OfferDetails                 |
|          241 | com.revenuecat.purchases.ui.revenuecatui.composables.PurchaseButton               | PurchaseButton               |
|          329 | com.revenuecat.purchases.ui.revenuecatui.templates.Template1                      | Template1                    |
|          330 | com.revenuecat.purchases.ui.revenuecatui.templates.Template1MainContent           | Template1MainContent         |
|          346 | com.revenuecat.purchases.ui.revenuecatui.templates.Template2                      | Template2                    |
|          347 | com.revenuecat.purchases.ui.revenuecatui.templates.Template2PortraitContent       | Template2PortraitContent     |
|          348 | com.revenuecat.purchases.ui.revenuecatui.templates.Template2LandscapeContent      | Template2LandscapeContent    |
|          349 | com.revenuecat.purchases.ui.revenuecatui.templates.IconImage                      | IconImage                    |
|          350 | com.revenuecat.purchases.ui.revenuecatui.templates.Title                          | Title                        |
[...]

PR #2608, opened 2025-08-18
Introduce runtime annotations library and add stability annotations for increasing UI performances

skydoves:
### Motivation

This is a follow-up task from [#2478](https://github.com/RevenueCat/purchases-android/pull/2478)

Introduce runtime annotations library (which is completely [JVM-based annotations](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/runtime/runtime-annotation/src/commonMain/kotlin/androidx/compose/runtime/annotation/compose-runtime-annotation-documentation.md)) and add stability annotations on the purchase core models.

~Unfortunately, the AndroidX runtime annotations library requires a compile SDK version above 34. So for now, I’ve switched to using [compose-stable-marker](https://github.com/skydoves/compose-stable-marker) instead. It offers the exact same functionality, was created two years earlier than the official library, and is already widely adopted by many libraries and SDKs.~


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -32,6 +33,7 @@ interface PaywallDisplayCallback {
  * This can be instantiated with an [ActivityResultCaller] instance
  * like a [ComponentActivity] or a [Fragment].
  */
+@Immutable
```
tonidero: Hmm is this needed? I believe `PaywallActivityLauncher` shouldn't really be used from compose... Not sure if it might be on Paywall tester so we can test launching an activity... But still, seems like for most cases this won't happen.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallResult.kt
```diff
@@ -9,6 +11,7 @@ import kotlinx.parcelize.Parcelize
 /**
  * Result of the paywall activity.
  */
+@Stable
```
tonidero: Same here, I think `PaywallResult` should only be generated in non-compose code normally, so I believe it might not be needed? (Again, it might be used in paywall tester, but I don't think that's representative of a real scenario)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
skydoves: Nice catch! I just removed some stability annotations that are not used by Compose UI on [c57f611](https://github.com/RevenueCat/purchases-android/pull/2608/commits/c57f611d9b95230bcdf2a0d7cc377f96547ca6ed)

test-apps/testpurchasesandroidcompatibility/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
 
 android {
     namespace = "com.revenuecat.testpurchasesandroidcompatibility"
-    compileSdk = 33 // Keeping it at 33 to test compatibility with purchases-android
+    compileSdk = 34 // Keeping it at 34 to test compatibility with purchases-android
```
tonidero, line 8: Hmm yeah... From what I remember, we added this because we added a dependency that made is incompatible with apps targeting the same SDK version as us... But I think it should be fine to test against 34 now 👍

tonidero (approved):
This looks great! Nice improvements! 🙌

test-apps/testpurchasesandroidcompatibility/build.gradle.kts
skydoves: Thank you for the details, Toni! ❤️

PR #2609, opened 2025-08-18
Update to Dokka 2.0.0

tonidero:
### Description
This updates to Dokka 2.0 which fixes some issues with the bump of gradle version to 9.0.0. I did appreciate some small changes on the designs, most problematic one might be that the order of rendering seems to be different... but the docs themselves seem to be mostly the same. Links seem to work fine and didn't appreciate any issues.


gradle/libs.versions.toml
```diff
@@ -23,7 +23,7 @@ datastorePreference = "1.0.0"
 dependencyGraph = "0.9"
 detekt = "1.23.8"
 detektRulesCompose = "0.4.10"
-dokka = "1.8.10"
+dokka = "2.0.0"
```
JayShortway, line 26: Just double checking, does our `DokkaHideInternal` plugin still work?

tonidero: Yup! At least, can't see any internal APIs in the generated docs.

PR #2610, opened 2025-08-18
Add APIs for hybrid SDKs to set presentedOfferingContext

tonidero:
### Description
This adds some new internal APIs to be used by our hybrid SDKs that allow to pass a `PresentedOfferingContext` when launching paywalls and passing a specific offering. 

- [ ] Actually use presented offering context data


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -13,22 +13,22 @@ internal sealed class OfferingSelection {
     data class OfferingType(val offeringType: Offering) : OfferingSelection()
 
     @Immutable
-    data class OfferingId(val offeringId: String) : OfferingSelection()
+    data class OfferingInfo(val offeringInfo: OfferingPresentationInfo) : OfferingSelection()
```
JayShortway: Do we need a new class (`OfferingPresentationInfo`) here? Can we not add the `PresentedOfferingContext` parameter to `OfferingId`? Like so: 
```kotlin
data class OfferingId(
  val offeringId: String, 
  val presentedOfferingContext: PresentedOfferingContext?,
) : OfferingSelection()
```

Reasoning also being that it's not super clear (at first glance) how `setOfferingInfo()` is different from `setOffering()`. That could also remain `setOfferingId()` in this case: 
```kotlin
internal fun setOfferingId(
  offeringId: String?,
  presentedOfferingContext: PresentedOfferingContext?,
)
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -86,6 +94,15 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
      * @param edgeToEdge Whether to display the paywall in edge-to-edge mode.
      * Default is true for Android 15+, false otherwise.
      */
+    @Deprecated(
+        message = "Use launch with presented offering context instead",
```
JayShortway: Is the `PresentedOfferingContext` always available? Should we default to `null`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -103,15 +111,39 @@ class PaywallView : CompatComposeView {
...
+            )
         }
         paywallOptions = paywallOptions.copy(offeringSelection = offeringSelection)
     }
 
+    @InternalRevenueCatAPI
+    fun setOfferingInfo(
+        offeringId: String,
+        presentedOfferingContext: PresentedOfferingContext?,
+    ) {
```
JayShortway: I'd argue that we can keep calling this `setOfferingId()`. We just need 1 extra argument if you give us the offering ID.

JayShortway (commented):
Thanks for taking care of this! Just a question on whether we need to introduce the new "OfferingInfo" concept for this.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
tonidero: For an offering, it should always be available yes. I had to make it nullable for backwards compatibility reasons, but otherwise, it would be good to assume that it should always be there.

JayShortway: Hmm, but what is the purpose of the "offering ID" API, if devs will always need to have a full `Offering` (because they now need the `PresentedOfferingContext`)? That makes the "offering ID" API obsolete no, or am I misunderstanding something?

tonidero: The main reason is that it's tricky to create a full `Offering` native object from the hybrids, since it has a lot of data and even some non-parcelable objects like the Store's product entities. So we can't easily recreate the offerings from the data in the hybrids, that's the reason we were only passing the offering id. 

This PR adds the data we were missing for proper traceability purposes. Otherwise, any placement/targeting data would be lost.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
tonidero: Yeah... we could add it as a new parameter, the main reason I joined them into a new class was because both types of data should move together through all the plumbing, and creating a new entity ensured that better than separating into different parameters and moving both around. That could lead to easily forgetting passing/setting one of them IMO. It's true it's a bit confusing with the other `setOffering` APIs though... not sure if we could come up with a better naming...

tonidero:
In this PR I wasn't using the newly provided presented offering context yet. So I added that in a separate PR: https://github.com/RevenueCat/purchases-android/pull/2612

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
JayShortway: What about `IdAndPresentedOfferingContext`? 😅 (It's already namespaced to `OfferingSelection`, so maybe we can get away with omitting "Offering" in the name.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
JayShortway: Right got it, but this is a public API too. Before this PR, it had its utility. You could launch a paywall with just an Offering ID, without having to get Offerings first. But now the utility of this public API is gone, as devs will need to get Offerings to be able to pass a `PresentedOfferingContext`. 

I was gonna suggest: 
1. we deprecate `launch(offeringIdentifier: String)` in favor of `launch(offering: Offering)` instead, and
2. we make `launch(offeringIdentifier: String, presentedOfferingContext: PresentedOfferingContext)` an `@InternalRevenueCatAPI`. 

But I see you already did 2 😄 So in that case I think we should still do 1, as the current deprecation message encourages the use of an internal API.

tonidero: Right... our docs did say not to use this method, but yeah, we probably should have made it internal from the beginning... In any case, I do agre that the ReplaceWith should point to the method taking an offering instead of the internal API. Good catch!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
tonidero: Hmm it's a bit of a longer name... certainly more explicit, but not sure if we win much that much clarity TBH... But I guess it could improve on the readability for the setOfferingId method... So will do the rename.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -116,9 +123,10 @@ class PaywallOptions internal constructor(
                 ?: OfferingSelection.None
         }
 
-        internal fun setOfferingId(offeringId: String?) = apply {
-            this.offeringSelection = offeringId?.let { OfferingSelection.OfferingId(it) }
-                ?: OfferingSelection.None
+        internal fun setOfferingIdAndPresentedOfferingContext(
+            offeringInfo: OfferingSelection.IdAndPresentedOfferingContext?,
```
JayShortway: nit:
```suggestion
            idAndPresentedOfferingContext: OfferingSelection.IdAndPresentedOfferingContext?,
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/OriginalTemplatePaywallFooterView.kt
```diff
@@ -130,15 +132,37 @@ open class OriginalTemplatePaywallFooterView : FrameLayout {
     /**
      * Sets the offering id to be used to display the Paywall. If not set, the default one will be used.
      */
+    @Deprecated(
+        "Use setOfferingInfo instead.",
+        ReplaceWith(
+            "setOfferingInfo(offeringId, presentedOfferingContext)",
+        ),
+    )
```
JayShortway: `setOfferingInfo` does not exist anymore, and `setOfferingIdAndPresentedOfferingContext` is internal. Should we add a `setOffering` function here? If not, we should remove the `ReplaceWith` from the deprecation message.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -146,7 +157,13 @@ class PaywallView : CompatComposeView {
     private fun parseAttributes(context: Context, attrs: AttributeSet?) {
         val (offeringId, fontProvider, shouldDisplayDismissButton, _) =
             PaywallViewAttributesReader.parseAttributes(context, attrs, R.styleable.PaywallView) ?: return
-        setOfferingId(offeringId)
+        this.initialOfferingInfo = offeringId?.let {
+            OfferingSelection.IdAndPresentedOfferingContext(
+                offeringId = offeringId,
+                // WIP: We do not support presentedOfferingContext when using the view in XML layouts.
+                presentedOfferingContext = null,
```
JayShortway, line 164: Not for this PR, just thinking out loud: what if we support passing a placement in XML layouts?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -101,13 +107,18 @@ class PaywallView : CompatComposeView {
     }
 
     /**
-     * Sets the offering id to be used to display the Paywall. If not set, the default one will be used.
+     * Sets the offering id and presented offering context to be used to display the Paywall.
+     * If not set, the default one will be used.
      */
-    fun setOfferingId(offeringId: String?) {
+    @JvmOverloads
+    fun setOfferingId(offeringId: String?, presentedOfferingContext: PresentedOfferingContext? = null) {
```
JayShortway, line 114: Do we need a `setOffering(Offering)` too?

tonidero: I believe we needed this for RN if I remember correctly, that required us to set the offering AFTER actually creating the paywall, which was pretty convoluted. We didn't need the `setOffering` API until now, and in general, I think it's better to just pass it in the constructor.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
tonidero: Yeah we could... TBH, I'm not sure how many people are using paywalls through XMLs... but it's certainly something we could add if needed.

PR #2612, opened 2025-08-19
Override presented offering context paywalls without offering

tonidero:
### Description
Based on #2610 

This actually makes use of the presented offering context to pass a modified offering to the paywall. This data will be sent as part of the post receipt when a purchase is made.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -368,8 +368,7 @@ internal class PaywallViewModelImpl(
                     is OfferingSelection.IdAndPresentedOfferingContext -> {
                         val offerings = purchases.awaitOfferings()
                         val presentedOfferingContext = offeringSelection.presentedOfferingContext
-                        val offering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
-                            ?: offerings.current
+                        val offering = offerings[offeringSelection.offeringId] ?: offerings.current
                         presentedOfferingContext?.let {
                             offering?.copy(presentedOfferingContext)
                         } ?: offering


purchases/api-defauts.txt
```diff
@@ -1299,9 +1299,12 @@ package com.revenuecat.purchases.models {
...
-    ctor public TestStoreProduct(String id, String name, String title, String description, com.revenuecat.purchases.models.Price price, com.revenuecat.purchases.models.Period? period, optional com.revenuecat.purchases.models.PricingPhase? freeTrialPricingPhase, optional com.revenuecat.purchases.models.PricingPhase? introPricePricingPhase);
+    ctor public TestStoreProduct(String id, String name, String title, String description, com.revenuecat.purchases.models.Price price, optional com.revenuecat.purchases.models.Period? period, optional com.revenuecat.purchases.models.PricingPhase? freeTrialPricingPhase);
+    ctor public TestStoreProduct(String id, String name, String title, String description, com.revenuecat.purchases.models.Price price, optional com.revenuecat.purchases.models.Period? period, optional com.revenuecat.purchases.models.PricingPhase? freeTrialPricingPhase, optional com.revenuecat.purchases.models.PricingPhase? introPricePricingPhase);
```
JayShortway, line 1306: The deleted line lacks the `optional` keyword for the `com.revenuecat.purchases.models.Period? period` parameter. The added lines all have the `optional` keyword for `period`. Any idea why? 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -362,11 +363,16 @@ internal class PaywallViewModelImpl(
...
                 if (currentOffering == null) {
                     val offerings = purchases.awaitOfferings()
-                    currentOffering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
+                    val presentedOfferingContext = (options.offeringSelection as? OfferingSelection.OfferingInfo)
+                        ?.offeringInfo?.presentedOfferingContext
+                    val offering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
                         ?: offerings.current
+                    currentOffering = presentedOfferingContext?.let {
+                        offering?.copy(it)
+                    } ?: offering
```
JayShortway: With this added logic I think it may be simpler to just do a `when` over `options.offeringSelection`, and split out both paths separately? (Avoiding the nullable `offeringSelection.offering`.)

```kotlin
val currentOffering = when (options.offeringSelection) {
  is OfferingSelection.Offering -> {}
  is OfferingSelection.OfferingInfo -> {}
}
```

purchases/api-defauts.txt
tonidero: I removed one of the constructors for the TestStoreProduct that made it non-optional: https://github.com/RevenueCat/purchases-android/pull/2612/files#diff-18a1e4e397d6d9a1ac93514cf38e78a9b038cb703d536ed5162cc44c0f2d848aL23-L39. I do think that shouldn't really matter, but I guess the path it follows now in this case would go to the top-level constructor instead of that constructor. Metalava probably generated symbols for that constructor, but I think the API itself should remain compatible?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2612/commits/6fc390445b9e06b3c8a81a0d8b43ffecd0d61773

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -362,11 +363,21 @@ internal class PaywallViewModelImpl(
...
-                    val offerings = purchases.awaitOfferings()
-                    currentOffering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
-                        ?: offerings.current
+                var currentOffering: Offering? = when (val offeringSelection = options.offeringSelection) {
+                    is OfferingSelection.OfferingType -> offeringSelection.offeringType
+                    is OfferingSelection.IdAndPresentedOfferingContext -> {
+                        val offerings = purchases.awaitOfferings()
+                        val presentedOfferingContext = offeringSelection.presentedOfferingContext
+                        val offering = options.offeringSelection.offeringIdentifier?.let { offerings[it] }
+                            ?: offerings.current
```
JayShortway: Using `offeringSelection` would allow it to smart cast, avoiding the need to handle a nullable `offeringIdentifier`.
```suggestion
                        val offering = offerings[offeringSelection.offeringIdentifier] ?: offerings.current
```

purchases/api-defauts.txt
JayShortway: Thanks! Yea I think so too 🤔

JayShortway (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: I think that doesn't work becuase `offeringSelection.offeringIdentifier` is nullable (Because it can be `OfferingSelection.None`).

JayShortway: Sorry I meant `offeringId`. This should work no? 

```kotlin
val offering = offerings[offeringSelection.offeringId] ?: offerings.current
```

tonidero: Oh yes yes, that works! Will do the change. Thanks!

PR #2618, opened 2025-08-21
Revert dokka 2 and gradle 9 update

tonidero:
### Description
We've been having some issues with Dokka 2. Our dokka-hide-internal plugin tests are not passing with not a great description. We've decided to revert the update for now. Unfortunately, that also means we need to revert the Gradle 9 update, to keep it compatible with Dokka 1. 

This PR also changes to run the dokka-hide-internal plugin tests on all branches so we don't miss this again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -918,7 +918,6 @@ workflows:
           requires:
             - prepare-tests
       - test_dokka_hide_internal:
-          <<: *release-branches
           requires:
             - prepare-tests
       - run-backend-integration-tests:


.circleci/config.yml
```diff
@@ -918,7 +918,6 @@ workflows:
           requires:
             - prepare-tests
       - test_dokka_hide_internal:
-          <<: *release-branches
```
tonidero, line 921: This should be fine to run on all branches and avoid us from finding issues later, like in this case.

PR #2620, opened 2025-08-21
Add preferred UI locale override for RevenueCat UI components

joshdholtz:
## Summary

This PR adds the ability to override the system locale for RevenueCat UI components (Paywalls and Customer Center), allowing developers to display the UI in a specific language regardless of the system locale setting. This implements the same functionality that was added to the iOS SDK in https://github.com/RevenueCat/purchases-ios/pull/5292.

### Configuration API
- Add `preferredUILocaleOverride` property to `PurchasesConfiguration`
- Add `preferredUILocaleOverride(String?)` builder method with comprehensive documentation

### Runtime API  
- Add `overridePreferredUILocale(String?): Boolean` method to `Purchases` class
- Add public getter `preferredUILocaleOverride` property
- **Automatic cache management**: Cache clearing happens automatically when locale changes
- **Rate limiting**: Built-in rate limiting (10 calls per 60 seconds) prevents excessive API calls
- Both APIs available for default and custom entitlement computation flavors

### UI Components Integration
- Update `PaywallViewModel` and `PaywallState` to use preferred locale override
- Add reflection-based approach to access preferred locale from UI components
- Support both "es-ES" and "es_ES" locale format patterns
- Proper locale matching logic that prioritizes preferred override
- Graceful fallback to system default for invalid locales with proper logging

### HTTP Request Integration
- Update `LocaleProvider` to include preferred locale override in API requests
- Preferred locale is sent first in Accept-Language header when set
- Ensures both UI rendering and API responses respect the same locale preference

### Paywall Tester Enhancements
- Add comprehensive locale selection tab with 15+ predefined locales
- Add search/filter functionality to offerings tab
- Support custom locale input with validation
- Optional preferred locale configuration via `Constants.kt`
- Real-time feedback on cache clearing status with rate limit handling

### Configuration time:
```kotlin
val configuration = PurchasesConfiguration.Builder(context, apiKey)
    .preferredUILocaleOverride("es-ES")  // or "es_ES" - both formats supported
    .build()
```

### Runtime override:
```kotlin
// Set preferred locale (automatically clears cache if changed)
val cacheCleared = Purchases.sharedInstance.overridePreferredUILocale("de-DE")
if (cacheCleared) {
    // Cache was cleared and offerings will be refetched with new locale
}

// Revert to system default
Purchases.sharedInstance.overridePreferredUILocale(null)
```

### Rate Limiting
- Cache clearing operations are rate limited to prevent excessive network requests
- Uses the same `RateLimiter` utility as subscriber attributes (10 calls per 60 seconds)
- Locale setting is immediate, cache clearing is rate limited separately

### Locale Resolution Priority
1. **Preferred override** (if set via configuration or runtime API)
2. **System default locales** (from `LocaleListCompat.getDefault()`)
3. **Paywall default locale** (fallback for UI components)

### Thread Safety
- All operations are properly synchronized
- Reflection-based access includes proper exception handling
- State management through PurchasesOrchestrator ensures consistency

## Test plan

- [x] All existing unit tests pass
- [x] Code compiles successfully for all modules  
- [x] Lint/detekt checks pass with no new violations
- [x] API compatibility checks pass (metalava)
- [x] Graceful error handling for invalid locale strings
- [x] Thread-safe access to shared configuration state
- [x] Rate limiting functionality verified
- [x] Paywall tester UI enhancements tested

## Testing Notes

- Manual testing shows paywalls now correctly display in the preferred locale
- HTTP requests include the preferred locale in Accept-Language headers
- Cache clearing triggers background refetch of offerings with new locale
- Rate limiting prevents excessive API calls when rapidly changing locales
- Paywall tester provides comprehensive UI for testing different locales

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -857,8 +857,8 @@ class Purchases internal constructor(
      *
      * @param localeString The locale string (e.g., "es-ES", "en-US") or null to use system default
      */
-    fun overridePreferredUILocale(localeString: String?) {
-        purchasesOrchestrator.overridePreferredUILocale(localeString)
+    fun overridePreferredUILocale(localeString: String?): Boolean {
+        return purchasesOrchestrator.overridePreferredUILocale(localeString)
     }
 
     /**
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -462,19 +462,21 @@ internal class PurchasesOrchestrator(
      * importing RevenueCatUI in your project. The locale override will take effect the next time
      * a paywall or customer center is displayed.
      */
-    fun overridePreferredUILocale(localeString: String?) {
+    fun overridePreferredUILocale(localeString: String?): Boolean {
         val previousLocale = _preferredUILocaleOverride
 
         if (previousLocale == localeString) {
-            return
+            debugLog { "Locale unchanged, no fresh fetch needed" }
+            return false
         }
 
         _preferredUILocaleOverride = localeString
         localeProvider.setPreferredLocaleOverride(localeString)
 
-        fetchOfferingsWithRateLimit { offerings, error ->
+        debugLog { "Locale changed, attempting to fetch fresh offerings" }
+        return fetchOfferingsWithRateLimit { offerings, error ->
             if (offerings != null) {
-                verboseLog { "Fresh offerings fetch completed successfully" }
+                debugLog { "Fresh offerings fetch completed successfully" }
             } else {
                 debugLog { "Fresh offerings fetch failed: ${error?.message}" }
             }
--- a/purchases/api-entitlement.txt
+++ b/purchases/api-entitlement.txt
@@ -340,7 +340,6 @@ package com.revenuecat.purchases {
     method public void getStorefrontCountryCode(com.revenuecat.purchases.interfaces.GetStorefrontCallback callback);
     method @kotlin.jvm.Synchronized public com.revenuecat.purchases.interfaces.UpdatedCustomerInfoListener? getUpdatedCustomerInfoListener();
     method public static boolean isConfigured();
-    method public void overridePreferredUILocale(String? localeString);
     method public void purchase(com.revenuecat.purchases.PurchaseParams purchaseParams, com.revenuecat.purchases.interfaces.PurchaseCallback callback);
     method public void removeUpdatedCustomerInfoListener();
     method public void restorePurchases(com.revenuecat.purchases.interfaces.ReceiveCustomerInfoCallback callback);


examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/locale/LocaleScreen.kt
```diff
@@ -0,0 +1,211 @@
...
+        )
+    }
+
+    Column(
+        modifier = modifier.fillMaxSize().padding(16.dp),
+        verticalArrangement = Arrangement.spacedBy(8.dp),
+    ) {
+        Text(
+            text = "Preferred UI Locale Override",
+            fontSize = 20.sp,
```
JZDesign, line 79: Do we regularly use sp instead of [material theme](https://developer.android.com/develop/ui/compose/designsystems/material3#typography)?

```kotlin
fontSize = MaterialTheme.typography.titleMedium.fontSize,
fontStyle = MaterialTheme.typography.titleMedium.fontStyle,
```

joshdholtz: Honestly... no idea 🤷‍♂️

This is just our internal paywalls tester app and I told Claude to add this because I'm not qualified 😇 But I can go look at what we do on the other screens for consistency. But...

I would like for this app to get an overhaul at some point!

tonidero: Yeah... I would say we haven't been doing a great job keeping these into styles, specially for our test apps... NABD for our test apps IMO though.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -111,15 +115,58 @@ private fun OfferingsListScreen(
...
+    onSearchQueryChange: (String) -> Unit,
     modifier: Modifier = Modifier,
 ) {
     var dropdownExpandedOffering by remember { mutableStateOf<Offering?>(null) }
     var displayPaywallDialogOffering by remember { mutableStateOf<Offering?>(null) }
 
     val showDialog = remember { mutableStateOf(false) }
 
+    // Filter offerings based on search query
+    val filteredOfferings = remember(offeringsState.offerings, offeringsState.searchQuery) {
```
tonidero, line 127: Should we extract this to a separate PR?

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -68,6 +71,7 @@ fun OfferingsScreen(
             tappedOnNavigateToOfferingCondensedFooter = tappedOnOfferingCondensedFooter,
             tappedOnNavigateToOfferingByPlacement = tappedOnOfferingByPlacement,
             tappedOnReloadOfferings = { viewModel.refreshOfferings() },
+            onSearchQueryChange = { query -> viewModel.updateSearchQuery(query) },
```
tonidero, line 74: Amazing!! 🙌 

One thing though, should we extract this to a separate PR? NABD though.

purchases/src/customEntitlementComputation/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -127,6 +127,28 @@ class Purchases internal constructor(
         purchasesOrchestrator.close()
     }
 
+    /**
+     * The current preferred UI locale override, if any.
+     * This value is used by RevenueCat UI components like Paywalls and Customer Center.
+     */
+    val preferredUILocaleOverride: String?
```
tonidero: Just checking... do we want to offer these APIs in the CEC flavor? I think the idea until now it's been to keep it minimal. Also, since this is meant mostly for paywalls, more work would be needed to actually support CEC mode in Paywalls I think. So I think we should remove these APIs from here?

purchases/src/main/kotlin/com/revenuecat/purchases/common/LocaleProvider.kt
```diff
@@ -7,6 +7,29 @@ internal interface LocaleProvider {
 }
 
 internal class DefaultLocaleProvider : LocaleProvider {
+    private var orchestratorProvider: (() -> com.revenuecat.purchases.PurchasesOrchestrator?)? = null
```
tonidero: Hmm TBH, I'm not liking adding a dependency to the purchases orchestrator from the locale provider... Could we create a separate holder like `LocaleOverrideData` which can be passed here? If we make the setter in `Purchases` a static method that sets this data statically, this could be even easier.

purchases/src/main/kotlin/com/revenuecat/purchases/common/LocaleProvider.kt
```diff
@@ -7,6 +7,29 @@ internal interface LocaleProvider {
...
+
+    fun setOrchestratorProvider(provider: () -> com.revenuecat.purchases.PurchasesOrchestrator?) {
```
tonidero: Should import instead of using fully qualified names (a few occurrences of this in this PR)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -839,6 +840,61 @@ class Purchases internal constructor(
...
+     * Override the preferred UI locale for RevenueCat UI components at runtime.
+     * This affects both API requests and UI rendering.
+     *
+     * If the locale changes, this will automatically clear the offerings cache and trigger
+     * a background refetch to get paywall templates with the correct localizations.
+     *
+     * @param localeString The locale string (e.g., "es-ES", "en-US") or null to use system default
+     * @return true if cache was cleared and refetch triggered, false if locale unchanged or cache clear rate limited
+     */
+    fun overridePreferredUILocale(localeString: String?): Boolean {
```
tonidero, line 860: Would it make sense to make this a "static" method instead of tying it to the Purchases instance? That might make it easier to access later on... I guess it wouldn't be cleared at any point, but normally that doesn't happen either...

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -195,9 +195,16 @@ internal class PurchasesOrchestrator(
...
 
+    @SuppressWarnings("MagicNumber")
+    private val preferredLocaleOverrideRateLimiter = RateLimiter(2, 60.seconds)
+
     var storefrontCountryCode: String? = null
         private set
 
+    @get:Synchronized
+    @set:Synchronized
+    var preferredUILocaleOverride: String? = initialConfiguration.preferredUILocaleOverride
```
tonidero: Hmm I'm not sure we should modify the `initialConfiguration` when modifying the locale... I think this should be a separate place to become the source of truth of the locale to be used IMO.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -839,6 +840,61 @@ class Purchases internal constructor(
...
+     * Fetches fresh offerings from the API to get updated localizations.
+     * This is useful after changing the preferred locale to get paywall templates
+     * with the correct localizations.
+     *
+     * This method is rate limited to 2 calls per 60 seconds to prevent excessive
+     * network requests.
+     *
+     * @return true if fresh fetch was triggered, false if rate limited
+     */
+    private fun fetchOfferingsIfPossible(): Boolean {
```
tonidero: Hmm I would remove this method and call the purchasesOrchestrator directly and add the logs there if needed. The least logic we have in the facade the better.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -900,6 +907,39 @@ internal class PurchasesOrchestrator(
...
+    /**
+     * Fetches fresh offerings with rate limiting to prevent excessive network requests.
+     *
+     * @param callback Callback to handle the result
+     * @return true if fresh fetch was triggered, false if rate limited
+     */
+    internal fun fetchOfferingsWithRateLimit(callback: (Offerings?, PurchasesError?) -> Unit): Boolean {
+        return if (preferredLocaleOverrideRateLimiter.shouldProceed()) {
+            log(LogIntent.DEBUG) { "Fetching fresh offerings" }
+            getOfferings(
```
tonidero, line 960: A couple possible edge case of this... If our backend is down (very unlikely), or this takes a bit longer for any reason, the paywall might be displayed with the cached offerings, which could be in the wrong locale... Not sure how big of a deal this is it though...

We could clear the offerings cache before fetching it to at least make sure we don't use the wrong locale... though I guess in that case we might be delaying displaying the paywall, or might cause some errors if the request fails for any reason...

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -839,6 +840,61 @@ class Purchases internal constructor(
...
+     *
+     * @param localeString The locale string (e.g., "es-ES", "en-US") or null to use system default
+     * @return true if cache was cleared and refetch triggered, false if locale unchanged or cache clear rate limited
+     */
+    fun overridePreferredUILocale(localeString: String?): Boolean {
+        val previousLocale = purchasesOrchestrator.preferredUILocaleOverride
+        debugLog { "overridePreferredUILocale: changing from '$previousLocale' to '$localeString'" }
+        purchasesOrchestrator.preferredUILocaleOverride = localeString
+
+        return if (previousLocale != localeString) {
```
tonidero: I think it might be good to move this logic to the `PurchasesOrchestrator` as well.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
joshdholtz: I originally thought so but I also needed it for my testing on this one 😛 I didn't like that I was struggling to scroll to find my paywall to test 🙃

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
joshdholtz: @tonidero The iOS PR has it on the instance so _probably_ best if we match consistency here? 🤷‍♂️  But also... I totally see why static would work here as well 😇

tonidero: That's fine, we can keep it to the instance... In that case, I think we should just create some kind of holder for this information that can be access more easily later. Related to this comment: https://github.com/RevenueCat/purchases-android/pull/2620#discussion_r2309509579

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/locale/LocaleScreen.kt
JZDesign: Yea I agree about the test app, just getting my bearings.

.kotlin/errors/errors-1756481779029.log
```diff
@@ -0,0 +1,4 @@
+kotlin version: 2.0.21
```
JZDesign: Not sure you meant to commit this

JZDesign (commented):
Potentially dumb question—new to the android sdk—But can't we allow the user to pass in a custom context on application initialization that overrides the locale and then we don't have to make these broad sweeping changes?

They could do something like:

```kotlin

    fun cloneContextAndSetLocale(context: Context, languageCode: String) : Context {
        val locale = Locale(languageCode)
        Locale.setDefault(locale) // Set default for the app process

        val config = Configuration(context.resources.configuration)
        config.setLocale(locale) // Set for the configuration
        val newContext = context.createConfigurationContext(config) // Create a new context
        newContext.resources.updateConfiguration(config, newContext.resources.displayMetrics)
        return newContext
    }
```

Then give us the context. 

Or is that less feasible than what we're doing here?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -127,6 +127,7 @@ internal class PurchasesOrchestrator(
     private val dispatcher: Dispatcher,
     private val initialConfiguration: PurchasesConfiguration,
     private val fontLoader: FontLoader,
+    private val localeProvider: com.revenuecat.purchases.common.DefaultLocaleProvider,
```
tonidero: Should add the import instead of the fully qualified name :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -839,6 +839,29 @@ class Purchases internal constructor(
...
+     * Override the preferred UI locale for RevenueCat UI components at runtime.
+     * This affects both API requests and UI rendering.
+     *
+     * If the locale changes, this will automatically clear the offerings cache and trigger
+     * a background refetch to get paywall templates with the correct localizations.
+     *
+     * @param localeString The locale string (e.g., "es-ES", "en-US") or null to use system default
+     * @return true if cache was cleared and refetch triggered, false if locale unchanged or cache clear rate limited
+     */
+    fun overridePreferredUILocale(localeString: String?): Boolean {
```
tonidero, line 860: Hmm is there a need for this method to return a boolean? Feels like this should not return anything IMO. It would also simplify the code a bit. Wdyt?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -400,7 +415,27 @@ internal class PaywallViewModelImpl(
     }
 
     private fun getCurrentLocaleList(): LocaleListCompat {
-        return LocaleListCompat.getDefault()
+        val preferredLocale = purchases.preferredUILocaleOverride
+        Logger.d("PaywallViewModel getCurrentLocaleList: preferredLocale = $preferredLocale")
```
tonidero: I think many of these logs should be verbose instead of debug... Wdyt? I would maybe only leave one as debug when "changing" locales.

tonidero (commented):
Another point related to the context approach. To override the locale in a context, you usually need to provide an "activity" context, but we recommend developers to pass us their application context, which is trickier to override.

In any case, I have a few more comments but smaller, I think this is pretty close to ready!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -400,7 +409,22 @@ internal class PaywallViewModelImpl(
...
     private fun getCurrentLocaleList(): LocaleListCompat {
-        return LocaleListCompat.getDefault()
+        val preferredLocale = purchases.preferredUILocaleOverride ?: return LocaleListCompat.getDefault()
+
+        return try {
+            val locale = createLocaleFromString(preferredLocale)
+            val localeList = LocaleListCompat.create(locale)
+            localeList
+        } catch (@Suppress("SwallowedException") e: IllegalArgumentException) {
+            Logger.w("Invalid preferred locale format: $preferredLocale. Using system default.")
```
tonidero: Nitpick, might be worth bumping this log to error and pass the exception to the log, so we get the exact error?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -839,6 +839,28 @@ class Purchases internal constructor(
...
         }
 
+    /**
+     * The preferred UI locale override for RevenueCat UI components.
+     * This affects both API requests and UI rendering.
+     *
+     * @return The preferred UI locale override, or null if using system default
+     */
+    val preferredUILocaleOverride: String?
+        @Synchronized get() = purchasesOrchestrator.preferredUILocaleOverride
```
JZDesign, line 849: Is a synchronized getter necessary? Wouldn't we also need an atomic setter for this to matter?

joshdholtz: Pretty sure that `@Synchronized` covers everything we need here! At least according to research that I just did  🤷‍♂️ 

But maybe @tonidero can answer better? 😅

tonidero: Oh hmm I missed this but I agree with @JZDesign actually... I don't think this is doing much right now, other than disallowing simultaneous getter operations on different threads... But you could totally override the locale on a different thread and there would be no guarantees...

Now, having said that, it does seem very unlikely for these APIs to be called multiple times in different threads I would say... So I do think that it's probably unnecessary? And the one in `PurchasesOrchestrator` is probably unnecessary too I would say.

PR #2621, opened 2025-08-21
Migrate RC SharedPreferences data to a separate file

tonidero:
### Description
We were storing data in the default shared preferences file, which could potentially be used by the app developer. That's a bad practice and it's more difficult to do some things with. This PR changes to a separate shared preferences file and adds a migration, so we migrate all RC data from the previous shared preferences file to the new one. We won't delete the data from the previous shared preferences file, but that will remain unused.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -253,6 +253,7 @@ internal class PurchasesOrchestrator(
         log(LogIntent.DEBUG) { ConfigureStrings.APP_FOREGROUNDED }
         appConfig.isAppBackgrounded = false
 
+        // $RCAnonymousID:a36af41a15e844dc8786bd6f0386d4ad
         enqueue {
             if (shouldRefreshCustomerInfo(firstTimeInForeground)) {
                 log(LogIntent.DEBUG) { CustomerInfoStrings.CUSTOMERINFO_STALE_UPDATING_FOREGROUND }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -253,7 +253,6 @@ internal class PurchasesOrchestrator(
         log(LogIntent.DEBUG) { ConfigureStrings.APP_FOREGROUNDED }
         appConfig.isAppBackgrounded = false
 
-        // $RCAnonymousID:a36af41a15e844dc8786bd6f0386d4ad
         enqueue {
             if (shouldRefreshCustomerInfo(firstTimeInForeground)) {
                 log(LogIntent.DEBUG) { CustomerInfoStrings.CUSTOMERINFO_STALE_UPDATING_FOREGROUND }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
@@ -5,8 +5,6 @@ import android.content.SharedPreferences
 import android.preference.PreferenceManager
 import androidx.annotation.VisibleForTesting
 import androidx.core.content.edit
-import java.util.concurrent.locks.ReentrantLock
-import kotlin.concurrent.withLock
 
 /**
  * Provides an instance of SharedPreferences to be used by the RevenueCat SDK.
@@ -30,13 +28,11 @@ internal class SharedPreferencesManager(
         const val SHARED_PREFERENCES_PREFIX = "com.revenuecat.purchases."
     }
 
-    private val migrationLock = ReentrantLock()
-
     /**
      * Gets the appropriate shared preferences, performing migration if needed
      */
     fun getSharedPreferences(): SharedPreferences {
-        migrationLock.withLock {
+        synchronized(this) {
             if (shouldPerformMigration()) {
                 performMigration()
             }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
@@ -40,25 +40,21 @@ internal class SharedPreferencesManager(
      * Gets the appropriate shared preferences, performing migration if needed
      */
     fun getSharedPreferences(): SharedPreferences {
-        synchronized(this) {
-            val alreadyHasVersion = hasRevenueCatVersion()
-            if (shouldPerformMigration(alreadyHasVersion)) {
-                performMigration()
-                updateSharedPreferencesVersion()
-            } else if (!alreadyHasVersion) {
-                updateSharedPreferencesVersion()
-            }
-        }
+        synchronized(this) { ensureMigrated() }
         return revenueCatSharedPreferences
     }
 
-    /**
-     * Checks if migration should be performed by checking if RevenueCat preferences are empty
-     * and legacy preferences contain RevenueCat data
-     */
-    private fun shouldPerformMigration(alreadyHasVersion: Boolean): Boolean {
-        return !alreadyHasVersion && legacySharedPreferences.value.all.keys.any { key ->
-            key.startsWith(SHARED_PREFERENCES_PREFIX)
+    private fun ensureMigrated() {
+        val alreadyHasVersion = hasRevenueCatVersion()
+        if (!alreadyHasVersion) {
+            if (legacySharedPreferences.value.all.keys.any {
+                        key ->
+                    key.startsWith(SHARED_PREFERENCES_PREFIX)
+                }
+            ) {
+                performMigration()
+            }
+            updateSharedPreferencesVersion()
         }
     }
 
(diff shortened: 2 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsSynchronizer.kt
```diff
@@ -30,12 +28,6 @@ internal class DiagnosticsSynchronizer(
 
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
         const val MAX_EVENTS_TO_SYNC_PER_REQUEST: Int = 200
-
-        fun initializeSharedPreferences(context: Context): SharedPreferences =
```
tonidero, line 34: This was unused, so I took the chance to delete it. I think it might be worth moving all the shared preferences handling to the new `SharedPreferencesManager`... but didn't want to overcomplicate this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,143 @@
...
+                }
+                is Long -> {
+                    editor.putLong(key, value)
+                    true
+                }
+                is Float -> {
+                    editor.putFloat(key, value)
+                    true
+                }
+                is Set<*> -> {
```
tonidero, line 122: These are the only types of data that we store in the shared preferences files (we actually don't use booleans/ints/floats... but I thought it might be good to add just in case.

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,143 @@
...
+    private fun performMigration() {
+        log(
+            LogIntent.DEBUG,
+        ) { "Starting shared preferences migration from legacy to RevenueCat-specific preferences" }
+
+        val revenueCatKeys = getRevenueCatKeysToMigrate()
+
+        val legacyPrefs = legacySharedPreferences
+        val revenueCatPrefs = revenueCatSharedPreferences
+        revenueCatPrefs.edit(commit = true) {
```
tonidero: I doubted here a bit... I used `commit = true`, which is synchronous. The main problem here is to possibly cause ANRs... However, doing it async could cause some annoying edge cases, and this should only run once... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -253,6 +253,7 @@ internal class PurchasesOrchestrator(
         log(LogIntent.DEBUG) { ConfigureStrings.APP_FOREGROUNDED }
         appConfig.isAppBackgrounded = false
 
+        // $RCAnonymousID:a36af41a15e844dc8786bd6f0386d4ad
```
JayShortway: Is this intentional? 😄

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
JayShortway: I think using `apply()` is fine here. The in-memory representation is updated immediately. The only edge cases could come from another editor calling `commit()` before our `apply()` wrote to disk. Ideally we avoid `commit()` in general.

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,143 @@
...
+        // If RevenueCat preferences already have data, no migration needed
+        if (revenueCatPrefs.all.isNotEmpty()) {
+            return false
+        }
+
+        // Check if legacy preferences contain any RevenueCat data
+        return legacyPrefs.all.keys.any { key ->
+            key.startsWith(SHARED_PREFERENCES_PREFIX)
+        }
+    }
```
JayShortway: Maybe we can write a specific key (like `version`) to the `revenueCatPrefs` when migration is done, and hardcode an `EXPECTED_VERSION` in the `SharedPreferencesManager`. That way we can:
- avoid loading the `legacySharedPreferences` in memory if the `version` key exists in `revenueCatPrefs` (constructor parameter should be a lambda), and
- support changing schemas and stuff in `revenueCatPrefs` by updating the version and writing version-specific migrations.

What do you think?

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,143 @@
...
+
+        val legacyPrefs = legacySharedPreferences
+        val revenueCatPrefs = revenueCatSharedPreferences
+        revenueCatPrefs.edit(commit = true) {
+            var migratedCount = 0
+            for (key in revenueCatKeys) {
+                if (migratePreferenceValue(legacyPrefs, this, key)) {
+                    migratedCount++
+                }
+            }
```
JayShortway: `migratedCount` seems to be unused?

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
JayShortway: Yea agreed!

purchases/src/test/java/com/revenuecat/purchases/common/SharedPreferencesManagerTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        every { mockRevenueCatPrefs.all } returnsMany listOf(
+            emptyMap(), // First call - empty, triggers migration
+            mapOf(REVENUECAT_KEY_1 to "value1") // After migration - has data, no more migration
+        )
+        every { mockLegacyPrefs.all } returns mapOf(REVENUECAT_KEY_1 to "value1")
+
+        val manager = SharedPreferencesManager(mockContext)
+        
+        val result1 = manager.getSharedPreferences()
+        val result2 = manager.getSharedPreferences()
```
JayShortway: This is not necessarily concurrent right, as the test runs on a single thread? Maybe I'm misunderstanding the MockK setup. Maybe this works?

```kotlin
var result1: SharedPreferences? = null
val thread1 = thread { result1 = manager.getSharedPreferences() }

var result2: SharedPreferences? = null
val thread2 = thread { result2 = manager.getSharedPreferences() }

thread1.join()
thread2.join()
```

JayShortway (commented):
Very nice to have this! Just some suggestions.

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
tonidero: I love that! I applied it in https://github.com/RevenueCat/purchases-android/pull/2621/commits/5528e5f87cb754fe3c64183af834d71cfde9bb13

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Checks if migration should be performed by checking if RevenueCat preferences are empty
+     * and legacy preferences contain RevenueCat data
+     */
+    private fun shouldPerformMigration(alreadyHasVersion: Boolean): Boolean {
+        return !alreadyHasVersion && legacySharedPreferences.value.all.keys.any { key ->
+            key.startsWith(SHARED_PREFERENCES_PREFIX)
+        }
+    }
```
JayShortway: Maybe this one can be called `ensureMigrated()` and handle all migration concerns, and then `getSharedPreferences()` can just be: 

```kotlin
fun getSharedPreferences(): SharedPreferences {
    synchronized(this) { ensureMigrated() }
    return revenueCatSharedPreferences
}
```

Right now the "when and how to migrate" logic is a little split between `getSharedPreferences()` and `shouldPerformMigration()`. Let me know what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,153 @@
...
+     * Performs the migration from legacy shared preferences to RevenueCat-specific ones
+     */
+    private fun performMigration() {
+        log(
+            LogIntent.DEBUG,
+        ) { "Starting shared preferences migration from legacy to RevenueCat-specific preferences" }
+
+        val revenueCatKeys = getRevenueCatKeysToMigrate()
+
+        val legacyPrefs = legacySharedPreferences
```
JayShortway:
```suggestion
        val legacyPrefs by legacySharedPreferences
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2621/commits/12aa814cfd76d0d68c00cafe6b841abfbc2086f1

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2621/commits/84029c81629246792a5c0e37ee009005527df562

PR #2623, opened 2025-08-21
MON-1193 Support delayed close button (Component Transitions)

JZDesign:
### Checklist

### Motivation
We have several users who want the ability to delay the appearance of the close paywall button. 

### Description

Implemented a Transitions mechanism that will optionally hide/show a view when the component is marked with a transition object. This lays the groundwork for accepting custom animation declarations so that it would not error out if/when we add that in the future.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentView.kt
```diff
@@ -70,105 +71,107 @@ internal fun ButtonComponentView(
     onClick: suspend (PaywallAction) -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    // Get a ButtonComponentState that calculates the stateful properties we should use.
-    val buttonState = rememberButtonComponentState(
-        style = style,
-        paywallState = state,
-    )
+    TransitionView(transition = style.transition) {
```
JZDesign, line 74: I recommend hiding whitespace when reviewing this file

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallTransition.kt
```diff
@@ -0,0 +1,68 @@
...
+
+        @SerialName("lazy")
+        LAZY,
+    }
+
+    @OptIn(ExperimentalSerializationApi::class)
+    @InternalRevenueCatAPI
+    @Serializable
+    @JsonClassDiscriminator("type")
+    sealed class TransitionType {
```
tonidero: Hmm I find a bit confusing we have both a `TransitionType` and a `AnimationType`. Do we need both? And in that case, should we document the difference?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/ButtonComponent.kt
```diff
@@ -25,6 +25,7 @@ import kotlinx.serialization.encoding.Encoder
 class ButtonComponent(
     @get:JvmSynthetic val action: Action,
     @get:JvmSynthetic val stack: StackComponent,
+    @get:JvmSynthetic val transition: PaywallTransition? = null,
```
tonidero, line 28: Nothing to do here but one thing I was seeing is that we won't allow to override the animation for different screen sizes or other parameters. Having said that, we also don't have overrides for button actions, so I think this is more than likely ok, and I don't think there would be many use cases for us to override a button action or transition depending on screen sizes or other parameters, at least for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -424,6 +424,7 @@ internal class StyleFactory(
             ButtonComponentStyle(
                 stackComponentStyle = stack,
                 action = action,
+                transition = component.transition,
```
tonidero, line 427: I guess we don't need to do any processing on the transition, so this seems good 👍 (I was just investigating whether we needed to do any transformations, same way we do for actions.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallTransition.kt
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * A [GREEDY] displacement will result in the space being taken up by the incoming view
+     * *before* it attempts to transition into the view hierarchy.
+     *
+     * A [LAZY] displacement will not do this, instead it will result in shifting the layout
+     * as the new view inserts itself.
+     */
+    @InternalRevenueCatAPI
+    @Serializable
+    enum class DisplacementStrategy {
```
tonidero, line 38: For all enums and sealed classes, we should write a custom serializer that provides a default value in case there is an issue deserializing, you can look at sample of deserializers inheriting `EnumDeserializerWithDefault`.

The reason for this was that, if we add more elements to an enum or sealed class, this could cause older SDKs to stop parsing the response, causing the paywall to fail to render. So we decided that, unless it's critical for displaying the paywall, all enums and sealed classes should have defaults

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
```diff
@@ -0,0 +1,53 @@
...
+    @get:JvmSynthetic
+    @SerialName("ms_duration")
+    val msDuration: Int?,
+) {
+
+    @OptIn(ExperimentalSerializationApi::class)
+    @InternalRevenueCatAPI
+    @Serializable
+    @JsonClassDiscriminator("type")
+    sealed class AnimationType {
```
tonidero: Same for sealed classes... We shoudl write a custom deserializer so we pick a default in case new values are added later on.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
```diff
@@ -0,0 +1,53 @@
...
+        object EaseOut : AnimationType()
+
+        @Serializable
+        @SerialName("linear")
+        object Linear : AnimationType()
+
+        @Poko
+        @Serializable
+        @SerialName("custom")
+        class Custom(
```
tonidero: Hmm I'm not sure how this will work... Would we want developers to give us an animation depending on the `value`? maybe planned for future PRs? If it's not needed for right now, maybe we could make this an enum with a default value and remove this `Custom` type for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/TransitionView.kt
```diff
@@ -0,0 +1,137 @@
...
+
+                transition.AnimatedVisibility { content() }
+            }
+        } else {
+            transition.AnimatedVisibility { content() }
+        }
+    }
+}
+
+private fun Modifier.hidden(): Modifier = this.layout { measurable, constraints ->
```
tonidero, line 48: Hmm I was thinking whether it would make more sense to just add the content again with an alpha of 0, but then we would be rendering the content twice. I think this seems like a smart solution! We should just test to make sure it behaves as expected, but seems right 👍

tonidero (commented):
I think the only real thing is to add defaults to the enums. Other than that, just some questions. Great job!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
JZDesign: Yea, the thought is to allow users to inject a function that will return an animation (or if looking at the other file, a transition) that way they get the ability to fully customize this feature, but I'm not handling that now, It in this PR really as a placeholder/conversation starter. 

I can remove custom, but I did want to get your thoughts on this concept

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallTransition.kt
JZDesign: I'll add docs. Yes I think we need both because the transition is the method of showing the view (slide, opacity, scale, etc.,) and the animation is the easing that is applied to the process. So we could have a Slide with linear easing, an opacity transition with EaseInOut easing etc.,

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/ButtonComponent.kt
JZDesign: And if the user DID want to do that, they could just conditionally show the button they want on a phone and another on a tablet with different actions or transitions.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
tonidero: I think it's an interesting idea! My only concern is the way for developers to feed us those animations... I'm thinking of some sort of callback that the developer can implement, where we give them the custom animation value, and they return us an animation... But this would be very tricky to support in the hybrids. Not sure if we want to get into that complexity TBH...

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallTransition.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Defines the type of transition to use for paywall transitions.
+     *
+     * [NOTE] This is a sealed class and not an enum because we see a future where we may want
+     * to pass back more verbose instructions to the view layer than a simple enum case
+     */
+    @InternalRevenueCatAPI
+    @Serializable(with = TransitionTypeDeserializer::class)
+    sealed class TransitionType {
```
tonidero: Now that we've removed the `Custom`, we could make this an enum, same for `AnimationType`. Since it's annotated with `InternalRevenueCatAPI`, we can always change it later if needed.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
```diff
@@ -0,0 +1,64 @@
...
+    val msDuration: Int?,
+) {
+
+    /**
+     * Defines the type of animation to use for paywall transitions.
+     *
+     * [NOTE] This is a sealed class and not an enum because we see a future where we may want
+     * to pass back more verbose instructions to the view layer than a simple enum case
+     */
+    @InternalRevenueCatAPI
```
tonidero: NABD, but inner classes/enums of an `@InternalRevenueCatAPI` do not need to be annotated again so we could remove this annotation from these.

tonidero (approved):
This looks great to me! Amazing job!

PR #2624, opened 2025-08-21
Fix embedded font family loading

tonidero:
### Description
Embedded font families with multiple fonts didn't seem to be working. We were not grouping the fonts together correctly as it was expected.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
@@ -8,7 +8,7 @@ import androidx.compose.ui.text.font.FontFamily
 import com.revenuecat.purchases.Purchases
 import com.revenuecat.purchases.UiConfig
 import com.revenuecat.purchases.paywalls.DownloadedFontFamily
-import com.revenuecat.purchases.ui.revenuecatui.utils.FontFamilyXMLParser
+import com.revenuecat.purchases.ui.revenuecatui.utils.FontFamilyXmlParser
 import java.util.Locale
 
 /**
@@ -67,11 +67,12 @@ internal class PaywallResourceProvider(
             return null
         }
         return try {
-            FontFamilyXMLParser.parse(parser)
+            FontFamilyXmlParser.parse(parser)
         } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
             // This can happen if the XML is malformed or not a valid font family.
             // We log the error and return null.
-            Logger.e("Error parsing XML font family with resource ID $resourceId", e)
+            val resourceName = resources.getResourceEntryNameOrNull(resourceId)
+            Logger.e("Error parsing XML font family with resource ID ${resourceName ?: resourceId}", e)
             null
         } finally {
             parser.close()
@@ -105,3 +106,10 @@ internal fun Context.toResourceProvider(): ResourceProvider {
 private fun Context.applicationName(): String {
     return applicationInfo.loadLabel(packageManager).toString()
 }
+
+private fun Resources.getResourceEntryNameOrNull(resourceId: Int): String? =
+    try {
+        getResourceEntryName(resourceId)
+    } catch (_: Resources.NotFoundException) {
+        null
+    }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXmlParser.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXmlParser.kt
@@ -6,14 +6,24 @@ import androidx.compose.ui.text.font.Font
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontStyle
 import androidx.compose.ui.text.font.FontWeight
-import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
 import org.xmlpull.v1.XmlPullParser
+import kotlin.jvm.Throws
+
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal data class ParsedFont(
+    val resId: Int,
+    val weight: Int,
+    val style: FontStyle,
+)
 
 @Suppress("NestedBlockDepth")
-internal object FontFamilyXMLParser {
-    private const val unrecognizedValue = -1
-    private const val defaultFontWeight = 400
+internal object FontFamilyXmlParser {
+    private const val UNRECOGNIZED_VALUE = -1
+    private const val DEFAULT_FONT_WEIGHT = 400
+    private const val ANDROID_NAMESPACE = "http://schemas.android.com/apk/res/android"
+    private const val APP_NAMESPACE = "http://schemas.android.com/apk/res-auto"
 
+    @Throws
     fun parse(parser: XmlResourceParser): FontFamily? {
         val parsedFonts = parseXmlData(parser)
         return if (parsedFonts.isNotEmpty()) {
@@ -31,57 +41,53 @@ internal object FontFamilyXMLParser {
         }
     }
 
-    @VisibleForTesting
-    internal fun parseXmlData(parser: XmlResourceParser): List<Triple<Int, Int, FontStyle>> {
-        return try {
-            val parsedFonts = mutableListOf<Triple<Int, Int, FontStyle>>()
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    @Throws
+    internal fun parseXmlData(parser: XmlResourceParser): List<ParsedFont> {
+        val parsedFonts = mutableListOf<ParsedFont>()
 
-            var eventType = parser.eventType
-            while (eventType != XmlPullParser.END_DOCUMENT) {
-                when (eventType) {
-                    XmlPullParser.START_TAG -> {
-                        if (parser.name == "font") {
-                            val parsedFont = parseFontData(parser)
-                            if (parsedFont != null) {
-                                parsedFonts.add(parsedFont)
-                            }
+        var eventType = parser.eventType
+        while (eventType != XmlPullParser.END_DOCUMENT) {
+            when (eventType) {
+                XmlPullParser.START_TAG -> {
+                    if (parser.name == "font") {
+                        val parsedFont = parseFontData(parser)
+                        if (parsedFont != null) {
+                            parsedFonts.add(parsedFont)
                         }
                     }
                 }
-                eventType = parser.next()
             }
-
-            parsedFonts
-        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
-            Logger.e("Error parsing XML font family", e)
-            emptyList()
+            eventType = parser.next()
         }
+
+        return parsedFonts
     }
 
-    private fun parseFontData(parser: XmlResourceParser): Triple<Int, Int, FontStyle>? {
+    private fun parseFontData(parser: XmlResourceParser): ParsedFont? {
         val fontResId = getFontResourceId(parser)
-        if (fontResId == unrecognizedValue) return null
+        if (fontResId == UNRECOGNIZED_VALUE) return null
 
         val fontWeight = getFontWeight(parser)
         val fontStyle = getFontStyle(parser)
 
-        return Triple(fontResId, fontWeight, fontStyle)
+        return ParsedFont(fontResId, fontWeight, fontStyle)
     }
 
     private fun getFontResourceId(parser: XmlResourceParser): Int {
         // Try app namespace first (http://schemas.android.com/apk/res-auto)
         var fontResId = parser.getAttributeResourceValue(
-            "http://schemas.android.com/apk/res-auto",
+            APP_NAMESPACE,
             "font",
-            unrecognizedValue,
+            UNRECOGNIZED_VALUE,
         )
 
         // Fallback to android namespace
-        if (fontResId == unrecognizedValue) {
+        if (fontResId == UNRECOGNIZED_VALUE) {
             fontResId = parser.getAttributeResourceValue(
-                "http://schemas.android.com/apk/res/android",
+                ANDROID_NAMESPACE,
                 "font",
-                unrecognizedValue,
+                UNRECOGNIZED_VALUE,
             )
         }
 
@@ -91,34 +97,34 @@ internal object FontFamilyXMLParser {
     private fun getFontWeight(parser: XmlResourceParser): Int {
         // Try app namespace first
         var fontWeight = parser.getAttributeIntValue(
-            "http://schemas.android.com/apk/res-auto",
+            APP_NAMESPACE,
             "fontWeight",
-            unrecognizedValue,
+            UNRECOGNIZED_VALUE,
         )
 
         // Fallback to android namespace
-        if (fontWeight == unrecognizedValue) {
+        if (fontWeight == UNRECOGNIZED_VALUE) {
             fontWeight = parser.getAttributeIntValue(
-                "http://schemas.android.com/apk/res/android",
+                ANDROID_NAMESPACE,
                 "fontWeight",
-                defaultFontWeight,
+                DEFAULT_FONT_WEIGHT,
             )
         }
 
-        return if (fontWeight == unrecognizedValue) defaultFontWeight else fontWeight
+        return if (fontWeight == UNRECOGNIZED_VALUE) DEFAULT_FONT_WEIGHT else fontWeight
     }
 
     private fun getFontStyle(parser: XmlResourceParser): FontStyle {
         // Try app namespace first
         var fontStyleValue = parser.getAttributeValue(
-            "http://schemas.android.com/apk/res-auto",
+            APP_NAMESPACE,
             "fontStyle",
         )
 
         // Fallback to android namespace
         if (fontStyleValue == null) {
             fontStyleValue = parser.getAttributeValue(
-                "http://schemas.android.com/apk/res/android",
+                ANDROID_NAMESPACE,
                 "fontStyle",
             )
         }
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXmlParserTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXmlParserTest.kt
@@ -3,22 +3,23 @@ package com.revenuecat.purchases.ui.revenuecatui.utils
 import android.content.res.XmlResourceParser
 import androidx.compose.ui.text.font.FontStyle
 import org.assertj.core.api.Assertions.assertThat
+import org.intellij.lang.annotations.Language
 import org.junit.Test
 import org.xmlpull.v1.XmlPullParser
 import org.xmlpull.v1.XmlPullParserFactory
 import java.io.StringReader
 
-class FontFamilyXMLParserTest {
+class FontFamilyXmlParserTest {
 
     @Test
     fun `parseXmlData returns empty list for empty XML`() {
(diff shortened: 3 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,9 +64,33 @@ internal sealed interface FontSpec {
 internal fun Map<FontAlias, FontsConfig>.determineFontSpecs(
     resourceProvider: ResourceProvider,
 ): Map<FontAlias, FontSpec> {
+    val resourceFontFamilies = values
+        .toSet()
+        .map { it.android }
+        .filterIsInstance<FontInfo.Name>()
+        .filter { it.family != null }
+        .groupBy { it.family!! }
```
JZDesign: !!'s make me nervous

tonidero: Totally agree... this is to make the compiler happy, we filter for null values on the previous line, but wasn't sure of a good way to make the compiler happy for this... Will try to clean that up tomorrow.

tonidero: Removed in [Remove force unwrap](https://github.com/RevenueCat/purchases-android/pull/2624/commits/3f8dc016d223e61521b60bbc4a4291ffc2bfea64)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -29,19 +29,21 @@ private val GoogleFontsProvider: GoogleFont.Provider = GoogleFont.Provider(
...
 /**
  * A `FontSpec` is a more detailed version of [FontInfo]. A [FontInfo.Name] can be resolved to a generic font
  * (e.g. "sans-serif"), a font resource provided by the app, or a device font provided by the OEM.
  *
  * Determining this is relatively costly for font resources. So this abstraction allows us to perform this logic only
- * once for each one (see [determineFontSpecs]) in the validation step, before resolving the actual font where needed.
- *
- * It also allows us to defer resolving the actual font to the UI layer, as only at that time do we know the exact
- * override that's being used. We need to know this, because we need to know the [FontWeight] for which to resolve the
- * font.
```
tonidero, line 41: I think this isn't really the case anymore, considering we need to create the font families with all weights. The only thing that needs to be determined in the call site is which font weight to use for the available font family.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,9 +60,50 @@ internal sealed interface FontSpec {
 internal fun Map<FontAlias, FontsConfig>.determineFontSpecs(
     resourceProvider: ResourceProvider,
 ): Map<FontAlias, FontSpec> {
+    val resourceFontFamilies = values
+        .toSet()
+        .map { it.android }
+        .filterIsInstance<FontInfo.Name>()
+        .mapNotNull { fontInfo -> fontInfo.family?.let { family -> fontInfo to family } }
+        .groupBy({ it.second }, { it.first })
```
tonidero: This is a way to group by non-null family properties while making the compiler happy and avoiding the force-unwrap... Lmk if you have other suggestions, maybe a force-unwrap isn't so bad if we previously verify that the values is not null though...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,9 +60,50 @@ internal sealed interface FontSpec {
...
+                            id = it,
+                            weight = fontInfo.weight,
+                            style = fontInfo.style?.toComposeFontStyle(),
+                        )
+                    }
+            }
+        }
+        .filterValues { it.isNotEmpty() }
+        .mapValues { (_, resourceFonts) ->
+            if (resourceFonts.size == 1) {
```
tonidero: I took the decission here to only check if the resource is an xml font if all the fonts in the same family point to the same resource. If you have different resources for different fonts in the same font family, it's very unlikely for it to be an xml I think. This way we avoid querying for these extra resource files in those cases.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
```diff
@@ -56,6 +62,36 @@ internal class PaywallResourceProvider(
     override fun getResourceIdentifier(name: String, type: String): Int =
         resources.getIdentifier(name, type, packageName)
 
+    @Suppress("ReturnCount")
+    @SuppressLint("RestrictedApi") // FontResourcesParserCompat.*
```
tonidero: This is a restricted API that should only be called from within the same library... But seemed to be the easiest way to parse an XML font family file that I could find.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,9 +60,50 @@ internal sealed interface FontSpec {
...
+        .map { it.android }
+        .filterIsInstance<FontInfo.Name>()
+        .mapNotNull { fontInfo -> fontInfo.family?.let { family -> fontInfo to family } }
+        .groupBy({ it.second }, { it.first })
+        .mapValues { (_, fontInfos) ->
+            val resourceIdsSeen = mutableSetOf<Int>()
+            fontInfos.mapNotNull { fontInfo ->
+                resourceProvider.getResourceIdentifier(name = fontInfo.value, type = "font")
+                    .takeIf { it != 0 && it !in resourceIdsSeen }
+                    ?.also { resourceIdsSeen.add(it) }
```
tonidero, line 74: No need to continue here with multiple fonts to check if they are pointing to the same resource id.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,9 +60,54 @@ internal sealed interface FontSpec {
...
+        .toSet()
+        .map { it.android }
+        .filterIsInstance<FontInfo.Name>()
+        .mapNotNull { fontInfo -> fontInfo.family?.let { family -> fontInfo to family } }
+        .groupBy({ it.second }, { it.first })
+        .mapValues { (_, fontInfos) ->
+            val resourceNamesSeen = mutableSetOf<String>()
+            fontInfos.mapNotNull { fontInfo ->
+                if (fontInfo.value in resourceNamesSeen) {
+                    null
```
tonidero: This is a small optimization to avoid checking for the same resource name multiple times within the same family (for example for xml families). I guess this does mean that the developer can't set multiple fonts to have the same font name... but seems like something they shouldn't do anyway.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
JayShortway: Maybe the `.map()` until `.groupBy()` can be replaced by this?
```kotlin
    val resourceFontFamilies = values
        .toSet()
        .mapNotNull { fontsConfig ->
            val fontInfo = fontsConfig.android as? FontInfo.Name
            fontInfo?.family?.let { family -> family to fontInfo }
        }
        .groupBy({ (family, _) -> family }, { (_, fontInfo) -> fontInfo })
```

Helps a bit with performance, as we're only looping once instead of 3 times. Note that I also switched the order of the pair, and added named arguments in the `groupBy` for better readability.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,9 +60,54 @@ internal sealed interface FontSpec {
...
+                FontFamily(
+                    resourceFonts.map { resourceFont ->
+                        Font(
+                            resId = resourceFont.id,
+                            weight = resourceFont.weight?.let { FontWeight(it) } ?: FontWeight.Normal,
+                            style = resourceFont.style ?: FontStyle.Normal,
+                        )
+                    },
+                ),
+            )
```
JayShortway: Here we are mapping values, filtering values, and then mapping values again. It's difficult to follow what's going on, and also could be improved performance wise. I think both `mapValues()` could be combined, if we move `filterValues()` to the end? I would also suggest coming up with some well-named extension functions to make all the mapping logic more readable.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -29,19 +29,21 @@ private val GoogleFontsProvider: GoogleFont.Provider = GoogleFont.Provider(
     certificates = R.array.com_google_android_gms_fonts_certs,
 )
 
+private data class ResourceFontSpec(
+    val id: Int,
+    val weight: Int?,
+    val style: FontStyle?,
+)
```
JayShortway: Having both `FontSpec.Resource` and `ResourceFontSpec` is a bit confusing haha. Maybe we can avoid the need for this type altogether though if we combine the 2 `mapValues()` calls (in my previous comment)?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
JayShortway: Hmm, but this is tricky though. We could run into binary compatibility problems if the app depends on a different version of `androidx.core`. I think we should see if we can find another way. Why do we need to parse the XML anyway? Compose should be able to deal with XML fonts. Can we provide this data to the Compose layer without parsing the XML?

JayShortway (commented):
Great job figuring this out! 💪 Some suggestions here and there.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
tonidero: > Compose should be able to deal with XML fonts.

Unfortunately that's not the case from testing... If you reference a xml resource id when creating a Font, it does not load all fonts, and only a single one. Also tried loading a Typeface from the XML, but that doesn't work either...

tonidero: We could protect ourselves against crashes caused by binary incompatibilities... Otherwise, I think we would need to write the parsing logic ourselves

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
tonidero: You do have a great point. Tried to clean it up a bit in https://github.com/RevenueCat/purchases-android/pull/2624/commits/176ef68662cb58980be1d364a64d41c4ee7e5888. Lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
tonidero: I was able to add a custom parser in https://github.com/RevenueCat/purchases-android/pull/2624/commits/6ea520422f02d375ea09d060e7bee568f35e792d and https://github.com/RevenueCat/purchases-android/pull/2624/commits/7853a0c8aa70f1c82cfe631efd7b240f1cdd5813

Lmk what you think!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -58,14 +54,71 @@ internal sealed interface FontSpec {
...
+        .mapNotNull { fontsConfig ->
+            val fontInfo = fontsConfig.android as? FontInfo.Name
+            fontInfo?.family?.let { family -> family to fontInfo }
+        }
+        .groupBy({ (family, _) -> family }, { (_, fontInfo) -> fontInfo })
+        .mapNotNull { (family, fontInfos) ->
+            fontInfos.toFontSpecResource(resourceProvider)?.let {
+                family to it
+            }
+        }.associateBy({ it.first }, { it.second })
```
JayShortway, line 68: Love it 😗👌

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParser.kt
```diff
@@ -0,0 +1,131 @@
...
+import androidx.annotation.VisibleForTesting
+import androidx.compose.ui.text.font.Font
+import androidx.compose.ui.text.font.FontFamily
+import androidx.compose.ui.text.font.FontStyle
+import androidx.compose.ui.text.font.FontWeight
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+import org.xmlpull.v1.XmlPullParser
+
+@Suppress("NestedBlockDepth")
+internal object FontFamilyXMLParser {
```
JayShortway: [Kotlin convention](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names) nit: 
```suggestion
internal object FontFamilyXmlParser {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParser.kt
```diff
@@ -0,0 +1,131 @@
...
+                    )
+                },
+            )
+        } else {
+            null
+        }
+    }
+
+    @VisibleForTesting
+    internal fun parseXmlData(parser: XmlResourceParser): List<Triple<Int, Int, FontStyle>> {
```
JayShortway: Would it be possible to test it without `@VisibleForTesting`? Can we not test all functionality through `parse()`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParser.kt
```diff
@@ -0,0 +1,131 @@
...
+        var fontWeight = parser.getAttributeIntValue(
+            "http://schemas.android.com/apk/res-auto",
+            "fontWeight",
+            unrecognizedValue,
+        )
+
+        // Fallback to android namespace
+        if (fontWeight == unrecognizedValue) {
+            fontWeight = parser.getAttributeIntValue(
+                "http://schemas.android.com/apk/res/android",
```
JayShortway: Maybe we can keep the namespaces as constants?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParser.kt
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private fun parseFontData(parser: XmlResourceParser): Triple<Int, Int, FontStyle>? {
+        val fontResId = getFontResourceId(parser)
+        if (fontResId == unrecognizedValue) return null
+
+        val fontWeight = getFontWeight(parser)
+        val fontStyle = getFontStyle(parser)
+
+        return Triple(fontResId, fontWeight, fontStyle)
```
JayShortway: This Triple could be a private (data) class named `ParsedFont`. What do you think?

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParserTest.kt
```diff
@@ -0,0 +1,238 @@
...
+        
+        val parser = TestXmlResourceParser(xmlContent, mapOf("@font/regular" to 123))
+        val result = FontFamilyXMLParser.parse(parser)
+        
+        assertThat(result).isNotNull()
+    }
+
+    // Test implementation of XmlResourceParser that uses real XML parsing
+    private class TestXmlResourceParser(
+        xmlContent: String,
```
JayShortway: If we add this annotation, Android Studio will treat every string passed-in as actual XML and do syntax highlighting and allows you to format it and stuff:

```suggestion
        @Language("xml") xmlContent: String,
```

(A `// language=xml` comment above every field works too, but that needs to be repeated everywhere.)

You can do Option+Enter and "Edit XML Fragment" to format it.
image

And it will underline syntax errors:
image

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
```diff
@@ -56,6 +59,25 @@ internal class PaywallResourceProvider(
...
+            resources.getXml(resourceId)
+        } catch (_: Resources.NotFoundException) {
+            return null
+        }
+        return try {
+            FontFamilyXMLParser.parse(parser)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+            // This can happen if the XML is malformed or not a valid font family.
+            // We log the error and return null.
+            Logger.e("Error parsing XML font family with resource ID $resourceId", e)
```
JayShortway: Printing the resource name is a bit more helpful, as the ID is basically random:
```suggestion
            Logger.e("Error parsing XML font family with resource ID ${resources.getResourceEntryNameOrNull(resourceId) ?: resourceId}", e)
```

with `getResourceEntryNameOrNull()` being: 
```kotlin
private fun Resources.getResourceEntryNameOrNull(resourceId: Int): String? =
    try {
        getResourceEntryName(resourceId)
    } catch (_: Resources.NotFoundException) {
        null
    }
```

JayShortway (approved):
Amazing job on this! The custom XML parser is not too bad indeed! Main concern is the `@VisibleForTesting` thing.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParser.kt
tonidero: It's not really possible since Compose's `FontFamily` doesn't expose any getters to obtain information about the fonts that are part of that font family, so we can't assert that the result is correct.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/FontFamilyXMLParserTest.kt
tonidero: Oh nice! TIL!

PR #2625, opened 2025-08-22
Add `RevenueCatBackupAgent`

tonidero:
### Description
This PR adds a new `RevenueCatBackupAgent` that consumers of this library can optionally use to backup revenuecat data accross installations. To use it, developers need to add this in the application's AndroidManifest `<application>` tag:
```
android:backupAgent="com.revenuecat.purchases.backup.RevenueCatBackupAgent"
```

After that, the shared preferences file with the main content from RevenueCat will be saved using the device's backup system. See official docs for this type of backup here: https://developer.android.com/identity/data/keyvaluebackup.

If users already have their own BackupAgentHelper configured for their app, they can just add a line to backup the `com_revenuecat_purchases_preferences` preferences file to their BackupAgentHelper, like:
```
SharedPreferencesBackupHelper(this, RevenueCatBackupAgent.REVENUECAT_PREFS_FILE_NAME).also {
   addHelper("revenuecat_prefs_backup", it)
}
```

This will allow customers, specially for apps using anonymous users, to immediately regain access to any entitlements they had before changing devices or uninstalling and reinstalling, effectively removing the need for these users to "restore purchases".


purchases/src/main/kotlin/com/revenuecat/purchases/backup/RevenueCatProactiveBackupAgent.kt
```diff
@@ -0,0 +1,29 @@
...
+import android.app.backup.BackupDataInput
+import android.app.backup.BackupDataOutput
+import android.app.backup.SharedPreferencesBackupHelper
+import android.os.ParcelFileDescriptor
+import com.revenuecat.purchases.common.SharedPreferencesManager
+import com.revenuecat.purchases.common.debugLog
+
+private const val REVENUECAT_PREFS_BACKUP_KEY = "revenuecat_prefs_backup"
+
+class RevenueCatProactiveBackupAgent : BackupAgentHelper() {
```
tonidero: TBH, I'm not sure of the name... Maybe just `RevenueCatBackupAgent` makes more sense? The idea was to add docs to make it a "proactive restores" feature...

purchases/api-defauts.txt
rickvdl, line 1: I think there is a typo in the filename `api-defauts.txt` -> `api-defaults.txt`?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1144,6 +1148,9 @@ internal class PurchasesOrchestrator(
 
     private fun getPurchaseCompletedCallbacks(): Pair<SuccessfulPurchaseCallback, ErrorPurchaseCallback> {
         val onSuccess: SuccessfulPurchaseCallback = { storeTransaction, info ->
+            // This lets the backup manager know a change in data happened that would be good to backup.
+            // In this case, we want to make sure that if there is a purchase, we make sure to
```
rickvdl: Is missing word at the end of the comment here? :)

purchases/src/main/kotlin/com/revenuecat/purchases/backup/RevenueCatProactiveBackupAgent.kt
```diff
@@ -0,0 +1,48 @@
...
+ * See https://developer.android.com/identity/data/keyvaluebackup for more details on how key-value backup works
+ * and how to test it.
+ * - Setting the back up agent in your AndroidManifest would disable auto backup for your app, if it was enabled.
+ * If you want to use auto backup to also back up your app's data or have your own Backup Agent to backup some of your
+ * files, please make sure you add the SharedPreferences file `com_revenuecat_purchases_preferences` to your auto backup
+ * configuration. See https://developer.android.com/identity/data/autobackup.
+ */
+class RevenueCatProactiveBackupAgent : BackupAgentHelper() {
+    override fun onCreate() {
+        SharedPreferencesBackupHelper(this, SharedPreferencesManager.REVENUECAT_PREFS_FILE_NAME).also {
```
rickvdl: This is currently a constant, and in your example on how developers might add it to their existing backup code you mentioned the string value of this constant:
> SharedPreferencesBackupHelper(this, "com_revenuecat_purchases_preferences").also {
        addHelper("revenuecat_prefs_backup", it)
}

Should we expose a method somewhere that returned the current file name / key for developers as well? This would allow us to make changes in the future as to how / where the file is stored.

purchases/src/main/kotlin/com/revenuecat/purchases/backup/RevenueCatProactiveBackupAgent.kt
rickvdl: My two cents here; maybe `RevenueCatAutoBackupAgent`? Maybe `proactive` is more like a marketing term that might change? Also not super sure though

tonidero: I wouldn't call it `AutoBackup` since it might be confused with Android's auto backup system... But I think `RevenueCatBackupAgent` might be enough yeah.

purchases/src/main/kotlin/com/revenuecat/purchases/backup/RevenueCatProactiveBackupAgent.kt
tonidero: Good idea! Exposed here: https://github.com/RevenueCat/purchases-android/pull/2625/commits/0bd30e99da30231720410f6aee23b286bf540c41

purchases/api-defauts.txt
tonidero: Will do this in a separate PR, since it could cause conflicts otherwise. Good catch!

purchases/src/main/kotlin/com/revenuecat/purchases/backup/RevenueCatBackupAgent.kt
```diff
@@ -0,0 +1,51 @@
...
+ * ```
+ * This will back up the SharedPreferences file used by the RevenueCat SDK, allowing it to keep the same user as
+ * was previously used in the same or different device with the same Google account, removing the need for users to
+ * restore purchases.
+ *
+ * Some important notes:
+ * - This backup may not work on all devices, as it's ultimately controlled by the Android system and the user settings.
+ * See https://developer.android.com/identity/data/keyvaluebackup for more details on how key-value backup works
+ * and how to test it.
+ * - Setting the back up agent in your AndroidManifest would disable auto backup for your app, if it was enabled.
```
vegaro: nitpick, but we have both `back up` and `backup`  in this sentence

purchases/src/test/java/com/revenuecat/purchases/PurchasesCommonTest.kt
```diff
@@ -1255,6 +1255,32 @@ internal class PurchasesCommonTest: BasePurchasesTest() {
         }
     }
 
+    @Test
+    fun `when making purchase, backup manager is notified of data change`() {
```
vegaro:
```suggestion
    fun `when making a purchase, backup manager is notified of data change`() {
```

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
```diff
@@ -524,6 +524,28 @@ internal class PurchasesTest : BasePurchasesTest() {
         }
     }
 
+    @Test
+    fun `login called with different appUserID notifies backup manager`() {
```
vegaro, line 528: maybe we need a test for logout as well?

vegaro (approved):

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
tonidero: I kinda reused the existing `logout called with identified user makes right calls` test, which tests a few things... Might be worth splitting up in multiple tests, but I think that might be better to do on a separate PR?

PR #2634, opened 2025-08-25
File Repository

JZDesign:
### Checklist

### Motivation

We will be supporting Video components in the Paywalls soon, to do that we need a good way to cache the large files to the device.

### Description

This PR is a carbon copy of what I did in the iOS sdk recently. There are 2 cache mechanisms.

1. A deferred value cache—This essentially allows us to reuse the same coroutine job across multiple call-sites. That way we can prefetch and move on. When the paywall is launched the same download can be awaited when it's partially completed. Until it finishes, we can show the fallback image that should download far faster.

2. The file repository—This knows when to fetch the data from the web or when to read it from the file system. There are likely things I've missed here. But there are a couple of important things to note. 
    a. We are not currently hosting the videos in our ecosystem so a checksum isn't realistic in our v1
    b. I'm good at Kotlin, but not at all experienced with Android SDK development . Totally open to feedback.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
@@ -6,6 +6,7 @@ import com.revenuecat.purchases.LogHandler
 import com.revenuecat.purchases.common.currentLogHandler
 import com.revenuecat.purchases.common.verboseLog
 import com.revenuecat.purchases.utils.DefaultUrlConnectionFactory
+import com.revenuecat.purchases.utils.UrlConnection
 import com.revenuecat.purchases.utils.UrlConnectionFactory
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
@@ -98,34 +99,34 @@ internal class DefaultFileRepository(
         }.await()
     }
 
-    private suspend fun downloadFile(url: URL): ByteArray {
-        return try {
-            withContext(Dispatchers.IO) {
-                verboseLog { "Downloading remote font from $url" }
+    private suspend fun downloadFile(url: URL): ByteArray = try {
+        withContext(Dispatchers.IO) {
+            verboseLog { "Downloading remote font from $url" }
 
-                val connection = urlConnectionFactory.createConnection(url.toString())
+            val connection = urlConnectionFactory.createConnection(url.toString())
 
-                if (connection.responseCode != HttpURLConnection.HTTP_OK) {
-                    throw IOException("HTTP ${connection.responseCode} when downloading file at: $url")
-                }
-                try {
-                    val bytes = connection.inputStream.use { inputStream ->
-                        inputStream.readBytes()
-                    }
-                    connection.disconnect()
-                    return@withContext bytes
-                } catch (e: IOException) {
-                    val message = "Failed to read input stream for file at: $url. Error: ${e.localizedMessage}"
-                    logHandler.e("FileRepository", message, e)
-                    connection.disconnect()
-                    throw Error.FailedToFetchFileFromRemoteSource(message)
-                }
+            if (connection.responseCode != HttpURLConnection.HTTP_OK) {
+                throw IOException("HTTP ${connection.responseCode} when downloading file at: $url")
             }
-        } catch (e: IOException) {
-            val message = "Failed to fetch file from remote source: $url. Error: ${e.localizedMessage}"
-            logHandler.e("FileRepository", message, e)
-            throw Error.FailedToFetchFileFromRemoteSource(message)
+            return@withContext connection.consumeBytes()
         }
+    } catch (e: IOException) {
+        val message = "Failed to fetch file from remote source: $url. Error: ${e.localizedMessage}"
+        logHandler.e("FileRepository", message, e)
+        throw Error.FailedToFetchFileFromRemoteSource(message)
+    }
+
+    private fun UrlConnection.consumeBytes(): ByteArray = try {
+        val bytes = inputStream.use { inputStream ->
+            inputStream.readBytes()
+        }
+        disconnect()
+        bytes
+    } catch (e: IOException) {
+        val message = "Failed to read input stream for file. Error: ${e.localizedMessage}"
+        logHandler.e("FileRepository", message, e)
+        disconnect()
+        throw Error.FailedToFetchFileFromRemoteSource(message)
     }
 
     private fun saveCachedFile(uri: URI, data: ByteArray) {


purchases/src/main/kotlin/com/revenuecat/purchases/storage/DeferredValueStore.kt
```diff
@@ -0,0 +1,61 @@
+package com.revenuecat.purchases.storage
+
+import kotlinx.coroutines.Deferred
+
+open class SingleDeferredValueStore<T>(
```
JZDesign: I don't think I'll use this version (only the keyed version) so I will probably just remove it.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+        class FailedToSaveCachedFile(message: String) : Error(message)
+
+        /**
+         * Used when fetching the data fails.
+         */
+        class FailedToFetchFileFromRemoteSource(message: String) : Error(message)
+    }
+}
+
+internal class DeviceCache(
```
tonidero: Hmm can this be private? If not, I would consider a different name so it's not confused with the existing `DeviceCache`... Like `FileDeviceCache`?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * A file cache.
+ */
+internal interface FileRepositoryType {
+    /**
+     * Prefetch files at the given urls.
+     * @param urls An array of URL to fetch data from.
+     */
+    fun prefetch(urls: List<URL>): Job?
```
tonidero: Hmm so this is a convenience method to prefetch all files right?  Do we need to return a job here though? I think this could be just a "fire and forget" kinda API?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+ * The file manager is a service capable of storing data and returning the URL where that stored data exists.
+ */
+internal interface LargeItemCacheType {
+    fun generateLocalFilesystemURI(remoteURL: URL): URI?
+    fun cachedContentExists(uri: URI): Boolean
+    fun saveData(data: InputStream, uri: URI)
+}
+
+internal class FileRepository(
+    private val fileManager: LargeItemCacheType,
```
tonidero: I think the name of this as `fileManager` can be very confusing, since we might think it's an actual `FileManager`... Maybe we can rename to `fileCachingManager`?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob()),
+    private val logHandler: LogHandler = currentLogHandler,
```
tonidero: Interesting... I guess we are injecting this in order to be able to test logs?

We haven't really been testing logs in the Android SDK so far, though it might be interesting to do, it can make some tests more difficult to maintain... I kinda prefer to test the effects rather than the logs... But not too opposed either. I think some logs are important and might be worth testing.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+ */
+internal interface LargeItemCacheType {
+    fun generateLocalFilesystemURI(remoteURL: URL): URI?
+    fun cachedContentExists(uri: URI): Boolean
+    fun saveData(data: InputStream, uri: URI)
+}
+
+internal class FileRepository(
+    private val fileManager: LargeItemCacheType,
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob()),
```
tonidero: [Not related to this PR!]

I've seen some corruption issues when downloading fonts (checksums not matching), I think they may be happening when there are too many parallel downloads. 

Maybe we should limit the number of parallel jobs the scope can run? I think we can use  `limitedParallelism` on the dispatcher to limit this, and then probably adding some retry mechanism... But then again, this should be done on a separate PR :) Just mentioning it here so we consider it into the mechanism

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+        val dir = File(context.cacheDir, "rc_files")
+        if (!dir.exists()) {
+            dir.mkdirs()
+        }
+        dir
+    }
+
+    override fun generateLocalFilesystemURI(remoteURL: URL): URI? {
+        // Using a simple approach of taking the last path component.
+        // A more robust implementation might use a hash of the URL.
```
tonidero: Might be worth using the hash of the URL from the start? I agree it would be more robust and migrating later would be more difficult than starting from scratch.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+    fun generateLocalFilesystemURI(remoteURL: URL): URI?
+    fun cachedContentExists(uri: URI): Boolean
+    fun saveData(data: InputStream, uri: URI)
+}
+
+internal class FileRepository(
+    private val fileManager: LargeItemCacheType,
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob()),
+    private val logHandler: LogHandler = currentLogHandler,
+    private val urlConnection: UrlConnectionFactory = DefaultUrlConnectionFactory(),
```
tonidero: Should we rename to `urlConnectionFactory`? Can be confusing otherwise

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DeferredValueStore.kt
```diff
@@ -0,0 +1,61 @@
...
+open class KeyedDeferredValueStore<H, T>(
+    private val lock: Any = object {},
+) {
+    val deferred: MutableMap<H, Deferred<T>> = mutableMapOf()
+
+    fun getOrPut(key: H, task: () -> Deferred<T>) = synchronized(lock) {
+        deferred.get(key) ?: forgettingFailure(key, task).also { deferred.put(key, it) }
+    }
+
+    fun replaceValue(key: H, task: () -> Deferred<T>) = synchronized(lock) {
```
tonidero: Hmm I don't think we would need this method? If so I would probably removing it for now.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DeferredValueStore.kt
```diff
@@ -0,0 +1,61 @@
...
+        forgettingFailure(key, task).also { deferred.put(key, it) }
+    }
+
+    fun forgettingFailure(key: H, task: () -> Deferred<T>): Deferred<T> =
```
tonidero: This could be private I think?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DeferredValueStore.kt
```diff
@@ -0,0 +1,61 @@
...
+                }
+            }
+        }
+
+    fun clear() = synchronized(lock) {
+        deferred = null
+    }
+}
+
+open class KeyedDeferredValueStore<H, T>(
```
tonidero: This should be `internal`

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DeferredValueStore.kt
```diff
@@ -0,0 +1,61 @@
...
+            invokeOnCompletion { exception ->
+                exception?.run {
+                    synchronized(lock) {
+                        deferred.remove(key)
+                    }
+                }
+            }
+        }
+
+    fun clear() = synchronized(lock) {
```
tonidero: This shouldn't normally be needed I think? If we need it for tests, I would annotate it with `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)`. However, I would try to test this with it's proper public API, in which case we might want to remove this method as well.

tonidero (commented):
Looking great! Much more generic and reusable than what we have for fonts :). Left some comments, lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: I didn't want to return a job actually, but my Kotlin skills are not quite as good as my swift, and I was struggling to make a deterministic test without adding a delay of 100 ms

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: I usually test logs, but I didn't see that happening here, so I didn't test them, but I still prefer to inject dependencies when I can.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
tonidero: I think we should inject a `TestScope` to the class so we can "advance time" with `testScope.advanceUntilIdle()`. Lmk if I can help with this!

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
tonidero: I usually just use the static logger helper methods we added like `warnLog`, `errorLog`, `debugLog`... 

I'm not opposed to inject the loggers on each class... I think the main advantage of that is for testability though, so I think we might only want to do that if there is an important log we want to test.

As you said, we haven't been testing logs, but I think it might be a good thing to do, at least for some of the logs.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,200 @@
...
+                try {
+                    connection = urlConnectionFactory.createConnection(url.toString())
+
+                    if (connection.responseCode != HttpURLConnection.HTTP_OK) {
+                        throw IOException("HTTP ${connection.responseCode} when downloading file at: $url")
+                    }
+
+                    return@withContext connection.inputStream
+                } finally {
+                    connection?.disconnect()
```
tonidero: Hmm I'm actually a bit confused on this... this function returns an `InputStream`, but we're disconnecting the connection before reading the stream right? I think the stream should ideally download to a temp folder at least, so we can close the stream within the function. Alternatively, we shouldn't close that at all and make sure we do that on the call site after we're done reading.

tonidero (commented):
Looking good, just something I just noticed related to closing the stream before having the chance to read it. The rest seems good!

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: That's a good call out. I had copied the logic from the font loader, modified it for my use case, and didn't consider that…

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: That will work! Thanks.

What I generally do in Swift is await a `yield` on a lower priority job and it forces the structured concurrency model to pick up the work I want it to. 

I found a yield function in coroutines but it did not work like I wanted 🤣

JZDesign: Actually… that's not working for me.

JZDesign: I ended up needing to expose access to the deferred value store so I could await the job to make it deterministic. the combination of `scope.launch` which then triggered `scope.async` made `testScope.advanceUntilIdle` non-deterministic

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: The limit is interesting, I've not hit that kind of issue before, so I'm not positive what a sensible limit would be.


Retries are a great idea, I find that we only really want to retry in a few scenarios, like 500 level errors from the server, timeout errors, or an error storing the file (not creating the file uri)

totally open to that conversation and figuring out what are good triggers for retry (or inversely—what triggers will never retry)

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+import java.io.InputStream
+import java.net.HttpURLConnection
+import java.net.URI
+import java.net.URL
+import java.security.MessageDigest
+
+/**
+ * A file cache.
+ */
+internal interface FileRepositoryType {
```
marek-dabek: Nit: Do we need the Type suffix? Most classes I've seen use a interface BaseType and class SpecializedBaseType notation.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Create and/or get the cached file url.
+     * @param url The url for the remote data to cache into a file.
+     * @return The local file's URI where the data can be found after caching is complete.
+     */
+    suspend fun generateOrGetCachedFileURL(url: URL): URI
+}
+
+/**
+ * The file manager is a service capable of storing data and returning the URL where that stored data exists.
```
marek-dabek: File manager?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+import java.io.FileOutputStream
+import java.io.IOException
+import java.io.InputStream
+import java.net.HttpURLConnection
+import java.net.URI
+import java.net.URL
+import java.security.MessageDigest
+
+/**
+ * A file cache.
```
marek-dabek: Nit: I guess the comment might be redundant. Also, technically it's a repository, it just uses a cache.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+        /**
+         * Used when fetching the data fails.
+         */
+        class FailedToFetchFileFromRemoteSource(message: String) : Error(message)
+    }
+}
+
+private class FileCache(
+    private val context: Context,
+) : LargeItemCacheType {
```
marek-dabek: Nit: LargeItemCacheType sounds really vague. `LocalFileCache : FileCache`?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    override suspend fun generateOrGetCachedFileURL(url: URL): URI {
+        return store.getOrPut(url) {
+            scope.async {
+                val cachedUri = fileCacheManager.generateLocalFilesystemURI(remoteURL = url)
+                    ?: {
+                        val error = Error.FailedToCreateCacheDirectory(url.toString())
+                        logHandler.e("FileRepository", "Failed to create cache directory for $url", error)
```
marek-dabek: Not sure, not an Android person, but is there a Sentry-equivalent thing we use for Android. Or logging an error is enough? If this fails a lot it might be a really bad.

tonidero: We don't. We don't want to add big dependencies like that to our SDK. We do have our own `Diagnostics` service, that allows us to track desired events, but it's opt in on the part of developers.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+ * The file manager is a service capable of storing data and returning the URL where that stored data exists.
+ */
+internal interface LargeItemCacheType {
+    fun generateLocalFilesystemURI(remoteURL: URL): URI?
+    fun cachedContentExists(uri: URI): Boolean
+    fun saveData(data: InputStream, uri: URI)
+}
+
+internal class FileRepository(
+    @get:VisibleForTesting
```
tonidero: Nit: We can add this
```suggestion
    @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+    internal val store: KeyedDeferredValueStore<URL, URI> = KeyedDeferredValueStore<URL, URI>(),
+    private val fileCacheManager: LargeItemCacheType,
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO + NonCancellable),
```
tonidero: This is my opinion but I usually prefer to name the scope related to what type of dispatcher they are using. In this case, could be good to rename to `ioScope` so it's clearer.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
tonidero: I kinda agree.. The pattern we've been following in Android has been to name the interface the "normal" name, in this case `FileRepository`, and then name the implementation with a `Default` prefix, like `DefaultFileRepository`. This is the case for interfaces that are only really implemented once.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }.await()
+    }
+
+    private suspend fun downloadFile(url: URL): InputStream {
+        return try {
+            withContext(Dispatchers.IO) {
+                verboseLog { "Downloading remote font from $url" }
+
+                val connection = urlConnectionFactory.createConnection(url.toString())
```
tonidero: Hmm with this we're not disconnecting the connection anymore right? I think we need to refactor this so we disconnect once the download has finished. Either joining the download + saving to a file, so we disconnect once everything has been downloaded, or just save to a temp file when downloading, then later just move the file to the final location.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
```diff
@@ -0,0 +1,197 @@
...
+
+    override fun generateLocalFilesystemURI(remoteURL: URL): URI? {
+        val urlHash = md5Hex(remoteURL.toString().toByteArray())
+        val fileName = File(urlHash).name
+        if (fileName.isEmpty()) return null
+        return File(cacheDir, fileName).toURI()
+    }
+
+    override fun cachedContentExists(uri: URI): Boolean =
+        File(uri).exists()
```
tonidero: We discussed this but I'm still scared that we don't perform any verification that the download was performed successfully before saving the file... we could end up with a corrupted cache that would require uninstall/reinstall to come out from. I've seen this happen in fonts, and I fear it might happen more often on videos which are likely larger. Are we planning to add any sort of verification here so at least we don't end up stuck? @JZDesign @joshdholtz 

I know we talked about maybe calculate a checksum on the backend side that can be used to perform this verification... but not sure if there are better ways TBH.

purchases/src/test/java/com/revenuecat/purchases/storage/FileRepositoryTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        val fileRepository = FileRepository(
+            fileCacheManager = mockCache,
+            logHandler = mockk<LogHandler>(relaxed = true),
+            urlConnectionFactory = factory,
+        )
+
+        val result = fileRepository.generateOrGetCachedFileURL(url)
+        assertThat(result).isEqualTo(cacheUri)
+        verify(exactly = 1) { mockCache.saveData(any(), cacheUri) }
+
```
tonidero: Nit: Extra line

tonidero (commented):
This looks great! Just 2 blockers for me:
- We're not disconnecting from the connection once the file finishes downloading
- We're not adding any sort of verification on files downloaded that could end up with a corrupted state.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: Yea, me too usually. The iOS repo had this naming convention that I got feedback on and was asked to follow, so I used the same naming on Android to make it easier to transition between contexts.

Open to making changes, but I would like to standardize across repositories for contextual purposes If I can. 

That way if I know what the thing is called in iOS I know what it's called in android and visa versa

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: copy pasta from the iOS implementation 😆 I'll fix it

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: Isn't that the default?

tonidero: Yeah it’s the default… I guess we did it to make it more explicit and now most of our usages include it, so I would add it for consistency

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
tonidero: Hmm I agree a standard in naming would be good. Probably a good guild sdk meeting point? However, since this is the android repo, I would try to keep consistency within the codebase IMO. We do follow this pattern much less in android, so if we like to follow the iOS convention, we can always rename Android usages later

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
JZDesign: @tonidero I changed things up a bit, would you give that a look?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -0,0 +1,195 @@
...
+
+                if (connection.responseCode != HttpURLConnection.HTTP_OK) {
+                    throw IOException("HTTP ${connection.responseCode} when downloading file at: $url")
+                }
+
+                val bytes = connection.inputStream.use { inputStream ->
+                    inputStream.readBytes()
+                }
+
+                connection.disconnect()
```
tonidero: We might want to wrap a try-finally over the code reading from the stream to make sure we close if there is an error reading.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/FileRepository.kt
joshdholtz: @tonidero @JZDesign Ahhh, this is a good point! Yeah yeah, it probably is a good idea to store a checksum with the files that we are downloading to verify 🤔 

It's a little hard to support that going backward with the paywalls that are already created (like if/when we swap images to use this) but I think its a thing we should start doing. 

I'll get some tasks created to track this so we can implement for all the files that we are storing 👍

PR #2638, opened 2025-08-26
Feat/support new key

facumenzella:
### Motivation
We want to be more explicit on how to edit the button text for the ScreenOffering.
 
[ios](https://github.com/RevenueCat/purchases-ios/pull/5501)

### Description
- Add button text to ScreenOffering

> This comes already translated. For backwards compatibility, we default to the key used right now.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -476,11 +476,9 @@ private fun MainScreenContent(
     } else {
         configuration.getNoActiveScreen()?.let { noActiveScreen ->
             NoActiveUserManagementView(
-                screenTitle = noActiveScreen.title,
-                screenSubtitle = noActiveScreen.subtitle,
+                screen = noActiveScreen,
                 contactEmail = configuration.support.email,
                 localization = configuration.localization,
-                supportedPaths = noActiveScreen.paths,
```
vegaro, line 483: nice clean up

PR #2641, opened 2025-08-28
MON-1193 flatten Transition JSON structure after chatting more thoroughly with team

JZDesign:
### Checklist

### Motivation

PR #2623 merged a day too soon. After Josh got back from a long weekend we had more time to chat. There are existing patterns in place to keep the json flat. Fortunately, if the code from that PR merges to production it will fallback to a Fade transition which is what we need for the immediate ask anyway. We also hashed out some details on what is optional.

### Description

Flattened the json from :

```JSON
"type": {
    "type": "some_type",
    "maybe_never_would_add_another_field": "some_value"
 }
```

to:

```JSON
"type": "some_type",
"maybe_never_would_add_another_field": "some_value"
```

and made the int values required values in the json


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
```diff
@@ -43,8 +34,14 @@ class PaywallAnimation(
 }
 
 @OptIn(InternalRevenueCatAPI::class)
-object AnimationTypeAsObjectSerializer : EnumAsObjectSerializer<PaywallAnimation.AnimationType>(
-    enumClass = PaywallAnimation.AnimationType::class,
```
tonidero, line 47: Hmm just to check how would `EnumAsObjectSerializer` handle receiving a string value instead of an object? Just wondering if we would be falling back to the default paywall in that case...

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
```diff
@@ -43,8 +34,14 @@ class PaywallAnimation(
 }
 
 @OptIn(InternalRevenueCatAPI::class)
-object AnimationTypeAsObjectSerializer : EnumAsObjectSerializer<PaywallAnimation.AnimationType>(
```
tonidero, line 46: Do we need to delete `EnumAsObjectSerializer` as well? I believe it's unused now?

tonidero (approved):
Right. I think this makes sense... Just a question about what would the old implementation fallback to

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallAnimation.kt
JZDesign: It actually fell back to the default that was passed in.

PR #2642, opened 2025-08-28
Fix PaywallDialog going over screen size on Android 35+

tonidero:
### Description
There is an issue in our current version of compose that causes the Dialog composable to extend beyond the screen size in Android 35+ devices: https://issuetracker.google.com/issues/246909281. This was fixed in a newer version of Compose, however, that introduces a breaking change by forcing apps to compile against Android 35.

This PR introduces a workaround by applying some bottom padding that consists on the status bar + navigation bars sizes as bottom padding. It also makes the PaywallDialog go Edge to Edge on Android 35+, which is the default starting on that version. The current fix has been tested on Android <35 and >35 and also when the client app uses a higher version of compsoe with the fix.

| Issue in Android 35+ | Android < 35 | Android >= 35 with older compose version | Android >= 35 with newer compose version |
| ------ | ----- | ----- | ----- |
| image  | image | image  | image |

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -1,10 +1,15 @@
 package com.revenuecat.purchases.ui.revenuecatui
 
+import android.os.Build
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.WindowInsets
+import androidx.compose.foundation.layout.asPaddingValues
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.navigationBars
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.statusBars
 import androidx.compose.material3.Scaffold
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


vegaro:
Have you noticed the "Restore purchases" is almost hidden at the bottom? I think it's unrelated and probably happening in Android < 35 in `main`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -55,30 +64,48 @@ fun PaywallDialog(
...
+        containerColor = Color.Black.copy(alpha = 0.4f),
     ) { paddingValues ->
+        val shouldApplyDialogBottomPadding = paddingValues.calculateBottomPadding() == 0.dp &&
+            paddingValues.calculateTopPadding() == 0.dp
         Box(
             modifier = Modifier
                 .fillMaxSize()
-                .padding(paddingValues),
+                .conditional(Build.VERSION.SDK_INT <= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) { padding(paddingValues) }
+                .padding(bottom = if (shouldApplyDialogBottomPadding) dialogBottomPadding else 0.dp),
```
vegaro, line 112: is that `0.dp` maybe causing "Restore purchases" at the bottom to be almost behind the navigation bar?

tonidero: Nope, I think that's part of the paywall and was already happening in Android <35. The developer should add some margin in the editor to add a bit of spacing.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -55,30 +64,48 @@ fun PaywallDialog(
...
 }
 
 @Composable
-private fun DialogScaffold(paywallOptions: PaywallOptions) {
+private fun DialogScaffold(paywallOptions: PaywallOptions, dialogBottomPadding: Dp) {
     Scaffold(
         modifier = Modifier
             .fillMaxWidth()
             .fillMaxHeight(getDialogMaxHeightPercentage()),
+        containerColor = Color.Black.copy(alpha = 0.4f),
```
vegaro, line 104: Maybe we should add a comment here explaining why this is needed

vegaro (approved):
thank you for fixing it and all the testing

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
tonidero: Good point! Added in https://github.com/RevenueCat/purchases-android/pull/2642/commits/94bfca45ac7f297bf28b66fd67cf1b273255c672

PR #2643, opened 2025-08-28
Add option to disable automatic ID collection when setting attribution network IDs at configuration time

tonidero:
### Description
This adds a new API that can be used to disable automatic collection of Advertising ID, IP and device info when using one of the attribution networks we support like Adjust or Appsflyer. This can be set at configuration time like this:
```
val purchasesConfiguration = PurchasesConfiguration.Builder(context, "YOUR_API_KEY")
   .enableIdentifierCollectionWhenUsingAttributionNetwork(enabled)
   .build()
Purchases.configure(purchasesConfiguration)
```

Note that you can still collect these identifiers by calling `Purchases.sharedInstance.collectDeviceIdentifiers()`

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
@@ -35,6 +35,7 @@ open class PurchasesConfiguration(builder: Builder) {
     val dangerousSettings: DangerousSettings
     val verificationMode: EntitlementVerificationMode
     val pendingTransactionsForPrepaidPlansEnabled: Boolean
+    val enableIdentifierCollectionWhenUsingAttributionNetwork: Boolean
 
     init {
         this.context =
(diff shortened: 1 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -256,6 +265,21 @@ open class PurchasesConfiguration(builder: Builder) {
...
 
+        /**
+         * Enable this setting to allow the collection of identifiers when setting the identifier for an
+         * attribution network. For example, when calling [Purchases.setAdjustID] or [Purchases.setAppsflyerID],
+         * the SDK would collect the Android advertising ID, IP and device versions, if available, and send them
+         * to RevenueCat. This is required by some attribution networks to attribute installs and re-installs.
+         *
+         * Default is enabled.
+         */
+        fun enableIdentifierCollectionWhenUsingAttributionNetwork(
```
tonidero: I admit this is a very long name... But I wanted to make it clear what it does.

PR #2646, opened 2025-08-28
Video Component Models (dark code)

JZDesign:
### Checklist

### Motivation

Creating a new paywall component can be a VERY large amount of code, I'm breaking things apart to make the PR's easier to review. This PR is just the creation of the VideoComponent model.

### Description

Add a video component model and serialization tests

> [!Note]
> I did not make it a member of the PaywallComponent sealed interface. I intentionally left that part out so I wouldn't have to have a style factory function setup and all of the other stuff that would domino into the PR.
> This is just the json -> model

### DIFF
--- a/config/detekt/detekt-baseline.xml
+++ b/config/detekt/detekt-baseline.xml
@@ -1,6 +1,8 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <SmellBaseline>
-  <ManuallySuppressedIssues/>
+  <ManuallySuppressedIssues>
+    <ID>LongMethod:VideoComponentTests.kt$VideoComponentTests$@Test fun `deserializes correctly`()</ID>
+  </ManuallySuppressedIssues>
   <CurrentIssues>
     <ID>ComposableNaming:PaywallAPI.kt$PaywallAPI$check</ID>
     <ID>ComposableNaming:PaywallAPI.kt$PaywallAPI$checkDialog</ID>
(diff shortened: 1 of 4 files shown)


config/detekt/detekt-baseline.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <SmellBaseline>
-  <ManuallySuppressedIssues/>
+  <ManuallySuppressedIssues>
+    <ID>LongMethod:VideoComponentTests.kt$VideoComponentTests$@Test fun `deserializes correctly`()</ID>
+  </ManuallySuppressedIssues>
```
JZDesign:
```suggestion
  <ManuallySuppressedIssues/>
```

JZDesign: This was a mistake

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/properties/VideoUrls.kt
```diff
@@ -0,0 +1,36 @@
...
+@Serializable
+@Immutable
+class VideoUrls(
+    @get:JvmSynthetic
+    val width: UInt,
+    @get:JvmSynthetic
+    val height: UInt,
+    @get:JvmSynthetic
+    @Serializable(with = URLSerializer::class)
+    val url: URL,
```
tonidero, line 31: I wonder... should we allow developers to set a reference to an embedded video? So basically, follow a similar pattern as we do with fonts. If we find the name of the video in the resources for the app, we use that video. If not, we download it. That might help with bandwidth... But then again, it's probably a future improvement we can do without causing a breaking change, so I think this is ok for now.

JZDesign: That is an interesting question. @joshdholtz what do you think?

PR #2649, opened 2025-09-01
Added 'deepLinkID' to PaywallComponentsData

rickvdl:
### Checklist

### Motivation

This property was added to the iOS SDK [here](https://github.com/RevenueCat/purchases-ios/pull/5509), so adding it to Android as well for consistency.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PaywallComponentsData.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PaywallComponentsData.kt
@@ -35,4 +35,7 @@ class PaywallComponentsData(
     @Serializable(with = GoogleListSerializer::class)
     @SerialName("zero_decimal_place_countries")
     val zeroDecimalPlaceCountries: List<String> = emptyList(),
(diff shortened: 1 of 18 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PaywallComponentsData.kt
```diff
@@ -35,4 +35,7 @@ class PaywallComponentsData(
     @Serializable(with = GoogleListSerializer::class)
     @SerialName("zero_decimal_place_countries")
     val zeroDecimalPlaceCountries: List<String> = emptyList(),
+    @get:JvmSynthetic
+    @SerialName("deep_link_id")
+    val deepLinkID: String,
```
tonidero, line 40: I see in iOS we catch decoding errors here to avoid the whole paywall from failing to decode. Should we do the same in Android? We would need a custom serializer/deserializer and we should consider making the property nullable in that case, instead of defaulting to an empty string?

rickvdl:
This PR is no longer necessary since the RC mobile app has been refactored to use a different API

PR #2652, opened 2025-09-03
Display Virtual Currencies in Customer Center

fire-at-will:
### Description
Adds support for displaying virtual currency balances in the Customer Center. This mimics the design found in the iOS Customer Center as much as possible. It adds a list of virtual currency balances to the`RelevantPurchasesListView` and `NoActiveUserManagementView` views if:
- The app has virtual currencies
- The "Display virtual currency balances" toggle is enabled in the Customer Center configuration in the dashboard

### Changes
- Adds `displayVirtualCurrencies` to the CustomerCenterConfig
- Adds a list of virtual currency balances to the `RelevantPurchasesListView` and `NoActiveUserManagementView`. It displays a max of 4 virtual currencies. If there are >4 virtual currencies, it will display the 3 VCs with the highest balance, and show a "See all in-app currencies" button that navigates to a full VC balances screen
- Adds a screen that displays all virtual currency balances. It refreshes the VC balances when this screen is loaded in case the balances have changed since they were last loaded.

### Testing
- Tested manually
- Added automated tests for the new view model functions

### Screenshots
On the `RelevantPurchasesListView`, with just 2 virtual currencies (no "see all" button):
Screenshot 2025-09-05 at 2 12 25 PM

On the `RelevantPurchasesListView`, with just 5 virtual currencies ("see all" button is visible):
Screenshot 2025-09-05 at 2 17 12 PM

VC Balances screen (accessed by tapping "See all in-app currencies" button)
Screenshot 2025-09-05 at 2 18 03 PM

On the `NoActiveUserManagementView` screen:
Screenshot 2025-09-05 at 2 19 20 PM


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -350,18 +354,6 @@ private fun CustomerCenterNavigationIcon(
     }
 }
 
-@Composable
```
fire-at-will, line 358: `CustomerCenterLoading` and `CustomerCenterError` were extracted out to separate files so they could be reused in the `VirtualCurrencyBalancesScreen` 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CustomerCenterErrorView.kt
```diff
@@ -0,0 +1,33 @@
+package com.revenuecat.purchases.ui.revenuecatui.customercenter.views
+
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview
+import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesErrorCode
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCenterState
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.theme.CustomerCenterPreviewTheme
+
```
fire-at-will, line 10: The only new things in `CustomerCenterErrorView` and `CustomerCenterLoadingView` are:
- Added `View` suffix to the names
- Made them `internal` instead of `private`
- Added previews

This allows us to reuse these views between the `InternalCustomerCenter` and the `VirtualCurrencyBalancesScreen` 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrenciesListView.kt
```diff
@@ -0,0 +1,265 @@
...
+
+@Composable
+internal fun VirtualCurrencyRow(
+    appearance: CustomerCenterConfigData.Appearance,
+    virtualCurrency: VirtualCurrency,
+    position: ButtonPosition,
+    modifier: Modifier = Modifier,
+) {
+    val formattedBalance = remember(virtualCurrency.balance) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
```
fire-at-will, line 122: I wanted to use `NumberFormatter` to format the VC's balance, but it looks like it was added in API level 30 ([source](https://developer.android.com/reference/android/icu/number/NumberFormatter)). This is what Claude came up with to ensure that the API is available - is this the best way to do this?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/RelevantPurchasesListView.kt
```diff
@@ -74,6 +76,22 @@ internal fun RelevantPurchasesListView(
             )
         }
 
+        virtualCurrencies?.let {
+            if (virtualCurrencies.all.isNotEmpty()) {
+                Spacer(modifier = Modifier.size(CustomerCenterConstants.Layout.SECTION_SPACING))
+
+                VirtualCurrenciesListView(
+                    virtualCurrencies = virtualCurrencies,
+                    appearance = CustomerCenterConfigTestData.standardAppearance,
```
facumenzella: We should use the appearance from the config, not the standard one. InternalCustomerCenter has `state.customerCenterConfigData.appearance`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -478,8 +478,10 @@ private fun MainScreenContent(
             NoActiveUserManagementView(
                 screen = noActiveScreen,
                 contactEmail = configuration.support.email,
+                appearance = configuration.appearance,
                 localization = configuration.localization,
                 offering = (state as? CustomerCenterState.Success)?.noActiveScreenOffering,
+                virtualCurrencies = (state as? CustomerCenterState.Success)?.virtualCurrencies,
```
facumenzella: `state` is already `CustomerCenterState.Success` 👀

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrenciesListView.kt
```diff
@@ -0,0 +1,265 @@
...
+            NumberFormatter.with()
+                .locale(Resources.getSystem().configuration.locales.get(0))
+                .format(virtualCurrency.balance)
+                .toString()
```
facumenzella: Can't we use the same one used inside `Purchases`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
vegaro: right, from the parameters of the function

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrencyBalancesScreen.kt
```diff
@@ -0,0 +1,271 @@
...
+internal fun VirtualCurrencyBalancesScreen(
+    appearance: CustomerCenterConfigData.Appearance,
+    localization: CustomerCenterConfigData.Localization,
+    modifier: Modifier = Modifier,
+    viewModel: VirtualCurrencyBalancesScreenViewModel = getVirtualCurrencyBalancesScreenViewModel(),
+) {
+    val viewState by viewModel.viewState.collectAsState()
+    LaunchedEffect(Unit) {
+        viewModel.onViewAppeared()
+    }
```
vegaro: This is considered an anti-pattern. Take a look at this blogpost https://proandroiddev.com/loading-initial-data-in-launchedeffect-vs-viewmodel-f1747c20ce62 by @skydoves 

The screen calls `onViewAppeared()` from `LaunchedEffect(Unit)`, so every time the composable re-enters composition you kick off `loadData()` again.

The article recommends lazy loading via `Flow`: start work when there’s a subscriber, not from `LaunchedEffect`. So basically, using `stateIn(..., SharingStarted.WhileSubscribed(...))` in the view model and collect with `collectAsStateWithLifecycle()`

vegaro (changes requested):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
fire-at-will: Removed the unnecessary casting for both the `offering` and `virtualCurrencies` params here in 365bffcde7fe0c7440ddc14a1fa6f79c6212f202 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/RelevantPurchasesListView.kt
fire-at-will: Good catch! We're now passing the appearance down to this view - change made in e793ab19766e33dbbec6c75888de42f338c07cdd 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrenciesListView.kt
fire-at-will: @facumenzella Good idea! It looks like we can either:
- Use the locale string from `PurchasesType.preferredUILocaleOverride` - this will require us to push down either `PurchasesType` or the value of `PurchasesType.preferredUILocaleOverride` to this view
- The locale string from `CustomerCenterConfigData.Localization.locale` - this would be easier since `localization` is available here. 

It feels like `PurchasesType.preferredUILocaleOverride` might be better since it'll let developers override the locale used for rendering - do you have a preference either way?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrencyBalancesScreen.kt
fire-at-will: Thanks @vegaro, I learned something new today 😄

I've refactored the logic of loading the balances to use a flow with `stateIn` and `SharingStarted.WhileSubscribed` in this commit: 9c19e267186a1b0695dd41bf66a5846a1d33d14d

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/VirtualCurrencyBalancesScreenViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@Stable
+internal class VirtualCurrencyBalancesScreenViewModel(
+    private val purchases: PurchasesType,
+) : ViewModel() {
+
+    val viewState: StateFlow<VirtualCurrencyBalancesScreenViewState> = flow {
+        emit(VirtualCurrencyBalancesScreenViewState.Loading)
+
+        purchases.invalidateVirtualCurrenciesCache()
```
vegaro, line 26: I am not completely sure, but I believe every time you go back to the screen (put app in background then come back to the app) this flow will be triggered, and the virtual currencies will be fetched again. Not sure if that's intended either

fire-at-will: Yeah, this is intentional and matches the behavior in the iOS Customer Center. The reasoning behind that is that, unlike things like entitlements, VC balances can change often, and we want to make sure we're always showing the latest balances when the user goes to look at them.

One quirk here that isn't present on iOS is that due to the `started = SharingStarted.WhileSubscribed(STOP_TIMEOUT_MILLIS)` in
```kotlin
.stateIn(
    scope = viewModelScope,
    started = SharingStarted.WhileSubscribed(STOP_TIMEOUT_MILLIS),
    initialValue = VirtualCurrencyBalancesScreenViewState.Loading, 
)
```
the flow doesn't get re-run if the user navigates away from the page and then back to it in less than 5 seconds. We're doing this in a few other places in the Customer Center, notably [here in the CustomerCenterViewModel](https://github.com/RevenueCat/purchases-android/blob/3e6f654e4fd6a8516a8176df5185b51aea45d864/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt#L180)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrenciesListView.kt
facumenzella: Yeah, I think that makes sense. 
Let's use `CustomerCenterConfigData.Localization.locale` to be consistent within customer center.

fire-at-will: Cool, I'll use that one!

fire-at-will: Updated this to use the locale from the Localization here: e35281230eb7709700b5c43d92a9e876d1bd0f2e 👍

PR #2653, opened 2025-09-03
Remove validation for no packages on paywalls

joshdholtz:
### Motivation

Don't validate if no package components on paywall

### Description

This shouldn't be a SDK validation but a server side one

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
@@ -38,7 +38,6 @@ internal sealed class PaywallValidationError : Throwable() {
             is AllLocalizationsMissing -> message
             is AllVariableLocalizationsMissing -> message
             is MissingPackage -> message
-            is MissingAllPackages -> message
             is MissingColorAlias -> message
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
```diff
@@ -93,14 +92,6 @@ internal sealed class PaywallValidationError : Throwable() {
             PaywallValidationErrorStrings.MISSING_PACKAGE
                 .format(missingPackageId, offeringId, allPackageIds.joinToString())
     }
-    data class MissingAllPackages(
```
tonidero, line 96: Might be good having a paywall without any packages so we can confirm it works as expected with no packages but from looking at the code and from what I remember, I don't think there should be any issues later 👍

tonidero (approved):
This seems ok 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/errors/PaywallValidationError.kt
joshdholtz: I did test this locally and we will have a new template that will have this that will get added to our cross platform validation soon 💪

PR #2655, opened 2025-09-03
Improve thread safety of setting paywalls preferred locale

joshdholtz:
### Motivation

Improve thread safety of setting preferred locale

### Description

Make the variable `@Volatile`


tonidero (approved):
I think this should be safe enough! Didn't see any of issues with it 👍

PR #2658, opened 2025-09-04
Support setting null offering id on PaywallView

tonidero:
### Description
Follow-up to #2610. We had a non-null offeringIdentifier parameter, which doesn't allow us to reset the offering selection to none. This would be very uncommon, but we should still support it.


rickvdl (approved):
Nice!

PR #2659, opened 2025-09-09
Fix CoroutineCreationDuringComposition lint error on AGP 8.13.0

vegaro:
Fixes https://github.com/RevenueCat/purchases-android/pull/2588#issuecomment-3149306146

A new lint issue that starts happening when upgrading to AGP 8.13.0

Also wrapped the `refreshStateIfColorsChanged` in a LaunchedEffect keyed by the inputs, so it only runs when the theme actually changes.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
@@ -81,7 +81,7 @@ internal fun InternalCustomerCenter(
     val isDark = isSystemInDarkTheme()
 
     LaunchedEffect(colorScheme, isDark) {
-        viewModel.refreshStateIfColorsChanged(colorScheme, isDark)
+        viewModel.refreshColors(colorScheme, isDark)
     }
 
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -77,14 +77,19 @@ internal fun InternalCustomerCenter(
     ),
     onDismiss: () -> Unit,
 ) {
-    viewModel.refreshStateIfColorsChanged(MaterialTheme.colorScheme, isSystemInDarkTheme())
+    val colorScheme = MaterialTheme.colorScheme
+    val isDark = isSystemInDarkTheme()
+
+    LaunchedEffect(colorScheme, isDark) {
```
tonidero, line 83: This is good! I wonder if we could remove the checks we do on the `refreshStateIfColorsChanged` method that check that the values haven't changed after this change... but I guess they won't hurt either so FFTI.

vegaro: nice! I will make the change now, otherwise I will forget lol

PR #2668, opened 2025-09-11
[Experimental]: Support Multi-Line Subscriptions

fire-at-will:
### Description
Adds support for purchasing add-on subscriptions in the `purchases-android` library. Through the APIs introduced in this PR, developers can purchase subscriptions with add-ons via a new `PurchaseParams.Builder` function that accepts a `List<StoreProduct>` or `List<Package>` of the add on products/packages that the user would like to purchase:

```kotlin
fun purchaseSubscriptionWithAddOns() {
    val purchaseParams = Builder(
        activity = activity,
        storeProduct = baseProduct
    )
        .addOnProducts(listOf(product2, product3))
        .build()

   val purchaseParams = Builder(
        activity = activity,
        packageToPurchase = basePackage
    )
        .addOnPackages(listOf(package2, package3))
        .build()

    Purchases.sharedInstance.purchase(
        purchaseParams = purchaseParams,
        callback = // Omitted for brevity
    )
}
```

Additionally, this PR introduces a new property to `StoreTransaction`: `val subscriptionOptionIdForProductIDs: Map<String, String>?`. This property contains a mapping of productIDs to the subscription option ID that was purchased. This is helpful in two scenarios:
- It allows developers to know which subscription options were purchased in a multi-line subscription
- It allows the SDK to include subscription option IDs for multi-line add-ons when POSTing the receipt to the RC backend.

### Notes
The new `PurchaseParams` function and `StoreTransaction.subscriptionOptionIdForProductIDs` are marked as `ExperimentalPreviewRevenueCatPurchasesAPI`, so it will require developers to opt-in to using an experimental API, as shown in the code snippet. We can remove in the future when the feature is more stable.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -111,6 +111,26 @@ internal class HTTPClient(
         fallbackBaseURLs: List<URL> = emptyList(),
         fallbackURLIndex: Int = 0,
     ): HTTPResult {
+        fun canUseFallback(): Boolean =
+            endpoint.supportsFallbackBaseURLs && fallbackURLIndex in fallbackBaseURLs.indices
+
+        fun performRequestToFallbackURL(): HTTPResult {
+            val fallbackBaseURL = fallbackBaseURLs[fallbackURLIndex]
+            log(LogIntent.DEBUG) {
+                NetworkStrings.RETRYING_CALL_WITH_FALLBACK_URL.format(endpoint.getPath(), fallbackBaseURL)
+            }
+            return performRequest(
+                fallbackBaseURL,
+                endpoint,
+                body,
+                postFieldsToSign,
+                requestHeaders,
+                refreshETag,
+                fallbackBaseURLs,
+                fallbackURLIndex + 1,
+            )
+        }
+
         if (appConfig.forceServerErrors) {
             warnLog { "Forcing server error for request to ${endpoint.getPath()}" }
             return HTTPResult(
@@ -127,6 +147,9 @@ internal class HTTPClient(
         try {
             callResult = performCall(baseURL, endpoint, body, postFieldsToSign, requestHeaders, refreshETag)
             callSuccessful = true
+        } catch (e: IOException) {
+            // Handle connection failures with fallback URLs
+            if (canUseFallback()) callResult = performRequestToFallbackURL() else throw e
         } finally {
             trackHttpRequestPerformedIfNeeded(
                 baseURL,
@@ -149,25 +172,9 @@ internal class HTTPClient(
                 fallbackBaseURLs,
                 fallbackURLIndex,
             )
-        } else if (RCHTTPStatusCodes.isServerError(callResult.responseCode) &&
-            endpoint.supportsFallbackBaseURLs &&
-            fallbackURLIndex in fallbackBaseURLs.indices
-        ) {
+        } else if (RCHTTPStatusCodes.isServerError(callResult.responseCode) && canUseFallback()) {
             // Handle server errors with fallback URLs
-            val fallbackBaseURL = fallbackBaseURLs[fallbackURLIndex]
-            log(LogIntent.DEBUG) {
-                NetworkStrings.RETRYING_CALL_WITH_FALLBACK_URL.format(endpoint.getPath(), fallbackBaseURL)
-            }
-            callResult = performRequest(
-                fallbackBaseURL,
-                endpoint,
-                body,
-                postFieldsToSign,
-                requestHeaders,
-                refreshETag,
-                fallbackBaseURLs,
-                fallbackURLIndex + 1,
-            )
+            callResult = performRequestToFallbackURL()
         }
         return callResult
     }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
@@ -1,7 +1,9 @@
 package com.revenuecat.purchases.common.networking
 
 import com.revenuecat.purchases.VerificationResult
+import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.isSuccessful
+import org.json.JSONException
 import org.json.JSONObject
 import java.util.Date
 
@@ -50,7 +52,17 @@ internal data class HTTPResult(
         BACKEND, CACHE
     }
 
-    val body: JSONObject = payload.takeIf { it.isNotBlank() }?.let { JSONObject(it) } ?: JSONObject()
+    val body: JSONObject = payload
+        .takeIf { it.isNotBlank() }
+        ?.let {
+            try {
+                JSONObject(it)
+            } catch (e: JSONException) {
+                errorLog(throwable = e) { "Failed to parse payload as JSON: $it" }
+                null
+            }
+        }
+        ?: JSONObject()
 
     val backendErrorCode: Int? = if (!isSuccessful()) body.optInt("code").takeIf { it > 0 } else null
     val backendErrorMessage: String? = if (!isSuccessful()) {
--- a/purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
@@ -26,7 +26,7 @@ import org.json.JSONException
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
-import java.net.URL
+import org.robolectric.ParameterizedRobolectricTestRunner
 import java.util.Date
 import kotlin.time.Duration.Companion.milliseconds
 import org.robolectric.annotation.Config as AnnotationConfig
@@ -84,23 +84,6 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
         server.takeRequest()
 
         assertThat(result.body.getString("response")).`as`("response is OK").isEqualTo("OK")
-    }
-
-    // Errors
-
-    @Test(expected = JSONException::class)
-    fun reWrapsBadJSONError() {
-        val endpoint = Endpoint.LogIn
-        enqueue(
-            endpoint,
-            expectedResult = HTTPResult.createResult(payload = "not uh jason")
-        )
-
-        try {
-            client.performRequest(baseURL, endpoint, body = null, postFieldsToSign = null, mapOf("" to ""))
-        } finally {
-            server.takeRequest()
-        }
     }
 
     // region forceServerErrors
@@ -974,6 +957,168 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
             )
         }
     }
+}
 
-    // endregion Fallback API host
+@RunWith(ParameterizedRobolectricTestRunner::class)
+internal class ParameterizedNonJsonResponseBodyTest(
+    private val endpoint: Endpoint,
+    private val statusCode: Int,
+) : BaseHTTPClientTest() {
+
+    companion object {
+        @JvmStatic
+        @ParameterizedRobolectricTestRunner.Parameters(name = "endpoint={0}, statusCode={1}")
+        fun parameters(): Collection<Array<Any>> {
+            return listOf(
+                arrayOf(Endpoint.GetOfferings("test_user"), 500),
+                arrayOf(Endpoint.GetOfferings("test_user"), 503),
+                arrayOf(Endpoint.GetOfferings("test_user"), 504),
+                arrayOf(Endpoint.GetProductEntitlementMapping, 500),
+                arrayOf(Endpoint.GetProductEntitlementMapping, 503),
+                arrayOf(Endpoint.GetProductEntitlementMapping, 504),
+            )
+        }
+    }
+
+    @Before
+    fun setupClient() {
+        mockSigningManager = mockk()
+        every { mockSigningManager.shouldVerifyEndpoint(any()) } returns false
+        client = createClient()
+    }
+
+    @Test
+    fun `performRequest should retry with fallback URL when server returns non-JSON response`() {
+        // Arrange
+        assert(endpoint.supportsFallbackBaseURLs) {
+            "This test is only meant to test endpoints supporting fallback URLs."
+        }
+        val fallbackServer = MockWebServer()
+        val fallbackBaseURL = fallbackServer.url("/v1").toUrl()
+        val invalidJsonPayload = "<html><body>504 Gateway Timeout</body></html>"
+        val validJsonPayload = """{"offerings": [], "current_offering_id": null}"""
+        val mainResponse = MockResponse()
+            .setBody(invalidJsonPayload)
+            .setResponseCode(statusCode)
+        val fallbackResponse = MockResponse()
+            .setBody(validJsonPayload)
+            .setResponseCode(RCHTTPStatusCodes.SUCCESS)
+        server.enqueue(mainResponse)
+        fallbackServer.enqueue(fallbackResponse)
+        every {
+            mockETagManager.getHTTPResultFromCacheOrBackend(
+                statusCode,
+                invalidJsonPayload,
+                eTagHeader = any(),
+                urlPath = endpoint.getPath(),
+                refreshETag = false,
+                requestDate = any(),
+                verificationResult = VerificationResult.NOT_REQUESTED
+            )
+        } returns HTTPResult.createResult(statusCode, invalidJsonPayload)
+        every {
+            mockETagManager.getHTTPResultFromCacheOrBackend(
+                RCHTTPStatusCodes.SUCCESS,
+                validJsonPayload,
+                eTagHeader = any(),
+                urlPath = endpoint.getPath(),
+                refreshETag = false,
+                requestDate = any(),
+                verificationResult = VerificationResult.NOT_REQUESTED
+            )
+        } returns HTTPResult.createResult(RCHTTPStatusCodes.SUCCESS, validJsonPayload)
+
+        // Act
+        try {
+            val result = client.performRequest(
+                baseURL,
+                endpoint,
+                body = null,
+                postFieldsToSign = null,
+                mapOf("" to ""),
+                fallbackBaseURLs = listOf(fallbackBaseURL),
+            )
+
+            // Assert
+            assertThat(server.requestCount).isEqualTo(1)
+            assertThat(fallbackServer.requestCount).isEqualTo(1)
+            assertThat(result.responseCode).isEqualTo(RCHTTPStatusCodes.SUCCESS)
+            assertThat(result.payload).isEqualTo(validJsonPayload)
+            assertThat(result.body.has("offerings")).isTrue
+        } finally {
+            fallbackServer.shutdown()
+        }
+    }
+}
+
+@RunWith(ParameterizedRobolectricTestRunner::class)
+internal class ParameterizedConnectionFailureFallbackTest(
+    private val endpoint: Endpoint,
+) : BaseHTTPClientTest() {
+
+    companion object {
+        @JvmStatic
+        @ParameterizedRobolectricTestRunner.Parameters(name = "endpoint={0}")
+        fun parameters(): Collection<Array<Any>> {
+            return listOf(
+                arrayOf(Endpoint.GetOfferings("test_user")),
+                arrayOf(Endpoint.GetProductEntitlementMapping),
+            )
+        }
+    }
+
+    @Before
+    fun setupClient() {
+        mockSigningManager = mockk()
+        every { mockSigningManager.shouldVerifyEndpoint(any()) } returns false
+        client = createClient()
+    }
+
+    @Test
+    fun `performRequest should retry with fallback URL when connection fails`() {
+        // Arrange
+        assert(endpoint.supportsFallbackBaseURLs) {
+            "This test is only meant to test endpoints supporting fallback URLs."
+        }
+        val fallbackServer = MockWebServer()
+        val fallbackBaseURL = fallbackServer.url("/v1").toUrl()
+        val validJsonPayload = """{"offerings": [], "current_offering_id": null}"""
+        // Shut down main server to cause IOException when client tries to connect
+        server.shutdown()
+        val fallbackResponse = MockResponse()
+            .setBody(validJsonPayload)
+            .setResponseCode(RCHTTPStatusCodes.SUCCESS)
+        fallbackServer.enqueue(fallbackResponse)
+        every {
+            mockETagManager.getHTTPResultFromCacheOrBackend(
+                RCHTTPStatusCodes.SUCCESS,
+                validJsonPayload,
+                eTagHeader = any(),
+                urlPath = endpoint.getPath(),
+                refreshETag = false,
+                requestDate = any(),
+                verificationResult = VerificationResult.NOT_REQUESTED
+            )
+        } returns HTTPResult.createResult(RCHTTPStatusCodes.SUCCESS, validJsonPayload)
+
+        // Act
+        try {
+            val result = client.performRequest(
+                baseURL,
+                endpoint,
+                body = null,
+                postFieldsToSign = null,
+                mapOf("" to ""),
+                fallbackBaseURLs = listOf(fallbackBaseURL),
+            )
+
+            // Assert
+            assertThat(fallbackServer.requestCount).isEqualTo(1)
+            assertThat(result.responseCode).isEqualTo(RCHTTPStatusCodes.SUCCESS)
+            assertThat(result.payload).isEqualTo(validJsonPayload)
+            assertThat(result.body.has("offerings")).isTrue
+        } finally {
+            fallbackServer.shutdown()
+        }
+    }
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -261,6 +261,18 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
                     } else {
                         viewModel.handlePackagePurchase(activity, pkg = action.rcPackage)
                     }
+                is PaywallAction.External.LaunchWebCheckout -> {
+                    val url = viewModel.getWebCheckoutUrl(action)
+                    if (url == null) {
+                        Logger.e("Web checkout URL cannot be found, not launching web checkout.")
+                    } else {
+                        context.handleUrlDestination(url, action.openMethod)
+                        if (action.autoDismiss) {
+                            Logger.d("Auto-dismissing paywall after launching web checkout.")
+                            viewModel.closePaywall()
+                        }
+                    }
+                }
 
                 is PaywallAction.External.NavigateBack -> viewModel.closePaywall()
                 is PaywallAction.External.NavigateTo -> when (val destination = action.destination) {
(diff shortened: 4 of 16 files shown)


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OfferingFragment.kt
```diff
@@ -130,6 +198,76 @@ class OfferingFragment : Fragment(), PackageCardAdapter.PackageCardAdapterListen
...
+                    }
+                }
+            },
+        )
+    }
+
+    private fun startAddOnPurchaseWithoutFinishingTransaction(selectedPackages: List<Package>) {
+        // For non-finishing transactions, we'll handle each package individually
+        // This is a simplified approach - in a real implementation you might want to handle this differently
+        selectedPackages.forEach { pkg ->
```
fire-at-will, line 266: I think this is wrong...I'll update this to make an add-on purchase instead of purchasing them individually

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -286,24 +294,26 @@ internal class BillingWrapper(
...
-                        replaceProductInfo?.oldPurchase?.productIds?.firstOrNull(),
-                    )
-                    launchBillingFlow(activity, result.value)
+        val result = buildPurchaseParams(
+            purchasingData,
+            replaceProductInfo,
+            appUserID,
+            isPersonalizedPrice,
+            onCompletion = { result ->
+                executeRequestOnUIThread {
```
MarkVillacampa, line 303: is moving `executeRequestOnUIThread` to the completion block intentional?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -906,6 +931,135 @@ internal class BillingWrapper(
...
+            is Result.Error -> {
+                onCompletion(productDetailsParamsListResult)
+                return@buildProductWithAddOnsPurchaseParams
+            }
+            is Result.Success<List<ProductDetailsParams>> -> {
+                productDetailsParamsList = productDetailsParamsListResult.value
+            }
+        }
+
+        // Determine if the base product has been purchased before
```
MarkVillacampa: To understand this logic, this is for the case where the developer is adding/removing add-ons for an existing subscription?

Currently for product changes, we detect it's a product change by the presence is `oldProductId` in the PurchaseParams instance, but here we wouldn't have it, unless we document explicitly that developers need to set it even if the base product is not changing, then we wouldn't need this logic?

purchases/src/main/kotlin/com/revenuecat/purchases/models/GooglePurchasingData.kt
```diff
@@ -27,5 +27,19 @@ sealed class GooglePurchasingData : PurchasingData {
             is Subscription -> {
                 ProductType.SUBS
             }
+            is ProductWithAddOns -> {
+                // Can be either INAPP or SUBS. Since all of the types in the
+                // multi-line purchase must match, we use the type from the
+                // base product.
+                baseProduct.productType
```
MarkVillacampa: Only subscriptions are supported right?

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
```diff
@@ -48,10 +49,30 @@ internal class PostTransactionWithProductDetailsHelper(
                                     product.id == transaction.productIds.firstOrNull()
                                 }
                             }
+
+                        val subscriptionOptionsForGoogleProductIDs = transaction
```
fire-at-will, line 54: When we generate the ReceiptInfo in `postReceiptHelper.postTransactionAndConsumeIfNeeded()`, we need a way to get the subscription options for the add-on purchases so they can be included when we post the receipt to the backend. Building this `subscriptionOptionsForGoogleProductIDs` map here allows us to do that :)

purchases/src/main/kotlin/com/revenuecat/purchases/google/PurchaseContext.kt
```diff
@@ -9,4 +9,5 @@ internal class PurchaseContext(
     val presentedOfferingContext: PresentedOfferingContext?,
     val selectedSubscriptionOptionId: String?,
     val replacementMode: GoogleReplacementMode?,
+    val subscriptionOptionIdForProductIDs: Map<String, String>?,
```
fire-at-will, line 12: This allows us to access which subscription options for add-ons were purchased for which product IDs after a purchase has completed, which then lets us send that information to the RC backend and to the developers using the SDK.

purchases/src/main/kotlin/com/revenuecat/purchases/models/GoogleSubscriptionOption.kt
```diff
@@ -128,5 +128,6 @@ class GoogleSubscriptionOption @JvmOverloads constructor(
             id,
             productDetails,
             offerToken,
+            addOnProducts = emptyList(),
```
fire-at-will, line 137: I believe this is fine because a subscription option shouldn't ever have add-ons, but let me know if there's something here I may be missing!

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
```diff
@@ -106,6 +114,41 @@ class StoreTransaction(
     val replacementMode: ReplacementMode?,
 ) : Parcelable {
 
+    constructor(
```
fire-at-will, line 126: Here, we recreate the old StoreTransaction constructor without the `subscriptionOptionIdForProductIDs` to avoid making a breaking change.

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
```diff
@@ -17,7 +18,8 @@ internal class PostTransactionWithProductDetailsHelper(
     /**
      * The callbacks in this method are called for each transaction in the list.
      */
-    @Suppress("LongParameterList")
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
```
fire-at-will, line 21: I was a bit surprised that opting in to `ExperimentalPreviewRevenueCatPurchasesAPI` didn't trickle up the call chain from here, but it doesn't appear to have for some reason 🤷‍♂️

purchases/src/main/kotlin/com/revenuecat/purchases/google/ProductValidation.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        throw PurchasesException(error)
+    }
+    val addOnProductsWithSameProductType: MutableList<GooglePurchasingData> = ArrayList()
+    val billingPeriods = mutableSetOf<Period>()
+
+    for (addOnProduct in addOnProducts) {
+        val addOnPurchasingData = addOnProduct.purchasingData
+        if (addOnPurchasingData is GooglePurchasingData &&
+            addOnPurchasingData::class == baseProductPurchasingData::class
```
tonidero: Hmm instead of checking the class like this, can we check that it's a `GooglePurchasingData.Subscription`? Feels cleaner to me...

purchases/src/main/kotlin/com/revenuecat/purchases/google/ProductValidation.kt
```diff
@@ -0,0 +1,77 @@
...
+                "(1 base + $MAX_NUMBER_OF_ADD_ON_PRODUCTS add-ons).",
+        ).also { errorLog(it) }
+
+        throw PurchasesException(error)
+    }
+
+    if (billingPeriods.size > 1) {
+        val error = PurchasesError(
+            PurchasesErrorCode.PurchaseInvalidError,
+            "All items in a multi-line purchase must have the same billing period.",
```
tonidero: Just curious, what happens if one of the add-on products has a free trial/intro price with a different period? Does that get ignored?

purchases/src/main/kotlin/com/revenuecat/purchases/google/ProductValidation.kt
```diff
@@ -0,0 +1,77 @@
...
+        ).also { errorLog(it) }
+
+        throw PurchasesException(error)
+    }
+
+    return addOnProductsWithSameProductType
```
tonidero: Should we also validate that the add-on products are not part of the same subscription in Google? (so not a different base plan within the same subscription). I guess you can potentially have multiple base plans with the same period but you can't buy that same product multiple times right?

purchases/src/main/kotlin/com/revenuecat/purchases/models/GoogleSubscriptionOption.kt
tonidero: Hmm well... I guess a developer might potentially want to add add-ons even when selecting a specific subscription option to purchase. A subscription option is just a way to select the specific offer that developers want within a base plan but doesn't exclude supporting add-ons IMO...

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -104,10 +107,8 @@ internal class PostReceiptHelper(
             marketplace = purchase.marketplace,
             initiationSource = initiationSource,
             onSuccess = { postReceiptResponse ->
-                // Currently we only support a single token per postReceipt call but multiple product Ids.
-                // The backend would fail if given more than one product id (multiline purchases which are
-                // not supported) so it's safe to pickup the first one.
-                // We would need to refactor this if/when we support multiple tokens per call.
+                // Currently we only support a single token per postReceipt call but multiple product Ids
+                // (for multi-line subscriptions).
```
tonidero, line 111: For the acknowledgement logic, I guess if we acknowledge the purchase that includes the add-ons, that will acknowledge all the subscriptions, correct?

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
tonidero: Hmm just a note that this might not work when restoring purchases... How critical is it to have this information? Just wondering if there would be issues in the case of purchasing one of these multi line subscriptions but the request failing, then us retrying to post it on next app open without this information.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
```diff
@@ -98,6 +98,16 @@ class StoreTransaction(
...
 
+    /**
+     * The id of the SubscriptionOption purchased for each product ID.
+     *
+     * In Google, this will be calculated from the basePlanId and offerId
+     * Null in Google for restored transactions and purchases initiated outside of the app.
+     * Null for Amazon purchases.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    val subscriptionOptionIdForProductIDs: Map<String, String>?,
```
tonidero, line 115: [Not related to this line!] 

Just thinking, for offline entitlements mode, we might need to also change the code to consider all the purchased subscriptions? TBH, I'm not familiarized in how these purchases are queried from Google, so if they appear as separate purchases, it might be ok as is, but wanted to confirm.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -93,6 +95,7 @@ internal class PostReceiptHelper(
             presentedOfferingContext = purchase.presentedOfferingContext,
```
tonidero, line 95: Hmm I'm thinking that the presented offering context we post to the backend will always be for the "base" package/product and won't include the add-ons if they were obtained from a different placement or something... I think it might be an acceptable compromise for now but I think we should be aware of this limitation.

tonidero (commented):
Looking great! Just some high level questions to understand this better before approving :)

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
tonidero: We could deprecate this constructor. We probably would keep it, but that way we ask people to move to the new one.

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
fire-at-will: Great point! My understanding is that Khepri uses the `product_ids` param in the POST body to perform its calculations, so it doesn't need this information. However, fortress relies on the `platform_product_ids` param because it needs both the `product_id` and the `base_plan_id`.

I'll look into the following here:
- Seeing if it's possible to get that information for a restore. You're probably right and it may not be available, but I'll see if I can find a way.
- If it's not possible, I'll document that `StoreTransaction.subscriptionOptionForGoogleProductIDs` isn't available for restores.

cc @greenietea just to make sure that our assumptions here about what the backend needs is correct :)

purchases/src/main/kotlin/com/revenuecat/purchases/google/ProductValidation.kt
fire-at-will: I agree that just checking for it being a `GooglePurchasingData.Subscription` would be cleaner. The reason I wrote it this way is so that it'll be one less thing that we need to change if/when we decide to support [multi-products for one-time purchases](https://developer.android.com/google/play/billing/multi-product-for-one-time-product). 

With the logic written as-is, this check just ensures that the add-on products just have the some `GooglePurchasingData` type as the base item, which allows it to support both subscriptions and one-time purchases. There's a check above that checks to ensure that the base item is a `GooglePurchasingData.Subscription`, so this effectively checks that all add-on products are `GooglePurchasingData.Subscription`s.

That said, it does feel a bit overkill for the current functionality, so if you think it's best to not over-engineer this check, I'm happy to change it to just check that each add-on is a `GooglePurchasingData.Subscription` :D

purchases/src/main/kotlin/com/revenuecat/purchases/google/ProductValidation.kt
fire-at-will: That is a _great_ question. I'll test this out today and see! If it causes an issue, I'll add a check here for it.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
fire-at-will: I think we can deprecate this existing constructor, but we should probably wait to do so until we remove the `@ExperimentalPreviewRevenueCatPurchasesAPI` annotation from `StoreTransaction.subscriptionOptionIdForProductIDs`. Otherwise, we wouldn't have a non-deprecated constructor that doesn't require you to opt-in to experimental APIs.

For my own curiosity, do you know why we have this constructor public? I usually think of these kinds of objects as something the SDK creates and returns to developers, but not really as something that developers would need to instantiate themselves.

purchases/src/main/kotlin/com/revenuecat/purchases/google/ProductValidation.kt
fire-at-will: Great idea, I've added a validation that throws if one or more of the purchases in a multi-line subscription are for the same subscription product + some unit tests for that here: 38555f86d41a63bf75e3cc4aea593483d5a898be 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -122,6 +128,64 @@ class PurchaseParams(val builder: Builder) {
...
+         * This will result in a multi-line purchase whose base product is the one passed in to the
+         * [PurchaseParams.Builder]'s constructor.
+         *
+         * The following restrictions apply to add-on purchases:
+         * - Add-on purchases are currently only supported for subscriptions on the Play Store.
+         * - The renewal periods of all add-on packages must be the same and match the period of the base product.
+         * - No more than 49 add-ons packages per multi-line purchase are allowed.
+         */
+        @ExperimentalPreviewRevenueCatPurchasesAPI
+        @Throws(PurchasesException::class)
```
tonidero: TBH I’m wondering if having this throw is the best API for this, since Kotlin doesn’t force you to handle it, I can foresee some people would not actually catch this exception…

One alternative I think could be to have a different method in Purchases where it allows you to pass the different packages/products to bundle them and create a new entity for this bundled subscription, this method can return an error. Then, allow to pass this new bundled entity when creating a PurchaseParams...

Wdyt @RevenueCat/coresdk ?

fire-at-will: Yeah, it's definitely possible (and maybe even likely?) for a developer to not try/catch this function, leading to app crashes.

Perhaps we could do something like:

```kotlin
val subscriptionAddOns = Purchases.sharedInstance.createSubscriptionAddOns(
   baseItem: StoreProduct, 
   storeProducts: List<StoreProduct>
) -> Result<SubscriptionAddOns>

// Developers can handle the result however they'd like, we'll use .getOrNull()
// here for simplicity
subscriptionWithAddOns.getOrNull()?.let {
   purchaseParams.addOns(subscriptionAddOns)            
}

// Make purchase
```

I can think of a few potential issues to this approach:
- It slightly complicates the purchase flow for developers, as they need to make an extra function call and handle the case where the add-ons aren't valid (though that's better than crashing)
- It's possible for a developer to pass in one baseItem to `bundleSubscriptionAddOns`, but then pass a different one in to the purchase params. This could lead to a scenario where the add-ons were validated against an add-on item that isn't actually being purchased.
   - In this case, perhaps we could just return an error in the `purchase()` function?

fire-at-will: Despite those drawbacks, it's definitely a safer API than having a function throw that people may not expect to throw!

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
fire-at-will: I did some testing this afternoon and confirmed that offline entitlements mode will need to be updated to support multi-line subscriptions...I'll spend some time tomorrow looking into what all needs to be updated to add support :)

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
fire-at-will: I was thinking about this while doing some yard work this evening and thought of another potential option. 

What if we make `PurchaseParamsBuilder.addOnProducts()` not throw and not perform any validations at all? Then, we perform the validations when the user calls the purchase() function with the purchaseParams and, if validation errors are found, return an error with `PurchasesErrorCode.PurchaseInvalidError` and a description of what is wrong with the purchase params.

This matches what the Billing Client does for multi-line subs (you can pass in any combination of products when launching the billing flow, but the purchase sheet will show an error if there's a compatibility issue between the products). In fact, this could be better than what Google does since developers could theoretically handle this without having Google show a sheet with an error on it.

I think this approach:
- Scales well to other things in the purchase params we may want to validate in the future since we don't need to create new top level functions/objects to perform validation
- Preserves the relative simplicity when creating the purchase params object
- Shouldn't require extra work for developers to handle the error cases since they should already be handling error scenarios from the purchase function.

What do you think @tonidero?

tonidero: I do think that makes a lot of sense and it's probably cleaner! Let's do that!

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
MarkVillacampa: I think this is ok, we probably only care about the base product in that scenario, but good callout

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
MarkVillacampa: yes it's the same purchase token

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
MarkVillacampa: I agree making the validations and throwing on the purchase method makes more sense than the params builder 👍

fire-at-will: The changes for this got bigger than I thought due to moving around the validations & their tests, plus adding a few new tests, so I opened up a PR going into this branch with the changes to make it easier to review: https://github.com/RevenueCat/purchases-android/pull/2706

purchases/src/main/kotlin/com/revenuecat/purchases/models/GoogleSubscriptionOption.kt
fire-at-will: Right! I think I had things twisted around in my head. We should add support for this, so that developers can do something like:

```kotlin
val purchaseParams = PurchaseParams.Builder(activity, subscriptionOption)
      .addOnProducts(...)
      .build()
```

We should also consider having a builder function that is `.addOnSubscriptionOptions()` to go with `.addOnProducts()` and `.addOnPackages()`.

Perhaps this could go in a follow-up PR since this one is already getting a bit large?

tonidero: Yup I think that’s ok!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
fire-at-will: Update: I've merged https://github.com/RevenueCat/purchases-android/pull/2706 into this branch, which moves the validation logic from the Builder to the purchase function :)

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
fire-at-will: I've confirmed that we're not sending up `platform_product_ids` for restores for purchases both with and without multi-line subscriptions, so adding support for this is likely out of the scope for this PR 👍

purchases/src/main/kotlin/com/revenuecat/purchases/models/GooglePurchasingData.kt
```diff
@@ -17,6 +18,10 @@ sealed class GooglePurchasingData : PurchasingData {
         val optionId: String,
         val productDetails: ProductDetails,
         val token: String,
+        @ExperimentalPreviewRevenueCatPurchasesAPI
+        val billingPeriod: Period? = null,
```
fire-at-will, line 29: Set the default values to `null` to not break the old constructor's signature

purchases/src/main/kotlin/com/revenuecat/purchases/models/GoogleSubscriptionOption.kt
```diff
@@ -128,5 +131,25 @@ class GoogleSubscriptionOption @JvmOverloads constructor(
...
+     * or if no infinitely recurring phase is found, then the last one.
+     */
+    private val primaryPricingPhase: PricingPhase?
+        get() {
+            val infiniteRecurringPricingPhase = pricingPhases
+                .firstOrNull { it.recurrenceMode == RecurrenceMode.INFINITE_RECURRING }
+
+            if (infiniteRecurringPricingPhase == null) {
+                log(LogIntent.WARNING) {
+                    OfferingStrings.GOOGLE_PRODUCT_MISSING_INFINITELY_RECURRING_BILLING_PHASE.format(this.productId)
```
tonidero, line 151: Hmm I guess this would always log for non subscriptions right? Maybe we should only log if the type is a subscription?

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreTransaction.kt
tonidero: I think they can be useful for some developers for testing purposes, but yeah, in general, it's not really needed to make them public.

purchases/api-defauts.txt
```diff
@@ -994,11 +996,15 @@ package com.revenuecat.purchases.models {
   }
 
   @dev.drewhamilton.poko.Poko public static final class GooglePurchasingData.Subscription extends com.revenuecat.purchases.models.GooglePurchasingData {
-    ctor public GooglePurchasingData.Subscription(String productId, String optionId, com.android.billingclient.api.ProductDetails productDetails, String token);
+    ctor public GooglePurchasingData.Subscription(String productId, String optionId, com.android.billingclient.api.ProductDetails productDetails, String token, optional @com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI com.revenuecat.purchases.models.Period? billingPeriod, optional @com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI java.util.List<? extends com.revenuecat.purchases.models.GooglePurchasingData>? addOnProducts);
```
tonidero: I think it could be better to mark the whole constructor as experimental, instead of just the parameter for the addons... NABD though.

purchases/api-defauts.txt
```diff
@@ -994,11 +996,15 @@ package com.revenuecat.purchases.models {
...
+    method public java.util.List<com.revenuecat.purchases.models.GooglePurchasingData>? getAddOnProducts();
```
tonidero: Seems we haven't added an experimental annotation for these?

purchases/api-defauts.txt
```diff
@@ -1270,6 +1277,7 @@ package com.revenuecat.purchases.models {
     method @Deprecated public java.util.List<java.lang.String> getSkus();
     method public String? getStoreUserID();
     method public String? getSubscriptionOptionId();
+    method public java.util.Map<java.lang.String,java.lang.String>? getSubscriptionOptionIdForProductIDs();
```
tonidero: Same here about the experimental annotations

tonidero (approved):
I think this is fine to merge as experimental, while we work on those remaining topics 👍

purchases/src/main/kotlin/com/revenuecat/purchases/models/GoogleSubscriptionOption.kt
fire-at-will: Does the SDK ever create a `GoogleSubscriptionOption` for non-subscription IAPs? I'm looking around the SDK for a case of that now

fire-at-will: We might need to create an abstraction like the `GoogleSubscriptionOption ` when we add support for OTP variants, but I think it'd make sense to create a new class named something like `GoogleOneTimePurchaseOption` for that to avoid confusion

tonidero: Ohh you're totally right 🤦 Yeah, this is only for subscriptions, so it's ok to leave as is. Thanks!

purchases/api-defauts.txt
fire-at-will: Yeah, looks like Poko didn't transitively apply the experimental annotation here. I'll see if I can add it manually by explicitly writing the getter functions.

fire-at-will: Okay, so this wasn't a Poko thing. It looks like Kotlin auto-generates these getters for us, and doesn't apply `@ ExperimentalPreviewRevenueCatPurchasesAPI` to them, potentially because requiring opt-in to use functions is a Kotlin feature. 

To get around this, I've applied `@get:JvmSynthetic` to these properties, which prevents these getters from being generated at all. This will prevent these from being accessible from Java, but I think that's okay for now. We can remove the `@get:JvmSynthetic` annotation when we remove the experimental annotation.

purchases/api-defauts.txt
fire-at-will: Makes sense! I've done that in this commit: 65d60fa90ea1a55febb4899dcc12f25073c9d0ed

To get this done, I:
- Removed the experimental annotations from the individual properties
- I still wanted to keep them in the primary constructor, so I made the primary constructor experimental
- Then, to not break the original contract, I re-created the constructor without the new params and didn't mark it as experimental.

Let me know if that looks correct to you!

MarkVillacampa (approved):

PR #2672, opened 2025-09-15
Fix bundled font families when using assets folder

tonidero:
### Description
As a follow-up from https://github.com/RevenueCat/purchases-android/pull/2624.

That PR fixed embedded font families when they were in the resources folder, as it's common in android. However, that didn't fix the issue when fonts are in the assets folder. For example, when being added in React native when using `npx react-native-asset`.

This will fix that so font families load correctly when using the assets folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -278,6 +278,35 @@ Using the SDK with your own IAP code is still supported in v9. Other than updati
 * Add V9 migration guide (#2508) via Toni Rico (@tonidero)
 * [AUTOMATIC][Paywalls V2] Updates paywall-preview-resources submodule (#2499) via RevenueCat Git Bot (@RCGitBot)
 
+## 8.24.0
+## RevenueCat SDK
+> [!NOTE]  
+> This release brings all non-breaking changes between SDK versions 9.4.0 and 9.6.0 to major 8, so developers that don't/can't use major 9, can still get the latest updates.
+
+### ✨ New Features
+* Add `RevenueCatBackupAgent` (#2625) via Toni Rico (@tonidero)
+* Add preferred UI locale override for RevenueCat UI components (#2620) via Josh Holtz (@joshdholtz)
+* Add option to disable automatic ID collection when setting attribution network IDs at configuration time (#2643) via Toni Rico (@tonidero)
+### 🐞 Bugfixes
+* Handle payment pending errors when restoring properly (#2635) via Toni Rico (@tonidero)
+
+## RevenueCatUI SDK
+### Paywallv2
+#### ✨ New Features
+* MON-1193 Support delayed close button (Component Transitions) (#2623) via Jacob Rakidzich (@JZDesign)
+#### 🐞 Bugfixes
+* Fix PaywallDialog going over screen size on Android 35+ (#2642) via Toni Rico (@tonidero)
+### Customer Center
+#### ✨ New Features
+* Add button_text to ScreenOffering (#2638) via Facundo Menzella (@facumenzella)
+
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
```diff
@@ -54,22 +54,29 @@ internal sealed interface FontSpec {
...
             fontInfo?.family?.let { family -> family to fontInfo }
         }
         .groupBy({ (family, _) -> family }, { (_, fontInfo) -> fontInfo })
+    val resourceFontFamilies = fontInfosByFamily
         .mapNotNull { (family, fontInfos) ->
             fontInfos.toFontSpecResource(resourceProvider)?.let {
                 family to it
             }
         }.associateBy({ it.first }, { it.second })
+    val assetFontFamilies = fontInfosByFamily.filterNot { it.key in resourceFontFamilies }
```
ajpallares, line 70: Just to make sure: we want to prioritize resources over assets right? In the sense that if a font is present as both an asset and a resource, then the resource one will prevail.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
```diff
@@ -79,12 +81,22 @@ internal class PaywallResourceProvider(
...
+                val nameWithExtension = if (name.endsWith(".ttf")) name else "$name.ttf"
+                val path = assetsList?.find { it == nameWithExtension }
+                    ?.let { "${ResourceProvider.ASSETS_FONTS_DIR}/$it" }
+                if (path != null) {
+                    name to path
+                } else {
+                    null
+                }
+            }
+            .associateBy({ it.first }, { it.second })
```
vegaro: this can be replace with `.toMap()`, since you already have a `List<Pair<String, String>>`

ajpallares (approved):
A bit out of my expertise but I think it all makes sense. Just a small question, nothing important really.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
tonidero: That's correct! Resources is how it usually works on Android... but RN does it differently. Still, we should default to the "normal" way of reading fonts. It should normally only be in one place though, otherwise it would just be increasing the bundle size

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallResourceProvider.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/2672/commits/2eff24b556f337ab3bb0f7951069737a96addcd3

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FontSpec.kt
ajpallares: Makes total sense. Thank you for the explanation!

PR #2680, opened 2025-09-19
Paywalls -> Video Component

JZDesign:
> [!IMPORTANT]
> This is behind a compiler flag. When it merges it will be dark code. There is a follow up PR planned to implement a 
> Checksum and more robust file storage to better handle the load on device memory

https://github.com/user-attachments/assets/6a6fd861-e243-4216-a8e2-272e6c7d3e28

> [!TIP]
> To test this, run PR 6173 in the revenue cat app

### Checklist

### Motivation

We have requests for video support.

### Description

Creates a video component that uses the native MediaPlayer under the hood.

It supports autoplay, looping, muting audio, showing playback controls, and content scale overrides like fill while retaining the aspect ratio _(which was surprisingly the hardest part)_


purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -160,6 +183,7 @@ internal class DefaultFileRepository(
     }
 }
 
+@InternalRevenueCatAPI
 private class DefaultFileCache(
```
vegaro, line 180: This one is private, why does it need the `@InternalRevenueCatAPI` annotation?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -25,7 +25,8 @@ import java.security.MessageDigest
 /**
  * A file repository that handles downloading and caching files from remote URLs.
  */
-internal interface FileRepository {
+@InternalRevenueCatAPI
+interface FileRepository {
```
vegaro, line 30: I believe the reason to make this public is to pass `repository` to `VideoView`. And the only usage of `repository` in `VideoView` is to get the URI. Maybe we can replace `repository` with the `URI` itself and we avoid exposing this. That way the API is closer to `RemoteImage` or `TextureVideoView`, where the URI is one of the params.

JZDesign: The main reason is that the URI is different on disk than from the remote source and if we do not yet have the video saved to disk when the paywall is being rendered, the file repository will resume the download that was started during cache warming and reduce the time the user will need to wait by not starting a new request.  So there are 2 methods I want shared with the UI layer: `getFile` which is the synchronous check and `generateOrGetCachedFileURL` which performs the download and saves to disk before returning the new file location. This second one is a suspend function and it shares the deferred task so that many callers can all await the same result without hitting the network again _(this is keyed by the URL of the request)_

RemoteImage uses Coil under the hood, so it's doing some of what I'm doing here, but it's behind a UI dependency. I couldn't think of a good clean way to abstract it and invert the dependency. I'm open to it though if you have a pattern in mind

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: Actually yes it is needed because it conforms to a `InternalRevenueCatAPI` interface

JZDesign: Though I can mark it : `@OptIn(InternalRevenueCatAPI::class)` and not `@ InternalRevenueCatAPI`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
```diff
@@ -0,0 +1,344 @@
...
+            selectedTabIndexProvider = selectedTabIndexProvider,
+        )
+    }.apply {
+        update(
+            windowSize = windowSize,
+            density = density,
+            darkMode = darkMode,
+            layoutDirection = layoutDirection,
+        )
+    }
```
skydoves: I’m curious why we’re not passing all these configurations into remember as parameters so the state can update dynamically.

```kotlin
remember(keys = style, windoeSize, densitry, darkMode, layoutDirection, etc)
```

Then I believe we can just remove the `update` function and the overused `derivedStateOf`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
```diff
@@ -0,0 +1,344 @@
...
+    @get:JvmSynthetic
+    val shadow by derivedStateOf { presentedPartial?.shadow ?: style.shadow }
+
+    @get:JvmSynthetic
+    val overlay: ColorStyles? by derivedStateOf { presentedPartial?.overlay ?: style.overlay }
+
+    @get:JvmSynthetic
+    val contentScale: ContentScale by derivedStateOf {
+        presentedPartial?.partial?.fitMode?.toContentScale() ?: style.contentScale
+    }
```
skydoves, line 234: I feel like we may be overusing `derivedStateOf` whenever lambdas are involved. Since `derivedStateOf` is actually quite expensive under the hood, we should consider it for cases where it truly helps, specifically to prevent unnecessary recompositions when the result hasn’t changed.

For example, the [official documentation](https://developer.android.com/develop/ui/compose/side-effects#derivedstateof) says:

```kotlin
// DO NOT USE. Incorrect usage of derivedStateOf.
var firstName by remember { mutableStateOf("") }
var lastName by remember { mutableStateOf("") }

val fullNameBad by remember { derivedStateOf { "$firstName $lastName" } } // This is bad!!!
val fullNameCorrect = "$firstName $lastName" // This is correct
```

Most of the current values from `derivedStateOf` depend on `presentedPartial`. If we include `windowSize`. the key factor in recalculating `presentedPartial`, as a parameter when creating `rememberUpdatedVideoComponentState`, then both `presentedPartial` and all other properties that rely on it will also be updated accordingly without using `derivedStateOf`.

skydoves (commented):
Amazing first feature implementation! I just left a few minor comments related to Compose state :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
JZDesign: Mostly because I didn't know I could. Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
JZDesign: Okay, so I've been doing a bit of reading on this, and I then had Gemini review it. Gemini came back and said essentially:

> Yes, chaining can be bad. This actually looks fine though

And some detail

> **Why Chaining Can Be Good**
>
>    **•Modularity and Readability:** It allows you to break down complex state derivations into smaller, more manageable, and understandable pieces.
>
>    **•Performance Optimization:** Each link in the chain only re-evaluates if its direct dependencies actually result in a new value. This can prevent a cascade of unnecessary recalculations down the chain. If an early derivedStateOf in the chain filters out "noise" (input changes that don't affect its output), subsequent derivedStateOf instances won't even be triggered for re-evaluation.

JZDesign: Now, I'm an iOS guy typically, so I'm a bit out of my element here. I will take your advice on the next steps. If you think I should do something different, point me in the right direction.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,508 @@
...
+    private var looping = false
+    private var autoStart = true
+    private var scaleType = ScaleType.FIT
+
+    // resume state across surface re-create (e.g., rotation)
+    private var resumePosMs = 0
+    private var resumePlayWhenReady = false
+
+    private val layoutListener = ViewTreeObserver.OnGlobalLayoutListener { applySizing() }
+    private var vtoListening = false
```
vegaro: nitpick, but it took me a bit to get that `vto` was `viewTreeObserver`, maybe it would be a good idea to rename this to `viewTreeObserverListening`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+                            padding = style.padding,
+                            margin = style.margin,
+                            shape = style.shape,
+                            border = style.border,
+                            shadow = style.shadow,
+                            overlay = style.overlay,
+                            contentScale = style.contentScale,
+                            rcPackage = style.rcPackage,
+                            tabIndex = style.tabIndex,
+                            overrides = emptyList(), // TODO
```
JZDesign: I'm not sure if this is actually supported or not yet, and if it is, if the localization key is the same as the one on the video component or if it's nested in the image data.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,508 @@
...
+            }
+        }
+    }
+
+    // Inside TextureVideoView.kt
+    class PlaybackState(val positionMs: Int, val playWhenReady: Boolean)
+
+    fun getPlaybackState(): PlaybackState {
+        // If not prepared yet, return last requested state (resumePosMs / resumePlayWhenReady)
+        val pos = if (prepared) {
```
vegaro: there's a mix of `pos` and `position` in this file. I think using `position` consistently would be better

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,508 @@
...
+        released = true
+        if (vtoListening) {
+            viewTreeObserver.removeOnGlobalLayoutListener(layoutListener)
+            vtoListening = false
+        }
+    }
+
+    private fun prepareIfNeeded() {
+        if (released) return
+        val u = uri ?: return
```
vegaro: I think you can just check if uri is null here instead of creating a variable, kotlin will know it's not nullable after this line

```
if (uri == null) return
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+            suspend fun fetchVideoUrl(withUrgency: Boolean) {
+                try {
+                    if (withUrgency) {
+                        videoUrl = videoState.videoUrls.urlLowRes?.toString()?.let(::URI)
+                    }
+
+                    val url = repository.generateOrGetCachedFileURL(videoState.videoUrls.url)
+                    videoUrl = url
+                    imageViewStyle = null
+                } catch (_: Exception) {
```
vegaro: instead of setting `imageViewStyle` here to null, could you just check if `videoUrl` is null when rendering the fallback image in line 130?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+            modifier = modifier
+                .size(videoState.sizePlusMargin)
+                .applyIfNotNull(videoState.marginAdjustedAspectRatio) { aspectRatio(it) }
+                .padding(videoState.margin)
+                .applyIfNotNull(shadowStyle) { shadow(it, composeShape) }
+                .clip(composeShape)
+                .applyIfNotNull(borderStyle) { border(it, composeShape).padding(it.width) },
+        ) {
+            var imageViewStyle: ImageComponentStyle? by rememberSaveable {
+                if (style.fallbackSources != null) {
```
vegaro: I would rename this to `fallbackImageViewStyle`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
JZDesign: I could, but part of why I'm doing this is to hide the fallback image view when the video view should be displayed. Technically, they _should_ have the same sizing and take up the same space, but if the aspect ratios are different, you may see the image underneath the video player, and that's not the desired behavior.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
JZDesign: This is why I did that

Screenshot 2025-09-24 at 6 40 11 PM

JZDesign: I was also going to use `uri?.run` or `uri?.let` but it felt wrong

vegaro: I see, `uri?.let` is actually pretty standard in Kotlin in this case of situation:

```
uri?.let {
    prepared = false
    player.reset()
    player.setDataSource(context, it)
} ?: return
```

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.revenuecat.purchases.paywalls.components.TabsComponent
+import com.revenuecat.purchases.paywalls.components.TextComponent
+import com.revenuecat.purchases.paywalls.components.TimelineComponent
+import com.revenuecat.purchases.paywalls.components.VideoComponent
+import com.revenuecat.purchases.storage.DefaultFileRepository
+import com.revenuecat.purchases.storage.FileRepository
+
+class OfferingVideoPredownloader(context: Context) {
+    @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    var fileRepository: FileRepository = DefaultFileRepository(context)
```
JayShortway: Can we avoid `﻿﻿﻿VisibleForTesting`? It's only on the getter anyway. The one creating the `OfferingVideoPredownloader` can keep a reference to the `FileRepository` too if needed.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    /**
+     * Returns all PaywallComponent that satisfy the predicate.
+     *
+     * Implemented as breadth-first search.
+     */
+    @Suppress("CyclomaticComplexMethod")
+    private fun PaywallComponent.filter(predicate: (PaywallComponent) -> Boolean): List<PaywallComponent> {
```
JayShortway: We can make this an internal extension function in a separate file, to avoid duplication in `OfferingImagePreDownloader`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Localization.kt
```diff
@@ -75,6 +77,34 @@ internal fun LocalizationDictionary.image(key: LocalizationKey): Result<ThemeIma
         ?.let { Result.Success(it) }
         ?: Result.Error(MissingImageLocalization(key))
 
+/**
+ * Retrieves an Video for all locales in this map, associated with the provided [key].
```
JayShortway:
```suggestion
 * Retrieves a Video for all locales in this map, associated with the provided [key].
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/VideoComponentStyle.kt
```diff
@@ -0,0 +1,71 @@
...
+     * available.
+     */
+    @get:JvmSynthetic
+    val rcPackage: Package?,
+    /**
+     * If this is non-null and equal to the currently selected tab index, the `selected` [overrides] will be used if
+     * available. This should only be set for images inside tab control elements. Not for all images within a tab.
+     */
+    @get:JvmSynthetic
+    val tabIndex: Int?,
```
JayShortway, line 63: Do we expect people to place videos inside tab controls? I guess it's technically possible, and we shouldn't place arbitrary limits, especially in the SDK. In that case: 

```suggestion
    /**
     * If this is non-null and equal to the currently selected tab index, the `selected` [overrides] will be used if
     * available. This should only be set for videos inside tab control elements. Not for all videos within a tab.
     */
    @get:JvmSynthetic
    val tabIndex: Int?,
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
JayShortway: Oh wow thanks for pointing this out @skydoves! TIL that `derivedStateOf` is expensive. This is on me, as I've been using it all over the place in every `*ComponentState`. 
   
If there's a quick win here, we can take it. Otherwise it might be good to stay consistent for now, and see if we can update all `*ComponentState`s to reduce the use of `derivedStateOf` later?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
```diff
@@ -0,0 +1,336 @@
...
+                thisDimensionPx = videoUrls.height,
+                otherDimensionPx = videoUrls.width,
+                density = density,
+            ),
+        )
+
+    /**
+     * Adjusts this size constraint to take into account the size of the video.
+     */
+    private fun SizeConstraint.adjustDimension(
```
JayShortway, line 271: Maybe this can also be an internal extension function? Not a blocker at all.

ui/revenuecatui/api.txt
```diff
@@ -182,6 +182,50 @@ package com.revenuecat.purchases.ui.revenuecatui.components.image {
...
+  public enum TextureVideoView.ScaleType {
+    enum_constant public static final com.revenuecat.purchases.ui.revenuecatui.components.video.TextureVideoView.ScaleType FILL;
+    enum_constant public static final com.revenuecat.purchases.ui.revenuecatui.components.video.TextureVideoView.ScaleType FIT;
+  }
+
+  public final class VideoViewKt {
+    method @androidx.compose.runtime.Composable public static void VideoView(String videoUri, optional androidx.compose.ui.Modifier modifier, optional boolean showControls, optional boolean autoPlay, optional boolean loop, optional boolean muteAudio, optional androidx.compose.ui.layout.ContentScale contentScale);
+  }
+
+}
```
JayShortway: Do any of these need to be public?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,509 @@
...
+    }
+}
+
+@Suppress("TooGenericExceptionCaught")
+private fun safely(execute: () -> Unit, failureMessage: (Exception) -> String? = { null }) {
+    try {
+        execute()
+    } catch (e: Exception) {
+        failureMessage(e)?.run {
+            Log.e("TextureVideoView", this)
```
JayShortway: We should probably use our own `Logger` for this. (It's very confusing that `revenuecatui` and `purchases` have separate loggers.)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,509 @@
...
+    },
+    restore = {
+        val restored = SavedPlayback(it[0] as Int, it[1] as Boolean)
+        mutableStateOf(restored)
+    },
+)
+
+@Suppress("LongMethod", "LongParameterList")
+@Composable
+private fun VideoCard(
```
JayShortway: Why is this named `*Card`? We could maybe just call this `Video`, like there's an `Image` too?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.components.style.ImageComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.style.VideoComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallState
+import com.revenuecat.purchases.ui.revenuecatui.extensions.applyIfNotNull
+import java.net.URI
+
+@Suppress("CyclomaticComplexMethod", "LongMethod", "ModifierNotUsedAtRoot", "ModifierReused")
+@JvmSynthetic
+@Composable
+internal fun VideoComponentView(
```
JayShortway, line 37: It's going to be really difficult avoiding regressions in all the sizing and scaling options without any previews. Previewing videos doesn't work, as we've found, but maybe we can have this composable detect the preview environment, and display an image instead of a video using the same size and scale options? Maybe that gives us _some_ value, but not sure. In any case I'd do it in a separate PR.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+    style: VideoComponentStyle,
+    state: PaywallState.Loaded.Components,
+    modifier: Modifier = Modifier,
+    repository: FileRepository = DefaultFileRepository(LocalContext.current),
```
JayShortway: Should we remember this?

```suggestion
    repository: FileRepository = remember { DefaultFileRepository(LocalContext.current) },
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
JayShortway: Is this something that Dan or Josh knows?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            // If both of the video files are not found on disk
+            if (videoUrl == null) {
+                LaunchedEffect(Unit) {
+                    fetchVideoUrl(withUrgency = fallbackImageViewStyle == null)
+                }
+            } else {
+                fallbackImageViewStyle = null
+            }
```
JayShortway: Maybe all of this can be extracted into a `@Composable fun rememberVideoUrl()` function, so it's easier to reason about outside of the main `VideoComponentView`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+                    // this may result in some paywalls not looking so good depending on the fallback image they used
+                    // and the styles applied to the video component
+                    videoUrl = videoState.videoUrls.url.toString().let(::URI)
+                }
+            }
+
+            // If the full size video wasn't found on disk try the low res
+            if (videoUrl == null) {
+                if (videoState.videoUrls.urlLowRes != null) {
+                    videoUrl = repository.getFile(videoState.videoUrls.urlLowRes!!)
```
JayShortway: Normally I'm not afraid of double bangs (😅) but in this case `VideoComponentView` shouldn't assume anything about `VideoComponentState` imo. We could do something like this:

```kotlin
videoState.videoUrls.urlLowRes?.also { urlLowRes ->
  videoUrl = repository.getFile(urlLowRes)
  // ...
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -0,0 +1,148 @@
...
+            // footprint of paywalls
+            suspend fun fetchVideoUrl(withUrgency: Boolean) {
+                try {
+                    if (withUrgency) {
+                        videoUrl = videoState.videoUrls.urlLowRes?.toString()?.let(::URI)
+                    }
+
+                    val url = repository.generateOrGetCachedFileURL(videoState.videoUrls.url)
+                    videoUrl = url
+                    fallbackImageViewStyle = null
```
JayShortway: I'm a bit confused about `withUrgency`. Maybe there's a more descriptive name? Is `generateOrGetCachedFileURL` so expensive that we need to set `videoUrl` to the low-res one first? Is `videoUrl` always local? If so maybe we can reflect that in the name?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.revenuecat.purchases.paywalls.components.TabControlComponent
+import com.revenuecat.purchases.paywalls.components.TabControlToggleComponent
+import com.revenuecat.purchases.paywalls.components.TabsComponent
+import com.revenuecat.purchases.paywalls.components.TextComponent
+import com.revenuecat.purchases.paywalls.components.TimelineComponent
+import com.revenuecat.purchases.paywalls.components.VideoComponent
+import com.revenuecat.purchases.storage.DefaultFileRepository
+import com.revenuecat.purchases.storage.FileRepository
+
+class OfferingVideoPredownloader(context: Context) {
```
JayShortway: I think this can be internal, just like `OfferingImagePredownloader`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
JZDesign: Josh is in India, but I meet with Barbara about it tomorrow. I believe it's not yet supported and the video component in the khepri isn't prepared for this yet. But I'm confirming where we are and what needs to be done. It's possible that the override will be a part of the containing video component and not in the nested fallback image property, in which case this would remain empty and I would just remove the comment.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
JZDesign: I'm not sure it's necessary. There is a singleton of the deferred value store being used in the file repository. If the activity is recreated, the instance of that is the same, so it should make no difference.

JayShortway: True, but we'd be reconstructing the `DefaultFileRepository` every time this gets recomposed, which can happen far more often than just Activity recreations. 

This implicitly assumes that `DefaultFileRepository` is cheap to instantiate and stateless. That might be true now, but we don't know if that will remain true in the future. And if it changes, we might forget to update it here. Also, it puts (a little) pressure on the garbage collector, which will have to clean up old instances.

Of course we shouldn't just `remember` everything and optimize prematurely, but maybe it's safer to do it here?

JZDesign: Okay. I can see that. On it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
JZDesign: This is not something we need to worry about after all. The video component will contain its own overrides and that would include the fallback image. So this can remain empty.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -39,25 +41,41 @@ internal interface FileRepository {
...
     private val fileCacheManager: LocalFileCache,
     private val ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO + NonCancellable),
     private val logHandler: LogHandler = currentLogHandler,
     private val urlConnectionFactory: UrlConnectionFactory = DefaultUrlConnectionFactory(),
 ) : FileRepository {
+    companion object {
+        // This allows us to share the asynchronous tasks across callsites
+        // and instances of the file repository, making it possible to pick up
+        // file downloads in the middle of the download.
+        val urlToUriStore = KeyedDeferredValueStore<URL, URI>()
```
tonidero: Hmm I don't like storing data statically like this... I would prefer to make the `FileRespository` a "singletonish" class in the Purchases instance (have a single instance stored as part of the `Purchases` instance), and pass that instance around where needed to keep this data. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -39,25 +41,41 @@ internal interface FileRepository {
...
     fun saveData(data: ByteArray, uri: URI)
 }
 
-internal class DefaultFileRepository(
+@InternalRevenueCatAPI
+class DefaultFileRepository
+internal constructor(
     @get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal val store: KeyedDeferredValueStore<URL, URI> = KeyedDeferredValueStore<URL, URI>(),
+    val store: KeyedDeferredValueStore<URL, URI> = urlToUriStore,
```
tonidero: Hmm any reason we had to make this parameter public instead of keeping it internal?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -39,25 +41,41 @@ internal interface FileRepository {
...
+@InternalRevenueCatAPI
+interface LocalFileCache {
     fun generateLocalFilesystemURI(remoteURL: URL): URI?
     fun cachedContentExists(uri: URI): Boolean
     fun saveData(data: ByteArray, uri: URI)
 }
 
-internal class DefaultFileRepository(
+@InternalRevenueCatAPI
+class DefaultFileRepository
```
tonidero: Hmm do we need to make the implementation class public? Or would it be enough with just making the interface public? I see that we're creating new instances as part of the `VideoComponentView` which is probably the reason, but if we apply my other suggestion of making the FileRespository a "singletonish" class, we can keep this one internal.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -0,0 +1,40 @@
...
+internal class OfferingVideoPredownloader(
+    context: Context,
+    private val fileRepository: FileRepository = DefaultFileRepository(context),
+) {
+    private val shouldPredownload: Boolean = try {
+        Class.forName("com.revenuecat.purchases.ui.revenuecatui.PaywallKt")
+        true
+    } catch (_: ClassNotFoundException) {
+        false
+    }
```
tonidero: Since this logic is duplicated in the OfferingImagePredownloader, it could make sense to move this to a separate place, maybe `AppConfig`, and pass that around?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,528 @@
...
+}
+
+/**
+ * Texture-backed video view with stock MediaController controls.
+ * Implements FIT (letterbox) and FILL (center-crop) by resizing the TextureView,
+ * letting the FrameLayout clip any overflow. No transforms, no buffer-size tricks.
+ */
+@SuppressLint("ViewConstructor")
+@Suppress("TooManyFunctions")
+internal class TextureVideoView @JvmOverloads constructor(
```
tonidero: Hmm can this one be private?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -400,6 +405,7 @@ internal class StyleFactory(
             is TabControlToggleComponent -> createTabControlToggleComponentStyle(component)
             is TabControlComponent -> tabControl.errorIfNull(nonEmptyListOf(PaywallValidationError.TabControlNotInTab))
             is TabsComponent -> createTabsComponentStyle(component)
+            is VideoComponent -> createVideoComponentStyle(component)
```
tonidero, line 427: [Not related to this line!]

We might need to handle "header video" components, same way we do for images: 
https://github.com/RevenueCat/purchases-android/blob/9ef023d18085f3083a584e2627ba670f1d4a0711/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt#L162

This way, it would extend beyond safe areas.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,528 @@
...
+        }
+    }
+
+    private fun prepareIfNeeded() {
+        if (released) return
+
+        uri?.let {
+            prepared = false
+            player.reset()
+            player.setDataSource(context, it)
```
tonidero, line 248: Is there a way to know if there is an issue with the file in the uri? If we can, we should delete the downloaded file in case we detect it's corrupted. If we can't, I think we need to rethink again the issue about the download possibly being corrupted. I really want to avoid the situation of ending up with a corrupted file which would only get fixed with an uninstall/reinstall.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JayShortway: Agreed, provided some pointers [here](https://revenuecat.slack.com/archives/C098ENSLWJK/p1758551928005399?thread_ts=1758117892.563609&cid=C098ENSLWJK).

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: That's what I had in mind originally. But with context being required to instantiate it, and me being newer to this code base, that one was harder for me to figure out how to do well.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: that was a mistake I think.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: Yes, if we can get that done, this can be internal

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JZDesign: Thanks. That got lost in all the things, @JayShortway 😆 I'll do that.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
JZDesign: I agree, we don't yet have a checksum. Open to other ideas though.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JZDesign: I actually had some of it done before I got scrambled

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
JZDesign: Oh, wow, I needed coffee when I read this before apparently. Yea that would work.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
JZDesign: Actually after implementing this, it crashes and the crash log is so long and repetitive it is truncating the beginning of the log in logcat and I'm not sure what's going on.

JZDesign: I'm increasing my logcat buffer and trying to track that down now

JZDesign: Ha, It's a stack overflow

JZDesign: Resolved

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
tonidero: I think the "proper" options that come to mind are:
- Changing how this feature works so we host the files and we can add a checksum or something to verify it
- When the url is added in the frontend, calculate a checksum of the file at that point in time from the frontend/backend (that way we also verify that it's a valid url) and pass that to the clients.

Boths are biggish changes, but I think they are kinda needed or I'm afraid of entering this wrong state permanently. I would consider this a blocker issue.

From a quick search I didn't find a way to detect whether the downloaded file is a valid video (any ideas @JayShortway @vegaro @skydoves?)... which could allow us to potentially clear the files if we detect it's a corrupted video. cc @joshdholtz

JZDesign: For now this will be behind a feature flag and we can add it before the official rollout. If you're comfortable with that.

tonidero: Hmm the problem I see of having it behind a feature flag only on the frontend/backend is that, once we enable the feature flag, a developer might start using it with an older version of the SDK that only has these changes, without any verification, which could potentially cause issues for these developers 😞.

tonidero: If we disable the feature entirely in this branch with a compile time flag that avoids the published version of the SDK to download the files in the first place, I'd be ok merging this for now.

JayShortway: I don't think it's possible to be certain the file is not corrupt without a checksum. We can try to read the file header to see if it's an MP4 file, but that's about it (practically speaking). It can still be corrupt past the header.

Hosting the files ourselves is probably too big of a change (but the ideal solution). I do like the idea of calculating a checksum when the URL is added to the dashboard. We could still calculate it on the backend, and delete the file once we're done. Maybe that's somewhat pragmatic?

Probably safest to keep this behind a build-time flag for now indeed. For inspiration, check out #2685 where we removed the `ENABLE_SIMULATED_STORE` flag.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -144,6 +145,13 @@ class Purchases internal constructor(
     val store: Store
         get() = purchasesOrchestrator.store
 
+    /**
+     * The currently configured FileRepository
+     */
+    @OptIn(InternalRevenueCatAPI::class)
```
tonidero, line 151: This should propagate the experimental label to consumers like this:
```suggestion
    @InternalRevenueCatAPI
```

This way, consumers need to opt in to access this

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.revenuecat.purchases.common.canUsePaywallUI
+import com.revenuecat.purchases.common.checkIfVideoComponentIsEnabled
+import com.revenuecat.purchases.paywalls.components.VideoComponent
+import com.revenuecat.purchases.storage.DefaultFileRepository
+import com.revenuecat.purchases.storage.FileRepository
+
+internal class OfferingVideoPredownloader(
+    context: Context,
+    canShowPaywalls: Boolean = canUsePaywallUI,
+    videoComponentIsEnabled: () -> Boolean = ::checkIfVideoComponentIsEnabled,
```
tonidero, line 17: Hmm I wonder if we should just access `BuildConfig.ENABLE_VIDEO_COMPONENT` directly... I guess the reason for doing it like this is to test it? We can mock static constants using `mockk` like `mockkStatic(BuildConfig::class)`, then `every { BuildConfig.ENABLE_VIDEO_COMPONENT } returns true`. Having said that, I don't think it's a blocker, since this would probably go away soon in any case.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
tonidero: Agreed that we should try to do something like this before shipping it to avoid regressions. These previews have proven very helpful to avoid breaking things :).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,528 @@
...
+            try {
+                player.currentPosition
+            } catch (_: Throwable) {
+                resumePosMs
+            }
+        } else {
+            resumePosMs
+        }
+        val play = if (prepared) {
+            (
```
tonidero, line 178: Hmm is this parentheses needed?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,528 @@
...
+        loop = loop,
+        muteAudio = muteAudio,
+        modifier = modifier,
+    )
+}
+
+/**
+ * Texture-backed video view with stock MediaController controls.
+ * Implements FIT (letterbox) and FILL (center-crop) by resizing the TextureView,
+ * letting the FrameLayout clip any overflow. No transforms, no buffer-size tricks.
```
tonidero, line 66: I feel the pain this must have been 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -0,0 +1,528 @@
...
+        }
+    }
+}
+
+@Stable
+private data class SavedPlayback(val position: Int, val play: Boolean)
+
+// Simple static state holder as fallback when rememberSaveable fails
+private object VideoStateHolder {
+    private val states = mutableMapOf<String, SavedPlayback>()
```
tonidero, line 385: Do we need to clear the saved state here? I think right now if leaving the paywall and coming back, it would remain on the previous state right? In that scenario I think it should start from the beginning... 

I'm also wondering if we could save this state in a ViewModel instead of static like here (view models survive reconfiguring the activity)... But I think we can try to improve this in follow-up PRs.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
JZDesign: Thanks. I'll make this change in the next PR. Missed this comment somehow

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: @JZDesign we should do this before making a release, so we don't create a new stable API and we're forced to make a breaking change later :)

JZDesign: https://github.com/RevenueCat/purchases-android/pull/2709

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
JZDesign: Going into #2717

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
JZDesign: Going into #2717

PR #2685, opened 2025-09-25
Enable Test Store

ajpallares:
This PR removes the ENABLE_SIMULATED_STORE build config to enable the Test Store functionality in the SDK

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -60,7 +60,6 @@ import java.util.concurrent.ThreadFactory
 internal class PurchasesFactory(
     private val isDebugBuild: IsDebugBuildProvider,
     private val apiKeyValidator: APIKeyValidator = APIKeyValidator(),
-    private val isSimulatedStoreEnabled: () -> Boolean = { BuildConfig.ENABLE_SIMULATED_STORE },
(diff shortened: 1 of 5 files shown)


JayShortway (approved):
Makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -77,7 +75,7 @@ internal class PurchasesFactory(
 
         with(configuration) {
             val finalStore = if (
-                apiKeyValidationResult == APIKeyValidator.ValidationResult.SIMULATED_STORE && isSimulatedStoreEnabled()
+                apiKeyValidationResult == APIKeyValidator.ValidationResult.SIMULATED_STORE
             ) {
                 Store.UNKNOWN_STORE // We should add a new store when we fully support the simulated store.
```
tonidero, line 80: I wonder, should we add a new store now for these purchases? I think that was added in iOS already, but we didn't do it in android to avoid the new APIs until we were ready.

ajpallares: We discussed it internally and we're going to keep using the unknown store for the Test Store purchases until we fully release the Test Store. See https://github.com/RevenueCat/purchases-ios/pull/5596#discussion_r2379436251

Once Test Store is released, we'll add the new Store case

PR #2690, opened 2025-09-29
Migrate deprecated RenderScript to RenderScriptToolKit

skydoves:
### Motivation
RenderScript APIs are deprecated starting in Android 12 (API level 31, but our target SDK has already been increased to 33). Device and component manufacturers have already stopped providing hardware acceleration support, and RenderScript support is expected to be removed entirely in a future release.

### Description
Following the [RenderScript Intrinsics Replacement Toolkit](https://github.com/android/renderscript-intrinsics-replacement-toolkit) now we use `RenderScriptToolKit` over deprecated `RenderScript`,

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -2,6 +2,41 @@
   <code_scheme name="Project" version="173">
     <option name="RIGHT_MARGIN" value="120" />
     <option name="SOFT_MARGINS" value="120" />
+    <JavaCodeStyleSettings>
+      <option name="IMPORT_LAYOUT_TABLE">
+        <value>
+          <package name="" withSubpackages="true" static="false" module="true" />
+          <package name="android" withSubpackages="true" static="true" />
+          <package name="androidx" withSubpackages="true" static="true" />
+          <package name="com" withSubpackages="true" static="true" />
+          <package name="junit" withSubpackages="true" static="true" />
+          <package name="net" withSubpackages="true" static="true" />
+          <package name="org" withSubpackages="true" static="true" />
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/PaywallBackground.kt
```diff
@@ -24,13 +23,14 @@ import com.revenuecat.purchases.ui.revenuecatui.isFullScreen
 internal fun BoxScope.PaywallBackground(templateConfiguration: TemplateConfiguration) {
     val supportsNativeBlurring = Build.VERSION.SDK_INT >= BackgroundUIConstants.minSDKVersionSupportingBlur
```
JayShortway, line 24: Just curious whether this PR is necessary, as we are already avoiding the use of RenderScript on API 31 and above, by checking `supportsNativeBlurring`.

skydoves: Oops, I just noticed we’re already using `blur` for API 31 and above, and falling back to `RenderScript` for lower versions 😄 I think this PR can be closed, though it might still be useful to keep as a reference for the future.

JayShortway: Ahh yes, indeed it can be a good reference! Thanks for checking 😄

Issue #2681, opened 2025-09-23
[Customer Center] Expired Status Badge hard to read in dark mode.


dalenjohnson:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
The subscription status badge in the customer center is hard to read in dark mode with a light text color. 

Internally there is a `CustomerCenterConstants.COLOR_BADGE_EXPIRED` that is assigned to `badgeInfo.color` for the badge background color and this is an off-white color. The text color is applied with: `badgeInfo.textColor ?: MaterialTheme.colorScheme.onSurface` but `badgeInfo.textColor` is never set and always null so it uses the `onSurface` color, which is typically a light color in dark mode making this hard to read. 

The quick and dirty solution is probably to add some opacity to `COLOR_BADGE_EXPIRED` like it is for the canceled and active badge color constants. Or change it to a border like in lifetime purchases.
<img width="196" height="131" alt="Image" src="https://github.com/user-attachments/assets/2b82377f-83f9-4148-8e85-da23d7111f36" /><img width="196" height="131" alt="Image" src="https://github.com/user-attachments/assets/0c983980-19c5-41b7-9fe6-0bcccd535694" />
<img width="196" height="131" alt="Image" src="https://github.com/user-attachments/assets/365e2daf-f168-4abd-92b6-557bbeb99472" /><img width="196" height="131" alt="Image" src="https://github.com/user-attachments/assets/86721e73-2fac-4187-914d-ddaa6ee3c883" />

The long term solution is probably to have better theming option for the customer center and probably allow us to assign these badge colors ourselves. Currently, I think that theming the customer center is kind of a mess with being able to define colors on the webUI, providing theming in the app, these internal constants, and all of these options being intermixed in the customer center ui. I know this is made worse by the fact that I am building my app using Compose Multiplatform so I have to deal with both iOS and Android theming. 

1. Environment
   1. Platform: Android
   2. SDK version: KMP 2.1.4+17.7.0
   3. OS version: Android 15/16
   4. Android Studio version:
[...]

PR #2693, opened 2025-09-29
Fix Expired badge in Customer Center dark mode

vegaro:
Reported in https://github.com/RevenueCat/purchases-android/issues/2681

- I added some opacity to the Expired badge, so it shows some contrast against the background, in case it's the same color as `onSurface`. This is the same for Cancelled and Active.
- Added previews for dark mode.
- Improved the border of lifetime, so it's based off the `onSurface` color with an opacity and not a hardcoded color


facumenzella (approved):
snaps look good 🚀

PR #2695, opened 2025-09-30
Expose `launch` and `launchIfNeeded` to Java

vegaro:
Required for https://github.com/RevenueCat/purchases-unity/pull/649

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
@@ -122,9 +122,9 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
     }
 
     @InternalRevenueCatAPI
-    @JvmSynthetic
+    @JvmOverloads
     fun launch(
-        offeringIdentifier: String,
+        offeringIdentifier: String?,
         presentedOfferingContext: PresentedOfferingContext,
         fontProvider: ParcelizableFontProvider? = null,
         shouldDisplayDismissButton: Boolean = DEFAULT_DISPLAY_DISMISS_BUTTON,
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -122,9 +122,9 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
     }
 
     @InternalRevenueCatAPI
-    @JvmSynthetic
+    @JvmOverloads
     fun launch(
-        offeringIdentifier: String,
+        offeringIdentifier: String?,
```
vegaro: this is required to be able to call `launch` without an offering identifier. I can't use the launch that accepts an `Offering` because I need the `presentedOfferingContext` as parameter

tonidero: Hmm I think `presentedOfferingContext` should always be null if the offering identifier/offering is null right? If so, I think we should be able to use that one?

vegaro: oh, ok, so if `offering` and `presentedOfferingContext` are `null`, call the one with `Offering`. Otherwise this one. Good to know!

vegaro: I need to pick your brain:
- If I remove `?` and keep the `@JvmOverloads`, there are two conflicting APIs in Java since `launcher.launch(null, null)` could be `launch(Offering,ParcelizableFontProvider)` and `launch(String,PresentedOfferingContext)`
- If I don't use `@JvmOverloads` I can't use the defaults in the optional parameters in `launch` and `launchIfNeeded`

Any ideas besides creating new calls? Or migrating this to Kotlin lol

vegaro: oh I know what to do! `@JvmName("launchWithOfferingId")`

PR #2698, opened 2025-09-30
mark subscriptionOptionIdForProductIDs as experimental

fire-at-will:
### Checklist


fire-at-will:
Merging this in without review since it's going into a feature branch that will be reviewed before going into main

PR #2699, opened 2025-10-01
Add flag to show / hide user section in CustomerCenter

facumenzella:
### Motivation
This PR follows https://github.com/RevenueCat/purchases-ios/pull/5609 and adds `displayUserDetailsSection` to `Support`. 
Android does not show that section yet, but this is adding support for it

### Description
- Parse flag to show / hide user details section
- Add missing UI

Screenshot_20251001_135810

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
@@ -43,6 +43,7 @@ import com.revenuecat.purchases.models.StoreProduct
 import com.revenuecat.purchases.models.SubscriptionOption
 import com.revenuecat.purchases.models.Transaction
 import com.revenuecat.purchases.models.googleProduct
+import com.revenuecat.purchases.ui.revenuecatui.BuildConfig
 import com.revenuecat.purchases.ui.revenuecatui.OfferingSelection
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivity
 import com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivityArgs
@@ -186,7 +187,6 @@ internal class CustomerCenterViewModelImpl(
     override val actionError: State<PurchasesError?>
         get() = _actionError
     private val _actionError: MutableState<PurchasesError?> = mutableStateOf(null)
-    private var latestCustomerInfo: CustomerInfo? = null
 
     override fun pathButtonPressed(
         context: Context,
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -186,6 +186,7 @@ internal class CustomerCenterViewModelImpl(
     override val actionError: State<PurchasesError?>
         get() = _actionError
     private val _actionError: MutableState<PurchasesError?> = mutableStateOf(null)
+    private var latestCustomerInfo: CustomerInfo? = null
```
tonidero: Hmm I'm not sure if we should cache the CustomerInfo here... Maybe we could return the CustomerInfo as part of the `loadPurchases` method (So return a Pair including the `CustomerInfo` and `PurchaseInformation`) so we can use that information... Otherwise, we could include the required information inside `PurchaseInformation`? Feels like that should be ok.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -830,6 +832,17 @@ internal class CustomerCenterViewModelImpl(
...
+            val originalAppUserId = latestCustomerInfo?.originalAppUserId ?: ""
+            val originalPurchaseDate = latestCustomerInfo?.originalPurchaseDate?.let {
+                dateFormatter.format(
+                    it,
+                    locale,
+                )
+            }
+            val shouldShowUserDetailsSection =
+                customerCenterConfigData.support.displayUserDetailsSection &&
+                    (originalAppUserId.isNotBlank() || originalPurchaseDate != null)
```
tonidero: One thing... Would we still want to show the account details section even if `originalPurchaseDate` is null? Accessing the userId could still be useful I think even if the user never made any purchases, so I'm not sure if we should tie it to that.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/AccountDetailsSection.kt
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        IconButton(
+            onClick = {
+                clipboard.setText(AnnotatedString(appUserId))
+            },
+        ) {
+            Icon(
+                imageVector = Icons.Outlined.ContentCopy,
+                contentDescription = "Copy",
```
tonidero: Hmm I think `contentDescription` here should be either `null` or be localized... I think until now, for icons like these, we've mostly left it as null but I guess we don't have any accompanying string here for that, so I think the proper thing to do here would be to localize it... Maybe we can do it in a separate PR though.

tonidero (commented):
Just some questions, specially about caching the CustomerInfo in the ViewModel... I think it would be better to avoid it if we can. Other than that, looks good. Will also leave to others with more context in the new section to approve.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
facumenzella: I am matching iOS implementation. We've been going back and forth with this, so the dashboard flag just hides it entirely. 😓 

1. The purchase date is always present
2. User id depends on DEBUG. 

This PR just adds a dashboard controlled flag to hide the section entirely.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -550,6 +560,9 @@ private val previewConfigData = CustomerCenterConfigData(
         localizedStrings = mapOf(
             "cancel" to "Cancel",
             "subscription" to "Subscription",
+            "account_details_title" to "Account details",
+            "account_details_user_id" to "App User ID",
+            "account_details_purchase_date" to "Original purchase date",
```
vegaro, line 565: Are these used? I don't think they even match the correct keys

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -859,6 +875,20 @@ internal class CustomerCenterViewModelImpl(
...
+        customerCenterConfigData: CustomerCenterConfigData,
+    ): Triple<String, String?, Boolean> {
+        val originalAppUserId = customerInfo.originalAppUserId ?: ""
+        val originalPurchaseDate = customerInfo.originalPurchaseDate?.let {
+            dateFormatter.format(it, locale)
+        }
+        val shouldShowUserDetailsSection =
+            customerCenterConfigData.support.displayUserDetailsSection &&
+                (BuildConfig.DEBUG || originalPurchaseDate != null)
+        return Triple(originalAppUserId, originalPurchaseDate, shouldShowUserDetailsSection)
```
vegaro, line 889: Instead of returning a Triple, I would create a small data class, specially since there are two Strings

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -822,14 +828,21 @@ internal class CustomerCenterViewModelImpl(
...
+            )
             val virtualCurrencies = if (customerCenterConfigData.support.displayVirtualCurrencies == true) {
                 purchases.invalidateVirtualCurrenciesCache()
                 purchases.awaitGetVirtualCurrencies()
             } else {
                 null
             }
 
+            val (originalAppUserId, originalPurchaseDate, shouldShowUserDetailsSection) =
+                computeUserDetailsInfo(customerInfo, customerCenterConfigData)
```
vegaro, line 844: Maybe `computeUserDetailsInfo` could be an extension on `CustomerInfo`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/AccountDetailsSection.kt
```diff
@@ -0,0 +1,297 @@
...
+
+@Composable
+internal fun AccountDetailsSection(
+    appUserId: String,
+    appearance: CustomerCenterConfigData.Appearance,
+    localization: CustomerCenterConfigData.Localization,
+    modifier: Modifier = Modifier,
+    originalPurchaseDate: String? = null,
+) {
+    if (BuildConfig.DEBUG) {
```
vegaro, line 41: I am not 100% sure this will work. This `BuildConfig` is not the app's, it's our own, and I am not sure if it's generated when our SDK is installed. You can double check in an Android project that uses our Maven dependency and see if it's available or not when building on debug vs release.

facumenzella: I misunderstood how this worked 😢

PR #2701, opened 2025-10-01
Revert kotlinCompilerExtensionVersion for MagicWeatherCompose and CustomEntitlementComputationSample

skydoves:
Revert kotlinCompilerExtensionVersion for MagicWeatherCompose and CustomEntitlementComputationSample since they use Kotlin 1.9.x. Thanks for finding this out, @JayShortway 👍 #2691


JayShortway (approved):
@skydoves Thanks so much for taking care of this! 🙏

PR #2706, opened 2025-10-02
Validate Add-On Products in Purchase Function

fire-at-will:
In [this PR comment thread](https://github.com/RevenueCat/purchases-android/pull/2668#discussion_r2396481257), we decided to move add-on product validation for multi-line subscriptions to the purchase() function to make the APIs safer for developers.

This PR:
- Moves the add-on product validation for multi-line subscriptions to the purchase() function
- Updates the product validation unit tests
- Unrelated, but added a few unit tests for testing products changes with multi-line subscriptions that I noticed were missing


purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -29,13 +28,21 @@ class PurchaseParams(val builder: Builder) {
     @get:JvmSynthetic
     internal var presentedOfferingContext: PresentedOfferingContext?
 
+    @get:JvmSynthetic
+    internal val baseItemProduct: StoreProduct?
```
fire-at-will: When we're in the purchase() function, we can only access the built purchaseParams, which only provides access the `GooglePurchasingData`s of the products. Here, we add the `StoreProduct`s for the base item & add-on products so that we can perform validations on their billing periods, since we don't have access to billing periods in `GooglePurchasingData`s.

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
```diff
@@ -1835,14 +1963,136 @@ internal class PurchasesTest : BasePurchasesTest() {
...
+
+            assertThat(capturedUserCancelled).withFailMessage(
+                "Expected userCancelled to be false for store: $store"
+            ).isNotNull().isFalse()
+        }
+    }
+
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    @Test
+    fun `product change with add-ons throws when Store is not PLAY_STORE`() {
```
fire-at-will, line 2087: The diff here looks a little funny. What's happening here is that I added a copy of the `initial purchase with add-ons throws when Store is not PLAY_STORE` test, but for product changes :)

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PurchaseParamsValidator.kt
```diff
@@ -0,0 +1,113 @@
...
+                return addOnProductsValidationError
+            }
+        }
+
+        return Result.Success(Unit)
+    }
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    @Suppress("ReturnCount", "CyclomaticComplexMethod", "LongMethod")
+    internal fun validateAddOnProducts(
```
tonidero: This could be private?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PurchaseParamsValidator.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.revenuecat.purchases.models.Period
+import com.revenuecat.purchases.strings.PurchaseStrings
+import kotlin.jvm.Throws
+
+/**
+ * Maximum number of add-on products allowed in a multi-line purchase.
+ */
+internal const val MAX_NUMBER_OF_ADD_ON_PRODUCTS = 49
+
+internal class PurchaseParamsValidator {
```
tonidero, line 15: NABD but we could make this an `object` and access these functions statically?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
tonidero: Hmm note that the minimal unit that can be purchased is a `SubscriptionOption`. Not sure if this would work with a `SubscriptionOption` instead? We can pickup the "default" SubscriptionOption and store that instead I guess... Though I guess that might not work in the future for non-subscriptions... Might be good adding the period information to PurchasingData instead if that's what we need?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1367,6 +1377,18 @@ internal class PurchasesOrchestrator(
             return
         }
 
+        if (
```
tonidero, line 1380: [Unrelated to this line]

Just curious, how does Google respond if trying to upgrade/downgrade from a previous subscription and when upgrading/downgrading to a different subscription that might not have all of the bundled subscriptions? Seems there are a lot of cases here that might be difficult to handle, so hopefully Google handles it on their own.

tonidero (commented):
In general I think this looks good. I'm just not sure about adding the new data to PurchaseParams and whether we can add it to PurchasingData (or GooglePurchasingData) instead... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PurchaseParamsValidator.kt
fire-at-will: Definitely - updated in fa142b8fbd59b55cfda358b532a5d07a267d4bc9

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PurchaseParamsValidator.kt
fire-at-will: I'm hesitant to make this a singleton via `object` because that would make it more difficult to inject the validator in to other classes and make it more difficult to test. For example, `PurchasesTest.'initial purchase with add-ons that fails purchaseParams validation throws an error'` relies on inserting a mocked validator into the `PurchasesOrchestrator` and forcing it to fail validations by modifying the mock's return values.

That said, `MAX_NUMBER_OF_ADD_ON_PRODUCTS` could be moved into a companion object so it's only visible here! I've made that change in ff9470a02592c0ac2f3ce9f85f2b1a5cda9b1913

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
fire-at-will: Luckily Google seems to handle all of the product change logic for us! For example, you can do a product change from product `A(base item), B` -> `C(base item), D` without any issues. The only complication is that depending on the type of product change you're doing, Google requires that you provide a replacement mode. We might be able to perform some validation on the purchase params to ensure that the replacement mode is set if needed, but it could introduce complexity a bit since we'd need to make a `queryAllPurchases()` call during the validation if `oldProductId` is set.

There's a super nice table with more info on the different scenarios [here](https://developer.android.com/google/play/billing/subscription-with-addons#subscription-modifications)!

purchases/src/main/kotlin/com/revenuecat/purchases/models/GooglePurchasingData.kt
```diff
@@ -17,6 +17,7 @@ sealed class GooglePurchasingData : PurchasingData {
         val optionId: String,
         val productDetails: ProductDetails,
         val token: String,
+        val billingPeriod: Period?,
```
fire-at-will, line 20: I don't love that this is an optional, but I made it optional to gracefully handle converting `GoogleSubscriptionOption`s to `GooglePurchasingData.Subscription`. When we do that, `GoogleSubscriptionOption` doesn't have a single billing period, but instead has a list of `PricingPhase`s. It's technically possible for that list to be empty, and when so, we set the `GooglePurchasingData.billingPeriod` to null.

This shouldn't happen, but I wanted to avoid throwing if possible. Let me know if you can think of a better way to handle this!

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PurchaseParamsValidator.kt
```diff
@@ -0,0 +1,116 @@
...
+        productIds.add(googleSubscriptionPurchasingData.productId)
+        for (addOnProduct in addOnProducts) {
+            if (addOnProduct is GooglePurchasingData.Subscription) {
+                val addOnProductId = addOnProduct.productId
+                if (productIds.contains(addOnProductId)) {
+                    // The developer is attempting to purchase multiple options on the same product.
+                    val error = PurchasesError(
+                        PurchasesErrorCode.PurchaseInvalidError,
+                        "Multi-line purchases cannot contain multiple purchases for the same product." +
+                            " Multiple purchases for the product $addOnProductId were provided.",
```
tonidero: Maybe 
```suggestion
                            " Multiple instances for the product $addOnProductId were provided.",
```

Since these are not purchases.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PurchaseParamsValidator.kt
tonidero: Hmm I would argue that it should be relatively easy to produce an error, for example, by passing an addon product with a repeated productId... In general, we've been mocking too many dependencies in Android, and we should try to minimize those. But having said that, I don't consider this a blocker.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -29,13 +28,17 @@ class PurchaseParams(val builder: Builder) {
     @get:JvmSynthetic
     internal var presentedOfferingContext: PresentedOfferingContext?
 
+    @get:JvmSynthetic
+    internal var containsAddOnItems: Boolean
```
tonidero: Hmm I'm wondering, could we just rely on the `GooglePurchasingData.Subscription` to check if there are any `addOnProducts` instead? To avoid having this extra state which could get out of sync. We could also add a computed property there, to make it easier

tonidero (approved):
Some comments but nothing really blocking. Looks great!!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
fire-at-will: Yeah, that works for me! Converted it to a computed property here: 66b3ee35ba6cd964343c8ceb57a2ee4f7ba842d8

purchases/src/main/kotlin/com/revenuecat/purchases/models/GooglePurchasingData.kt
tonidero: As you said, this should never happen since there should always be at least one pricing phase... We might want to add a log in this case, but yeah, I think it's ok to keep nullable 👍

fire-at-will: Sounds good, I'll add in a log 👍

PR #2710, opened 2025-10-06
Add TEST_STORE to Store enum

tonidero:
### Description
This adds a new value to the `Store` enum: `TEST_STORE`. This will be the value used when using RevenueCat's test store.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
@@ -238,14 +238,27 @@ enum class Store {
      */
     @SerialName("paddle")
     PADDLE,
+
+    @SerialName("simulated_store")
+    TEST_STORE,
     ;
 
     internal companion object {
         @JvmSynthetic
         fun fromString(text: String): Store {
-            return runCatching {
-                enumValueOf<Store>(text.uppercase())
-            }.getOrDefault(UNKNOWN_STORE)
+            return when (text) {
+                "app_store" -> APP_STORE
+                "mac_app_store" -> MAC_APP_STORE
+                "play_store" -> PLAY_STORE
+                "stripe" -> STRIPE
+                "promotional" -> PROMOTIONAL
+                "amazon" -> AMAZON
+                "rc_billing" -> RC_BILLING
+                "external" -> EXTERNAL
+                "paddle" -> PADDLE
+                "simulated_store" -> TEST_STORE
+                else -> UNKNOWN_STORE
+            }
         }
     }
 }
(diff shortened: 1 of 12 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -253,6 +253,9 @@ data class CustomerCenterConfigData(
             @SerialName("unknown_store")
             UNKNOWN_STORE,
 
+            @SerialName("test_store")
+            TEST_STORE,
```
tonidero, line 257: @RevenueCat/customer-center We might want to add this value to the localizations from the backend.

purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
```diff
@@ -238,14 +238,27 @@ enum class Store {
...
+    TEST_STORE,
     ;
 
     internal companion object {
         @JvmSynthetic
         fun fromString(text: String): Store {
-            return runCatching {
-                enumValueOf<Store>(text.uppercase())
-            }.getOrDefault(UNKNOWN_STORE)
+            return when (text) {
```
tonidero, line 257: I noticed we were parsing this in 2 different ways here and in `EntitlementInfoFactories`. Considering that the code here didn't work with the simulated store (since the name in the response doesn't match the enum value), I copied the other code here and added the new value.

purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
ajpallares, line 1: If I'm reading correctly, this approach wouldn't fix the issue where adding a new `Store` enum case in the future would produce a parsing error in `SubscriptionInfoResponse`. Or am I missing something?

ajpallares (approved):
Looks good! Just a question about potential serialization errors with the `Store` enum in `SubscriptionInfoResponse`

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
facumenzella: https://github.com/RevenueCat/khepri/pull/15295 I'll post the keys in random acts of copy just to make sure we're all in the same page

purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
tonidero: That's correct, I had not fixed that issue in this PR... But I think I can add a fix, now that you mention it :)

ajpallares: Ohh sorry about that! You can do it in a separate PR if you prefer. I just wanted to make sure that I wasn't missing anything. Thanks!

PR #2711, opened 2025-10-07
Add test store maestro test

tonidero:
### Description
Add test store maestro test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -803,6 +803,42 @@ jobs:
           command: |
             ./gradlew ui:revenuecatui:connectedDebugAndroidTest
 
+  run-maestro-paywall-tester-tests:
+    description: "Run Maestro tests for Paywall Tester app"
+    <<: *android-machine-emulator
+    steps:
+      - checkout
+      - install-sdkman
+      - run:
+          name: Replace API_KEY
+          working_directory: examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/
+          command: |
+            sed -i s/\"API_KEY_A\"/\"$TEST_STORE_API_KEY\"/ Constants.kt
+      - android/create_avd:
+          avd_name: test-paywall-tester
+          system_image: system-images;android-35;google_apis;x86_64
+          install: true
+      - android/start_emulator:
+          avd_name: test-paywall-tester
+          post_emulator_launch_assemble_command: ./gradlew examples:paywall-tester:assembleDebug
+      - android/restore_gradle_cache
+      - run:
+          name: Install Paywall Tester on emulator
+          command: |
+            adb install examples/paywall-tester/build/outputs/apk/debug/paywall-tester-debug.apk
+      - run:
+          name: Install Maestro
+          command: |
+            curl -Ls "https://get.maestro.mobile.dev" | bash
+            echo 'export PATH="$HOME/.maestro/bin:$PATH"' >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: Run Maestro tests
+          command: |
+            maestro test maestro_tests/paywall_tester/
+      - store_artifacts:
+          path: maestro_tests/paywall_tester/.maestro/screenshots
+
   update-paywall-preview-resources-submodule:
     docker:
       - image: cimg/ruby:3.2.0
@@ -930,6 +966,9 @@ workflows:
           requires:
             - prepare-tests
       - emerge_purchases_ui_snapshot_tests:
+          requires:
+            - prepare-tests
+      - run-maestro-paywall-tester-tests:
           requires:
             - prepare-tests
       - integration-tests-build: *release-branches
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
@@ -55,7 +55,10 @@ fun AppInfoScreen(
         val state by viewModel.state.collectAsState()
         val currentUserID by remember { derivedStateOf { state.appUserID } }
         val currentApiKeyDescription by remember { derivedStateOf { state.apiKeyDescription } }
+        val currentActiveEntitlements by remember { derivedStateOf { state.activeEntitlements } }
+        Spacer(modifier = Modifier.weight(1f))
         Text(text = "Current user ID: $currentUserID")
+        Text(text = "Current active entitlements: $currentActiveEntitlements")
         Text(text = "Current API key: $currentApiKeyDescription")
         Button(onClick = { showLogInDialog = true }) {
             Text(text = "Log in")
@@ -73,6 +76,10 @@ fun AppInfoScreen(
             tappedOnCustomerCenter()
         }) {
             Text(text = "Show customer center")
+        }
+        Spacer(modifier = Modifier.weight(1f))
+        Button(onClick = { viewModel.refresh() }) {
+            Text(text = "Refresh")
         }
     }
 
@@ -190,11 +197,18 @@ fun AppInfoScreenPreview() {
     AppInfoScreen(
         viewModel = object : AppInfoScreenViewModel {
             override val state: StateFlow<UiState>
-                get() = MutableStateFlow(UiState(appUserID = "test-user-id", apiKeyDescription = "test-api-key"))
+                get() = MutableStateFlow(
+                    UiState(
+                        appUserID = "test-user-id",
+                        apiKeyDescription = "test-api-key",
+                        activeEntitlements = listOf("pro", "premium"),
+                    ),
+                )
 
             override fun logIn(newAppUserId: String) { }
             override fun logOut() { }
             override fun switchApiKey(newApiKey: String) { }
+            override fun refresh() { }
         },
         tappedOnCustomerCenter = {},
     )
--- a/maestro_tests/paywall_tester/test_store_annual_purchase.yaml
+++ b/maestro_tests/paywall_tester/test_store_annual_purchase.yaml
@@ -1,0 +1,20 @@
+appId: com.revenuecat.paywall_tester
+---
+- launchApp:
+    clearState: true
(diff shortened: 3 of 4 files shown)


tonidero:
I would like to add more tests for purchasing the other products and also adding some verifications, but wanted to grab feedback first

.circleci/config.yml
```diff
@@ -803,6 +803,44 @@ jobs:
...
+      - run:
+          name: Install Paywall Tester on emulator
+          command: |
+            adb install examples/paywall-tester/build/outputs/apk/debug/paywall-tester-debug.apk
+      - run:
+          name: Install Maestro
+          command: |
+            curl -Ls "https://get.maestro.mobile.dev" | bash
+            echo 'export PATH="$HOME/.maestro/bin:$PATH"' >> $BASH_ENV
+            source $BASH_ENV
```
JayShortway: I think we either have this or _should_ have this in the orb, as we have some Maestro tests in purchases-ios too.

.circleci/config.yml
```diff
@@ -803,6 +803,44 @@ jobs:
           command: |
             ./gradlew ui:revenuecatui:connectedDebugAndroidTest
 
+  run-maestro-paywall-tester-tests:
+    description: "Run Maestro tests for Paywall Tester app"
```
JayShortway: I don't think we should use Paywall Tester for this, as that's a tester app used by humans. It's easy to break, because not everyone using that project will know that it's used in automated tests too. 

We should use the [Automated SDK Tests](https://app.revenuecat.com/projects/6627e5bc/apps) RC project instead imo, which Antonio created a while back. Its name already implies that humans shouldn't touch it haha.

Maybe we should create a dedicated Android app module too, so it can be fully tailored to testing and doesn't have to take human interaction into account. What do you think?

maestro_tests/paywall_tester/test_store_annual_purchase.yaml
```diff
@@ -0,0 +1,20 @@
...
+---
+- launchApp:
+    clearState: true
+- tapOn: App Info
+- assertVisible:
+    text: "Current active entitlements: []"
+- tapOn: Offerings
+- tapOn: default
+- tapOn: Navigate to paywall
+- assertVisible: Get CatPrrro
```
rickvdl: Is it safe to assume the paywall / products / offerings in this project are stable and won't change? If not, would it make sense to introduce a specific "readonly" project for these automated UI tests? Or would that just complicate things

JayShortway (commented):
Love it! This is the direction we should go imo. Just one comment on which app to use for this.

maestro_tests/paywall_tester/test_store_annual_purchase.yaml
tonidero: It's actually the same project we use for android integration tests and I just added a test store app config and products. Ideally that project should be considered read only (exccept when needing changes), so I think it's ok to assume product/paywall info here IMO.

.circleci/config.yml
tonidero: We actually are using the [same project](https://app.revenuecat.com/projects/2a0cfe53/overview) we use for the purchases module integration tests. I would consider that read-only. Note that we don't really care about the package name for the test store, so we can use whatever app we want.

As for using a separate app/module for this in the project, I'm not completely opposed... but I was thinking we already have a ton of different apps in this repo and it's getting a bit unwieldly... (note that there is a build cost for each new module). Also, considering this will run on every commit, people will know immediately if they've broken the tests I guess. Considering this, I think it might be ok to use paywall tester? If you feel strongly this should be a separate app, I can add a simple module to run some of these tests, and maybe add extra data that can be validated visually through Maestro.

.circleci/config.yml
tonidero: You're right this is likely better in the orb. I will move it there 👍

tonidero: Added in https://github.com/RevenueCat/sdks-circleci-orb/pull/21. Once it's merged, I will update to use it here.

.circleci/config.yml
JayShortway: Ah nice, didn't realize we were using the integration test project. Yea I think we can keep using that one for now, it's also very easy to change indeed.

I do think I prefer a dedicated app module for this. I agree that it's currently a bit disorganized, but the existing `test-apps` folder seems suitable? (We can name the new app `end-to-end` or something?) This allows us to build the app completely for automated testing, without having to consider user experience at all. 

As an example, we could have a single screen per product and easily verify that all the product details are correct, without having to worry (too much) about scrolling or false positives. (With multiple products on a single screen, it's hard to isolate them when asserting things in Maestro.)

maestro_tests/paywall_tester/test_store_annual_purchase.yaml
JayShortway, line 1: Nitpick, but we seem to mostly use hyphens instead of underscores in folder names, which would make it `maestro-tests/paywall-tester/`. I'd also be okay with just `maestro/` btw.

.circleci/config.yml
```diff
@@ -803,6 +803,44 @@ jobs:
...
+          command: |
+            curl -Ls "https://get.maestro.mobile.dev" | bash
+            echo 'export PATH="$HOME/.maestro/bin:$PATH"' >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: Run Maestro tests
+          command: |
+            maestro test maestro_tests/paywall_tester/
+      - store_artifacts:
+          path: ~/.maestro/tests/
```
JayShortway: Btw this crossed my mind! Could we: 

```suggestion
            maestro test --format junit maestro_tests/paywall_tester/
      - store_artifacts:
          path: ~/.maestro/tests/
      - store_test_results:
          path: ~/.maestro/tests/ # No idea if this folder is correct, but we can use --output if needed.
```

([Docs](https://docs.maestro.dev/cli/test-suites-and-reports#generating-reports))

tonidero: Hmm indeed, this would be a nice improvement. Great idea!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/appinfo/AppInfoScreen.kt
```diff
@@ -55,7 +55,10 @@ fun AppInfoScreen(
         val state by viewModel.state.collectAsState()
         val currentUserID by remember { derivedStateOf { state.appUserID } }
         val currentApiKeyDescription by remember { derivedStateOf { state.apiKeyDescription } }
+        val currentActiveEntitlements by remember { derivedStateOf { state.activeEntitlements } }
```
tonidero, line 82: These changes in Paywall tester aren't really needed anymore... But I thought to let them, since they can be useful.

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/main/CustomerInfoPage.kt
```diff
@@ -207,12 +207,7 @@ private fun InfoSection(
 private fun InfoItem(label: String, value: String) {
     Column {
         Text(
-            text = label,
-            style = MaterialTheme.typography.labelMedium,
-            color = MaterialTheme.colorScheme.onSurfaceVariant,
-        )
-        Text(
-            text = value,
+            text = "$label: $value",
```
tonidero, line 210: I did this change so it's easier to test in maestro, since we can't really use tags there, we can really just check for the text itself.

.circleci/config.yml
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2711/commits/bba49ea38394a920fdda83174fd574f149b5aa0b

.circleci/config.yml
```diff
@@ -803,6 +803,41 @@ jobs:
           command: |
             ./gradlew ui:revenuecatui:connectedDebugAndroidTest
 
+  run-maestro-e2e-tests:
```
facumenzella, line 806: Are these working? I couldn't make these work on iOS the last time I tried.

I'll try to mimic this 🤔

tonidero: This one tests seems to work fine, specially now that we don't need to interact with Google play :)

facumenzella: I've re-tested in iOS and I think I am still hitting the same [wall](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/30632/workflows/6854af69-ac5c-47f8-aed3-7b62a3f56e96/jobs/385230):

Looks like more than one fails 🤔 

```
Requested 1 shards, but it cannot be higher than the number of flows (0). Will use 1 shards instead.

Waiting for flows to complete...
[Passed] Manage Subscriptions Empty (4s)
[Failed] Manage Subscriptions Restore Purchases (25s) (Assertion is false: "No past purchases" is visible)

Flow Manage Subscriptions Restore Purchases failed and continueOnFailure is set to false, aborting running sequential Flows

```

tonidero: Were you testing using the test store? FWIW, the test store right now is still pretty limited in terms of functionality (you can't really restore purchases for example, since there is no storage to store existing purchases in the client side).

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/main/CustomerInfoPage.kt
JayShortway: We should be able to use tags with Maestro like so: 

```kotlin
Modifier
  .semantics {
    testTagsAsResourceId = true
    testTag = "test"
  }
```
([docs](https://docs.maestro.dev/platform-support/android-jetpack-compose#interacting-with-composable-with-accessibility))

However, just checking the text itself is totally fine too!

tonidero: Ah nice TIL! But yeah... if we want to share these tests accross SDKs, checking text might be more reliable...

JayShortway: Yea that's a good point indeed!

PR #2713, opened 2025-10-07
Add Customer Center per-view listener for hybrids

facumenzella:
### Motivation
- Add a per-view CustomerCenterListener hook so hybrid apps can override callbacks without relying solely on the global Purchases listener.
- Ensure the non-Compose API mirrors PaywallView.
- Provide a working sample and API coverage so partners can see and compile against the new surface.
- Needed for https://github.com/RevenueCat/purchases-flutter/pull/1476 and https://github.com/RevenueCat/react-native-purchases/pull/1411

### Description
- CustomerCenterView now installs an internal proxy listener, accepts an optional listener via constructor or setter, and keeps the global listener as the fallback.
- Updated the Paywall Tester app: the App Info screen launches a full-screen dialog that hosts CustomerCenterView, logging/toasting to differentiate the per-view listener from the global handler.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CustomerCenterView.kt
```diff
@@ -26,12 +31,47 @@ public class CustomerCenterView : CompatComposeView {
     constructor(
         context: Context,
         dismissHandler: (() -> Unit)? = null,
+        customerCenterListener: CustomerCenterListener? = null,
```
tonidero, line 42: Hmm I would leave the `dismissHandler` as the last parameter, so devs can still do 
```
CustomerCenterView(context) {
  // Dismiss
}
```

This would break with this change I think.

However, I think adding it as a parameter before this, it would also break in Java... So maybe we should add a separate constructor with the 3 parameters to avoid the breaking change in Java.

tonidero (commented):
Looking good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CustomerCenterView.kt
```diff
@@ -50,12 +107,12 @@ public class CustomerCenterView : CompatComposeView {
 
     @Composable
     override fun Content() {
-        CustomerCenterUI(dismissHandler = dismissHandler)
+        CustomerCenterUI(options = customerCenterOptions)
```
vegaro: isn't this losing the `dismissHandler`?

vegaro: ok, nevermind, it calls the private `CustomerCenterUI` which accesses it directly. Maybe we can just remove the private function? It looks like it's useless

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CustomerCenterView.kt
vegaro: Will it break in Java if it has the `@JvmOverloads`?

It will automatically create these three:

```kotlin
(context)
(context, dismissHandler)
(context, dismissHandler, customerCenterListener)
```

I actually think this other constructor is not needed:

```kotlin
constructor(
    context: Context,
    dismissHandler: (() -> Unit)? = null,
)
```

vegaro: just tested it and with those changes it would look like this:

```kotlin
CustomerCenterView(context)
CustomerCenterView(context, dismissHandler)
CustomerCenterView(context, customerCenterListener = listener)
CustomerCenterView(context, dismissHandler, listener)
```

```java
new CustomerCenterView(context);
new CustomerCenterView(context, dismissHandlerFunction);
new CustomerCenterView(context, null, listener);
new CustomerCenterView(context, dismissHandlerFunction, listener);
```

facumenzella: I need more context to address this 😓

vegaro: @facumenzella let's sync, I will help you out. I was actually missing the fact that `dismissHandler` needs to be last

facumenzella: Ok just synced, and we're good to go! cc @tonidero

PR #2715, opened 2025-10-07
Add E2E Test app for integration tests

tonidero:
### Description
Added new test app to run E2E tests with the test store. I did the minimal to make it work, so no view models, and no great navigation system... but it's usable for automated testing. The screen/page structure looks like:
- `MainPage` is the root/initial page, that has 2 tabs `CustomerInfoPage` and `OfferingsPage`. 
- From the `OfferingsPage`, you can navigate to `OfferingScreen`.
- From the `OfferingScreen` you can navigate to `PackageScreen`.

This is the first part to get #2711 in


JayShortway (approved):
This is perfect! Indeed we don't need viewmodels nor a navigation system. Thanks so much for adding this! ❤️

PR #2716, opened 2025-10-07
Add `TestStoreSimulatedPurchaseError` error code

tonidero:
### Description
Adds a new error code to be used when simulating purchase errors in the test store.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
@@ -92,4 +92,5 @@ enum class PurchasesErrorCode(val code: Int, val description: String) {
         36,
         "Request failed signature verification. Please see https://rev.cat/trusted-entitlements for more info.",
     ),
+    TestStoreSimulatedPurchaseError(42, "Purchase failure simulated successfully in Test Store."),
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/simulatedstore/SimulatedStoreBillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/simulatedstore/SimulatedStoreBillingWrapper.kt
@@ -194,8 +194,8 @@ internal class SimulatedStoreBillingWrapper(
             onNegativeButtonClicked = {
                 purchasesUpdatedListener?.onPurchasesFailedToUpdate(
                     PurchasesError(
-                        PurchasesErrorCode.ProductNotAvailableForPurchaseError,
-                        "Test purchase failure: no real transaction occurred",
+                        PurchasesErrorCode.TestStoreSimulatedPurchaseError,
+                        "Simulated error successfully.",
                     ),
                 )
             },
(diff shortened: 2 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt
```diff
@@ -92,4 +92,5 @@ enum class PurchasesErrorCode(val code: Int, val description: String) {
         36,
         "Request failed signature verification. Please see https://rev.cat/trusted-entitlements for more info.",
     ),
+    TestStoreSimulatedPurchaseError(42, "Purchase failure simulated successfully in Test Store."),
```
tonidero, line 95: We need to make sure we sync on the number here @ajpallares, since these are used in the hybrids to parse the correct error across platforms. I think this is the next one in iOS (we didn't seem to have added some in Android), so I just continue from the last one in iOS.

ajpallares: That's correct! I also use code 42 for this one here: https://github.com/RevenueCat/purchases-ios/blob/72c42f633834ef40523391f74534550fbd76c115/Sources/Error%20Handling/ErrorCode.swift#L67

Done in this PR https://github.com/RevenueCat/purchases-ios/pull/5634

purchases/src/main/kotlin/com/revenuecat/purchases/simulatedstore/SimulatedStoreBillingWrapper.kt
```diff
@@ -194,8 +194,8 @@ internal class SimulatedStoreBillingWrapper(
             onNegativeButtonClicked = {
                 purchasesUpdatedListener?.onPurchasesFailedToUpdate(
                     PurchasesError(
-                        PurchasesErrorCode.ProductNotAvailableForPurchaseError,
-                        "Test purchase failure: no real transaction occurred",
+                        PurchasesErrorCode.TestStoreSimulatedPurchaseError,
+                        "Simulated error successfully.",
```
ajpallares, line 200: What would be the final log that developers see? Not critical, but it would be great to have the logs be as similar as possible in all platforms

ajpallares: I ask because in iOS, I'm not adding an extra message here when creating the error. It would be redundant perhaps?

This is the error log currently in iOS:
> ERROR: [TEST STORE] Purchase failure simulated successfully in Test Store.

tonidero: Well this error is not currently logged in this PR. I added the logs in [this other PR](https://github.com/RevenueCat/purchases-android/pull/2710/files#diff-6100e84aaaffb8823983da3606e5cffac20e73efa50a6decf030be17c02b71ecR196). I will change that error log in that PR to match iOS :)

tonidero: Changed log [here](https://github.com/RevenueCat/purchases-android/pull/2710/commits/1adacca9adaa3bc1a596efcdd9315fe9ff693a61).

tonidero: Regarding the added underlying error message... I'm just adding something in case some user is accessing the `underlyingErrorMessage` instead of the `message` property.

ajpallares: Ohh, I see. Thanks for explaining!

PR #2717, opened 2025-10-07
Finalize video component to include a checksum, and optimize memory usage for large file downloads

JZDesign:
### Checklist

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -51,7 +51,6 @@ import com.revenuecat.purchases.subscriberattributes.caching.SubscriberAttribute
 import com.revenuecat.purchases.utils.CoilImageDownloader
 import com.revenuecat.purchases.utils.IsDebugBuildProvider
 import com.revenuecat.purchases.utils.OfferingImagePreDownloader
-import com.revenuecat.purchases.utils.OfferingVideoPredownloader
 import com.revenuecat.purchases.utils.isAndroidNOrNewer
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencyManager
 import java.net.URL
@@ -335,7 +334,6 @@ internal class PurchasesFactory(
                 backend,
                 OfferingsFactory(billing, offeringParser, dispatcher),
                 OfferingImagePreDownloader(coilImageDownloader = CoilImageDownloader(application)),
-                OfferingVideoPredownloader(application),
                 diagnosticsTracker,
                 offeringFontPreDownloader = offeringFontPreDownloader,
             )
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
@@ -1,0 +1,125 @@
+package com.revenuecat.purchases.models
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import java.io.File
+import java.io.FileInputStream
+import java.io.InputStream
+import java.security.MessageDigest
+
+/**
+ * A checksum for validating file integrity
+ */
+@InternalRevenueCatAPI
+@Serializable
+data class Checksum(
+    /** The algorithm used to generate the checksum */
+    @SerialName("algo") val algorithm: Algorithm,
+
+    /** The checksum hash value */
+    @SerialName("value") val value: String
+) {
+    /**
+     * Supported hashing algorithms
+     */
+    @Serializable
+    enum class Algorithm(val algorithmName: String) {
+        @SerialName("sha256") SHA256("SHA-256"),
+        @SerialName("sha384") SHA384("SHA-384"),
+        @SerialName("sha512") SHA512("SHA-512"),
+        @SerialName("md5") MD5("MD5");
+
+        companion object {
+            fun fromString(value: String): Algorithm? = when (value.lowercase()) {
+                "sha256" -> SHA256
+                "sha384" -> SHA384
+                "sha512" -> SHA512
+                "md5" -> MD5
+                else -> null
+            }
+        }
+    }
+
+    companion object {
+        private const val BUFFER_SIZE = 256 * 1024 // 256KB
+
+        /**
+         * Generate a checksum from data
+         */
+        fun generate(data: ByteArray, algorithm: Algorithm): Checksum {
+            val digest = MessageDigest.getInstance(algorithm.algorithmName)
+            val hash = digest.digest(data)
+            return Checksum(algorithm, hash.toHexString())
+        }
+
+        /**
+         * Generate a checksum from a file using streaming
+         * (memory efficient - doesn't load entire file)
+         */
+        fun generate(file: File, algorithm: Algorithm): Checksum {
+            val digest = MessageDigest.getInstance(algorithm.algorithmName)
+
+            FileInputStream(file).use { inputStream ->
+                val buffer = ByteArray(BUFFER_SIZE)
+                var bytesRead: Int
+
+                while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                    digest.update(buffer, 0, bytesRead)
+                }
+            }
+
+            val hash = digest.digest()
+            return Checksum(algorithm, hash.toHexString())
+        }
+
+        /**
+         * Generate checksum from an InputStream while reading it
+         * (useful for validating during download)
+         */
+        fun generateAndConsume(
+            inputStream: InputStream,
+            algorithm: Algorithm,
+            onData: (ByteArray, Int) -> Unit
+        ): Checksum {
+            val digest = MessageDigest.getInstance(algorithm.algorithmName)
+            val buffer = ByteArray(BUFFER_SIZE)
+            var bytesRead: Int
+
+            while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                digest.update(buffer, 0, bytesRead)
+                onData(buffer, bytesRead)
+            }
+
+            val hash = digest.digest()
+            return Checksum(algorithm, hash.toHexString())
+        }
+    }
+
+    /**
+     * Compare this checksum to another
+     * @throws ChecksumValidationException if checksums don't match
+     */
+    fun compare(other: Checksum) {
+        if (this.algorithm != other.algorithm) {
+            throw ChecksumValidationException(
+                "Algorithm mismatch: expected ${other.algorithm}, got ${this.algorithm}"
+            )
+        }
+
+        if (this.value.lowercase() != other.value.lowercase()) {
+            throw ChecksumValidationException(
+                "Checksum mismatch: expected ${other.value}, got ${this.value}"
+            )
+        }
+    }
+
+    /**
+     * Exception thrown when checksum validation fails
+     */
+    class ChecksumValidationException(message: String) : Exception(message)
+}
+
+internal fun ByteArray.toHexString(): String {
+    return joinToString("") { "%02x".format(it) }
+}
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
@@ -3,9 +3,6 @@ package com.revenuecat.purchases.models
 import com.revenuecat.purchases.InternalRevenueCatAPI
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import java.io.File
-import java.io.FileInputStream
-import java.io.InputStream
 import java.security.MessageDigest
 
 /**
@@ -18,17 +15,25 @@ data class Checksum(
     @SerialName("algo") val algorithm: Algorithm,
 
     /** The checksum hash value */
-    @SerialName("value") val value: String
+    @SerialName("value") val value: String,
 ) {
     /**
      * Supported hashing algorithms
      */
     @Serializable
     enum class Algorithm(val algorithmName: String) {
-        @SerialName("sha256") SHA256("SHA-256"),
-        @SerialName("sha384") SHA384("SHA-384"),
-        @SerialName("sha512") SHA512("SHA-512"),
-        @SerialName("md5") MD5("MD5");
+        @SerialName("sha256")
+        SHA256("SHA-256"),
+
+        @SerialName("sha384")
+        SHA384("SHA-384"),
+
+        @SerialName("sha512")
+        SHA512("SHA-512"),
+
+        @SerialName("md5")
+        MD5("MD5"),
+        ;
 
         companion object {
             fun fromString(value: String): Algorithm? = when (value.lowercase()) {
@@ -42,8 +47,6 @@ data class Checksum(
     }
 
     companion object {
-        private const val BUFFER_SIZE = 256 * 1024 // 256KB
-
         /**
          * Generate a checksum from data
          */
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
@@ -52,48 +52,6 @@ data class Checksum(
             val hash = digest.digest(data)
             return Checksum(algorithm, hash.toHexString())
         }
-
-        /**
-         * Generate a checksum from a file using streaming
-         * (memory efficient - doesn't load entire file)
-         */
-        fun generate(file: File, algorithm: Algorithm): Checksum {
-            val digest = MessageDigest.getInstance(algorithm.algorithmName)
-
-            FileInputStream(file).use { inputStream ->
-                val buffer = ByteArray(BUFFER_SIZE)
-                var bytesRead: Int
-
-                while (inputStream.read(buffer).also { bytesRead = it } != -1) {
-                    digest.update(buffer, 0, bytesRead)
-                }
-            }
-
-            val hash = digest.digest()
-            return Checksum(algorithm, hash.toHexString())
-        }
-
-        /**
-         * Generate checksum from an InputStream while reading it
-         * (useful for validating during download)
-         */
-        fun generateAndConsume(
-            inputStream: InputStream,
-            algorithm: Algorithm,
-            onData: (ByteArray, Int) -> Unit
-        ): Checksum {
-            val digest = MessageDigest.getInstance(algorithm.algorithmName)
-            val buffer = ByteArray(BUFFER_SIZE)
-            var bytesRead: Int
-
-            while (inputStream.read(buffer).also { bytesRead = it } != -1) {
-                digest.update(buffer, 0, bytesRead)
-                onData(buffer, bytesRead)
-            }
-
-            val hash = digest.digest()
-            return Checksum(algorithm, hash.toHexString())
-        }
     }
 
     /**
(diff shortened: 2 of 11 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -41,7 +41,7 @@ internal fun VideoComponentView(
     repository: FileRepository = Purchases.sharedInstance.fileRepository,
 ) {
     val videoState = rememberUpdatedVideoComponentState(style, state)
-
+    println("HELP_ME_OB1 - In video view")
```
JZDesign:
```suggestion
```

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -21,20 +24,23 @@ internal class OfferingVideoPredownloader(
 
     fun downloadVideos(offering: Offering) {
         if (shouldPredownload) {
-            // WIP: We will add a remote flag in the offering metadata that will indicate if we should download
-            // the high res videos or not. For now, we want to only download the low-res to ensure we don't rack up
-            // high cloudfront costs
+            // WIP: We will add a remote flag in the offering metadata that will indicate if we should
```
rickvdl, line 27: Is this still WIP?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -335,7 +334,6 @@ internal class PurchasesFactory(
                 backend,
                 OfferingsFactory(billing, offeringParser, dispatcher),
                 OfferingImagePreDownloader(coilImageDownloader = CoilImageDownloader(application)),
-                OfferingVideoPredownloader(application),
```
rickvdl, line 338: Where was the initialization moved? :)

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +209,102 @@ private class DefaultFileCache(
...
 
-    override fun saveData(data: ByteArray, uri: URI) =
-        writeBytes(data, File(uri))
+    override fun saveData(inputStream: InputStream, uri: URI, checksum: Checksum?) {
+        val finalFile = File(uri)
+        val tempFile = File.createTempFile(
+            "rc_download_",
+            ".tmp",
+            finalFile.parentFile,
+        )
```
JayShortway, line 234: I've never used this. Does this work without storage permissions?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +209,102 @@ private class DefaultFileCache(
...
+            val buffer = ByteArray(BUFFER_SIZE)
+            var bytesRead: Int
+
+            while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                // Update checksum
+                digest.update(buffer, 0, bytesRead)
+
+                // Write to file
+                outputStream.write(buffer, 0, bytesRead)
+            }
```
JayShortway, line 291: Nice!

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +209,102 @@ private class DefaultFileCache(
...
+
+        val hash = digest.digest()
+
+        val computedChecksum = Checksum(
+            checksum.algorithm,
+            hash.toHexString(),
+        )
+
+        // throws if comparison is invalid
+        computedChecksum.compare(checksum)
```
JayShortway: Can we make this function return a boolean instead? It's very easy to forget that we need to catch an exception, as Kotlin doesn't make us do it.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +209,102 @@ private class DefaultFileCache(
...
+        FileOutputStream(file).use { outputStream ->
+            val buffer = ByteArray(BUFFER_SIZE)
+            var bytesRead: Int
+
+            while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                outputStream.write(buffer, 0, bytesRead)
+            }
+
+            outputStream.flush()
+        }
```
JayShortway, line 270:
```suggestion
        FileOutputStream(file).use { outputStream ->
            inputStream.copyTo(outputStream, bufferSize = BUFFER_SIZE)
        }
```

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +209,102 @@ private class DefaultFileCache(
...
+            // Stream to temp file, optionally calculating checksum if available
+            if (checksum != null) {
+                streamToFileAndCompareChecksum(inputStream, tempFile, checksum)
+            } else {
+                streamToFile(inputStream, tempFile)
+            }
+
+            // We will first attempt to just move the file in one shot. This is an atomic and very fast operation
+            // but it is prone to failure if the temp directory and cache directory are on different volumes
+            // like the SDCard and the device's internal storage
```
JayShortway, line 246: Can we not ensure that the `tempFile` and `finalFile` are on the same volume by making `tempFile` the same as `finalFile` except the extension? E.g. swap `.mp4` with `.tmp`.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -30,128 +33,136 @@ import java.security.MessageDigest
...
     private val urlConnectionFactory: UrlConnectionFactory = DefaultUrlConnectionFactory(),
 ) : FileRepository {
 
+    // Cache key now includes checksum
+    internal data class CacheKey(
+        val url: URL,
+        val checksum: Checksum?,
+    )
+
     // Convenience constructor for Android
```
JayShortway: Claude is a bit overly descriptive in its comments imo 😄 
```suggestion
    internal data class CacheKey(
        val url: URL,
        val checksum: Checksum?,
    )

```

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +209,102 @@ private class DefaultFileCache(
...
+            // but it is prone to failure if the temp directory and cache directory are on different volumes
+            // like the SDCard and the device's internal storage
+            if (!tempFile.renameTo(finalFile)) {
+                // If that rename fails, we will try again with a slower,
+                // non-atomic, operation that will work across volumes
+                tempFile.copyTo(finalFile, overwrite = true)
+            }
+        } finally {
+            tempFile.delete()
+        }
```
JayShortway, line 260: We should also delete `finalFile` I think? It could have (partially) been written.

purchases/src/test/java/com/revenuecat/purchases/models/ChecksumTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Test
+    fun `generate checksum from data - SHA512`() {
+        val data = "Test".toByteArray()
+        val checksum = Checksum.generate(data, Checksum.Algorithm.SHA512)
+
+        assertThat(checksum.algorithm).isEqualTo(Checksum.Algorithm.SHA512)
+        assertThat(checksum.value).hasSize(128) // SHA512 produces 128 hex chars
+    }
```
JayShortway, line 50: I think these should assert the literal checksum value instead of only the length, like we do for SHA256.

purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Compare this checksum to another
+     * @throws ChecksumValidationException if checksums don't match
+     */
+    fun compare(other: Checksum) {
+        if (this.value.lowercase() != other.value.lowercase() || this.algorithm != other.algorithm) {
+            throw ChecksumValidationException()
+        }
+    }
```
JayShortway: I think we should override `equals()` instead of writing a new function. That allows us to compare checksums using standard equality operators (`=` / `!=`) and returns a boolean instead of throwing an exception.

purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsManagerTest.kt
```diff
@@ -442,73 +435,6 @@ class OfferingsManagerTest {
 
     // endregion getOfferings
 
-    // region pre download offering videos
```
JayShortway, line 445: I understand that we're still predownloading low-res videos, and potentially high-res based on a remote flag. If so, we should adjust these tests instead of removing them imo.

purchases/src/test/java/com/revenuecat/purchases/storage/DefaultFileCacheTest.kt
```diff
@@ -0,0 +1,335 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.io.ByteArrayInputStream
+import java.io.File
+import java.io.IOException
+import java.io.InputStream
+import java.net.URL
+
+@OptIn(InternalRevenueCatAPI::class)
```
JayShortway: This should not be needed? At least not on my machine haha.

purchases/src/test/java/com/revenuecat/purchases/storage/DefaultFileCacheTest.kt
```diff
@@ -0,0 +1,335 @@
...
+@RunWith(RobolectricTestRunner::class)
```
JayShortway: Nitpick:

```suggestion
@RunWith(AndroidJUnit4::class)
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoPlaybackViewModel.kt
```diff
@@ -0,0 +1,54 @@
...
+import androidx.compose.runtime.Stable
+import androidx.lifecycle.ViewModel
+import com.revenuecat.purchases.InternalRevenueCatAPI
+
+/**
+ * ViewModel that manages video playback state across activity recreation.
+ * Survives configuration changes and activity recreation while maintaining
+ * playback position and play/pause state for each video.
+ */
+internal class VideoPlaybackViewModel : ViewModel() {
```
JayShortway: We should not use ViewModels for low-level composables, only for screen-level ones. See also [here](https://developer.android.com/topic/libraries/architecture/viewmodel#jetpack-compose). 

A `ViewModel` is not needed to survive Activity recreation. `rememberSaveable()` can be used for that (which I see is already being used).

This `ViewModel` is mostly wrapping a map, which we can do directly in the composable using `mutableStateMapOf()`. This will automatically recompose the `Video` if the map changes. If you don't want that we can use a regular `mutableMapOf()`.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: I believe so yes.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: Ah, I'll need to create a try catch around the copy to and do a delete there. We don't want to delete the final file unless there is an error, but we always want to delete the temp file

purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsManagerTest.kt
JZDesign: Actually, we're kinda wishy-washy on that right now. Josh mentioned something that happened with a large app going viral and the infra costs getting out of hand with images, so we chose to prewarm only low res images.

Our prewarming is still not intelligent in any way. If an app has 50 paywalls we prewarm everything. This is part of why it was so expensive.

But videos are larger files. So if we have a similar situation arise where there are many paywalls with many video files, the prewarm of the low res video could cost us a LOT of money.

So I've temporarily pulled the video prewarming out, though we are prewarming the first frame image still.

We will be weighing options and planning on how to deduce what should be prewarmed later. So I've left most of the plumbing and just removed the usage of it for now.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
JZDesign: https://github.com/RevenueCat/purchases-android/pull/2717#discussion_r2413577076

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
JZDesign: Yes, see https://github.com/RevenueCat/purchases-android/pull/2717#discussion_r2413577076

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoPlaybackViewModel.kt
JZDesign: I will try the `mutableStateMapOf` function. It was not preserving state on rotation when using mutableStateOf and remember so that's why I opted for ViewModel

JZDesign: From those docs:

>  Don't use them as state holders of reusable UI components such as chip groups or forms. Otherwise, you'd get the same ViewModel instance in different usages of the same UI component under the same ViewModelStoreOwner unless you use an explicit view model key per chip.


Ah… I get it. But I technically could use the view model if I specify unique keys per video component.

I'll still try the mutableStateMapOf function first

JayShortway: Yes it's a bit of "best practice" to not use ViewModels in that way.

If you use `rememberSaveable` (instead of `remember`) it should survive Activity recreation. Let me know how it goes!

JZDesign: That failed too actually, but I'll try again. I was having some interesting compile time issues yesterday where it wasn't recognizing serializable so that may have played a part. I'll invalidate and restart before trying again

JZDesign: done. Thanks for pushing on this one.

purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsManagerTest.kt
JZDesign: I suppose I could put that all back in and add a compiler flag like the one that is blocking the whole video component

purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
```diff
@@ -65,5 +65,5 @@ internal fun checkIfVideoComponentIsEnabled(): Boolean = BuildConfig.ENABLE_VIDE
 
 // Expose for UI package
 @InternalRevenueCatAPI
-val isVideoComponentEnabled: Boolean
-    get() = checkIfVideoComponentIsEnabled()
+val isVideoComponentEnabled: Boolean = true
+//    get() = checkIfVideoComponentIsEnabled()
```
JayShortway: Was this supposed to be committed? 😄

purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    override fun equals(other: Any?): Boolean {
+        if (other !is Checksum) return false
+        if (this.value.lowercase() != other.value.lowercase() || this.algorithm != other.algorithm) {
+            return false
+        }
+        return true
+    }
```
JayShortway, line 66: Nice! 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/models/Checksum.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Exception thrown when checksum validation fails
+     */
+    class ChecksumValidationException : Exception()
```
JayShortway, line 71: This can go I think?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JayShortway: Oh right of course!

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +207,103 @@ private class DefaultFileCache(
...
+        val computedChecksum = Checksum(
+            checksum.algorithm,
+            hash.toHexString(),
+        )
+
+        return if (checksum == computedChecksum) {
+            Result.success(Unit)
+        } else {
+            Result.failure(Checksum.ChecksumValidationException())
         }
```
JayShortway: Could this be: 

```suggestion
        return checksum == computedChecksum
```

JayShortway: Then at the call site (above), we can do something like: 

```kotlin
val checksumValid = if (checksum != null) {
  streamToFileAndCompareChecksum(inputStream, tempFile, checksum).getOrThrow()
} else {
  streamToFile(inputStream, tempFile)
  true
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -417,13 +386,8 @@ private fun Video(
...
 
-    val saved = rememberSaveable(
-        key = key,
-        saver = SavedPlaybackSaver,
-    ) {
-        // Try to restore from static holder first, fallback to default
-        val restored = VideoStateHolder.restore(key)
-        mutableStateOf(restored)
+    val savedState = rememberSaveable(key = key) {
+        mutableMapOf<String, VideoPlaybackState>()
```
JayShortway, line 390: 🙌

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoPlaybackViewModel.kt
JayShortway: Very nice!! Thank you! 🙏

purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsManagerTest.kt
JayShortway: Ah got it, thanks for explaining the context! If the current state of the world is that we're not prewarming videos, then I'd leave the code out. No need to keep code around "just in case", as we can always look back in our git history to find it.

purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
JZDesign: lol… 

I thought I unchecked that box in the git gui…

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -193,26 +207,103 @@ private class DefaultFileCache(
...
+            // but it is prone to failure if the temp directory and cache directory are on different volumes
+            // like the SDCard and the device's internal storage
+            if (!tempFile.renameTo(finalFile)) {
+                // If that rename fails, we will try again with a slower,
+                // non-atomic, operation that will work across volumes
+                @Suppress("TooGenericExceptionCaught")
+                try {
+                    tempFile.copyTo(finalFile, overwrite = true)
+                } catch (_: Exception) {
+                    finalFile.delete()
```
tonidero, line 255: Could be good to add a log in this case

PR #2718, opened 2025-10-08
Disable offline entitlements in test store

tonidero:
### Description
We don't want to use offline entitlements for the test store, since the store in the client won't actually return any purchases, so it's kinda useless.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/OfflineEntitlementsManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/OfflineEntitlementsManager.kt
@@ -3,6 +3,7 @@ package com.revenuecat.purchases.common.offlineentitlements
 import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesErrorCode
+import com.revenuecat.purchases.Store
 import com.revenuecat.purchases.common.AppConfig
 import com.revenuecat.purchases.common.Backend
 import com.revenuecat.purchases.common.caching.DeviceCache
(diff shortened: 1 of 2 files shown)


purchases/src/test/java/com/revenuecat/purchases/common/offlineentitlements/OfflineEntitlementsManagerTest.kt
```diff
@@ -200,6 +204,13 @@ class OfflineEntitlementsManagerTest {
         assertThat(offlineEntitlementsManager.shouldCalculateOfflineCustomerInfoInPostReceipt(isServerError)).isFalse
     }
 
+    @Test
+    fun `shouldCalculateOfflineCustomerInfoInPostReceipt returns false if using test store`() {
```
tonidero, line 208: Plan to add some integration tests for this, but for now, this should work.

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/OfflineEntitlementsManager.kt
```diff
@@ -116,10 +117,12 @@ internal class OfflineEntitlementsManager(
...
 
     // We disable offline entitlements in observer mode (finishTransactions = true) since it doesn't
-    // provide any value and simplifies operations in that mode.
+    // provide any value and simplifies operations in that mode. Also on test store, since we don't have a store
+    // to store purchases in the client.
     private fun isOfflineEntitlementsEnabled() = appConfig.finishTransactions &&
         appConfig.enableOfflineEntitlements &&
-        !appConfig.customEntitlementComputation
+        !appConfig.customEntitlementComputation &&
+        appConfig.store != Store.TEST_STORE
```
rickvdl, line 125: Nitpick: On iOS using enums I'd do something like this (not sure if this is a thing in Kotlin), but it avoids using the Store cases throughout the codebase which could be kind of nice.
```
enum Store {
  case appStore
  case test
}

extension Store {
  var isOfflineEntitlementsSupported: Bool {
    switch self {
    case .appStore: return true
    case .test: return false
    }
  }
}

tonidero: I like the cleanup! Will do so tomorrow, thanks!

ajpallares: I'd say in this case it would be a matter of style. Even in iOS, I would keep it like Toni did because this way, the logic of "Offline entitlements is not supported for the Test Store" is implemented in the OfflineEntitlementsManager, which is where it should belong IMHO.
However, it would be OK to do if the `Store` extension were defined in the `OfflineEntitlementsManager` file

tonidero: Yeah, I do agree it's a matter of style... but I feel having a check in AppConfig, might be a bit cleaner. We can probably do the check on the store there. Will do this in a follow up PR since I want to get started on the SDK releases with this.

PR #2724, opened 2025-10-10
Extracted the currency locale matching logic from the PaywallState to the Purchases class for reuse

rickvdl:
### Context

The public APIs for getting the price string of a product take `Locale` which is used for formatting the price. These methods also have a default value for the locale, which is `Locale.getDefault()`. Which seems fine, since it will format the price according to the user's locale. 

### The problem
The problem here arises when the `storefrontCountryCode` is different from the country/region in the user's locale. This can cause situations where price is formatted using the store's currency code, but using the user's locale. Whereas the localized price string returned by the store is using the storefront's country code as the region for formatting the price string.

### The solution
In https://github.com/RevenueCat/purchases-android/pull/2604 a fix was made specifically for paywalls which added some logic to find the best matching locale between the `storefrontCountryCode` and the user's device locale. 

In this PR this logic was extracted to `CurrencyLocaleResolver` and reused across paywalls, `StoreProduct` and `PricingPhase`. Making sure all APIs (AFAK) are using the same locale for formatting prices.

### Changes introduces
- `CurrencyLocaleResolver` centralizes the logic introduced in https://github.com/RevenueCat/purchases-android/pull/2604
- Public APIs in `StoreProduct` and `PricingPhase` use the value derived from `CurrencyLocaleResolver` instead of `Locale.getDefault()` as default locale
- Added unit tests for `CurrencyLocaleProvider` and it's uses

### Checklist

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/CurrencyLocaleResolver.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/CurrencyLocaleResolver.kt
@@ -1,0 +1,53 @@
+package com.revenuecat.purchases.utils
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import java.util.Locale
+
+/**
+ * Utility class for resolving the appropriate Locale for currency formatting based on storefront country code
+ * and device locales.
+ */
+@InternalRevenueCatAPI
+object CurrencyLocaleResolver {
+
+    /**
+     * Determines the Locale used for formatting currencies based on the `storefrontCountryCode` and the device's
+     * available locale's. The `storefrontCountryCode` argument will be used instead of the cached
+     * `storefrontCountryCode` if provided. The `locale` argument is used as fallback in case no
+     * `storefrontCountryCode` is available or when there are no matching device locale's.
+     *
+     * @param storefrontCountryCode The storefront country code (e.g., "US", "NL"). If null, only the locale
+     * parameter is returned.
+     * @param locale The device locale to use as a fallback. Defaults to system default locale.
+     * @return The best matching locale for currency formatting.
+     */
+    @JvmStatic
+    fun resolve(
+        storefrontCountryCode: String?,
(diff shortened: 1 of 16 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/utils/CurrencyLocaleResolver.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import java.util.Locale
+
+/**
+ * Utility class for resolving the appropriate Locale for currency formatting based on storefront country code
+ * and device locales.
+ */
+@InternalRevenueCatAPI
+object CurrencyLocaleResolver {
```
tonidero, line 11: Can this be `internal` now?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -174,22 +159,10 @@ internal sealed interface PaywallState {
...
-                    val javaLocale = availableStorefrontCountryLocalesByLanguage[deviceLanguageCode]
-                        ?: availableStorefrontCountryLocalesByLanguage.values.firstOrNull()
-                        ?: Locale.Builder()
-                            .setLocale(locale.toJavaLocale())
-                            .setRegion(storefrontCountryCode.uppercase())
-                            .build()
-
-                    javaLocale.toComposeLocale()
-                }
+                CurrencyLocaleResolver.resolve(
```
tonidero, line 162: Hmm I wonder, could we just use `purchases.getDefaultCurrencyLocale()`? That way we can make `CurrencyLocaleResolver` be `internal`.

tonidero (commented):
Just one comment, the rest looks great to me!

JZDesign (approved):

PR #2725, opened 2025-10-10
Turn on Video Component

JZDesign:


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -148,6 +148,7 @@ class Purchases internal constructor(
     /**
      * The currently configured FileRepository
      */
+    @get:JvmSynthetic
```
tonidero, line 151: Hmm I think this is fine since it's internal, but any reason to make the getter not available in Java? Fine to leave as is btw, just curious.

JayShortway: I asked for this since they were showing up in the API dump files. 😇

PR #2726, opened 2025-10-14
Adds new experimental `storefrontLocale` APIs.

JayShortway:
## Description
Adds experimental APIs providing the storefront country as `Locale` instead of as 2-letter code. This allows developers more flexibility in handling this information.

Relates to [SDK-4119](https://linear.app/revenuecat/issue/SDK-4119/consolidate-storefront-format-between-ios-and-android).

### DIFF
--- a/purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
+++ b/purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
@@ -177,7 +177,8 @@ internal class PurchasesTest : BasePurchasesTest() {
     @Test
     fun `storefront locale is correctly constructed from storefront country`() {
         // Arrange
-        val regionCode = "US"
+        val regionCode = "NL"
+        val expectedIso3Country = "NLD"
         val expectedLocale = Locale.Builder().setRegion(regionCode).build()
         every { mockBillingAbstract.getStorefront(onSuccess = captureLambda(), onError = any()) }.answers {
             lambda<(String) -> Unit>().captured.invoke(regionCode)
@@ -191,6 +192,9 @@ internal class PurchasesTest : BasePurchasesTest() {
         // Assert
         assertThat(actualLocaleFromCallback).isEqualTo(expectedLocale)
         assertThat(actualLocaleFromProperty).isEqualTo(expectedLocale)
+        // The below assertion is added so we're notified if our assumptions are no longer true. If it starts failing,
+        // we should look into providing another way of getting the 3-letter storefront country code.
+        assertThat(actualLocaleFromProperty?.isO3Country).isEqualTo(expectedIso3Country)
     }
 
     @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -174,6 +187,27 @@ class Purchases internal constructor(
...
+     * This method will try to obtain the Store (Google/Amazon) locale.
+     * If there is any error, it will return null and log said error.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    fun getStorefrontLocale(callback: GetStorefrontLocaleCallback) {
+        purchasesOrchestrator.getStorefrontCountryCode(
+            object : GetStorefrontCallback {
+                override fun onReceived(storefrontCountryCode: String) {
+                    callback.onReceived(
+                        storefrontLocale = Locale.Builder().setRegion(storefrontCountryCode).build(),
```
tonidero: Hmm I wonder if we should just add a method in `PurchasesOrchestrator` in order to move this bit of logic there... I prefer not to add any logic if I can to the facade `Purchases.kt` files.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -104,6 +108,15 @@ class Purchases internal constructor(
...
         @Synchronized get() = purchasesOrchestrator.storefrontCountryCode
 
+    /**
+     * The storefront locale. **Note:** this locale only has a region set.
+     * This may be null if the store hasn't connected yet or fetching the country code hasn't finished or failed.
+     * To get the country code asynchronously use [getStorefrontLocale] or [awaitStorefrontLocale].
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    val storefrontLocale: Locale?
+        get() = storefrontCountryCode?.let { Locale.Builder().setRegion(it).build() }
```
tonidero: Same as previous comment

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -174,6 +187,27 @@ class Purchases internal constructor(
         purchasesOrchestrator.getStorefrontCountryCode(callback)
     }
 
+    /**
+     * This method will try to obtain the Store (Google/Amazon) locale.
+     * If there is any error, it will return null and log said error.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    fun getStorefrontLocale(callback: GetStorefrontLocaleCallback) {
```
tonidero, line 196: We might want to add some basic tests for this?

JayShortway: Good point, added here: https://github.com/RevenueCat/purchases-android/pull/2726/commits/b02df8d1af42379dee0872ed8533f9f06b6623d7

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: Moved here: https://github.com/RevenueCat/purchases-android/pull/2726/commits/ddcd66e89285236d75a05398c09548ba2a0402d4

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: Moved here: https://github.com/RevenueCat/purchases-android/pull/2726/commits/ddcd66e89285236d75a05398c09548ba2a0402d4

purchases/src/testDefaults/kotlin/com/revenuecat/purchases/PurchasesTest.kt
```diff
@@ -172,6 +173,26 @@ internal class PurchasesTest : BasePurchasesTest() {
...
+        }
+
+        // Act
+        var actualLocaleFromCallback: Locale? = null
+        purchases.getStorefrontLocaleWith { actualLocaleFromCallback = it }
+        val actualLocaleFromProperty = purchases.storefrontLocale
+
+        // Assert
+        assertThat(actualLocaleFromCallback).isEqualTo(expectedLocale)
+        assertThat(actualLocaleFromProperty).isEqualTo(expectedLocale)
```
tonidero, line 194: Maybe worth adding an assertion that the `locale.isO3Country` is the expected value? I guess we're testing the `Locale` type in that case, but still seems useful since it's one of the main reasons for this new API.

JayShortway: Hmm, I don't think that's very actionable. In the hypothetical scenario it fails, there's not much we can do. I did add another test for the non-happy case though: https://github.com/RevenueCat/purchases-android/pull/2726/commits/15b86e9aaeefcb71c6fbcdd1327ed9aee84370c1.

tonidero: Right I agree there isn't much we can do if it fails. I think it can be useful so we're aware of some change breaking our assumptions. But I agree, it's not that useful, so happy to keep as is.

JayShortway: Yea that's fair. I added the assertion with a comment explaining why we have it: https://github.com/RevenueCat/purchases-android/pull/2726/commits/36647db122a842ea462b52942d66522d4f9115e6

PR #2728, opened 2025-10-14
Add non paid revenue reporting infra

tonidero:
### Description
Implements ad event tracking infrastructure for RevenueCat. This provides the client-side foundation needed to report ad impressions, clicks, and revenue alongside subscription data for comprehensive LTV tracking.

In follow-up PRs:
- [x] Add size limit to ad events file https://github.com/RevenueCat/purchases-android/pull/2841
- [x] Flush multiple batches, instead of a single one (with a limit) https://github.com/RevenueCat/purchases-android/pull/2842

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -347,11 +347,12 @@ platform :android do
 
         # Create PR against the current branch (not main)
         create_pr_if_necessary(
-          repo: repo_name,
-          branch_name: pr_branch_name,
+          github_pr_token: ENV["GITHUB_TOKEN"],
+          repo_name: repo_name,
+          base_branch: current_branch,
+          head_branch: pr_branch_name,
           title: title,
           body: body,
-          base: current_branch,
           labels: ["pr:other"],
           team_reviewers: ["coresdk"]
         )
@@ -807,25 +808,6 @@ DESC
     end
   end
 
-  private_lane :create_or_checkout_branch do |options|
-    branch_name = options[:branch_name] || UI.user_error!("No branch name provided")
-
-    # Check if branch exists locally or remotely
-    branch_exists_locally = sh("git", "branch", "--list", branch_name).length > 0
-    branch_exists_remotely = sh("git", "ls-remote", "--heads", "origin", branch_name).length > 0
-
-    if branch_exists_locally || branch_exists_remotely
-      UI.message("Branch #{branch_name} already exists, checking it out")
-      sh("git", "checkout", branch_name)
-      if branch_exists_remotely
-        sh("git", "pull", "origin", branch_name)
-      end
-    else
-      UI.message("Creating new branch #{branch_name}")
-      sh("git", "checkout", "-b", branch_name)
-    end
-  end
-
   private_lane :commit_and_push do |options|
     branch_name = options[:branch_name] || UI.user_error!("No branch_name provided")
     commit_message = options[:commit_message] || UI.user_error!("No commit_message provided")
@@ -849,41 +831,6 @@ DESC
     end
   end
 
-  private_lane :create_pr_if_necessary do |options|
-    branch_name = options[:branch_name] || UI.user_error!("No branch_name provided")
-    title = options[:title] || UI.user_error!("No title provided")
-    body = options[:body] || UI.user_error!("No body provided")
-    repo = options[:repo] || repo_name
-    repo_with_owner = "RevenueCat/#{repo}"
-    base = options[:base] || "main"
-    labels = options[:labels] || []
-    team_reviewers = options[:team_reviewers] || []
-
-    existing_pr = github_api(
-      api_token: ENV["GITHUB_TOKEN"],
-      path: "/repos/#{repo_with_owner}/pulls?head=RevenueCat:#{branch_name}&state=open"
-    )
-
-    if existing_pr[:json].length == 0
-      pr_url = create_pull_request(
-        repo: repo_with_owner,
-        title: title,
-        body: body,
-        base: base,
-        head: branch_name,
-        api_token: ENV["GITHUB_TOKEN"],
-        labels: labels,
-        team_reviewers: team_reviewers
-      )
-      
-      if pr_url.nil?
-        UI.user_error!("Failed to create pull request.")
-      end
-    else
-      UI.message("PR already exists.")
-    end
-  end
-
 
   private_lane :commit_push_and_create_pr_if_necessary do |options|
     branch_name = options[:branch_name] || UI.user_error!("No branch_name provided")
@@ -902,8 +849,10 @@ DESC
       )
 
       create_pr_if_necessary(
-        repo: repo,
-        branch_name: branch_name,
+        github_pr_token: ENV["GITHUB_TOKEN"],
+        repo_name: repo,
+        base_branch: "main",
+        head_branch: branch_name,
         title: title,
         body: body,
         labels: labels,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -12,6 +12,7 @@ import com.revenuecat.purchases.Store
 import com.revenuecat.purchases.VerificationResult
 import com.revenuecat.purchases.api.BuildConfig
 import com.revenuecat.purchases.common.diagnostics.DiagnosticsTracker
+import com.revenuecat.purchases.common.networking.ConnectionErrorReason
 import com.revenuecat.purchases.common.networking.ETagManager
 import com.revenuecat.purchases.common.networking.Endpoint
 import com.revenuecat.purchases.common.networking.HTTPRequest
@@ -168,6 +169,7 @@ internal class HTTPClient(
         val requestStartTime = dateProvider.now
         var callResult: HTTPResult? = null
         var requestResult: HTTPTimeoutManager.RequestResult = HTTPTimeoutManager.RequestResult.OTHER_RESULT
+        var exceptionHit: IOException? = null
 
         try {
             callResult = performCall(
@@ -185,6 +187,7 @@ internal class HTTPClient(
                 requestResult = HTTPTimeoutManager.RequestResult.SUCCESS_ON_MAIN_BACKEND
             }
         } catch (e: IOException) {
+            exceptionHit = e
             if (e is SocketTimeoutException && isMainBackend && canUseFallback()) {
                 requestResult = HTTPTimeoutManager.RequestResult.TIMEOUT_ON_MAIN_BACKEND_FOR_FALLBACK_SUPPORTED_ENDPOINT
                 callResult = performRequestToFallbackURL()
@@ -203,6 +206,7 @@ internal class HTTPClient(
                 callSuccessful,
                 callResult,
                 isRetry = refreshETag,
+                connectionException = exceptionHit,
             )
         }
         if (callResult == null) {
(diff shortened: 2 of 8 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -18,25 +20,30 @@ import kotlinx.serialization.ExperimentalSerializationApi
...
  * @property identityManager Manages user identity within the system.
  * @property eventsDispatcher Dispatches event-related operations.
  * @property postEvents Function for sending events to the backend.
  */
+@Suppress("LongParameterList")
 internal class EventsManager(
     private val appSessionID: UUID = Companion.appSessionID,
     private val legacyEventsFileHelper: EventsFileHelper<PaywallStoredEvent>,
     private val fileHelper: EventsFileHelper<BackendStoredEvent>,
+    private val adFileHelper: EventsFileHelper<BackendStoredEvent.Ad>,
```
tonidero: This PR is still WIP but this is the last approach I mentioned [here](https://github.com/RevenueCat/purchases-ios/pull/5640#discussion_r2429269778). I believe it minimizes the code duplication quite a bit. Wdyt @polmiro @JayShortway ?

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
```diff
@@ -0,0 +1,71 @@
+package com.revenuecat.purchases.ads.events
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.common.events.EventsManager
+
+@InternalRevenueCatAPI
+class AdTracker internal constructor(
```
tonidero, line 54: This is kinda the public API I was thinking. This would be called by users like:
```
Purchases.sharedInstance.adTracker.trackAdDisplayed(....)
```

I thought about having a single "event" object that users need to build, but feels like extra complexity and the API can just flatten it out. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
tonidero: Lmk what you think about this approach (different track methods for different types of events). I think we could also pass the URL as a parameter to the events manager and create multiple event managers. However, I don't think the current approach is that error-prone since the types of events for ads and current are different.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
```diff
@@ -90,9 +115,9 @@ internal class EventsManager(
         }
     }
 
-    @get:Synchronized
-    @set:Synchronized
-    private var flushInProgress = false
+    private var flushInProgress = AtomicBoolean(false)
```
tonidero, line 113: This felt safer 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
polmiro: I agree with having different storage files given they function as a queue that gets cleared independently.

When it comes to event managers I am a bit in the middle, let me share my thoughts see what you think:
- Tracking:
    - From a caller perspective all the event tracking calls should be going through the same object. And how that gets delivered is an implementation detail. This hints to me that we should have a single events manager when it comes to tracking.
- Flushing:
    - Paywall events flushing: There are a couple of trigger points that specific only to the paywall events. These in my opinion should only go to the event manager that handles paywall events and not the one handling ad events.
    - Ad events flushing: Given the amount of events, there may be added complexities to handle in the ad event manager (just throwing some rough potential ideas):
        - sending more one batch of events at a time.
        - having some sort of time-based flushing to avoid waiting too long
        - having some event count flushing to prevent accumulating too many events.
    - Maybe these apply to both event managers, maybe not, I am not sure.
    - These hint to me that we may want separate events manager when it comes to flushing.

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
polmiro: I like it, specially since it's an API that will likely not be the "go-to" recommended solution in the future and more of an alternative to only implement ad tracking data in RC.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/EventsManager.kt
tonidero: Hmm well, regarding flushing behavior, if we want to have different behavior between ad and paywalls&customerCenter events, having it be 2 instances of the EventManger might be a bit trickier... Right now the EventManger, syncs only the oldest 50 events for paywalls/customer center events. But maybe we want to perform multiple requests for Ads, if there are more than 50 events, so they don't accumulate, which would require changes inside the EventManager...

Having said that, I think it would be better to try to keep the logic for flushing the same/similar, since the events seem relatively similar in importance. At this point, I think I could go either way. Will implement the 2 instances of the EventManager though, since it might be slightly cleaner 👍

polmiro: I agree with the thoughts, this is why I wanted to separate this part in a different PR since it seems prone to require some more back and forth.

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
polmiro: For the record: I already ported this to `purchases-ios`.

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
```diff
@@ -0,0 +1,148 @@
...
+                else -> AdRevenuePrecision(value)
+            }
+        }
+    }
+}
+
+/**
+ * Tracks ad-related events such as ad displays, opens, and revenue.
+ */
+@InternalRevenueCatAPI
```
tonidero, line 53: Note that I'm marking this as `@InternalRevenueCatAPI`, so developers would need to explicitedly opt-in to use it, but it can still theoretically be used in the wild (very unlikely), unlike in iOS, where it's behind a compiler flag. I can add a compiler flag to disable tracking at all until we're ready, but I was wondering what the plan is to ship this? @polmiro

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
```diff
@@ -0,0 +1,148 @@
...
+class AdTracker internal constructor(
+    private val eventsManager: EventsManager?,
```
JayShortway, line 55: Why is this nullable? Does the `AdTracker` have any use without an `EventsManager`?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -420,6 +422,36 @@ internal class PurchasesFactory(
...
 
+    private fun createAdEventsManager(
+        context: Context,
+        identityManager: IdentityManager,
+        eventsDispatcher: Dispatcher,
+        backend: Backend,
+    ): EventsManager? {
+        // RevenueCatUI is Android 24+ so it should always enter here when using RevenueCatUI.
+        // Still, we check for Android N or newer since we use Streams which are 24+ and the main SDK supports
+        // older versions.
```
JayShortway: Where are we using `Stream`s? I remember refactoring to move away from streams as it requires core library desugaring. We can use Kotlin's `Sequence` instead.
- https://github.com/RevenueCat/purchases-android/pull/1943
- https://github.com/RevenueCat/purchases-android/pull/1944

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
tonidero: It would be nullable in Android < 24. This is how it worked for the existing paywalls/customer center events, and I wanted to minimize changes here. As you mentioned in the other comment, I think it's due to the previous usage of Streams.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: This is a leftover comment from that refactor, same as the fact that we don't support events in Android < 24. We should totally refactor this but I think it's better to split that into a different PR. One thing I can do is to remove some duplication in this method that creates the events manager itself in this PR.

tonidero: I did this small refactor to avoid the duplication, and we can work on removing the optionality of the events manager in a follow-up PR?

tonidero: Applied the changes to remove this in https://github.com/RevenueCat/purchases-android/pull/2854

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
tonidero: Applied the changes to remove this in https://github.com/RevenueCat/purchases-android/pull/2854

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
ajpallares: What there be the risks of someone using this? From the stability of public APIs perspective, I think this is OK, even if we want to make changes to it while marked as `@InternalRevenueCatAPI`

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1578,6 +1581,8 @@ internal class PurchasesOrchestrator(
     private fun flushPaywallEvents() {
         if (isAndroidNOrNewer()) {
             eventsManager?.flushEvents()
+            // WIP: Decide when/frequency to flush events
```
ajpallares: Just flagging this, in case it's something we know already. I'm OK with merging it like this for now as everything is marked as internal-only APIs

tonidero: I think in the end we decided it's ok so sync on app foreground + background + on purchases, the way it is now. cc @polmiro could you confirm? 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
tonidero: It's behind a feature flag in the backend as well, so it would basically have events that would fail to send, and since it's with a 4xx, it would mark them as sent, so I think very low right now.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
polmiro: Yes, that's right. I would leave it as a future improvement for now since we don't have a good hunch yet on what should it be.

tonidero: Will remove the comment for now, and we can revisit this later 👍

PR #2731, opened 2025-10-15
Fix default Customer Center UI by using `surfaceContainer` instead of `surface`

vegaro:
By default `surface` and `background` have the same color in Material3 themes. This was causing apps that don't wrap the Customer Center with a Material3 theme to have the wrong default colors. **This was actually happening in all hybrids since there's no way to wrap the Customer Center with Material3 theme**. Example:

image

By changing to use `surfaceContainer` we make sure the UI has some defaults that make sense. `surfaceContainer` adds some contrast on a background/surface: This would be the default in a hybrid SDK now for example:

image

I know this is technically a "breaking" change in the UI, and I've thought of alternatives to try to keep supporting `surface`. For example this is the one I considered the most:

- Change it to `surfaceContainer` unless it's the same as `background`. In that case we use `surface` if it's not the same as `background`. This way we make sure we don't break any UI. Worst case is we show a color that's different to what they were setting before, but still has some contrast with the `background`.

But doing a fallback mechanism like that needs to be documented since it's possible someone has unwanted behavior, for example someone might actually be setting `surfaceContainer` on purpose the same as `background` if they want a UI like that. 

Since this wasn't documented and because we want to make sure the docs are clear and we specify `surfaceContainer` is the color to override, I think we can consider this a bug. We've actually have many reports of people using Material2/XML with this issue, and all the hybrids were broken.

The final solution we need to do is accept a `CustomerCenterColors` option and make this configurable from the dashboard.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
@@ -108,7 +108,7 @@ private fun ContentUnavailableView(
     Surface(
         modifier = modifier,
         shape = RoundedCornerShape(CustomerCenterConstants.Card.ROUNDED_CORNER_SIZE),
-        color = MaterialTheme.colorScheme.surface,
+        color = MaterialTheme.colorScheme.surfaceContainer,
     ) {
         Column(
             modifier = Modifier
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrenciesListView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/VirtualCurrenciesListView.kt
@@ -152,7 +152,7 @@ internal fun VirtualCurrencyRow(
     Surface(
         modifier = modifier.fillMaxWidth(),
         shape = shape,
-        color = MaterialTheme.colorScheme.surface,
+        color = MaterialTheme.colorScheme.surfaceContainer,
     ) {
         Row(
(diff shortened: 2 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
```diff
@@ -108,7 +108,7 @@ private fun ContentUnavailableView(
     Surface(
         modifier = modifier,
         shape = RoundedCornerShape(CustomerCenterConstants.Card.ROUNDED_CORNER_SIZE),
-        color = MaterialTheme.colorScheme.surface,
+        color = MaterialTheme.colorScheme.surfaceContainer,
```
tonidero, line 111: I guess this might cause a change in looks for people that already were using a theme and overwriting this color right? Not sure how to avoid that behavior change though... Maybe it's worth passing an updated theme in the hybrids, since it's where this is mostly broken? I guess since this is relatively broken in Android as well, it might be ok to ship it, but I think we should add a warning in the release notes as well..

facumenzella: We've discussed this, and sadly this is more a bug than a breaking change 😢 . A warning in the release notes would be nice for sure

facumenzella (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/NoActiveUserManagementView.kt
vegaro: Just implemented an alternative approach in https://github.com/RevenueCat/purchases-android/pull/2741

PR #2732, opened 2025-10-15
Fixes more Icons crashing with `NoClassDefFoundError` when using Material 1.4.0

vegaro:
#2727 fixed some Icons, but I somehow missed some other `Icons` being used in other parts of the code. This PR adds the missing ones:

```
java.lang.NoClassDefFoundError: Failed resolution of: Landroidx/compose/material/icons/Icons$Rounded;
```

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/Info.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/Info.kt
@@ -1,0 +1,41 @@
+@file:Suppress("MagicNumber", "ObjectPropertyName")
+
+package com.revenuecat.purchases.ui.revenuecatui.icons
+
+import androidx.compose.material.icons.materialIcon
+import androidx.compose.material.icons.materialPath
+import androidx.compose.ui.graphics.vector.ImageVector
+
+internal val Info: ImageVector
+    get() {
+        if (_info != null) {
+            return _info!!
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/Info.kt
```diff
@@ -0,0 +1,41 @@
+@file:Suppress("MagicNumber", "ObjectPropertyName")
+
+package com.revenuecat.purchases.ui.revenuecatui.icons
+
+import androidx.compose.material.icons.materialIcon
+import androidx.compose.material.icons.materialPath
+import androidx.compose.ui.graphics.vector.ImageVector
+
+internal val Info: ImageVector
```
tonidero, line 9: Hmm I wonder if we should wrap all these icons into a single namespace like Google's `Icons`... It feels weird to have this as a top level constant.

tonidero (approved):
Just a possible cleanup... but not a blocker. Looks good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/icons/Info.kt
vegaro: I agree, while doing this I also noticed it would be good to know by the name if it's a filled button or not. I can do that as a followup

PR #2746, opened 2025-10-16
Add extra message test store dialog

tonidero:
### Description
This brings some extra context message to the purchase dialog for the test store, copying what we do in iOS
image


rickvdl (approved):
Nice, looks good!

PR #2747, opened 2025-10-17
Add webCheckoutURL property to Offering and Package

tonidero:
### Description
This adds a new API for both `Offering` and `Package`, the `webCheckoutURL`. This property will have the Web Purchase Link associated with that Offering and/or Package, so it can be opened if needed. This will be needed to support Web Purchase Buttons in Paywalls, coming in a follow-up PR.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsFactoryTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsFactoryTest.kt
@@ -486,6 +486,34 @@ class OfferingsFactoryTest {
     }
 
     @Test
+    fun `createOfferings with invalid URL in WPL`() {
+        val productIds = listOf(productId)
+        mockStoreProduct(productIds, emptyList(), ProductType.SUBS)
+        mockStoreProduct(productIds, productIds, ProductType.INAPP)
+
+        val invalidUrlWPL = oneOfferingWithWPL.apply {
+            val offering = getJSONArray("offerings").getJSONObject(0)
+            offering.put("web_checkout_url", "ht!tp:/invalid-url")
+            val pkg = offering.getJSONArray("packages").getJSONObject(0)
+            pkg.put("web_checkout_url", "ht!tp:/invalid-url")
+        }
+
+        var offerings: Offerings? = null
+        offeringsFactory.createOfferings(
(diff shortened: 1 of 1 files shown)


purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsFactoryTest.kt
```diff
@@ -29,138 +29,200 @@ import org.junit.Before
...
+                    "identifier": "$STUB_OFFERING_IDENTIFIER",
+                    "description": "This is the base offering",
+                    "packages": [
+                        {
+                            "identifier": "${'$'}rc_monthly",
+                            "platform_product_identifier": "$STUB_PRODUCT_IDENTIFIER",
+                            "web_checkout_url": "http://revenuecat.com?package_id=${'$'}rc_monthly"
+                        }
+                    ],
+                    "web_checkout_url": "http://revenuecat.com"
```
rickvdl, line 216: What happens when the URL format is invalid? Will just the url be null or will the parsing fail completely

tonidero: We will log and fallback to null, but I indeed didn't add a test, good call! Will do that, thanks!

api-tester/src/main/java/com/revenuecat/apitester/kotlin/PackageAPI.kt
```diff
@@ -4,6 +4,7 @@ import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.PackageType
 import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.models.StoreProduct
+import java.net.URL
```
joshdholtz, line 7: This is a `java.net` package?! I don't know if I ever realized that 😅

joshdholtz (commented):
Looks good to me! But will save "Approve" for somebody else 😇

api-tester/src/main/java/com/revenuecat/apitester/kotlin/PackageAPI.kt
tonidero: Hahah there is also an Android `Uri`... but prefer to use this one when I can so it's not dependent of Android yeah...

PR #2749, opened 2025-10-17
Add PurchaseButtonComponent WPL field deserialization

tonidero:
### Description
This just adds the new fields we will need to support WPL in paywalls, but they are not yet in use as of this PR.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal
-  revision: 25c7fb83d3e6f590a7130547f75135abda9f86a2
+  revision: 525d48c7adac8d3d958abe3502db8fb78ff54ceb
   branch: main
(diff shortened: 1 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PurchaseButtonComponent.kt
```diff
@@ -13,4 +17,84 @@ import kotlinx.serialization.Serializable
 @Immutable
 class PurchaseButtonComponent(
     @get:JvmSynthetic val stack: StackComponent,
-) : PaywallComponent
+    @get:JvmSynthetic val action: Action? = null,
```
tonidero, line 20: Action is pretty much deprecated, but still supported in iOS. We basically transform the action into method if method is not available and action is. In order to avoid any weird scenario, we are keeping this backwards compatibility in android.

PR #2750, opened 2025-10-17
Keep cache logic similar to iOS + ensure path extension is on file

JZDesign:
This mainly is to keep the logic as similar as possible, but if for some reason there are specific versions of android that fail like iOS did when the file extension was not on the file, this will fix that.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
@@ -209,12 +209,11 @@ internal class DefaultFileCache(
 
     override fun generateLocalFilesystemURI(remoteURL: URL, checksum: Checksum?): URI? {
         val urlHash = md5Hex(remoteURL.toString().toByteArray())
-        val fileName = File(urlHash).name
+        // Use checksum value as part of the file name (like iOS)
+        val fileName = File(urlHash).name + (checksum?.value ?: "")
         if (fileName.isEmpty()) return null
 
-        // Use checksum value as extension (like iOS), fallback to URL extension
-        val extension = (checksum?.value ?: "") + remoteURL
-            .path.substringAfterLast('.', "")
+        val extension = remoteURL.path.substringAfterLast('.', "")
         val fileWithExtension = "$fileName.$extension"
 
         return File(cacheDir, fileWithExtension).toURI()
--- a/purchases/src/test/java/com/revenuecat/purchases/storage/DefaultFileCacheTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/storage/DefaultFileCacheTest.kt
@@ -268,13 +268,13 @@ class DefaultFileCacheTest {
     }
 
     @Test
-    fun `generateLocalFilesystemURI uses checksum as extension`() {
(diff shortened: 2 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -213,8 +213,9 @@ internal class DefaultFileCache(
         if (fileName.isEmpty()) return null
 
         // Use checksum value as extension (like iOS), fallback to URL extension
-        val extension = checksum?.value ?: remoteURL.path.substringAfterLast('.', "")
-        val fileWithExtension = if (extension.isNotEmpty()) "$fileName.$extension" else fileName
+        val extension = (checksum?.value ?: "") + remoteURL
+            .path.substringAfterLast('.', "")
```
JayShortway: Doesn't this mean we get 2 "extensions" if we have a checksum and an extension in the URL?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -213,8 +213,9 @@ internal class DefaultFileCache(
...
+        val fileWithExtension = "$fileName.$extension"
```
JayShortway: What happens with files that were saved using the old extension scheme? Are they still being used or will they be downloaded again? Will they get cleaned up?

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: Wow, this is not at all what I meant to write. I needed that weekend apparently.

JZDesign: I meant to append the checksum to the filename, then make the extension just the plain old .mp4 or .mov

But I was not seeing clearly. I'll fix that

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: In this case it would redownload the file and the old file will be cleaned up by the OS when it deems appropriate (since it will live in the cache's directory). The video component will get turned on tomorrow I think, so there will be some of this logic in the wild, but if this merges and is released the next day, it should be quite minimal.

If we think it's necessary to proactively clean up the cache directory, I can add that in there.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
JZDesign: Now, it's just the one extension. Thanks for asking that question btw.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
tonidero: Hmm well, the OS will only clear these files if it runs out of disk storage, which won't happen in most cases... I think it might be a good idea to create some sort of fallback system and/or cleanup old files to avoid duplicating these files? Not sure if a blocker though...

PR #2751, opened 2025-10-20
Add support for Web Purchase buttons in paywalls

tonidero:
### Description
Adds support for Web Purchase buttons in paywalls, both ones using Web Billing and using Custom URL checkouts.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -182,36 +182,28 @@ internal class PaywallViewModelImpl(
             Logger.e("Web checkout URL can only be constructed for loaded Components paywalls")
             return null
         }
-        val packageParamBehavior = launchWebCheckout.packageParamBehavior
-        val packageToUse = when (packageParamBehavior) {
-            is PaywallAction.External.LaunchWebCheckout.PackageParamBehavior.DoNotAppend -> null
+        val behavior = launchWebCheckout.packageParamBehavior
+        val (packageToUse, packageParam) = when (behavior) {
             is PaywallAction.External.LaunchWebCheckout.PackageParamBehavior.Append ->
-                packageParamBehavior.rcPackage ?: state.selectedPackageInfo?.rcPackage
+                (behavior.rcPackage ?: state.selectedPackageInfo?.rcPackage) to behavior.packageParam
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -167,6 +174,46 @@ internal class PaywallViewModelImpl(
...
+                    } else {
+                        url
+                    }
+                }
+            }
+            finalUrl.toString()
+        } else {
+            packageToUse?.webCheckoutURL?.toString() ?: state.offering.webCheckoutURL.toString()
+        }
+    }
```
ajpallares, line 207: I feel we could avoid repeating the `when (packageParamBehavior)`.

Proposal (FFTI)

```suggestion
    override fun getWebCheckoutUrl(launchWebCheckout: PaywallAction.External.LaunchWebCheckout): String? {
        val customUrl = launchWebCheckout.customUrl
        val state = state.value as? PaywallState.Loaded.Components
        if (state == null) {
            Logger.e("Web checkout URL can only be constructed for loaded Components paywalls")
            return null
        }

        val behavior = launchWebCheckout.packageParamBehavior
        val (packageToUse, packageParam) = when (behavior) {
            is PaywallAction.External.LaunchWebCheckout.PackageParamBehavior.Append ->
                (behavior.rcPackage ?: state.selectedPackageInfo?.rcPackage) to behavior.packageParam
            is PaywallAction.External.LaunchWebCheckout.PackageParamBehavior.DoNotAppend ->
                null to null
        }

        if (customUrl != null) {
            val url = try {
                URL(customUrl)
            } catch (e: MalformedURLException) {
                Logger.e("Invalid custom URL: $customUrl", e)
                return null
            }

            val finalUrl = if (packageParam != null && packageToUse != null) {
                url.appendQueryParameter(packageParam, packageToUse.identifier)
            } else url

            return finalUrl.toString()
        }

        return packageToUse?.webCheckoutURL?.toString() ?: state.offering.webCheckoutURL.toString()
    }
```

tonidero: Oh nice! I like it, thanks for the suggestion!

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2751/commits/9013fa830f787e65ca8884f2de1103268e7235e9

PR #2752, opened 2025-10-21
Add `underlyingErrorMessage` to `PurchasesException` message

tonidero:
### Description
This PR adds the `underlyingErrorMessage` to the message of the `PurchasesException`. This can be specially useful in hybrids, where access to the underlying error message is not as straightforward, since it's added as part of the info map. This will however also affect particularly users using coroutines.

This is an attempt to improve the situation where the app crashes if a developer uses the test store API key in production builds. This error message is not at all easy to parse in the hybrids since it's not included in the exception message.

I've found in many occasions that the split between the `message` and `underlyingErrorMessage` has been a bit confusing... The main problem of this approach is that exceptions may be split in crash reporting tools if they use the message as a way to split those exceptions... But I do think it improves usability of the errors, so would appreciate some feedback.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/PurchasesExceptionTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/PurchasesExceptionTest.kt
@@ -46,6 +46,6 @@ class PurchasesExceptionTest {
 
         val purchasesException = PurchasesException(purchasesError)
 
-        assertThat(purchasesError.message).isEqualTo("${purchasesException.message} Underlying error: $underlyingErrorMessage")
+        assertThat(purchasesException.message).isEqualTo("${purchasesError.message} Underlying error: $underlyingErrorMessage")
     }
 }
--- a/purchases/src/test/java/com/revenuecat/purchases/PurchasesTransactionExceptionTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/PurchasesTransactionExceptionTest.kt
@@ -46,7 +46,7 @@ class PurchasesTransactionExceptionTest {
 
         val purchasesException = PurchasesTransactionException(purchasesError, userCancelled = false)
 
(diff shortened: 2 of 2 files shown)


ajpallares (commented):
Yeah, I have nothing to add to your comments @tonidero. This makes the `message` much more useful by itself, not requiring going through the underlying error message to know what's going on. I see the point of the tools that use the error message to classify errors; this would break that. 

In my opinion, errors should be classified by code instead of by message, so maybe this is a way to enforce that implementation.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesException.kt
```diff
@@ -9,5 +9,9 @@ open class PurchasesException(val error: PurchasesError) : Exception() {
         get() = error.underlyingErrorMessage
 
     override val message: String
-        get() = error.message
+        get() = error.message + if (error.underlyingErrorMessage != null) {
+            " Underlying error: ${error.underlyingErrorMessage}"
+        } else {
+            ""
+        }
```
vegaro, line 16: hmm isn't the `underlyingErrorMessage` most of the time something not that user friendly? If someone is displaying just the message, this change will show the whole thing. I beleive for request errors, the underlying error message is particularly dev-focused.

For example the other day building an error view for the Customer Center, in my first iteration I added the underlying message in an initial iteration, but then I noticed it was very developer focused and decided to remove it from the UI. I kept the message though ("Error performing request"). I believe the underlying error message said something about can't decode "<html" tag but I don't recall exactly.

tonidero: I would argue that the `message` isn't user friendly either... But having said that, the underlying error message is indeed much more technical and I do think some devs might be using it to display user facing messages, and this would break them...

So I'm thinking to move this so we do it in the RN layer for exceptions when configuring the SDK, which is the case we're trying to improve right now. We can iterate on this if we detect it becomes a problem in other platforms.

PR #2753, opened 2025-10-21
Add hook to delay payment screens

JZDesign:
### Checklist

### Motivation

Some of our customers have requested a way to intercept the on purchase flow before the purchase happens so they can ensure other critical steps are done first like authentication, or age verification.

### Description

Creates the ability to gate and intercept when a purchase / redeem button is pressed. This should allow the consumer to log or intercept the flow and present an auth flow for example.

### DIFF
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallListenerAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallListenerAPI.java
@@ -14,7 +14,7 @@ final class PaywallListenerAPI {
     void check() {
         PaywallListener listener = new PaywallListener() {
             @Override
-            public void onPurchasePackageInitiated(@NonNull String packageId, @NonNull Resumable resume) {}
+            public void onPurchasePackageInitiated(@NonNull Package rcPackage, @NonNull Resumable resume) {}
 
             @Override
             public void onPurchaseStarted(@NonNull Package rcPackage) {}
--- a/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallListenerAPI.kt
+++ b/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallListenerAPI.kt
@@ -11,7 +11,7 @@ import com.revenuecat.purchases.ui.revenuecatui.utils.Resumable
 private class PaywallListenerAPI {
     fun check() {
         val listener = object : PaywallListener {
-            override fun onPurchasePackageInitiated(packageId: String, resume: Resumable) {}
+            override fun onPurchasePackageInitiated(rcPackage: Package, resume: Resumable) {}
 
             override fun onPurchaseStarted(rcPackage: Package) {}
 
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallViewActivity.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallViewActivity.kt
@@ -1,6 +1,5 @@
 package com.revenuecat.paywallstester
 
-import android.content.DialogInterface
 import android.os.Bundle
 import android.util.Log
 import android.view.WindowManager
@@ -29,7 +28,7 @@ class PaywallViewActivity : AppCompatActivity() {
         setContentView(binding.root)
         binding.paywallView.setOfferingId(offeringId)
         binding.paywallView.setPaywallListener(object : PaywallListener {
-            override fun onPurchasePackageInitiated(packageId: String, resume: Resumable) {
+            override fun onPurchasePackageInitiated(rcPackage: Package, resume: Resumable) {
                 AlertDialog
                     .Builder(this@PaywallViewActivity)
                     .setTitle("On Purchase Initiated Hook")
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/paywall/PaywallScreenViewModel.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/paywall/PaywallScreenViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.AndroidViewModel
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.viewModelScope
 import com.revenuecat.purchases.CustomerInfo
+import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.Purchases
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesException
@@ -12,13 +13,11 @@ import com.revenuecat.purchases.awaitOfferings
 import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
 import com.revenuecat.purchases.ui.revenuecatui.utils.Resumable
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 
 interface PaywallScreenViewModel : PaywallListener {
     companion object {
@@ -49,7 +48,7 @@ class PaywallScreenViewModelImpl(
     }
 
     @Suppress("MagicNumber")
-    override fun onPurchasePackageInitiated(packageId: String, resume: Resumable) {
+    override fun onPurchasePackageInitiated(rcPackage: Package, resume: Resumable) {
         resume()
     }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
@@ -11,11 +11,11 @@ interface PaywallListener {
      * Called when a package purchase is about to be initiated, before the payment sheet is displayed.
      * This allows the app to perform any necessary preparation (e.g., authentication) before proceeding.
      *
-     * @param packageId The identifier of the package being purchased.
+     * @param rcPackage: The Package being purchased.
      * @param resume A callback that must be invoked to continue with the purchase flow.
      *               If not called, the purchase flow will not proceed.
      */
-    fun onPurchasePackageInitiated(packageId: String, resume: Resumable) {
+    fun onPurchasePackageInitiated(rcPackage: Package, resume: Resumable) {
         // Default implementation immediately resumes
         resume()
     }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -349,12 +349,13 @@ internal class PaywallViewModelImpl(
         // Call onPurchasePackageInitiated and wait for resume() to be called
 
         val shouldResume = suspendCoroutine { continuation ->
-            listener?.onPurchasePackageInitiated(packageToPurchase.identifier) { shouldResume ->
+            listener?.onPurchasePackageInitiated(packageToPurchase) { shouldResume ->
                 continuation.resume(shouldResume)
             } ?: continuation.resume(true)
         }
 
         if (!shouldResume) {
+            Logger.d("Purchase cancelled listener.onPurchasePackageInitiated returned false")
             return
         }
 
--- a/ui/revenuecatui/api.txt
+++ b/ui/revenuecatui/api.txt
@@ -52,7 +52,7 @@ package com.revenuecat.purchases.ui.revenuecatui {
     method public default void onPurchaseCancelled();
     method public default void onPurchaseCompleted(com.revenuecat.purchases.CustomerInfo customerInfo, com.revenuecat.purchases.models.StoreTransaction storeTransaction);
     method public default void onPurchaseError(com.revenuecat.purchases.PurchasesError error);
-    method public default void onPurchasePackageInitiated(String packageId, com.revenuecat.purchases.ui.revenuecatui.utils.Resumable resume);
+    method public default void onPurchasePackageInitiated(com.revenuecat.purchases.Package rcPackage, com.revenuecat.purchases.ui.revenuecatui.utils.Resumable resume);
     method public default void onPurchaseStarted(com.revenuecat.purchases.Package rcPackage);
     method public default void onRestoreCompleted(com.revenuecat.purchases.CustomerInfo customerInfo);
     method public default void onRestoreError(com.revenuecat.purchases.PurchasesError error);
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
@@ -69,8 +69,8 @@ class PaywallView : CompatComposeView {
     private var listener: PaywallListener? = null
     private var shouldDisplayDismissButton: Boolean? = null
     private var internalListener: PaywallListener = object : PaywallListener {
-        override fun onPurchasePackageInitiated(packageId: String, resume: Resumable) {
-            listener?.onPurchasePackageInitiated(packageId, resume) ?: resume()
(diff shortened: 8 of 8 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -199,6 +201,7 @@ internal class PaywallViewModelImpl(
         if (verifyNoActionInProgressOrStartAction()) {
             return
         }
+
```
JZDesign:
```suggestion
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
```diff
@@ -6,6 +6,19 @@ import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.models.StoreTransaction
 
 interface PaywallListener {
+    /**
+     * Called when a package purchase is about to be initiated, before the payment sheet is displayed.
+     * This allows the app to perform any necessary preparation (e.g., authentication) before proceeding.
+     *
+     * @param packageId The identifier of the package being purchased.
+     * @param resume A callback that must be invoked to continue with the purchase flow.
+     *               If not called, the purchase flow will not proceed.
```
tonidero, line 16: Hmm I think with the current implementation, the coroutine waiting for the callback to be called could remain waiting forever? 

Maybe we should make calling the resume function mandatory to be called, and return a parameter of whether to resume or cancel the purchase... Still there are high chances of devs forgetting to call it I would say... not sure if there is a better way though...

JZDesign: Since this is opt in, and it's used for this one purpose, I think it's a lower risk. But we could add logs in a timer or something that would alert them that this wasn't invoked yet or something like that

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -344,6 +346,13 @@ internal class PaywallViewModelImpl(
 
     @Suppress("LongMethod", "NestedBlockDepth")
     private suspend fun performPurchase(activity: Activity, packageToPurchase: Package) {
+        // Call onPurchasePackageInitiated and wait for resume() to be called
+        suspendCancellableCoroutine { continuation ->
+            listener?.onPurchasePackageInitiated(packageToPurchase.identifier) {
+                continuation.resume(Unit)
+            } ?: continuation.resume(Unit)
+        }
```
JayShortway: Since we're not invoking `continuation.invokeOnCancellation { }`, I think we can just use `suspendCoroutine { }`:

```suggestion
        suspendCoroutine { continuation ->
            listener?.onPurchasePackageInitiated(packageToPurchase.identifier) {
                continuation.resume(Unit)
            } ?: continuation.resume(Unit)
        }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
```diff
@@ -6,6 +6,19 @@ import com.revenuecat.purchases.PurchasesError
...
+     */
+    fun onPurchasePackageInitiated(packageId: String, resume: () -> Unit) {
```
JayShortway: Would be good to add API tests for this, in Java too. I think that lambda parameter might be cumbersome in Java unfortunately. We might solve that with a `fun interface`. That feels like a lambda from the Kotlin side and makes it bearable from Java.

JZDesign: I think I got this where it should be. Take another look and let me know if there is something I'm missing.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
tonidero: Relatively yeah... but I wonder if we should make it so the callback receives a boolean parameter indicating whether it should proceed with the purchase, or cancel it... So developers that want to cancel the purchase attempt, still call the `resume` callback, but with a `false` parameter. This way, we won't get this coroutine waiting forever in this situation (unless on a bad implementation by the developer)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
```diff
@@ -4,8 +4,22 @@ import com.revenuecat.purchases.CustomerInfo
...
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.models.StoreTransaction
+import com.revenuecat.purchases.ui.revenuecatui.utils.Invokable
 
 interface PaywallListener {
+    /**
+     * Called when a package purchase is about to be initiated, before the payment sheet is displayed.
+     * This allows the app to perform any necessary preparation (e.g., authentication) before proceeding.
+     *
+     * @param packageId The identifier of the package being purchased.
```
tonidero: I've been thinking if there is anything else we would like to give the developer... Would it make more sense to return the entire `Package`? I think it would be much more useful than just returning the ID

JayShortway: Good point. Maybe we should provide an object, like `SomethingSomethingParams`, which would allow us to evolve this without breaking the API.

JZDesign: What else do you think would be useful?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Invokable.kt
```diff
@@ -0,0 +1,8 @@
+package com.revenuecat.purchases.ui.revenuecatui.utils
+
+fun interface Invokable {
+    fun performAction()
+
+    operator fun invoke() = performAction()
+}
```
JayShortway: I was more thinking of something like: 

```suggestion
fun interface Resumable {
    fun resume()
}
```

I don't think we need the `operator fun`, but maybe I'm missing something.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
JZDesign: Good point. I will change this up to allow for canceling.

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallListenerAPI.kt
JayShortway, line 1: I think we should add a Java one to `api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui` too, if only to get a feel for what the API is like for Java consumers.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
tonidero: I think the `Package` would have most important information... Other things could be things like `context`, including locale, darkMode, or things like this... But well, I don't think it's needed for V1. Making the API extensible would be great though indeed.

JayShortway: Yea I would only add now what we _know_ people want, and add more later if we get feedback.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
JayShortway: Thanks! Added a comment [here](https://github.com/RevenueCat/purchases-android/pull/2753#discussion_r2490605078).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Invokable.kt
JZDesign: It just makes it easy for the consumer to call the callback when ready

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
JZDesign: > Other things could be things like context, including locale, darkMode, or things like this

These are all things they can gather without us sending it back right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Invokable.kt
JZDesign: Really, it's the call site here that I care about :


```kotlin

…, resume: { package, resume -> 
    someLongTask()
    resume()
}

```


now, for the users who actually want to cancel it would look like this;


```kotlin

…, resume: { package, complete ->
   try {
       performGating()
       complete()
   } catch (e: Throwable) {
       complete(shouldResume = false)
   }
}
```

JZDesign: as opposed to:

```kotlin
…, resume: { package, completionHandler ->
    completionHandler.resume()
}
```

JZDesign: Also, I had renamed to Resumable yesterday before I saw this comment, so I'm glad we're thinking the same way.

api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallListenerAPI.java
```diff
@@ -0,0 +1,41 @@
...
+import com.revenuecat.purchases.models.StoreTransaction;
+import com.revenuecat.purchases.ui.revenuecatui.PaywallListener;
+import com.revenuecat.purchases.ui.revenuecatui.utils.Resumable;
+
+@SuppressWarnings({"unused", "FieldCanBeLocal"})
+final class PaywallListenerAPI {
+    void check() {
+        PaywallListener listener = new PaywallListener() {
+            @Override
+            public void onPurchasePackageInitiated(@NonNull String packageId, @NonNull Resumable resume) {}
```
JayShortway: This is nice!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
```diff
@@ -343,7 +343,7 @@ private fun DisplayOfferingMenu(
             onClick = { activity.launchPaywall(offering, edgeToEdge = true) },
         )
         DropdownMenuItem(
-            text = { Text(text = "Display paywall as view in an activity") },
+            text = { Text(text = "Display paywall as view in an activity (Purchase button gating example)") },
```
JayShortway, line 346: Is this intended?

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/paywall/PaywallScreenViewModel.kt
```diff
@@ -45,6 +48,14 @@ class PaywallScreenViewModelImpl(
         updateOffering()
     }
 
+    @Suppress("MagicNumber")
+    override fun onPurchasePackageInitiated(packageId: String, resume: Resumable) {
+        runBlocking {
+            delay(2_000L)
+            resume(true)
+        }
```
JayShortway: We should not be doing this imo, not even in a sample app. This blocks the main thread if we're not careful, essentially freezing the app for 2 seconds. 

This also makes me think: what's the ideal use case for this new API and what does it look like? As in, what kind of code do we think people will call before they call `resume()`?

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/PaywallViewActivity.kt
```diff
@@ -26,6 +29,32 @@ class PaywallViewActivity : AppCompatActivity() {
...
+                builder.setNegativeButton(
+                    "Cancel",
+                ) { dialog, which ->
+                    dialog.dismiss() // Dismiss the dialog
+                    resume(false)
+                }
+                builder.setCancelable(false) // Prevents dialog from being dismissed by tapping outside or back button
+
+                val dialog: AlertDialog = builder.create()
+                dialog.show()
```
JayShortway: Minor, but you can chain calls to the builder, and you can call `show()` directly on the builder without calling `create()`:

```kotlin
AlertDialog.Builder()
  .setTitle("")
  .setMessage("")
  .show()
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Invokable.kt
JayShortway: Yea that's pretty nice actually! I didn't know `fun interface` allowed for 2 functions if one of them is an `operator fun`, TIL!

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
JZDesign: Yes, so we can test it and know which one it was

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/paywall/PaywallScreenViewModel.kt
JZDesign: 😬 

That is a relic from me testing things and I accidentally committed it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/Invokable.kt
JZDesign: fun interfaces are "SAM" interfaces Single Abstract Method interfaces. They can have lots, but only one that needs to be implemented by a conforming type.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/paywall/PaywallScreenViewModel.kt
JZDesign: I think the best answer to the question of what do we think people will do before they call resume is more like the dialogue example I put in the PaywallActivity. Perform other actions and present other UI to perform Age Gating or Authentication before allowing them to pay.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -344,6 +346,18 @@ internal class PaywallViewModelImpl(
...
+        // Call onPurchasePackageInitiated and wait for resume() to be called
+
+        val shouldResume = suspendCoroutine { continuation ->
+            listener?.onPurchasePackageInitiated(packageToPurchase.identifier) { shouldResume ->
+                continuation.resume(shouldResume)
+            } ?: continuation.resume(true)
+        }
+
+        if (!shouldResume) {
+            return
```
tonidero, line 359: Maybe worth adding a debug log here saying the purchase will be cancelled?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallListener.kt
tonidero: Right, I think it's probably ok for now to just add the `Package` as a parameter, but I would do that now, since it could be a breaking change later otherwise (or add overloads which is not great). Possibly we could create a new type like `PurchasePackageInitiatedParams` or something like that, which would be much easier to extend in the future, that for now can just contain the `Package`.

PR #3, opened 2018-05-01
[FINAL] Deliver Feature Complete Build

jeiting:

PR #2754, opened 2025-10-21
Use the fallback endpoint in more failure scenarios

JayShortway:
## Description
Adds 2 tests:
- Asserting we use the fallback URLs when the backend responds with a non-JSON body.
- Asserting we use the fallback URLs when we fail to connect to the backend.

and fixes them.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
@@ -957,11 +957,8 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
             )
         }
     }
-
-    // endregion Fallback API host
 }
 
-// region Non-JSON response with fallback URLs
 @RunWith(ParameterizedRobolectricTestRunner::class)
 internal class ParameterizedNonJsonResponseBodyTest(
     private val endpoint: Endpoint,
@@ -1054,9 +1051,6 @@ internal class ParameterizedNonJsonResponseBodyTest(
     }
 }
 
-// endregion Non-JSON response with fallback URLs
-
-// region Connection failure with fallback URLs
 @RunWith(ParameterizedRobolectricTestRunner::class)
 internal class ParameterizedConnectionFailureFallbackTest(
     private val endpoint: Endpoint,
@@ -1128,5 +1122,3 @@ internal class ParameterizedConnectionFailureFallbackTest(
         }
     }
 }
-
-// endregion Connection failure with fallback URLs


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -50,7 +52,17 @@ internal data class HTTPResult(
...
+        .takeIf { it.isNotBlank() }
+        ?.let {
+            try {
+                JSONObject(it)
+            } catch (e: JSONException) {
+                errorLog(throwable = e) { "Failed to parse payload as JSON: $it" }
+                null
+            }
+        }
+        ?: JSONObject()
```
JayShortway, line 65: This means `HTTPResult` no longer assumes every response is JSON, which is more true to reality.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -127,6 +147,9 @@ internal class HTTPClient(
         try {
             callResult = performCall(baseURL, endpoint, body, postFieldsToSign, requestHeaders, refreshETag)
             callSuccessful = true
+        } catch (e: IOException) {
+            // Handle connection failures with fallback URLs
+            if (canUseFallback()) callResult = performRequestToFallbackURL() else throw e
```
JayShortway, line 152: This handles the connection error case. In this case, we do not retry with a fresh ETag, but go straight to the fallback URL. I think that makes sense, but curious to hear thoughts.

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
```diff
@@ -86,23 +86,6 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
...
-            endpoint,
-            expectedResult = HTTPResult.createResult(payload = "not uh jason")
-        )
-
-        try {
-            client.performRequest(baseURL, endpoint, body = null, postFieldsToSign = null, mapOf("" to ""))
-        } finally {
-            server.takeRequest()
-        }
-    }
```
JayShortway, line 104: I think this is no longer valid, as `HTTPResult` no longer throws `JSONException`, but I'm not quite sure what behavior this was testing. It was added by [the man himself in 2018](https://github.com/RevenueCat/purchases-android/pull/3/commits/061a503f4c86cbe0a6eada86a803c60fc0945ee3#diff-30a3571e93a10afe5fce6be86dacedf4cc4c1db3e11640b2eb790f3a0c628de2R60), so I hope it doesn't come back to haunt me.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: I do think that makes sense... I'm honestly not sure if ETags are supported in fortress/fallback urls... I guess if they were, we might be changing the ETag once, and once we go back to the main endpoint, we would have lost it and then refetched the whole response... But I believe that fortress/fallback urls are unlikely to implement etags + this would just mean one extra body in this very unlikely scenario, so I think this is fine yeah.

JayShortway: Yea that's a good point as well. I was also thinking that if the initial request fails because of a connection problem, it's better to go straight to the fallback URL and not keep hitting the main endpoint.

PR #2762, opened 2025-10-22
Add more proper error message when using test store API key in release builds

tonidero:
### Description
We were using the ConfigurationError code message and passing the details in the underlying error message. However, this made it so the proper error isn't that accessible, specially for hybrids developers. This modifies the exception message to be more precise instead of leaving the generic one. 

This is a different approach than the one in #2752

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesException.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesException.kt
@@ -1,6 +1,11 @@
 package com.revenuecat.purchases
 
-open class PurchasesException(val error: PurchasesError) : Exception() {
+open class PurchasesException internal constructor(
+    val error: PurchasesError,
+    internal val overridenMessage: String? = null,
+) : Exception() {
+
+    constructor(error: PurchasesError) : this(error, null)
 
     val code: PurchasesErrorCode
         get() = error.code
@@ -9,5 +14,5 @@ open class PurchasesException(val error: PurchasesError) : Exception() {
         get() = error.underlyingErrorMessage
 
     override val message: String
-        get() = error.message
+        get() = overridenMessage ?: error.message
 }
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -433,11 +433,12 @@ internal class PurchasesFactory(
...
                 throw PurchasesException(
-                    PurchasesError(
+                    error = PurchasesError(
                         code = PurchasesErrorCode.ConfigurationError,
-                        underlyingErrorMessage = "Please configure the Play Store/Amazon store app on the " +
-                            "RevenueCat dashboard and use its corresponding API key before releasing.",
                     ),
+                    overridenMessage = "Please configure the Play Store/Amazon store app on the " +
+                        "RevenueCat dashboard and use its corresponding API key before releasing. " +
+                        "Test Store is not supported in production builds.",
```
tonidero, line 441: The other alternative I thought about was adding a new code for this error... But I would like to avoid adding new error codes unless needed I guess, at least until we give it a proper pass to our error handling APIs.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesException.kt
```diff
@@ -9,5 +14,5 @@ open class PurchasesException(val error: PurchasesError) : Exception() {
         get() = error.underlyingErrorMessage
 
     override val message: String
-        get() = error.message
+        get() = overridenMessage ?: error.message
```
tonidero, line 17: By default we're always using the error code message. This will allow to overwrite that message in specific situations like the one in this PR.

vegaro: I think this makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
vegaro: I like it. This help with not having to create a lot of custom error codes for everything

PR #2763, opened 2025-10-22
[CAT-2067 EXPERIMENTAL]: Add-On Subscription Options

fire-at-will:
### Description
Adds support for specifying specific `SubscriptionOption`s when purchasing a multi-line subscription.

### New APIs
The new API is consistent with the add-on item APIs introduced for `StoreProduct`s and `Package`s, introduced in https://github.com/RevenueCat/purchases-android/pull/2668. They are annotated as experimental for now, and the annotation will be removed in a future PR.
```kotlin
PurchaseParams.Builder.addOnSubscriptionOptions(addOnSubscriptionOptions: List<SubscriptionOption>)
```

Usage:
```kotlin
// Fetch subscription options from getOfferings()

val purchaseParamsBuilder = PurchaseParams.Builder(
   activity = currentActivity,
   subscriptionOption = baseSubscriptionOption,  // Base item
)
   // Add-ons, purchased in the order provided
   .addOnSubscriptionOptions(listOf(addOnSubOption1, addOnSubOption2))
   .build()

// Standard purchase function call...
```

#### Add On Builder APIs Now Combine Add Ons Across Multiple Calls
Previously, calling `PurchaseParams.Builder`’s `.addOnStoreProducts()` or `.addOnPackages()` would overwrite any add-ons that had already been added. This made it cumbersome to include multiple add-ons when combining different `addOn` builder functions in the same `PurchaseParams.Builder`. Now, repeated calls to `.addOnStoreProducts()`, `.addOnPackages()`, or `.addOnSubscriptionOptions()` will append their add-ons instead, allowing multiple sets to stack together.

For an example, consider this code:
```kotlin
val purchaseParamsBuilder = PurchaseParams.Builder(
   activity = currentActivity,
   subscriptionOption = subscriptionOptionForProduct1,  // Base item
)
   // Add-ons, purchased in the order provided
   .addOnSubscriptionOptions(listOf(subscriptionOptionForProduct2))
   .addOnPackages(listOf(packageForProduct3))
   .build()
```

Before the changes this would result in this purchase:
- Base item: subscription option for product 1
- Add-ons: product 3's default subscription option

After the changes this would result in this purchase:
- Base item: subscription option for product 1
- Add-ons: 
   - subscription option for product 2
   - product 3's default subscription option

#### Purchase Tester Updates
The purchase tester now takes into account the subscription option that is selected in the UI when purchasing add-ons. If one is selected, it uses the selected subscription option. Otherwise, it uses the default one.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/QueryProductDetailsParamsBuilderException.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/QueryProductDetailsParamsBuilderException.kt
@@ -1,0 +1,6 @@
+package com.revenuecat.purchases.google
+
+internal class QueryProductDetailsParamsBuilderException(message: String?, cause: Throwable?) : Exception(
+    message,
+    cause,
+)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/billingClientParamBuilders.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/billingClientParamBuilders.kt
@@ -4,6 +4,7 @@ import com.android.billingclient.api.BillingClient
 import com.android.billingclient.api.QueryProductDetailsParams
 import com.android.billingclient.api.QueryPurchaseHistoryParams
 import com.android.billingclient.api.QueryPurchasesParams
+import com.revenuecat.purchases.common.errorLog
 
 internal fun @receiver:BillingClient.ProductType String.buildQueryPurchaseHistoryParams(): QueryPurchaseHistoryParams? {
     return when (this) {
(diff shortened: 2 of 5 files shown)


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/OfferingFragment.kt
```diff
@@ -196,16 +202,33 @@ class OfferingFragment : Fragment(), PackageCardAdapter.PackageCardAdapterListen
     }
 
     @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
-    private fun startAddOnPurchase(selectedPackages: List<Package>) {
+    private fun startAddOnPurchase(
```
fire-at-will, line 205: We've modified this purchase function to use the subscription option APIs when a subscription option is available. Otherwise, it defaults to using the package APIs.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -161,22 +169,48 @@ class PurchaseParams(val builder: Builder) {
...
+        fun addOnSubscriptionOptions(addOnSubscriptionOptions: List<SubscriptionOption>) = apply {
+            val compatibleAddOnProducts: List<GooglePurchasingData> = addOnSubscriptionOptions
+                .mapNotNull { it.purchasingData as? GooglePurchasingData.Subscription }
+
+            attachSubscriptionAddOns(addOns = compatibleAddOnProducts)
+        }
+
+        @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+        private fun attachSubscriptionAddOns(addOns: List<GooglePurchasingData>) = apply {
+            if (addOns.isEmpty()) {
```
tonidero, line 198: NABD but should we just return after logging in this case?

tonidero (approved):
I think this makes sense, thanks for adding this!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
fire-at-will: Makes sense! I've added that return here: ae6763d09a294f3049ec004a7025e83056fe83ba

PR #2765, opened 2025-10-23
Change test server down mechanism to use new server down endpoint

tonidero:
### Description
In order to test the server being down, we had a mocked response in our integration test. However, this didn't test the exact behavior of the backend. This PR just changes the mechanism we use to test these situations to instead use a specific endpoint that will always behave as if the server was down.


purchases/src/main/kotlin/com/revenuecat/purchases/ForceServerErrorStrategy.kt
```diff
@@ -0,0 +1,25 @@
+package com.revenuecat.purchases
+
+import com.revenuecat.purchases.common.AppConfig
+import com.revenuecat.purchases.common.networking.Endpoint
+import java.net.URL
+
+internal interface ForceServerErrorStrategy {
+    companion object {
+        const val serverErrorURL = "https://api.revenuecat.com/force-server-failure"
+        val failAll = object : ForceServerErrorStrategy {
```
tonidero, line 14: These are currently unused, but plan to use them in a follow-up PR.

purchases/src/test/java/com/revenuecat/purchases/common/AppConfigTest.kt
```diff
@@ -221,21 +221,6 @@ class AppConfigTest {
...
-            context = mockk(relaxed = true),
-            purchasesAreCompletedBy = REVENUECAT,
-            showInAppMessagesAutomatically = false,
-            platformInfo = PlatformInfo(flavor = "native", version = "3.2.0"),
-            proxyURL = null,
-            store = Store.PLAY_STORE,
-            isDebugBuild = false,
-            apiKeyValidationResult = APIKeyValidator.ValidationResult.VALID,
-        )
-        assertThat(appConfig.forceServerErrors).isFalse
```
tonidero, line 236: Removing this test since this doesn't live here anymore

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
```diff
@@ -86,49 +86,6 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
         assertThat(result.body.getString("response")).`as`("response is OK").isEqualTo("OK")
     }
 
-    // region forceServerErrors
-
-    @Test
-    fun `returns server error result when forcing server errors`() {
```
tonidero, line 92: These cannot be tested well with unit tests with the new changes.

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
```diff
@@ -464,6 +463,7 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
             every { customEntitlementComputation } returns false
             every { isDebugBuild } returns true
             every { isAppBackgrounded } returns false
+            every { runningTests } returns false
```
ajpallares: Just to be extra safe. Shouldn't this be `true`?

tonidero: Yup! Fixed in https://github.com/RevenueCat/purchases-android/pull/2765/commits/13c06c16bf8a99cd141a655fd232db51be7f8d55. Nice catch!!

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
ajpallares: I did add some unit tests that check the `ForceServerErrorStrategy` implementation on `HTTPClient` (see https://github.com/RevenueCat/purchases-ios/blob/4ab4375207b3f28cc9b5de3e8cefb66ee6051903/Tests/UnitTests/Networking/HTTPClientTests.swift).

In case it's useful here

purchases/src/main/kotlin/com/revenuecat/purchases/ForceServerErrorStrategy.kt
JayShortway: Beautiful pattern! Maybe in the follow-up PR we could make it a `fun interface`?

```kotlin
internal fun interface ForceServerErrorStrategy
```

That should make instantiating implementations a bit more ergonomic.

tonidero: @ajpallares's idea! It's indeed great :) And yeah, that will definitely be nicer, great idea!

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2773

PR #2769, opened 2025-10-24
Add fallback URL integration tests

tonidero:
### Description
Adds integration tests to verify behavior when main url and load shedder are down and we fallback to fallback URL.

### DIFF
--- a/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
+++ b/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
@@ -62,6 +62,11 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Test
     fun customerInfoCanBeFetched() {
+        if (isRunningLoadShedderIntegrationTests()) {
+            // Test won't work correctly in load shedder since we can only fetch customer info once a purchase happens
+            return
+        }
+
         val lock = CountDownLatch(1)
 
         onActivityReady {
@@ -77,6 +82,11 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Test
     fun customerInfoCanBeFetchedFromBackendAndThenGottenFromCache() {
+        if (isRunningLoadShedderIntegrationTests()) {
+            // Test won't work correctly in load shedder since we can only fetch customer info once a purchase happens
+            return
+        }
+
         val lock = CountDownLatch(1)
 
         activityScenarioRule.scenario.onActivity {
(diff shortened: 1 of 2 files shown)


purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/fallbackurl/PurchasesFallbackURLTest.kt
```diff
@@ -0,0 +1,182 @@
...
+    fun customerInfoCannotBeFetchedFromFallbackURL() {
+        ensureBlockFinishes { latch ->
+            Purchases.sharedInstance.getCustomerInfoWith({
+                assertThat(it.code).isEqualTo(PurchasesErrorCode.UnknownBackendError)
+                latch.countDown()
+            }) {
+                fail("should be error")
+            }
+        }
+    }
```
ajpallares: I'm implementing a similar test in iOS but I'm not getting an error here. Instead, I'm getting the Customer Info computed offline. Should this be the same in Android?

tonidero: ~~Hmm I think offline entitlements would only be used if we find any purchases in Android... In this test, there aren't any purchases from Google, so we just return an error.~~ Actually ignore that, I think it should have indeed entered offline entitlements, checking

tonidero: ~~Ok I see what's happening. The issue is that I don't have set it up to wait until product entitlement mappings have finished downloading, so it doesn't have them in order to enter offline entitlements, so it would fail instead.~~

Edit: It's actually using offline entitlements, so it's the same behavior in iOS, but seems to be flaky due to the above. Will fix that so it's not flaky.

tonidero: Changed temporarily in https://github.com/RevenueCat/purchases-android/pull/2769/commits/b713fbf8827d90989e3cba3b55161b1af342de68

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/FallbackURLBackendIntegrationTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        assertThat(error).isNull()
+        verify(exactly = 1) {
+            // Verify we save the backend response in the shared preferences
+            sharedPreferencesEditor.putString(Endpoint.GetProductEntitlementMapping.getPath(), any())
+        }
+        verify(exactly = 1) { sharedPreferencesEditor.apply() }
+        assertSigningNotPerformed()
+    }
+
+//    Commenting out test until we fix signing in fallback URL. See Linear SDK-4152
```
ajpallares: I have this PR up already https://github.com/RevenueCat/purchases-android/pull/2780 implementing the new fallback URLs. If it's useful to you, you can change the base branch of this PR into my PR's branch and then you should be able to run these tests

tonidero: Thanks! Did that and after a couple fixes it's working now 🙌

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/fallbackurl/PurchasesFallbackURLTest.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    @Test
+    fun customerInfoCannotBeFetchedFromFallbackURLSoUsesAnOfflineCalculatedOne() = runTest {
+        waitForProductEntitlementMappingToUpdate()
+
+        val appUserID = Purchases.sharedInstance.appUserID
+
+        val customerInfo = Purchases.sharedInstance.awaitCustomerInfo()
+        assertThat(customerInfo.entitlements.active).isEmpty()
+        assertThat(customerInfo.originalAppUserId).isEqualTo(appUserID)
```
ajpallares, line 50: Should we also assert here this?
```
assertThat(customerInfo.entitlements.verification).isEqualTo(VerificationResult.VERIFIED_ON_DEVICE)
```

tonidero: Good call! https://github.com/RevenueCat/purchases-android/pull/2769/commits/56a90667885fd79474775eb761942e036253786c

PR #2772, opened 2025-10-24
Use cached offerings on network errors

tonidero:
### Description
This PR changes the behavior when get offerings fails with a network error of any kind so we use the cached offerings if any. Until now, we only used those when the server returned a 5xx. This adds new cases where we do this.

### DIFF
--- a/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
+++ b/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
@@ -81,7 +81,7 @@ class CachedOfferingsUsageIntegrationTest : BasePurchasesIntegrationTest() {
                     return false
                 }
 
-                override fun shouldFakeResponseWithoutPerformingRequest(baseURL: URL, endpoint: Endpoint): HTTPResult? {
+                override fun fakeResponseWithoutPerformingRequest(baseURL: URL, endpoint: Endpoint): HTTPResult? {
                     if (endpoint is Endpoint.GetOfferings) {
                         return HTTPResult(
                             responseCode = 401,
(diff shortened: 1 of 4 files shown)


purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
```diff
@@ -0,0 +1,107 @@
...
+            setUpTest {
+                latch.countDown()
+            }
+        }
+
+        mockBillingAbstract.mockQueryProductDetails()
+    }
+
+    @Test
+    fun cachedOfferingsAreUsedWhenCachedOfferingsAndServerErrorWith5xx() = runTest {
```
tonidero, line 42: I decided to write these as coroutines tests. I think we probably should be moving all these tests to this syntax since it's much nicer to read IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/ForceServerErrorStrategy.kt
```diff
@@ -16,4 +17,7 @@ internal fun interface ForceServerErrorStrategy {
         get() = "https://api.revenuecat.com/force-server-failure"
 
     fun shouldForceServerError(baseURL: URL, endpoint: Endpoint): Boolean
+    fun shouldFakeResponseWithoutPerformingRequest(baseURL: URL, endpoint: Endpoint): HTTPResult? {
```
ajpallares: I think the function name could be misleading. The `should` prefix could be assumed to return a `Boolean`. Maybe this is clearer?
```suggestion
    fun fakeResponseWithoutPerformingRequest(baseURL: URL, endpoint: Endpoint): HTTPResult? {
```
I think a `null` here is more easily interpreted as a "no fake response without performing the request"

tonidero: You're totally right, thanks!

tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/2772/commits/13eaebc07c24aac3186d26b8b2eb82f7284d26a3

PR #2773, opened 2025-10-24
Change ForceServerErrorStrategy to be a `fun interface` to simplify instantiation

tonidero:
### Description
As title says. Followup from #2765


purchases/src/main/kotlin/com/revenuecat/purchases/ForceServerErrorStrategy.kt
```diff
@@ -4,22 +4,12 @@ import com.revenuecat.purchases.common.AppConfig
 import com.revenuecat.purchases.common.networking.Endpoint
 import java.net.URL
 
-internal interface ForceServerErrorStrategy {
+internal fun interface ForceServerErrorStrategy {
```
vegaro, line 7: Lol I was reading the title and thought, how can an interface be a fun thing?

ajpallares: Same! I was also expected something "fun_er_ 🤣

PR #2775, opened 2025-10-24
Add caret to subscriptions and only open purchase detail if there is detailed actions available

vegaro:
We were only showing caret right in the in app purchases, not in the subscriptions, which was confusing since it wasn't easy to understand when things were clickable. 

| Header | After |
|--------|--------|
| image | image | 

We were also allowing clicks even if there was no particular action in the detailed view, so this PR also fixes that. This showcases what I mean, and it also shows the wrong carets:

[Screen_recording_20251024_182630.webm](https://github.com/user-attachments/assets/c7e97447-9566-4608-926c-f5ac955d9855)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PurchaseInformationCardView.kt
```diff
@@ -83,16 +83,14 @@ internal fun PurchaseInformationCardView(
...
                         Row(
                             horizontalArrangement = Arrangement.spacedBy(
                                 CustomerCenterConstants.Card.BADGE_HORIZONTAL_PADDING,
                             ),
                             verticalAlignment = Alignment.CenterVertically,
                         ) {
-                            if (purchaseInformation.isLifetime) {
-                                PurchaseStatusBadge(purchaseInformation, localization)
-                            }
+                            PurchaseStatusBadge(purchaseInformation, localization)
```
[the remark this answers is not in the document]
vegaro: This is ok

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterConfigTestData.kt
```diff
@@ -238,7 +238,7 @@ internal object CustomerCenterConfigTestData {
         isExpired = false,
         isTrial = false,
         isCancelled = true,
-        isLifetime = false,
+        isLifetime = true,
```
facumenzella, line 262: Is this related?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -846,11 +875,16 @@ internal class CustomerCenterViewModelImpl(
                 detailScreenPaths = emptyList(), // Will be computed when a purchase is selected
                 noActiveScreenOffering = noActiveScreenOffering,
                 virtualCurrencies = virtualCurrencies,
+                purchasesWithActions = emptySet(), // Will be computed below
             )
             val mainScreenPaths = computeMainScreenPaths(successState)
+            val purchasesWithActions = computePurchasesWithActions(successState)
```
facumenzella: This whole dance is confusing IMO. Can't we have `computePurchasesWithActions` receive individual paremeters rather than Succes state?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -466,13 +466,14 @@ private fun MainScreenContent(
                 appearance = configuration.appearance,
                 localization = configuration.localization,
                 onPurchaseSelect = { purchase ->
-                    // Only allow selection if there are multiple purchases
-                    if (state.purchases.size > 1) {
+                    // Only allow selection if there are multiple purchases and the purchase has actions
+                    if (state.purchases.size > 1 && purchase in state.purchasesWithActions) {
```
facumenzella, line 566: What purchase would have no actions at all? 🤔 

I wonder if this check is necessary. In iOS we always navigate, even if there are no actions.

vegaro: A purchase can have no detail-screen actions when it's a non-Google subscription (no Cancel/Resubscribe from our config), or it's expired with no Resubscribe path, or the management screen only has general paths (e.g. Restore, Contact support) and no subscription specific paths for that purchase. In those cases the detail screen would be empty.

The check avoids navigating to an empty detail view and only shows the caret when there’s at least one action (e.g. Cancel/Resubscribe). If we want to align with iOS and always navigate even when the detail is empty, we can remove the check but I think this is nicer? Idk

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterConfigTestData.kt
vegaro: yes, this was wrong test data setup

PR #2776, opened 2025-10-24
Unlock entitlements from add-ons in offline entitlements mode

fire-at-will:
### Description
Adds support for unlocking entitlements from products purchased as a subscription add-on. This is done by modifying the `OfflineCustomerInfoCalculator` to consider all product IDs in a `StoreTransaction` when calculating the purchased products, instead of just the first one.

 The PR has the `pr:other` tag since support for subscriptions with add-ons is still experimental.

### DIFF
--- a/purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
+++ b/purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
@@ -11,9 +11,11 @@ import com.revenuecat.purchases.models.StoreTransaction
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.slot
+import io.mockk.verify
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.runBlocking
 import org.assertj.core.api.Assertions
+import org.assertj.core.api.Assertions.fail
 import org.junit.After
 import org.junit.BeforeClass
 import org.junit.Rule
@@ -21,6 +23,10 @@ import java.net.URL
 import java.util.Date
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
 
 open class BasePurchasesIntegrationTest {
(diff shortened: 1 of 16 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/PurchasedProductsFetcher.kt
```diff
@@ -36,30 +36,31 @@ internal class PurchasedProductsFetcher(
...
-        return PurchasedProduct(
-            productIdentifier,
-            mapping?.basePlanId,
-            transaction,
-            mapping?.entitlements ?: emptyList(),
-            expirationDate,
-        )
+        return transaction.productIds.mapNotNull { productIdentifier ->
+            val mapping = productEntitlementMapping.mappings[productIdentifier] ?: return@mapNotNull null
+            PurchasedProduct(
```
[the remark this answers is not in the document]
fire-at-will: addressed in 781da080810c1d701e289b754be893165071f3e3

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/PurchasedProductsFetcher.kt
```diff
@@ -36,30 +36,34 @@ internal class PurchasedProductsFetcher(
...
-            mapping?.basePlanId,
-            transaction,
-            mapping?.entitlements ?: emptyList(),
-            expirationDate,
-        )
+
+        return transaction.productIds
+            .filter { productIdentifier -> productEntitlementMapping.mappings.containsKey(productIdentifier) }
+            .map { productIdentifier ->
+                val mapping = productEntitlementMapping.mappings.getValue(productIdentifier)
```
tonidero: ~~Hmm this can crash if the product identifier is not found in the mappings right? Until now we were just defaulting to no entitlements, which seems like a safer approach?~~ 

Actually, I see we're filtering out before so we wouldn't crash, but we wouldn't have the PurchasedProduct at all either... I think it's better to have it with no entitlements to properly reflect that there were purchased products, just that with no entitlements right?

fire-at-will: You're right, I think there are two issues with the proposed change:
1. The code won't crash if the mapping isn't found because of the preceding `.filter`. However, it's not the most safe code in the long term since the crash in the proceeding `.map` won't be immediately obvious to someone who removes the filter in the future
2. This filters out the purchased products without entitlements, which is a breaking change in behavior. The difference is subtle but meaningful. This actually means that a few of my tests related to add-ons were asserting the incorrect behavior.

I've made a few changes in commit 0874b2cceea1fe233e53004bfbe1c9ad5176bf29:
- Modified the logic in `PurchasedProductsFetcher`:
   - Removed the `.filter` entirely
   - Adjusted the `.map` function to use a safe lookup when looking up the entitlement in the `mappings` map
   - Adjusted the `.map` function to return products without entitlements, using the same logic that we had before this PR
- Modified two add-on tests `PurchasedProductsFetcherTest` to reflect that purchased products should still return the products even if they don't have associated entitlements
- Added a new test to `PurchasedProductsFetcherTest`: `returns an active purchased product without an entitlement` to assert that we return purchased products without an entitlement.

Let me know what you think!

PR #2780, opened 2025-10-27
[SDK-4152] Use static fallback urls

ajpallares:
### Motivation
The signature for requests on the fallback server is done only for static endpoints, not supporting dynamic paths like the GET offerings one (`/v1/subscribers/APP_USER_ID/offerings`).

### Description
Changes the fallback urls into static ones to support signature verification.

In particular:
* The fallback url for GET offerings now uses the `v1/offerings` path (previously it was the same relative path as for the main server: `/v1/subscribers/APP_USER_ID/offerings`)
* The fallback url for GET product entitlement mapping is maintained to use the `v1/product_entitlement_mapping` relative path.

iOS counterpart: https://github.com/RevenueCat/purchases-ios/pull/5709

### DIFF
--- a/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
+++ b/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
@@ -62,6 +62,11 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Test
     fun customerInfoCanBeFetched() {
+        if (isRunningLoadShedderIntegrationTests()) {
+            // Test won't work correctly in load shedder since we can only fetch customer info once a purchase happens
+            return
+        }
+
         val lock = CountDownLatch(1)
 
         onActivityReady {
@@ -77,6 +82,11 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Test
     fun customerInfoCanBeFetchedFromBackendAndThenGottenFromCache() {
+        if (isRunningLoadShedderIntegrationTests()) {
+            // Test won't work correctly in load shedder since we can only fetch customer info once a purchase happens
+            return
+        }
+
         val lock = CountDownLatch(1)
 
         activityScenarioRule.scenario.onActivity {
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -55,11 +55,11 @@ internal class ETagManager(
 ) {
 
     internal fun getETagHeaders(
-        path: String,
+        urlString: String,
```
ajpallares, line 58: Although this is only a parameter name change in `EtagManager`, the idea here is to stop using only the path as the key for caching the responses and use the full URL of the request. The reason is that we might be making requests with the same path to different endpoints (e.g. `vi/product_entitlement_mapping`  to both the main server and to the fallback server) and we don't want to return the response of a server for requests to the other. Please, lmk if you don't agree with this change @RevenueCat/coresdk 

Note that the same is done already in iOS (see https://github.com/RevenueCat/purchases-ios/blob/303f7dfb7b27fe24db61a3cfa851e35571447355/Sources/Networking/HTTPClient/ETagManager.swift#L132-L134)

tonidero: I think it makes sense... Only concern is whether this could cause current etags to stop being found... But even if that's the case, it's probably ok I think, just means one extra body to be sent IMO...

tonidero (commented):
I think changes make sense! Seems there are some failing tests and there is that small concern of stop using our current Etag cache... but I wouldn't consider that a blocker.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
ajpallares: Yep. That is a side effect of this. I checked and the cache seems to only be cleaned up whenever the user ID changes. 
In addition, with the current change it would mean that the cache will still (always) contain the previous values.

Do you think it's worth keeping the old cache keys and only add new keys for the endpoints of the new fallback URL? I don't really know what's best

tonidero: Hmm well... I don't think it's a huge deal to leave the old keys as they are since it's basically a small entry per each endpoint that supports etags so not a lot of space and special casing is probably more complex... So I think it's fine to leave it this way IMO.

ajpallares: Yes, makes sense. Thank you!

PR #2781, opened 2025-10-28
Add golden tests for requests and responses in backend integration tests

tonidero:
### Description
This will add golden tests for the requests and responses in backend integration tests, which will allow us to detect changes both in khepri and load shedder and cloudfront requests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,6 @@ purchases-android-snapshots
 
 # Make sure we share the same Detekt IntelliJ plugin settings
 !.idea/detekt.xml
+
+# Kotlin error logs
+.kotlin/errors/


.kotlin/errors/errors-1761217878314.log
ajpallares, line 1: Should this file be ignored?

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -256,6 +275,40 @@ internal class HTTPClient(
             throw IOException(NetworkStrings.HTTP_RESPONSE_PAYLOAD_NULL)
         }
 
+        // Notify listener if present
```
ajpallares, line 278: Should we wrap this in
```kotlin
if (appConfig.runningTests) {
```
?

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/GoldenFileRecorder.kt
```diff
@@ -0,0 +1,253 @@
...
+                } else {
+                    navigateAndFilter(current.get(part), parts, index + 1)
+                }
+            }
+        }
+    }
+
+    private fun recordGoldenFile(sequenceNumber: String, request: RecordedRequest, response: RecordedResponse) {
+        val requestFile = File(testDirectory, "request_$sequenceNumber.json")
+        val responseFile = File(testDirectory, "response_$sequenceNumber.json")
```
ajpallares, line 249: I see requests/responses are being stored as `request_001`, `request_002`.

I'm wondering if we should be so strict about checking the requests (we'd be also checking the requests order). My thinking is that this could break fairly easily if we add new requests, change the requests orders, etc. I'm not sure anyway

tonidero: Well note that this right now is run only for the unit tests in `backend_integration_tests` that mostly verify actual calls, but it's not a "real" E2E test that verifies by configuring the entire SDK and verifies all the requests performed. This is similar to backend integration tests in iOS.

In this case, I think it's proper to be strict about the number and order of requests. Right now the only requests that actually do multiple requests are those that hit the fallback URL (where we first hit the server error endpoint, then use the fallback). For all others, a single request/response fair is generated. I do think these tests should be kept minimal.

To add E2E tests, I don't think this kind of golden tests would be the best approach IMO since as you said there is much more uncertainty.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: That's a good idea 👍 Will add it, thanks!

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/GoldenFileRecorder.kt
ajpallares: Oh yes. Good points. If we're limiting this to minimal `backend_integration_tests` then I think it's fine to be this strict 👍 
Thank you for the explanation!

.circleci/config.yml
```diff
@@ -696,8 +696,21 @@ jobs:
...
+          command: |
+            if ! git diff --quiet purchases/src/test/resources/backend_integration_tests_golden/; then
+              echo "ERROR: Golden test files have been modified!"
+              echo "The following changes have been found:"
+              git diff purchases/src/test/resources/backend_integration_tests_golden/
+              echo ""
+              echo "Golden test files should not be modified during test execution."
+              echo "If you need to update the golden files, commit the changes explicitly."
+              exit 1
+            fi
```
ajpallares, line 710: Perhaps not a huge deal if kept like this, as the job is still reported as failed (e.g. [here](https://circleci.com/gh/RevenueCat/purchases-android/108507)) but it would be great if this check could be part of the individual tests, so that we get a better report of the failure(s).

Right now, we are getting all tests in green (passed) and that could be confusing.

Image

Still, the error log is clear enough but could be more difficult to understand that it's coming from a test failure for someone less familiar with this (like me).

tonidero: Hmm I think that would be doable by adding some sort of "Verifying" mode, where we don't generate new snapshots but just compare them as we go through the test, and fail the moment there is a request that doesn't match the golden... But it's actually tricky due to the way the tests are written (The "ensureBlockFinishes" would swallow exceptions in this case, so we would need to rewrite most tests...). I can see how it could be confusing, but would prefer to dedicate the time to the other reliability tasks for now and we can revisit this if it's a problem. Wdyt?

ajpallares: If it's not something obvious I think it could totally wait. Thank you for the explanation! 👌

PR #2782, opened 2025-10-29
Add package fallback to tab selection when there's no default selected

vegaro:
We got a report of a paywall that in Android is not showing the Annual product and has some odd behavior when toggling on/off on the switch as the UI Footer moves up and down. iOS behaves as expected.

Screenshot 2025-10-29 at 08 42 20

I pinpointed it to be related to the default package of the OFF tab. It looks like there's no default selected and that's making Android default to null package. 

Screenshot 2025-10-29 at 08 42 56

I looked at the iOS implementation and added a fallback here in the Android code.

https://linear.app/revenuecat/issue/DENG-1483/i-have-paywall-trial-paywall-1

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
@@ -245,7 +245,8 @@ internal sealed interface PaywallState {
                         ?: packages.packagesByTab[selectedTabIndex]?.firstOrNull()?.pkg?.also {
                             Logger.w(
                                 "Could not find default package for tab $selectedTabIndex. " +
-                                    "Using first package instead. This could be caused by not having any package marked as selected by default.",
+                                    "Using first package instead. " +
+                                    "This could be caused by not having any package marked as selected by default.",
                             )
                         }
                 }


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -240,7 +240,15 @@ internal sealed interface PaywallState {
...
                     if (packagesOutsideTabs.contains(selectedPackage)) return
 
-                    selectedPackage = selectedPackageByTab[selectedTabIndex] ?: initialSelectedPackageOutsideTabs
+                    selectedPackage = selectedPackageByTab[selectedTabIndex]
+                        ?: initialSelectedPackageOutsideTabs
+                        ?: packages.packagesByTab[selectedTabIndex]?.firstOrNull()?.pkg?.also {
+                            Logger.w(
+                                "Could not find default package for tab $selectedTabIndex. " +
+                                    "Using first package instead. This could be caused by a package that " +
+                                    "isn't marked as selected by default in the tab configuration.",
```
JayShortway:
```suggestion
                                    "Using first package instead. This could be caused by not having any package marked as selected by default.",
```

JayShortway (approved):
Thanks for handling this! 💪 

I think it makes sense for the SDK to be extra careful, but I'd also argue that the real bug is in the editor, which should always have 1 package selected by default. Could we raise that in the paywalls team? 🙏

PR #2783, opened 2025-10-29
Skip tests not supported in fortress

tonidero:
### Description
We used to run these test in load shedder integration tests. It wasn't an issue since we weren't actually redirecting customer info requests to the load shedder. We added a new rule to do that now, which was causing these tests to fail when run against the load shedder. They don't really apply since these tests would require to purchase before fetching customer info.

In its place, we added one more test to cover these new behavior.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/LoadShedderBackendIntegrationTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/LoadShedderBackendIntegrationTest.kt
@@ -1,7 +1,6 @@
 package com.revenuecat.purchases.backend_integration_tests
 
 import com.revenuecat.purchases.PurchasesError
-import com.revenuecat.purchases.VerificationResult
 import com.revenuecat.purchases.common.networking.Endpoint
 import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMapping
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode


purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/LoadShedderBackendIntegrationTest.kt
```diff
@@ -155,52 +154,4 @@ internal class LoadShedderBackendIntegrationTest: BaseBackendIntegrationTest() {
         }
         assertSigningPerformed()
     }
-
-    @Test
-    fun `can perform login backend request`() {
```
tonidero, line 160: I've removed the logIn tests from the load shedder integration tests since it's not actually supported. It used to work since we weren't actually pointing to the load shedder. The behavior is still tested in production in `ProductionBackendIntegrationTest`

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
```diff
@@ -165,47 +175,21 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Test
     fun canPurchaseSubsProduct() {
-        val lock = CountDownLatch(1)
-
-        val storeProduct = StoreProductFactory.createGoogleStoreProduct()
-        val storeTransaction = StoreTransactionFactory.createStoreTransaction()
-        mockBillingAbstract.mockQueryProductDetails(queryProductDetailsSubsReturn = listOf(storeProduct))
+        performPurchase()
```
tonidero, line 178: I've just basically moved the whole test into this function so I can reuse it in the new test

PR #2784, opened 2025-10-29
FIX: Video component - felt slow on first appearance

JZDesign:
### Checklist

### Motivation

The image would not render when the low-res video was starting to load before the initial cache was populated. So we need to explicitly set and unset the image at the appropriate times.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
@@ -57,7 +57,7 @@ internal fun VideoComponentView(
                 .clip(composeShape)
                 .applyIfNotNull(borderStyle) { border(it, composeShape).padding(it.width) },
         ) {
-            if (videoUrl == null && fallbackImageViewStyle != null) {
+            if (fallbackImageViewStyle != null) {
                 ImageComponentView(fallbackImageViewStyle, state, modifier)
             }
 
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -57,7 +57,7 @@ internal fun VideoComponentView(
                 .clip(composeShape)
                 .applyIfNotNull(borderStyle) { border(it, composeShape).padding(it.width) },
         ) {
-            if (videoUrl == null && fallbackImageViewStyle != null) {
+            if (fallbackImageViewStyle != null) {
```
fire-at-will, line 60: This will cause the `ImageComponentView` to be rendered pretty much all the time (even though it'd be covered up by the video view once loaded), right? Would it be better from a resource management and accessibility perspective to remove this `ImageComponentView` from the view hierarchy when the video view is rendered in the `.let { }` block below?

fire-at-will: Oh, I just saw that we're setting `fallbackImageViewStyle = null` later in the file after a video is loaded 😅

JZDesign: Yea, the idea is:

1. Show the image
1. If there is a low res video and no cached video file - show this on top of the image
1. when the full res file is cached, show that and hide the image

PR #2786, opened 2025-10-29
Video Background

JZDesign:
### Checklist

### Motivation

Make it easy to add videos to a paywall.

### Description

Rendering the video as a background requires that it be in a z-stack and THAT seems to require very explicit sizing for android to actually render the video component at all. The only way I could figure out how to do that and support the custom layout that we're doing for buttons and no-badge stacks was to essentially duplicate the sizing logic that happens in side of the stack function. If we apply those same modifiers on the optional video view when it's not a video, it's not put in a box, and then we get a crash for making an infinitely scrolling view. So I have wrapped it in another check for now. It's ugly, and I know it, so I'd like someone more skilled with Compose to look at it. 

https://github.com/user-attachments/assets/d4537f17-63ee-4806-a3fa-77e5ba3856c7

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
@@ -20,8 +20,8 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.nonEmptyMapOf
 internal fun ViewWithVideoBackground(
     state: PaywallState.Loaded.Components,
     background: BackgroundStyle?,
+    modifier: Modifier = Modifier,
     shape: Shape = RectangleShape,
-    modifier: Modifier,
     content: @Composable BoxScope.() -> Unit,
 ) {
     Box(modifier = modifier) {
@@ -67,8 +67,8 @@ internal fun ViewWithVideoBackground(
 @Composable
 internal fun WithOptionalVideoBackground(
     state: PaywallState.Loaded.Components,
-    modifier: Modifier = Modifier,
     background: BackgroundStyle?,
+    modifier: Modifier = Modifier,
     shape: Shape = RectangleShape,
     content: @Composable () -> Unit,
 ) {


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -1,3 +1,4 @@
+
```
JZDesign, line 1: Detekt was mad about this line when I was in the file… I kinda want to revert this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -39,4 +40,9 @@ internal fun Modifier.background(
                     contentScale = background.contentScale,
                 )
                 .applyIfNotNull(background.colorOverlay) { underlay(it, shape) }
+        is BackgroundStyle.Video ->
+            // Video backgrounds are handled specially - they need to be rendered
+            // in a Box behind the content, so we just clip the shape here
+            // TODO…………
```
JZDesign:
```suggestion
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -616,22 +618,53 @@ private fun MainStackComponent(
...
+                    )
+                }
+            }
+        } else {
+            stack(
+                outerShapeModifier.then(borderModifier).then(
+                    innerShapeModifier
+                        .conditional(stackState.applyBottomWindowInsets) {
+                            windowInsetsPadding(systemBarInsets.only(WindowInsetsSides.Bottom))
+                        },
```
JayShortway: Not sure if it matters in practice, but the chaining here seems different from what it was before? If we add the same line breaks we get:


```kotlin
                outerShapeModifier
                    .then(borderModifier)
                    .then(
                        innerShapeModifier
                            .conditional(stackState.applyBottomWindowInsets) {
                                windowInsetsPadding(systemBarInsets.only(WindowInsetsSides.Bottom))
                            },
```

instead of:

```kotlin
                outerShapeModifier
                    .then(borderModifier)
                    .then(innerShapeModifier)
                    .conditional(stackState.applyBottomWindowInsets) {
                        windowInsetsPadding(systemBarInsets.only(WindowInsetsSides.Bottom))
                     },
```

The only change is missing closing parenthesis in `then(innerShapeModifier)`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.revenuecat.purchases.paywalls.components.properties.Size
+import com.revenuecat.purchases.paywalls.components.properties.SizeConstraint
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toLocaleId
+import com.revenuecat.purchases.ui.revenuecatui.components.properties.BackgroundStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.style.VideoComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.components.video.VideoComponentView
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallState
+import com.revenuecat.purchases.ui.revenuecatui.helpers.nonEmptyMapOf
+
+@Suppress("ModifierNotUsedAtRoot")
```
JayShortway: Any issues with using the Modifier for the root `Box`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
```diff
@@ -0,0 +1,89 @@
...
+                    contentScale = background.contentScale,
+                    rcPackage = null,
+                    tabIndex = null,
+                    ignoreTopWindowInsets = true,
+                    overrides = emptyList(),
+                    padding = PaddingValues(),
+                    margin = PaddingValues(),
+                    border = null,
+                    shadow = null,
+                )
```
JayShortway, line 52: Can we do this when creating the `BackgroundStyle.Video`, so we keep all "style creation" logic contained? Maybe `BackgroundStyle.Video` only needs 1 property of type `VideoComponentStyle`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -39,4 +40,8 @@ internal fun Modifier.background(
                     contentScale = background.contentScale,
                 )
                 .applyIfNotNull(background.colorOverlay) { underlay(it, shape) }
+        is BackgroundStyle.Video ->
+            // Video backgrounds are handled specially - they need to be rendered
+            // in a Box behind the content, so we just clip the shape here
+            this.clip(shape)
```
JayShortway: I'm thinking whether we should just no-op in this case. It might be _more_ unexpected that `.background(Video)` clips vs when it just doesn't do anything. Not a strong opinion.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway, line 1: Where's the duplicated sizing logic you mentioned in the PR description? 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
```diff
@@ -0,0 +1,89 @@
...
+
+                VideoComponentView(
+                    style = videoComponent,
+                    state = state,
+                    modifier = modifier.align(Alignment.Companion.Center),
```
JayShortway: Not sure if it helps, but more of an FYI: children of `Box` can also use `.matchParentSize()`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -616,22 +618,53 @@ private fun MainStackComponent(
...
+                ) {
+                    stack(
+                        Modifier
+                            .then(innerShapeModifier)
+                            .conditional(stackState.applyBottomWindowInsets) {
+                                windowInsetsPadding(systemBarInsets.only(WindowInsetsSides.Bottom))
+                            },
+                    )
+                }
+            }
```
JZDesign: @JayShortway my old comment was auto resolved… or I never hit the button. This is where the duplicated logic is. I had to apply it to the box here but some of the same logic is applied in the stack function.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
JZDesign: You may have advice for me here, but the modifier I'm passing in is `matchParentSize()` while in a different box scope because I need the video view to have a size in order for it to render. This is the trickiest bit and you or another Android pro may have a better way to do this than I figured out.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
JZDesign: You know… maybe that's all I need to do here and not pass in a modifier at all. I thought I had a bug where the container had to pass in that modifier, but I could be mis-remembering

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
JZDesign: In order for us to make the VideoComponentStyle we need to have access to the paywall state—at the very least we need access to the paywall's current locale. I'm not sure we should go through the effort to centralize the style generation for this case. Unless there is a way to access the current localization in a better way

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
joshdholtz: I don't know if what I'm going to say makes sense but... I _think_ clipping might make sense if somebody isn't showing the paywall with a video background NOT taking up the entire screen? So that the video stays within the bounds of the paywall view?

But IDK HOW ANYTHING WORKS

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
joshdholtz: @JayShortway Is it still good to merge this without that refactor? I don't know enough about this part of the codebase to know 🤷‍♂️

joshdholtz (approved):
This looks good to me! I'd _probably_ like one more approval from some Android folk

cc: @JayShortway @tonidero

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
JayShortway: Hmm, this would intersect some architectural layers that haven't been intersected before. A `*ComponentStyle` generally contains information for all locales. The `*ComponentState` is the one picking the right locale because that happens in the view layer. 

We are already creating a `VideoComponentStyle` in `StyleFactory` with all locales (for the `VideoComponentView`). Can we use the same pattern here, picking the right locale after creating the `VideoComponentStyle`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
JayShortway: Regular Composables always stay in the bounds of their parents (which is different from SwiftUI afaik), but I can see video being different in its rendering behavior, so would be good to check.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JayShortway: Ah okay, I don't think this is too bad!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
JZDesign: I don't think we can yet because the backgrounds don't yet have the `overrideSourceLid` which is the key used to find those values

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -616,22 +618,50 @@ private fun MainStackComponent(
...
-            outerShapeModifier
-                .then(borderModifier)
-                .then(innerShapeModifier)
-                .conditional(stackState.applyBottomWindowInsets) {
-                    windowInsetsPadding(systemBarInsets.only(WindowInsetsSides.Bottom))
-                },
-        )
+        if (backgroundStyle is BackgroundStyle.Video) {
+            // Video backgrounds require a Box wrapper with explicit sizing
+            Box(
```
tonidero: Hmm I'm wondering, could this work by passing these modifiers to the box inside the `ViewWithVideoBackground` composable, instead of wrapping the `WithOptionalVideoBackground` inside a `Box`. Feels like a level on indirection that might not be needed?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
tonidero: Hmmm ok... I do think this is indeed a bit of an architectural smell... but I don't want to block this due to this, maybe something we can address shortly after shipping this? Seems that it shouldn't cause breaking changes later if we change it, as far as I understand...

tonidero (approved):
Going to approve to not block the release of this feature more, but indeed, ideally we separate the style creation from the composable itself as soon as we can, to avoid later us mixing things up.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JZDesign: Thanks. Brain was apparently fried when I did this 😆

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
JZDesign: Yea… We'll get there. Not in love with this myself 😬

PR #2788, opened 2025-10-30
Adds Google login Maestro flow

JayShortway:
## Description
Adds my attempt from a few weeks ago logging in to a Google account using Maestro. It might need some further tweaking and obviously isn't used yet, but this way it's not some random file on my desktop haha.


maestro/google-login.yml
```diff
@@ -0,0 +1,42 @@
...
+# adb shell am start -a android.settings.ADD_ACCOUNT_SETTINGS -n com.android.settings/.accounts.AddAccountSettings;
+
+# This flow requires passing the user name and password as parameters like so:
+# maestro test -e USERNAME=user@example.com -e PASSWORD=123 google-login.yml
+
+appId: "com.android.settings"
+---
+- launchApp
+- scrollUntilVisible:
+    element: "Passwords, passkeys & accounts"
```
tonidero: This string seems to only be available in newer android versions (in API 33, it says Passwords and accounts), but as you said in the comments, we might be able to bypass those steps with the adb command above.

tonidero (commented):
This does seem to work very well for me! Tried a few times, and worked all the times 🙌 

Before approving, we might want to move it somewhere else, like "util_flows" or something along those lines, separate from the tests themselves since this flow could potentially be reused for multiple tests. Also, I think we would need to adapt the device where tests run to android 36 or change to use the ADB command for tests to pass.

maestro/google-login.yml
JayShortway: Yes, good point! I thought about being smart with regex, but it seemed simpler and less confusing to make 2 separate flows:
- `maestro/google-login/api/36.yml` for API >= 36 
- `maestro/google-login/api/32.yml` for API >= 32 

I'm not entirely sure if these are the exact API boundaries, but we can tweak as needed.

There's also a `maestro/google-login/google-login.yml` file, which selects the right API flow to use based on the `API_VERSION` parameter that is passed. I'm not sure of a way to automatically get the Android version in Maestro, but we can improve it if we find one. So to be clear, this is how it would work now: 

```shell
maestro test \
  -e API_VERSION=36 \
  -e USERNAME=user@example.com \
  -e PASSWORD=123 \
  google-login.yml;
```

maestro/google-login/api/32.yml
```diff
@@ -0,0 +1,46 @@
...
+      - repeat:
+          while:
+            visible: "Cancel"
+          commands:
+            - tapOn: "Cancel"
+      - repeat:
+          while:
+            visible: "Skip"
+          commands:
+            - tapOn: "Skip"
```
ajpallares, line 38: What happens if there is a "Cancel" after a "Skip"? I think this would not work.

I guess we know that this won't happen?

maestro/google-login/google-login.yml
```diff
@@ -0,0 +1,15 @@
+# This flow requires passing the Android API version, user name and password as parameters like so:
+# maestro test -e API_VERSION 36 -e USERNAME=user@example.com -e PASSWORD=123 google-login.yml
```
ajpallares, line 2:
```suggestion
# maestro test -e API_VERSION=36 -e USERNAME=user@example.com -e PASSWORD=123 google-login.yml
```

maestro/google-login/google-login.yml
```diff
@@ -0,0 +1,15 @@
...
+
+appId: "com.android.settings"
+---
+- runFlow:
+    label: "Log in on API >= 36"
```
rickvdl, line 7: This is great 🙌

maestro/google-login/api/32.yml
```diff
@@ -0,0 +1,46 @@
...
+          id: identifierId
+      - inputText: ${USERNAME}
+      - tapOn: NEXT
+      - assertVisible: "Show password"
+      - inputText: ${PASSWORD}
+      - tapOn: NEXT
+      - tapOn: "I agree"
+      - waitForAnimationToEnd
+      # Skip and cancel whatever Google asks us
+      - repeat:
```
rickvdl, line 29: Smart one 😅

PR #2791, opened 2025-10-30
Unify integration tests jobs

tonidero:
### Description
This PR unifies the existing integration test jobs for production and load shedder into a single one, with some internal code in fastlane to decide what environment variables to use. In future PRs, we will change the environment variables to try to unify as much as possible, and only split where needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@
 * Change ForceServerErrorStrategy to be a `fun interface` to simplify instantiation (#2773) via Toni Rico (@tonidero)
 * Stores backend integration test result on CircleCI (#2771) via JayShortway (@JayShortway)
 * Change test server down mechanism to use new server down endpoint (#2765) via Toni Rico (@tonidero)
+* [CAT-2067 EXPERIMENTAL]: Add-On Subscription Options (#2763) via Will Taylor (@fire-at-will)
+* Avoids the `PaywallComponentsTemplate_Preview` from being rendered by Emerge (#2761) via JayShortway (@JayShortway)
+* Handle error building `QueryProductDetailsParams` in some devices (#2764) via Toni Rico (@tonidero)
+* Add more proper error message when using test store API key in release builds (#2762) via Toni Rico (@tonidero)
 
 ## 9.12.0
 > [!WARNING]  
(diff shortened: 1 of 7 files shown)


fastlane/Fastfile
```diff
@@ -581,20 +611,35 @@ DESC
...
-      success: success,
-      default_payloads: [:git_branch]
-    )
+      if success
+        "Android purchases integration tests (#{suite}) finished successfully"
+      else
+        "Android purchases integration tests (#{suite}) failed"
+      end
+
+    slack_url = ENV["SLACK_URL_LOAD_SHEDDER_INTEGRATION_TESTS"]
```
tonidero, line 625: With this change, we would also send slack messages for production integration tests to the slack channel. Open question, should we do this? I think the ones we care about to notify the channel are the nightly ones, which are only run for the load shedder. Aside from that, these tests usually only run on releases.

I don't think it should be too noisy, but we might want to rename the slack channel... thoughts @RevenueCat/coresdk?

JayShortway: I'm okay with sending all integration test results to the slack channel. We can rename the channel. 

Also, I wouldn't be opposed to sending slack updates on release PRs as well. (Regular PRs is too noisy I think.)

tonidero: > I wouldn't be opposed to sending slack updates on release PRs as well

I think that already happens :)

fastlane/Fastfile
```diff
@@ -581,20 +611,35 @@ DESC
...
+        default_payloads: [:git_branch]
+      )
+    else
+      UI.message("Skipping Slack notification for #{suite} suite - no webhook URL configured")
+    end
+  end
+
+  lane :send_slack_load_shedder_integration_test do |options|
+    # Keep for backward compatibility, delegates to unified method
+    send_slack_purchases_integration_test(suite: 'loadshedder', success: options[:success] || false)
```
JayShortway: Could you explain why we need this? I don't see `send_slack_load_shedder_integration_test` being called anywhere 🤔

tonidero: We don't, I actually removed it in a follow-up PR instead of here 😓 https://github.com/RevenueCat/purchases-android/pull/2793. Will remove it here since it's the proper place.

tonidero: Removed here: https://github.com/RevenueCat/purchases-android/pull/2791/commits/03f3df5664eaeef06f3ec294b09bfb512c38fe96

JayShortway: Ah okay great, thanks! I thought I was overlooking something.

.circleci/config.yml
```diff
@@ -984,16 +977,15 @@ workflows:
...
-          requires:
-            - purchases-integration-tests-build
-      - run-firebase-tests-purchases-load-shedder-integration-test:
+      - run-firebase-tests-purchases-integration-tests:
           <<: *release-branches
           context:
             - slack-secrets
+          matrix:
+            parameters:
+              suite: ["production", "loadshedder"]
```
rickvdl, line 988: We ended up going for `backend_environment` for this variable. Ideally it is also updated throughout this PR for consistency

rickvdl (commented):
Nice cleanup, one comment :)

.circleci/config.yml
tonidero: yes! I did the rename in the follow-up PR to avoid conflicts later: https://github.com/RevenueCat/purchases-android/pull/2792. Sorry, forgot to mention it here

rickvdl: Ah, great! Never mind then :)

ajpallares (approved):
Great one! 🚀

PR #2793, opened 2025-10-30
Support fortress us-east-2 integration tests

tonidero:
### Description
Add fortress us-east-2 integration tests

### DIFF
--- a/fastlane/README.md
+++ b/fastlane/README.md
@@ -233,6 +233,14 @@ Builds a Purchase Tester APK and prompts for:
 
 
 
+### android send_slack_load_shedder_integration_test
+
+```sh
+[bundle exec] fastlane android send_slack_purchases_integration_test
+```
+
+
+
 ### android update_paywall_preview_resources_submodule
 
 ```sh


.circleci/config.yml
```diff
@@ -594,7 +594,7 @@ jobs:
     parameters:
       backend_environment:
         type: enum
-        enum: ["production", "loadshedder"]
+        enum: ["production", "loadshedder", "loadshedder-us-east-2"]
```
rickvdl: On the iOS side I've named these cases
```
enum: ["production", "load_shedder_primary", "load_shedder_secondary", "fallback"]
```
Also happy to make the load shedder one more explicit (having the region in the name), but would do the same for both in that case. LMK :)

tonidero: Hmm I think having the region in the name would be more precise... I haven't added fallback yet, since it wouldn't make sense to run most of our current test suite against the fallback URL. For now, I've only added backend integration tests verying the offerings/product entitlement mapping behavior with the fallback URL. But will try adding some minimal full integration tests as well for the fallback URL in a follow-up PR 👍

tonidero: So if you like, we can name them `load_shedder_us_east_1` and `load_shedder_us_east_2`?

rickvdl: Sure, understand! Yes let's do that :) Will update on my side as well

tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/2793/commits/b7ef35e9c84ce5d8798334718c4ec4a1101f3c26

PR #2794, opened 2025-10-31
Add extra request logging

tonidero:
### Description
From time to time, I've found myself wanting to get a curl request for the requests performed by the SDK during a test run, which makes it annoying to get it through the network inspector, since you need to run while debugging to be able to get them to show and it's a bit annoying. In those situations, I've added this small piece of code to log the curl request + full response. I thought it could be useful to have it in the repo but wanted to gather thoughts first. Wdyt?

### DIFF
--- a/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
+++ b/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
@@ -44,8 +44,6 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Before
     fun setup() {
-        Purchases.logLevel = LogLevel.VERBOSE
-
         ensureBlockFinishes { latch ->
             setUpTest {
                 latch.countDown()


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -265,6 +270,14 @@ internal class HTTPClient(
             debugLog { NetworkStrings.API_REQUEST_STARTED.format(connection.requestMethod, path) }
             responseCode = connection.responseCode
             payload = inputStream?.let { readFully(it) }
+            if (ENABLE_EXTRA_REQUEST_LOGGING) {
+                verboseLog { "Request extra logging. Response code: $responseCode. Body: $payload" }
```
tonidero: I thought this could be useful as normal logs... but we could be logging a lot of information, so prefer to leave it behind this flag.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -73,6 +73,7 @@ internal class HTTPClient(
     internal companion object {
         // This will be used when we could not reach the server due to connectivity or any other issues.
         const val NO_STATUS_CODE = -1
+        const val ENABLE_EXTRA_REQUEST_LOGGING = false
```
tonidero: This will be disabled by default, and we can just enable it by modifying this flag.

JayShortway: Should this be a build config field maybe?

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -252,6 +253,10 @@ internal class HTTPClient(
 
             val httpRequest = HTTPRequest(fullURL, headers, jsonBody)
 
+            if (ENABLE_EXTRA_REQUEST_LOGGING) {
+                verboseLog { "Request extra logging. Will perform request with curl: ${toCurlRequest(httpRequest)}" }
```
JayShortway: Maybe something like 

```suggestion
                verboseLog { "HTTP request:\\n ${toCurlRequest(httpRequest)}" }
```
wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
JayShortway: I agree!

JayShortway: What do you think of this? 

```suggestion
                verboseLog { "HTTP response:\\n  status code: $responseCode \\n  body: $payload" }
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -336,6 +349,26 @@ internal class HTTPClient(
...
+            builder.append(entry.value)
+            builder.append("\" \\\n  ")
+        }
+
+        if (httpRequest.body != null) builder.append("-d '").append(httpRequest.body.toString()).append("' \\\n  ")
+
+        builder.append("\"").append(httpRequest.fullURL).append("\"")
+
+        return builder.toString()
+    }
```
JayShortway, line 372: 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -265,6 +270,14 @@ internal class HTTPClient(
             debugLog { NetworkStrings.API_REQUEST_STARTED.format(connection.requestMethod, path) }
             responseCode = connection.responseCode
             payload = inputStream?.let { readFully(it) }
+            if (ENABLE_EXTRA_REQUEST_LOGGING) {
+                verboseLog { "Request extra logging. Response code: $responseCode. Body: $payload" }
+            }
+        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+            if (ENABLE_EXTRA_REQUEST_LOGGING) {
+                verboseLog { "Request extra logging. Exception thrown: $e" }
```
JayShortway: Shouldn't this be an error log? 

```suggestion
                errorLog(e) { "HTTP request failed" }
```

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
```diff
@@ -44,6 +44,8 @@ class PurchasesIntegrationTest : BasePurchasesIntegrationTest() {
 
     @Before
     fun setup() {
+        Purchases.logLevel = LogLevel.VERBOSE
```
JayShortway: Would we be logging anything sensitive if this is enabled? If so, we should be careful with enabling this on CI.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Good idea! That's much cleaner 😅

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/PurchasesIntegrationTest.kt
tonidero: In theory CircleCI should obfuscate any occurences of environment variables in the logs automatically... but yeah, I guess better to be careful. We just would need to remember to enable this if we need to get the logs above locally... Or possibly I could change them to debugLog which should already be logged (as long as the config setting is on). Wdyt?

tonidero: In the end I removed it and I made the logs debugLog so they are printed if the build config is enabled (which shouldn't be in CI)

tonidero: Also changed it to only print on debug builds just in case...

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Made a buildConfig and can be enabled in local.properties. To avoid possible accidents of having this enabled locally and making a release locally, I also added a check to only log in debug builds.

ajpallares (commented):
Wow, really useful indeed. We probably should copy this in iOS 🤓

ajpallares (approved):

PR #2796, opened 2025-10-31
Add internal CustomerInfo source properties

tonidero:
### Description
The idea here is to use this information during tests to make sure the information was obtained from the expected source and try to avoid false positives/negatives during tests.


purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
```diff
@@ -19,6 +20,43 @@ import kotlinx.parcelize.TypeParceler
...
+
+/**
+ * Internal enum representing the current source of CustomerInfo.
+ */
+internal enum class CustomerInfoSource {
+    MAIN,
+    LOAD_SHEDDER,
+    OFFLINE_ENTITLEMENTS,
+    CACHE,
+}
```
ajpallares: I was implementing the same thing in iOS (see https://github.com/RevenueCat/purchases-ios/pull/5737) and I thought that it could be a bit confusing to have the two enums `CustomerInfoSource` and `CustomerInfoOriginalSource`, as they sometimes can be redundant and also require to be in sync (especially when not coming from the cache).

What I did in iOS was to keep `CustomerInfoOriginalSource` as it is here and replace `CustomerInfoSource` with a boolean
```swift
loadedFromCache: Bool
```

This way the two properties are always complementary.

tonidero: Right... That could indeed work and avoid some duplication. I will do the change. Thanks!!

tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/2796/commits/48c793ef54849b0c507cecfee152c484e445c416

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -12,18 +12,21 @@ private const val SERIALIZATION_NAME_PAYLOAD = "payload"
...
     val origin: Origin,
     val requestDate: Date?,
     val verificationResult: VerificationResult,
+    val isLoadShedderResponse: Boolean? = null,
 ) {
     companion object {
         const val ETAG_HEADER_NAME = "X-RevenueCat-ETag"
         const val SIGNATURE_HEADER_NAME = "X-Signature"
         const val REQUEST_TIME_HEADER_NAME = "X-RevenueCat-Request-Time"
+        const val LOAD_SHEDDER_HEADER_NAME = "x-revenuecat-fortress"
```
rickvdl, line 29: Is case-sensitivity a thing here? If it's only used for extracting a value from the http header then it shouldn't be so please ignore me in that case :)

rickvdl (commented):
💪  LGTM code-wise! Maybe @ajpallares can give the final approval to make sure it's aligned with iOS :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
tonidero: It's indeed used only for extracting the value of the response http header, so it doesn't matter, but good callout! Probably good to be consistent and use the same casing as the ones above

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -76,13 +76,15 @@ internal class ETagManager(
         refreshETag: Boolean,
         requestDate: Date?,
         verificationResult: VerificationResult,
+        isLoadShedderResponse: Boolean? = null,
```
ajpallares: When can this be `null`? I ask for two reasons:
1. In case it could just be just `Boolean` (not optional)
2. Because I think we should probably avoid having a default value (to force always passing the parameter;  default values has bitten me in the past sometimes 😅).

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -491,4 +493,12 @@ internal class HTTPClient(
             Date(it)
         }
     }
+    private fun getLoadShedderHeader(connection: URLConnection): Boolean? {
+        val loadShedderHeader = connection.getHeaderField(HTTPResult.LOAD_SHEDDER_HEADER_NAME)
+        return if (loadShedderHeader?.lowercase() == "true") {
+            true
+        } else {
+            null
```
ajpallares: Related to my question before. Maybe this should just be `false` and this way we don't need the optional `Boolean`?

purchases/src/main/kotlin/com/revenuecat/purchases/CustomerInfo.kt
```diff
@@ -248,5 +304,6 @@ private data class ComparableData(
         firstSeen = customerInfo.firstSeen,
         originalAppUserId = customerInfo.originalAppUserId,
         originalPurchaseDate = customerInfo.originalPurchaseDate,
+        // Note: originalSource and source are excluded from equality comparison as they are metadata
```
ajpallares:
```suggestion
        // Note: originalSource and loadedFromCache are excluded from equality comparison as they are metadata
```

purchases/src/test/java/com/revenuecat/purchases/common/CustomerInfoFactoryTest.kt
```diff
@@ -60,4 +62,62 @@ class CustomerInfoFactoryTest {
...
+            VerificationResult.NOT_REQUESTED,
+            isLoadShedderResponse = null,
+        )
+        val customerInfo = CustomerInfoFactory.buildCustomerInfo(httpResult)
+        assertThat(customerInfo.originalSource).isEqualTo(CustomerInfoOriginalSource.MAIN)
+        assertThat(customerInfo.loadedFromCache).isFalse
+    }
+
+    @Test
+    fun `builds CustomerInfo with custom source parameters`() {
```
ajpallares: Nit: I don't understand what "custom source" means here. Maybe this?
```suggestion
    fun `builds CustomerInfo with offline entitlements parameters`() {
```

purchases/src/test/java/com/revenuecat/purchases/common/CustomerInfoTest.kt
```diff
@@ -243,6 +245,23 @@ class CustomerInfoTest {
...
+        assertThat(info.loadedFromCache).isEqualTo(false)
+        val infoWithDifferentMetadata = CustomerInfoFactory.buildCustomerInfo(
+            body = jsonObject,
+            overrideRequestDate = null,
+            verificationResult = VerificationResult.NOT_REQUESTED,
+            originalSource = CustomerInfoOriginalSource.LOAD_SHEDDER,
+            loadedFromCache = true,
+        )
+
+        assertThat(info).isEqualTo(infoWithDifferentMetadata)
```
ajpallares, line 264: Maybe my biggest nit (I don't know if I can't go further 😆), but, in case we changed default values in the future for `CustomerInfoFactory.buildCustomerInfo`, to keep having this test be true ("with different original source and loadedFromCache"), maybe we should either skip the test if both `loadedFromCache`s are equal or if both `originalSource`s are equal, or perhaps just assert for them being different.

ajpallares (approved):
Ups, forgot to approve in my review! 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
tonidero: Yeah makes sense, changed in https://github.com/RevenueCat/purchases-android/pull/2796/commits/79d67f8eb817dc449ff080f99600017814d31996

PR #2797, opened 2025-10-31
Normalize product ids

vegaro:
`getProducts` would return an empty list when passed product IDs with base plans (e.g., `connect:connect-monthly`). This happened because Google Play Billing only accepts the product ID without the base plan part.

We got this report in React Native https://github.com/RevenueCat/react-native-purchases/issues/989 and others in Flutter. This is particularly problematic because the CustomerInfo has the base plan in the ids, so when devs query those ids directly, it will fail.

With the changes in this PR, we return only the particular base plan ids that are queried


tonidero (commented):
Code looks great 🙌 . Seems some tests are failing but not much to add

tonidero (approved):

PR #2799, opened 2025-11-03
Check if maven has published current version before deploying

tonidero:
### Description
This adds some checks to the deploy job to make sure we don't start deploying if the version trying to be deployed has already been fully or partially deployed.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -201,52 +201,51 @@ platform :android do
   end
 
   def check_version_not_in_maven_central(version, group_id, artifact_ids)
-    base_url = "https://repo1.maven.org/maven2"
+    require 'rest-client'
+    require 'json'
+    require 'uri'
+    
+    auth_token = ENV["FETCH_PUBLICATIONS_USER_TOKEN_MAVEN_CENTRAL"]
+    if auth_token.nil? || auth_token.empty?
+      UI.user_error!("FETCH_PUBLICATIONS_USER_TOKEN_MAVEN_CENTRAL environment variable is not set. Please provide a valid token to check Maven Central publications.")
+    end
+    
+    base_url = "https://central.sonatype.com/api/v1/publisher/published"
     existing_artifacts = []
     
     artifact_ids.each do |artifact_id|
-      # Convert group ID dots to slashes for URL path
-      group_path = group_id.gsub(".", "/")
-      url = "#{base_url}/#{group_path}/#{artifact_id}/#{version}/"
+      # Build query parameters for the API endpoint with proper URI encoding
+      # The API uses 'name' parameter which corresponds to the artifact_id
+      api_url = "#{base_url}?namespace=#{URI.encode_www_form_component(group_id)}&name=#{URI.encode_www_form_component(artifact_id)}&version=#{URI.encode_www_form_component(version)}"
       
-      UI.verbose("Checking Maven Central: #{url}")
+      UI.verbose("Checking Sonatype API for publication status: #{group_id}:#{artifact_id}:#{version}")
       
-      # Use curl to check if the version directory exists (HTTP HEAD request)
-      # Don't use -f flag so we can check the HTTP status code
-      # Use -L to follow redirects
       begin
-        # Capture both stdout (HTTP code) and stderr separately
-        result = sh("curl -s -S -o /dev/null -w '%{http_code}' -L -I '#{url}' 2>&1", log: false).strip
+        response = RestClient.get(
+          api_url,
+          {
+            'Authorization' => "Bearer #{auth_token}",
+            'accept' => 'application/json'
+          }
+        )
         
-        # Extract the HTTP status code (last 3 digits at the end of the output)
-        match_result = result.match(/(\d{3})\s*$/)
-        if match_result.nil?
-          UI.important("Could not parse HTTP status code for #{url}. Response: #{result}")
-          next
-        end
+        # Parse JSON response
+        response_data = JSON.parse(response.body)
         
-        http_code = match_result[1]
-        
-        if http_code == "200"
+        # Check if published field is true
+        if response_data["published"] == true
           existing_artifacts << artifact_id
           UI.important("Artifact #{group_id}:#{artifact_id}:#{version} already exists in Maven Central")
-        elsif http_code != "404"
-          # If we get an unexpected status code, warn but don't fail
-          UI.user_error!("Unexpected HTTP status #{http_code} when checking #{url}")
         end
       rescue => e
-        UI.user_error!("Failed to check #{url}: #{e.message}")
-        # Continue checking other artifacts even if one fails
-        next
+        UI.user_error!("Failed to check #{group_id}:#{artifact_id}:#{version}: #{e.message}")
       end
     end
     
     unless existing_artifacts.empty?
       error_message = "Version #{version} already exists in Maven Central for the following artifacts:\n"
       existing_artifacts.each do |artifact_id|
-        group_path = group_id.gsub(".", "/")
-        maven_url = "#{base_url}/#{group_path}/#{artifact_id}/#{version}/"
-        error_message += "  - #{group_id}:#{artifact_id}:#{version}\n    #{maven_url}\n"
+        error_message += "  - #{group_id}:#{artifact_id}:#{version}\n"
       end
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -105,6 +105,24 @@ platform :android do
     )
   end
 
+  desc "Check if version already exists in Maven Central"
+  lane :validate_version_not_in_maven_central do |options|
+    version = options[:version] || current_version_number
+    UI.message("Checking if version #{version} already exists in Maven Central...")
+    
+    group_id = "com.revenuecat.purchases"
```
tonidero, line 113: I'm assuming the group id, which should be fine for this repo.

build.gradle.kts
```diff
@@ -79,3 +79,41 @@ tasks.named<org.jetbrains.dokka.gradle.DokkaMultiModuleTask>("dokkaHtmlMultiModu
     outputDirectory.set(file("docs/${project.property("VERSION_NAME")}"))
     includes.from("README.md")
 }
+
+tasks.register("listPublications") {
```
tonidero, line 83: Honestly not sure if it's worth the extra complexity here. The idea is to be able to know what we publish to maven dynamically so, if we add more modules we publish in the future, we don't need to update any hardcoded list.

build.gradle.kts
```diff
@@ -79,3 +79,41 @@ tasks.named<org.jetbrains.dokka.gradle.DokkaMultiModuleTask>("dokkaHtmlMultiModu
...
+        val version = project.findProperty("VERSION_NAME") as String? ?: project.version.toString()
+        val pomArtifactId = project.findProperty("POM_ARTIFACT_ID") as String?
+
+        subprojects.forEach { subproject ->
+            // Check if maven publish plugin is applied
+            val hasMavenPublish = subproject.plugins.hasPlugin("com.vanniktech.maven.publish")
+
+            if (hasMavenPublish) {
+                // Get the publishing extension
+                val publishing = subproject.extensions.findByType<org.gradle.api.publish.PublishingExtension>()
```
JayShortway: Do we need to check `hasMavenPublish`? Can we not just try to get the `PublishingExtension` right away? If that fails there's no publication. 

Reason for asking is because this tightly couples this task to the specific publish plugin we're using. It's unlikely to change, but would be nice if we can avoid that coupling easily.

build.gradle.kts
JayShortway: I think this is nice!

fastlane/Fastfile
```diff
@@ -105,6 +105,24 @@ platform :android do
     )
   end
 
+  desc "Check if version already exists in Maven Central"
+  lane :validate_version_not_in_maven_central do |options|
```
JayShortway, line 109: Are we going to move this to the Fastlane plugin? That way we can reuse for PHC and KMP.

fastlane/Fastfile
```diff
@@ -143,6 +165,94 @@ platform :android do
...
+  end
+
+  def check_version_not_in_maven_central(version, group_id, artifact_ids)
+    base_url = "https://repo1.maven.org/maven2"
+    existing_artifacts = []
+    
+    artifact_ids.each do |artifact_id|
+      # Convert group ID dots to slashes for URL path
+      group_path = group_id.gsub(".", "/")
+      url = "#{base_url}/#{group_path}/#{artifact_id}/#{version}/"
```
JayShortway: This is an improvement but it's not airtight, given the publishing delay Maven Central has after uploading. 

Maybe the `/api/v1/publisher/published` endpoint would not have that delay? As it will return the "published-status", which I'm assuming to mean it will return `false` if the package was uploaded but not yet published. (And something like 404 if the package is not uploaded.)

- [API docs](https://central.sonatype.org/publish/publish-portal-api/)
- [OpenAPI spec](https://central.sonatype.com/api-doc)

build.gradle.kts
JayShortway, line 1: (This comment is not related to this file at all 😄) 

Another idea I had was to separate the upload step from the publish step, ([plugin docs](https://vanniktech.github.io/gradle-maven-publish-plugin/central/#uploading-with-manual-publishing)), and only publish at the end, once we verify that all packages have been uploaded. (And if not, we could e.g. drop whatever we did upload to try again.)

We can publish (and drop) uploaded packages using [the API](https://central.sonatype.org/publish/publish-portal-api/#publish-or-drop-the-deployment).

fastlane/Fastfile
tonidero: Indeed, I think it will be needed there as well. Will try to move it there instead

Edit: Waiting until it's ready to have all the reviews in one place. Once it's ready, I will move to the plugin.

build.gradle.kts
tonidero: Hmm indeed I think it's another valid approach... But I guess there is still a chance of the process being stopped midway in this approach when publishing right? Not sure if having all the artifacts uploaded would be enough for the plugin to fail on the next attempt though...

fastlane/Fastfile
tonidero: I was checking these APIs, it seems we need a deployment ID to query this way, which might not be straightforward to obtain since it's only obtained when uploading it seems, and considering we're using a plugin for that, we would need to cache this somewhere or change the way we deploy things, which seems overkill...

Regarding the possible delays, I think indeed there could be delays, but not sure how long they could be... I guess the problem could be if in that timeframe the tag is removed, and a new deploy started, which seems like a small chance, and this still feels like an improvement over what we had before.

But TBH, I'm just wondering again if we should just block deleting tags except by just a few that know what's happening... If something broke midway, might be worth to just do another release including the changes from the previous one... Wdyt @RevenueCat/coresdk ?

JayShortway: `/api/v1/publisher/published` doesn't seem to need a deploymentId: 

afbeelding  
  
---------    

Regarding the delay, [docs for the vanniktech plugin](https://vanniktech.github.io/gradle-maven-publish-plugin/central/#publishing-releases) state: 
  

    
>Step 4 can take **10 to 30 minutes** and only after it completed the published artifacts will be available for download.


I think it would be a good idea to only allow removing tags by CoreSDK. Does GitHub have a role for that? I'd assume everyone needs the Write role.

build.gradle.kts
JayShortway: >there is still a chance of the process being stopped midway in this approach when publishing

Hmm good point, but this should be much faster (seconds vs minutes), so the chance is very low I'd say.

>Not sure if having all the artifacts uploaded would be enough for the plugin to fail on the next attempt though...

If we (are able to) check uploaded artifacts, we could make the release job fail if it finds any.

fastlane/Fastfile
tonidero: Thanks! I was looking at the wrong place 😅 . Changed to use the API you mentioned [here](https://github.com/RevenueCat/purchases-android/pull/2799/commits/52461406682b761a072266b30a7153d57bb5e2ac). We should still consider the other option you mentioned, added to our sync.

fastlane/Fastfile
```diff
@@ -143,6 +165,93 @@ platform :android do
...
+  end
+
+  def check_version_not_in_maven_central(version, group_id, artifact_ids)
+    require 'rest-client'
+    require 'json'
+    require 'uri'
+    
+    auth_token = ENV["FETCH_PUBLICATIONS_USER_TOKEN_MAVEN_CENTRAL"]
+    if auth_token.nil? || auth_token.empty?
+      UI.user_error!("FETCH_PUBLICATIONS_USER_TOKEN_MAVEN_CENTRAL environment variable is not set. Please provide a valid token to check Maven Central publications.")
```
tonidero, line 210: This has already been set in CircleCI android (need to add it in PHC and KMP once we move this to the plugin)

vegaro (approved):
I think this makes sense. It won't prevent race conditions (imagine another job in parallel actually publishing) but I don't think we can do much about it

fastlane/Fastfile
tonidero: I extracted part of the logic here to the plugin: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/101. We would still need to keep some logic to get the artifacts to check... but this felt more like a gradle task, and separate from the plugin action... I guess we could assume the gradle command would be available in all SDKs using it but preferred to keep it separate for now. lmk what you think there! (still need to adapt this PR to use that lane from the plugin)

PR #2802, opened 2025-11-04
Add internal `Offerings` and `ProductEntitlementMapping` source information

tonidero:
### Description
This adds source information, similar to what was done in #2796 but for `Offerings` and `ProductEntitlementMapping`. I kept the source enums different, since these have different values than CustomerInfo. Same as in the previous PR, these will be used to improve our testing.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsCacheTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/offerings/OfferingsCacheTest.kt
@@ -8,6 +8,8 @@ import com.revenuecat.purchases.common.FakeLocaleProvider
 import com.revenuecat.purchases.common.HTTPResponseOriginalSource
 import com.revenuecat.purchases.common.caching.DeviceCache
 import com.revenuecat.purchases.utils.add
+import com.revenuecat.purchases.utils.copy
+import io.mockk.InternalPlatformDsl.toArray
 import io.mockk.Runs
 import io.mockk.every
 import io.mockk.just
@@ -72,7 +74,13 @@ class OfferingsCacheTest {
         assertThat(offeringsCache.cachedOfferings).isNull()
         offeringsCache.cacheOfferings(offerings, offeringsResponse)
         assertThat(offeringsCache.cachedOfferings).isEqualTo(offerings)
-        verify(exactly = 1) { deviceCache.cacheOfferingsResponse(offeringsResponse) }
+        verify(exactly = 1) {
+            deviceCache.cacheOfferingsResponse(
+                match {
+                    it.getString(OfferingsCache.ORIGINAL_SOURCE_KEY) == HTTPResponseOriginalSource.MAIN.name
+                }
+            )
+        }
     }
 
     // region offerings cache


purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
```diff
@@ -46,7 +46,11 @@ class CachedOfferingsUsageIntegrationTest : BasePurchasesIntegrationTest() {
 
         val cachedOfferings = Purchases.sharedInstance.awaitOfferings()
 
-        assertThat(cachedOfferings).isEqualTo(networkOfferings)
+        val networkOfferingsWithCacheOrigin = networkOfferings.copy(
+            loadedFromCache = true,
+        )
+
+        assertThat(cachedOfferings).isEqualTo(networkOfferingsWithCacheOrigin)
```
ajpallares: Does this mean that `loadedFromCache` is also required to be equal for two Offerings to be equal? (I think it should not affect for equality, as discussed [here](https://github.com/RevenueCat/purchases-ios/pull/5749#discussion_r2495535597))

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
```diff
@@ -67,7 +71,11 @@ class CachedOfferingsUsageIntegrationTest : BasePurchasesIntegrationTest() {
...
```
ajpallares: Same question as before

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
ajpallares: Note that in iOS, the `Offerings` object is not `Equatable` anyway

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -94,6 +96,8 @@ internal class ETagManager(
                         // when enabling verification.
                         verificationResult = verificationResult,
                         requestDate = requestDate ?: storedResult.requestDate,
+                        // Preserve fallback flag from stored result
+                        isFallbackURL = storedResult.isFallbackURL,
```
ajpallares: Would this be equivalent to just not adding this `isFallbackURL` parameter in the `copy` method call?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -21,6 +22,7 @@ internal data class HTTPResult(
     val requestDate: Date?,
     val verificationResult: VerificationResult,
     val isLoadShedderResponse: Boolean? = null,
+    val isFallbackURL: Boolean? = null,
```
ajpallares: Similar question as in the other PR. Should it be nullable or can it just be `Boolean` (without "?")?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -52,7 +54,20 @@ internal data class HTTPResult(
             } else {
                 null
             }
-            return HTTPResult(responseCode, payload, origin, requestDate, verificationResult, isLoadShedderResponse)
+            val isFallbackURL: Boolean? = if (jsonObject.has(SERIALIZATION_NAME_IS_FALLBACK_URL)) {
+                jsonObject.getBoolean(SERIALIZATION_NAME_IS_FALLBACK_URL)
+            } else {
+                null
```
ajpallares: Related to my previous comment: and this would be just `false`. Mostly I'm afraid that we have `false` and `null` mean the same here

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
```diff
@@ -24,6 +25,8 @@ internal class OfferingsFactory(
     @SuppressWarnings("TooGenericExceptionCaught", "LongMethod")
     fun createOfferings(
         offeringsJSON: JSONObject,
+        originalDataSource: HTTPResponseOriginalSource,
+        loadedFromCache: Boolean,
```
ajpallares: Here I realized that, in iOS, we have the in-memory cache for `Offerings`. That's why I called it `loadedFromDiskCache`. I'm not seeing such behavior in Android, in which case this name is OK 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMapping.kt
```diff
@@ -1,19 +1,29 @@
...
         private const val BASE_PLAN_ID_KEY = "base_plan_id"
         private const val ENTITLEMENTS_KEY = "entitlements"
+        private const val ORIGINAL_SOURCE_KEY = "rc_original_source"
 
-        fun fromJson(json: JSONObject): ProductEntitlementMapping {
+        fun fromJson(
+            json: JSONObject,
+            originalSource: HTTPResponseOriginalSource = HTTPResponseOriginalSource.MAIN,
+            loadedFromCache: Boolean = false,
+        ): ProductEntitlementMapping {
```
ajpallares, line 27: I'm a bit confused here because `fromJson` requires passing `originalSource` when the `toJson()` below does put the `originalSource` into the JSON. Is it correct? or are we losing the `originalSource` stored in the JSON?

ajpallares (commented):
Some initial comments. I'll keep reviewing this! But amazing job (based on what I've seen so far) 👏

purchases/src/main/kotlin/com/revenuecat/purchases/Offerings.kt
```diff
@@ -70,6 +73,57 @@ class Offerings internal constructor(
...
+    val targeting: Offerings.Targeting?,
+) {
+    constructor(
+        offerings: Offerings,
+    ) : this(
+        current = offerings.current,
+        all = offerings.all,
+        placements = offerings.placements,
+        targeting = offerings.targeting,
+        // Note: originalSource and source are excluded from equality comparison as they are metadata
```
ajpallares:
```suggestion
        // Note: originalSource and loadedFromCache are excluded from equality comparison as they are metadata
```

purchases/src/test/java/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMappingSourceTest.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        // Cache the mapping
+        deviceCache.cacheProductEntitlementMapping(originalMapping)
+
+        // Mock retrieval with preserved originalSource
+        every { deviceCache.getProductEntitlementMapping() } returns ProductEntitlementMapping.fromJson(
+            sampleResponseJson,
+            HTTPResponseOriginalSource.FALLBACK,
+            loadedFromCache = true,
+        )
```
ajpallares, line 140: The test name seems to imply that we're testing the cache behavior. But that's being mocked and what is being tested here is actually `ProductEntitlementMapping.fromJson`. And also, I'm confused (maybe it relates to my comment above): the original source initially in `httpResult` is gotten from `isFallbackURL = true`. But the we pass `HTTPResponseOriginalSource.FALLBACK` to the `fromJson` method. That means that we're actually ignoring the saved information stored for the `ORIGINAL_SOURCE_KEY`.

Or am I fixated with this and I'm totally wrong? (maybe I'm not thinking clearly anymore today; I should stop soon 😅)

ajpallares (commented):
Kept reviewing, and went back to a previous comment that I'd love to clarify 🙏

purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/cachedofferings/CachedOfferingsUsageIntegrationTest.kt
tonidero: > Does this mean that loadedFromCache is also required to be equal for two Offerings to be equal? (I think it should not affect for equality, as discussed https://github.com/RevenueCat/purchases-ios/pull/5749#discussion_r2495535597)

It wasn't actually needed, removed here: https://github.com/RevenueCat/purchases-android/pull/2802/commits/d8ac213ad763e566a95217ed165b015df44e9de0

tonidero: > Note that in iOS, the Offerings object is not Equatable anyway

In Android it is, but we indeed don't take into account this new data to perform the comparison

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
tonidero: Yup 🤦 Removed

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsFactory.kt
tonidero: Good point, in Android we also have an in memory cache for offerings, so renaming this makes sense, thanks!

tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/2802/commits/494e19753970823c326d265cd3d6cbcc620040a5

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMapping.kt
tonidero: Yeah, it's indeed confusing... Maybe it would have been worth to refactor it to store an object containing both the product entitlement mapping response + source instead, but wanted to avoid dealing with the migration... For now, I've refactored this a bit [here](https://github.com/RevenueCat/purchases-android/pull/2802/commits/b1ba302dad2cf45c85c2d3e1b26900f54cb27da3), though I noticed I had to create a copy of the json to avoid possible unexpected side effects. I believe it shouldn't be a performance hit the way we're doing it now. Lmk what you think!

purchases/src/test/java/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMappingSourceTest.kt
tonidero: Yeah, this test was assuming things before, which was wrong... It did work before since we were reading the value from the cached JSON beforehand, before passing it back to this method. In any case, I've refactored it which should be a bit cleaner. Lmk what you think! (Also I renamed the test 👍 )

ajpallares (approved):
Thank you so much for addressing all the comments! Amazing! 🙌

PR #2807, opened 2025-11-05
Add client side timeout logic for endpoints that support fallback URLs

tonidero:
### Description
This adds some logic to implement a client side connection timeout. The logic is like this:
- If the request supports fallback URLs and we're attempting the "main" backend, timeout will be 5s.
- If the request is performed to a fallback URL or does not support fallbacks, timeout will be 30s
- Once we timeout in one of the endpoints that support fallbacks, the new timeout for these endpoints that support timeouts in the main server will be 2s, so we can fail early and try the fallback.
- All requests will always hit the "main" backend first.
- Whenever the main server is hit successfully, the timeout is reset to the initial 5s for the endpoints that support fallbacks

This is pretty delicate, so lmk what you think!


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.util.concurrent.atomic.AtomicLong
+
+internal class HTTPTimeoutManager(
+    private val appConfig: AppConfig,
+    private val dateProvider: DateProvider = DefaultDateProvider(),
+) {
+    companion object {
+        const val SUPPORTED_FALLBACK_TIMEOUT_MS = 5000L // 5 seconds for requests with fallback support
+        const val REDUCED_TIMEOUT_MS = 2000L // 2 seconds for requests with fallback support after timeout
+        const val DEFAULT_TIMEOUT_MS = 30000L // 30 seconds for requests without fallback support and fallback requests
```
tonidero, line 20: This is pretty long... but right now we aren't setting anything, and this felt pretty safe IMO.

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
```diff
@@ -982,6 +985,186 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
             )
         }
     }
+
+    // region Timeout Management
+
+    @Test
+    fun `HTTPClient records SUCCESS_ON_MAIN_BACKEND when successful request to main backend`() {
```
rickvdl, line 1012: Great set of tests as well! 

I think ideally we also mock the HTTPTimeoutManager and assert on the recorded events.I think these tests are very useful, but they assert on the behavior of the timeout manager, assuming it recorded what we think it recorded.

I think there could be a small role for this mock, making 100% sure it recorded the right event. What do you think?

purchases/src/test/java/com/revenuecat/purchases/common/networking/HTTPTimeoutManagerTest.kt
```diff
@@ -0,0 +1,274 @@
...
+    fun setUp() {
+        appConfig = mockk<AppConfig>().apply {
+            every { runningTests } returns false
+        }
+        dateProvider = FakeDateProvider()
+        timeoutManager = HTTPTimeoutManager(appConfig, dateProvider)
+    }
+
+    @Test
+    fun `getTimeoutForRequest returns SUPPORTED_FALLBACK_TIMEOUT_MS for endpoints with fallback support when no timeout occurred`() {
```
rickvdl, line 48: Awesome set of tests! Took some inspiration from them :)

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
rickvdl: Added some on the iOS side [here](https://github.com/RevenueCat/purchases-ios/pull/5760/commits/d22950ed4a2a91d40239ddce4991e4f67910b3f8)

tonidero: Right, I've wanted to avoid mocking things in tests when possible for a long time, and test a more comprehensive behavior. Since this is relatively coupled, I thought we could use the real dependency, but I think having some extra tests that mock it and verify that calls happen as expected, does make sense as well. I will add those aside from the already added ones 👍  Thanks for the feedback!

tonidero: In the end what I did was use spies and reuse the existing tests, so I could just verify the correct calls were being made: https://github.com/RevenueCat/purchases-android/pull/2807/commits/47a8212045b549800354209d6443887917df9061. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.revenuecat.purchases.common.DateProvider
+import com.revenuecat.purchases.common.DefaultDateProvider
+import java.util.concurrent.atomic.AtomicLong
+
+internal class HTTPTimeoutManager(
+    private val appConfig: AppConfig,
+    private val dateProvider: DateProvider = DefaultDateProvider(),
+) {
+    companion object {
+        const val READ_TIMEOUT_MS = 30000L // We always use 30 seconds for read timeout, once we have established a connection
```
tonidero: I decided to change the read timeout to 30 seconds always. I feel that if it's a connectivity issue with the backend (because it's queuing or something like that), it would be catched during the connect timeout, which we can keep at 5s for endpoints supporting fallback. But for the read timeout, that could be an indication of slow connections, so I think it might be better to be more conservative. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
ajpallares: I agree. I'd be more worried about the `SUPPORTED_FALLBACK_TIMEOUT_MS`. I wonder if 5 seconds is too short?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * Records the result of a request and updates internal state accordingly.
+     * @param result The result of the request
+     */
+    fun recordRequestResult(result: RequestResult) {
+        when (result) {
+            RequestResult.SUCCESS_ON_MAIN_BACKEND -> {
+                resetTimeout()
```
ajpallares, line 77: Nice! I was exactly looking for this! Makes total sense 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -453,8 +474,10 @@ internal class HTTPClient(
             .filterNotNullValues()
     }
 
-    private fun getConnection(request: HTTPRequest): HttpURLConnection {
+    private fun getConnection(request: HTTPRequest, timeoutMs: Long): HttpURLConnection {
         return (request.fullURL.openConnection() as HttpURLConnection).apply {
+            connectTimeout = timeoutMs.toInt()
+            readTimeout = timeoutManager.getReadTimeout().toInt()
```
ajpallares: I'm curious. Do we need to set this because the default was very different? Or is it for tests?

purchases/src/test/java/com/revenuecat/purchases/common/networking/HTTPTimeoutManagerTest.kt
```diff
@@ -0,0 +1,280 @@
...
+        val endpoint = Endpoint.GetOfferings("test_user_id")
+        assert(endpoint.supportsFallbackBaseURLs)
+
+        val timeout = timeoutManager.getTimeoutForRequest(endpoint, isFallback = false)
+        assertThat(timeout).isEqualTo(HTTPTimeoutManager.SUPPORTED_FALLBACK_TIMEOUT_MS)
```
ajpallares, line 53: Why not `HTTPTimeoutManager.SUPPORTED_FALLBACK_TIMEOUT_MS / 100` here? 🤔

ajpallares: Nevermind
```kotlin
        appConfig = mockk<AppConfig>().apply {
            every { runningTests } returns false
        }
```

ajpallares (approved):
Wow! This is a very clean implementation and it was a great idea to separate it into its own object. And as Rick mentioned, great set of tests 👌 
Thanks for being so thorough! 🙌 

I have some small comments.

The first thing that came to mind was when reading the description.
> Once we timeout in one of the endpoints that support fallbacks, the new timeout for these endpoints that support timeouts in the main server will be 2s, so we can fail early and try the fallback.

I later realized it is indeed implemented, but probably I'd add that

> Whenever the main server is hit successfully, the timeout is reset to the initial 5s for the endpoints that support fallbacks

Other than that, some small comments! Thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: No, the default is the same (no timeout), but I think it's safer to have a bigger read timeout, which means that the connection was established with the backend, but it might just take longer to download the response, especially for big responses like offerings. I think if the connection was established, the chances of it being a backend issue are much smaller.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -86,9 +90,45 @@ internal class HTTPClient(
...
+     *
+     * @param inputStream The input stream to read from
+     * @param timeoutMs The maximum time in milliseconds to wait for the entire read operation
+     * @throws SocketTimeoutException if the read operation exceeds the timeout
+     * @throws IOException for other IO errors
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     @Throws(IOException::class)
-    private fun readFully(inputStream: InputStream): String {
-        return buffer(inputStream).readText()
```
tonidero, line 91: I changed this as well after more testing on this PR. Reason is that the `readText` method we were using, may extend beyond the given read timeout for very large responses (i.e. offerings). It seems like it's reading in chunks and applying the read timeout to each chunk, so it can take ~2-3min to timeout when simulating bad connectivity settings. 

The new approach takes more control and makes sure we stop reading in, at worst 60 seconds (30 seconds for the read timeout of the current chunk + 30 seconds for the total timeout while reading since the last chunk)

Would appreciate a double check on this behavior @JayShortway @vegaro

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
rickvdl: Looks great @tonidero 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: We discussed it during our sync up today. In the end, we decided to leave the read timeout default (infinite), and just add a connection timeout, to avoid changing the behavior unnecessarily. Please lmk if you have any concerns. cc @RevenueCat/coin

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -453,8 +474,10 @@ internal class HTTPClient(
             .filterNotNullValues()
     }
 
-    private fun getConnection(request: HTTPRequest): HttpURLConnection {
+    private fun getConnection(request: HTTPRequest, timeoutMs: Long): HttpURLConnection {
         return (request.fullURL.openConnection() as HttpURLConnection).apply {
+            connectTimeout = timeoutMs.toInt()
+            // readTimeout = 0 We let the read timeout to the default, which is infinite.
```
ajpallares: This comment could be confusing. Because the beginning "readtimeout = 0" could seem like we commented the code by mistake?

I'd probably say:
```suggestion
            // We leave the read timeout to the default (readTimeout = 0), which means infinite.
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -0,0 +1,107 @@
...
+     * Calculates the timeout for a request based on the endpoint and whether it's a fallback call.
+     * @param endpoint The endpoint being requested
+     * @param isFallback Whether this is a fallback request
+     * @return The timeout in milliseconds
+     */
+    fun getTimeoutForRequest(endpoint: Endpoint, isFallback: Boolean): Long {
+        // Check if reset is needed (10 minutes elapsed)
+        if (shouldResetTimeout()) {
+            resetTimeout()
+        }
```
JayShortway, line 54: If my mental compiler is correct I guess we could have 2 threads both resetting the time out, if the second thread reads `lastTimeoutRequestTime` before the first resets it, but that's probably fine?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -0,0 +1,107 @@
...
+    enum class RequestResult {
+        /**
+         * Request succeeded on the main backend
+         */
+        SUCCESS_ON_MAIN_BACKEND,
+
+        /**
+         * Request timed out on the main backend and supports fallback URLs
+         */
+        TIMEOUT_ON_MAIN_BACKEND_WITH_FALLBACK,
```
JayShortway: I had to look at the implementation to understand what this exactly means. Maybe we can name it something like this?

```suggestion
        TIMEOUT_ON_MAIN_BACKEND_WILL_USE_FALLBACK,
```

or


```suggestion
        WILL_USE_FALLBACK_AFTER_TIMEOUT_ON_MAIN_BACKEND,
```

JayShortway (approved):
Great stuff!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
tonidero: Indeed, that would be possible. But as you said, it wouldn't be a problem. We would just be setting to 0 multiple times. Great catch though!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
tonidero: Right... The idea here is that we only enter this "lowered timeout" state, if we hit a timeout on one of the endpoints that supports fallback. So, if we hit a timeout in, for example, POST /receipt, we won't lower the timeout. That's the reason for the name.

I think your suggestions could work, but I don't like mixing what will happen with the event itself... Maybe we could rename to something like `TIMEOUT_ON_FALLBACK_SUPPORTED_ENDPOINT_ON_MAIN_BACKEND`. That's quite a mouthful though 😓. Wdyt @JayShortway ?

JayShortway: Good points! How about `TIMEOUT_ON_MAIN_BACKEND_FOR_FALLBACK_SUPPORTED_ENDPOINT`? It's a mouthful, but I think it covers it clearly.

tonidero: Like that! Will do the rename. cc @rickvdl

tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/2807/commits/b566181f28b9c36f3e959aeb2c0f9c1c44eb3f8f

rickvdl: Nice one, will make the change as well

PR #2809, opened 2025-11-05
Adds `TestStoreErrorDialogActivity`

JayShortway:
## Description
This is an activity and not a dialog because a dialog can't be shown with an Application context, which is what we have in `PurchasesFactory`.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -438,7 +438,7 @@ internal class PurchasesFactory(
                     error = PurchasesError(
                         code = PurchasesErrorCode.ConfigurationError,
                         underlyingErrorMessage = "Test Store API key used in release build. Please configure the " +
-                            "Play Store/Amazon app on the RevenueCat dashboard and use its corresponding Google API key " +
+                            "Play Store/Amazon app on the RevenueCat dashboard and use its corresponding API key " +
                             "before releasing. Visit https://rev.cat/sdk-test-store to learn more.",
                     ),
                 )


purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
```diff
@@ -0,0 +1,48 @@
+package com.revenuecat.purchases
+
+import android.app.Activity
+import android.app.AlertDialog
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+
+internal class TestStoreErrorDialogActivity : Activity() {
```
JayShortway, line 9: This is an activity and not a dialog because Claude tells me a dialog can't be shown with an Application context, which is what we have in `PurchasesFactory`. I did not verify that though.

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        
+        setFinishOnTouchOutside(false)
+        
+        AlertDialog.Builder(this)
+            .setTitle("Configuration Error")
+            .setMessage("Test Store API key detected in release build. Please configure the Play Store/Amazon store app on the RevenueCat dashboard and use its corresponding API key before releasing.")
```
tonidero: Probably worth also displaying this as part of the logs? Possibly before displaying the activity even.

tonidero: Well... it would show in the crash after clicking in the alert I guess... But not sure if we want to show a log before that as well... maybe it's ok, not a strong opinion

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
JayShortway: I verified this now. Showing a Dialog directly from an Application context crashes due to a missing window token. So using an Activity seems to be the right approach.

tonidero: One small concern I have over this... If the app is initiated and configured in the background (from the application), this would initiate UI right? I guess it might be initiated from a notification or something, and start the app suddenly 🥴

JayShortway: Yea true, and that will likely also crash because showing UI from a background process is no longer allowed. However, I'm thinking this is okay for now? This is an "incorrect configuration" scenario after all.

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
JayShortway: Good point. Added some logs.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -434,14 +434,17 @@ internal class PurchasesFactory(
             if (!isDebugBuild() &&
                 apiKeyValidationResult == APIKeyValidator.ValidationResult.SIMULATED_STORE
             ) {
-                throw PurchasesException(
+                errorLog(
                     error = PurchasesError(
                         code = PurchasesErrorCode.ConfigurationError,
+                        underlyingErrorMessage = "Test Store API key used in release build. Please configure the " +
+                            "Play Store app on the RevenueCat dashboard and use its corresponding Google API key " +
```
ajpallares:
```suggestion
                            "Play Store/Amazon app on the RevenueCat dashboard and use its corresponding Google API key " +
```
(it was mentioned before this change)

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun onPause() {
+        super.onPause()
+        crashApp()
+    }
+
+    private fun crashApp() {
+        throw PurchasesException(
+            error = PurchasesError(code = PurchasesErrorCode.ConfigurationError),
+            overridenMessage = "Test Store API key used in release build. Please configure the " +
+                "Play Store app on the RevenueCat dashboard and use its corresponding Google API key " +
```
ajpallares:
```suggestion
                "Play Store/Amazon app on the RevenueCat dashboard and use its corresponding Google API key " +
```

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
JayShortway: Oh of course!

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
tonidero: Good point. As long as the crash makes sense I’m ok with that!

PR #2810, opened 2025-11-05
CC-581 | Allow for support ticket creation

RosieWatson:
### Motivation

This will allow customers to create a support ticket in order to contact the developer, rather than directly emailing the developer.

### Description

- If the correct config is passed through (from enabling support tickets in the dashboard) the "Contact us" button will no longer open the native mail client, but will instead show a new screen. This screen allows customer to give their email and a description of the issue. This will then be passed on to the developer from the backend.

I tested my changes locally (emulator only, due to no Android device) both against my local API and the production API, using existing CC configs and some with the new changes. I also added unit tests for the new screen.

[iOS PR](https://github.com/RevenueCat/purchases-ios/pull/5779)


purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -215,3 +215,26 @@ suspend fun Purchases.awaitStorefrontLocale(): Locale {
...
+ *
+ * @param email The user's email address for the support ticket.
+ * @param description The description of the support request.
+ * @return Boolean indicating whether the ticket was successfully sent.
+ * @throws [PurchasesException] with a [PurchasesError] if there's an error creating the support ticket.
+ */
+@JvmSynthetic
+@Throws(PurchasesException::class)
+@InternalRevenueCatAPI
+suspend fun Purchases.awaitCreateSupportTicket(email: String, description: String): Boolean {
```
tonidero: Not sure if there are other considerations but just wondering, would it make sense to return the ticket number that was just created (if it was created)? Seems that it could be helpful for traceability.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -750,6 +754,70 @@ internal class Backend(
...
+                        try {
+                            val wasSent = result.body.optBoolean("sent", false)
+                            onSuccessHandler(wasSent)
+                        } catch (e: SerializationException) {
+                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
+                        } catch (e: IllegalArgumentException) {
+                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
+                        }
+                    } else {
+                        onErrorHandler(result.toPurchasesError().also { errorLog(it) })
```
tonidero, line 806: Just wondering, if we hit this endpoint a lot, I imagine it will be rate limited? If so what's the error code we return in that scenario?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -844,6 +844,21 @@ internal class PurchasesOrchestrator(
         )
     }
 
+    fun createSupportTicket(
+        email: String,
+        description: String,
+        onSuccess: (Boolean) -> Unit,
+        onError: (PurchasesError) -> Unit,
+    ) {
+        backend.postCreateSupportTicket(
```
tonidero, line 853: Hmm I wonder if we should also have some SDK side throttle... but maybe it's early optimizing, and we can rely on the backend for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/composables/SettingsButton.kt
```diff
@@ -73,21 +95,70 @@ internal fun SettingsButton(
...
 @Composable
 private fun SettingsButton_Preview() {
+    fun onClick(){
+    }
+
     SettingsButton(
         title = "Cancel subscription",
         style = SettingsButtonStyle.FILLED,
-    ) {}
+        onClick = { onClick() },
```
tonidero: Hmm is this change needed? Just wondering if we should just leave the callback empty (same in the other previews)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -302,6 +307,90 @@ internal class CustomerCenterViewModelImpl(
...
+                            handleSupportTicketSubmit(email, description, onSuccess, onError)
+                        },
+                        onCancel = {
+                            goBackToMain()
+                        },
+                        onClose = {
+                            goBackToMain()
+                        },
+                    ),
+                    title = "Create Support Ticket", // TODO: Use localized string
```
tonidero: Still need to do this :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,148 @@
...
+internal fun CreateSupportTicketView(
+    data: CreateSupportTicketData,
+    modifier: Modifier = Modifier,
+    localization: CustomerCenterConfigData.Localization,
+) {
+    var email by remember { mutableStateOf("") }
+    var description by remember { mutableStateOf("") }
+    var isSubmitting by remember { mutableStateOf(false) }
+    var hasError by remember { mutableStateOf(false) }
+    val snackbarHostState = remember { SnackbarHostState() }
```
tonidero: Hmm since we're using the snackbar for errors only, would it be worth renaming this to `errorSnackbarHostState` or something like that?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/RelevantPurchasesListView.kt
```diff
@@ -33,6 +33,7 @@ internal fun RelevantPurchasesListView(
     virtualCurrencies: VirtualCurrencies?,
     appearance: CustomerCenterConfigData.Appearance,
     localization: CustomerCenterConfigData.Localization,
+    supportTicket: CustomerCenterConfigData.Support.SupportTickets?,
```
tonidero: Should this be plural? `supportTickets`. Or maybe the type should be renamed to `SupportTicket`...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -363,11 +371,32 @@ private fun CustomerCenterLoaded(
...
-        onAction = onAction,
+    val snackbarHostState = remember { SnackbarHostState() }
+    val successMessage = state.customerCenterConfigData.localization.commonLocalizedString(
+        CustomerCenterConfigData.Localization.CommonLocalizedString.SENT,
     )
+
+    LaunchedEffect(state.showSupportTicketSuccessSnackbar) {
+        if (state.showSupportTicketSuccessSnackbar) {
+            snackbarHostState.showSnackbar(successMessage)
+            onAction(CustomerCenterAction.DismissSupportTicketSuccessSnackbar)
```
tonidero: Hmm this is a bit confusing, as in sending the dismiss action for the snackbar when we are displaying it? Should it be on the `else` of the condition?

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -750,6 +754,70 @@ internal class Backend(
         }
     }
 
+    fun postCreateSupportTicket(
```
tonidero, line 758: Would be great to add some unit tests for this :) I thought about adding some backend integration tests... but then we would be creating tickets very often I guess... but could be good to in some test project?

tonidero (commented):
I don't have much context so will leave to others to approve, but this is looking great! Just some comments.

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
RosieWatson: At the moment this is a suuuper rudimentary implementation of "tickets" that is essentially just sending an email to a developer behind the scenes, so we don't have any concept of a ticket number _yet_

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
RosieWatson: Good point, will make sure that route is covered as well!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/composables/SettingsButton.kt
RosieWatson: Android studio was yelling loudly at me, but maybe it was something else at the time, will try to take out and see if it's happy

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/RelevantPurchasesListView.kt
RosieWatson: it's coming from the backend with a pluralisation, will try to not have it swap about so much between things

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
RosieWatson: I thought this was a fairly common pattern, as it's basically just setting the state back so that it doesn't get re-triggered and show again

tonidero: Oh right 🤦 I didn't remember how that work. Please ignore me

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
tonidero: 👍 Might make sense to return an object so it's extensible in the future, in case we want to return more data then

RosieWatson: That makes sense, will update!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
RosieWatson: Just been looking at this a bit more, so what I think will currently happen is that it will be treated like a generic error as a 429 would be returned. I can't see other endpoints handing this specifically, but maybe I'm reading things wrong - if you have an example that would be stunning

tonidero: No, I think we don't handle throttling from the backend at all right now 😓 It would probably be good to do so but can be done in the future, I think it's fine as long as the message or underlyingErrorMessage contain the reason.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
vegaro: Agree, to prevent abuse it might be smart to only allow to send a message every app session maybe?

RosieWatson: How is an app session defined? Would that mean to send another they would have to close and reopen the app?

vegaro: yes, exactly

RosieWatson: Okay cool, can get that added in then

vegaro: I also think we can tune this later in another PR btw, maybe worth checking with Engin

tonidero: Yeah, I would be careful with such a long throttle... but yeah definitely good to check first on the desired behavior.

RosieWatson: Engin wasn't keen on the full session throttle right now. The backend has a decent throttle on it and if we start seeing some abuse we can update things, if that sounds good

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,219 @@
...
+import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterConstants.Layout.SECTION_SPACING
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterConstants.Layout.SECTION_TITLE_BOTTOM_PADDING
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.composables.SettingsButton
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.composables.SettingsButtonConfig
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CreateSupportTicketData
+
+@JvmSynthetic
+@Composable
+internal fun CreateSupportTicketView(
```
vegaro, line 55: It would be great to add a preview here so we get Emerge warnings if something breaks

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,219 @@
...
+                    CustomerCenterConfigData.Localization.CommonLocalizedString.EMAIL,
+                ),
+            )
+        },
+        placeholder = {
+            Text(
+                localization.commonLocalizedString(
+                    CustomerCenterConfigData.Localization.CommonLocalizedString.ENTER_EMAIL,
+                ),
+            )
```
vegaro, line 245: might be worth it to add some basic email validation to prevent them from reaching our backend

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,219 @@
...
+    OutlinedTextField(
+        value = description,
+        onValueChange = onDescriptionChange,
+        label = {
+            Text(
+                localization.commonLocalizedString(
+                    CustomerCenterConfigData.Localization.CommonLocalizedString.DESCRIPTION,
+                ),
+            )
+        },
```
vegaro, line 297: should we set a max length? just a sanity thing. Does the backend prevent from sending huge chunks of text?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/RelevantPurchasesListView.kt
```diff
@@ -102,6 +103,14 @@ internal fun RelevantPurchasesListView(
...
             localization = localization,
             contactEmail = contactEmail,
+            addCreateTicketButton = supportTickets.allowCreation && (
+                supportTickets.customerType == CustomerCenterConfigData.Support.SupportTickets.CustomerType.ALL ||
+                    supportTickets.customerType == CustomerCenterConfigData.Support.SupportTickets.CustomerType.ACTIVE
+                ),
+            addContactButton = (
+                supportTickets.customerType == CustomerCenterConfigData.Support.SupportTickets.CustomerType.ALL ||
+                    supportTickets.customerType == CustomerCenterConfigData.Support.SupportTickets.CustomerType.ACTIVE
+                ),
```
vegaro: I think extracting some small functions, or maybe making extension functions on `supportTickets` could help legibility

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketViewTest.kt
vegaro, line 1: Nice tests!!

[nicerocket](https://app.graphite.com/settings/meme-library?org=RevenueCat)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
RosieWatson: Good shout, will set to something sensible!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.CustomerCenterConstants.Layout.SECTION_TITLE_BOTTOM_PADDING
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.composables.SettingsButton
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.composables.SettingsButtonConfig
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CreateSupportTicketData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.data.CustomerCenterConfigTestData
+import com.revenuecat.purchases.ui.revenuecatui.customercenter.theme.CustomerCenterPreviewTheme
+
+private const val MAX_DESCRIPTION_LENGTH = 250
+
+private fun isValidEmail(email: String): Boolean {
```
vegaro: super nit but generally private members go at the end of the class. So I would move this function and the two private data classes to the bottom

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+
+@JvmSynthetic
+@Composable
+internal fun CreateSupportTicketView(
+    data: CreateSupportTicketData,
+    localization: CustomerCenterConfigData.Localization,
+    modifier: Modifier = Modifier,
+) {
+    var email by remember { mutableStateOf("") }
+    var emailTouched by remember { mutableStateOf(false) }
```
vegaro: Do we need this? or we can check if `email != ""` instead? Also maybe `emailDirty` fits better, so it doesn't seem like it means that the email field was touched with the finger

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+    var emailHasFocus by remember { mutableStateOf(false) }
+    var description by remember { mutableStateOf("") }
+    var isSubmitting by remember { mutableStateOf(false) }
+    var hasError by remember { mutableStateOf(false) }
```
vegaro: I believe these should be `rememberSaveable` so they survive device rotation, otherwise they will be cleared

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+                description = description,
+                onDescriptionChange = { newValue ->
+                    if (newValue.length <= MAX_DESCRIPTION_LENGTH) {
+                        description = newValue
+                    }
+                },
+                enabled = !isSubmitting,
+            ),
+            isSubmitting = isSubmitting,
+            onSubmit = {
```
vegaro: I think it would be a good idea to trim the email here before doing anything

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+            .fillMaxWidth()
+            .onFocusChanged { focusState ->
+                state.onFocusChanged(focusState.isFocused)
+            }
+            .testTag("email_field"),
+        singleLine = true,
+        keyboardOptions = KeyboardOptions(
+            keyboardType = KeyboardType.Email,
+            imeAction = ImeAction.Next,
+        ),
```
vegaro, line 270: Does this work? Or does it need the following?

```
keyboardActions = KeyboardActions(onNext = { focusManager.moveFocus(FocusDirection.Down) }),
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+                style = MaterialTheme.typography.bodySmall,
+            )
+        },
+        enabled = enabled,
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(200.dp)
+            .testTag("description_field"),
+        minLines = 6,
+        maxLines = 10,
```
vegaro, line 310: You can add the following so it shows `Done` in the keyboard and it submits the form:

```
keyboardOptions = KeyboardOptions(imeAction = ImeAction.Done),
keyboardActions = KeyboardActions(onDone = { if (!isSubmitting && isValidEmail(trimmedEmail) && trimmedDescription.isNotBlank()) onSubmit() }),
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+private fun DescriptionInputField(
+    description: String,
+    onDescriptionChange: (String) -> Unit,
+    enabled: Boolean,
+    localization: CustomerCenterConfigData.Localization,
+) {
+    val remainingChars = MAX_DESCRIPTION_LENGTH - description.length
+    val remainingCharsText = localization.commonLocalizedString(
+        CustomerCenterConfigData.Localization.CommonLocalizedString.CHARACTERS_REMAINING,
+    ).replace("{{ count }}", remainingChars.toString())
```
vegaro: this is tricky but it's the first time we face a string that might require pluralization (`1 char remaining` vs `1 chars remaining`). Maybe we can change it to (1/MAX_DESCRIPTION_LENGTH)?

Your call. It's very minor

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+    onErrorShow: () -> Unit,
+    localization: CustomerCenterConfigData.Localization,
+    modifier: Modifier = Modifier,
+) {
+    val errorSnackbarHostState = remember { SnackbarHostState() }
+    val errorMessage = localization.commonLocalizedString(
+        CustomerCenterConfigData.Localization.CommonLocalizedString.SUPPORT_TICKET_FAILED,
+    )
+
+    LaunchedEffect(hasError, onErrorShow) {
```
vegaro: What's the reason for `onErrorShow` here? Shouldn't `hasError` be enough?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
```diff
@@ -0,0 +1,328 @@
...
+            onSubmit = onSubmit,
+            localization = localization,
+        )
+    }
+}
+
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO or Configuration.UI_MODE_TYPE_NORMAL)
+@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES or Configuration.UI_MODE_TYPE_NORMAL)
+@Composable
+private fun CreateSupportTicketView_Preview() {
```
vegaro: it would be nice to have a preview with error states and with some data filled in. But that might be a hard to mock. If you think it's an easy win, let's add one. I don't think you can do it without modifying `CreateSupportTicketData` and adding an initial email:


```suggestion
@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO or Configuration.UI_MODE_TYPE_NORMAL)
@Preview(uiMode = Configuration.UI_MODE_NIGHT_YES or Configuration.UI_MODE_TYPE_NORMAL)
@Composable
private fun CreateSupportTicketView_Preview() {
    val mockData = CreateSupportTicketData(
        onSubmit = { _, _, _, _ -> },
        onCancel = { },
        onClose = { },
    )

    CustomerCenterPreviewTheme {
        Box(
            modifier = Modifier
                .fillMaxSize()
                .background(MaterialTheme.colorScheme.background),
        ) {
            CreateSupportTicketView(
                mockData,
                CustomerCenterConfigTestData.customerCenterData().localization,
            )
        }
    }
}

@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO or Configuration.UI_MODE_TYPE_NORMAL)
@Composable
private fun CreateSupportTicketView_WithDataPreview() {
    val mockData = CreateSupportTicketData(
        onSubmit = { _, _, _, _ -> },
        onCancel = { },
        onClose = { },
        initialEmail = "user@example.com",
        initialDescription = "I'm having an issue with my subscription."
    )

    CustomerCenterPreviewTheme {
        Box(
            modifier = Modifier
                .fillMaxSize()
                .background(MaterialTheme.colorScheme.background),
        ) {
            CreateSupportTicketView(
                mockData,
                CustomerCenterConfigTestData.customerCenterData().localization,
            )
        }
    }
}

@Preview(uiMode = Configuration.UI_MODE_NIGHT_NO or Configuration.UI_MODE_TYPE_NORMAL)
@Composable
private fun CreateSupportTicketView_WithErrorsPreview() {
    val mockData = CreateSupportTicketData(
        onSubmit = { _, _, _, _ -> },
        onCancel = { },
        onClose = { },
        initialEmail = "invalid-email",  // Invalid email format to trigger error
        initialDescription = ""  // Empty description to trigger error
    )

    CustomerCenterPreviewTheme {
        Box(
            modifier = Modifier
                .fillMaxSize()
                .background(MaterialTheme.colorScheme.background),
        ) {
            CreateSupportTicketView(
                mockData,
                CustomerCenterConfigTestData.customerCenterData().localization,
            )
        }
    }
}
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
RosieWatson: I am unsure on this one, I can't tell if Im getting the accurate flow with just the emulator - I'll change to your suggestion and if you have a device I would appreciate a check

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
RosieWatson: The reason I added this was if I just checked if it was empty, then it would show the validation error from the start, which didn't feel right to me (but happy to change it to be that way)

I will update the var name so it's clearer

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
RosieWatson: That's a very good thing to know, thank you!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
RosieWatson: I clearly forgot basic english here 🤦 will sort

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CreateSupportTicketView.kt
RosieWatson: I have one added with extra info, but the error reproduction wasn't simple as its based on the inputs being touched, which I didnt feel was worth an extra param for just a preview, but let me know if you think it could be

PR #2816, opened 2025-11-06
Add back BC7 support in different flavor

tonidero:
### Description
This splits some of the billing client logic into different flavors BC7 and BC8, each using a different version of Google's Billing client. This is intended to be able to support BC7 for a longer time, and delay the issue presented by Google removing the method to query consumed one time products.

Right now, this will be deployed in different artifacts:
- `com.revenuecat.purchases:purchases-bc7`
- `com.revenuecat.purchases:purchases-ui-bc7`
- `com.revenuecat.purchases:purchases-store-amazon-bc7`

### DIFF
--- a/purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
+++ b/purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
@@ -8,7 +8,9 @@ import com.android.billingclient.api.ProductDetails
 import com.android.billingclient.api.UnfetchedProduct
 import com.revenuecat.purchases.NO_CORE_LIBRARY_DESUGARING_ERROR_MESSAGE
 import com.revenuecat.purchases.ProductType
+import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesErrorCallback
+import com.revenuecat.purchases.PurchasesErrorCode
 import com.revenuecat.purchases.common.DateProvider
 import com.revenuecat.purchases.common.DefaultDateProvider
 import com.revenuecat.purchases.common.LogIntent
@@ -17,6 +19,7 @@ import com.revenuecat.purchases.common.between
 import com.revenuecat.purchases.common.diagnostics.DiagnosticsTracker
 import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.log
+import com.revenuecat.purchases.google.QueryProductDetailsParamsBuilderException
 import com.revenuecat.purchases.google.buildQueryProductDetailsParams
 import com.revenuecat.purchases.google.toGoogleProductType
 import com.revenuecat.purchases.google.toStoreProducts
@@ -55,12 +58,21 @@ internal class QueryProductDetailsUseCase(
         withConnectedClient {
             val googleType: String = useCaseParams.productType.toGoogleProductType() ?: BillingClient.ProductType.INAPP
 
-            queryProductDetailsAsyncEnsuringOneResponse(
-                this,
-                googleType,
-                nonEmptyProductIds,
-                ::processResult,
-            )
+            try {
+                queryProductDetailsAsyncEnsuringOneResponse(
+                    this,
+                    googleType,
+                    nonEmptyProductIds,
+                    ::processResult,
+                )
+            } catch (e: QueryProductDetailsParamsBuilderException) {
+                onError(
+                    PurchasesError(
+                        code = PurchasesErrorCode.StoreProblemError,
+                        underlyingErrorMessage = "${e.message}: ${e.cause?.message}",
+                    ),
+                )
+            }
         }
     }
 
--- a/purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
+++ b/purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
@@ -20,7 +20,10 @@ import com.revenuecat.purchases.utils.mockProductDetails
 import io.mockk.Runs
 import io.mockk.every
 import io.mockk.just
+import io.mockk.mockk
+import io.mockk.mockkStatic
 import io.mockk.slot
+import io.mockk.unmockkStatic
 import io.mockk.verify
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.fail
@@ -244,6 +247,55 @@ internal class QueryProductDetailsUseCaseTest : BaseBillingUseCaseTest() {
     }
 
     // region retries
+
+    @Test
+    fun `If ExceptionInInitializerError is thrown when building QueryProductDetailsParams, returns StoreProblemError`() {
+        val productIDs = setOf("product_a")
+        var receivedError: PurchasesError? = null
+
+        // Mock QueryProductDetailsParams.Builder to throw ExceptionInInitializerError when setProductList is called
+        val mockBuilder = mockk<QueryProductDetailsParams.Builder>()
+        every { mockBuilder.setProductList(any()) } throws ExceptionInInitializerError(
+            RuntimeException("Simulated ExceptionInInitializerError"),
+        )
+
+        mockkStatic(QueryProductDetailsParams::class)
+        every { QueryProductDetailsParams.newBuilder() } returns mockBuilder
+
+        val useCase = QueryProductDetailsUseCase(
+            QueryProductDetailsUseCaseParams(
+                mockDateProvider,
+                mockDiagnosticsTracker,
+                productIDs,
+                ProductType.SUBS,
+                appInBackground = false,
+            ),
+            { _ ->
+                fail("shouldn't be success")
+            },
+            { error ->
+                receivedError = error
+            },
+            withConnectedClient = {
+                it.invoke(mockClient)
+            },
+            executeRequestOnUIThread = { _, request ->
+                request(null)
+            },
+        )
+
+        useCase.run()
+
+        unmockkStatic(QueryProductDetailsParams::class)
+
+        assertThat(receivedError).isNotNull
+        assertThat(receivedError!!.code).isEqualTo(PurchasesErrorCode.StoreProblemError)
+        assertThat(
+            receivedError!!.underlyingErrorMessage,
+        ).isEqualTo(
+            "Error while building QueryProductDetailsParams in Billing client: Simulated ExceptionInInitializerError",
+        )
+    }
 
     @Test
     fun `If service is disconnected, re-executeRequestOnUIThread`() {


.circleci/config.yml
```diff
@@ -947,10 +968,13 @@ workflows:
...
       - prepare-tests
-      - test_defaults_debug:
+      - test_defaults_bc8_debug:
+          requires:
+            - prepare-tests
+      - test_defaults_bc8_release:
           requires:
             - prepare-tests
-      - test_defaults_release:
+      - test_defaults_bc7_release:
```
tonidero, line 977: Note that I'm only running tests for bc7 in release + defaults variant to avoid to have too much extra cardinality and repeated tests.

.run/BC7 Debug tests.run.xml
```diff
@@ -1,5 +1,5 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="Debug tests" type="GradleRunConfiguration" factoryName="Gradle">
+  <configuration default="false" name="BC7 Debug tests" type="GradleRunConfiguration" factoryName="Gradle">
```
tonidero, line 2: Renamed and added run configurations for Android Studio to be able to run tests for a specific flavor more easily.

api-tester/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ android {
         create("defaults") {
             dimension = "apis"
             isDefault = true
+            missingDimensionStrategy("billingclient", "bc8")
```
tonidero, line 16: This flavor dimension doesn't AND shouldn't modify the public API, so I don't think it's worth adding api tests for the bc7 flavor.

examples/paywall-tester/build.gradle.kts
```diff
@@ -18,6 +18,18 @@ android {
 
         missingDimensionStrategy("apis", "defaults")
 
+        flavorDimensions += "billingclient"
```
tonidero, line 21: Decided to create the flavors for paywall tester and purchase tester so it's easier to test in both bc7 and bc8

fastlane/Fastfile
```diff
@@ -124,17 +124,33 @@ platform :android do
       properties: gradle_properties
     )
 
+    UI.verbose("Deploying purchases Billing Client 7 version")
+    deploy_specific_package(
+      "defaultsBc7Release",
+      "purchases-bc7",
```
tonidero, line 130: I still doubted on whether it was worth it splitting into a different version or a different artifact. In the end, the reasoning of the automated version updaters leaned me towards having a separate artifact (haven't tested an actual deployment yet).

fastlane/Fastfile
```diff
@@ -124,17 +124,33 @@ platform :android do
...
+      "customEntitlementComputationBc8Release",
       "purchases-custom-entitlement-computation",
       "purchases",
       gradle_properties
     )
 
     UI.verbose("Deploying Debug view version")
     deploy_specific_package(
-      "defaultsDebug",
+      "defaultsBc8Debug",
```
tonidero, line 161: Not deploying a bc7 version of the debug view nor BOM right now... TBH, I'm not sure how many devs use this... we can always add it if needed in the future...

fastlane/Fastfile
```diff
@@ -941,7 +957,7 @@ def build_purchases_android_test_apk(package_name, apk_name, build_type, flavor)
       "minSdkVersion" => 21
     }
   )
-  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-#{flavor}-#{build_type}-androidTest.apk"
+  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-*-androidTest.apk"
```
tonidero, line 968: Had to change this because the naming of the APK doesn't match the flavor name now that we have multiple dimensions, this was easier than finding another workaround

feature/amazon/build.gradle.kts
```diff
@@ -13,12 +13,20 @@ android {
     namespace = "com.revenuecat.purchases.amazon"
 
     flavorDimensions += "apis"
+    flavorDimensions += "billingclient"
```
tonidero, line 16: I considered if it was worth adding this dimension to the amazon flavor, considering if you're using it, normally you aren't pointing to google... But we do import this in hybrids, and we need to split it up there, so this made the most sense IMO.

purchases/src/bc7/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
```diff
@@ -39,7 +42,7 @@ internal class QueryProductDetailsUseCase(
     val onError: PurchasesErrorCallback,
     val withConnectedClient: (BillingClient.() -> Unit) -> Unit,
     executeRequestOnUIThread: ExecuteRequestOnUIThreadFunction,
-) : BillingClientUseCase<QueryProductDetailsResult>(useCaseParams, onError, executeRequestOnUIThread) {
+) : BillingClientUseCase<List<ProductDetails>>(useCaseParams, onError, executeRequestOnUIThread) {
```
tonidero, line 45: From here, take account of the folder of each file, since that determines if it will be used in the bc7 or bc8 flavor.

purchases/src/bc7/kotlin/com/revenuecat/purchases/google/BillingClientVersion.kt
```diff
@@ -0,0 +1,3 @@
+package com.revenuecat.purchases.google
+
+internal const val BillingClientVersion = "bc7"
```
tonidero, line 3: This is just a way to pass this information on a header to the backend. I thought it could be difficult to track how many devs are using this new version.

purchases/src/main/kotlin/com/revenuecat/purchases/common/purchaseHistoryRecordExtensions.kt
```diff
@@ -0,0 +1,25 @@
+package com.revenuecat.purchases.common
+
+import com.android.billingclient.api.PurchaseHistoryRecord
+import com.revenuecat.purchases.strings.BillingStrings
+import java.util.ArrayList
+
+internal fun PurchaseHistoryRecord.toHumanReadableDescription() =
```
tonidero, line 7: This is interesting... so seems that Google removed the methods for querying purchase history, but didn't remove the models... So for now, I've kept these extension functions in the shared source set. If they end up being removed, we can then move it to the bc7 flavor.

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -1118,27 +1114,6 @@ class BillingWrapperTest {
         }
     }
 
-    @Test
-    fun `getting all purchases gets both subs and inapps`() {
```
tonidero, line 1122: Many of these tests I had to split them in the bc7 and bc8 flavor versions of the tests due to how the code was split up :(

.circleci/config.yml
```diff
@@ -243,7 +243,7 @@ jobs:
           paths:
             - .
 
-  test_defaults_debug:
+  test_defaults_bc8_debug:
```
JayShortway, line 246: Could we give this job, and the `_release` one, a parameter to indicate which bc version it should use? That might avoid some duplication? Just an idea.

api-tester/build.gradle.kts
JayShortway: Agreed, but maybe we can still enable metalava for this flavor?

fastlane/Fastfile
JayShortway: I think different artifact makes sense yea. Worth thinking through the "graduation flow", by which I mean how someone is going to go from `-bc7` to the regular artifact. Current train of thought:

1. When BC9 comes out,
2. we create a new major (v10), and
3. we stop updating `-bc7`.
4. Anyone on `-bc7` will have to either
    a. drop the `-bc7`, or
    b. go straight to v10.

I think that makes sense? Not sure if we can make that smoother. We can publish a relocation POM.xml for `-bc7` too, like I did for [kobankat](https://mvnrepository.com/artifact/io.shortway.kobankat/kobankat-core). That shows up in supported places like so:

Image

fastlane/Fastfile
JayShortway: Agreed

feature/amazon/build.gradle.kts
JayShortway: Wait, can you explain why we need this? Wouldn't both bc7 and bc8 flavors of the hybrids point to the same amazon dependency?

gradle/libs.versions.toml
```diff
@@ -13,6 +13,7 @@ annotationExperimental = "1.5.1"
 appcompat = "1.4.1"
 assertJ = "3.22.0"
 browser = "1.8.0"
+billingBc7 = "7.1.1"
 billing = "8.0.0"
```
JayShortway, line 17: Lowest of priorities:

```suggestion
bc7 = "7.1.1"
bc8 = "8.0.0"
```

gradle/libs.versions.toml
```diff
@@ -96,6 +97,7 @@ androidx-test-runner = { module = "androidx.test:runner", version.ref = "testLib
 assertJ = { module = "org.assertj:assertj-core", version.ref = "assertJ" }
 amazon-appstore-sdk = { module = "com.amazon.device:amazon-appstore-sdk", version.ref = "amazon" }
 billing = { module = "com.android.billingclient:billing", version.ref = "billing" }
+billing-bc7 = { module = "com.android.billingclient:billing", version.ref = "billingBc7" }
```
JayShortway, line 100:
```suggestion
billing-bc8 = { module = "com.android.billingclient:billing", version.ref = "billing" }
billing-bc7 = { module = "com.android.billingclient:billing", version.ref = "billingBc7" }
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -573,14 +602,14 @@ internal class BillingWrapper(
...
                         notNullPurchasesList.takeUnless { it.isEmpty() }?.let { purchase ->
-                            " Purchases:" + purchase.joinToString(
+                            "Purchases:" + purchase.joinToString(
                                 ", ",
                                 transform = { it.toHumanReadableDescription() },
                             )
-                        } ?: " No purchases received"
-                        )
+                        }
+                    }"
```
JayShortway, line 612: What changed here?

purchases/src/bc7/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
JayShortway: Maybe moving the current code to the `/bc8/` folder could have reduced the diff? 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/purchaseHistoryRecordExtensions.kt
JayShortway: Might be safer to move them? I'm thinking of a scenario in which they are removed in BC 8.1, we still depend on BC 8.0, but the developer has declared BC 8.1 separately. That means we would get BC 8.1 at runtime. I guess if we're not calling that function it'd be fine.

purchases/src/main/kotlin/com/revenuecat/purchases/common/purchaseHistoryRecordExtensions.kt
```diff
@@ -0,0 +1,25 @@
...
+ * **Important**: until multi-line subscriptions are released,
+ * we assume that there will be only a single sku in the purchase.
+ * https://android-developers.googleblog.com/2021/05/whats-new-in-google-play-2021.html
+ */
+internal val PurchaseHistoryRecord.firstSku: String
+    get() = skus[0].also {
+        if (skus.size > 1) {
+            log(LogIntent.GOOGLE_WARNING) { BillingStrings.BILLING_PURCHASE_HISTORY_RECORD_MORE_THAN_ONE_SKU }
+        }
+    }
```
JayShortway, line 22: Unrelated tho this PR, but isn't this released now? 😅 (cc @fire-at-will)

purchases/src/main/kotlin/com/revenuecat/purchases/google/errors.kt
```diff
@@ -4,34 +4,12 @@ import com.android.billingclient.api.BillingClient
...
-        "NO_APPLICABLE_SUB_RESPONSE_CODE"
-    BillingClient.OnPurchasesUpdatedSubResponseCode.PAYMENT_DECLINED_DUE_TO_INSUFFICIENT_FUNDS ->
-        "PAYMENT_DECLINED_DUE_TO_INSUFFICIENT_FUNDS"
-    BillingClient.OnPurchasesUpdatedSubResponseCode.USER_INELIGIBLE -> "USER_INELIGIBLE"
-    else -> "UNKNOWN_SUB_RESPONSE_CODE ($this)"
+    val allPossibleBillingResponseCodes = BillingClient.BillingResponseCode::class.java.declaredFields
+    return allPossibleBillingResponseCodes
+        .firstOrNull { it.getInt(it) == this }
+        ?.name
+        ?: "$this"
```
JayShortway, line 12: Why did this change?

purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
JayShortway, line 1: I'm assuming this file is unchanged vs what we have on `main`?

purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCase.kt
JayShortway, line 1: Same here, assuming this file is unchanged vs what we have on `main`?

purchases/src/test/java/com/revenuecat/purchases/utils/billingClientStubs.kt
```diff
@@ -172,6 +163,23 @@ fun stubGooglePurchase(
     signature,
 )
 
+fun stubPurchaseHistoryRecord(
```
JayShortway, line 166: Was this added?

scripts/api-dump.sh
```diff
@@ -2,7 +2,7 @@
 
 exit_code=0
 
-./gradlew metalavaGenerateSignatureDefaultsRelease || exit_code=$?
-./gradlew metalavaGenerateSignatureCustomEntitlementComputationRelease || exit_code=$?
+./gradlew metalavaGenerateSignatureDefaultsBc8Release || exit_code=$?
+./gradlew metalavaGenerateSignatureCustomEntitlementComputationBc8Release || exit_code=$?
```
JayShortway, line 6: As I said in a previous comment, might make sense to enable metalava for both bc7 and bc8?

feature/amazon/build.gradle.kts
tonidero: So the artifact `purchases-store-amazon` has a dependency on the `purchases` artifact, and hybrids include the amazon dependency out of the box. If we don't split the amazon artifact, the hybrids, once we support bc7 versions of them, would depend on `purchases-bc7` and `purchases-store-amazon`, which depends on `purchases`, so we would end up with both `purchases-bc7` and `purchases`, resulting in duplicated classes.

purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCase.kt
tonidero: Well, this one doesn't really exist on `main`, this basically adds a bc8 version that fallsback to just querying products using the available `queryPurchasesAsync`. So this is a bit confusing in bc8, because it's very similar to the other existing `QueryPurchasesUseCase`, but has some differences in the expected type returned (I was trying to minimize the changes to existing code here)

api-tester/build.gradle.kts
tonidero: Great point, indeed I will add that, thanks!

fastlane/Fastfile
tonidero: Yup, that flow makes sense IMO.  And I didn't know about the relocation POM.xml, that's cool!

purchases/src/bc7/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCase.kt
tonidero: Well, my approach was actually to first revert all the changes we did with the bc8 migration. Then, move towards the bc8 implementation from there, so it was kinda a new implementation, even if I took code from current `main`, which was why github didn't pick the correct file for the diff :(

.circleci/config.yml
tonidero: There are some steps we currently only do for the bc8 release version (kover reports)... But I think that could also be extracted to a parameter yeah, good idea!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Nothing really... was the result of a rewrite I did and then reverted, and didn't end up matching the correct style as before 🤦 Will revert this changes

purchases/src/test/java/com/revenuecat/purchases/utils/billingClientStubs.kt
tonidero: This existed in the previous major and I just brought it back

feature/amazon/build.gradle.kts
JayShortway: Ohh right, makes sense, thanks! (I'm thinking maybe `purchases-store-amazon` should use `compileOnly`, but that's a question for another day haha.)

tonidero: That would probably have been good indeed 😅

PR #2817, opened 2025-11-07
Fix dark mode in CustomerCenterView

vegaro:
We got a report that in React Native, the `CustomerCenterView` doesn't switch to dark mode correctly:

**Before**
image

**After**
image

The fix is similar to what we do in CustomerCenterActivity


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CustomerCenterView.kt
```diff
@@ -107,8 +111,13 @@ public class CustomerCenterView : CompatComposeView {
 
     @Composable
     override fun Content() {
-        CustomerCenter(options = customerCenterOptions) {
-            onBackPressed()
+        val isDarkTheme = isSystemInDarkTheme()
+        val colorScheme = if (isDarkTheme) darkColorScheme() else lightColorScheme()
+
+        MaterialTheme(colorScheme = colorScheme) {
```
tonidero, line 117: Just a note that this means we might be overwriting any material theme the developer may currently be using... but since we're doing this in the view, I think the risk of messing up something is small

tonidero (approved):
🚢

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CustomerCenterView.kt
vegaro: yeah I agree. I think they would need to be adding this view using Compose, right? At that point they should be using the Composable 😆

tonidero: Indeed, so I think it's fine.

Issue #2818, opened 2025-11-07
ANRs when calling Purchases#configure


RenanLukas:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
We're experiencing ANRs in our app. After debugging with StrictMode enabled, we discovered that `Purchases.configure()` is performing disk I/O operations on the main thread.

1. Environment
   1. Platform: Android
   2. SDK version: `9.12.2`
   3. OS version: minSdk: 28, targetSdk: 35, compileSdk: 35
   4. Android Studio version: `Android Studio Otter | 2025.2.1`
   5. How widespread is the issue. Percentage of devices affected: Unknown - detected through StrictMode in debug builds, likely affects all devices when `Purchases.configure()` is called on the main thread

2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
```
StrictMode policy violation; ~duration=170 ms: android.os.strictmode.DiskReadViolation
	at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1683)
	at libcore.io.BlockGuardOs.access(BlockGuardOs.java:74)
	at libcore.io.ForwardingOs.access(ForwardingOs.java:128)
	at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:8582)
	at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:332)
	at java.io.File.exists(File.java:829)
	at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:819)
	at android.app.ContextImpl.ensurePrivateCacheDirExists(ContextImpl.java:815)
	at android.app.ContextImpl.getCacheDir(ContextImpl.java:926)
	at android.content.ContextWrapper.getCacheDir(ContextWrapper.java:328)
	at com.revenuecat.purchases.paywalls.FontLoader.<init>(FontLoader.kt:30)
	at com.revenuecat.purchases.PurchasesFactory.createPurchases(PurchasesFactory.kt:325)
[...]

PR #2819, opened 2025-11-07
Fix ANR when initializing FontLoader

vegaro:
Reported here https://github.com/RevenueCat/purchases-android/issues/2818

I believe the cause is we are doing `context.cacheDir` on initialization (in main thread).  Looking at the stacktrace it looks like it does some checks on the disk on main thread that could be causing ANRs.

I tested with StrictMode and after this fix, the warning is gone.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -27,12 +27,16 @@ import com.revenuecat.purchases.utils.Result as RCResult
 @OptIn(InternalRevenueCatAPI::class)
 internal class FontLoader(
     private val context: Context,
-    private val cacheDir: File = File(context.cacheDir, "rc_paywall_fonts"),
```
tonidero, line 30: Hmm does creating a `File` pointer actually perform disk operations? I believe this just creates a pointer, but doesn't actually interact with disk, until you actually want to perform some operations (which I think should be happening in the ioScope?)

vegaro: Right... I think you're correct 🤔

vegaro: What if the issue is `context.cacheDir`? 

Looking at the stacktrace it looks like it does some checks?

```
	at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:8582)
	at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:332)
	at java.io.File.exists(File.java:829)
	at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:819)
	at android.app.ContextImpl.ensurePrivateCacheDirExists(ContextImpl.java:815)
	at android.app.ContextImpl.getCacheDir(ContextImpl.java:926)
	at android.content.ContextWrapper.getCacheDir(ContextWrapper.java:328)
```

vegaro: For reference https://android.googlesource.com/platform/frameworks/base/+/HEAD/core/java/android/app/ContextImpl.java#822

I did check and this PR's changes removes the StrictMode warnings

tonidero: ohhh right... my guess would be that indeed it has something to do with `context.cacheDir` as you said. I guess that's checking for the cache dir existance, and so performing disk operations... Good to know!

vegaro: yes! very sneaky haha

vegaro: btw there are so many other warnings with StrictMode related to shared preferences, but that's another story :D

JayShortway: >checking for the cache dir existance

Yes, I've seen that.

RE StrictMode, I think we should enable it in our Maestro test app (and fix all errors). But indeed, different story.

JayShortway (approved):
Nice, thanks for fixing this!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -27,12 +27,16 @@ import com.revenuecat.purchases.utils.Result as RCResult
...
+    private val providedCacheDir: File? = null,
     private val ioScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
     private val urlConnectionFactory: UrlConnectionFactory = DefaultUrlConnectionFactory(),
 ) {
     private var hasCheckedFoldersExist: AtomicBoolean = AtomicBoolean(false)
 
+    private val cacheDirectory by lazy(LazyThreadSafetyMode.NONE) {
```
tonidero, line 36: Should we use a synchronized safety mode? It shouldn't happen that we initialize this from multiple threads... But might be slightly safer.

vegaro: I actually don't think that would add any safety since there is no need to lock anything. If two threads hit this at the same time without locking they would just return a `File` pointing to the same directory, since `File(context.cacheDir, "rc_paywall_fonts")` is idempotent. And if they don't create the File, they would just return `providedCacheDir` which is also not necessary to synchronize since it's just a `get`

PR #2826, opened 2025-11-10
Adds slack alert and heartbeat monitor to backend integration tests

JayShortway:
## Description
Uses the new actions from the Fastlane plugin.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -378,10 +378,12 @@ platform :android do
 
       run_firebase_integration_tests(app_name)
     rescue => exception
-      send_slack_purchases_integration_test(backend_environment: backend_environment, success: false)
+      slack_backend_integration_test_results(environment: backend_environment, success: false)
       raise
     else
-      send_slack_purchases_integration_test(backend_environment: backend_environment, success: true)
+      slack_backend_integration_test_results(environment: backend_environment, success: true)
+    ensure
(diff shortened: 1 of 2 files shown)


rickvdl (approved):
🙌  Plug and play, nice!

ajpallares (approved):
Nice!

fastlane/Fastfile
```diff
@@ -378,10 +378,12 @@ platform :android do
 
       run_firebase_integration_tests(app_name)
     rescue => exception
-      send_slack_purchases_integration_test(backend_environment: backend_environment, success: false)
+      slack_backend_integration_test_results(environment: backend_environment, success: false)
```
tonidero, line 381: I just noticed the name could be a bit confusing with our current naming for android tests (which is already confusing). I think we have 3 sets of tests:
- Backend integration tests. These are actually unit tests that actually perform backend requests.
- Purchases integration tests. These are instrumentation tests, that test almost E2E behavior, though we do mock Google.
- Maestro E2E tests. These test an actual app using RevenueCat (using test store right now).

This naming might indicate it's used for the first set of tests, but it's actually used for the second... Having said that, I don't think it's a huge deal and something we can fix at any point.

JayShortway: Right, yes we should definitely clean up that naming. I figured they all test the integration of the SDK with the backend, so in that sense they're all "backend integration tests". The action can also be used for all 3 if we want, because it will log the job name too.

PR #2827, opened 2025-11-11
Configure build logic and restructure gradle files

skydoves:
Extracted duplicated Gradle build configuration from individual module build files into a centralized build-logic convention plugin system with reusable plugins, the build-logic configuration. I followed our KMP project (https://github.com/RevenueCat/purchases-kmp) for the overall structures.
 
Advantages of this build logic structure:

1. Reduce duplication: Common configuration (Android settings, flavors, publishing, Metalava API tracking) is defined once and reused across all 7+ modules instead of being copy-pasted.
2. Type-safety and IDE support: Kotlin DSL convention plugins provide compile-time validation, auto-completion, and refactoring capabilities that Groovy scripts don't have.
3. Centralized maintenance: Changes to build configuration (like updating compile SDK or Kotlin options) only need to be made in one place rather than hunting through multiple build files.
4. Enforces consistency: All library modules automatically get the same configuration through the convention plugins, preventing configuration drift between modules.

### DIFF
--- a/ui/debugview/build.gradle.kts
+++ b/ui/debugview/build.gradle.kts
@@ -7,6 +7,18 @@ plugins {
 android {
     namespace = "com.revenuecat.purchases.ui.debugview"
 
+    // billingclient dimension needed to match dependencies
+    flavorDimensions += "billingclient"
+    productFlavors {
+        create("bc8") {
+            dimension = "billingclient"
+            isDefault = true
+        }
+        create("bc7") {
+            dimension = "billingclient"
+        }
+    }
+
     defaultConfig {
         minSdk = 21
     }


build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
```diff
@@ -0,0 +1,60 @@
...
+
+        compileOptions {
+            sourceCompatibility = JavaVersion.VERSION_1_8
+            targetCompatibility = JavaVersion.VERSION_1_8
+        }
+    }
+
+    extensions.configure<KotlinAndroidProjectExtension> {
+        compilerOptions {
+            jvmTarget.set(org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8)
```
tonidero: Should we add imports for these instead of using the fully qualified name?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureApisFlavorsForApplication.kt
```diff
@@ -0,0 +1,25 @@
...
+    extensions.configure<ApplicationExtension> {
+        flavorDimensions += "apis"
+        productFlavors {
+            create("defaults") {
+                dimension = "apis"
+                isDefault = true
+            }
+            create("customEntitlementComputation") {
+                dimension = "apis"
+                missingDimensionStrategy("apis", "defaults")
```
tonidero: Hmm I know this was like this before but is this needed? I would think that in this case, there wouldn't be a missing dimension case, since the dimension is created right?

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureConditionalPublishing.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.revenuecat.purchases.android.buildlogic.ktx.libs
+import com.revenuecat.purchases.android.buildlogic.ktx.plugins
+import org.gradle.api.Project
+
+/**
+ * Configures conditional Maven publishing based on the ANDROID_VARIANT_TO_PUBLISH property.
+ * Only applies publishing for variants that don't contain "customEntitlementComputation"
+ */
+internal fun Project.configureConditionalPublishing() {
+    if (!project.properties["ANDROID_VARIANT_TO_PUBLISH"].toString().contains("customEntitlementComputation")) {
```
tonidero: Hmm I think this is applied to all public libraries, but we still want to be able to publish the CEC version of the `purchases` module right? We do want to skip it for the `purchases-ui` dependency + bom, so maybe we need to split this out somehow...

tonidero (commented):
Looking amzing! Thanks for doing this! 🙇  I think the only important question is whether we're still publishing the customEntitlementComputation in the purchases module.

skydoves:
Awesome, @tonidero! I just applied your comments and resolved some issues on [57cdb12](https://github.com/RevenueCat/purchases-android/pull/2827/commits/57cdb12eb5ebf90e1fea10bfda03624aa7cd77ff).

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureApisFlavors.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.android.build.api.dsl.LibraryExtension
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+
+/**
+ * Configures the standard "apis" flavor dimension used across all library modules
+ */
+internal fun Project.configureApisFlavors() {
+    extensions.configure<LibraryExtension> {
+        flavorDimensions += "apis"
```
tonidero, line 12: We will need to add another one like this for the `billingclient` dimension, that may have bc8 (default) or bc7 flavors. This should be configured for the `purchases`, `purchases-ui` and `puchases-store-amazon` dependencies. No need for the `debugview` dependency. Sorry for the conflicts 🙏

skydoves: Thank you for the detailed histories! 🙏  I just resolved the conflicts and configured the `billingclient` dimension on [b011871](https://github.com/RevenueCat/purchases-android/pull/2827/commits/b011871d6d860cd79f2879f213f82be3fb1c2629).

PR #2829, opened 2025-11-11
Add `bc8` default flavor

tonidero:
### Description
First step to split bc7 and bc8 flavors. For now, just creating an empty default bc8 flavor, which will be used by all code. In follow-up PRs, I will start moving code to the correct flavor, before applying the changes in the pretty big #2816 . This PR is pointing to a base branch to merge all PRs, not to main.

No changes should happen as consequence of this PR.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -909,6 +909,9 @@ end
 
 def build_purchases_android_test_apk(package_name, apk_name, build_type, flavor)
   task_to_run = flavor == 'defaultsBc8' ? 'assembleDefaultsBc8AndroidTest' : 'assembleCustomEntitlementComputationBc8AndroidTest'
+  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-*-androidTest.apk"
+  # Delete any preexisting apks to avoid possible interferences
+  sh "rm -f #{apk_path}"
   gradle(
     task: ":purchases:#{task_to_run}",
     properties: {
@@ -917,7 +920,6 @@ def build_purchases_android_test_apk(package_name, apk_name, build_type, flavor)
       "minSdkVersion" => 21
     }
   )
-  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-*-androidTest.apk"
   new_apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/#{apk_name}.apk"
   sh "mv #{apk_path} #{new_apk_path}"
   copy_artifacts(


.run/Debug tests.run.xml
```diff
@@ -10,14 +10,15 @@
...
-          <option value="testDefaultsDebugUnitTest" />
+          <option value="testDefaultsBc8DebugUnitTest" />
         </list>
       </option>
       <option name="vmOptions" />
     </ExternalSystemSettings>
     <ExternalSystemDebugServerProcess>true</ExternalSystemDebugServerProcess>
     <ExternalSystemReattachDebugProcess>true</ExternalSystemReattachDebugProcess>
     <DebugAllEnabled>false</DebugAllEnabled>
+    <RunAsTest>false</RunAsTest>
```
ajpallares, line 21: I'm curious. What does this do?

purchases/build.gradle.kts
```diff
@@ -37,6 +38,10 @@ android {
         create("customEntitlementComputation") {
             dimension = "apis"
         }
+        create("bc8") {
+            dimension = "billingclient"
+            isDefault = true
+        }
```
ajpallares, line 44: I'm guessing this means that this will be transparent for devs, in the sense that by default they will be using the bc8 flavor with no changes required, right?

ajpallares (approved):
Changes look good to me. Basically a rename and adding the new "billingclient" flavor dimension. Would be ideal to have someone more familiarized with Android take a look as well though!

purchases/build.gradle.kts
tonidero: Yup pretty much, this change should be transparent to users of the SDK. As for us, it will cause the rename of some gradle tasks, but right now, no changes, since this is the only flavor available in this dimension (will add bc7 in follow-up PRs)

.run/Debug tests.run.xml
tonidero: TBH, I'm not sure, this was added by AS, but it still runs tests correctly as far as I can tell

fastlane/Fastfile
```diff
@@ -917,7 +917,7 @@ def build_purchases_android_test_apk(package_name, apk_name, build_type, flavor)
       "minSdkVersion" => 21
     }
   )
-  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-#{flavor}-#{build_type}-androidTest.apk"
+  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-*-androidTest.apk"
```
vegaro: what does this change do?

tonidero: Not much right now... before we were able to accurately match the name of the generated apk for tests, but now it's trickier, because the flavor is in the form of `defaultsBc8` while the apk is in the form of `purchases-defaults-bc8-debug-androidTest.apk`. We could try to split the flavors like that, but since there should only be one apk built there when running these normally, I'm just matching with "any" apk.

fastlane/Fastfile
```diff
@@ -400,7 +400,7 @@ platform :android do
       # We default to empty active entitlements since we are using an expired token for these integration tests.
       # So no active entitlements will be returned by our servers.
       active_entitlements_to_verify: ENV['CUSTOM_ENTITLEMENT_COMPUTATION_ACTIVE_ENTITLEMENTS_TO_VERIFY'] || '',
-      flavor: 'customEntitlementComputation'
+      flavor: 'customEntitlementComputationBc8'
```
JayShortway, line 403: Do we need the BC split for CEC mode too?

fastlane/Fastfile
JayShortway: Maybe clearing the `outputs/apk` folder at the start of the lane can add some safety?

purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
```diff
@@ -22,20 +23,23 @@ import io.mockk.mockkStatic
 import io.mockk.slot
 import io.mockk.unmockkStatic
 import io.mockk.verify
-import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.Assertions
```
JayShortway: Why was this changed? I think undoing this change can reduce the diff a lot.

purchases/build.gradle.kts
JayShortway: The `isDefault` mostly means Android Studio will select the `bc8` variant by default. So this is helpful for us when developing the SDK.

gradle.properties
```diff
@@ -47,7 +47,7 @@ paywallTesterVersionName=1.0
 
 #Do not sign releases. When calling uploadArchives pass -PRELEASE_SIGNING_ENABLED=true
 RELEASE_SIGNING_ENABLED=false
-ANDROID_VARIANT_TO_PUBLISH=defaultsRelease
+ANDROID_VARIANT_TO_PUBLISH=defaultsBc8Release
```
JayShortway, line 50: I'm assuming you tested this with mavenLocal? 😄

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -1367,78 +1364,6 @@ class BillingWrapperTest {
 
     // region diagnostics tracking
 
-    @Test
-    fun `querySkuDetailsAsync tracks diagnostics call with correct parameters`() {
```
JayShortway, line 1371: I was expecting these tests to be moved somewhere, but I can find where to. Was my expectation incorrect haha? If so, why are they removed?

JayShortway (approved):
Looking good! Just some questions.

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
JayShortway: Nvm, found in the follow-up PR!

fastlane/Fastfile
tonidero: Well, I think not unless users of CEC mode ask for it? Seems unnecessary otherwise IMO...

tonidero: But having the dimension exist there makes it easier for us to organize things IMO

purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
tonidero: Indeed... I don't remember why I did this, but will revert the changes 👍

fastlane/Fastfile
tonidero: Good idea! will do that 👍

gradle.properties
tonidero: Well, it does publish successfully, with the expected version number, but I haven't tested that artifact yet. I will run some tests on the final PR of this chain.

fastlane/Fastfile
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2829/commits/e7c3420b8e3079d921563c16a2aa68153cb14a30

purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
tonidero: Reverted in https://github.com/RevenueCat/purchases-android/pull/2829/commits/9db17b6169ad45682975002fba2997096cbcd932

PR #2831, opened 2025-11-12
Add `QueryPurchaseHistoryUseCase` to BC8 implementation

tonidero:
### Description
This PR includes a few changes:
- Includes a new `QueryPurchaseHistoryUseCase` to the BC8 implementation, which under the hood, just queries active purchases
- Readds the catching of the `NoCoreLibraryDesugaringException`
- Readds diagnostics tracking for purchase history events

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -909,6 +909,9 @@ end
 
 def build_purchases_android_test_apk(package_name, apk_name, build_type, flavor)
   task_to_run = flavor == 'defaultsBc8' ? 'assembleDefaultsBc8AndroidTest' : 'assembleCustomEntitlementComputationBc8AndroidTest'
+  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-*-androidTest.apk"
+  # Delete any preexisting apks to avoid possible interferences
+  sh "rm -f #{apk_path}"
   gradle(
     task: ":purchases:#{task_to_run}",
     properties: {
@@ -917,7 +920,6 @@ def build_purchases_android_test_apk(package_name, apk_name, build_type, flavor)
       "minSdkVersion" => 21
     }
   )
-  apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/purchases-*-androidTest.apk"
   new_apk_path = "../purchases/build/outputs/apk/androidTest/#{flavor}/#{build_type}/#{apk_name}.apk"
   sh "mv #{apk_path} #{new_apk_path}"
   copy_artifacts(
--- a/purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
+++ b/purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryProductDetailsUseCaseTest.kt
@@ -23,7 +23,7 @@ import io.mockk.mockkStatic
 import io.mockk.slot
 import io.mockk.unmockkStatic
 import io.mockk.verify
-import org.assertj.core.api.Assertions
+import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.fail
 import org.assertj.core.api.AssertionsForClassTypes
 import org.assertj.core.data.Offset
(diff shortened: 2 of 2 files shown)


purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCase.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlin.time.Duration
+
+internal data class QueryPurchaseHistoryUseCaseParams(
+    val dateProvider: DateProvider = DefaultDateProvider(),
+    val diagnosticsTrackerIfEnabled: DiagnosticsTracker?,
+    @BillingClient.ProductType val productType: String,
+    override val appInBackground: Boolean,
+) : UseCaseParams
+
+internal class QueryPurchaseHistoryUseCase(
```
vegaro, line 38: The fact that this is now called `QueryPurchaseHistory` but it's doing `queryPurchasesAsync` is confusing me tbh. I understand it's called like that because bc7 will actually query the history right?

tonidero: Thats correct. The file needs to be called the same way to make importing it easier in the main source set

tonidero: It’s indeed confusing… not sure if we could come up with a better name… but basically each flavor will be doing something different

vegaro: It returns transactions. So maybe `QueryTransactions`? and then if bc7 uses the history and bc8 the current active subscriptions it's implementation details? Even `QueryPurchases`?

tonidero: There is already a QueryPurchasesUseCase that only queries active purchases, so needs to be different to differentiate them… about QueryTransactions, I think it could still be very confusing since, what would be the difference with QueryPurchasesUseCase?

vegaro: Right. Maybe the current `QueryPurchases` should be `QueryActivePurchases`, and this should be `QueryPurchases` instead of `QueryPurchasesHistory`. But that's a bigger refactor and I don't have a better 

It's ok, I won't block the PR since I don't have a better idea than this 😅 haha

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -888,17 +917,21 @@ internal class BillingWrapper(
...
+                        isPersonalizedPrice?.let {
+                            setIsOfferPersonalized(it)
+                        }
                     }
-                }
-                .build(),
-        )
+                    .build(),
+            )
+        } catch (e: NoClassDefFoundError) {
```
vegaro, line 932: Can we add a comment here about why this is needed?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -909,22 +942,26 @@ internal class BillingWrapper(
...
```
vegaro, line 965: Same here

vegaro (approved):

purchases/src/bc8/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCase.kt
JayShortway: Might be worth it to add some _KDoc_ to this class to explain why it is named the way it is?

purchases/src/testBc8/kotlin/com/revenuecat/purchases/google/usecase/QueryPurchaseHistoryUseCaseTest.kt
```diff
@@ -0,0 +1,768 @@
...
+        verify(exactly = 2) {
+            mockClient.queryPurchasesAsync(any<QueryPurchasesParams>(), any())
+        }
+        clearStaticMockk(QueryPurchasesParams::class)
+    }
+
+    // region diagnostics tracking
+
+    @Test
+    fun `queryPurchaseHistoryAsync tracks diagnostics call with correct parameters`() {
```
JayShortway, line 242: Found them! 😄

PR #2833, opened 2025-11-12
Add BC7 unit tests job and add metalava check for bc7 flavor

tonidero:
### Description
This refactors our unit tests jobs to be a single one with parameters for what flavors to run them on, so we can run them in a matrix-like pattern.

Additionally, we also add a metalava check for the bc7 flavor, so we make sure the API is not different accidentally.

### DIFF
--- a/purchases/api-defaults-bc7.txt
+++ b/purchases/api-defaults-bc7.txt
@@ -1,0 +1,1452 @@
+// Signature format: 4.0
+package com.revenuecat.purchases {
+
+  public enum AmazonLWAConsentStatus {
+    enum_constant public static final com.revenuecat.purchases.AmazonLWAConsentStatus CONSENTED;
+    enum_constant public static final com.revenuecat.purchases.AmazonLWAConsentStatus UNAVAILABLE;
+  }
+
+  public enum CacheFetchPolicy {
+    enum_constant public static final com.revenuecat.purchases.CacheFetchPolicy CACHED_OR_FETCHED;
+    enum_constant public static final com.revenuecat.purchases.CacheFetchPolicy CACHE_ONLY;
+    enum_constant public static final com.revenuecat.purchases.CacheFetchPolicy FETCH_CURRENT;
+    enum_constant public static final com.revenuecat.purchases.CacheFetchPolicy NOT_STALE_CACHED_OR_CURRENT;
+    field public static final com.revenuecat.purchases.CacheFetchPolicy.Companion Companion;
+  }
+
+  public static final class CacheFetchPolicy.Companion {
+    method public com.revenuecat.purchases.CacheFetchPolicy default();
+  }
+
+  public final class CoroutinesExtensionsCommonKt {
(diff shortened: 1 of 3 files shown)


purchases/build.gradle.kts
```diff
@@ -104,7 +104,13 @@ metalava {
 
     val name = if (variantName.lowercase().contains("defaults")) {
         excludeSourceSets.add("src/customEntitlementComputation/kotlin")
-        "api-defauts.txt"
+        if (variantName.lowercase().contains("bc8")) {
+            "api-defauts.txt"
```
rickvdl, line 108: Noticed this typo before as well. If easy we can change it, if not let's leave it for another time :)

tonidero: Yeah I actually thought about fixing it here... But I wanted to avoid having a different file with these PRs, that could make it easy to miss API changes. I think this should be fixed, but probably better to do it in a separate PR chain 😅

.circleci/config.yml
```diff
@@ -243,95 +243,55 @@ jobs:
...
-      - store_test_results:
-          path: build/test-results-defaults-bc8-release
-      - store_artifacts:
-          path: build/reports
-
-  test_cec_debug:
+  test:
     <<: *android-executor
     shell: /bin/bash --login -o pipefail
+    parameters:
```
vegaro, line 249: nice!

purchases/build.gradle.kts
vegaro: I can't stop seeing it now

purchases/api-defaults-bc7.txt
```diff
@@ -0,0 +1,1452 @@
...
+    method public String getCode();
+    method public String getName();
+    method public String? getServerDescription();
+    property public final int balance;
+    property public final String code;
+    property public final String name;
+    property public final String? serverDescription;
+  }
+
+}
```
vegaro, line 1451: can we compare this with version 8.24.0? Maybe you have already but might be a good way to check if something is missing?

.circleci/config.yml
```diff
@@ -1013,6 +993,7 @@ workflows:
           requires:
             - test_defaults_bc8_debug
             - test_defaults_bc8_release
+            - test_defaults_bc7_release
```
vegaro, line 996: just double checking we don't want `bc7_debug`?

purchases/api-defaults-bc7.txt
tonidero: Yeah, there are a bunch of changes, since we did stop a bunch of classes from being data classes in this major, but nothing unexpectedly missing I would say. Lmk if you have a specific  concern!

.circleci/config.yml
tonidero: We could add it... I just wanted to avoid all the extra cardinality, which I feel doesn't provide that much value IMO. Happy to add it if you feel we should have it though!

.circleci/config.yml
JayShortway: Beautiful!

purchases/build.gradle.kts
JayShortway: I read it like it's French, _defáuts_ 🇫🇷

.circleci/config.yml
vegaro: Just wanted to make sure you didn't miss it and it was a concious decision

PR #2834, opened 2025-11-12
Add new header to detect what version of SDK is being used

tonidero:
### Description
This adds a new header `X-Billing-Client-Sdk-Version` that will indicate what version of the SDK (bc8 or bc7) is used

### DIFF
--- a/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/request_001.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/request_001.json
@@ -3,7 +3,7 @@
   "method": "GET",
   "headers": {
     "Content-Type": "application\/json",
-    "X-Billing-Client-Sdk-Version": "bc8",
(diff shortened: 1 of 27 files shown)


ajpallares (approved):
Right! Makes sense!

purchases/src/bc8/kotlin/com/revenuecat/purchases/google/BillingClientVersion.kt
```diff
@@ -0,0 +1,3 @@
+package com.revenuecat.purchases.google
+
+internal const val BillingClientVersion = "bc8"
```
JayShortway: Maybe we should make this the actual version, e.g. `8.0.0` (same for bc7)? That would give us a bit more granularity. 

It could be set as a BuildConfig value that reads from libs.versions.toml, for instance. I did something similar in purchases-kmp.

tonidero: Great idea! Done in https://github.com/RevenueCat/purchases-android/pull/2834/commits/e73f715bce8bf41ed58fb087869e999894d8eac0

PR #2837, opened 2025-11-13
Countdown Paywalls component

vegaro:
Equivalent of ([#5790](https://github.com/RevenueCat/purchases-ios/pull/5790)).

### Overview
Adds support for countdown timers in paywalls that can display different content based on whether the countdown is active or has ended.

### Note on prewarming vides/images
When a countdown ends and transitions from `countdownStack` to `endStack`, any images or videos in the end stack (which can contain anything) need to be immediately available.

[Screen_recording_20251113_125835.webm](https://github.com/user-attachments/assets/9f794ab6-b848-45af-a847-d77cd10d7dad)

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
@@ -7,8 +7,6 @@ import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.common.canUsePaywallUI
 import com.revenuecat.purchases.models.Checksum
-import com.revenuecat.purchases.paywalls.components.CountdownComponent
-import com.revenuecat.purchases.paywalls.components.StackComponent
 import com.revenuecat.purchases.paywalls.components.VideoComponent
 import com.revenuecat.purchases.paywalls.components.properties.ThemeVideoUrls
 import com.revenuecat.purchases.storage.DefaultFileRepository
@@ -27,25 +25,13 @@ internal class OfferingVideoPredownloader(
             // WIP: We will add a remote flag in the offering metadata that will indicate if we should
             // pre-download videos or not. For now, we want to only download the low-res to ensure we
             // don't rack up high cloudfront costs
-            offering.paywallComponents?.data?.componentsConfig?.base?.stack?.let { stack ->
-                downloadVideosFromStack(stack)
-            }
-        }
-    }
-
-    private fun downloadVideosFromStack(stack: StackComponent) {
-        stack.components.forEach { component ->
-            when (component) {
-                is VideoComponent -> {
-                    fileRepository.prefetch(component.source.checkedUrls())
+            offering.paywallComponents?.data?.componentsConfig?.base?.stack
+                ?.filter { it is VideoComponent }
+                ?.forEach { component ->
+                    if (component is VideoComponent) {
+                        fileRepository.prefetch(component.source.checkedUrls())
+                    }
                 }
-                is CountdownComponent -> {
-                    downloadVideosFromStack(component.countdownStack)
-                    component.endStack?.let { downloadVideosFromStack(it) }
-                    component.fallback?.let { downloadVideosFromStack(it) }
-                }
-                else -> {}
-            }
         }
     }
 }


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentView.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.compositionLocalOf
+import androidx.compose.ui.Modifier
+import com.revenuecat.purchases.ui.revenuecatui.components.PaywallAction
+import com.revenuecat.purchases.ui.revenuecatui.components.stack.StackComponentView
+import com.revenuecat.purchases.ui.revenuecatui.components.style.CountdownComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.data.PaywallState
+
+@Suppress("CompositionLocalAllowlist")
+internal val LocalCountdownTime = compositionLocalOf<CountdownTime?> { null }
```
vegaro: I am using `CompositionLocal` here to avoid prop drilling `CountdownTime` through every component layer
`(CountdownComponentView → StackComponentView → ComponentView → TextComponentView)`.

The alternative would be to add a `countdownTime` parameter to all intermediate components, which would pollute their APIs since they only pass it down and it's not used very often by `TextComponentView`.

I think this is a [legitimate use](https://developer.android.com/develop/ui/compose/compositionlocal#deciding) of `CompositionLocal`: "scoped context for a specific subtree that updates frequently". But I am unsure. 

What do you think @tonidero @JayShortway

tonidero: Right I think it's an option. However, I do think that for other similar use cases, like passing down the selected `Package` or `tabIndex` or things like that, we've used the `StyleFactoryScope`, to scope that data to that subtree. Not sure if we want to try to do the same thing here? 

`CompositionLocal` might be simpler though, so I could be convinced to use it.

JayShortway: Yes I was going to suggest the same. We know which subtree (branch) needs the countdown at creation time, so we can use the `StyleFactoryScope` for it. I'd really prefer to avoid `CompositionLocal`s. It's easy now, but it makes testing and previewing harder, as it introduces implicit dependencies.

vegaro: Perfect. Thank you! I will check the `StyleFactoryScope` stuff you're talking about

JayShortway: We could create a `withCountdown()` function, like we have `withTabIndex()`, `withTabControl()` etc.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
```diff
@@ -0,0 +1,34 @@
...
+    @get:JvmSynthetic val endStack: StackComponent? = null,
+    @get:JvmSynthetic val fallback: StackComponent? = null,
+) : PaywallComponent {
+
+    @InternalRevenueCatAPI
+    @Poko
+    @Serializable
+    @Immutable
+    class CountdownStyle(
+        @get:JvmSynthetic val type: String,
```
tonidero, line 30: Just curious, I see in the examples that we have `date` as the type. What other types could there be? And if there are other types, should the field below be nullable? Though TBH, I don't think this would really work without a date 😅, but maybe we want to make it be supported without a date and use the fallback stack, if a date is not present? (just future proofing what will happen if we add other types).

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
```diff
@@ -0,0 +1,34 @@
...
+        @Serializable(with = ISO8601DateSerializer::class)
+        @get:JvmSynthetic val date: Date,
```
tonidero, line 32: I'm wondering... should this date be location aware? As in, maybe someone would want to make the countdown finish at 9pm local time everywhere. I guess the backend could try to decide that, but it might be much trickier there IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlinx.serialization.Serializable
+import java.util.Date
+
+@InternalRevenueCatAPI
+@Poko
+@Serializable
+@SerialName("countdown")
+@Immutable
+class CountdownComponent(
+    @get:JvmSynthetic val style: CountdownStyle,
```
tonidero, line 17: Hmm I was wondering if we would want to create a `PartialCountdownComponent` to support overrides, the same way we do in other components... But maybe it's not needed for now? We would only be overriding the date right now, which feels unnecessary IMO. So I think we're good 😅

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StackComponentStyle.kt
```diff
@@ -55,6 +55,11 @@ internal data class StackComponentStyle(
      */
     @get:JvmSynthetic
     val tabIndex: Int?,
+    /**
+     * If this is non-null, it means this stack is inside a countdown component.
+     */
+    @get:JvmSynthetic
+    val countdownDate: java.util.Date?,
```
tonidero: Should we import, instead of using `java.util.Date`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -112,6 +113,10 @@ internal class StyleFactory(
          * If this is non-null, it means the branch currently being built is inside a tab component.
          */
         var tabIndex: Int? = null,
+        /**
+         * If this is non-null, it means the branch currently being built is inside a countdown component.
+         */
+        var countdownDate: java.util.Date? = null,
```
tonidero: Same about the import, and below as well

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/ISO8601DateSerializer.kt
```diff
@@ -12,7 +12,7 @@ import java.util.Date
 internal object ISO8601DateSerializer : KSerializer<Date> {
 
     override val descriptor: SerialDescriptor
-        get() = PrimitiveSerialDescriptor("Date", PrimitiveKind.STRING)
+        get() = PrimitiveSerialDescriptor("ISO8601Date", PrimitiveKind.STRING)
```
vegaro, line 15: Updated this because there's a `DateSerializer` with "Date" already as descriptor. I think this is only for debugging and such, so not really breaking anything

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -425,9 +433,34 @@ internal class StyleFactory(
...
         }
     }
 
+    private fun StyleFactoryScope.createCountdownComponentStyle(
+        component: CountdownComponent,
+    ): Result<CountdownComponentStyle, NonEmptyList<PaywallValidationError>> =
+        withSelectedScope(
+            packageInfo = null,
+            tabControlIndex = null,
+            countdownDate = component.style.date,
```
tonidero: Hmm something I just thought about, would the `null` values for package/tabControlIndex override the existing data if any? 
For example, what would happen if the countdown timer is inside some tabs or within a package? Would that info keep getting propagated inside a CountdownComponent? Say for example, that I have 2 tabs, with a package each and inside each, I have a paywall component where I want to put the selected package name/duration or something like that.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/TextComponentStyle.kt
```diff
@@ -55,6 +55,12 @@ internal class TextComponentStyle(
      */
     @get:JvmSynthetic
     val tabIndex: Int?,
+    /**
+     * If this is non-null, it means this text is inside a countdown component and countdown variables should be
+     * replaced with values calculated from this date.
+     */
+    @get:JvmSynthetic
+    val countdownDate: java.util.Date?,
```
tonidero: Same about imports.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
```diff
@@ -57,10 +59,16 @@ internal fun TextComponentView(
...
+    // Calculate countdown only if this text is inside a countdown
+    val countdownTime = style.countdownDate?.let { date ->
+        rememberCountdownState(date).countdownTime
+    }
+
     // Process any variables in the text.
     val text = rememberProcessedText(
         state = state,
         textState = textState,
+        countdownTime = countdownTime,
```
tonidero: Hmm I wondered if we should move this to the `TextComponentState`, same as the package/tabIndex... but not sure if that could be a problem considering this will be updated every second? If there are no performance concerns, I would consider moving it there.

tonidero (commented):
Some initial comments, but looking amazing!!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentState.kt
```diff
@@ -0,0 +1,74 @@
...
+                hasEnded = true
+                break
+            }
+
+            countdownTime = CountdownTime.fromInterval(delta)
+            hasEnded = false
+
+            delay(1.seconds)
+        }
+    }
```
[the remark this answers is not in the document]
vegaro: fixed in [83d601c](https://github.com/RevenueCat/purchases-android/pull/2837/commits/83d601c292d49ad51aade608a9a63de8f7a0d0d3)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
vegaro: I believe I asked this to @joshdholtz and he mentioned we can skip it for now. But let's confirm

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
vegaro: Right! I created a `withCountdown()` after Joop's suggestion, and I noticed that as well, that the  `packageInfo` and `tabControlIndex` would get lost. I think with the `withCountdown()` that should be fixed right?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
vegaro: I believe this date should be the client's timezone, but I am going to raise it because the frontend it's confusing and I am actually not that sure what it represents.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentViewModel.kt
```diff
@@ -0,0 +1,74 @@
...
+
+            return CountdownTime(days, hours, minutes, seconds)
+        }
+    }
+}
+
+@Composable
+internal fun rememberCountdownState(targetDate: Date): CountdownState {
+    var countdownTime by remember { mutableStateOf(CountdownTime.ZERO) }
+    var hasEnded by remember { mutableStateOf(false) }
```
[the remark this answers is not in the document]
vegaro: fixed in [c48e2b6](https://github.com/RevenueCat/purchases-android/pull/2837/commits/c48e2b69b8f0b6ab4ce4a2ca5f7b7165b1c80b96)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
vegaro: It's only date for now by looking at the backend code

https://github.com/RevenueCat/khepri/pull/15960/files#diff-2cdc9c6d57838d95711b57f3bd73f86d74c399be24dfb2611c361c91523731cfR18

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
tonidero: That's right! I think this is not an issue anymore 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
tonidero: Hmm if it's the date without timezone information, we should make sure we correct it client side... I don't think the backend would send different info depending on the client, since every customization in such big jsons is very costly... So yeah, better to know what this means, and see if we need to process the date in any way client-side... Maybe it can be just an additional setting for "LOCAL_TIME"/"GLOBAL_TIME", that indicates whether the SDK should process it one way or another.

vegaro: ok double checked this with the team and the backend is always going to send UTC, so we should be fine since `Date().time` returns UTC/GMT https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#getTime--

JayShortway: Does ISO-8601 always contain a time zone? I think we should be very explicit about the time zone, even if it's always UTC. We can also use millis since epoch?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
vegaro: Done in https://github.com/RevenueCat/purchases-android/pull/2837/commits/3b107f51dd38ef6407a085cbdf4a890aa1b83ec0

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentState.kt
```diff
@@ -0,0 +1,100 @@
...
+    val lifecycleOwner = LocalLifecycleOwner.current
+
+    LaunchedEffect(targetDate) {
+        launch {
+            lifecycleOwner.lifecycle.currentStateFlow
+                .map { it.isAtLeast(Lifecycle.State.STARTED) }
+                .distinctUntilChanged()
+                .collect { isStarted ->
+                    isCountingEnabled = isStarted
+                }
```
JayShortway, line 68: Nice!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StackComponentStyle.kt
```diff
@@ -55,6 +56,11 @@ internal data class StackComponentStyle(
      */
     @get:JvmSynthetic
     val tabIndex: Int?,
+    /**
+     * If this is non-null, it means this stack is inside a countdown component.
+     */
+    @get:JvmSynthetic
+    val countdownDate: Date?,
```
JayShortway, line 63: 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -299,6 +305,23 @@ internal class StyleFactory(
...
+        ): T {
+            val currentScope = copy()
+            this.countdownDate = countdownDate
+
+            val result = block()
+
+            this.countdownDate = currentScope.countdownDate
+
+            return result
+        }
```
JayShortway, line 323: Nice!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
```diff
@@ -406,6 +421,15 @@ internal object VariableProcessorV2 {
...
         Variable.PRODUCT_STORE_PRODUCT_NAME -> rcPackage.product.name
+
+        Variable.COUNT_DAYS_WITH_ZERO -> countdownTime?.days?.let { String.format(Locale.US, "%02d", it) } ?: ""
+        Variable.COUNT_DAYS_WITHOUT_ZERO -> countdownTime?.days?.toString() ?: ""
+        Variable.COUNT_HOURS_WITH_ZERO -> countdownTime?.hours?.let { String.format(Locale.US, "%02d", it) } ?: ""
+        Variable.COUNT_HOURS_WITHOUT_ZERO -> countdownTime?.hours?.toString() ?: ""
+        Variable.COUNT_MINUTES_WITH_ZERO -> countdownTime?.minutes?.let { String.format(Locale.US, "%02d", it) } ?: ""
+        Variable.COUNT_MINUTES_WITHOUT_ZERO -> countdownTime?.minutes?.toString() ?: ""
+        Variable.COUNT_SECONDS_WITH_ZERO -> countdownTime?.seconds?.let { String.format(Locale.US, "%02d", it) } ?: ""
+        Variable.COUNT_SECONDS_WITHOUT_ZERO -> countdownTime?.seconds?.toString() ?: ""
```
JayShortway: I'm not sure if hardcoding `Locale.US` can bite us? We do have `dateLocale` available, maybe we should use that?

JayShortway: I'm curious about the exact behavior here:
- My target date is 2 days from now.
- I display only the `hours` variable.
- Do I:
    - get text that counts down twice from 24 to 0, or
    - get text that counts down once from 48 to 0?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentState.kt
JayShortway, line 1: This file should probably be named `CountdownComponentState` 😄

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentViewModelTests.kt
```diff
@@ -0,0 +1,105 @@
+package com.revenuecat.purchases.ui.revenuecatui.components.countdown
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class CountdownComponentViewModelTests {
```
JayShortway:
```suggestion
class CountdownComponentStateTests {
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
JayShortway: Might be good to add tests for this behavior too so we're sure it doesn't change, maybe on the `TextComponentView` level? We can see if it fits in `TextComponentViewVariableTests`, or otherwise a separate test class is fine too imo.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
tonidero: Ok so just to confirm, we won't be supporting specific local time countdowns right? As in, if we want an offer to expire at 9pm in local time of the device? I think this can be added later if needed with an extra property and some date manipulation, so not a blocker, but just to confirm.

vegaro: Right. For now the scope is all "deadlines" are in UTC

JayShortway (commented):
Great job on this! 💪 Some questions and suggestions here and there.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
tonidero: Sounds good!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/countdown/CountdownComponentState.kt
vegaro: lol yes

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
vegaro: Oh that's a good point. It would be "get text that counts down twice from 24 to 0"

vegaro: Ok, just discussed with @joshdholtz 

For now it will count twice, but we can add a future property that can change the behavior (“count down from days” or “countdown from hours”, etc)

For now the behavior will be this. I am going to see if I can add a test for this

JayShortway: Thank you! Yes I think it's good to be explicit about this with tests 😄

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
vegaro: Confirmed with @joshdholtz we don't need this. Countdown is made of Stacks and Stacks have overrides so it's not needed for countdown

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CountdownComponent.kt
vegaro: > Does ISO-8601 always contain a time zone? 

yes, right now it's just `Z`, which means UTC

> I think we should be very explicit about the time zone, even if it's always UTC. We can also use millis since epoch?

when we start supporting setting client timezones epoch will not work, so we need to use full dates (with the explicit timezones that they already have)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
vegaro: Added tests in [86f1095](https://github.com/RevenueCat/purchases-android/pull/2837/commits/86f109532a6b54f1bfcf65e1b622707bb8f76f62)

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
```diff
@@ -25,13 +27,25 @@ internal class OfferingVideoPredownloader(
...
+        }
+    }
+
+    private fun downloadVideosFromStack(stack: StackComponent) {
+        stack.components.forEach { component ->
+            when (component) {
+                is VideoComponent -> {
+                    fileRepository.prefetch(component.source.checkedUrls())
+                }
+                is CountdownComponent -> {
```
tonidero: Hmm we don't predownload for other stack like component... Just wondering, why we need to do this only for this one?

tonidero:
> Hmm it looks like [3b107f5](https://github.com/RevenueCat/purchases-android/pull/2837/commits/3b107f51dd38ef6407a085cbdf4a890aa1b83ec0) broke it and the timer is static. Checking what's going on

Hmm possibly the provider that was creater, the derivedStateOf doesn't respond to changes with that? We would need to store the countdownTime as a state so the derivedStateOf responds to changes to that I think... FWIW, I would be ok reverting to what was before, I'm not sure how to store this in the state right now 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/utils/OfferingVideoPredownloader.kt
vegaro: Oh! Interesting. I added this becuase it was being added in iOS https://github.com/RevenueCat/purchases-ios/pull/5790/files#diff-7679550051a68d69abd3dc4dbd6283add4a4835d77c9c6fccf1551840230156aR231

but now that you mention it, it looks like in iOS we are prewarming in more cases and in Android only if `VideoComponent`?

tonidero: cc @JZDesign Not sure if you have any input on this one? I don't think it's a blocker though...

vegaro: Reverted these chagnes. `downloadVideos` is actually not being called

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
vegaro: we are adding support for whoever wants to display hours as 48 case! It will come in a future PR

PR #2841, opened 2025-11-14
Add file size limit to events tracking files

tonidero:
### Description
This adds a file size limit to the events manager, so if it grows too much, it doesn't grow infinitely. This also affects paywall, and customer center.


tonidero:
> in the iOS one I separated them into two different storage files but I believe you are already aware of this.

Right, in Android, I'm creating 2 different EventManager, each with its own storage file. That's part of the base PR [here](https://github.com/RevenueCat/purchases-android/pull/2728)

PR #2843, opened 2025-11-14
Conditional Visibility

JZDesign:
### Checklist

### Motivation

Users want a way to control when to show or hide certain elements of their paywall. Commonly this is based on screen size, but there are other reasons this may occur as well.

### Description

Josh wrote the majority of the iOS PR months ago. I picked it up and added the selected package bit, then I had codex look at those changes and replicate them here in the android project.


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -50,6 +87,9 @@ internal object ConditionSerializer : SealedDeserializerWithDefault<Condition>(
         "intro_offer" to { Condition.IntroOffer.serializer() },
         "multiple_intro_offers" to { Condition.MultipleIntroOffers.serializer() },
         "selected" to { Condition.Selected.serializer() },
+        "orientation" to { Condition.Orientation.serializer() },
+        "screen_size" to { Condition.ScreenSize.serializer() },
```
JayShortway, line 141: Just curious why we need an extra `screen_size` one? We already have the `compact`, `medium` and `expanded` ones.

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
```diff
@@ -73,6 +75,28 @@ class UiConfig(
...
+        ) {
+            companion object Defaults {
+                @JvmField
+                val mobile = ScreenSize(name = "mobile", width = 375)
+
+                @JvmField
+                val tablet = ScreenSize(name = "tablet", width = 700)
+
+                @JvmField
+                val desktop = ScreenSize(name = "desktop", width = 1024)
```
JayShortway, line 94: We decided a while back to use window sizes "compact", "medium" and "expanded" instead of device sizes. Window size is not always equal to device size, and it's window size we care about.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ScreenConditionSnapshot.kt
```diff
@@ -0,0 +1,19 @@
...
+)
+
+internal enum class ScreenOrientation {
+    PORTRAIT,
+    LANDSCAPE,
+    UNKNOWN,
+}
+
+@Suppress("CompositionLocalAllowlist")
+internal val LocalScreenCondition = staticCompositionLocalOf { ScreenConditionSnapshot() }
```
JayShortway: Let's try to avoid CompositionLocals unless absolutely necessary. Most of the window metrics can be read in individual composables when needed. See for instance [StackComponentState](https://github.com/RevenueCat/purchases-android/blob/269609fff5435cc2452b686cdc915f45781739f6/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt#L54).

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
JZDesign: We are allowing custom sizing from the editor.

JZDesign: Actually… It doesn't look like we are allowing custom sizing in the editor.

I'm going off of Josh's old iOS PR, but that part is conflicting the editor. Now I have to look into things.

JayShortway: Ha alright. Just to clarify, the existing _compact, medium, expanded_ mechanism uses Compose's built-in definitions for these window sizes.

https://developer.android.com/develop/ui/compose/layouts/adaptive/use-window-size-classes

JZDesign: Okay. Yes I did understand the mission correctly.

We are going to allow for custom breakpoints. Those will be configured in the UI config. Then conditions can be written against them. 

We have some users who need to be able to fine tune their breakpoints for tiny devices, that way the we don't rely on what apple or android determine to be compact, but what the developer wants. For example, the iPhone SE is the same "compact" as the iPhone Xr, even though it's considerably smaller. For users who want to have a paywall for that specific screen size, we need to be more precise than whatever apple and google say.

If you see great issue with it then we can hash things out and go back to the drawing board.

JZDesign: The editor & Khepri portions of this are not done, but we fall back to the 3 defaults in the sdk so we can add that functionality later.

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
JZDesign: The nomenclature of ScreenSize is for user understanding. We use the window size under the hood.

I think the decision you're referring to is now outdated.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
JayShortway: Ahh okay, custom breakpoints is nice! Maybe we should remove `compact`, `medium` and `expanded` in favor of `screen_size` in that case, to avoid ourselves getting confused?

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -17,39 +18,112 @@ class ComponentOverride<T : PartialComponent>(
...
     }
 }
 
 @OptIn(InternalRevenueCatAPI::class)
 internal object ConditionSerializer : SealedDeserializerWithDefault<Condition>(
     serialName = "Condition",
     serializerByType = mapOf(
-        "compact" to { Condition.Compact.serializer() },
-        "medium" to { Condition.Medium.serializer() },
-        "expanded" to { Condition.Expanded.serializer() },
```
vegaro, line 49: I confirmed there are no paywalls with these in the DB and they are safe to remove

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -17,39 +18,112 @@ class ComponentOverride<T : PartialComponent>(
...
+            @SerialName("in")
+            IN,
+
+            @SerialName("not_in")
+            NOT_IN,
+        }
+
+        @Serializable
+        enum class EqualityOperatorType {
+            @SerialName("=") // WIP… Should we make this human language from the server like in and not_in?
```
tonidero: Hmm could be `eq` and `not_eq` to have the same style as the other operators indeed... but not a strong preference.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -17,39 +18,112 @@ class ComponentOverride<T : PartialComponent>(
...
+        /**
+         * Compares the selected package's intro-offer eligibility (not the whole paywall) against [value].
+         * This matches the package the customer currently has highlighted in the UI.
+         *
+         * Default values provide backward compatibility with legacy JSON that only has `{"type": "intro_offer"}`.
+         */
         @Serializable
-        object Compact : Condition
+        data class IntroOffer(
+            @SerialName("operator") val operator: EqualityOperatorType = EqualityOperatorType.EQUALS,
```
tonidero, line 29: Hmm I see we've split the operators between `EqualityOperatorType` and `ArrayOperatorType` and have different conditions have different supported types. I'm wondering if we should unify these conditional logic, so it's the same for all conditions (`=`/`!=` can be represented with `in`/`not_in` for example). Then, both the operator and value could be part of the `Condition` interface, and each type can just pass those in.
This might simplify the condition logic down the line?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ScreenConditionSnapshot.kt
```diff
@@ -0,0 +1,15 @@
+package com.revenuecat.purchases.ui.revenuecatui.components
+
+import com.revenuecat.purchases.UiConfig
+
+internal data class ScreenConditionSnapshot(
+    val condition: ScreenCondition = ScreenCondition.COMPACT,
+    val orientation: ScreenOrientation = ScreenOrientation.UNKNOWN,
+    val screenSize: UiConfig.AppConfig.ScreenSize? = null,
```
tonidero: Hmm I'm a bit confused with having both `condition` and `screenSize`. What's the difference? And maybe we would like to add some docs explaining the difference if we do need to have both.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -17,39 +18,112 @@ class ComponentOverride<T : PartialComponent>(
...
-        "medium" to { Condition.Medium.serializer() },
-        "expanded" to { Condition.Expanded.serializer() },
         "intro_offer" to { Condition.IntroOffer.serializer() },
         "multiple_intro_offers" to { Condition.MultipleIntroOffers.serializer() },
+        "introductory_offer_available" to { Condition.AnyIntroOffer.serializer() },
+        "multiple_intro_offers_available" to { Condition.AnyMultipleIntroOffers.serializer() },
         "selected" to { Condition.Selected.serializer() },
+        "orientation" to { Condition.Orientation.serializer() },
+        "screen_size" to { Condition.ScreenSize.serializer() },
+        "selected_package" to { Condition.SelectedPackage.serializer() },
```
tonidero, line 142: What's the difference between `selected` and `selected_package`? And also between `intro_offer` and `introductory_offer_available` + `multiple_intro_offers` and `multiple_intro_offers_available`? Feels like a rename to support the new condition system, but basically this also means it would default to `unsupported` in older versions of the SDK... Not sure if that's worth it?

If I'm wrong and they are supposed to be different conditions, we might want to document the difference

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
```diff
@@ -41,6 +41,11 @@ internal fun rememberUpdatedStackComponentState(
         style = style,
         selectedPackageProvider = { paywallState.selectedPackageInfo?.rcPackage },
         selectedTabIndexProvider = { paywallState.selectedTabIndex },
+        screenConditionProvider = { paywallState.screenConditionSnapshot },
+        introOfferAvailability = IntroOfferAvailability(
```
tonidero, line 45: Hmm I'm wondering, does this need to be updated when the selected package changes? (since the eligibility for that new package may have changed). 

Just wondering if this should be within a "provider" as well, or even unify with the `Package` that is selected in a new data class.

tonidero: Also, something we should confirm is that this uses the correct selected package in each tab, since the selected package would depend on the tab in some paywalls.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentState.kt
```diff
@@ -42,32 +47,36 @@ private fun rememberUpdatedCarouselComponentState(
     style: CarouselComponentStyle,
     selectedPackageProvider: () -> Package?,
     selectedTabIndexProvider: () -> Int,
+    screenConditionProvider: () -> ScreenConditionSnapshot,
+    introOfferAvailability: IntroOfferAvailability = IntroOfferAvailability(),
```
tonidero: Hmm I'm wondering, should we remove the default parameter here? For consistency and to make sure we remember to pass that parameter in when needed.
```suggestion
    introOfferAvailability: IntroOfferAvailability,
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ScreenConditionState.kt
```diff
@@ -0,0 +1,110 @@
...
+) {
+    private var breakpoints: List<UiConfig.AppConfig.ScreenSize> = resolveBreakpoints(screenSizes)
+    private var layoutWidthDp: Float? = null
+    private var layoutHeightDp: Float? = null
+    private var windowWidthSizeClass: WindowWidthSizeClass = WindowWidthSizeClass.COMPACT
+
+    var snapshot by mutableStateOf(ScreenConditionSnapshot())
+        private set
+
+    fun updateLayoutSize(widthDp: Float, heightDp: Float) {
```
tonidero, line 27: Hmm I'm not in love with making all these methods available to callers, since they could potentially be called somewhere else, when they are only intended to be called by the `rememberScreenConditionState` and `trackScreenCondition` below... But I'm not sure what would be the best approach... 

We could make `rememberScreenConditionState` a companion object function, and would allow us to make some of these private, but wouldn't work for `trackScreenCondition` since that's an extension function of `Modifier`... So I think it's ok to leave as is for now...

tonidero (commented):
Still reviewing but posting what I have so far

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
JZDesign: Turns out, Khepri requires this. It's used a lot in other areas.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
JZDesign: Not entirely sure I understand. Would you show me a simple code example of what you mean exactly?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ScreenConditionSnapshot.kt
JZDesign: Ah… I can remove condition

JZDesign: I think

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
JZDesign: Good questions. Long story short, our foresight was only kind of right when we started this work many months ago. As development continued, other scenarios came up, and the use case for the keys that were already in the code could have been clearer but these are the meanings. I have a doc on this that I can throw into notion, or perhaps this is a safe enough one that could live in the repo… though it would be duplicated.

|Key|Purpose|
|:--:|:--:|
|`selected`| The component itself is selected, like a tab, a package, a switch, etc.,|
|`selected_package`| The selected package matches the given array conditions|
|`intro_offer`|The selected package includes an intro offer you are eligible for|
|`introductory_offer_available`|any package in the paywall includes an introductory offer you are eligible for, selected or not|
|`multiple_intro_offers`|The selected package has multiple intro offers that you are eligible for|
|`multiple_intro_offers_available`|any package in the paywall includes multiple intro offers you are eligible for, selected or not|
|`app_version`|The consuming application version matches the given comparison|

> [!NOTE]
> Edited, adding application version to allow for fine tuned control for various reasons like being able to add new conditions that are not supported by older SDKs and giving the user the ability to ensure the paywall renders appropriately for their use case.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
JZDesign: I'll test this more thoroughly today. I thought it was working as expected, but I'll triple check to be sure.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
tonidero: Hmm right... nvm, I think it would be much worse since we would lose a lot of processing already done by the parsing... For posterity, I was thinking on something like this:
```
    sealed interface Operator {
        @Serializable
        @SerialName("not_in")
        object NOT_IN : Operator

        @Serializable
        @SerialName("in")
        object IN : Operator
        
        ....
    }

    sealed interface Condition {
        val operator: Operator
        val value: List<String>

        @Serializable
        data class IntroOffer(
            @SerialName("operator") override val operator: Operator = Operator.IN,
            @SerialName("value") override val value: List<String> = listOf("true"),
        ) : Condition

        ....
    }
```

But I think that way we would then need to parse the value from a list of strings, to whatever type each condition has and would probably be more logic on that side... Better to make the parsing expect a narrower type so we don't need any extra processing. 

So yeah, please ignore me 😓

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
tonidero: Ohh thanks for the explanation. That makes sense! Then, I would indeed add that as docs to the repo, I think it should be fine, and could help understand the difference between them.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentState.kt
tonidero: After your previous explanation in the previous comment, I believe this might be right though. I did not understand the new types completely.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
```diff
@@ -45,16 +45,17 @@ import dev.drewhamilton.poko.Poko
...
     initialDarkMode: Boolean,
     initialLayoutDirection: LayoutDirection,
     private val style: VideoComponentStyle,
     private val localeProvider: () -> Locale,
     private val selectedPackageProvider: () -> com.revenuecat.purchases.Package?,
     private val selectedTabIndexProvider: () -> Int,
+    private val introOfferAvailability: IntroOfferAvailability,
 ) {
-    private var windowSize by mutableStateOf(initialWindowSize)
+    private val screenConditionSnapshot = initialScreenCondition
```
vegaro: I believe you can simplify this by making the constructor parameter directly a `private val` and removing this redundant property assignment. 

```
internal class VideoComponentState(
...
    private val initialScreenCondition: ScreenConditionSnapshot,
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ScreenConditionState.kt
```diff
@@ -0,0 +1,110 @@
...
+
+internal class ScreenConditionState(
+    screenSizes: List<UiConfig.AppConfig.ScreenSize>?,
+) {
+    private var breakpoints: List<UiConfig.AppConfig.ScreenSize> = resolveBreakpoints(screenSizes)
+    private var layoutWidthDp: Float? = null
+    private var layoutHeightDp: Float? = null
+    private var windowWidthSizeClass: WindowWidthSizeClass = WindowWidthSizeClass.COMPACT
+
+    var snapshot by mutableStateOf(ScreenConditionSnapshot())
```
vegaro: I am a bit confused by the name of this variable. Actually by the usage of "snapshot" in general like in `ScreenConditionSnapshot`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -137,6 +145,12 @@ internal sealed interface PaywallState {
                 ?.pkg
             private val packagesOutsideTabs: Set<Package> = packages.packagesOutsideTabs
                 .mapTo(mutableSetOf()) { it.pkg }
+            private val allAvailablePackages: Set<Package> = mutableSetOf<Package>().apply {
```
vegaro: super nit, but can you add an empty line above and after this property? I also noticed `packagesOutsideTabs` is missing one above. Thank you!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -137,6 +145,12 @@ internal sealed interface PaywallState {
...
+                addAll(packagesOutsideTabs)
+                packages.packagesByTab.values.forEach { infoList ->
+                    infoList.forEach { add(it.pkg) }
+                }
+            }
```
vegaro: This whole thing can also be achieved with the following:

```kotlin
private val allAvailablePackages: Set<Package> = 
    packagesOutsideTabs + packages.packagesByTab.values.flatten().map { it.pkg }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedPartial.kt
```diff
@@ -92,42 +102,113 @@ internal fun <T : PresentedPartial<T>> List<PresentedOverride<T>>.buildPresented
 
 @Suppress("ReturnCount")
 private fun <T : PresentedPartial<T>> PresentedOverride<T>.shouldApply(
-    windowSize: ScreenCondition,
-    introOfferEligibility: IntroOfferEligibility,
+    screenCondition: ScreenConditionSnapshot,
+    introOfferSnapshot: IntroOfferSnapshot,
+    state: ComponentViewState,
+    selectedPackageIdentifier: String?,
+): Boolean = this.conditions.all { condition ->
```
vegaro, line 109: There's an issue with doing `all` here. If we add support for new conditions, and the dev uses them, this will never return true, because `Unsupported` returns `false`. @facumenzella @kukukaka and I were actually discussing this morning about this, and we thought that developers probably want to ignore the new rules in this cases right?

JZDesign: Should we filter out Unsupported then?

JZDesign: Or have unsupported return true?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/timeline/TimelineComponentState.kt
```diff
@@ -39,32 +44,36 @@ private fun rememberUpdatedTimelineComponentState(
     style: TimelineComponentStyle,
     selectedPackageProvider: () -> Package?,
     selectedTabIndexProvider: () -> Int,
+    screenConditionProvider: () -> ScreenCondition,
+    introOfferAvailability: IntroOfferAvailability = IntroOfferAvailability(),
```
tonidero, line 48: Same here. We probably should remove the default `IntroOfferAvailability`, probably a leftover? :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
```diff
@@ -235,12 +243,10 @@ internal class VideoComponentState(
 
     @JvmSynthetic
     fun update(
-        windowSize: WindowWidthSizeClass? = null,
```
tonidero, line 238: Hmm I'm not sure this is right... in the other components states, the `update` method includes the `screenCondition`, so we can update it, but not for the VideoComponentState. Is there a reason for that difference? This also means that we will keep using the `initialScreenCondition` for the `PresentedPartial`  even if the screen changes, so not sure if that's right... Maybe we're handling that in the view layer?

tonidero (commented):
Only that last VideoComponentState question, but it's looking great!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentState.kt
JZDesign: Ah, this here isn't actually the issue but I'm glad it flagged things for you, because there was an issue that I fixed where build partial was called

PR #2844, opened 2025-11-14
Add `countFrom` support for `CountdownComponent`

vegaro:
Follow up on #2837 

Adds `count_from` support so devs can specify if they want to show for example 48 hours when the countdown ends in 2 days.

Example with `count_from` set to minutes and 6 days are remaining:

image


joshdholtz (approved):
:shipit: :shipit: :shipit:

PR #2854, opened 2025-11-18
Make events manager be supported in Android < 24

tonidero:
### Description
Originally we used streams which weren't supported in Android < 24, but that was removed in https://github.com/RevenueCat/purchases-android/pull/1943 and https://github.com/RevenueCat/purchases-android/pull/1944. However, we have some leftover code checking for the android version to create the events manager itself. This removes that, which makes events now available for older android versions.


ajpallares (approved):
Makes sense!

PR #2855, opened 2025-11-18
Track `connection_error_reason` property in diagnostics

tonidero:
### Description
This adds a new `connection_error_reason` to the `http_request_performed` diagnostics event, with possible values `TIMEOUT`, `NO_NETWORK`, `OTHER`.


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ConnectionErrorReason.kt
```diff
@@ -0,0 +1,22 @@
...
+    NO_NETWORK,
+    OTHER,
+    ;
+
+    companion object {
+        fun fromIOException(ioException: IOException): ConnectionErrorReason {
+            return when (ioException) {
+                is java.net.SocketTimeoutException -> TIMEOUT
+                is java.net.ConnectException,
+                is java.net.UnknownHostException,
```
rickvdl, line 16: I guess this also covers DNS issues, which is fine I guess, I'll see if I can cover the same set of codes here

tonidero: Yeah indeed... I tried to get more granularity, but couldn't find a way... I think this is ok for now and we can revisit if needed

PR #2857, opened 2025-11-19
[MON-1122] Revert variable rounding logic to not round up

polpielladev:
When shipping Paywalls v2, there was a regression with the way that we show prices. We changed from not rounding or rounding down to actually rounding up.

Where this is specially important is when a dev picks a price that is explicitly designed to get just under a dollar threshold, like $83.99, which is either 6.99 or 7.00 — and 6.99 is the actual price they're aiming to put in front of customers.

### Checklist

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -421,29 +421,21 @@ internal class PurchasesFactory(
         legacyEventsFileHelper: EventsFileHelper<PaywallStoredEvent>?,
         fileHelper: EventsFileHelper<BackendStoredEvent>,
         baseURL: URL,
-    ): EventsManager? {
-        // RevenueCatUI is Android 24+ so it should always enter here when using RevenueCatUI.
-        // Still, we check for Android N or newer since we use Streams which are 24+ and the main SDK supports
-        // older versions.
-        return if (isAndroidNOrNewer()) {
-            EventsManager(
-                legacyEventsFileHelper = legacyEventsFileHelper,
-                fileHelper = fileHelper,
-                identityManager = identityManager,
-                eventsDispatcher = eventsDispatcher,
-                postEvents = { request, onSuccess, onError ->
-                    backend.postEvents(
-                        paywallEventRequest = request,
-                        baseURL = baseURL,
-                        onSuccessHandler = onSuccess,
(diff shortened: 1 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/utils/PriceFactory.kt
```diff
@@ -19,15 +21,19 @@ internal object PriceFactory {
         locale: Locale,
     ): Price {
         val currency = Currency.getInstance(currencyCode)
+        val digits = currency.defaultFractionDigits.coerceAtLeast(0)
+
+        val divisor = Math.pow(DECIMAL_BASE, digits.toDouble())
+        val valueInCurrency = amountMicros / MICRO_MULTIPLIER
+        val truncatedValue = Math.floor(valueInCurrency * divisor) / divisor
```
tonidero: Should we `round` instead of `floor`? Just in case there was some kind of floating point error and this truncating to a different cent than it should?

Aside from that, could be nice to add an extension function on double that just passes the number of decimals we want, and rounds to that number... But just a possible cleanup, this looks good!

polpielladev: Hey!! 👋 Good point! Unfortunately, using `round` would mean that we always round to the closest integer, but we actually always want to round down (or `floor`) to truncate and fix the issue.

Lmk if that makes sense, I have also moved the code to an extension that we can reuse throughout the codebase too.

polpielladev: If you're happy with the approach, I will go ahead and approve the emerge snapshots

tonidero: So just to confirm, if due to some floating point error, we have something like `1.999999999...`, this would round to `1.99`, but maybe we would want to round to `2.00` in this case? Granted this shouldn't usually happen... 

I guess the question is, in what cases we want to floor, that wouldn't work using round? If we want X decimals, the X+1 decimal should always be due to some floating point error, so rounding to the closest feels safer... Happy to discuss it more of course!

tonidero: We discussed this internally, this should be ok since it's only used for the pricePerPeriod calculations, and in this case, we do want to truncate the decimals, so this is ok!

PR #2860, opened 2025-11-20
Fix backend integration tests

tonidero:
### Description
Seems the backend integration tests were flaky after adding the timeouts in #2807. This is because the timeout was too short during tests, causing it to be hit frequently.

### DIFF
--- a/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
@@ -1,7 +1,6 @@
 {
   "statusCode": 200,
   "headers": {
-    "Accept-Ranges": "bytes",
     "Connection": "keep-alive",
     "Content-Length": "87466",
     "Server": "cloudflare",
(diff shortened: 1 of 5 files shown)


purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
```diff
@@ -1,6 +1,7 @@
 {
   "statusCode": 200,
   "headers": {
+    "Accept-Ranges": "bytes",
```
rickvdl: Do you know why this changed?

ajpallares: I'm curious. Wasn't this field received before?

tonidero: Good question, I want to ask this to the COIN team, but it indeed seems that something we started receiving... I wonder if we should start doing an allowlist, instead of a blocklist for headers... But I think it's good to be aware of what we receive

tonidero: See https://github.com/RevenueCat/purchases-android/pull/2860#discussion_r2545672998

tonidero: Seems it's not always sent... So I'm just going to ignore this header for now.

PR #2861, opened 2025-11-20
Select default package on Sheet dismissal

vegaro:
Equivalent of https://github.com/RevenueCat/purchases-ios/pull/5797

Setting the default package back to default after dismissing Sheet

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
@@ -265,6 +265,13 @@ internal sealed interface PaywallState {
                 if (currentTabContainsThisPackage) selectedPackageByTab[currentTabIndex] = selectedPackage
             }
 
+            fun resetToDefaultPackage() {
+                selectedPackage =
+                    packages.packagesByTab[selectedTabIndex]?.firstOrNull { it.isSelectedByDefault }?.pkg
+                        ?: initialSelectedPackageOutsideTabs
+                        ?: selectedPackageByTab[selectedTabIndex]
+            }
+
             private fun LocaleList.toLocaleId(): LocaleId {
                 val preferredOverride = purchases.preferredUILocaleOverride
                 val deviceLocales = map { it.toLocaleId() }.plus(locales.head)
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedPaywallComponents.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedPaywallComponents.kt
@@ -135,7 +135,6 @@ private fun SimpleSheetState.show(
     state: PaywallState.Loaded.Components,
     onClick: suspend (PaywallAction) -> Unit,
 ) {
-    val previouslySelectedPackage = state.selectedPackageInfo?.rcPackage
     show(
         backgroundBlur = sheet.backgroundBlur,
         content = {
@@ -154,9 +153,7 @@ private fun SimpleSheetState.show(
             )
         },
         onDismiss = {
-            // Restore the previously selected package when sheet is dismissed to prevent
-            // purchasing a hidden package that was selected in the sheet
-            previouslySelectedPackage?.let { state.update(selectedPackage = it) }
+            state.resetToDefaultPackage()
         },
     )
 }


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -265,6 +265,12 @@ internal sealed interface PaywallState {
                 if (currentTabContainsThisPackage) selectedPackageByTab[currentTabIndex] = selectedPackage
             }
 
+            fun getDefaultSelectedPackage(): Package? {
+                return initialSelectedPackageOutsideTabs
+                    ?: selectedPackageByTab[selectedTabIndex]
+                    ?: packages.packagesByTab[selectedTabIndex]?.firstOrNull()?.pkg
```
tonidero: I don't have a lot of context, but should this logic be inverted? Like:

```suggestion
                return packages.packagesByTab[selectedTabIndex]?.firstOrNull()?.pkg
                    ?: selectedPackageByTab[selectedTabIndex]
                    ?: initialSelectedPackageOutsideTabs
```

Just thinking that we should try to keep the previously selected package within the tab after dismissing the sheet? Not sure if this makes a lot of sense though, I think I'm missing context on the issue.

vegaro: Let me record a video and debug here but if I understand correctly, the tab is the same within the sheet, so `selectedPackageByTab`​ will be actually the one in the sheet? And we want to as soon as the sheeet is closed, select the default package. Let me record a video and see if I can explain it better. And also debug because I might be wrong

vegaro: Oh I think I get what you are saying now after doing some debugging. You're saying if we should select back what it was selected before opening the sheet.  
  
So we have decided to just select the default package back when closing the sheet. That's why `initialSelectedPacakageOutsideTabs`​ takes preference. That's what I've done in iOS as well.  
  
One thing though... debugging a bit closely I am noticing that `selectedPackageByTab`​ is empty and `packagesByTab`​ as well. I find that strange... but maybe that only exists if there are tabs? In the Paywall I am using for testing I just have a package list and a button that opens the sheet

tonidero: > selectedPackageByTab​ is empty and packagesByTab​ as well

I do think that's normal if the paywall doesn't have tabs. Those are only used in that case. 

> So we have decided to just select the default package back when closing the sheet.

Ok, we should probably test what happens in case we have tabs with different packages on each tab (to make sure we return to the default tab + default package within that tab. But I think in general the code makes sense!

JayShortway: Can confirm that the `*ByTab` things are only used if the paywall has tabs.

vegaro: made a little change and the logic is now:
- First select default pacakge in tab
- If null select default package outside of tabs

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -265,6 +265,13 @@ internal sealed interface PaywallState {
                 if (currentTabContainsThisPackage) selectedPackageByTab[currentTabIndex] = selectedPackage
             }
 
+            fun resetToDefaultPackage() {
+                selectedPackage =
+                    packages.packagesByTab[selectedTabIndex]?.firstOrNull { it.isSelectedByDefault }?.pkg
+                        ?: initialSelectedPackageOutsideTabs
+                        ?: selectedPackageByTab[selectedTabIndex]
```
tonidero, line 272: Should this also reset the selected tab to the default?

vegaro: you mean if we reach the `selectedPackageByTab[selectedTabIndex]`?

tonidero: Well, I was actually not thinking about the package, but about the tab index: https://github.com/RevenueCat/purchases-android/blob/41a5c9102121cdfbf874c5af48e6e632e680ad71/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt#L203. 

Mostly thinking of a case like this:
- A paywall with 2 tabs, each with different packages. The customer moves to a non-default tab and selects a package.
- Then, it opens a sheet that also has a package itself, the selected package is now that.
- Then, the user dismisses the sheet.

Should it be kept in the selected tab, instead of the default? Just thinking, since we're resetting the selected package in each tab to the default, we might also want to reset the selected tab... But that might not be what we want, in which case, this is ok I think.

vegaro: oh I see. Thanks for the explanation. We want to keep the tab that was open before opening the sheet, so no defaulting the tab.

Issue #2606, opened 2025-08-16
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'void com.android.okhttp.internal.http.HttpStream.writeRequestHeaders(com.android.okhttp.Request)' on a null object reference


wonderbook-dev:
Saw this crash in Firebase this morning. Seems SDK issue.

SDK: 9.3.0
device: Android 15 / Redmi 14C

Small number of devices.

 ```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke interface method 'void com.android.okhttp.internal.http.HttpStream.writeRequestHeaders(com.android.okhttp.Request)' on a null object reference
       at com.android.okhttp.internal.http.HttpEngine.readResponse(HttpEngine.java:606)
       at com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:475)
       at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getResponse(HttpURLConnectionImpl.java:411)
       at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getErrorStream(HttpURLConnectionImpl.java:154)
       at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.getErrorStream(DelegatingHttpsURLConnection.java:98)
       at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.getErrorStream(HttpsURLConnectionImpl.java:30)
       at com.revenuecat.purchases.common.HTTPClient.getInputStream(HTTPClient.kt:80)
       at com.revenuecat.purchases.common.HTTPClient.performCall(HTTPClient.kt:178)
       at com.revenuecat.purchases.common.HTTPClient.performRequest(HTTPClient.kt:126)
       at com.revenuecat.purchases.common.HTTPClient.performRequest$default(HTTPClient.kt:104)
       at com.revenuecat.purchases.common.Backend$getCustomerInfo$call$1.call(Backend.kt:163)
       at com.revenuecat.purchases.common.Dispatcher$AsyncCall.run(Dispatcher.kt)
       at com.revenuecat.purchases.common.Dispatcher.enqueue$lambda$2$lambda$1(Dispatcher.kt:67)
       at com.google.firebase.appcheck.internal.DefaultFirebaseAppCheck$$ExternalSyntheticLambda0.run(R8$$SyntheticClass:31)
[...]

PR #2865, opened 2025-11-24
Handle error reading `errorStream` in some devices

tonidero:
### Description
We've received reports of some crashes in some devices in what seems to be an issue with OkHTTP: https://github.com/RevenueCat/purchases-android/issues/2606. While there isn't much we can do, since it seems more device dependent, these changes makes it so we don't crash and instead give an error back to the developer


ajpallares (approved):
Nice! 👍

PR #2866, opened 2025-11-24
Adds Maestro test making a purchase on a v2 Paywall

JayShortway:
## Description
**Note:** based on https://github.com/RevenueCat/purchases-android/pull/2788. 

Adds a Maestro test that makes a Google Play purchase and immediately revokes it with the RevenueCat API, to allow testing again.

This can be merged, but it's just a POC. Still to do:
- aligning this with the existing iOS tests, and
- running this on CI.

### DIFF
--- a/maestro/rc-api/v1/api.js
+++ b/maestro/rc-api/v1/api.js
@@ -1,0 +1,22 @@
+for (var EXPECTED_PARAMS = ['apiKey'], i = 0; i < EXPECTED_PARAMS.length; i++) {
+  if (!eval(EXPECTED_PARAMS[i])) throw new Error(EXPECTED_PARAMS[i] + ' is not provided');
+}
+
+function revokeGooglePlaySubscription(appUserId, productId) {
+  return http.post(
+    'https://api.revenuecat.com/v1/subscribers/' +
+      encodeURIComponent(appUserId) +
+      '/subscriptions/' +
+      encodeURIComponent(productId) +
+      '/revoke', {
+        headers: {
+          'Content-Type': 'application/json',
+          'Authorization': 'Bearer ' + apiKey,
+        },
+        body: ''
+      });
+}
+
+output.rcApiV1 = {
+  revokeGooglePlaySubscription: revokeGooglePlaySubscription
+}
--- a/maestro/test_paywall_v2.yaml
+++ b/maestro/test_paywall_v2.yaml
@@ -1,0 +1,35 @@
+appId: com.revenuecat.automatedsdktests
+env:
+  PRODUCT_ID: subscription_monthly
+onFlowStart:
+  - runScript:
+      when:
+        platform: Android
+      label: "Load RevenueCat API v1"
+      file: ./rc-api/v1/api.js
+      env:
+        apiKey: ${RC_SECRET_API_KEY_V1}
+  - runFlow:
+      when:
+        platform: Android
+      label: "Log in with a Google account"
+      file: google-login/google-login.yml
+---
+- launchApp:
+    appId: com.revenuecat.automatedsdktests
+    clearState: true
+- tapOn: Offerings
+- tapOn: default
+- tapOn: Show Paywall
+- assertVisible: Paywall V2
+- tapOn: Monthly
+- tapOn: Continue
+- runFlow:
+    label: "Subscribe"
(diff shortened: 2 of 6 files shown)


maestro/rc-api/v1/api.js
```diff
@@ -0,0 +1,22 @@
...
+          'Content-Type': 'application/json',
+          'Authorization': 'Bearer ' + apiKey,
+        },
+        body: ''
+      });
+}
+
+output.rcApiV1 = {
+  revokeGooglePlaySubscription: revokeGooglePlaySubscription
+}
```
JayShortway, line 22: This follows the [Shared Functions](https://docs.maestro.dev/advanced/javascript/make-http-s-requests#shared-functions) approach from the Maestro docs, which makes calling these APIs convenient from Maestro flows. We can add more RC API v1 calls to this file as needed.

test-apps/e2etests/build.gradle.kts
```diff
@@ -9,7 +9,7 @@ android {
     compileSdk = 36
 
     defaultConfig {
-        applicationId = "com.revenuecat.e2etests"
+        applicationId = "com.revenuecat.automatedsdktests"
```
JayShortway, line 12: What shall we do here? Some options:
1. Change the application ID in the Production Integration Tests Android RC project.
2. Remove the Production Integration Tests Android RC project in favor of the Automated SDK Testing RC project.
3. Make the application ID configurable from the command line.

I think option 2 has my preference.

tonidero: Hmm I think we don't use `com.revenuecat.e2etests` anywhere right now? (it works since it used to use the test store). But yeah, we do need a real package name for tests that test the Google flow.

I'm fine with option 2, but we will need to change the android backend integration tests as well to use the new API key and see if there are any changes. Additionally, for changing the "Purchases integration tests" it will likely require changing a bunch of environment variables.

But in any case, I do think option 2 makes the most sense, even if it's a bit more work 👍

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/main/CustomerInfoPage.kt
```diff
@@ -203,11 +209,25 @@ private fun InfoSection(
...
+        Text(
+            text = value,
+            modifier = Modifier.semantics {
+                // Setting the label as testTag for the value Text is on purpose. It allows us to do thinks like this
+                // in Maestro:
+                // - copyTextFrom:
+                //    id: "Original App User ID"
+                // to extract the value.
+                testTag = label
+                testTagsAsResourceId = true
```
tonidero, line 229: 😍

maestro/test_paywall_v2.yaml
```diff
@@ -0,0 +1,35 @@
...
+- tapOn: Show Paywall
+- assertVisible: Paywall V2
+- tapOn: Monthly
+- tapOn: Continue
+- runFlow:
+    label: "Subscribe"
+    file: subscribe/subscribe.yml
+- tapOn: Customer Info
+- assertVisible: "pro: "
+- assertVisible: "${'Active: true, Product: ' + PRODUCT_ID}"
```
rickvdl, line 32: This is something would be nice to align with iOS

maestro/test_paywall_v2.yaml
```diff
@@ -0,0 +1,35 @@
...
+- copyTextFrom:
+    id: "Original App User ID"
+- evalScript: ${output.rcApiV1.revokeGooglePlaySubscription(maestro.copiedText, PRODUCT_ID)}
```
rickvdl, line 35: Very cool!

maestro/rc-api/v1/api.js
```diff
@@ -0,0 +1,22 @@
+for (var EXPECTED_PARAMS = ['apiKey'], i = 0; i < EXPECTED_PARAMS.length; i++) {
+  if (!eval(EXPECTED_PARAMS[i])) throw new Error(EXPECTED_PARAMS[i] + ' is not provided');
+}
+
+function revokeGooglePlaySubscription(appUserId, productId) {
+  return http.post(
```
rickvdl, line 6: Great that you've managed to get this to work with the limited scripting abilities

PR #2867, opened 2025-11-24
Restore Purchases config automatically in CustomerCenter

facumenzella:
### Motivation & Description
This is a follow-up of https://github.com/RevenueCat/purchases-android/pull/1872

We're getting some errors like this one in flutter:

```
Purchases$Companion.getSharedInstance
pf.K - There is no singleton instance. Make sure you configure Purchases before trying to get the default instance.
```

The theory is that this might be related to a similar bug we fixed for paywalls of the app getting backgrounded and killed while on the CustomerCenter activity, and then crashing when coming back


tonidero (approved):
It looks great to me! Thanks for the fix!

PR #2871, opened 2025-11-25
Extract parameters for non paid revenue tracking API to use objects

tonidero:
### Description
This is a follow-up from #2728 

In iOS, we're passing an object to the track methods for non paid revenue tracking, but we were passing the individual parameters in Android. After thinking about it, passing an object allows for better extensibility, so modifying the API for android.

This is technically a breaking change, but the API is still internal, so it should be ok.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
@@ -27,11 +27,15 @@ import com.revenuecat.purchases.utils.Result as RCResult
 @OptIn(InternalRevenueCatAPI::class)
 internal class FontLoader(
     private val context: Context,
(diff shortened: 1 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/types/AdDisplayedData.kt
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @property networkName The name of the ad network.
+ * @property mediatorName The name of the ad mediator. See [AdMediatorName] for common values.
+ * @property placement The placement of the ad, if available.
+ * @property adUnitId The ad unit ID.
+ * @property impressionId The impression ID.
+ */
+@InternalRevenueCatAPI
+@Poko
+class AdDisplayedData(
```
tonidero, line 17: I did wonder about adding a builder for all of these... But maybe it's ok for now since the parameters are pretty simple IMO. We can add it later if it's needed I guess

PR #2872, opened 2025-11-25
Add AIDL to query purchase history inapp products

tonidero:
### Description
With BC8, we lost the ability of querying consumed one time products, which sometimes we needed to unlock entitlements correctly on a restore, in cases where the products were accidentally consumed. This uses the old AIDL service to be able to query those products instead of using the billing library.


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -359,18 +367,73 @@ internal class BillingWrapper(
...
-                appInBackground,
-            ),
-            onReceivePurchaseHistory,
-            onReceivePurchaseHistoryError,
-            ::withConnectedClient,
-            ::executeRequestOnUIThread,
-        ).run()
+
+        // For INAPP, use PurchaseHistoryManager (AIDL)
+        if (productType == BillingClient.ProductType.INAPP) {
```
tonidero: Note that I changed to use the AIDL only for INAPP purchases since I don't think we care (other than for importing devs from previous migrations) about expired subscriptions. Lmk if you think we should also use it for SUBS for completeness. I was mostly trying to minimize the aspects where this is necessary.

tonidero:
This shouldn't be merged yet, but planning to make a beta build with these changes so we can test them more internally. I would appreciate a review before then @RevenueCat/sdk @RevenueCat/catforms

tonidero:
Moving back to draft to put everything behind a feature flag

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+            if (!result.isSuccess()) {
+                errorLog { "Error querying purchase history through AIDL: ${result.getResponseCodeString()}" }
+                break
+            }
+
+            allRecords.addAll(result.records)
+            continuationToken = result.continuationToken
+
+            debugLog { "Retrieved ${result.records.size} records from AIDL queryPurchaseHistory" }
+        } while (continuationToken != null)
```
tonidero: I believe you already had pagination done in your PoC @JayShortway ? Not sure if you meant something else though.

purchases/build.gradle.kts
```diff
@@ -71,6 +72,11 @@ android {
             name = "ENABLE_EXTRA_REQUEST_LOGGING",
             value = (localProperties["ENABLE_EXTRA_REQUEST_LOGGING"] as? String ?: "false"),
         )
+        buildConfigField(
+            type = "boolean",
+            name = "ENABLE_QUERY_PURCHASE_HISTORY_AIDL",
+            value = (localProperties["ENABLE_QUERY_PURCHASE_HISTORY_AIDL"] as? String ?: "false"),
```
tonidero: Note that this is currently disabled by default, but we still include the AIDL in the SDK, so we should confirm it wouldn't cause issues with automated app reviews mostly.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+
+        if (!shouldStart) {
+            return deferred.await()
+        }
+
+        return try {
+            val result = operation()
+            deferred.complete(result)
+            result
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
```
JayShortway: I think this should catch `Throwable`, which is even more generic.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+            val response = billingService!!.getPurchaseHistory(
+                BillingConstants.BILLING_API_VERSION,
+                context.packageName,
+                type,
+                continuationToken,
+                extraParams,
+            )
+
+            parseResponse(response)
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
```
JayShortway: Same here, I think we should catch `Throwable` here.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
JayShortway: Oh right, but it's blocking. Maybe `performQuery` should be `suspend`? I see it's already called from a suspend function, `queryAllPurchaseHistory`.

tonidero: Right, it can be made suspend... but would still execute everything in the "current" context I guess... It's the responsibility of the coroutine scope to know what thread to executed this on (right now, I'm using an IO scope, to execute this in an IO thread)

tonidero: Or well, we could change to an IO context within this function true, instead of leaving it to the caller to specify the thread... I guess I preferred the caller to own that logic, since that way, everything gets executed in that same background thread... But happy to change the context here if you think it would be better.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/BillingConstants.kt
```diff
@@ -0,0 +1,53 @@
+package com.revenuecat.purchases.google.history
+
+/**
+ * Constants for the AIDL-based billing integration.
+ * These match the constants used by the Play Billing Library internally.
+ */
+internal object BillingConstants {
+    // Billing API version - use 6 for purchase history support (introduced in v6)
+    const val BILLING_API_VERSION = 6
```
ajpallares: If it's still supported by BC7, why not use v7 here?

tonidero: Hmm no reason, it does work in both, so I think it makes sense to use the latest indeed. Thanks! Updated [here](https://github.com/RevenueCat/purchases-android/pull/2872/commits/0a830ef97968c6250d78deff206d695aa50a62c2)

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseData.kt
```diff
@@ -0,0 +1,58 @@
...
+                errorLog(e) { "Error parsing AIDL purchase data JSON: $json" }
+                null
+            }
+        }
+    }
+
+    /**
+     * Get a human-readable string for the purchase state.
+     */
+    fun getPurchaseStateString(): String {
```
vegaro: I think this is not used

purchases/src/main/aidl/com/android/vending/billing/IInAppBillingService.aidl
```diff
@@ -0,0 +1,36 @@
+// IInAppBillingService.aidl
```
JayShortway:
```suggestion
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/BillingConstants.kt
```diff
@@ -0,0 +1,53 @@
+package com.revenuecat.purchases.google.history
+
+/**
+ * Constants for the AIDL-based billing integration.
+ * These match the constants used by the Play Billing Library internally.
+ */
+internal object BillingConstants {
+    // Billing API version - use 6 for purchase history support (introduced in v6)
```
JayShortway:
```suggestion
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/BillingConstants.kt
```diff
@@ -0,0 +1,53 @@
...
+    // Response codes
+    const val BILLING_RESPONSE_RESULT_OK = 0
+    const val BILLING_RESPONSE_RESULT_USER_CANCELED = 1
+    const val BILLING_RESPONSE_RESULT_SERVICE_UNAVAILABLE = 2
+    const val BILLING_RESPONSE_RESULT_BILLING_UNAVAILABLE = 3
+    const val BILLING_RESPONSE_RESULT_ITEM_UNAVAILABLE = 4
+    const val BILLING_RESPONSE_RESULT_DEVELOPER_ERROR = 5
+    const val BILLING_RESPONSE_RESULT_ERROR = 6
+    const val BILLING_RESPONSE_RESULT_ITEM_ALREADY_OWNED = 7
+    const val BILLING_RESPONSE_RESULT_ITEM_NOT_OWNED = 8
```
JayShortway: Don't we have these already defined somewhere?

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/BillingConstants.kt
```diff
@@ -0,0 +1,53 @@
...
+            BILLING_RESPONSE_RESULT_SERVICE_UNAVAILABLE -> "Service Unavailable"
+            BILLING_RESPONSE_RESULT_BILLING_UNAVAILABLE -> "Billing Unavailable"
+            BILLING_RESPONSE_RESULT_ITEM_UNAVAILABLE -> "Item Unavailable"
+            BILLING_RESPONSE_RESULT_DEVELOPER_ERROR -> "Developer Error"
+            BILLING_RESPONSE_RESULT_ERROR -> "Error"
+            BILLING_RESPONSE_RESULT_ITEM_ALREADY_OWNED -> "Item Already Owned"
+            BILLING_RESPONSE_RESULT_ITEM_NOT_OWNED -> "Item Not Owned"
+            else -> "Unknown ($responseCode)"
+        }
+    }
```
JayShortway: Maybe we can use our existing error messages instead of these?

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseData.kt
```diff
@@ -0,0 +1,58 @@
+package com.revenuecat.purchases.google.history
+
+import com.revenuecat.purchases.common.errorLog
+import org.json.JSONObject
+
+/**
+ * Data class representing a purchase from the Play Store.
+ * This is used to parse the JSON strings returned by both the Play Billing Library
+ * and the manual AIDL getPurchaseHistory call.
```
JayShortway:
```suggestion
 * This is used to parse the JSON strings returned by the AIDL getPurchaseHistory call.
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseData.kt
```diff
@@ -0,0 +1,58 @@
...
+                    packageName = jsonObject.optString("packageName", ""),
+                    productId = jsonObject.optString("productId", ""),
+                    purchaseTime = jsonObject.optLong("purchaseTime", 0),
+                    purchaseState = jsonObject.optInt("purchaseState", 0),
+                    purchaseToken = jsonObject.optString("purchaseToken", ""),
+                    quantity = jsonObject.optInt("quantity", 1),
+                    acknowledged = jsonObject.optBoolean("acknowledged", false),
+                    autoRenewing = jsonObject.optBoolean("autoRenewing", false),
+                )
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
```
JayShortway:
```suggestion
            } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseData.kt
```diff
@@ -0,0 +1,58 @@
...
+                errorLog(e) { "Error parsing AIDL purchase data JSON: $json" }
```
JayShortway, line 40: Would it make sense to add diagnostics for getPurchaseHistory-related errors like this? If yes, let's do that separately.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+import kotlinx.coroutines.sync.withLock
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+/**
+ * Manager for querying purchase history using the AIDL-generated stub.
+ *
+ * This provides access to the deprecated getPurchaseHistory() method that was
+ * removed from Play Billing Library 8.0.0 but is still supported by Google Play.
```
JayShortway: Maybe this is not needed.

```suggestion
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseData.kt
```diff
@@ -0,0 +1,58 @@
...
+internal data class PurchaseData(
+    val orderId: String,
+    val packageName: String,
+    val productId: String,
+    val purchaseTime: Long,
+    val purchaseState: Int,
+    val purchaseToken: String,
+    val quantity: Int = 1,
+    val acknowledged: Boolean = false,
+    val autoRenewing: Boolean = false,
```
vegaro: I was wondering why these three have defaults but the others don't

tonidero: No reason, none of them should be optional 🥴 I will clean that up

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+            return deferred.await()
+        }
+
+        return try {
+            val result = operation()
+            deferred.complete(result)
+            result
+        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+            deferred.completeExceptionally(e)
+            throw e
```
JayShortway, line 85: This will also catch Coroutines' `CancellationException`. We are rethrowing it, so that's good, but I'm not sure if calling `completeExceptionally` is appropriate in that case, as being cancelled is not an error. Also not quite sure what to do instead. Maybe `deferred` has functions to indicate it was cancelled? See also https://kotlinlang.org/docs/cancellation-and-timeouts.html

purchases/src/test/java/com/revenuecat/purchases/google/history/PurchaseHistoryManagerTest.kt
```diff
@@ -0,0 +1,595 @@
...
+        try {
+            manager.connect()
+            assert(false) { "Expected exception" }
+        } catch (e: Exception) {
+            assertThat(e).isEqualTo(expectedException)
+        }
+    }
+
+    @Test
+    fun `concurrent connect() calls reuse the same operation`() = runTest {
```
JayShortway, line 128: Nice!!

purchases/src/test/java/com/revenuecat/purchases/google/history/PurchaseHistoryManagerTest.kt
```diff
@@ -0,0 +1,595 @@
...
+        val result1 = async { manager.connect() }
+        val result2 = async { manager.connect() }
+        val result3 = async { manager.connect() }
+
+        // Yield to allow the async tasks to call bindService
+        kotlinx.coroutines.yield()
```
JayShortway: Nit, we could import this function so it's just `yield()`.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+        },
+        debugMessage = "Query for type $type already in progress, hooking into existing operation",
+    ) {
+        performQuery(type)
+    }
+
+    /**
+     * Performs the actual query operation with pagination.
+     */
+    private fun performQuery(type: String): List<StoreTransaction> {
```
JayShortway: Maybe this can be inlined in `queryAllPurchaseHistory`? Mostly suggesting that because `performQuery` is a rather generic name, and I think the best name for this function is also `queryAllPurchaseHistory` haha.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
JayShortway: No I totally agree that the caller should decide the context.

What I'm mostly referring to is that this function makes multiple blocking IPC calls (as long as `continuationToken` is not null) and there's nothing to indicate that. Making it `suspend` has 2 advantages in my opinion:

1. It's a bit clearer that this is an async operation. (Although this can also be solved by suffixing it with `*Blocking()` or something.)
2. Most importantly, it allows us to support cancellation. In the `while` loop, we can check `isActive` to see if our CoroutineContext is still active, and if not, break the loop.

```kotlin
while (continuationToken != null && isActive)
```

([docs](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancellation-is-cooperative))

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
```diff
@@ -0,0 +1,309 @@
...
+                override fun onServiceConnected(name: ComponentName?, service: IBinder?) {
+                    debugLog { "AIDL Billing service connected" }
+                    billingService = IInAppBillingService.Stub.asInterface(service)
+                    serviceConnection = this
+                    continuation.resume(true)
+                }
+
+                override fun onServiceDisconnected(name: ComponentName?) {
+                    debugLog { "AIDL Billing service disconnected" }
+                    billingService = null
```
JayShortway: Should we not do the same here as in `disconnect()`? Basically referring to all/most of these:
```kotlin
connectDeferred?.cancel()
queryDeferreds.forEach {
    it.value.cancel()
}
serviceConnection?.let { connection ->
    try {
        context.unbindService(connection)
        debugLog { "AIDL Billing service disconnected" }
    } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
        errorLog(e) { "Error disconnecting from AIDL Billing service" }
    }
}
billingService = null
serviceConnection = null
connectDeferred = null
queryDeferreds.clear()
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -79,15 +86,18 @@ import kotlin.math.min
...
 internal class BillingWrapper(
     private val clientFactory: ClientFactory,
     private val mainHandler: Handler,
     private val deviceCache: DeviceCache,
     @Suppress("unused")
     private val diagnosticsTrackerIfEnabled: DiagnosticsTracker?,
     purchasesStateProvider: PurchasesStateProvider,
+    private val purchaseHistoryManager: PurchaseHistoryManager,
     private val dateProvider: DateProvider = DefaultDateProvider(),
+    private val coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
```
JayShortway, line 98: Not now, but at some point we should probably start managing our scopes a bit more centrally.

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -359,18 +369,71 @@ internal class BillingWrapper(
...
+                    mainHandler.post {
+                        onReceivePurchaseHistory(transactions)
+                    }
+                } finally {
+                    purchaseHistoryManager.disconnect()
+                }
+            } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+                try {
+                    purchaseHistoryManager.disconnect()
+                } catch (@Suppress("TooGenericExceptionCaught") disconnectException: Throwable) {
```
JayShortway, line 424: Since we're in a suspend context here, I think we should check if `e` and `disconnectException` is a `CancellationException`, and if so, rethrow it and not treat it as an error.

purchases/src/test/java/com/revenuecat/purchases/google/history/PurchaseHistoryManagerTest.kt
```diff
@@ -0,0 +1,595 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(manifest = Config.NONE)
+internal class PurchaseHistoryManagerTest {
```
JayShortway, line 37: Maybe we can test some cancellation scenarios, like cancelling halfway through pagination for instance, or during the initial connection.

JayShortway, line 37: Amazing tests!!

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseData.kt
tonidero: Hmm it could indeed be useful... And yeah, let's do that separately 👍

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
tonidero: Hmm I'm not sure if we want to support cancellable coroutines here... As in, I'm thinking in what situations we would like to support cancelling this operation midway, considering we own the coroutine scope and I think there wouldn't be situations where this gets cancelled. However, your suggestion may be trying to future-proof it in case we sometimes support cancelling coroutines in the public API of the SDK, correct?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: I've been back and forth on this... Considering this coroutine gets executed within a non-coroutine context, if we just propagate the exception, it would just fail on the coroutineScope and fail silently, without calling the callback to finish the operation outside of the coroutine scope. So I think we should still call the callbacks with an error in this case. As I mentioned, there isn't currently a valid of a developer-initiated cancellation, so we can handle that later if we refactor the SDK to use more coroutines.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/BillingConstants.kt
tonidero: I changed to use the Billing library ones... I guess they could potentially change in newer versions of the SDK... but I think it would be very unlikely

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/BillingConstants.kt
tonidero: Yup changed, thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
tonidero: I followed your other comment to inline this function into the getAllPurchaseHistory one, which is already suspend. Also added the check for whether the coroutine has been cancelled. Lmk what you think about it!

purchases/src/test/java/com/revenuecat/purchases/google/history/PurchaseHistoryManagerTest.kt
tonidero: Added one test... It's quite tricky to test these scenarios though, since it's all executed synchronously during tests though.

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
JayShortway: Yea, it's more a matter of playing nice within the Coroutines system. We're already half-supporting cancellation here, because we're rethrowing exceptions (meaning the `CancellationException` will also be rethrown). I think the only thing left to do to fully support it is:
```suggestion
        } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
            if (e is CancellationException) deferred.cancel(e)
            else deferred.completeExceptionally(e)
            throw e
```

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
JayShortway: Hmm that's a good point. I agree that we should ensure we never fail silently. One way we could do that is by calling the callback in a [finally](https://kotlinlang.org/docs/cancellation-and-timeouts.html#closing-resources-with-finally) block: 

```kotlin
} finally {
  if (isActive) // success callback
  else // error callback (cancelled)
}
```
(did not test that code 😄) 

I'm also okay leaving it for now, as I agree it should be fine. I just know that catching `CancellationException`s can lead to subtle bugs, so that's something we should watch out for when touching this code.

purchases/src/test/java/com/revenuecat/purchases/google/history/PurchaseHistoryManagerTest.kt
```diff
@@ -0,0 +1,758 @@
...
+                "com.test.package",
+                BillingClient.ProductType.INAPP,
+                "continuation_token_1",
+                any()
+            )
+        }
+    }
+
+    @Test
+    fun `queryAllPurchaseHistory() stops pagination when coroutine is cancelled`() = runTest {
```
JayShortway, line 310: 😻

JayShortway (approved):
Looks great! I'd like us to really consider playing nice with Coroutines' cancellation system. Not doing that can lead to subtle bugs, either now or after a future refactor. Other than that it's good to ship imo! 💪

purchases/src/main/kotlin/com/revenuecat/purchases/google/history/PurchaseHistoryManager.kt
tonidero: Right, this is now done but with a different `catch` clause instead of an if-else :) So I think we're in the same page 🙌

JayShortway: Oh nice! Yea that works too!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Hmm so right now this is part of the catch clause that catches any exception and we propagate to the error callback always (not the success callback). If there was a cancellation exception thrown, it would be caught here already and be propagated to the `onReceivePurchaseHistoryError` below. So I think that should already cover these cases? 

Do lmk if you were thinking of something else though!

pauledwardtang:
Quick follow up.

In the pull request I noticed that the AIDL setting is **off** by default yet I got purchase information regardless.

As a sanity check I re-tested with `9.16.3` and `9.0.0` (fresh install, backups disabled, no backup agent configured) and I now get purchase information when I wasn't before. It's as if something has changed upstream and the issue is resolved for me or I did not test this with the right account / tested incorrectly.

I'm happy to share my app and product identifiers if it'll help.

pauledwardtang:
@tonidero 

I understand how that would be the case if the properties & key are injected from a CI server, but I'm looking at the generated build config and I don't see that flag is set to `true` when using the analyzer on a debug APK and a release AAB in Android Studio. I checked if `isAIDLEnabled` is set somewhere else when instantiating `BillingWrapper`, but `BillingFactory` instantiates it with the defaults.

image

image

I'm pretty sure I'm missing something here, just letting you know what I see on my end. Either way, the issue seems to have gone away for me regardless of the AIDL fix.

PR #2873, opened 2025-11-25
Fix predownloading of fonts if first offering doesn't have paywall components

vegaro:
I noticed that if the first offering returned by the backend doesn't have paywall components, we were not pre-downloading the fonts. This PR should prevent that. The added test doesn't pass in `main`


tonidero (approved):
Thanks for doing this!

PR #2876, opened 2025-11-26
Adds infra to support recording Maestro tests

JayShortway:
## Description
**Note:** based on https://github.com/RevenueCat/purchases-android/pull/2866. 

This adds the ability to record individual Maestro tests. 

### New infra
Here's the new infra:

- A single-file HTTP server, called `ShellServer`, which can run arbitrary shell commands on the (CI) host.
- A `maestro/shell/shell.js` script to make it easier to run shell commands through this server.
- A  `record/scripts/android.js` script that calls `shell.js` and runs the correct ADB commands to control screen recording.
- `record/start.yml` and `record/stop.yml` which make it convenient to start and stop screen recording from Maestro flows.

### Can be shared with iOS in the future
The above can be shared with iOS too, if we create `record/scripts/ios.js` which runs something like
```shell
xcrun simctl io booted recordVideo video.mov
```
(We can pre-compile the `ShellServer` and run it as jar or something.)

### How it's used
And this is how it's used:
- `maestro/test_paywall_v2.yaml` starts recording in `onFlowStart` and stops recording in `onFlowComplete`.
- The recording is only saved if the flow failed. Otherwise it is discarded.
- Since `onFlowComplete` runs regardless of success, I added `conventions/mark-successful.yml` to mark the flow successful. This should be run at the end of the regular flow steps.

### DIFF
--- a/maestro/record/scripts/android.js
+++ b/maestro/record/scripts/android.js
@@ -1,0 +1,40 @@
+function normalizeFilename(filename) {
+    return filename.replace(/\.[^.]+$/, '') + '.mp4';
+}
+
+function start(filename) {
+    var name = normalizeFilename(filename);
+
+    var pid = output.shell.start('adb shell screenrecord /sdcard/' + name);
+
+    if (!output.recordingPids) {
+        output.recordingPids = {};
+    }
+    output.recordingPids[name] = pid;
+
+    return pid;
+}
+
+function stop(filename, outputDir) {
+    var name = normalizeFilename(filename);
+
+    var pid = output.recordingPids ? output.recordingPids[name] : null;
+
+    if (pid && pid !== -1) {
+        output.shell.stop(pid);
+        delete output.recordingPids[name];
+    }
+
+    if (outputDir) {
+        var destination = outputDir + '/' + name;
+        output.shell.run('adb pull /sdcard/' + name + ' ' + destination);
+    }
+
+    // Clean up the recording file from the emulator
+    output.shell.run('adb shell rm -f /sdcard/' + name);
+}
+
+output.record = {
+    start: start,
+    stop: stop
+};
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -137,5 +137,5 @@ purchases-android-snapshots
 # Make sure we share the same Detekt IntelliJ plugin settings
 !.idea/detekt.xml
 
-# Kotlin error logs
-.kotlin/errors/
+# Kotlin compiler
+.kotlin
```
JayShortway, line 141: The `shell-over-http` project caused some more files being created in this folder. Ignoring this entire folder is [recommended](https://kotlinlang.org/docs/whatsnew20.html#new-directory-for-kotlin-data-in-gradle-projects) anyway:
>Do not commit the .kotlin directory to version control. For example, if you are using Git, add .kotlin to your project's .gitignore file.

maestro/record/scripts/android.js
```diff
@@ -0,0 +1,40 @@
+function normalizeFilename(filename) {
+    return filename.replace(/\.[^.]+$/, '') + '.mp4';
+}
+
+function start(filename) {
+    var name = normalizeFilename(filename);
+
+    var pid = output.shell.start('adb shell screenrecord /sdcard/' + name);
```
rickvdl, line 8: This is nice to be able to run full commands through the webserver 👏

tonidero, line 8: Hmm I'm not sure, is the `/sdcard/` folder always guaranteed to exist? Not sure if we should record in a documents folder or something like...

maestro/shell/shell.js
```diff
@@ -0,0 +1,46 @@
...
+        });
+        return response.status === 200;
+    } catch (e) {
+        return false;
+    }
+}
+
+function run(cmd) {
+    if (!isServerRunning()) {
+        return { exitCode: -1, stdout: '', stderr: 'shell-over-http server not running' };
```
tonidero, line 17: I'm probably missing something... but do we need to start the server for `shell-over-http` in the circleci job/fastlane lane before running the tests?

PR #2885, opened 2025-11-28
Fixes `ConcurrentModificationException` in `FontLoader`

JayShortway:
## Description
Fixes https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/6507804763198446256/crashes/a557155f/details

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
@@ -42,6 +42,7 @@ internal class FontLoader(
     }
 
     private val fontInfosForHash = mutableMapOf<String, MutableSet<DownloadableFontInfo>>()
+    private val lock = Any()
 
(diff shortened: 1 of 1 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -79,7 +80,7 @@ internal class FontLoader(
             val extension = url.substringAfterLast('.', missingDelimiterValue = "")
             val cachedFile = File(cacheDirectory, "$urlHash.$extension")
 
-            synchronized(this) {
+            synchronized(lock) {
```
JayShortway, line 83: This (and below on L114) is the cause. `this` here refers to the `ioScope`, not the `FontLoader`, so it's not actually mutually exclusive. 

We could've changed it to `this@FontLoader` and be done with it, but I thought it's better to be explicit.

tonidero (approved):
Ohhh ugh, great catch!! And yeah I agree with being explicit. Thanks for fixing!!

PR #2887, opened 2025-12-01
Remove data binding purchase tester + Update AGP

tonidero:
### Description
This PR removes data binding usages from the XML directly in purchase tester. Instead changes to use the newer view binding. Additionally, it updates the android gradle plugin which was stuck in an older version due to some incompatibilities with data binding.

For reviewing, I recommend hiding white spaces, specially in the XML files.

### DIFF
--- a/examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedPackageCardAdapter.kt
+++ b/examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedPackageCardAdapter.kt
@@ -43,9 +43,8 @@ class DeprecatedPackageCardAdapter(
             binding.packageProductTitle.text = "${product.title}${if (isActive) " (active)" else ""}"
             binding.packageProductDescription.text = product.description
 
-            updateRowView(binding.packageProductSku, "Sku:", product.id)
-            updateRowView(
-                binding.packageType,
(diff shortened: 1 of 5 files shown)


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/DeprecatedPackageCardAdapter.kt
```diff
@@ -59,24 +85,24 @@ class DeprecatedPackageCardAdapter(
...
+            )
 
             binding.root.setOnClickListener {
                 with(binding.packageDetailsContainer) {
                     visibility = if (visibility == View.GONE) View.VISIBLE else View.GONE
                 }
             }
         }
+
+        private fun updateRowView(rowViewBinding: RowViewBinding, header: String, detail: String?) {
```
JayShortway: Nit, would be nice if `RowViewBinding` is the receiver here (and in other places where we have a similar function):

```suggestion
        private fun RowViewBinding.updateRowView(header: String, detail: String?) {
```

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/2887/commits/dc2c0fd24e037131d1e401e50dec1607b4b368a9

PR #2891, opened 2025-12-02
Show redacted Test Api key in alert when detected in Release build

ajpallares:
### Description
Adds a redacted version of the Test Store API key to the alert (and log) shown when using the Test Store API key in a Release configuration.

image

iOS counterpart: https://github.com/RevenueCat/purchases-ios/pull/5897

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/APIKeyValidator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/APIKeyValidator.kt
@@ -1,6 +1,5 @@
 package com.revenuecat.purchases
 
-import androidx.annotation.VisibleForTesting
 import com.revenuecat.purchases.common.debugLog
 import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.warnLog


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -497,3 +498,29 @@ internal class PurchasesFactory(
         }
     }
 }
+
+internal val String.asRedactedAPIKey: String
+    get() {
+        val underscoreIndex = indexOf('_')
+        if (underscoreIndex == -1) {
+            return this // no underscore → do not redact
```
tonidero: Same as in iOS. I think we should redact in this case as well, just in case.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -497,3 +498,29 @@ internal class PurchasesFactory(
         }
     }
 }
+
+internal val String.asRedactedAPIKey: String
```
tonidero: Also, I would either make this private, or better yet, move somewhere else so it's better namespaced... Maybe to the `APIKeyValidator`? Or create a new `APIKeyUtils` file or something like that...

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
```diff
@@ -9,14 +9,20 @@ import android.os.Bundle
 internal class TestStoreErrorDialogActivity : Activity() {
```
tonidero: Something I just noticed, but this class should have been called `SimulatedStoreErrorDialogActivity` I guess 😅. Not to fix in this PR, just something I noticed now.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Also nitpicky and this might be my personal preference, but I would make this a function... `toRedactedAPIKey()` or something like that... Not a blocker at all though.

ajpallares: Moved to `APIKeyValidator` in https://github.com/RevenueCat/purchases-android/pull/2891/commits/996d8a743d530a773d19577128ed3a79ee552e3a (also made a function of `APIKeyValidator`)
Lmk if you still prefer to have it separate

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
ajpallares: Rewritten to redact the no-underscore case in https://github.com/RevenueCat/purchases-android/pull/2891/commits/996d8a743d530a773d19577128ed3a79ee552e3a

purchases/src/main/kotlin/com/revenuecat/purchases/APIKeyValidator.kt
```diff
@@ -72,4 +80,29 @@ internal class APIKeyValidator {
             else -> APIKeyPlatform.OTHER_PLATFORM
         }
     }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
```
tonidero: In this case, it's not actually visible for testing, since it actually needs to be used by production code right? I would remove this annotation :)

purchases/src/main/kotlin/com/revenuecat/purchases/TestStoreErrorDialogActivity.kt
ajpallares: Right! Good catch! Well, I think it's fine to do it in this PR, since it's only an internal rename (https://github.com/RevenueCat/purchases-android/pull/2891/commits/7275e71452483a536b0d7ce88d582f6b1577894a)

I have a different opinion on something that is more user/developer facing like [this](https://github.com/RevenueCat/purchases-ios/pull/5897#discussion_r2582020892), which I think should go into a separate PR

purchases/src/main/kotlin/com/revenuecat/purchases/APIKeyValidator.kt
ajpallares: Ohh right! Thank you! I think I left it over from the previous implementation 😅 
Removed in https://github.com/RevenueCat/purchases-android/pull/2891/commits/63440f0677a9e4cd4f7bf39aa59cf202ea5b3a50

PR #2892, opened 2025-12-03
Add extra non subscription events

tonidero:
### Description
This adds 2 new events that can be tracked:
- `trackAdLoaded` 
- `trackAdFailedToLoad`

This also 
- Makes the `impressionId` field for these events nullable, since it won't be available for failed to load events. 
- Adds a `mediator_error_code` field for failed to load events.

Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/5895


ajpallares (approved):
Looks great!

PR #2894, opened 2025-12-03
Conditional visibility -- app version condition

JZDesign:
### Checklist

### Motivation

As we evolve the conditional visibility feature and add new conditions, our users will likely wind up in a spot where their paywalls don't render in an expected way. We need to provide an option to allow them to account for old SDKs and to control how their paywall renders if they use newer conditions that old SDKs are unaware of.

### Description

Adds the ability for users to create a condition that will only show the component for a given app version

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
@@ -4,6 +4,7 @@ import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.paywalls.components.PartialComponent
 import com.revenuecat.purchases.paywalls.components.common.ComponentOverride.Condition
 import com.revenuecat.purchases.utils.serializers.SealedDeserializerWithDefault
+import com.revenuecat.purchases.utils.serializers.VersionIntSerializer
 import dev.drewhamilton.poko.Poko
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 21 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -99,6 +100,17 @@ class ComponentOverride<T : PartialComponent>(
             @SerialName("packages") val packages: List<String>,
         ) : Condition
 
+        /**
+         * Compares the app version against the [version].
+         * */
+        @Serializable
+        data class AppVersion(
+            @SerialName("operator") val operator: ComparisonOperatorType,
```
tonidero, line 108: Should we add `in` or `not_in` operators as well? Alternatively, we could add `and` and `or`... but maybe `in`/`not_in` would be enough.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/VersionIntSerializer.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.intOrNull
+import kotlinx.serialization.json.jsonPrimitive
+
+/**
+ * Serializer that converts version strings like "3.2.0" to integers like 320 by keeping only digits.
+ * Also handles integer inputs directly for backward compatibility.
```
tonidero, line 15: Hmm does this work for comparison purposes? For example, 11.2.0 would get translated to 1120, and 2.11.0 would get translated to 2110, and it would say that the second is bigger than the first? I think we need proper semver parsing and comparison instead of parsing to an int.

tonidero (commented):
Looking good but I do think we need to fix the semver number parsing/comparison

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/VersionIntSerializer.kt
JZDesign: Man… you're right. Thank you for the sanity check here.

JayShortway: Would be good to add some tests for this as well. Using parameterized tests we can cover a lot of cases easily.

PR #2895, opened 2025-12-03
Conditional visibility — fallback visibility

JZDesign:
### Checklist
- [x] If applicable, create follow-up issues for `purchases-android` and hybrids

### Motivation

As we evolve the conditional visibility feature and add new conditions, our users will likely wind up in a spot where their paywalls don't render in an expected way. We need to provide an option to allow them to account for old SDKs and to control how their paywall renders if they use newer conditions that old SDKs are unaware of.

### Description

Adds a property to the components that defines how the SDK should handle unsupported conditions


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CarouselComponent.kt
```diff
@@ -68,6 +68,9 @@ class CarouselComponent(
     val autoAdvance: AutoAdvancePages? = null,
     @get:JvmSynthetic
     val overrides: List<ComponentOverride<PartialCarouselComponent>> = emptyList(),
+    @get:JvmSynthetic
+    @SerialName("evaluate_unknown_conditions_as")
+    val evaluateUnknownConditionsAs: Boolean? = null,
```
tonidero, line 73: I'm unsure about  the position of the property... should it be part of the `condition` itself? As in, we might want to ignore some specific condition if unrecognized, but not another one. It would also be better encapsulated IMO. And if we do need component level support, we can just set the same value for all conditions. Wdyt?

tonidero (commented):
Just a thought on the general architecture... Code changes themselves look good but would like some thoughts on that first

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/CarouselComponent.kt
JZDesign: We debated on this for a while, the condition level override lost the debate and the team wanted to have a component level override for all unknown conditions. 

To set the same value for all conditions in a component is possible… I'll ask the team what they think.

PR #2898, opened 2025-12-05
Fallback to Billing library when error querying products through AIDL

tonidero:
### Description
This is based on https://github.com/RevenueCat/purchases-android/pull/2872

This just makes sure we correctly handle issues when querying products through the AIDL, by falling back to the billing library mechanisms.


ajpallares (approved):
Right! Makes total sense! 👍

PR #2900, opened 2025-12-05
Add Galaxy Store

fire-at-will:
### Description
Adds the `GALAXY` to the list of supported Store types. 

> Note: This PR will have one failing integration test for the Customer Center's localized strings not containing the Galaxy Store string until a backend PR has been merged.

### DIFF
--- a/purchases/src/test/resources/backend_integration_tests_golden/ProductionBackendIntegrationTest/can get customer center config data/response_001.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/ProductionBackendIntegrationTest/can get customer center config data/response_001.json
@@ -3,7 +3,7 @@
   "headers": {
     "Alt-Svc": "h3=\":443\"; ma=86400",
     "Connection": "keep-alive",
-    "Content-Length": "9117",
+    "Content-Length": "9147",
     "Content-Type": "application\/json",
     "X-Cache": "Miss from cloudfront",
     "access-control-allow-origin": "*",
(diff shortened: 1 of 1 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -247,6 +247,9 @@ data class CustomerCenterConfigData(
             @SerialName("amazon_store")
             AMAZON_STORE,
 
+            @SerialName("galaxy_store")
+            GALAXY_STORE,
```
fire-at-will, line 251: Named this `GALAXY_STORE` instead of galaxy since the string is literally "Galaxy Store" to match the name of the store

purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt
```diff
@@ -234,6 +234,11 @@ enum class Store {
      * For entitlements granted via RevenueCat's Test Store.
      */
     TEST_STORE,
+
+    /**
+     * For entitlement granted via the Galaxy store.
+     */
+    GALAXY,
```
tonidero, line 241: A bit late but should we mark this enum value as Experimental? I guess it could still be considered a small breaking change for devs using switch-case statements... NABD  for now since we're merging to an integration branch though :)

PR #2901, opened 2025-12-05
Galaxy Scaffolding Classes

fire-at-will:
### Description
Introduces the scaffolding for several classes that will be foundational to Samsung development as we move forward:
- `GalaxyBillingWrapper`: implementation of `BillingAbstract`. The class currently satisfies all of the BillingAbstract's requirements, but all functions are stubbed with `TODO`s
- `GalaxyPurchasingData`: will be used when processing samsung purchases
- `GalaxyStoreProduct`: represents a product from the Samsung store

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyPurchasingData.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyPurchasingData.kt
@@ -1,14 +1,14 @@
-package com.revenuecat.purchases.samsung
+package com.revenuecat.purchases.galaxy
 
 import com.revenuecat.purchases.ProductType
 import com.revenuecat.purchases.models.PurchasingData
 import dev.drewhamilton.poko.Poko
 
-sealed class SamsungPurchasingData : PurchasingData {
+sealed class GalaxyPurchasingData : PurchasingData {
     @Poko
     class Product(
-        val storeProduct: SamsungStoreProduct,
-    ) : SamsungPurchasingData() {
+        val storeProduct: GalaxyStoreProduct,
+    ) : GalaxyPurchasingData() {
         override val productId: String
             get() = storeProduct.id
         override val productType: ProductType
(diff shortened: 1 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
```diff
@@ -0,0 +1,125 @@
...
+     */
+    override val price: Price,
+
+    /**
+     * Name of the product. This will match the name for Galaxy products.
+     */
+    override val name: String,
+
+    /**
+     * Name of the product. This will match the name for Galaxy products.
```
fire-at-will, line 37: This is not a typo: `ProductVo`s do not have a `title` property, so we'll need to just put the name property in here (or leave it as a blank string)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyPurchasingData.kt
```diff
@@ -0,0 +1,17 @@
+package com.revenuecat.purchases.galaxy
+
+import com.revenuecat.purchases.ProductType
+import com.revenuecat.purchases.models.PurchasingData
+import dev.drewhamilton.poko.Poko
+
+sealed class GalaxyPurchasingData : PurchasingData {
```
tonidero, line 7: Same as in the preview PR, might make sense to make these experimental... but NABD since we're merging to an integration branch

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
```diff
@@ -0,0 +1,125 @@
...
+    override val type: ProductType,
+
+    /**
+     * Price information for a non-subscription product.
+     * Term price for a subscription.
+     */
+    override val price: Price,
+
+    /**
+     * Name of the product. This will match the name for Galaxy products.
```
tonidero, line 32: A bit weird doc, saying that it will match itself :P Probably not needed? Like
```suggestion
     * Name of the product.
```

PR #2904, opened 2025-12-08
Configure SDK with Galaxy Store

fire-at-will:
### Description
This PR makes it so that SDK consumers can configure the Android SDK to use the Galaxy Store, like so:

```kotlin
val configuration = GalaxyConfiguration.Builder(
     context,
     apiKey,
     GalaxyBillingMode.PRODUCTION,
).build()

```

It also:
- Introduces `GalaxyBillingMode`, an enum required for Galaxy SDK configurations, which determines the environment that the Galaxy Store processes IAPs with (prod, test, etc.).
- Creates a new `galaxy` module that loads in the Samsung IAP SDK from the local disk. The location of the Samsung SDK  can be configured in `local.properties`, like so: `samsungIapSdkPath=/.../SamsungInAppPurchaseSDK_v6.5.0/Libs/samsung-iap-6.5.0.aar`. 
   - This is temporary and will be changed later
   - As a result of relying on a local copy of the Samsung SDK, **CI will fail for the time being** 🤷‍♂️ 
- When a Galaxy configuration is used, `BillingFactory` will wire up a `GalaxyBillingWrapper`
-  `GalaxyBillingWrapper` now initializes and configures `SamsungIAP`'s `IAPHelper` class, and passes in the IAP operation mode, depending on the value provided by the developer when configuring the SDK
- Added unit tests :)

### DIFF
--- a/feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/GalaxyConfigurationTest.kt
+++ b/feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/GalaxyConfigurationTest.kt
@@ -1,9 +1,7 @@
 package com.revenuecat.purchases.galaxy
 
 import android.content.Context
-import android.os.Build
 import androidx.test.ext.junit.runners.AndroidJUnit4
-import com.revenuecat.purchases.Purchases
 import com.revenuecat.purchases.Store
 import io.mockk.every
 import io.mockk.mockk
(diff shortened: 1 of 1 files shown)


feature/galaxy/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    defaultConfig {
+        missingDimensionStrategy("apis", "defaults")
+    }
+
+    testOptions {
+        // TODO: Remove this?
+        // Avoid merging Android manifests for JVM unit tests to prevent minSdk conflicts from optional AARs.
+        unitTests.isIncludeAndroidResources = false
```
fire-at-will, line 31: This is a hack proposed by ChatGPT to allow us to execute unit tests even though the Samsung SDK's minSDK != our minSDK. I think we'll want to remove it once we solidify our module/dependency structure.

PR #2905, opened 2025-12-08
Support Configuring the SDK with the Galaxy Store

fire-at-will:
### Description
This PR creates a new module for the Galaxy Store, and allows developers to set up the Android SDK to use the Galaxy Store when they configure the SDK, like so:

```kotlin
val configuration = GalaxyConfiguration.Builder(
     context,
     apiKey,
     GalaxyBillingMode.PRODUCTION,
).build()

```

It also:
- Introduces `GalaxyBillingMode`, an enum required when configuring the SDK to use the Galaxy Store, which determines the environment that the Galaxy Store processes IAPs with (prod, test, etc.).
- When a Galaxy configuration is used, `BillingFactory` will wire up a `GalaxyBillingWrapper`
- Adds unit tests :)

### DIFF
--- a/api-tester/src/main/java/com/revenuecat/apitester/kotlin/GalaxyConfigurationAPI.kt
+++ b/api-tester/src/main/java/com/revenuecat/apitester/kotlin/GalaxyConfigurationAPI.kt
@@ -14,13 +14,13 @@ private class GalaxyConfigurationAPI {
     ) {
         val builder1: GalaxyConfiguration.Builder = GalaxyConfiguration.Builder(
             context,
-            apiKey
+            apiKey,
         )
 
         val builder2: GalaxyConfiguration.Builder = GalaxyConfiguration.Builder(
             context,
             apiKey,
-            galaxyBillingMode
+            galaxyBillingMode,
         )
 
         val config1: PurchasesConfiguration = builder1.build()
@@ -34,10 +34,9 @@ private class GalaxyConfigurationAPI {
     ) {
         val config: PurchasesConfiguration = GalaxyConfiguration.Builder(
             context,
-            apiKey
+            apiKey,
         )
             .galaxyBillingMode(galaxyBillingMode = galaxyBillingMode)
             .build()
     }
 }
-


feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import androidx.annotation.NonNull;
+
+import com.revenuecat.purchases.PurchasesConfiguration;
+import com.revenuecat.purchases.Store;
+
+/**
+ * Holds parameters to initialize the SDK for the Galaxy Store. Create an instance of this class using the [Builder]
+ * and pass it to [Purchases.configure].
+ */
+public final class GalaxyConfiguration extends PurchasesConfiguration {
```
tonidero, line 27: Hmm any reason this is Java instead of Kotlin?

feature/galaxy/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+
+    // billingclient dimension is added for bc7/bc8 support
+    flavorDimensions += "billingclient"
+
+    productFlavors {
+        create("bc8") {
+            dimension = "billingclient"
+            isDefault = true
+        }
+        create("bc7") {
```
tonidero, line 16: We can probably remove the bc7 flavor here since it's not actually needed. We can probably just remove the product flavors entirely and add `missingDimensionStrategy("billingclient", "bc8")` below

feature/galaxy/consumer-rules.pro
```diff
@@ -0,0 +1,4 @@
+-dontwarn com.samsung.**
+-keep class com.samsung.** {*;}
```
tonidero: Just to check, are these rules recommended by samsung? Since this basically keeps everything which is not a great practice (though we also need to clean this up 😅 )

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+    public GalaxyConfiguration(@NonNull Builder builder) {
+        super(builder);
+    }
+
+    public static final class Builder extends PurchasesConfiguration.Builder {
+
+        public Builder(
+                @NonNull Context context,
+                @NonNull String apiKey,
+                @NonNull GalaxyBillingMode galaxyBillingMode
```
tonidero, line 47: I see we're making this mandatory for galaxy configurations, but should we default here to a value and provide a builder method to change it? I guess it's ok if we want to make it mandatory for now though.

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+        ) {
+            super(context, apiKey);
+            this.store(Store.GALAXY);
+            this.galaxyBillingMode(galaxyBillingMode);
```
tonidero, line 51: Note that it's possible someone uses the `PurchasesConfiguration` builder directly, instead of this one, by passing the GALAXY store and forgets to pass the galaxy billing mode, in which case, we will need to decide on a default for this value, which I see we already have as `PRODUCTION`, so I wonder if we should default to that in the constructor here as well.

tonidero (commented):
Looking nice! Just some questions

feature/galaxy/build.gradle.kts
tonidero: Oh wait, I remember why I did this now 😅. Mostly it's because hybrids import these libraries, and since the amazon library also includes Google, we wanted to make sure we only have a single flavor of google in the amazon library. 

So basically, this is good as it is. We will likely be removing the BC7 flavors soon'ish in favor of the AIDL but we can keep this for now 👍

feature/galaxy/consumer-rules.pro
fire-at-will: I haven't seen this recommended by samsung - I copied this over directly from our Amazon module and replaced `amazon` with `samsung`. Do you think we should remove this?

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
fire-at-will: I wrote this in Java because it looks like there was a pretty good reason that we wrote the `AmazonConfiguration` in Java instead of Kotlin, and I was copying over the model we used for Amazon 😅

There are more details at the top of this file and in the [AmazonConfiguration](https://github.com/RevenueCat/purchases-android/blob/30dcacc9cd44ce11b6033ea18c20ff11e06700af/feature/amazon/src/main/java/com/revenuecat/purchases/amazon/AmazonConfiguration.java#L1) file:

```java
/**
 * This class is in Java because there is a weird issue when written in Kotlin. The class
 * can't be used as a PurchasesConfiguration.Builder from a Java class when written in Kotlin.
 *
 * If written in Kotlin this wouldn't work:
 *
 * PurchasesConfiguration.Builder builder = new AmazonConfiguration.Builder(this, "");
 *
 * And this either:
 *
 * AmazonConfiguration.Builder builder = new AmazonConfiguration.Builder(this, "");
 * builder.build();
 */
```

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
fire-at-will: Yeah, I'm okay with adding in a default value here! Since this is written in Java, I think it'll just be a separate constructor.

fire-at-will: Okay, added a second constructor + unit tests for `GalaxyConfiguration` in 1c175d047d77ad41b4b438f31d73bc3bcae1eb2d :D

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
fire-at-will: Yeah, I think that defaulting to `PRODUCTION` in that scenario is the best thing to do! Samsung explicitly calls out not to use the other modes when submitting apps for distribution

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
fire-at-will: > Should we ... and provide a builder method to change it?

Yeah, there's a `.galaxyBillingMode()` builder function on `PurchasesConfiguration.Builder` that the `GalaxyConfiguration.Builder` class gets through inheritance!

feature/galaxy/consumer-rules.pro
tonidero: Unless we have a reason to disable obfuscated of those classes, I would say yes, let's remove this 👍

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
tonidero: Ohhh I should read 😅 But yeah, let's keep this one in java then, thanks for explaining!

feature/galaxy/src/main/java/com/revenuecat/purchases/galaxy/GalaxyConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+import androidx.annotation.NonNull;
+
+import com.revenuecat.purchases.PurchasesConfiguration;
+import com.revenuecat.purchases.Store;
+
+/**
+ * Holds parameters to initialize the SDK for the Galaxy Store. Create an instance of this class using the [Builder]
+ * and pass it to [Purchases.configure].
+ */
+public final class GalaxyConfiguration extends PurchasesConfiguration {
```
tonidero, line 27: Should we add API tests for this as well? Alternatively, we could mark these public APIs as experimental, but since we're merging to an integration branch, I think it's ok to add that later only if needed.

fire-at-will: I'll add some API tests! As for marking them as experimental, I think the plan is to not do that for now, keep merging into the dev branch, and then create a beta release for the first few customers who use the Galaxy store.

PR #2906, opened 2025-12-08
Handle Unsupported BillingWrapper Functions on the GalaxyBillingWrapper

fire-at-will:
### Description
The Galaxy Store doesn't support fetching the current storefront/country code or showing in-app messages, so this PR:
- Logs + throws an error when `GalaxyBillingWrapper.getStorefront()` is called
- Adds docs explaining that the storefront functions/properties aren't supported for the Galaxy Store
- Adds a comment to `GalaxyBillingWrapper` to explicitly call out that in-app messages aren't supported


tonidero (approved):
🚢

PR #2907, opened 2025-12-09
Fix missing events tracking by removing delay to flush when backgrounding app

tonidero:
### Description
We had some issues with the events data we sent related to paywalls and customer center. This fixes that to minimize the chances of losing this data.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/PurchasesCommonTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/PurchasesCommonTest.kt
@@ -1155,7 +1155,7 @@ internal class PurchasesCommonTest: BasePurchasesTest() {
                 onSuccess = any(),
                 onError = any()
             )
-            mockEventsManager.flushEvents(any())
+            mockEventsManager.flushEvents(Delay.NONE)
         }
     }
 


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -491,7 +493,7 @@ internal class PurchasesFactory(
         override fun newThread(r: Runnable?): Thread {
             val wrapperRunnable = Runnable {
                 r?.let {
-                    android.os.Process.setThreadPriority(android.os.Process.THREAD_PRIORITY_LOWEST)
+                    android.os.Process.setThreadPriority(android.os.Process.THREAD_PRIORITY_BACKGROUND)
```
tonidero, line 496: This is not the fix but felt it would be better to use a higher priority for events, while still not impacting UI.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -640,7 +641,7 @@ internal class Backend(
                 eventsDispatcher,
                 paywallEventRequest.cacheKey,
                 onSuccessHandler to onErrorHandler,
-                Delay.LONG,
+                delay,
```
tonidero, line 644: The main fix is basically applying a `Delay.NONE` when the app is backgrounded, instead of a long delay like it was now. I still kept a default delay when foregrounding, to avoid making too many requests in case the app is foregrounded due to something else that could run on a schedule.

tonidero, line 644: Note that this fix doesn't apply to diagnostics currently. We only sync diagnostics on first app open, and it feels lower priority than events/ads. We could however revisit to also flush on app background though.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1299,7 +1299,7 @@ internal class PurchasesOrchestrator(
                 )
 
                 // Synchronize paywall events after a new purchase
-                flushPaywallEvents()
+                flushPaywallEvents(Delay.NONE)
```
ajpallares, line 1302: I think it makes sense to have no delay when flushing the events after a purchase, but wanted to make sure, since a comment above suggests that `Delay.NONE` was only added to flushes when the app got backgrounded.

purchases/src/test/java/com/revenuecat/purchases/PurchasesCommonTest.kt
```diff
@@ -1154,7 +1155,7 @@ internal class PurchasesCommonTest: BasePurchasesTest() {
                 onSuccess = any(),
                 onError = any()
             )
-            mockEventsManager.flushEvents()
+            mockEventsManager.flushEvents(any())
```
ajpallares: Maybe we could require the exact `Delay` expected? Not a blocker though

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Yup yup. I didn't mention it in that comment, but indeed I do think we should post events immediately after a purchase to minify chances of missing events in that scenario, which can be very useful as well.

PR #2908, opened 2025-12-09
Fetch Products from Galaxy Store

fire-at-will:
### Description
Supports fetching products from the Galaxy Store.

#### Notable Files
- `GalaxyBillingWrapper`: Main abstraction class for the Galaxy Store.
- `SerialRequestExecutor`: Responsible for ensuring that we only make one call to the Galaxy Store at a time, since calling multiple functions that make networks requests to the store is unsupported and causes the Samsung SDK to throw an exception.
- `ProductDataHandler`: Handles the fetching of products from the Samsung SDK.
- `StoreProductConversions`: Handles the conversions from Samsung `ProductVo` instances to our `GalaxyStoreProduct` class

#### Samsung IAP SDK Inclusion
We're now pulling in the Samsung IAP SDK from the local machine while we determine how the SDK should be distributed.
   - You can include the Samsung SDK on your machine by including the following in your `local.properties` file: `samsungIapSdkPath=/.../SamsungInAppPurchaseSDK_v6.5.0/Libs/samsung-iap-6.5.0.aar`
   - This means that **CI will fail** until we determine how the SDK should be properly distributed.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/IAPHelperProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/IAPHelperProvider.kt
@@ -1,8 +1,11 @@
 package com.revenuecat.purchases.galaxy
 
+import com.revenuecat.purchases.galaxy.utils.GalaxySerialOperation
 import com.samsung.android.sdk.iap.lib.listener.OnGetProductsDetailsListener
 
 internal interface IAPHelperProvider {
+
+    @GalaxySerialOperation
     fun getProductsDetails(
         productIDs: String,
         onGetProductsDetailsListener: OnGetProductsDetailsListener,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/DefaultIAPHelperProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/DefaultIAPHelperProvider.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.purchases.galaxy
 
+import com.revenuecat.purchases.galaxy.utils.GalaxySerialOperation
 import com.samsung.android.sdk.iap.lib.helper.IapHelper
 import com.samsung.android.sdk.iap.lib.listener.OnGetProductsDetailsListener
 
@@ -7,6 +8,7 @@ internal class DefaultIAPHelperProvider(
     val iapHelper: IapHelper,
 ) : IAPHelperProvider {
 
+    @GalaxySerialOperation
     override fun getProductsDetails(
         productIDs: String,
         onGetProductsDetailsListener: OnGetProductsDetailsListener,


purchases/build.gradle.kts
```diff
@@ -69,6 +69,10 @@ android {
                 it.exclude("com/revenuecat/purchases/backend_integration_tests/**")
             }
         }
+
+        // TODO: Remove this when we figure out how to properly integrate the Samsung SDK
+        // Avoid merging Android manifests for JVM unit tests to prevent minSdk conflicts from optional AARs.
+        unitTests.isIncludeAndroidResources = false
```
fire-at-will, line 76: This is a temporary hack that I introduced so that I can run the unit tests locally despite the fact that the Samsung SDK's minSDK != our minSDK requirement.

We should remove this when we properly add in the Samsung SDK as a dependency to our SDK.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -81,9 +110,7 @@ internal class GalaxyBillingWrapper(
         TODO("Not yet implemented")
     }
 
-    override fun isConnected(): Boolean {
-        TODO("Not yet implemented")
-    }
+    override fun isConnected(): Boolean = true
```
fire-at-will, line 122: As far as I can tell, connecting/disconnecting isn't really a thing for the Samsung SDK so long as you have an instance of the `IapHelper` class

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/constants/GalaxyErrorCode.kt
```diff
@@ -0,0 +1,7 @@
+package com.revenuecat.purchases.galaxy.constants
+
+@SuppressWarnings("MagicNumber")
+internal enum class GalaxyErrorCode(val code: Int, val description: String) {
```
fire-at-will, line 4: There are more error codes used by the Galaxy Store, we'll add them in as needed

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,182 @@
...
+            // returns all products for the app, which wouldn't give us the product we want (none).
+            log(LogIntent.DEBUG) { GalaxyStrings.EMPTY_GET_PRODUCT_DETAILS_REQUEST }
+            onReceive(emptyList())
+            return
+        }
+
+        if (inFlightRequest != null) {
+            log(LogIntent.GALAXY_ERROR) { GalaxyStrings.ANOTHER_GET_PRODUCT_DETAILS_REQUEST_IN_FLIGHT }
+            val error = PurchasesError(
+                code = PurchasesErrorCode.OperationAlreadyInProgressError,
```
tonidero, line 57: Hmm If the parameters matched, we could try to reuse the existing operation... I wonder if we should try to build some kind of queue for these, so we don't fail... But maybe it's an optimization we can work on later.

Specially considering we're executing these requests in serial, it seems this optimization is not really needed right now.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,182 @@
...
+                productType = productType,
+                onReceive = onReceive,
+                onError = onError,
+            )
+
+            if (productDataCache.keys.containsAll(productIds)) {
+                val cachedProducts = productIds.mapNotNull(productDataCache::get)
+
+                synchronized(this) {
+                    this.inFlightRequest = request
```
tonidero: I guess there is a small chance of a race condition between this being set, and when we check at the beginning of this method. I wonder if we should maybe only check whether there is an inflight request here?

Again, this probably doesn't apply right now since we're executing requests serially though.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,182 @@
...
+                // - A string with one product ID in it: queries for that one product
+                // - A string with multiple product IDs in it, delimited by a comma
+                val productIdRequestString = productIds.joinToString(separator = ",")
+                iapHelper.getProductsDetails(
+                    productIDs = productIdRequestString,
+                    onGetProductsDetailsListener = this,
+                )
+
+                synchronized(this) {
+                    this.inFlightRequest = request
```
tonidero: Same here about possible race conditions.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    private data class Request(
+        val productIds: Set<String>,
+        val productType: ProductType,
+        val onReceive: StoreProductsCallback,
+        val onError: PurchasesErrorCallback,
+    )
+
+    @get:Synchronized
+    internal val productDataCache = mutableMapOf<String, ProductVo>()
```
tonidero: Just wondering, would it be worth caching the `StoreProduct` itself instead of `ProductVo`? So we don't need to transform it on every call to get product details. Might be missing a reason why we need to cache the original ProductVo though.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,182 @@
...
+    private fun clearInFlightRequest() {
+        inFlightRequest = null
+    }
+
+    private fun addTimeoutToProductDataRequest(request: Request) {
+        mainHandler.postDelayed(
+            {
+                val shouldHandleTimeout = synchronized(this) {
+                    if (inFlightRequest === request) {
+                        clearInFlightRequest()
```
tonidero: Hmm I'm thinking this won't actually cancel the Galaxy products request right? It will just allow to start a new one if I'm correct, which could cause an issue since you mentioned that we can't perform simultaneous requests that do network requests.

TBH, not sure there is a way to cancel an inflight request with the Galaxy SDK, looking only at the API they provide 😢. Considering this, I think we might want to be more permissive in our timeout here, and allow a long time for the operation to finally return from the Galaxy store? Otherwise we might be in an inconsistent state. Alternatively, we could even leave it without a timeout? I guess it depends on whether  we've been able to reproduce this getting stuck...

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
tonidero: Maybe we can just indicate it in the `startConnection` and `endConnection` methods in this class that they should no-op instead of a warning as we have now?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
```diff
@@ -0,0 +1,114 @@
...
+
+internal fun ProductVo.toStoreProduct(): StoreProduct {
+    val type = this.type.createRevenueCatProductTypeFromSamsungIAPTypeString()
+    val period: Period? = if (type == ProductType.SUBS) {
+        this.createPeriod()
+    } else {
+        null
+    }
+
+    return GalaxyStoreProduct(
```
tonidero, line 26: Haven't checked the purchase API but will we need the `ProductVo` to initate a purchase? If so, maybe we want to introduce it as an `internal` property of the `GalaxyStoreProduct` class. But this can be done in follow-up PRs though.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
```diff
@@ -0,0 +1,114 @@
...
+        id = this.itemId,
+        type = type,
+        price = this.createPrice(),
+        name = this.itemName,
+        title = this.itemName,
+        description = this.itemDesc,
+        period = period,
+        subscriptionOptions = null,
+        defaultOption = null,
```
tonidero, line 35: I think it would be great that we also build the subscriptionOptions/defaultOption at least for subscription, so we keep a similar API than we do for Google. We don't for Amazon, since we thought it wasn't needed, but I think it would be great to do it there as well, so the same data types can be access for all stores. Again, can be done in a separate PR.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.revenuecat.purchases.common.log
+import com.revenuecat.purchases.models.Period
+import com.revenuecat.purchases.models.Price
+import com.revenuecat.purchases.models.StoreProduct
+import com.samsung.android.sdk.iap.lib.vo.ProductVo
+
+internal fun ProductVo.toStoreProduct(): StoreProduct {
+    val type = this.type.createRevenueCatProductTypeFromSamsungIAPTypeString()
+    val period: Period? = if (type == ProductType.SUBS) {
+        this.createPeriod()
```
tonidero: Should we fail here if the period couldn't be parsed for a subscription product? Seems that shouldn't happen, so it would be good to fail (gracefully) in case it does 😅

purchases/src/main/kotlin/com/revenuecat/purchases/utils/SerialRequestExecutor.kt
```diff
@@ -0,0 +1,67 @@
...
+        if (shouldStartNext) {
+            executeNextRequest()
+        }
+    }
+
+    private fun executeNextRequest() {
+        val nextRequest = synchronized(lock) {
+            pendingRequests.firstOrNull()
+        } ?: run {
+            synchronized(lock) { hasActiveRequest = false }
```
tonidero: There is this super small chance of a race condition of this getting set to false if we check in thread 1 that there is nothing in the `pendingRequests` and right after but before this block is executed, on thread 2 the synchronized block in the `executeSerially` method above is executed.

Could we unify this to the `synchronized` block above that gets the first pending request?

purchases/src/main/kotlin/com/revenuecat/purchases/utils/SerialRequestExecutor.kt
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if (shouldStartNext) {
+            executeNextRequest()
+        }
+    }
+
+    private fun executeNextRequest() {
+        val nextRequest = synchronized(lock) {
+            pendingRequests.firstOrNull()
```
tonidero: Are we allowing retries of requests here? Seems like we don't so maybe we could just remove the request to be executed here instead of doing it on the `onRequestFinished`? TBH, this seems more of a style thing, I don't think it would affect behavior at all.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: If the `SerialRequestExecutor` is doing it's job properly, this `if` block should never be executed, it's really here as an extra layer of protection 🤞

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Honestly, I think that all of the synchronization in this class is _probably_ unnecessary since we're executing the requests serially. That said, would it just be safer to make this entire `getProductDetails()` function synchronized by annotating it with `@Synchronized`?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Oooh, that's a good idea! Honestly, the only reason is so that we can re-use the same `handleSuccessfulProductsResponse()` function, which takes in `ProductVo`s as a param.

I'll move the code around a little so that we can cache the `StoreProduct` instead :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
tonidero: Hmm well, I guess it's dangerous to synchronize on a possibly long block of code. Specially considering that may be doing network requests (the actual fetching of products). So I would be against that. I do agree it's probably unnecessary considering we're executing all of this serially, but just concerning that it's not part of the contract, and this could potentially be executed from somewhere else (though it shouldn't)... Still, I guess I wouldn't consider it a blocker

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Made this optimization in ef187747b8dfe2d44cef8c94abbd3b9cad2af4e6 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Oof, great point. FWIW, I haven't been able to produce a timeout yet, but I'm running on a brand new phone made in 2025 on great WiFi, so those aren't the best conditions for reproducing time-outs.

Even if we were able to produce what the `ProductDataHandler` would consider a timeout, you're right, this code would leave us in a bad state since `IapHelper` has no mechanism to cancel an in-flight request. We'd allow execution to continue, and all subsequent requests to the would fail when they call the `IapHelper` functions until the `IapHelper` declares that the original request has timed out. Given that, it's probably best to remove the timeout altogether. Are you good with that approach?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
fire-at-will: You'd expect it to take in a `ProductVo` (I did too!), but it just takes in a string of the product's ID! The SDK's APIs are all almost exclusively `Strings`, which makes things both simpler and more complex 😅

```java
public boolean startPayment(@NotNull String _itemId, @NotNull OnPaymentListener _onPaymentListener) {
    return this.executePayment(_itemId, (String)null, (String)null, (String)null, _onPaymentListener);
}

public boolean startPayment(@NotNull String _itemId, @Nullable String _obfuscatedAccountId, @Nullable String _obfuscatedProfileId, @NotNull OnPaymentListener _onPaymentListener) {
    return this.executePayment(_itemId, (String)null, _obfuscatedAccountId, _obfuscatedProfileId, _onPaymentListener);
}
```

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
fire-at-will: That makes sense! I'll work on this in a follow-up PR :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
tonidero: 😅 Oh well, I guess in that case, nvm 😂

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
tonidero: Yeah, I think we can start with that. If we get evidence of this possibly taking a long time, we can reconsider it. But for now, we will have to trust the galaxy SDK not to leave us hanging 🥁

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Added no-op comments in 7b314530520f25afc2a37b5cfaefa0ba6b8d3fcb 👍

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
fire-at-will: It looks like for both Amazon and Google, we just pass in a null period without doing any checking for the product even though it's type is still subscription 😅

- Amazon: https://github.com/RevenueCat/purchases-android/blob/123f0bc295042a816a73c2edf3382791837afbc6/purchases/src/main/kotlin/com/revenuecat/purchases/amazon/storeProductConversions.kt#L29
- Google: https://github.com/RevenueCat/purchases-android/blob/123f0bc295042a816a73c2edf3382791837afbc6/purchases/src/main/kotlin/com/revenuecat/purchases/google/storeProductConversions.kt#L39

Maybe we should keep the same behavior here for consistency? At the very least, I'll add in a log for it 😎

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
tonidero: Might be worth removing all the synchronous code, and instead create a single thread to execute all these Galaxy operations. We've created single thread executors for backend requests, and we could follow a similar approach. See: https://github.com/RevenueCat/purchases-android/blob/123f0bc295042a816a73c2edf3382791837afbc6/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt#L485

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
fire-at-will: Log added here: 2449973da8d16411cad4f04f33e8956e3d4ebb95 👍

purchases/src/main/kotlin/com/revenuecat/purchases/utils/SerialRequestExecutor.kt
fire-at-will: If it can happen, it probably will at some point! Good catch, unified the synchronized block here: d4c961afd954cff218175fd8aeea3fd6283ee13c

purchases/src/main/kotlin/com/revenuecat/purchases/utils/SerialRequestExecutor.kt
fire-at-will: > Are we allowing retries of requests here?
We don't have retries in the `SerialRequestExecutor` level today, it's up to consumers to implement that if they want it!

If we remove the request in `executeNextRequest()`, then we'd lose a reference to the current in-flight request. I don't think that affects the class's behavior, but it could make it more difficult to debug any issues that pop up. I could go either way on this, do you have strong opinions either way?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Removed the timeout functionality from `ProductDataHandler` in c474fee2ea7cb756f050a3ada6fe62b82b829d56 👍

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/StoreProductConversions.kt
tonidero: Oh right... I guess we indeed can keep the same behavior for consistency 👍

purchases/src/main/kotlin/com/revenuecat/purchases/utils/SerialRequestExecutor.kt
tonidero: Hmm TBH, I think either behavior is fine, so let's keep it as is :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Would this have any advantages over the `SerialRequestExecutor`/synchronized model we're using now? I'm not super familiar with the executors, so I asked Codex about it, and it gave me this:

> A single-thread ScheduledExecutorService only serializes the submitted runnables. Our requests are async: we enqueue a block, it kicks off work, returns
  immediately, and completes later via a callback. That means the executor thread would finish the runnable while the request is still in flight, and the next runnable
  would start right away—breaking the one-at-a-time guarantee. SerialRequestExecutor keeps the head request “active” until its finish is called, so even though the work
  hops off to other threads, we don’t start the next until the callback signals completion. To drop it, we’d need every request to fully execute and finish within the
  executor’s runnable, or refactor to chain scheduling on callback completion, which is what SerialRequestExecutor already does.

tonidero: Well, this would be mostly a replacement for the synchronized model we have, not for the SerialRequestExecutor, since I assume the galaxy SDK operations are done in another thread, so that wouldn't protect us against doing multiple requests to the galaxy SDK.

And well, as I mentioned, it could potentially simplify the synchronization. But well, considering we have the SerialRequestExecutor, we can probably get away with removing all that, and documenting well that requests are expected to be performed in serial... Not ideal in case we miss that constraint, since we don't have any compile time guarantees... but might be the simplest for now.

fire-at-will: Sounds good! Removed all `synchronized` blocks in 43a226ff73bf057212669b4b9abc555939555032

I also introduced a new `@GalaxySerialOperation` annotation that I applied to the `getProductDetails()` function. This annotation acts similarly to how the `@ExperimentalPreviewRevenueCatPurchasesAPI` works, where it forces any consumers of the annotated functions to explicitly opt-in to it. This will at least ensure that in the future, any other developer that tries to call these functions must acknowledge that the function must be called serially with other annotated functions. Let me know what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,146 @@
...
+            handleStoreProducts(
+                storeProducts = cachedProducts,
+            )
+        } else {
+            // When requesting products from the Samsung IAP SDK, the `_productIds` param is a string where
+            // the following contents product the following results:
+            // - An empty string: queries all products
+            // - A string with one product ID in it: queries for that one product
+            // - A string with multiple product IDs in it, delimited by a comma
+            val productIdRequestString = productIds.joinToString(separator = ",")
```
tonidero, line 86: I guess if we have some of the products in the cache, we would still be querying for all of them. Probably a small optimization though, so not a blocker.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -0,0 +1,146 @@
...
+            iapHelper.getProductsDetails(
+                productIDs = productIdRequestString,
+                onGetProductsDetailsListener = this,
+            )
+
+            this.inFlightRequest = request
+        }
+    }
+
+    override fun onGetProducts(error: ErrorVo, products: ArrayList<ProductVo>) {
```
tonidero, line 96: Just to check, the API would return either a non-success ErrorVo or a list of products, but not both correct? For example, if part of the productIds requested do not exist, or we're passing an incorrect productId format for some of them, would it return the valid products, without an error? Or would it return the valid ones + an error? 

In case it's the second case, we probably should log the error but still return a sucess result I think.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/IAPHelperProvider.kt
```diff
@@ -0,0 +1,10 @@
+package com.revenuecat.purchases.galaxy
+
+import com.samsung.android.sdk.iap.lib.listener.OnGetProductsDetailsListener
+
+internal interface IAPHelperProvider {
+    fun getProductsDetails(
```
tonidero, line 9: I'm wondering if we should also annotate this method with the `@GalaxySerialRequest`, so if/when we add other operation helpers, it's easier to remember we need to add that annotation there as well.

purchases/build.gradle.kts
```diff
@@ -149,6 +154,20 @@ fun obtainTestBuildType(): String =
...
     }
 
+// TO DO: Bring in Samsung SDK from somewhere else
+val samsungIapSdkPath = providers.provider {
+    providers.gradleProperty("samsungIapSdkPath").orNull
+        ?: providers.environmentVariable("SAMSUNG_IAP_SDK_PATH").orNull
+        ?: localProperties.getProperty("samsungIapSdkPath")
+}.map { path ->
+    val aar = file(path)
+    check(aar.exists()) {
```
tonidero, line 164: I guess this will fail to work if the aar is not set somehow... we shouldn't get this merged to main, but it's ok to keep it on the integration branch. Hopefully we can figure out a way to integrate it 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Will add this to a list of future optimizations to make!

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/IAPHelperProvider.kt
fire-at-will: This is a great idea! Added in a8a0d840f6ff6f8839d2d9aaab7b434292d60e49

purchases/build.gradle.kts
fire-at-will: Yeah, this is 1000% a temporary thing and should not be merged into main. We'll work with @MarkVillacampa at some point to find a proper way to integrate the Samsung SDK into our SDK 😄

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: I just tested this out! If you request 2 valid product IDs and 1 invalid product IDs, it returns:
- A non-error (success) `ErrorVo`
- An ArrayList with the 2 valid products 😅

I'll see if we can create an error log for this in a follow-up PR

fire-at-will: Created Linear ticket CAT-2148 to track this

PR #2914, opened 2025-12-10
Fix custom schemes on paywall purchase buttons

tonidero:
### Description
If a user used a URL with a custom scheme in a purchase button deep link. It failed to parse using `URL`. Using `URI` instead to be more permissive.


ajpallares (approved):
Makes sense!

JayShortway (approved):

PR #2917, opened 2025-12-10
Galaxy Attribution Fetcher

fire-at-will:
### Description
Without an AttributionFetcher for the Galaxy Store, the SDK will crash at runtime when configured for the Galaxy Store, so we need to create one. Since the Samsung IAP SDK doesn't support fetching device identifiers, this PR creates a no-op `GalaxyDeviceIdentifiersFetcher` that always returns 0 identifiers, and wires it into the `AttributionFetcherFactory`.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/AttributionFetcherFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/AttributionFetcherFactory.kt
@@ -1,8 +1,8 @@
 package com.revenuecat.purchases
 
+import com.revenuecat.purchases.amazon.attribution.AmazonDeviceIdentifiersFetcher
 import com.revenuecat.purchases.common.Dispatcher
 import com.revenuecat.purchases.common.errorLog
-import com.revenuecat.purchases.common.subscriberattributes.DeviceIdentifiersFetcher
 import com.revenuecat.purchases.galaxy.attribution.GalaxyDeviceIdentifiersFetcher
 import com.revenuecat.purchases.google.attribution.GoogleDeviceIdentifiersFetcher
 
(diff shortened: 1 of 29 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/AttributionFetcherFactory.kt
```diff
@@ -22,6 +22,16 @@ internal object AttributionFetcherFactory {
                 throw e
             }
         }
+        Store.GALAXY -> {
+            try {
+                Class.forName("com.revenuecat.purchases.galaxy.attribution.GalaxyDeviceIdentifiersFetcher")
```
tonidero: Hmm I understand this looks like this for Amazon, but looking at it, this seems to be a leftover from before we refactored the modules to have the amazon code be part of the purchases module as a compile only dependency.

Basically, I don't think we need to use reflection to instantiate this class since it will always be there (it's part of the same module) and we can just do `GalaxyDeviceIdentifiersFetcher()`.

I think the same applies to the amazon one, but we can split that one in a separate PR to main.

tonidero (commented):
Just a simplification that I think would be nice to have

purchases/src/main/kotlin/com/revenuecat/purchases/AttributionFetcherFactory.kt
fire-at-will: Removed the reflection for the GalaxyDeviceIdentifiersFetcher in 3db7de76b7225bc384a81f0296e896bb0f20d8fb

I'll remove it for Amazon in a follow-up PR into main to clean it up :)

fire-at-will: PR to remove reflection from the Amazon instantiation here: https://github.com/RevenueCat/purchases-android/pull/2919 :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/attribution/GalaxyDeviceIdentifiersFetcher.kt
```diff
@@ -0,0 +1,15 @@
...
+import android.app.Application
+import com.revenuecat.purchases.common.subscriberattributes.DeviceIdentifiersFetcher
+
+internal class GalaxyDeviceIdentifiersFetcher : DeviceIdentifiersFetcher {
+
+    override fun getDeviceIdentifiers(
+        applicationContext: Application,
+        completion: (deviceIdentifiers: Map<String, String>) -> Unit,
+    ) {
+        // Samsung IAP library doesn't support fetching device identifiers, so this is a no-op
```
tonidero, line 12: For Amazon, we also don't get this from their IAP library, but from the device settings. Is that something that would work for Galaxy devices? Can be done in a separate PR though of course!

tonidero (approved):
Makes sense 🚢 !

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/attribution/GalaxyDeviceIdentifiersFetcher.kt
fire-at-will: Yeah, I don't see why not! We can come back and do that in another PR, I think for now we've cut the device identifiers from the scope to ship quickly

PR #2919, opened 2025-12-11
Don't use reflection to instantiate AmazonDeviceIdentifiersFetcher

fire-at-will:
### Description
Follow-up PR from [this comment thread](https://github.com/RevenueCat/purchases-android/pull/2917#discussion_r2610896275) on PR https://github.com/RevenueCat/purchases-android/pull/2917.

We no longer need to use reflection to instantiate the `AmazonDeviceIdentifiersFetcher` class since it is now in the main module of the SDK, so this PR changes the `AmazonDeviceIdentifiersFetcher`'s instantiation from using reflection to using a normal constructor.


tonidero (approved):
Thanks for cleaning this up!! 🙌

PR #2922, opened 2025-12-11
Galaxy Store Purchasing Happy Path

fire-at-will:
### Description
This PR enables support for making purchases through the Galaxy Store. To keep the PR size down, it primarily considers the happy path where the purchase succeeds with no issues.

It:
- Sets the `IapHelper`'s operation mode when the `GalaxyBillingWrapper` is initialized.
- Forwards purchase requests to the `IapHelper` through the `PurchaseHandler` class
- Converts `PurchaseVo` responses to `StoreTransactions` for successful purchases
- Forwards errors to the purchase listener, but doesn't attempt to parse specific errors
- Adds unit tests :)

To keep the PR size down, the PR does not:
- Acknowledge purchases with Samsung
- Attempt to parse specific error codes from the `PurchaseVo` object.
We'll tackle these features in future PRs.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
@@ -27,7 +27,7 @@ internal fun PurchaseVo.toStoreTransaction(
         productIds = listOf(productId),
         type = type,
         purchaseTime = purchaseDate.time,
-        purchaseToken = this.orderId,
+        purchaseToken = this.purchaseId,
         purchaseState = purchaseState,
         isAutoRenewing = type == ProductType.SUBS,
         signature = null,
--- a/purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
@@ -209,28 +209,6 @@ class GalaxyBillingWrapperTest : GalaxyStoreTest() {
 
     @OptIn(GalaxySerialOperation::class)
     @Test
-    fun `makePurchaseAsync exits when finishTransactions is false`() {
-        val wrapper = createWrapper(finishTransactions = false)
-        val purchasesUpdatedListener = mockk<BillingAbstract.PurchasesUpdatedListener>(relaxed = true)
-        wrapper.purchasesUpdatedListener = purchasesUpdatedListener
-
-        wrapper.makePurchaseAsync(
-            activity = mockk<Activity>(),
-            appUserID = "user",
-            purchasingData = GalaxyPurchasingData.Product(createStoreProduct()),
-            replaceProductInfo = null,
-            presentedOfferingContext = null,
-            isPersonalizedPrice = null,
-        )
-
-        verify(exactly = 0) {
-            purchaseHandlerMock.purchase(any(), any(), any(), any())
-        }
-        verify { purchasesUpdatedListener wasNot Called }
-    }
-
-    @OptIn(GalaxySerialOperation::class)
-    @Test
     fun `makePurchaseAsync exits when replaceProductInfo is provided`() {
         val wrapper = createWrapper()
         wrapper.purchasesUpdatedListener = mockk(relaxed = true)
--- a/purchases/src/test/java/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversionsTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversionsTest.kt
@@ -25,6 +25,7 @@ class StoreTransactionConversionsTest {
     private val purchaseDateString = "2024-01-15 13:45:20"
     private val presentedOfferingContext = PresentedOfferingContext("offering_id")
     private val defaultOrderId = "order-123"
+    private val defaultPurchaseId = "purchase-456"
     private lateinit var originalTimeZone: TimeZone
 
     @BeforeTest
@@ -43,6 +44,7 @@ class StoreTransactionConversionsTest {
         val productId = "expected_product_id"
         val purchaseVo = createMockPurchaseVo(
             orderId = defaultOrderId,
+            purchaseId = defaultPurchaseId,
             purchaseDate = purchaseDateString,
             type = "item",
             itemId = "original_item_id",
@@ -55,7 +57,7 @@ class StoreTransactionConversionsTest {
         )
 
         assertThat(storeTransaction.orderId).isEqualTo(defaultOrderId)
-        assertThat(storeTransaction.purchaseToken).isEqualTo(defaultOrderId)
+        assertThat(storeTransaction.purchaseToken).isEqualTo(defaultPurchaseId)
         assertThat(storeTransaction.productIds).containsExactly(productId)
         assertThat(storeTransaction.purchaseState).isEqualTo(PurchaseState.PURCHASED)
         assertThat(storeTransaction.purchaseTime)
@@ -75,6 +77,7 @@ class StoreTransactionConversionsTest {
     fun `toStoreTransaction uses provided product id`() {
         val purchaseVo = createMockPurchaseVo(
             orderId = defaultOrderId,
+            purchaseId = defaultPurchaseId,
             purchaseDate = purchaseDateString,
             type = "subscription",
             itemId = "original_item_id",
@@ -94,6 +97,7 @@ class StoreTransactionConversionsTest {
     fun `toStoreTransaction maps subscription type and sets auto renewing`() {
         val purchaseVo = createMockPurchaseVo(
             orderId = defaultOrderId,
+            purchaseId = defaultPurchaseId,
             purchaseDate = purchaseDateString,
             type = "subscription",
         )
@@ -112,6 +116,7 @@ class StoreTransactionConversionsTest {
     fun `toStoreTransaction maps inapp type and disables auto renewing`() {
         val purchaseVo = createMockPurchaseVo(
             orderId = defaultOrderId,
+            purchaseId = defaultPurchaseId,
             purchaseDate = purchaseDateString,
             type = "item",
         )
@@ -130,6 +135,7 @@ class StoreTransactionConversionsTest {
     fun `toStoreTransaction maps unknown type to UNKNOWN product type`() {
         val purchaseVo = createMockPurchaseVo(
             orderId = defaultOrderId,
+            purchaseId = defaultPurchaseId,
             purchaseDate = purchaseDateString,
             type = "weird_type",
         )
@@ -148,6 +154,7 @@ class StoreTransactionConversionsTest {
     fun `toStoreTransaction throws IllegalArgumentException for invalid purchase date`() {
         val purchaseVo = createMockPurchaseVo(
             orderId = defaultOrderId,
+            purchaseId = defaultPurchaseId,
             purchaseDate = "not-a-date",
             type = "item",
         )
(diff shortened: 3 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/DefaultIAPHelperProvider.kt
```diff
@@ -1,13 +1,24 @@
...
+import com.samsung.android.sdk.iap.lib.constants.HelperDefine
 import com.samsung.android.sdk.iap.lib.helper.IapHelper
 import com.samsung.android.sdk.iap.lib.listener.OnGetProductsDetailsListener
+import com.samsung.android.sdk.iap.lib.listener.OnPaymentListener
 
 internal class DefaultIAPHelperProvider(
     val iapHelper: IapHelper,
 ) : IAPHelperProvider {
 
+    override fun setOperationMode(
```
fire-at-will, line 15: This function doesn't need to be `@GalaxySerialOperation`. Since the operation doesn't make a network request, Samsung doesn't take this function into account when restricting which operations can be performed in parallel.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/PurchaseHandler.kt
```diff
@@ -0,0 +1,125 @@
...
+        if (!requestWasDispatched) {
+            log(LogIntent.GALAXY_ERROR) { GalaxyStrings.GALAXY_STORE_FAILED_TO_ACCEPT_PAYMENT_REQUEST }
+            onError(
+                PurchasesError(
+                    code = PurchasesErrorCode.StoreProblemError,
+                    underlyingErrorMessage = "The Galaxy Store failed to accept the purchase request.",
+                ),
+            )
+            clearInFlightRequest()
+            return
```
fire-at-will, line 74: This return isn't really necessary, but I included it in case we ever add anything beneath this block in the future :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/PurchaseHandler.kt
```diff
@@ -0,0 +1,125 @@
...
+        val underlyingErrorMessage = error.errorString
+        log(LogIntent.GALAXY_ERROR) {
+            GalaxyStrings.PURCHASE_REQUEST_ERRORED.format(
+                inFlightRequest?.storeProduct?.id ?: "[none]",
+                underlyingErrorMessage,
+            )
+        }
+
+        // TO DO: Map galaxy errors to PurchaseErrors so we can give better errors
+        val purchasesError = PurchasesError(PurchasesErrorCode.StoreProblemError, underlyingErrorMessage)
```
fire-at-will, line 116: We'll flesh this out and parse specific error codes in a future PR

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    // throws IllegalArgumentException if the string is an invalid format
+    val purchaseDate = this.purchaseDate.parseDateFromGalaxyDateString()
+
+    return StoreTransaction(
+        orderId = this.orderId,
+        productIds = listOf(productId),
+        type = type,
+        purchaseTime = purchaseDate.time,
+        purchaseToken = this.orderId,
```
tonidero: So the orderId is what we need in the backend to validate the purchase, correct?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/GalaxyDateParser.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.revenuecat.purchases.galaxy.GalaxyStrings
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+
+internal fun String.parseDateFromGalaxyDateString(): Date {
+    val formatter = SimpleDateFormat("yyyy-MM-dd HH:mm:ss", Locale.ROOT).apply {
+        timeZone = TimeZone.getDefault()
```
tonidero, line 14: Hmm is the date sent from the galaxy library timezone aware? Not sure if maybe matching the Samsung account timezone? We should make sure it's correct on different device timezones and with different Samsung account timezones

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/DefaultIAPHelperProvider.kt
```diff
@@ -18,4 +29,23 @@ internal class DefaultIAPHelperProvider(
...
+    override fun startPayment(
+        itemId: String,
+        obfuscatedAccountId: String?,
+        obfuscatedProfileId: String?,
+        onPaymentListener: OnPaymentListener,
+    ): Boolean {
+        // Return values:
+        // true: The request was sent to server successfully and the result will be sent
+        //       to OnPaymentListener interface listener.
+        // false: The request was not sent to server and was not processed.
```
tonidero, line 43: Would have been great to get an error in this case, instead of a boolean :(

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -116,7 +137,70 @@ internal class GalaxyBillingWrapper(
...
+                    "GalaxyPurchasingData",
+                ),
+            )
+            errorLog(error)
+            purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
+            return
+        }
+        val storeProduct = galaxyPurchaseInfo.storeProduct
+
+        if (!shouldFinishTransactions()) { return }
```
tonidero: Hmm I think the meaning of finishing a transaction is wrong here... In this case, seems we wouldn't be finishing the transaction at all. Whereas it should mean we won't actually "acknowledge" and/or "consume" the purchase. But we would still allow purchasing with our SDK even with finishTransactions set to false.

However, I'm not sure if Galaxy store has a similar "acknowledge" and/or "consume" step to consider this 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -150,4 +234,17 @@ internal class GalaxyBillingWrapper(
...
+
+    private fun onPurchaseError(error: PurchasesError) {
+        purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
+    }
+
+    private fun shouldFinishTransactions(): Boolean {
+        return if (finishTransactions) {
+            true
+        } else {
+            log(LogIntent.AMAZON_WARNING) { AmazonStrings.WARNING_AMAZON_NOT_FINISHING_TRANSACTIONS }
```
tonidero: This shouldn't be Amazon :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/GalaxyDateParser.kt
fire-at-will: Yeah, we should probably get confirmation from Samsung on this. I'll see if I can find it in the SDK's decompiled code, and if not, we can see if we can ask Samsung. In my testing, it came back in my device's timezone (which for some reason is 1 hour ahead of my own timezone 😅)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
fire-at-will: I believe so, but will confirm with the backend engineers!

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/DefaultIAPHelperProvider.kt
fire-at-will: Yeah...it's odd that they don't just send an `ErrorVo` to the purchase listener 🤷‍♂️

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Yeah, the Galaxy Store has a concept of acknowledging purchases, and it's something we'll need to implement in a follow up PR. Here are their docs on it: https://developer.samsung.com/iap/programming-guide/iap-helper-programming.html#Notify-Samsung-IAP-that-the-purchase-was-processed

I think you're right, maybe we shouldn't do this here. I included this because this is what we're doing today on Amazon, not sure if that's a bug or not: https://github.com/RevenueCat/purchases-android/blob/5589c06eb9601cc20e65f808706dd805dbf5c8b9/purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt#L294

I'll go ahead and remove this check here for now, and then when we implement the logic to acknowledge Galaxy purchases, we can surround that with an `if(shouldFinishTransactions())` check

fire-at-will: Removed the check in c20d602bb6465bbeee80d91a2f5585d18369a971 :)

tonidero: Hmm that's indeed a good point... we do allow purchasing in Google when not finishing transactions, but seems we don't for Amazon... From a conceptual POV, I think not allowing to purchase when we're not going to finish transactions might make sense (we can discuss it with the team), but in that case, we should return an error, unlike what we have currently in Amazon, where it seems it would be left hanging 😬

tonidero: But yeah, allowing to purchase, and just not handle the acknowledge makes sense, since it's what we do for Google

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
fire-at-will: Checked with the backend engineers, and this value should be the `PurchaseVo`'s `purchaseId`. Updated this to use `purchaseId` in b978aa3d16142de8331828915205ac525bbc5b3b 👍

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/GalaxyDateParser.kt
fire-at-will: @tonidero I think we're good here now. 

For future reference in case anyone sees this, it looks like this is how the date string is created:
1. Samsung Store server sends down a timestamp in a JSON object
2. PurchaseVo gets created from that JSON
2. When PurchaseVo is created, it’s using simple DateFormat.format parsing:

```java
String result = "";
String dateFormat = "yyyy-MM-dd HH:mm:ss";
result = DateFormat.format(dateFormat, _timeMills).toString();
```

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Removed this helper for now since we don't need finishTransactions at the moment 👍

PR #2925, opened 2025-12-12
Improve accuracy of transactions origin Part 1: Refactor to allow caching transaction metadata

tonidero:
### Description
The final objective of these PRs are to have more reliability on all transaction metadata that normally gets sent with a purchase POST /receipt. In rare scenarios, when this request fails, the transaction metadata is currently lost. This will cache this information on disk, so we can recover it and post it (in follow-up PRs).

Main changes in this PR:
- Made `ReceiptInfo` serializable by making it include only the information that we currently post to the backend. This cascaded into a whole lot of changes unfortunately. 
- Move pending transactions handling to the PostReceiptHelper, so we can cache metadata for these pending purchases in the same location we currently cache non pending transactions.


purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -145,10 +147,21 @@ internal class PostReceiptHelper(
         storeUserID: String?,
         marketplace: String?,
         initiationSource: PostReceiptInitiationSource,
+        purchaseState: PurchaseState,
         onSuccess: (PostReceiptResponse) -> Unit,
         onError: PostReceiptDataErrorCallback,
     ) {
         val presentedPaywall = paywallPresentedCache.getAndRemovePresentedEvent()
+
+        if (purchaseState == PurchaseState.PENDING) {
```
tonidero, line 156: This check has been moved from the previous position in the `PostTransactionWithProductDetailsHelper`. Plan is to have pending transactions get to this point. And before returning an error here, cache all transaction metadata, unifying where we cache all this transaction metadata into a single place.

purchases/src/main/kotlin/com/revenuecat/purchases/PostTransactionWithProductDetailsHelper.kt
```diff
@@ -29,84 +27,77 @@ internal class PostTransactionWithProductDetailsHelper(
         transactionPostError: (ErrorPurchaseCallback)? = null,
     ) {
         transactions.forEach { transaction ->
-            if (transaction.purchaseState != PurchaseState.PENDING) {
```
tonidero, line 32: This is pretty much the same as before but just removing this check, which now happens internally in the `PostReceiptHelper`

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -3,52 +3,37 @@ package com.revenuecat.purchases.common
...
 import com.revenuecat.purchases.models.SubscriptionOption
+import kotlinx.parcelize.IgnoredOnParcel
+import kotlinx.serialization.Serializable
 
-@SuppressWarnings("LongParameterList")
-internal class ReceiptInfo
-@OptIn(InternalRevenueCatAPI::class)
-constructor(
+@Serializable
+internal data class ReceiptInfo(
```
tonidero, line 16: In here are the biggest changes in this PR, from which most others derive. Basically, we've changed the type here to only store the information we need to have in the `POST /receipt` and made this serializable, so we can store it on disk. This had ramifications, specially in tests.

purchases/src/test/java/com/revenuecat/purchases/PostTransactionWithProductDetailsHelperTest.kt
```diff
@@ -63,49 +63,6 @@ class PostTransactionWithProductDetailsHelperTest {
         )
     }
 
-    @Test
-    fun `if pending transaction, error callback is called`() {
```
tonidero, line 67: These tests are added back in a follow-up PR: #2979

PR #2927, opened 2025-12-12
Properly Map Galaxy Errors -> PurchasesError

fire-at-will:
### Description
This PR extends the `Error.toPurchasesError()` function to properly map all error codes that are documented to us in the Samsung SDK's documentation to a `PurchaseError` with the most appropriate `PurchaseErrorCode`. This will allow us to return the most descriptive error as we can when we receive an error from the Galaxy Store.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/galaxy/utils/ErrorVoUtilsTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/galaxy/utils/ErrorVoUtilsTest.kt
@@ -37,7 +37,7 @@ class ErrorVoUtilsTest {
 
     @Test
     fun `toPurchasesError maps known galaxy error codes to purchases error codes`() {
-        val expectedMappings = listOf(
+        val expectedMappings = mapOf(
             GalaxyErrorCode.IAP_PAYMENT_IS_CANCELED to PurchasesErrorCode.PurchaseCancelledError,
             GalaxyErrorCode.IAP_ERROR_ALREADY_PURCHASED to PurchasesErrorCode.ProductAlreadyPurchasedError,
             GalaxyErrorCode.IAP_ERROR_PRODUCT_DOES_NOT_EXIST to PurchasesErrorCode.ProductNotAvailableForPurchaseError,
@@ -57,7 +57,8 @@ class ErrorVoUtilsTest {
             GalaxyErrorCode.IAP_ERROR_NONE to PurchasesErrorCode.UnknownError,
         )
 
-        expectedMappings.forEach { (galaxyCode, purchasesErrorCode) ->
+        GalaxyErrorCode.values().forEach { galaxyCode ->
+            val purchasesErrorCode = expectedMappings.getValue(galaxyCode)
             val message = "message for ${galaxyCode.name}"
             val errorVo = mockErrorVo(galaxyCode.code, message)
 


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/constants/GalaxyErrorCode.kt
```diff
@@ -1,7 +1,22 @@
...
+    IAP_ERROR_PRODUCT_DOES_NOT_EXIST(-1005),
+    IAP_ERROR_CONFIRM_INBOX(-1006),
+    IAP_ERROR_ITEM_GROUP_DOES_NOT_EXIST(-1007),
+    IAP_ERROR_NETWORK_NOT_AVAILABLE(-1008),
+    IAP_ERROR_IOEXCEPTION_ERROR(-1009),
+    IAP_ERROR_SOCKET_TIMEOUT(-1010),
+    IAP_ERROR_CONNECT_TIMEOUT(-1011),
+    IAP_ERROR_NOT_EXIST_LOCAL_PRICE(-1012),
+    IAP_ERROR_NOT_AVAILABLE_SHOP(-1013),
+    IAP_ERROR_INVALID_ACCESS_TOKEN(-1015),
```
fire-at-will, line 23: This isn't a typo, there is no error with the code `-1014` in the docs: https://developer.samsung.com/iap/programming-guide/iap-helper-programming.html#ErrorVo-and-Response-code

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/ErrorVoUtils.kt
```diff
@@ -1,8 +1,44 @@
...
+        -> PurchasesErrorCode.NetworkError
+        GalaxyErrorCode.IAP_ERROR_WHILE_RUNNING -> PurchasesErrorCode.PurchaseInvalidError
+        GalaxyErrorCode.IAP_ERROR_NEED_APP_UPGRADE,
+        GalaxyErrorCode.IAP_ERROR_INITIALIZATION,
+        GalaxyErrorCode.IAP_ERROR_COMMON,
+        GalaxyErrorCode.IAP_ERROR_CONFIRM_INBOX,
+        -> PurchasesErrorCode.StoreProblemError
+        GalaxyErrorCode.IAP_ERROR_NOT_AVAILABLE_SHOP -> PurchasesErrorCode.PurchaseNotAllowedError
+        GalaxyErrorCode.IAP_ERROR_INVALID_ACCESS_TOKEN -> PurchasesErrorCode.InvalidCredentialsError
+        GalaxyErrorCode.IAP_ERROR_NONE -> return null // This means that the ErrorVo isn't an error
```
fire-at-will: It didn't feel right to return a PurchasesError with an unknown code or something like that here, since this galaxy error code explicitly means that an error didn't occur.

This code really shouldn't be executed, since most calling code will be checking our `ErrorVo.isError()` function, but I still added this here to exercise defensive programming.

purchases/src/test/java/com/revenuecat/purchases/galaxy/utils/ErrorVoUtilsTest.kt
```diff
@@ -34,8 +35,63 @@ class ErrorVoUtilsTest {
...
+            GalaxyErrorCode.IAP_ERROR_WHILE_RUNNING to PurchasesErrorCode.PurchaseInvalidError,
+            GalaxyErrorCode.IAP_ERROR_NEED_APP_UPGRADE to PurchasesErrorCode.StoreProblemError,
+            GalaxyErrorCode.IAP_ERROR_INITIALIZATION to PurchasesErrorCode.StoreProblemError,
+            GalaxyErrorCode.IAP_ERROR_COMMON to PurchasesErrorCode.StoreProblemError,
+            GalaxyErrorCode.IAP_ERROR_CONFIRM_INBOX to PurchasesErrorCode.StoreProblemError,
+            GalaxyErrorCode.IAP_ERROR_NOT_AVAILABLE_SHOP to PurchasesErrorCode.PurchaseNotAllowedError,
+            GalaxyErrorCode.IAP_ERROR_INVALID_ACCESS_TOKEN to PurchasesErrorCode.InvalidCredentialsError,
+        )
+
+        expectedMappings.forEach { (galaxyCode, purchasesErrorCode) ->
```
tonidero: Instead of iterating over the hardcoding mappings here, should we iterate over all values of the `GalaxyErrorcode` enum, so we make sure we catch all of them? Then, we can convert `expectedMappings` to a map, for easier access to the expected PurchasesErrorCode.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/ErrorVoUtils.kt
tonidero: Hmm I guess the main problem is that this makes the return nullable, which we need to handle everywhere... Maybe it's worth returning the unknown message here but just log something before that?

fire-at-will: Yeah, it definitely muddies up the call sites. Updated to return an UnknownError in 132448cfc46d8403131112697ea78ee233d24d08 👍

purchases/src/test/java/com/revenuecat/purchases/galaxy/utils/ErrorVoUtilsTest.kt
fire-at-will: I love that idea! That'll make sure that we remember to add future mappings to this test :)

Made the change in 09c48d51955ee72872e05cf6b7f1a678ed297817

PR #2928, opened 2025-12-13
Feat(Paywalls) Support Underlined Text

JZDesign:
### Checklist
- [x] If applicable, create follow-up issues for `purchases-ios` and hybrids https://github.com/RevenueCat/purchases-ios/pull/5935

### Motivation

We've had customers asking for this for months. And @bab-s would be happy.

PF-62

### Description

Screenshot 2025-12-12 at 8 37 23 PM

---

> [!NOTE]
> **Low Risk**
> Scoped to RevenueCat UI markdown text styling with defensive handling for malformed tags; no billing, auth, or network behavior changes.
> 
> **Overview**
> Adds **underline** rendering for paywall markdown that uses inline HTML `<u>...</u>`, which standard CommonMark does not style on its own.
> 
> The `Markdown` composable now handles `HtmlInline` nodes and strips/processes `<u>` tags via shared `MarkdownState`, applying `TextDecoration.Underline` only after a matching close tag so ranges can cross bold/italic/code nodes (e.g. `<u>**bold**</u>`) and unclosed tags do not underline trailing text. Plain `Text` and `Code` literals go through `appendTextWithUnderlines` instead of raw `append`.
> 
> Adds a Compose preview sample and a new `MarkdownTests` suite covering tag parsing, state depth, and combined formatting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd40b9a2fb8d612cc95b379da5c600022381b1d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -58,6 +61,21 @@ private val parser = Parser.builder()
...
+ */
+private class MarkdownState {
+    var underlineDepth = 0
+}
+
+object MarkdownTagDefinitions {
+    const val UNDERLINE_OPEN_TAG = "<u>"
+    const val UNDERLINE_CLOSE_TAG = "</u>"
+    const val UNDERLINE_OPEN_TAG_LENGTH = 3
+    const val UNDERLINE_CLOSE_TAG_LENGTH = 4
```
tonidero: Seems these 2 could be calculated from the tags themselves instead of hardcoding them?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -582,3 +659,15 @@ private fun MarkdownText(
             },
     )
 }
+
+@Preview
+@Composable
+@Suppress("MaxLineLength")
+fun PreviewText() {
```
tonidero: This needs to be private, we don't want to expose it as part of the public API :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -58,6 +61,21 @@ private val parser = Parser.builder()
...
 
+/**
+ * Tracks state during markdown AST traversal, specifically for handling
+ * underline tags that span across multiple AST nodes.
+ */
+private class MarkdownState {
+    var underlineDepth = 0
+}
+
+object MarkdownTagDefinitions {
```
tonidero: This also should be private

tonidero (approved):
I believe this makes sense. Thanks for adding support for this!! 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -539,22 +558,80 @@ private fun AnnotatedString.Builder.appendMarkdownChildren(
...
+                    MarkdownTagDefinitions.UNDERLINE_OPEN_TAG -> {
+                        pushStyle(SpanStyle(textDecoration = TextDecoration.Underline))
+                        state.underlineDepth++
+                    }
+                    MarkdownTagDefinitions.UNDERLINE_CLOSE_TAG -> {
+                        if (state.underlineDepth > 0) {
+                            pop()
+                            state.underlineDepth--
+                        }
+                    }
```
JayShortway: Could you explain why we need the `state`? I read it's for tags spanning multiple nodes, but it's not clicking for me haha. My naive thought is that we can just push when we see `<u>` and pop when we see `</u>`, but maybe I'm missing something.

JZDesign: If the paywall—for some odd reason—has a `</u>` tag that doesn't belong we could pop a style off and remove things we shouldn't

An edge case, to be sure.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -550,22 +567,82 @@ private fun AnnotatedString.Builder.appendMarkdownChildren(
...
+    when (tag) {
+        MarkdownTagDefinitions.UNDERLINE_OPEN_TAG -> {
+            pushStyle(SpanStyle(textDecoration = TextDecoration.Underline))
+            state.underlineDepth++
+        }
+        MarkdownTagDefinitions.UNDERLINE_CLOSE_TAG -> {
+            if (state.underlineDepth > 0) {
+                pop()
+                state.underlineDepth--
+            }
```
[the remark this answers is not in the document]
JayShortway: Might be good to fix this one.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
```diff
@@ -517,15 +533,16 @@ private fun AnnotatedString.Builder.appendMarkdownChildren(
...
     baseFontWeight: FontWeight?,
+    state: MarkdownState = MarkdownState(),
 ) {
     var child = parent.firstChild
     while (child != null) {
         when (child) {
-            is Paragraph -> appendMarkdownChildren(child, color, allowLinks, baseFontWeight)
-            is Text -> append(child.literal)
+            is Paragraph -> appendMarkdownChildren(child, color, allowLinks, baseFontWeight, state)
+            is Text -> appendTextWithUnderlines(child.literal, state)
```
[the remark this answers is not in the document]
JayShortway: Not sure if this is a problem? Worth verifying consistency with the other platforms at least.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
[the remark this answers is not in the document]
JayShortway: Not sure how though, `pop()` is the only way to pop styles afaik. No way to specify which style you want to pop, it's always the latest.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/composables/Markdown.kt
JayShortway: Right, probably good to be defensive indeed.

JayShortway (approved):

PR #2931, opened 2025-12-15
Adds `setAppsFlyerConversionData` to conveniently track AppsFlyer conversion data

JayShortway:
### Motivation
[Mini-spec](https://docs.google.com/document/d/1bUztyfG_mVlHsi0KfuJaa4Cgj7Ok14SJW4CiCGivrOg) 

### Description
Introduces a new public function `fun setAppsFlyerConversionData(data: Map<*, *>?)` to `Purchases` that processes AppsFlyer conversion data and sets corresponding RevenueCat subscriber attributes.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/MapExtensions.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/MapExtensions.kt
@@ -15,10 +15,10 @@ internal fun <K, V, R : Any> Map<K, V>.mapNotNullKeys(transform: (Map.Entry<K, V
     return destination
 }
 
-internal fun Map<*, *>.getStringValue(key: String): String? {
-    return when (val value = this[key]) {
-        is String -> value.takeIf { it.isNotBlank() }
-        null -> null
-        else -> value.toString().takeIf { it.isNotBlank() }
+internal fun Map<*, *>.getStringValueForPrimitive(key: String): String? {
+    val value = this[key] ?: return null
+    return when (value) {
+        is String, is Number, is Boolean, is Char -> value.toString().takeIf { it.isNotBlank() }
+        else -> null
     }
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/subscriberattributes/SubscriberAttributesManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/subscriberattributes/SubscriberAttributesManager.kt
@@ -11,7 +11,7 @@ import com.revenuecat.purchases.strings.AttributionStrings
 import com.revenuecat.purchases.subscriberattributes.caching.AppUserID
 import com.revenuecat.purchases.subscriberattributes.caching.SubscriberAttributeMap
 import com.revenuecat.purchases.subscriberattributes.caching.SubscriberAttributesCache
-import com.revenuecat.purchases.utils.getStringValue
+import com.revenuecat.purchases.utils.getStringValueForPrimitive
 import java.util.Observable
 
 @Suppress("TooManyFunctions")
@@ -170,33 +170,33 @@ internal class SubscriberAttributesManager(
 
         val attributes = mutableMapOf<String, String?>()
 
-        val mediaSource = data.getStringValue("media_source")
+        val mediaSource = data.getStringValueForPrimitive("media_source")
         if (mediaSource != null) {
             attributes[SubscriberAttributeKey.CampaignParameters.MediaSource.backendKey] = mediaSource
-        } else if (data.getStringValue("af_status")?.equals("Organic", ignoreCase = true) == true) {
+        } else if (data.getStringValueForPrimitive("af_status")?.equals("Organic", ignoreCase = true) == true) {
             attributes[SubscriberAttributeKey.CampaignParameters.MediaSource.backendKey] = "Organic"
         }
 
-        data.getStringValue("campaign")?.also {
+        data.getStringValueForPrimitive("campaign")?.also {
             attributes[SubscriberAttributeKey.CampaignParameters.Campaign.backendKey] = it
         }
 
-        val adGroup = data.getStringValue("adgroup") ?: data.getStringValue("adset")
+        val adGroup = data.getStringValueForPrimitive("adgroup") ?: data.getStringValueForPrimitive("adset")
         adGroup?.also {
             attributes[SubscriberAttributeKey.CampaignParameters.AdGroup.backendKey] = it
         }
 
-        val ad = data.getStringValue("af_ad") ?: data.getStringValue("ad_id")
+        val ad = data.getStringValueForPrimitive("af_ad") ?: data.getStringValueForPrimitive("ad_id")
         ad?.also {
             attributes[SubscriberAttributeKey.CampaignParameters.Ad.backendKey] = it
         }
 
-        val keyword = data.getStringValue("af_keywords") ?: data.getStringValue("keyword")
+        val keyword = data.getStringValueForPrimitive("af_keywords") ?: data.getStringValueForPrimitive("keyword")
         keyword?.also {
             attributes[SubscriberAttributeKey.CampaignParameters.Keyword.backendKey] = it
         }
 
-        val creative = data.getStringValue("creative") ?: data.getStringValue("af_creative")
+        val creative = data.getStringValueForPrimitive("creative") ?: data.getStringValueForPrimitive("af_creative")
         creative?.also {
             attributes[SubscriberAttributeKey.CampaignParameters.Creative.backendKey] = it
         }


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -816,6 +816,26 @@ class Purchases internal constructor(
...
+     * - `$mediaSource`: From `media_source`, or "Organic" if `af_status` is "Organic"
+     * - `$campaign`: From `campaign`
+     * - `$adGroup`: From `adgroup`, with fallback to `adset`
+     * - `$ad`: From `af_ad`, with fallback to `ad_id`
+     * - `$keyword`: From `af_keywords`, with fallback to `keyword`
+     * - `$creative`: From `creative`, with fallback to `af_creative`
+     *
+     * @param data The conversion data map from AppsFlyer's `onConversionDataSuccess` callback.
+     */
+    fun setAppsFlyerAttributionData(data: Map<*, *>?) {
```
tonidero: Maybe it's simpler this way... but I was wondering whether it would be worth adding a `AppslyerAttributionData` class that can be constructed with this map, so we perform some validations there and it's more flexible in case we want to change the signature later... 

But honestly, don't think it's a blocker. Probably shouldn't throw in case it doesn't pass validations, and we can adapt the signature later if needed, so FFTI!!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1044,6 +1044,61 @@ internal class PurchasesOrchestrator(
...
+        if (data == null) {
+            return
+        }
+
+        val attributes = mutableMapOf<String, String?>()
+
+        val mediaSource = data.getStringValue("media_source")
+            ?: data.getStringValue("af_status")
+                .takeIf { it.equals("Organic", ignoreCase = true) }
+                ?.let { "Organic" }
```
tonidero: Hmm do we need to get `af_status` at all considering we only get it it if's `Organic` if we are falling back to `Organic` in any case?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1044,6 +1044,61 @@ internal class PurchasesOrchestrator(
         )
     }
 
+    @Suppress("CyclomaticComplexMethod")
+    fun setAppsFlyerAttributionData(data: Map<*, *>?) {
```
tonidero: Nitpicky, but I would try to move this logic out, maybe to the SubscriberAttributesManger?... I think it would be better to minimize the logic here for testability purposes :)

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1044,6 +1044,61 @@ internal class PurchasesOrchestrator(
...
+        creative?.also {
+            attributes[SubscriberAttributeKey.CampaignParameters.Creative.backendKey] = it
+        }
+
+        if (attributes.isNotEmpty()) {
+            subscriberAttributesManager.setAttributes(attributes, appUserID)
+        }
+    }
+
+    private fun Map<*, *>.getStringValue(key: String): String? {
```
tonidero: Hmm might be worth extracting to `utils.kt` maybe?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1044,6 +1044,61 @@ internal class PurchasesOrchestrator(
         )
     }
 
+    @Suppress("CyclomaticComplexMethod")
+    fun setAppsFlyerAttributionData(data: Map<*, *>?) {
+        log(LogIntent.DEBUG) { AttributionStrings.METHOD_CALLED.format("setAppsFlyerAttributionData") }
+
+        if (data == null) {
+            return
```
rickvdl: Same question here as on iOS, is it ok that we're not support unsetting attributes?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
JayShortway: We're only falling back to `Organic` if `af_status` is `Organic`. That said, I do see how this can be unclear, so I made it a bit more obvious in https://github.com/RevenueCat/purchases-android/pull/2931/changes/5a111dd287519a4cdde1e6606968b43a5e80129b.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
JayShortway: That's a good idea! Done here: https://github.com/RevenueCat/purchases-android/pull/2931/commits/65dec7e7c05442d1535bdea36d9d60aaf2ca2994.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
JayShortway: Moved to `MapExtensions.kt` (https://github.com/RevenueCat/purchases-android/pull/2931/commits/76a3ee31a9f81d5a454a4a37bec20daa31be3c1b).

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
JayShortway: Right, I'm very much pro wrapper classes in the public API to allow us to evolve it in a non-breaking way. However in this case I'm not sure if it's worth it. It's really meant to be called like this:
```kotlin
override fun onConversionDataSuccess(data: MutableMap<String, Any>?) {
  Purchases.sharedInstance.setAppsFlyerAttributionData(data)
}
```

If AppsFlyer changes the `data` type, we would add a new function and deprecate the old one. If we had a wrapper class, we would add a new constructor and deprecate the old one. That's pretty much the same, with a tiny bit of added friction because developers need to understand that they need to construct the wrapper class first.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
JayShortway: I think that's okay. Conversion attribution is determined once for an install, and never changes. It might also be confusing. For instance, if you've set campaign data manually, and then call this function with a map that doesn't contain campaign data, it would be cleared. That might be unexpected.

I thought about supporting unsetting all attributes at once if the developer passes a `nil` dictionary. However I also think that would be confusing, if sometimes `nil` unsets and sometimes it doesn't. I chose to clarify this behavior in the docs instead. (https://github.com/RevenueCat/purchases-android/pull/2931/commits/ae39ec764b0e0f4946a146213fdfda02a85695ec).

rickvdl: Thanks for clarifying, agree!

purchases/src/main/kotlin/com/revenuecat/purchases/utils/MapExtensions.kt
```diff
@@ -14,3 +14,11 @@ internal fun <K, V, R : Any> Map<K, V>.mapNotNullKeys(transform: (Map.Entry<K, V
...
     return destination
 }
+
+internal fun Map<*, *>.getStringValue(key: String): String? {
+    return when (val value = this[key]) {
+        is String -> value.takeIf { it.isNotBlank() }
+        null -> null
+        else -> value.toString().takeIf { it.isNotBlank() }
+    }
+}
```
ajpallares: Nit: Couldn't this be simplified to

```
internal fun Map<*, *>.getStringValue(key: String): String? =
    this[key]?.toString()?.takeIf { it.isNotBlank() }
```

.gitignore
```diff
@@ -137,5 +137,5 @@ purchases-android-snapshots
 # Make sure we share the same Detekt IntelliJ plugin settings
 !.idea/detekt.xml
 
-# Kotlin error logs
-.kotlin/errors/
+# Kotlin compiler
+.kotlin
```
ajpallares: This change seems unrelated to this PR. I think the change makes sense, but idk if maybe should be done in a separate PR? (for easier tracking mostly)

JayShortway: Good point, done! Changed the base branch of this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/MapExtensions.kt
JayShortway: Ah yes, other than calling `toString` on a `String`, those should be equivalent indeed! Done here: https://github.com/RevenueCat/purchases-android/pull/2931/commits/23d7411a04f5416b6c51ae2739b906714747cee8.

PR #2938, opened 2025-12-15
Add exit offers support to PaywallActivity and PaywallDialog

vegaro:
Adds exit offers support to PaywallActivity

The idea is to show another activity on top with the exit offer paywall when the activtiy is dismissed

https://github.com/user-attachments/assets/9c77bd06-9512-465a-9ef0-5e2dd3d9f1e0

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.paywallstester.ui.screens.main.offerings
 
+import android.util.Log
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
@@ -41,16 +42,21 @@ import androidx.compose.ui.window.DialogProperties
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.revenuecat.paywallstester.MainActivity
+import com.revenuecat.purchases.CustomerInfo
 import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.Offerings
 import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.getOfferingsWith
+import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.ui.revenuecatui.PaywallDialog
 import com.revenuecat.purchases.ui.revenuecatui.PaywallDialogOptions
+import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import com.revenuecat.purchases.Package as RCPackage
 
 @SuppressWarnings("LongParameterList")
 @Composable
@@ -237,6 +243,34 @@ private fun OfferingsListScreen(
             PaywallDialogOptions.Builder()
                 .setDismissRequest { displayPaywallDialogOffering = null }
                 .setOffering(displayPaywallDialogOffering)
+                .setListener(object : PaywallListener {
+                    override fun onPurchaseStarted(rcPackage: RCPackage) {
+                        Log.d("PaywallDialog", "onPurchaseStarted: ${rcPackage.identifier}")
+                    }
+
+                    override fun onPurchaseCompleted(
+                        customerInfo: CustomerInfo,
+                        storeTransaction: StoreTransaction,
+                    ) {
+                        Log.d("PaywallDialog", "onPurchaseCompleted: ${storeTransaction.productIds}")
+                    }
+
+                    override fun onPurchaseError(error: PurchasesError) {
+                        Log.e("PaywallDialog", "onPurchaseError: ${error.message}")
+                    }
+
+                    override fun onRestoreStarted() {
+                        Log.d("PaywallDialog", "onRestoreStarted")
+                    }
+
+                    override fun onRestoreCompleted(customerInfo: CustomerInfo) {
+                        Log.d("PaywallDialog", "onRestoreCompleted: ${customerInfo.activeSubscriptions}")
+                    }
+
+                    override fun onRestoreError(error: PurchasesError) {
+                        Log.e("PaywallDialog", "onRestoreError: ${error.message}")
+                    }
+                })
                 .build(),
         )
     }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.ui.revenuecatui
 
 import android.os.Build
+import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.asPaddingValues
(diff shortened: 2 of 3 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -41,10 +51,20 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.saveSdkConfiguration
 internal class PaywallActivity : ComponentActivity(), PaywallListener {
     companion object {
         const val ARGS_EXTRA = "paywall_args"
-
         const val RESULT_EXTRA = "paywall_result"
     }
 
+    private var purchaseCompleted = false
+    private var currentOfferingSelection by mutableStateOf<OfferingSelection?>(null)
+    private var preloadedExitOfferId: String? = null
```
JZDesign: These wouldn't survive a device rotation right? I imagine that could cause issues

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ExitOffersConfiguration.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import dev.drewhamilton.poko.Poko
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@InternalRevenueCatAPI
+@Poko
+@Serializable
+class ExitOffersConfiguration(
```
facumenzella: I am using ExitOffers & ExitOffer which I am not sure I like. Not a fan of `Configuration` either hehe 💅

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -106,19 +124,76 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
...
+        exitOfferLauncher.launch(exitOfferArgs)
+    }
+
     override fun onSaveInstanceState(outState: Bundle) {
         saveSdkConfiguration(outState)
         super.onSaveInstanceState(outState)
     }
 
     override fun onPurchaseCompleted(customerInfo: CustomerInfo, storeTransaction: StoreTransaction) {
+        purchaseCompleted = true
```
facumenzella: I ended up using `PurchaseResultData` instead, in case it serves a greater purpose

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -41,10 +51,20 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.saveSdkConfiguration
...
     }
 
+    private var purchaseCompleted = false
+    private var currentOfferingSelection by mutableStateOf<OfferingSelection?>(null)
+    private var preloadedExitOfferId: String? = null
+
+    private val exitOfferLauncher: ActivityResultLauncher<PaywallActivityArgs> =
+        registerForActivityResult(PaywallContract()) { result ->
+            // Forward the exit offer's result and finish this activity
+            setResult(RESULT_OK, createResultIntent(result))
```
tonidero, line 56: Hmm this does remind me... Is the `PaywallResult` returned from the activity going to be the one from the first activity? Or would it be the one for the exit offer activity? I think ideally it should be the one for the exit offer activity.

Maybe we would need to piggyback the result of the exit offer activity while keeping the original paywall activity open?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
tonidero: Right, I think some of this data would just get calculated again after a reconfiguration, except `purchaseCompleted`... But it does feel very brittle to keep state like this. I think we should fix these states... I wonder if we could move these to compose world, so they could potentially be reused by other ways to access the paywall, and just expose some "internal" listener for now that the activity can use to decide to display the next offering Id.

tonidero: Also, maybe 
```suggestion
    private var purchaseCompleted = false
    private var currentOfferingSelection by mutableStateOf<OfferingSelection?>(null)
    private var preloadedExitOfferingId: String? = null
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -106,19 +124,76 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
...
                             },
                     ) {
+                        // Preload exit offer ID so it's ready when dismiss is called
+                        val selection = currentOfferingSelection
+                        val offeringId = selection?.offeringIdentifier
+                        LaunchedEffect(offeringId) {
+                            try {
+                                val offerings = Purchases.sharedInstance.awaitOfferings()
+                                val currentOffering = if (offeringId != null) {
+                                    offerings[offeringId]
```
tonidero: Should we log something if we can't find the offeringId for the exit offer for any reason?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
vegaro: Good catch! I've refactored this a bit in a9515b9754e94379112cc91c1055f5146165db4a so use the view model we use in Compose. This will come handy for the `PaywallDialog` actually. Let me know what you think

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
vegaro: with the current code the intention was to return the one for the exit offer activity. I believe that's what this is doing?

tonidero: 🤦 Ok you're right. It was early in the morning ☕ . This is good, please ignore me!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
vegaro: oops I pushed to the wrong branch lol!

vegaro: https://github.com/RevenueCat/purchases-android/pull/2938/commits/4b553aea1644f08a1a162ebfb297b77b588b2c43 has the refactor

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -41,10 +49,17 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.saveSdkConfiguration
 internal class PaywallActivity : ComponentActivity(), PaywallListener {
     companion object {
         const val ARGS_EXTRA = "paywall_args"
-
         const val RESULT_EXTRA = "paywall_result"
     }
 
+    private var currentOfferingSelection by mutableStateOf<OfferingSelection?>(null)
```
facumenzella: Do we need this to be `mutableStateOf`. It's being set, and then used right after

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -106,13 +119,54 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
                                 padding(paddingValues)
                             },
                     ) {
+                        val selection = currentOfferingSelection
+
+                        val paywallOptions = PaywallOptions.Builder(dismissRequest = {})
```
facumenzella: Is this being set `{}` intentionally? 🤔 Maybe a comment for posterity would help

vegaro: it's overriden further down by `setDismissRequestWithExitOffering`, but it's still a required parameter... Just added a comment

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
vegaro: No, we don't need mutableStateOf! Since currentOfferingSelection is only set once and never changes, it can just be a regular property. Let me fix it.

PR #2939, opened 2025-12-15
Acknowledge Galaxy Subscription Purchases

fire-at-will:
### Description
Adds support for acknowledging Galaxy subscription purchases.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
@@ -14,7 +14,6 @@ import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.ReplaceProductInfo
 import com.revenuecat.purchases.common.StoreProductsCallback
 import com.revenuecat.purchases.common.caching.DeviceCache
-import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.common.warnLog
 import com.revenuecat.purchases.galaxy.constants.GalaxyConsumeOrAcknowledgeStatusCode
@@ -196,18 +195,29 @@ internal class GalaxyBillingWrapper(
     ) {
         val galaxyPurchaseInfo = purchasingData as? GalaxyPurchasingData.Product
         if (galaxyPurchaseInfo == null) {
+            val errorMessage = PurchaseStrings.INVALID_PURCHASE_TYPE.format(
+                "Galaxy",
+                "GalaxyPurchasingData",
+            )
             val error = PurchasesError(
                 PurchasesErrorCode.UnknownError,
-                PurchaseStrings.INVALID_PURCHASE_TYPE.format(
-                    "Galaxy",
-                    "GalaxyPurchasingData",
-                ),
+                errorMessage,
             )
-            errorLog(error)
+            log(LogIntent.GALAXY_ERROR) { errorMessage }
             purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
             return
         }
         val storeProduct = galaxyPurchaseInfo.storeProduct
+
+        if (storeProduct.type == ProductType.INAPP) {
+            val error = PurchasesError(
+                PurchasesErrorCode.UnsupportedError,
+                GalaxyStrings.GALAXY_OTPS_NOT_SUPPORTED
+            )
+            log(LogIntent.GALAXY_ERROR) { GalaxyStrings.GALAXY_OTPS_NOT_SUPPORTED }
+            purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
+            return
+        }
 
         if (replaceProductInfo != null) {
             log(LogIntent.GALAXY_WARNING) { GalaxyStrings.PRODUCT_CHANGES_NOT_SUPPORTED }
--- a/purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
@@ -319,6 +319,30 @@ class GalaxyBillingWrapperTest : GalaxyStoreTest() {
         verify(exactly = 1) { purchasesUpdatedListener.onPurchasesFailedToUpdate(capture(errorSlot)) }
         assertThat(errorSlot.captured).isEqualTo(expectedError)
     }
+    
+    fun `makePurchaseAsync errors when purchasing OTP`() {
+        val purchasesUpdatedListener = mockk<BillingAbstract.PurchasesUpdatedListener>(relaxed = true)
+        val wrapper = createWrapper()
+        wrapper.purchasesUpdatedListener = purchasesUpdatedListener
+
+        val storeProduct = createStoreProduct(type = ProductType.INAPP)
+
+        wrapper.makePurchaseAsync(
+            activity = mockk<Activity>(),
+            appUserID = "user",
+            purchasingData = GalaxyPurchasingData.Product(storeProduct),
+            replaceProductInfo = null,
+            presentedOfferingContext = null,
+            isPersonalizedPrice = null,
+        )
+
+        val errorSlot = slot<PurchasesError>()
+        verify(exactly = 1) { purchasesUpdatedListener.onPurchasesFailedToUpdate(capture(errorSlot)) }
+        assertThat(errorSlot.captured.code).isEqualTo(PurchasesErrorCode.UnsupportedError)
+        assertThat(errorSlot.captured.underlyingErrorMessage).isEqualTo(GalaxyStrings.GALAXY_OTPS_NOT_SUPPORTED)
+
+        verify(exactly = 0) { purchaseHandlerMock.purchase(any(), any(), any(), any()) }
+    }
 
     @OptIn(GalaxySerialOperation::class)
     @Test
@@ -427,10 +451,11 @@ class GalaxyBillingWrapperTest : GalaxyStoreTest() {
 
     private fun createStoreProduct(
         id: String = "productId",
+        type: ProductType = ProductType.SUBS,
         presentedOfferingContext: PresentedOfferingContext? = null,
     ): GalaxyStoreProduct = GalaxyStoreProduct(
         id = id,
-        type = ProductType.SUBS,
+        type = type,
         price = Price(
             formatted = "$1.00",
             amountMicros = 1_000_000,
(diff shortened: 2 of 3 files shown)


purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyStoreTest.kt
```diff
@@ -20,43 +21,35 @@ open class GalaxyStoreTest {
...
-                "mCurrencyCode": "$currencyCode",
-                "mItemDesc": "$itemDescription",
-                "mType": "$type",
-                "mSubscriptionDurationMultiplier": "$subscriptionDurationMultiplier",
-                "mSubscriptionDurationUnit": "$subscriptionDurationUnit"
-            }
-        """.trimIndent()
-
-        return ProductVo(json)
+        return mockk<ProductVo>(relaxed = true).also { productVo ->
```
fire-at-will, line 24: I started having some issues instantiating `ProductVo` and `PurchaseVo` directly, so I decided to mock them out instead

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/AcknowledgePurchaseHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+            )
+            onError(error)
+            return
+        }
+
+        if (!iapHelperProvider.isAcknowledgeAvailable(context = context)) {
+            log(LogIntent.GALAXY_WARNING) {
+                GalaxyStrings.WARNING_ACKNOWLEDGING_PURCHASES_UNAVAILABLE
+            }
+            return
```
tonidero, line 61: This won't call neither the success nor error callbacks, basically leaving it stuck. I think we should either call the onError, or still perform the request and just process the error later.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/AcknowledgePurchaseHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+            }
+            val purchasesError = PurchasesError(
+                PurchasesErrorCode.StoreProblemError,
+                GalaxyStrings.ACKNOWLEDGE_REQUEST_RETURNED_SUCCESS_BUT_NO_ACKNOWLEDGEMENT_RESULTS,
+            )
+            val onError = inFlightRequest?.onError
+            clearInFlightRequest()
+            onError?.invoke(purchasesError)
+            return
+        } else if (acknowledgementResults.count() > 1) {
```
tonidero: Hmm so I guess you can acknowledge multiple purchases simultaneously? Interesting... but yeah I think it's fine to do one per request as we're doing now, so I imagine this should never happen indeed 👍

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -112,7 +128,101 @@ internal class GalaxyBillingWrapper(
         shouldConsume: Boolean,
         initiationSource: PostReceiptInitiationSource,
     ) {
-        warnLog { "Unimplemented: GalaxyBillingWrapper.consumeAndSave" }
+        if (!finishTransactions || purchase.type == ProductType.UNKNOWN) return
```
tonidero: In these cases, I think we should still call `addSuccessfullyPostedToken` so we're not posting the same tokens over and over? As long as it's not in a pending state though.

purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
```diff
@@ -318,6 +320,141 @@ class GalaxyBillingWrapperTest : GalaxyStoreTest() {
         assertThat(errorSlot.captured).isEqualTo(expectedError)
     }
 
+    @Test
+    fun `consumeAndSave does nothing when finishTransactions is false`() {
+        val wrapper = spyk(createWrapper())
```
tonidero: Hmm not a super fan of creating a spy just to verify that a method of the class we're testing is not called... I think the proper thing to do would be to verify whether the expected side effects (in this case, calling the IAPHelper consume/acknowledge methods are not called)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/AcknowledgePurchaseHandler.kt
fire-at-will: Yeah, you can acknowledge/consume multiple purchases at once. In fact, their docs even strongly recommend that you don't acknowledge them serially:

> We recommend reporting purchased consumable items immediately after verifying their purchase and reporting all unreported products in one consumePurchasedItems() or acknowledgePurchases() method call in order to avoid system overload or malfuction.

However, our system is only architected today to handle acknowleding/consuming one purchase at a time - see the `BillingAbstract.consumeAndSave` function's signature:

```kotlin
abstract fun consumeAndSave(
        finishTransactions: Boolean,
        purchase: StoreTransaction,
        shouldConsume: Boolean,
    )
```

I think that this is okay for the purchase scenario - the Galaxy Store doesn't have multi-line subscriptions, so only one purchase should go through at a time. The only case I can think of where we might have multiple to do at once is in a restore scenario, or maybe when multiple distinct POST /receipt requests have failed, and the SDK needs to re-POST receipts. Even in that scenario though, we'll be calling this endpoint serially, so I _think_ it should be fine, what do you think?

tonidero: Yeah, I think it should be fine... probably worth seeing how it behaves in the wild though, hopefully it's not rate limited on their side 😬

fire-at-will: cc @MarkVillacampa this might be a good thing to ask the Samsung team

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/AcknowledgePurchaseHandler.kt
fire-at-will: GREAT catch. I've made it call the onError callback and have added a test to ensure this behavior in 2a37f1de74752b972c3c91e5ccedb1f58c3717f1

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: It looks like we aren't calling `addSuccessfullyPostedToken` for UNKNOWN product types on the the Amazon or Play Store today:

**Google:**
https://github.com/RevenueCat/purchases-android/blob/d67337ace879f5e40753e6c3308a0bbb80369ee1/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt#L469

**Amazon:**
https://github.com/RevenueCat/purchases-android/blob/d67337ace879f5e40753e6c3308a0bbb80369ee1/purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt#L227

Where those two stores differ is whether they call `addSuccessfullyPostedToken` in observer mode 😅 
- We don't for Amazon
- We do for the Play Store

tonidero: I think Amazon is wrong in this case... If we don't add it, we will be posting the token on every app open I think?

purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
fire-at-will: Removed the spies from the tests here! 🚫🕵️‍♂️

d45f2060080a43eab59f7637631778d9a5921f61

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Yep, I believe you're correct 😅 I'll add the token for the Galaxy Store for both UNKNOWN product types and when observer mode is enabled

fire-at-will: Updated in 7dbaab997a9c8b8cde502f370d69d72cbaf0b065 to store the token when the product type is Unknown or when observer mode is enabled 👍

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -37,25 +41,30 @@ import com.samsung.android.sdk.iap.lib.vo.PurchaseVo
 internal class GalaxyBillingWrapper(
     stateProvider: PurchasesStateProvider,
     private val context: Context,
-    private val finishTransactions: Boolean,
+    private val deviceCache: DeviceCache,
     val billingMode: GalaxyBillingMode,
-    private val iapHelperProvider: IAPHelperProvider = DefaultIAPHelperProvider(
+    private val iapHelper: IAPHelperProvider = DefaultIAPHelperProvider(
```
tonidero, line 45: I'm wondering if we should call it `IAPHelperWrapper` instead of `Provider`... but that should probably be in a separate PR.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -94,7 +94,7 @@ internal class ProductDataHandler(
         }
     }
 
-    override fun onGetProducts(error: ErrorVo, products: ArrayList<ProductVo>) {
+    override fun onGetProducts(error: ErrorVo, products: ArrayList<ProductVo?>) {
```
tonidero, line 97: It's quite interesting this can return a nullable product list... Any idea in what scenarios this would return a list with null elements?

tonidero (approved):
This looks good to me!

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: I cant' think of any scenarios where this list should contain null elements, but because of the way that the function's signature is written in Java, it's a scenario that is allowed by the API. The signature in Java looks like this:

```java
public interface OnGetProductsDetailsListener {
    void onGetProducts(@NotNull ErrorVo var1, @NotNull ArrayList<ProductVo> var2);
}
```

`var2` is marked as `@NotNull`, which means that `var2` itself is not null, but it doesn't guarantee that the objects inside the list are non-null. In fact, when we use Android Studio's auto complete to write the function, it includes the `?` on the `ProductVo`. If we didn't mark `ProductVo` as optional in this signature, I think that we would get a crash when the code bridges from Java to Kotlin if there are any null elements in the list 😬

I realized this while working on the acknowledgement listener, which also receives an ArrayList of nullable objects, so I went back and fixed it here before I forgot about it :)

tonidero: Oh right 😓 Makes sense. thanks for explaining!!

PR #2940, opened 2025-12-16
Make close() method also clear the shared instance of the SDK

tonidero:
### Description
When calling the `close` function, we cleared a bunch of internal state that leaves the SDK in an inconsistent state. Specifically, we clear the billing client from the BillingWrapper, among other things. If after that, you tried to reconfigure the SDK with the exact same configuration, it wouldn't let you, since we have logic to prevent reconfigurations with the same config. This means that in some cases, we would end up in this inconsistent state, without a great way to get out of it.

I think the `close` function means the instance shouldn't be called more, so we should also clear the shared instance, so we can actually reinitialize later if needed and `Purchases.isConfigured` is false. This is however a behavior change.


tonidero:
> I think the behavior change is fine because it wouldn't work anyways right?

Indeed, I think it should be fine... I guess before this, you could continue doing some operations with the SDK if you had some cached values, but it's indeed in an inconsistent state, so I think normally the changes of this PR include how it should behave

vegaro (approved):

PR #2943, opened 2025-12-16
Return Error When Purchasing Galaxy Store OTP

fire-at-will:
### Description
The SDK won't support purchasing and handling OTPs on the Galaxy Store for now. This PR updates the purchase flow so that all OTP purchases will gracefully fail.


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -138,19 +138,30 @@ internal class GalaxyBillingWrapper(
...
             val error = PurchasesError(
                 PurchasesErrorCode.UnknownError,
-                PurchaseStrings.INVALID_PURCHASE_TYPE.format(
-                    "Galaxy",
-                    "GalaxyPurchasingData",
-                ),
+                errorMessage,
             )
-            errorLog(error)
+            log(LogIntent.GALAXY_ERROR) { errorMessage }
```
fire-at-will, line 148: I realized that `errorLog` can use the `GOOGLE_ERROR` log intent, so this sets it to use the correct log intent.

PR #2947, opened 2025-12-17
Galaxy Store Free Trial Eligibility Detection

fire-at-will:
### Description
This PR adds support for including free trials for Galaxy Store products in the `pricingPhases` property of the `GalaxySubscriptionOption`. It does so by creating one (and only one) SubscriptionOption for each Galaxy subscription product, and, when the user is eligible for a free trial, it includes the free trial in that SubscriptionOption's `pricingPhases`.

As a part of this exercise, we also needed to modify `GalaxyPurchasingData` to keep a reference to just the product ID and product type, instead of the entire `StoreProduct` to avoid a circular reference between `GalaxyStoreProduct` and `GalaxySubscriptionOption`.

### Out of Scope
This PR intentionally does not add introductory subscription pricing phases when a user is eligible for one. Support for those will come in the future :)

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxySubscriptionOption.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxySubscriptionOption.kt
@@ -37,6 +37,19 @@ class GalaxySubscriptionOption(
     override val installmentsInfo: InstallmentsInfo?,
 ) : SubscriptionOption {
 
+    internal constructor(
+        subscriptionOption: GalaxySubscriptionOption,
+        presentedOfferingContext: PresentedOfferingContext?,
+    ) :
+        this(
+            id = subscriptionOption.id,
+            pricingPhases = subscriptionOption.pricingPhases,
+            tags = subscriptionOption.tags,
+            presentedOfferingContext = presentedOfferingContext,
+            purchasingData = subscriptionOption.purchasingData,
+            installmentsInfo = subscriptionOption.installmentsInfo,
+        )
+
     /**
      * The offering ID this `GalaxySubscriptionOption` was returned from.
      *
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.galaxy
 
 import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.ProductType
+import com.revenuecat.purchases.models.GoogleSubscriptionOption
 import com.revenuecat.purchases.models.Period
 import com.revenuecat.purchases.models.Price
 import com.revenuecat.purchases.models.StoreProduct
@@ -84,6 +85,17 @@ class GalaxyStoreProduct(
     }
 
     override fun copyWithPresentedOfferingContext(presentedOfferingContext: PresentedOfferingContext?): StoreProduct {
+        val subscriptionOptionsWithContext = subscriptionOptions?.mapNotNull {
+            (it as? GalaxySubscriptionOption)?.let { galaxyOption ->
+                GalaxySubscriptionOption(
+                    subscriptionOption = galaxyOption,
+                    presentedOfferingContext = presentedOfferingContext,
+                )
+            }
+        }
+
+        val defaultOptionWithContext = subscriptionOptionsWithContext?.firstOrNull()
+
         return GalaxyStoreProduct(
             id = this.id,
             type = this.type,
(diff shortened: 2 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -3,16 +3,31 @@ package com.revenuecat.purchases.galaxy.conversions
...
+import java.util.HashSet
 
-internal fun ProductVo.toStoreProduct(): StoreProduct {
-    val type = this.type.createRevenueCatProductTypeFromSamsungIAPTypeString()
-    val period: Period? = if (type == ProductType.SUBS) {
+private const val ELIGIBILITY_PRICING_FREE_TRIAL = "FreeTrial"
+private const val ELIGIBILITY_PRICING_TIERED_PRICE = "TieredPrice"
+
+internal fun ProductVo.toStoreProduct(
+    promotionEligibilities: List<PromotionEligibilityVo>? = null,
```
fire-at-will, line 25: You may be wondering why `ProductVo.storeProduct` now takes in a `List<PromotionEligibilityVo>?` instead of a single `PromotionEligibilityVo?`. This is because the `IapHelper.getPromotionEligibility()`'s listener provides us with an ArrayList of PromotionEligibilityVo? objects:

```java
public interface OnGetPromotionEligibilityListener {
    void onGetPromotionEligibility(@NotNull ErrorVo var1, @NotNull ArrayList<PromotionEligibilityVo> var2);
}
```

And thus, it's allowed by the API to receive multiple `PromotionEligibilityVo`s for the same product ID. I haven't seen this happen in my testing, but since it's technically allowed, it's a case we need to handle.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -25,20 +40,108 @@ internal fun ProductVo.toStoreProduct(): StoreProduct {
...
+): List<PricingPhase> {
+    val pricingPhases: MutableList<PricingPhase> = mutableListOf()
+    val type = this.type.createRevenueCatProductTypeFromSamsungIAPTypeString()
+
+    val eligibilityPricings: MutableSet<String> = HashSet()
+    promotionEligibilities
+        ?.filter { it.itemId == this.itemId }
+        ?.mapTo(eligibilityPricings) { it.pricing }
+
+    if (
```
fire-at-will, line 93: We still haven't seen this case appear in our testing, but since it's possible through the API, it's probably worth handling in the code :)

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -25,20 +40,108 @@ internal fun ProductVo.toStoreProduct(): StoreProduct {
...
+        PricingPhase(
+            billingPeriod = Period(
+                value = freeTrialInDays,
+                unit = Period.Unit.DAY,
+                iso8601 = "P${freeTrialInDays}D",
+            ),
+            recurrenceMode = RecurrenceMode.NON_RECURRING,
+            billingCycleCount = null,
+            price = Price(
+                formatted = "%s%.2f".format(currencyUnit, 0.0),
```
tonidero, line 137: FWIW, I think Google returns `Free` as a string in this case... But that's problematic for us since we would need to localize that string... So I'm ok just putting `0.0` for now, and we can reassess later

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxySubscriptionOption.kt
```diff
@@ -0,0 +1,51 @@
...
+     */
+    override val id: String,
+
+    /**
+     * Pricing phases defining a user's payment plan for the product over time.
+     */
+    override val pricingPhases: List<PricingPhase>,
+
+    /**
+     * Always empty.
```
tonidero: Maybe
```suggestion
     * Always empty for Galaxy products.
```

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxySubscriptionOption.kt
```diff
@@ -0,0 +1,51 @@
...
+     * The offering ID this `GalaxySubscriptionOption` was returned from.
+     *
+     * Null if not using RevenueCat offerings system, or if fetched directly via `Purchases.getProducts`
+     */
+    @Deprecated(
+        "Use presentedOfferingContext instead",
+        ReplaceWith("presentedOfferingContext.offeringIdentifier"),
+    )
+    override val presentedOfferingIdentifier: String?
+        get() = presentedOfferingContext?.offeringIdentifier
```
tonidero, line 63: Is this populated later at some point? for google, we add a `copyWithPresentedOfferingContext` method that creates a copy of subscription options with the context added. Seems we're missing that here?

fire-at-will: Great callout, it looks like this happens at the `StoreProduct` level. I'll add that logic to `GalaxyStoreProduct.copyWithPresentedOfferingContext()`

fire-at-will: Updated to do this in cd237ec155c5d179abe25e5c29456464dd241e45 👍 

I'm going to go ahead and merge this so I can start working on including intro offers :)

PR #2953, opened 2025-12-18
Load Samsung IAP aar from libs Directory

fire-at-will:
### Description
This PR switches the Samsung IAP library loading mechanism from a `local.properties` path dependency to a relative directory lookup. This ensures that the Samsung SDK is correctly resolved when building the SDK or purchase-tester.

To add the Samsung SDK to the RevenueCat SDK:
1. Clone the `samsung-dev` branch of `purchases-android`
2. Place the Samsung IAP SDK's `aar` file in the `purchases-android/libs` directory

This is likely not the final iteration of how we will include the Samsung SDK with the RC SDK, but it unblocks apps that want to try out the dev branch as it stands today.

### DIFF
--- a/feature/galaxy/build.gradle.kts
+++ b/feature/galaxy/build.gradle.kts
@@ -1,5 +1,3 @@
-import java.io.FileInputStream
-import java.util.Properties
 
 plugins {
     id("revenuecat-public-library")
@@ -26,26 +24,10 @@ android {
     }
 }
 
-// TO DO: Bring in Samsung SDK from somewhere else
-val samsungIapAar: File? =
-    sequenceOf(
-        providers.gradleProperty("samsungIapSdkPath").orNull,
-        providers.environmentVariable("SAMSUNG_IAP_SDK_PATH").orNull,
-        rootProject.file("../samsung-iap-6.5.0.aar").takeIf { it.exists() }?.path,
-    )
-        .firstOrNull { !it.isNullOrBlank() }
-        ?.let { path ->
-            val aar = file(path)
-            check(aar.exists()) {
-                "Samsung IAP SDK AAR not found at $path. Override with samsungIapSdkPath property, " +
-                    "SAMSUNG_IAP_SDK_PATH env var, or local.properties"
-            }
-            aar
-        }
-
 dependencies {
     implementation(project(":purchases"))
 
-    compileOnly(files(samsungIapAar))
+    compileOnly("com.samsung.iap:samsung-iap:6.5.0@aar")
     testImplementation(libs.bundles.test)
+    testImplementation("com.samsung.iap:samsung-iap:6.5.0@aar")
 }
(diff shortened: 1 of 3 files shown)


feature/galaxy/build.gradle.kts
```diff
@@ -27,26 +27,25 @@ android {
...
+                    "SAMSUNG_IAP_SDK_PATH env var, or local.properties"
+            }
+            aar
+        }
 
 dependencies {
     implementation(project(":purchases"))
 
-    implementation(files(samsungIapSdkPath))
+    implementation(files(samsungIapAar))
```
JayShortway: If we make this `compileOnly`, we shouldn't need to care about where the developer places their own copy of the Samsung `.aar`. 


```suggestion
    compileOnly(files(samsungIapAar))
```

tonidero: In this scenario, I think we don't really need to use the `feature/galaxy` module for anything (yet) since it's already compileOnly on the `purchases` module and this module was meant to only include that library. But yeah, I think that makes sense.

JayShortway: Oh right, yes very true.

feature/galaxy/build.gradle.kts
```diff
@@ -27,26 +27,25 @@ android {
...
+                    "SAMSUNG_IAP_SDK_PATH env var, or local.properties"
+            }
+            aar
+        }
 
 dependencies {
     implementation(project(":purchases"))
 
-    implementation(files(samsungIapSdkPath))
+    compileOnly(files(samsungIapAar))
```
vegaro: I am very surprised this works since google says compileonly doesn't work with aars

https://developer.android.com/build/dependencies#dependency_configurations

fire-at-will: Interesting...I have it working, but they very clearly state that it won't work with AARs. Trying to see if there's a workaround

fire-at-will: @vegaro in 750656425c795cfa6d34efaf29a55796fa727fa8, I've modified this to still use `compileOnly`, but now instead of loading the AAR directly in from a file, it loads it from a `flatDir` repo, which Codex says gets around compileOnly's "no AAR" requirement 😅

It also modifies Gradle to look for the Samsung SDK in the `purchases-android/libs/` folder, instead of the parent directory that contains `purchases-android`.

I've tested this out in the separate test app and can confirm that it works. Let me know what you think!

fire-at-will:
Got verbal approval from @vegaro via a Slack DM, going to go ahead and merge

feature/galaxy/build.gradle.kts
vegaro: for future reference, when using `aars` even with `implementation` the consuming application still needs to include it, since our own aar won't include our aar dependency (unless it's a fat aar)

So I think that's mainly why Google states that?

PR #2954, opened 2025-12-18
Add tiered subscriptions (intro offers) to Galaxy Store pricing phases

fire-at-will:
### Description
Adds support for including tiered subscription (intro offers) in `GalaxySubscriptionOption`'s pricing phases

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
@@ -99,7 +99,7 @@ private fun ProductVo.createPricingPhases(
     } else if (eligibilityPricings.contains(ELIGIBILITY_PRICING_FREE_TRIAL)) {
         this.createFreeTrialPricingPhase()?.let { pricingPhases.add(it) }
 
-        if(this.hasTieredSubscription()) {
+        if (this.hasTieredSubscription()) {
             // When a Galaxy product has both a trial and a tiered subscription, and the user is eligible for both,
             // IapHelper.getPromotionEligibility only returns a value stating that the user is eligible for the trial.
             // However, when the user proceeds to purchase the option, both the trial and tiered subscription are
@@ -147,8 +147,8 @@ private fun ProductVo.createFreeTrialPricingPhase(): PricingPhase? {
     } // This returns null if the string couldn't be parsed to an Int
 }
 
+@SuppressWarnings("ReturnCount")
 private fun ProductVo.createTieredSubscriptionPricingPhase(): PricingPhase? {
-
     if (!this.hasTieredSubscription()) { return null }
     val tieredPrice = this.tieredPrice.toDoubleOrNull() ?: return null
 
@@ -156,7 +156,7 @@ private fun ProductVo.createTieredSubscriptionPricingPhase(): PricingPhase? {
     val tieredSubscriptionCount = this.tieredSubscriptionCount.toIntOrNull() ?: return null
     val billingPeriod = createPeriodFromGalaxyData(
         durationMultiplier = this.tieredSubscriptionDurationMultiplier,
-        durationUnit = this.tieredSubscriptionDurationUnit
+        durationUnit = this.tieredSubscriptionDurationUnit,
     ) ?: return null
 
     return PricingPhase(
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -94,19 +94,23 @@ private fun ProductVo.createPricingPhases(
         eligibilityPricings.contains(ELIGIBILITY_PRICING_FREE_TRIAL) &&
         eligibilityPricings.contains(ELIGIBILITY_PRICING_TIERED_PRICE)
     ) {
-        // We are temporarily not handling Tiered Pricing price phases when there is both a trial and a tiered
-        // price available
-        log(LogIntent.GALAXY_WARNING) {
-            GalaxyStrings.PARSING_INTRO_PRICING_PHASES_FOR_SUBS_TIERED_PRICING_NOT_SUPPORTED
-        }
-        this.createFreeTrialPricingPhase()?.let { pricingPhases.addFirst(it) }
+        this.createFreeTrialPricingPhase()?.let { pricingPhases.add(it) }
```
fire-at-will, line 97: Replacing `addFirst` with `add` to avoid some Android OS level availability requirements

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -142,30 +147,80 @@ private fun ProductVo.createFreeTrialPricingPhase(): PricingPhase? {
     } // This returns null if the string couldn't be parsed to an Int
 }
 
-private fun ProductVo.createPrice(): Price =
-    Price(
+private fun ProductVo.createTieredSubscriptionPricingPhase(): PricingPhase? {
```
fire-at-will, line 151: Extracted the price creation logic from `ProductVo.createPrice() to `createPriceFromGalaxyData()` so it can be reused

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -142,30 +147,80 @@ private fun ProductVo.createFreeTrialPricingPhase(): PricingPhase? {
...
         // This way, we can get strings like "$3.00" instead of "$3"
         formatted = "%s%.2f".format(currencyUnit, itemPrice),
         amountMicros = (itemPrice * 1_000_000L).toLong(),
         currencyCode = currencyCode,
     )
+}
 
 @SuppressWarnings("MagicNumber", "ReturnCount")
-private fun ProductVo.createPeriod(): Period? {
+private fun ProductVo.createPeriod(): Period? = createPeriodFromGalaxyData(
```
fire-at-will, line 200: Extracted the period creation logic from ProductVo.createPeriod() to createPeriodFromGalaxyData()` so it can be reused

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreProductConversions.kt
```diff
@@ -142,30 +147,80 @@ private fun ProductVo.createFreeTrialPricingPhase(): PricingPhase? {
...
+private fun createPriceFromGalaxyData(
+    currencyUnit: String,
+    currencyCode: String,
+    itemPrice: Double,
+): Price {
+    return Price(
         // Here, we manually build the formatted string instead of using ProductVo.itemPriceString
         // because itemPriceString doesn't include the decimal values if the amount is an integer with no decimal value.
         // This way, we can get strings like "$3.00" instead of "$3"
         formatted = "%s%.2f".format(currencyUnit, itemPrice),
```
tonidero: I missed this in previous PRs.... But I think we should use `itemPriceString` if available? Note that there are currencies where there are no cents or that it's preferred to show the price without cents when it's a whole number. We could try to handle it ourselves, but I would probably default to what the Galaxy store provides normally.

fire-at-will: I think that makes sense! Updated to use `itemPriceString` in 9992e979ca617f703b1a0302189e7044792ba4c0

PR #2955, opened 2025-12-18
Implement restore/sync purchases for Galaxy Store

fire-at-will:
### Description
Implements the following functions in `GalaxyBillingWrapper`:
- `queryAllPurchases()`
- `findPurchaseInPurchaseHistory()`

This unlocks both syncPurchases() and restorePurchases() behavior in the SDK for the Galaxy Store.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/ProductTypeConversions.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/ProductTypeConversions.kt
@@ -1,9 +1,9 @@
 package com.revenuecat.purchases.galaxy.conversions
 
+import com.revenuecat.purchases.ProductType
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.galaxy.GalaxyStrings
-import com.revenuecat.purchases.ProductType
 import com.samsung.android.sdk.iap.lib.constants.HelperDefine
 
 internal fun String.createRevenueCatProductTypeFromSamsungIAPTypeString(): ProductType {
@@ -18,4 +18,3 @@ internal fun String.createRevenueCatProductTypeFromSamsungIAPTypeString(): Produ
         }
     }
 }
-      
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/GetOwnedListHandler.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/GetOwnedListHandler.kt
@@ -1,18 +1,15 @@
 package com.revenuecat.purchases.galaxy.handler
 
-import com.revenuecat.purchases.ProductType
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesErrorCode
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.galaxy.GalaxyStrings
 import com.revenuecat.purchases.galaxy.IAPHelperProvider
-import com.revenuecat.purchases.galaxy.conversions.toGalaxyProductTypeString
 import com.revenuecat.purchases.galaxy.listener.GetOwnedListResponseListener
 import com.revenuecat.purchases.galaxy.utils.GalaxySerialOperation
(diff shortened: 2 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/ProductTypeConversions.kt
```diff
@@ -0,0 +1,11 @@
+package com.revenuecat.purchases.galaxy.conversions
+
+import com.revenuecat.purchases.ProductType
+
+internal fun ProductType.toGalaxyProductTypeString(): String? {
+    return when (this) {
+        ProductType.SUBS -> "subscription"
```
tonidero: Just to check but is there a constant in the galaxy SDK we could reference instead of hardcoding the string value? This is fine if not.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
```diff
@@ -41,3 +42,32 @@ internal fun PurchaseVo.toStoreTransaction(
...
     )
 }
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal fun OwnedProductVo.toStoreTransaction(
+    purchaseState: PurchaseState,
+): StoreTransaction {
+    val productType = this.type.createRevenueCatProductTypeFromSamsungIAPTypeString()
+
+    // throws IllegalArgumentException if the string is an invalid format
```
tonidero, line 53: Should we mark this method as `@Throws` even though it's not actually enforced in Kotlin?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -77,13 +84,42 @@ internal class GalaxyBillingWrapper(
...
     override fun queryAllPurchases(
         appUserID: String,
         onReceivePurchaseHistory: (List<StoreTransaction>) -> Unit,
         onReceivePurchaseHistoryError: PurchasesErrorCallback,
     ) {
-        warnLog { "Unimplemented: GalaxyBillingWrapper.queryAllPurchases" }
-        onReceivePurchaseHistory(emptyList())
+        serialRequestExecutor.executeSerially { finish ->
+            getOwnedListHandler.getOwnedList(
+                productType = null, // Passing null here queries all product types
```
tonidero: Hmm so, this will always be null... Would it make sense to simplify and remove this parameter entirely, and just always query for all products?

tonidero (approved):
Looks great! I think we could simplify by always passing `all` to the galaxy SDK. Wdyt? other than that :shipit:

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/ProductTypeConversions.kt
fire-at-will: It looks like they do! I've switched us over to use them here df25d3a0d47b337800dc2eeb550fc397473a76d3

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreTransactionConversions.kt
fire-at-will: Yeah, I think so, even though the compiler doesn't enforce it, it could still help a developer realize that it throws. Annotated it with `@Throws` in a9a920b81126e9c4dc3dade554aa48e86ffd3d6b

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Yeah, that could help simplify things! I'll make the change, and if we ever need to query for a specific product type later, we can add it back in.

fire-at-will: Switched it to always request all products here: 85089833a9a7c1dca424d3c10013914d46363626 👍

PR #2956, opened 2025-12-19
Add size testing app and related CI tasks

tonidero:
### Description
This adds a new app, straight out of the AS template that will be used to properly verify the size of the SDK in release builds. Right now the app we use for these tests is built in debug mode which won't properly reflect the final size of the SDK and is also used for snapshots that would be stripped aways from the app (once we enable proper minification).

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/main/offerings/OfferingsScreen.kt
@@ -1,5 +1,6 @@
 package com.revenuecat.paywallstester.ui.screens.main.offerings
 
+import android.util.Log
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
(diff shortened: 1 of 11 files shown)


fastlane/Fastfile
```diff
@@ -66,6 +67,21 @@ platform :android do
...
+  lane :emerge_sdk_size_tests do
+     gradle_properties = {
+       "signing.keyId" => ENV['GPG_SIGNING_KEY_ID_NEW'],
+       "signing.password" => ENV['GPG_SIGNING_KEY_PW_NEW'],
+       "signing.secretKeyRingFile" => "/home/circleci/.gnupg/secring.gpg",
+       "RELEASE_SIGNING_ENABLED" => true
+     }
+     gradle(task: ":purchases:publishToMavenLocal", properties: gradle_properties)
+     gradle(task: ":ui:revenuecatui:publishToMavenLocal", properties: gradle_properties)
+     Dir.chdir("../test-apps/sdksizetesting/app") do
```
tonidero, line 79: This is a bit weird... I need to enter the module folder, because fastlane's `gradle` command searches the `gradlew` executable in the parent folder

JayShortway: Directory handling in Fastlane is a bit odd.

fastlane/Fastfile
```diff
@@ -66,6 +67,21 @@ platform :android do
...
+  desc "Emerge size tests"
+  lane :emerge_sdk_size_tests do
+     gradle_properties = {
+       "signing.keyId" => ENV['GPG_SIGNING_KEY_ID_NEW'],
+       "signing.password" => ENV['GPG_SIGNING_KEY_PW_NEW'],
+       "signing.secretKeyRingFile" => "/home/circleci/.gnupg/secring.gpg",
+       "RELEASE_SIGNING_ENABLED" => true
+     }
+     gradle(task: ":purchases:publishToMavenLocal", properties: gradle_properties)
+     gradle(task: ":ui:revenuecatui:publishToMavenLocal", properties: gradle_properties)
```
JayShortway: Not important, but isn't this enough?


```suggestion
     gradle(task: ":publishToMavenLocal", properties: gradle_properties)
```

test-apps/sdksizetesting/settings.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+                // Exclude to make sure we use dependency from mavenLocal
+                excludeGroupByRegex("com\\.revenuecat.*")
+            }
+        }
+        mavenCentral {
+            content {
+                // Exclude to make sure we use dependency from mavenLocal
+                excludeGroupByRegex("com\\.revenuecat.*")
+            }
+        }
```
JayShortway, line 35: 😚👌

JayShortway (approved):
Beautiful! Thanks for doing this! 🙏

fastlane/Fastfile
tonidero: Hmm good point... I guess this might save a bit of time by not publishing anything else that we don't need (like the debug view library)... But I think that might be ok. Will change it, thanks!

PR #2958, opened 2025-12-19
Add exit offers support to PaywallDialog

vegaro:
On top of #2938 

This one adds exit offers support to PaywallDialog

Same concept, whenever the paywall is dismissed we display a new Dialog with the new paywall

https://github.com/user-attachments/assets/363d63db-23fc-406e-b3d2-d27b9d1a059b

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
@@ -18,10 +18,8 @@ import androidx.compose.runtime.ReadOnlyComposable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
-import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallDialog.kt
```diff
@@ -53,69 +54,144 @@ fun PaywallDialog(
...
+) {
+    val paywallOptions = remember(paywallDialogOptions, offeringSelection) {
+        buildPaywallOptions(
+            paywallDialogOptions = paywallDialogOptions,
+            offeringSelection = offeringSelection,
+            dismissRequest = {},
+        )
+    }
+
+    val viewModel = getPaywallViewModel(paywallOptions)
```
tonidero: Hmm have you tested whether this works with a `requiredEntitlementIdentifier` (or `shouldDisplayBlock`) parameter in the PaywallDialogOptions builder? Seems we're not passing the `shouldDisplayBlock` anymore to this method, so I'm not sure if that would eventually get called if, say, a user gets the required entitlement after restoring?

vegaro: good point. I did test it and it wasn't working properly. I tested `shouldDisplayBlock` I still need to test `requiredEntitlementIdentifier`.

One thing I noticed is that on reopening of the paywall `purchaseCompleted` stayed as `true` , do the paywall was dismissing automatically. It makes sense since the view model is getting reused. I fixed with a `snapshotFlow` in 4ccf964d1d04bbb0f07835cc0972a485e1ec6bb5. Wdyt? do you have a better idea?

tonidero: Oh hmm right... It's very annoying the VM doesn't get cleared after dismissing the paywall, and a potential big source of bugs... Maybe we can try to add some sort of `reset` method to the VM to be called when presenting a paywall that clears any leftover state...

The current snapshotFlow approach seems brittle... but I *think* it might work as a temporal solution.

vegaro: Yea it feels it will break somewhere else

vegaro: I actually just removed the snapshotFlow since it was causing issues with the `dismissRequest`. Commit [79b5f2f](https://github.com/RevenueCat/purchases-android/pull/2958/commits/79b5f2f1decd923de30a33b5d7b64c430e9ab9c1)

Just did another round of testing and it looks like everything is working fine now. DismissRequest and shouldDisplayBlock get called when they should, with and without exit offers.

tonidero (approved):
🚢

PR #2961, opened 2025-12-19
Fix crash when context.cacheDir is null in FontLoader

vegaro:
On some devices looks like especially Samsung and Android 16 maybe with aggressive battery optimization, context.cacheDir can return null when the app is in the background. 

This caused a `NullPointerException` at `FontLoader.kt:36.

This PR handles null `cacheDir` gracefully by logging an error and skipping the font download instead of crashing. The paywall will fall back to system fonts in this edge case.

This is the reported stacktrace:

```
 Fatal Exception: java.lang.NullPointerException:
at kotlin.UnsafeLazyImpl.getValue(Lazy.kt:98)
at com.revenuecat.purchases.paywalls.FontLoader.getCacheDirectory(FontLoader.kt:36)
at com.revenuecat.purchases.paywalls.FontLoader.ensureFoldersExist(FontLoader.kt:164)
at com.revenuecat.purchases.paywalls.FontLoader.access$getContext$p(FontLoader.kt:27)
at com.revenuecat.purchases.paywalls.FontLoader.access$ensureFoldersExist(FontLoader.kt:27)
at com.revenuecat.purchases.paywalls.FontLoader$startFontDownload$1.invokeSuspend(FontLoader.kt:77)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:820)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```

And the stats:

```
Devices: 65% Samsung
Operating systems: 42% Android 16
Device states: 81% background
```

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
@@ -33,8 +33,8 @@ internal class FontLoader(
 ) {
     private var hasCheckedFoldersExist: AtomicBoolean = AtomicBoolean(false)
 
-    private val cacheDirectory by lazy(LazyThreadSafetyMode.NONE) {
-        providedCacheDir ?: File(context.cacheDir, "rc_paywall_fonts")
+    private val cacheDirectory: File? by lazy(LazyThreadSafetyMode.NONE) {
+        providedCacheDir ?: context.cacheDir?.let { File(it, "rc_paywall_fonts") }
     }
 
     private val md: MessageDigest by lazy {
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/FontLoader.kt
```diff
@@ -158,13 +164,13 @@ internal class FontLoader(
         }
     }
 
-    private fun ensureFoldersExist() {
+    private fun ensureFoldersExist(cacheDir: File) {
         if (hasCheckedFoldersExist.getAndSet(true)) return
```
tonidero, line 168: Hmm I know this is not part of this PR… but with this, isn’t it possible that it tries to download later and doesn’t check whether the folder exists?

This should probably be set to false if the folder doesn’t end up being created

vegaro: Right! let me merge this since I want to make a patch release for christmas 🎅  and will open a PR

PR #2962, opened 2025-12-19
Use @ExperimentalPreviewRevenueCatPurchasesAPI for non subscription tracking API

polmiro:
## Changes

Updates the non subscription event tracking API annotations from `@InternalRevenueCatAPI` to `@ExperimentalPreviewRevenueCatPurchasesAPI`.

## Motivation

The non subscription event tracking API was previously marked as internal, preventing external developers from using it. This change makes the API experimental, allowing developers to opt-in while signaling that the API may change in future releases.


purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
```diff
@@ -1,5 +1,11 @@
+@file:OptIn(
+    com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI::class,
+    com.revenuecat.purchases.InternalRevenueCatAPI::class,
+)
+
```
tonidero: Hmm I would probably opt in at the class level below instead of at a file level?

purchases/src/test/java/com/revenuecat/purchases/ads/events/AdTrackerTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:OptIn(com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI::class)
```
tonidero: Same as before, I would probably try to opt in at the class level, instead of file level. Also if we move it there, we should avoid using the fully qualified name and just import the annotation.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -171,8 +171,9 @@ class Purchases internal constructor(
      * The AdTracker used to track ad attribution data.
      */
     @get:JvmSynthetic
-    @InternalRevenueCatAPI
+    @ExperimentalPreviewRevenueCatPurchasesAPI
     val adTracker: AdTracker
+        @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
```
tonidero: Hmm is this one needed? I think you're already propagating the annotation above so I *think* it shouldn't be needed?

polmiro: I think that's right! Giving it a shot. Just amended.

purchases/src/test/java/com/revenuecat/purchases/ads/events/AdTrackerTest.kt
polmiro: Sounds good, just amended.

purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
polmiro: Sounds good, just amended.

PR #2968, opened 2025-12-22
Improve event tracking reliability

tonidero:
### Description
This modifies a few aspecs of our event tracking infrastructure to improve its reliability:
- It flushes events on any activity onPause, in addition to the application onStop.
- It increases the priority of the thread uses to flush events to a normal priority (before it was background priority).
- Changes the order of operations to flush events before flushing legacy events.


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -324,6 +324,10 @@ internal class PurchasesOrchestrator(
         }
     }
 
+    override fun onActivityPaused(activity: Activity) {
```
tonidero, line 327: This has the potential to increase the number of flushes a lot. For example, for any activity change, it would flush events... I think it might be ok and safer than what we do right now where it gets flushed after a timeout when the whole application is backgrounded (I still left the old mechanism as a final catch all though)

vegaro (approved):

PR #2969, opened 2025-12-22
implement GalaxyBillingWrapper.queryPurchases()

fire-at-will:
### Description
This PR:
- Implements `queryPurchases()`, which is used when posting pending receipts
- Calls `purchasesUpdatedListener?.onPurchasesFailedToUpdate()` in `makePurchaseAsync()` when making a product change, which is currently unsupported

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
@@ -100,7 +100,7 @@ internal class GalaxyBillingWrapper(
                 onSuccess = { ownedProducts ->
                     val storeTransactions = ownedProducts.map {
                         try {
-                            it.toStoreTransaction(purchaseState = PurchaseState.UNSPECIFIED_STATE)
+                            it.toStoreTransaction(purchaseState = PurchaseState.PURCHASED)
                         } catch (e: IllegalArgumentException) {
                             val errorMessage = GalaxyStrings.ERROR_CANNOT_PARSE_PURCHASE_RESULT.format(e.message)
                             log(LogIntent.GALAXY_ERROR) { errorMessage }
@@ -373,7 +373,7 @@ internal class GalaxyBillingWrapper(
                         }
                         .map {
                             try {
-                                it.toStoreTransaction(purchaseState = PurchaseState.UNSPECIFIED_STATE)
+                                it.toStoreTransaction(purchaseState = PurchaseState.PURCHASED)
                             } catch (e: IllegalArgumentException) {
                                 val errorMessage = GalaxyStrings.ERROR_CANNOT_PARSE_PURCHASE_RESULT.format(e.message)
                                 log(LogIntent.GALAXY_ERROR) { errorMessage }


purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -347,13 +356,50 @@ internal class GalaxyBillingWrapper(
...
-        warnLog { "Unimplemented: GalaxyBillingWrapper.queryPurchases" }
-        onSuccess(emptyMap())
+        log(LogIntent.DEBUG) { RestoreStrings.QUERYING_PURCHASE }
+        serialRequestExecutor.executeSerially { finish ->
+            getOwnedListHandler.getOwnedList(
+                onSuccess = { ownedProducts ->
+                    val storeTransactions = ownedProducts
+                        .filter {
+                            // TO DO: Find out what this returns for OTPs when we support OTPs
+                            it.subscriptionEndDate.parseDateFromGalaxyDateString() > dateProvider.now
```
tonidero, line 372: Hmm does Galaxy store return all expired subscriptions always? Sad we have to do the filtering locally (time could be modified client side, which could be problematic...). But not familiar if there is a better way.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -347,13 +356,50 @@ internal class GalaxyBillingWrapper(
...
+                        }
+                        .map {
+                            try {
+                                it.toStoreTransaction(purchaseState = PurchaseState.UNSPECIFIED_STATE)
```
tonidero: Just to check, is there a concept of pending purchases in the galaxy store? And if so, there is no way to obtain it? Just wondering if there is a way to avoid using `UNSPECIFIED_STATE` and just default to `PURCHASED`.

purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
```diff
@@ -63,11 +70,17 @@ class GalaxyBillingWrapperTest : GalaxyStoreTest() {
...
         mockkStatic(Looper::class)
+        mockkStatic(Base64::class)
         val mockMainLooper = mockk<Looper> {
             every { thread } returns Thread.currentThread()
         }
         every { Looper.getMainLooper() } returns mockMainLooper
 
+        every { Base64.encode(any(), any()) } answers {
+            val data = it.invocation.args[0] as ByteArray
+            java.util.Base64.getEncoder().encode(data)
```
tonidero, line 81: Nitpick, could be worth adding an import for this.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Their docs say:

> getOwnedList() returns a list of in-app products that the app user currently has from previous purchases

In sandbox, purchases disappear here once they expire, so it probably isn't necessary, but I included this since we don't really know what will happen in production and it looked like we were doing something similar for Amazon

purchases/src/test/java/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperTest.kt
fire-at-will: Line 79 is mocking out `android.util.Base64`, because if we don't, we get an error when our `utils.sha1()` is called:

```
Method encode in android.util.Base64 not mocked. See https://developer.android.com/r/studio-ui/build/not-mocked for details.
```

This mocks it out with `java.util.Base64`, and since the class names are the same, the distinction is necessary

tonidero: Ahhh make sense. Thanks for explaining!

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
fire-at-will: Since there's no concept of pending purchases for the Galaxy store like there is for Google, I've changed this from `UNSPECIFIED` to `PURCHASED` in dde41298ee71611ac0c058f909932813b54f3a02 in both `override fun queryPurchases()` and `queryAllPurchases()`

PR #2974, opened 2025-12-24
Improve error messages when displaying `CustomerCenterActivity` or `PaywallActivity` from other sources

tonidero:
### Description
We got reports of some crashes that might seem to indicate that the activity was not being initiated correctly through the SDK. In order to better debug these situations, here we add a dummy parameter to the CustomerCenterActivity and PaywallActivity that, in case these activities are shown without it, it would fail with a more meaningful message.


vegaro (approved):
interesting!

PR #2975, opened 2025-12-24
Track exit offers

vegaro:
Adds tracking for exit offers. The event means "An exit offer should be loaded now", or "An exit offer is configured and will be followed". The exit offer paywall will track an impression itself.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendPaywallEventTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendPaywallEventTest.kt
@@ -59,6 +59,26 @@ class BackendPaywallEventTest {
         )
     ).map { it.toBackendEvent() })
 
+    private val exitOfferEventRequest = EventsRequest(listOf(
+        BackendStoredEvent.Paywalls(
+            BackendEvent.Paywalls(
+                id = "exit-offer-id",
+                version = 1,
+                type = PaywallEventType.EXIT_OFFER.value,
+                appUserID = "appUserID",
+                sessionID = "sessionID",
+                offeringID = "offeringID",
+                paywallRevision = 3,
+                timestamp = 123456789,
+                displayMode = "fullscreen",
+                darkMode = false,
+                localeIdentifier = "en_US",
+                exitOfferType = "dismiss",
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEvent.kt
```diff
@@ -27,6 +27,27 @@ enum class PaywallEventType(val value: String) {
...
+enum class ExitOfferType(val value: String) {
+    /**
+     * An exit offer shown when the user attempts to dismiss the paywall without interacting.
+     */
+    DISMISS("dismiss"),
+
+    /**
+     * An exit offer shown when the user dismisses the paywall after some interaction (e.g., cancelling a purchase).
+     */
+    ABANDONMENT("abandonment"),
```
facumenzella: We don't support this yet, so I would just remove it

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -183,13 +185,17 @@ internal class PaywallViewModelImpl(
...
         trackPaywallClose()
+        val exitOffering = if (!_purchaseCompleted.value) {
+            _preloadedExitOffering.value
+        } else {
+            null
+        }
+        if (exitOffering != null) {
+            trackExitOffer(ExitOfferType.DISMISS, exitOffering.identifier)
+        }
+        paywallPresentationData = null
```
facumenzella, line 196: Is it possible that we are tracking exit offer, even if `shouldDisplay` will return false?

facumenzella: I guess it's not THAT important because we'll get another impression, right?

vegaro: right exit offer basically means "an exit offer should be displayed now". If there's no impression it means it was never shown

PR #2979, opened 2025-12-29
Improve accuracy of transactions origin Part 2: Store transaction metadata

tonidero:
### Description
Second part of #2925 

This PR adds:
- `LocalTransactionMetadata` data class to contain all cached metadata. This is stored in a map format with the key being a hash of the purchase token.
- `LocalTransactionMetadataCache` class to handle caching and retrieval of the above data.


tonidero:
We're going with a different approach where we will cache at the request level, so this becomes unnecessary. Will close this for now, and will pick parts as needed

PR #2980, opened 2025-12-29
Add better error message when opening SimulatedStoreErrorDialogActivity from non-supported sources

tonidero:
### Description
This improves the error message in some situations that seem to be attributed to Google's automated testing opening activities when it shouldn't.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -104,7 +104,8 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
         if (!wasLaunchedThroughSDK && !Purchases.isConfigured) {
             throw IllegalStateException(
                 "PaywallActivity was not launched through the SDK. " +
-                    "Please use the SDK methods to open the Paywall.",
+                    "Please use the SDK methods to open the Paywall. " +
+                    "This might happen on some Google automated testing, but shouldn't happen to customers.",
```
rickvdl: Maybe users -> customers here would make a bit more sense.

rickvdl (approved):
Thanks for handling this 🙌

PR #2983, opened 2025-12-30
Make Endpoint serializable

tonidero:
### Description
Makes `Endpoint` serializable, so we can cache it in future PRs


tonidero:
Closing for now since we're not making the request cache for now.

PR #2987, opened 2025-12-31
Improve accuracy of transactions origin Part 1: Refactor to allow caching transaction metadata

tonidero:
### Description
The final objective of these PRs are to have more reliability on all transaction metadata that normally gets sent with a purchase POST /receipt. In rare scenarios, when this request fails, the transaction metadata is currently lost. This will cache this information on disk, so we can recover it and post it (in follow-up PRs).

Main changes in this PR:
- Made `ReceiptInfo` serializable by making it include only the information that we currently post to the backend. This cascaded into a whole lot of changes unfortunately. 
- Move pending transactions handling to the PostReceiptHelper, so we can cache metadata for these pending purchases in the same location we currently cache non pending transactions.

### DIFF
--- a/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
@@ -2,7 +2,7 @@
   "statusCode": 200,
   "headers": {
     "Connection": "keep-alive",
-    "Content-Length": "92142",
+    "Content-Length": "92166",
     "Server": "cloudflare",
     "x-revenuecat-app-id": "appc735cd95ea"
   },
@@ -4388,7 +4388,8 @@
       "variable_config": {
         "variable_compatibility_map": {},
         "function_compatibility_map": {}
-      }
+      },
+      "custom_variables": {}
     }
   }
 }
--- a/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform verified offerings backend request/response_002.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform verified offerings backend request/response_002.json
@@ -2,7 +2,7 @@
   "statusCode": 200,
   "headers": {
     "Connection": "keep-alive",
-    "Content-Length": "92142",
+    "Content-Length": "92166",
     "Server": "cloudflare",
     "x-revenuecat-app-id": "appc735cd95ea"
   },
@@ -4388,7 +4388,8 @@
       "variable_config": {
         "variable_compatibility_map": {},
         "function_compatibility_map": {}
-      }
+      },
+      "custom_variables": {}
     }
   }
 }


purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
```diff
@@ -21,6 +22,7 @@ private object PeriodConstants {
  */
 @Parcelize
 @Poko
+@Serializable
 class Period(
```
JayShortway: I wonder if it's worth it to try and avoid making these tiny public classes serializable, as it's more of an implementation detail that we're now committing to not changing. 

We can write a custom serializer, either for each public class separately or only for the `ReceiptInfo` as a whole. Alternatively we could add internal representations (e.g. `PeriodSurrogate`) that we make serializable instead. That is the surrogate pattern described [here](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#composite-serializer-via-surrogate).

purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
```diff
@@ -798,25 +797,7 @@ class PostReceiptHelperTest {
     }
 
     @Test
-    fun `postTransactionAndConsumeIfNeeded posts storeProduct`() {
```
JayShortway, line 801: Are these no longer valid? (I was looking if they are moved instead of removed, but I can't find where they would be moved to haha.)

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -145,10 +147,21 @@ internal class PostReceiptHelper(
...
         val presentedPaywall = paywallPresentedCache.getAndRemovePresentedEvent()
+
+        if (purchaseState == PurchaseState.PENDING) {
+            onError(
+                PurchasesError(PurchasesErrorCode.PaymentPendingError).also { errorLog(it) },
+                PostReceiptErrorHandlingBehavior.SHOULD_NOT_CONSUME,
+                null,
+            )
+            return
+        }
```
JayShortway, line 163: If I understand correctly this is existing behavior, just moved here?

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -60,46 +45,40 @@ constructor(
...
+                platformProductIds = platformProductIds,
+            )
         }
-
-    override fun hashCode(): Int {
-        var result = productIDs.hashCode()
-        result = 31 * result + (presentedOfferingContext?.hashCode() ?: 0)
-        result = 31 * result + (storeProduct?.hashCode() ?: 0)
-        result = 31 * result + (subscriptionOptionId?.hashCode() ?: 0)
-        return result
```
JayShortway, line 82: Any reason why this was custom before?

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: That's correct. I moved this check from the `PostTransactionWithProductDetailsHelper` to the `PostReceiptHelper`. The main reason being that we're adding this caching logic in the lower level `PostReceiptHelper`, and before, pending transactions didn't get here at all in order to have their information cached. 

For developers, the same error should be returned, and should only result in some slightly extra processing in pending transactions (but that can be ignored I think)

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
tonidero: Not that I could find. I tried checking in places where we compare or hash by ReceiptInfo, but it seemed like we don't need a custom implementation here.

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
tonidero: I see your point... I thought about it as this potentially being a new feature, but it's true that then we need to make sure we maintain it correctly. Better to avoid that unless we need to. Will look into this, thanks!!

rickvdl (commented):
Nice foundational work! lot's of changes 🚀  Changes look fine to me, but it might be good to get an extra pair of eyes on these.

purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
tonidero: Right, the ReceiptInfo no longer contains the entire `StoreProduct` or the `subscriptionOptionId` directly anymore, so these tests became invalid. We do verify that most of the ReceiptInfo is posted with the expected data in the `postTransactionAndConsumeIfNeeded posts with expected default parameters` test, which I thought was enough, though it's true that it makes the check against test that could also be tested, so I will change it to make it more deterministic.

purchases/src/main/kotlin/com/revenuecat/purchases/models/Period.kt
ajpallares: The same applies to iOS, btw. I was worried about making `PresentedOfferingContext` implement `Codable`. I will take a similar approach to avoid making the serialization public

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -2,53 +2,48 @@ package com.revenuecat.purchases.common
...
+        @Serializable(with = PricingPhaseSerializer::class)
+        PricingPhase,
+        >? = null,
+    @Serializable(with = ReplacementModeSerializer::class)
     val replacementMode: ReplacementMode? = null,
+    val platformProductIds: List<Map<String, String?>> = emptyList(),
 ) {
+    companion object {
+        @OptIn(InternalRevenueCatAPI::class)
+        fun from(
```
JayShortway, line 39: `ReceiptInfo` is already `internal` right?

purchases/src/test/java/com/revenuecat/purchases/common/ReceiptInfoTest.kt
```diff
@@ -6,25 +6,40 @@
...
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
 import org.assertj.core.api.Assertions.assertThat
 import org.json.JSONObject
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.annotation.Config
 
+@OptIn(ExperimentalSerializationApi::class)
```
JayShortway, line 41: Why is this needed? Is it caused by a serial descriptor of some of our custom serializers? Maybe we can avoid it if we serialize differently, e.g. using the surrogate pattern? Or is the experimental api usage only in tests?

purchases/src/test/java/com/revenuecat/purchases/common/ReceiptInfoTest.kt
```diff
@@ -6,25 +6,40 @@
...
 @RunWith(AndroidJUnit4::class)
 @Config(manifest = Config.NONE)
 class ReceiptInfoTest {
 
     private val productIdentifier = "com.myproduct"
+    private val json = Json.Default
```
JayShortway, line 47: Very nice additional serialization tests!

purchases/src/test/java/com/revenuecat/purchases/common/ReceiptInfoTest.kt
```diff
@@ -145,4 +163,463 @@ class ReceiptInfoTest {
...
+            productIDs = listOf(productIdentifier),
+            price = 4.99,
+            currency = "USD",
+            period = period
+        )
+
+        val encoded = json.encodeToString(receiptInfo)
+        val decoded = json.decodeFromString<ReceiptInfo>(encoded)
+
+        // language=JSON
```
JayShortway, line 185: Big fan of this 😄

purchases/src/test/java/com/revenuecat/purchases/common/ReceiptInfoTest.kt
tonidero: It's actually needed because of the tests I added that use reflection to verify that all fields are serialized, like `PresentedOfferingContextSerializer includes all constructor parameters`. That uses some experimental APIs. I thought that since it's only on tests, it should be ok to use those. Production code is not using any of those.

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
tonidero: This is because of the `SharedConstants.MICRO_MULTIPLIER` we use, which is public with the `InternalRevenueCatAPI` annotation (because it's used in purchases-ui I think).

purchases/src/test/java/com/revenuecat/purchases/common/ReceiptInfoTest.kt
JayShortway: Alright yes, thanks for confirming. Totally fine indeed.

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
JayShortway: Ah right, makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/ReplacementMode.kt
```diff
@@ -12,3 +21,31 @@ interface ReplacementMode : Parcelable {
...
+
+            while (true) {
+                when (val index = decodeElementIndex(descriptor)) {
+                    0 -> name = decodeStringElement(descriptor, 0)
+                    -1 -> break
+                    else -> error("Unexpected index: $index")
+                }
+            }
+
+            GoogleReplacementMode.valueOf(name)
```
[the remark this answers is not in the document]
JayShortway: This is a good one.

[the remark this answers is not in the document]
tonidero: Indeed a good one, fixed in https://github.com/RevenueCat/purchases-android/pull/2987/commits/b6a3f302706df6eeee14cb3595dbe3589411de89

PR #2988, opened 2025-12-31
Improve accuracy of transactions origin Part 2: Store transaction metadata

tonidero:
### Description
Second part of #2987 

This PR adds:
- `LocalTransactionMetadata` data class to contain all cached metadata. This is stored in a map format with the key being a hash of the purchase token.
- `LocalTransactionMetadataCache` class to handle caching and retrieval of the above data.

### DIFF
--- a/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallTrackingAPI.kt
+++ b/api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallTrackingAPI.kt
@@ -48,6 +48,7 @@ private class PaywallTrackingAPI {
             PaywallEventType.IMPRESSION,
             PaywallEventType.CANCEL,
             PaywallEventType.CLOSE,
+            PaywallEventType.EXIT_OFFER,
             -> {}
         }
     }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
@@ -69,6 +69,8 @@ internal sealed class BackendEvent : Event {
      * @property displayMode The display mode of the Paywall.
      * @property darkMode Whether the app was in dark mode at the time of the event.
      * @property localeIdentifier The locale identifier of the device.
+     * @property exitOfferType The type of exit offer shown. Only for exit offer events.
+     * @property exitOfferingID The offering ID of the exit offer shown. Only for exit offer events.
      */
     @Serializable
     @SerialName("paywalls")
@@ -91,6 +93,10 @@ internal sealed class BackendEvent : Event {
         val darkMode: Boolean,
         @SerialName("locale")
         val localeIdentifier: String,
+        @SerialName("exit_offer_type")
+        val exitOfferType: String? = null,
+        @SerialName("exit_offering_id")
+        val exitOfferingID: String? = null,
     ) : BackendEvent()
 
     @Serializable
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
@@ -83,6 +83,8 @@ internal fun PaywallEvent.toBackendStoredEvent(
             displayMode = data.displayMode,
             darkMode = data.darkMode,
             localeIdentifier = data.localeIdentifier,
+            exitOfferType = data.exitOfferType?.value,
+            exitOfferingID = data.exitOfferingIdentifier,
         ),
     )
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -834,6 +834,14 @@ internal class BillingWrapper(
                     debugLog { "Activity is null, not showing Google Play in-app message." }
                     return@withConnectedClient
                 }
+                if (activity.isFinishing) {
+                    debugLog { "Activity is finishing, not showing Google Play in-app message." }
+                    return@withConnectedClient
+                }
+                if (activity.isDestroyed) {
+                    debugLog { "Activity is destroyed, not showing Google Play in-app message." }
+                    return@withConnectedClient
+                }
                 showInAppMessages(activity, inAppMessageParams) { inAppMessageResult ->
                     when (val responseCode = inAppMessageResult.responseCode) {
                         InAppMessageResult.InAppMessageResponseCode.NO_ACTION_NEEDED -> {
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEvent.kt
@@ -27,6 +27,22 @@ enum class PaywallEventType(val value: String) {
      * The paywall was dismissed.
      */
     CLOSE("paywall_close"),
+
+    /**
+     * An exit offer will be shown to the user.
+     */
+    EXIT_OFFER("paywall_exit_offer"),
+}
+
+/**
+ * Types of exit offers. Meant for RevenueCatUI use.
+ */
+@InternalRevenueCatAPI
+enum class ExitOfferType(val value: String) {
+    /**
+     * An exit offer shown when the user attempts to dismiss the paywall without interacting.
+     */
+    DISMISS("dismiss"),
 }
 
 /**
@@ -57,6 +73,8 @@ data class PaywallEvent(
         val displayMode: String, // Refer to PaywallMode in the RevenueCatUI module.
         val localeIdentifier: String,
         val darkMode: Boolean,
+        val exitOfferType: ExitOfferType? = null,
+        val exitOfferingIdentifier: String? = null,
     )
 
     internal fun toPaywallPostReceiptData(): PaywallPostReceiptData {
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallStoredEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallStoredEvent.kt
@@ -21,6 +21,7 @@ internal data class PaywallStoredEvent(
         }
     }
 
+    @OptIn(InternalRevenueCatAPI::class)
     fun toBackendEvent(): BackendEvent.Paywalls {
         return BackendEvent.Paywalls(
             id = event.creationData.id.toString(),
@@ -34,6 +35,8 @@ internal data class PaywallStoredEvent(
             displayMode = event.data.displayMode,
             darkMode = event.data.darkMode,
             localeIdentifier = event.data.localeIdentifier,
+            exitOfferType = event.data.exitOfferType?.value,
+            exitOfferingID = event.data.exitOfferingIdentifier,
         )
     }
 
--- a/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/GoldenFileRecorder.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/GoldenFileRecorder.kt
@@ -118,6 +118,7 @@ class GoldenFileRecorder(
             "Vary",
             "Authorization",
             "X-Version",
+            "Server-Timing",
             "Accept-Ranges",
         )
 
--- a/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendPaywallEventTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendPaywallEventTest.kt
@@ -59,6 +59,26 @@ class BackendPaywallEventTest {
         )
     ).map { it.toBackendEvent() })
 
+    private val exitOfferEventRequest = EventsRequest(listOf(
+        BackendStoredEvent.Paywalls(
+            BackendEvent.Paywalls(
+                id = "exit-offer-id",
+                version = 1,
+                type = PaywallEventType.EXIT_OFFER.value,
+                appUserID = "appUserID",
+                sessionID = "sessionID",
+                offeringID = "offeringID",
+                paywallRevision = 3,
+                timestamp = 123456789,
+                displayMode = "fullscreen",
+                darkMode = false,
+                localeIdentifier = "en_US",
+                exitOfferType = "dismiss",
+                exitOfferingID = "exit-offering-id",
+            )
+        )
+    ).map { it.toBackendEvent() })
+
     private lateinit var appConfig: AppConfig
     private lateinit var httpClient: HTTPClient
 
@@ -121,6 +141,40 @@ class BackendPaywallEventTest {
                         "\"display_mode\":\"footer\"," +
                         "\"dark_mode\":true," +
                         "\"locale\":\"en_US\"" +
+                    "}" +
+                "]" +
+            "}"
+        )
+    }
+
+    @Test
+    fun `postPaywallEvents posts exit offer events correctly`() {
+        mockHttpResult()
+        backend.postEvents(
+            exitOfferEventRequest,
+            baseURL = AppConfig.paywallEventsURL,
+            delay = Delay.DEFAULT,
+            onSuccessHandler = {},
+            onErrorHandler = { _, _ -> },
+        )
+        verifyCallWithBody(
+            "{" +
+                "\"events\":[" +
+                    "{" +
+                        "\"discriminator\":\"paywalls\"," +
+                        "\"id\":\"exit-offer-id\"," +
+                        "\"version\":1," +
+                        "\"type\":\"paywall_exit_offer\"," +
+                        "\"app_user_id\":\"appUserID\"," +
+                        "\"session_id\":\"sessionID\"," +
+                        "\"offering_id\":\"offeringID\"," +
+                        "\"paywall_revision\":3," +
+                        "\"timestamp\":123456789," +
+                        "\"display_mode\":\"fullscreen\"," +
+                        "\"dark_mode\":false," +
+                        "\"locale\":\"en_US\"," +
+                        "\"exit_offer_type\":\"dismiss\"," +
+                        "\"exit_offering_id\":\"exit-offering-id\"" +
                     "}" +
                 "]" +
             "}"
--- a/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
@@ -1519,6 +1519,8 @@ class BillingWrapperTest {
     @Test
     fun `showing inapp messages calls show inapp messages correctly`() {
         val activity = mockk<Activity>()
+        every { activity.isFinishing } returns false
+        every { activity.isDestroyed } returns false
         every { mockClient.showInAppMessages(activity, any(), any()) } returns billingClientOKResult
 
         wrapper.showInAppMessagesIfNeeded(activity, InAppMessageType.values().toList()) {
@@ -1531,6 +1533,8 @@ class BillingWrapperTest {
     @Test
     fun `showing inapp messages handles inapp messages listener response correctly when no messages`() {
         val activity = mockk<Activity>()
+        every { activity.isFinishing } returns false
+        every { activity.isDestroyed } returns false
         val listenerSlot = slot<InAppMessageResponseListener>()
         every { mockClient.showInAppMessages(activity, any(), capture(listenerSlot)) } returns billingClientOKResult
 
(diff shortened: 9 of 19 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
```diff
@@ -0,0 +1,41 @@
...
+    val schemaVersion: Int = SCHEMA_VERSION,
+) {
+    companion object {
+        const val SCHEMA_VERSION = 1
+    }
+
+    @Serializable
+    data class TransactionMetadata(
+        @SerialName("user_id")
+        val userID: String,
```
rickvdl, line 27: On iOS I see this often (always?) mentioned as appUserID so I did the same in this structure on iOS.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -151,8 +154,22 @@ internal class PostReceiptHelper(
         onSuccess: (PostReceiptResponse) -> Unit,
         onError: PostReceiptDataErrorCallback,
     ) {
+        val hasCachedTransactionMetadata = hasCachedTransactionMetadata(purchaseToken)
```
rickvdl, line 157: Nit: on iOS we use a `storeIfNeeded` helper method that performs these checks, should save a bit of duplicate code :)

purchases/src/test/java/com/revenuecat/purchases/common/caching/LocalTransactionMetadataCacheTest.kt
```diff
@@ -0,0 +1,496 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.json.JSONObject
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(manifest = Config.NONE)
+class LocalTransactionMetadataCacheTest {
```
rickvdl: Nice set of tests 😄

rickvdl (commented):
Nice work! Also great set of tests 🙌

rickvdl (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
tonidero: You're right, I should use that naming here as well, thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataCache.kt
```diff
@@ -0,0 +1,121 @@
+package com.revenuecat.purchases.common.caching
+
+import com.revenuecat.purchases.JsonTools
+import com.revenuecat.purchases.common.debugLog
+import com.revenuecat.purchases.common.errorLog
+import com.revenuecat.purchases.common.sha1
+import com.revenuecat.purchases.common.verboseLog
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+import java.util.concurrent.atomic.AtomicReference
```
JayShortway: This needs [core library desugaring](https://developer.android.com/studio/write/java8-support-table) afaik 🙁 

Maybe we can just use a good ol' lock?

tonidero: I *think* it's only some methods on this class, but not the entire class right?
image

Still, great callout! might indeed be worth to avoid using this altogether, since it could indeed be easy to use the affected methods or try to use it in other parts of the code... So I will change to use a Lock just in case 👍

tonidero: After taking a closer look, we don't really need the `AtomicReference` at all, since the methods in the cache are already `@Synchronized`, so I removed that. Lmk what you think 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * Contains ephemeral data associated with purchases that may be lost during retry attempts.
+ * This data is cached before posting receipts and cleared upon a successful post attempt.
+ */
+@Serializable
+internal data class LocalTransactionMetadata(
+    @SerialName("purchase_data_by_token_hash")
+    val purchaseDataByTokenHash: Map<String, TransactionMetadata>,
```
ajpallares, line 15: Probably this has been discussed already (and I may have missed it) but, why store only one `local_transaction_metadata` file with all the transactions metadata instead of one transaction metadata file per transaction?
I do think this makes sense btw, but wanted to know the reason. I'll probably do the same in iOS

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    private var cachedData: LocalTransactionMetadata? = null
+
+    @Synchronized
+    fun cacheLocalTransactionMetadata(
+        purchaseToken: String,
+        data: LocalTransactionMetadata.TransactionMetadata,
+    ) {
+        if (hasCachedData(purchaseToken)) {
+            debugLog { "Purchase data already cached for token: $purchaseToken. Skipping cache." }
```
ajpallares, line 27: Just making sure we **always** want this. Could we ever want to update this cache with additional data that was missing before?

tonidero: I don't think additional data after the purchase is made would be useful I think... at least nothing comes to mind but lmk if you have other thoughts!

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
tonidero: TBH, either could work... my thinking was to try to keep a version of this in memory, and for that, having it on a single file was easier. I'm actually using SharedPreferences right now for simplicity, but we could change it to a separate single file if needed.
One thing is to make sure we use the api key as part of the key as well. I don't think changing api keys is that common, but this way we avoid shipping this to the wrong app in that scenario.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
ajpallares: Yeah, I agree. I can't think of anything either

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
ajpallares: I like the idea, yes. I will do the same in iOS to also have a version of it in memory. I think it's safe to assume that we won't have a big object in memory due to multiple transactions metadata being cached at the same time time, so there's no really use to storing them separately 👍 
> One thing is to make sure we use the api key as part of the key as well. I don't think changing api keys is that common, but this way we avoid shipping this to the wrong app in that scenario.

That's a very good point, yes. I'll take that into account for iOS and make sure I'll be storing using an api-key based key. Great suggestion!

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
```diff
@@ -0,0 +1,119 @@
...
+            localTransactionMetadata
+        } catch (e: SerializationException) {
+            errorLog(e) { "Failed to deserialize local transaction metadata. Clearing cache." }
+            deviceCache.remove(CACHE_KEY)
+            null
+        }
+    }
+
+    private fun getTokenHash(purchaseToken: String): String {
+        return purchaseToken.sha1()
```
ajpallares, line 109: Probably not a big deal, but, SHA1 is considered [cryptographically insecure](https://developer.apple.com/documentation/cryptokit/insecure/sha1). Is there any objection to using a secure hash like SHA512, for example?

tonidero: TBH, we might not even need to do this since we're already storing the original purchase token as well... This is mostly a way to key them by something unique that is not super long, and wanted something quick (sha1 is supposed to be faster). But honestly not a strong opinion, lmk and happy to change it to something else (probably sha512 no since I think that may be slower, no need to use too much computation for this I think)

ajpallares: Right! It's true that the token itself is being saved "plain" in the `TransactionMetadata`. I hadn't though of that. I don't have a strong opinion on this either. We can leave it as sha1 then 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
tonidero: Did this rename in this other PR: https://github.com/RevenueCat/purchases-android/pull/2993

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Looking at the latest logic [in the latest PR](https://github.com/RevenueCat/purchases-android/pull/2993), we need to know whether it was cached or not in order to clear it from cache. We can probably do this in a different way, but I think I might leave the logic as it is for now. Do lmk if you feel strongly about it and can dedicate some more time to fix it.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -151,8 +154,22 @@ internal class PostReceiptHelper(
...
 
+        if (shouldCacheTransactionMetadata) {
+            val dataToCache = LocalTransactionMetadata.TransactionMetadata(
+                userID = appUserID,
+                token = purchaseToken,
+                receiptInfo = receiptInfo,
+                paywallPostReceiptData = presentedPaywall?.toPaywallPostReceiptData(),
+                observerMode = !finishTransactions,
+            )
+            localTransactionMetadataCache.cacheLocalTransactionMetadata(purchaseToken, dataToCache)
```
JayShortway: This does 2 cache reads and a write. In between those calls the cache state could have changed. 
- `hasCachedTransactionMetadata` calls `localTransactionMetadataCache.getLocalTransactionMetadata`
- `shouldCacheTransactionMetadata` calls `localTransactionMetadataCache.getLocalTransactionMetadata`
- lastly `localTransactionMetadataCache.cacheLocalTransactionMetadata` is called.

Maybe the `LocalTransactionMetadataCache` could have a `storeIfNeeded` function that handles thread safety internally, avoiding the need for external locks?

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataCache.kt
```diff
@@ -0,0 +1,119 @@
...
+
+import com.revenuecat.purchases.JsonTools
+import com.revenuecat.purchases.common.debugLog
+import com.revenuecat.purchases.common.errorLog
+import com.revenuecat.purchases.common.sha1
+import com.revenuecat.purchases.common.verboseLog
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+
+internal class LocalTransactionMetadataCache(
```
JayShortway: I know I used the term "cache", and this is definitely not the biggest of deals, but this is not really a cache 😅 Maybe we can call it something else, like `LocalTransactionMetadataStore`? FFTI

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataCache.kt
```diff
@@ -0,0 +1,119 @@
...
+            errorLog(e) { "Failed to serialize local transaction metadata" }
+        }
+    }
+
+    fun getLocalTransactionMetadata(purchaseToken: String): LocalTransactionMetadata.TransactionMetadata? {
+        return getCachedData()?.purchaseDataByTokenHash?.get(getTokenHash(purchaseToken))
+    }
+
+    @Synchronized
+    fun clearLocalTransactionMetadata(purchaseTokens: List<String>) {
```
JayShortway: Should this be a set?
```suggestion
    fun clearLocalTransactionMetadata(purchaseTokens: Set<String>) {
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
```diff
@@ -0,0 +1,119 @@
...
+        val tokenHashesToClear = purchaseTokens.map { getTokenHash(it) }.toSet()
+        val existingTokenHashes = existingData.purchaseDataByTokenHash.keys
+
+        val tokensNotFound = tokenHashesToClear - existingTokenHashes
+        if (tokensNotFound.isNotEmpty()) {
+            debugLog {
+                "Some transaction metadata not found in cache when trying to clear: ${tokensNotFound.size} tokens. " +
+                    "Ignoring."
+            }
+        }
```
JayShortway: Do we need to log this? I had to do a double take to understand. At first I thought we were early returning.
If we want to log it, maybe we can phrase it differently:

```kotlin
"Ignoring clearing metadata for ${tokensNotFound.size} transaction(s) as they are not locally persisted."
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
```diff
@@ -0,0 +1,119 @@
...
+import com.revenuecat.purchases.common.verboseLog
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+
+internal class LocalTransactionMetadataCache(
+    private val deviceCache: DeviceCache,
+    private val json: Json = JsonTools.json,
+) {
+    private companion object {
+        const val CACHE_KEY = "local_transaction_metadata"
```
JayShortway, line 16: Instead of using `DeviceCache` and a single key, we could also use a separate `SharedPreferences` file for this, and each transaction could have its own key? That way we don't need to read everything in memory each time, and it's easier to support backwards compatibility.

tonidero: So I'm ok separating to a different SharedPreferences file if we prefer, but I might want to keep it under a single key. Reason is that we still need to access all the cached values when attempting to flush any remaining data, as can be seen in this follow-up PR: https://github.com/RevenueCat/purchases-android/pull/2993/

JayShortway: Hmm can't we use [`SharedPreferences.getAll()`](https://developer.android.com/reference/android/content/SharedPreferences#getAll()) for that?

tonidero: Indeed we can, but basically this means we need to also fetch this in memory on every app foreground anyway right? My thoughts were to have more control by having a single serializable object stored, and then, we have an instance of this "global" object cached in memory for faster access (note that I don't expect this object to ever get too big)

TBH, I think either approach could work and I don't feel too strongly either case, so I can change it if you prefer.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: You have a point... @rickvdl also suggested this, and I see the possible edge cases. Changing it, thanks!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataCache.kt
tonidero: cc @ajpallares

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
JayShortway: True, but we would read the whole thing in memory only once, instead of every time we want to read or write a single transaction. 

I also think it would be easier to handle backwards compatibility if we are dealing with individual `TransactionMetadata` objects. For instance in the future we could have `TransactionMetadataV1` and `TransactionMetadataV2`, and when deserializing we check which one we actually got. 

That to me sounds easier than having to manage a `Map<String, TransactionMetadataV1OrV2>`. Both are definitely possible though. It's not a blocker for me.

tonidero: I've applied this change in a follow-up PR to avoid conflicts: https://github.com/RevenueCat/purchases-android/pull/3002

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
tonidero: Removed log in https://github.com/RevenueCat/purchases-android/pull/2988/commits/86440d047626a1be43a35db96fe07edd2ef31eb0. Agreed, seemed a bit unnecessary.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: I ended up moving the refactors here to this last PR: https://github.com/RevenueCat/purchases-android/pull/3002. Note that I ended up having the synchonous method live in the PostReceiptHelper. Reason is that between the fetching and caching of data, there is some orchestration that needs to happen, and that's much simpler in the PostReceiptHelper. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
JayShortway: Thank you!

purchases/src/test/java/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStoreTest.kt
```diff
@@ -0,0 +1,492 @@
...
+                    "session_id":"session_id",
+                    "revision":1,
+                    "display_mode":"full_screen",
+                    "dark_mode":false,
+                    "locale":"en_US",
+                    "offering_id":"offering_id"
+                 },
+                 "observer_mode":false
+              }
+           }
```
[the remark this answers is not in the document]
JayShortway: Hmm, I would expect tests to fail if this were true.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
JayShortway: Works for me!

purchases/src/test/java/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStoreTest.kt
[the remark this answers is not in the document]
tonidero: Hmm right... this is because that value is a default, and the json instance we're using does not encode defaults... This could indeed be a problem, if we ever need to migrate data (not right now, but with more migrations, it could...).

I think I might remove the field for now, and we can add it once we need to do the first migration? We don't really know if/when we would need to migrate data, so probably better to not change the encoder without need I think. Will remove this field for now in the [last PR](https://github.com/RevenueCat/purchases-android/pull/3002) of the chain. Thanks for mentioning this!!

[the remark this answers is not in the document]
tonidero: As mentioned, I ended up just removing this property in the follow-up PR. https://github.com/RevenueCat/purchases-android/pull/3002/commits/3df3c900103507c9139c540f881280e55794a887. We can add it once we actually need to do a migration I think. cc @ajpallares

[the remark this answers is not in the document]
ajpallares: Makes sense! The lack of a value can be itself a value when we need to do the migration. Thanks for the ping!

rickvdl (approved):
Thanks for all the iterations 🙌

PR #2989, opened 2025-12-31
Improve accuracy of transactions origin Part 3: Merge cached data when posting receipts + Cache amazon data

tonidero:
### Description
These are the main changes in this PR:
- Now, when posting a receipt, if we find that there is any cached information we will include this information in that POST /receipt. This doesn't handle edge cases that may happen when the purchase is not found from the store anymore (will handle in follow-up PRs)
- Include amazon store user id + marketplace in ReceiptInfo, so it can be cached and posted later in POST /receipt when needed.
- Include original `observer_mode` in post receipt request body. The cached value will have priority over the current one.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
@@ -1893,7 +1893,6 @@ class PostReceiptHelperTest {
             ),
             type = PaywallEventType.IMPRESSION,
         )
-        val presentedPaywallData = presentedEvent.toPaywallPostReceiptData()
         paywallPresentedCache.cachePresentedPaywall(presentedEvent)
 
         mockPostReceiptSuccess()


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -275,12 +272,12 @@ internal class Backend(
             "applied_targeting_rule" to receiptInfo.presentedOfferingContext?.targetingContext?.let {
                 return@let mapOf("revision" to it.revision, "rule_id" to it.ruleId)
             },
-            "observer_mode" to !finishTransactions,
+            "observer_mode" to (originalObserverMode ?: !finishTransactions),
```
JayShortway: Should we even fall back to the current value? That makes the field ambiguous, as it can now mean either "observer mode at time of transaction" or "observer mode at time of posting". Do we even care about the latter? If yes, maybe we should make separate fields?

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -66,12 +69,47 @@ internal data class ReceiptInfo(
...
+                storeUserID = storeTransaction.storeUserID,
+                marketplace = storeTransaction.marketplace,
             )
         }
     }
 
     @IgnoredOnParcel
     val duration: String? = period?.iso8601?.takeUnless { it.isEmpty() }
+
+    fun merge(receiptInfo: ReceiptInfo): ReceiptInfo {
```
JayShortway: It would be helpful if this function had a docstring explaining that values in `this` ReceiptInfo have priority over the provided `receiptInfo`. 

(Also to me a name like `mergeWith()` reflects that prioritization better but I realize that that's entirely subjective haha.)

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -147,17 +141,20 @@ internal class PostReceiptHelper(
...
         onError: PostReceiptDataErrorCallback,
     ) {
-        val hasCachedTransactionMetadata = hasCachedTransactionMetadata(purchaseToken)
-        val shouldCacheTransactionMetadata = shouldCacheTransactionMetadata(purchaseToken, initiationSource)
+        val cachedTransactionMetadata = localTransactionMetadataCache.getLocalTransactionMetadata(purchaseToken)
+        val shouldCacheTransactionMetadata = shouldCacheTransactionMetadata(cachedTransactionMetadata, initiationSource)
 
         val presentedPaywall = paywallPresentedCache.getAndRemovePresentedEvent()
+        val effectivePaywallData = presentedPaywall?.toPaywallPostReceiptData()
+            ?: cachedTransactionMetadata?.paywallPostReceiptData
```
JayShortway: Could this be a potential source of bugs if we're posting cached receipts while a paywall is showing? It might accidentally attribute the cached receipt to the currently showing paywall, if I understand correctly, right?

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -261,11 +257,10 @@ internal class PostReceiptHelper(
...
-        return !hasCachedTransactionMetadata(purchaseToken) && source == PostReceiptInitiationSource.PURCHASE
-    }
-
-    private fun hasCachedTransactionMetadata(purchaseToken: String): Boolean {
-        return localTransactionMetadataCache.getLocalTransactionMetadata(purchaseToken) != null
+    private fun shouldCacheTransactionMetadata(
+        cachedTransactionMetadata: LocalTransactionMetadata.TransactionMetadata?,
+        source: PostReceiptInitiationSource,
+    ): Boolean {
+        return cachedTransactionMetadata == null && source == PostReceiptInitiationSource.PURCHASE
```
JayShortway, line 271: Does this prevent caching the paywall information when using observer mode with paywalls? (I'm assuming the `PostReceiptInitiationSource` would not be `PURCHASE` in this case.)

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -66,12 +69,47 @@ internal data class ReceiptInfo(
...
+            } else {
+                this
+            }
+        }
+        return ReceiptInfo(
+            productIDs = this.productIDs,
+            presentedOfferingContext = this.presentedOfferingContext.merge(receiptInfo.presentedOfferingContext),
+            price = this.price ?: receiptInfo.price,
+            formattedPrice = this.formattedPrice ?: receiptInfo.formattedPrice,
+            currency = this.currency ?: receiptInfo.currency,
```
JayShortway: Minor:

```suggestion
            formattedPrice = this.formattedPrice?.takeUnless { it.isBlank() } ?: receiptInfo.formattedPrice,
            currency = this.currency?.takeUnless { it.isBlank() } ?: receiptInfo.currency,
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -66,12 +69,47 @@ internal data class ReceiptInfo(
...
+            period = this.period ?: receiptInfo.period,
+            pricingPhases = this.pricingPhases ?: receiptInfo.pricingPhases,
+            replacementMode = this.replacementMode ?: receiptInfo.replacementMode,
+            platformProductIds = this.platformProductIds.ifEmpty {
+                receiptInfo.platformProductIds
+            },
```
JayShortway, line 122: Just double checking: should these lists and/or maps be merged?

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -23,6 +23,9 @@ internal data class ReceiptInfo(
     val pricingPhases: List<PricingPhase>? = null,
     val replacementMode: ReplacementMode? = null,
     val platformProductIds: List<Map<String, String?>> = emptyList(),
+    // Amazon-only fields
+    val storeUserID: String? = null,
+    val marketplace: String? = null,
```
JayShortway, line 38: ❤️

JayShortway (commented):
Awesomeee! Some minor comments and some philosophical ones 😄

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Hmm you're indeed right that this may potentially happen... and could also be happening right now I think, since the event storage mechanism is completely unattached to the posting of the receipt here, we don't know if it's related... We should think on how to fix this issue more permanently... Great catch!!

For this PR, I'm thinking we could potentially check, if there is a cached value for the given token we don't even need to call `getAndRemovePresentedEvent` above, to avoid missattributions later, but as mentioned, this wouldn't fix the existing issue I think.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Right, I went back and forth on this (see [commit](https://github.com/RevenueCat/purchases-android/pull/2989/commits/c3bf6dfd8be744de101c238dbf838f2be9ac1678) 😅 ).

My thoughts in the end were, we should always use the cached value if we have any cached data. If we don't, we should use the current value, since we can assume the purchase was just performed and/or comes from the queue. 

My main reason for not separating it is that we're already sending this information on a header for all requests, so maybe we don't really need to send it at all unless we have a cached value. However, not sending it would mean possible changes in the backend, so I decided to leave the current behavior if the purchase is not cached. 

But yeah, this is indeed confusing, so might be worth sending this in a new field to avoid confusion.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Hmm you're totally right... I was thinking to avoid storing useless information (when it comes from the queue/restores) but I had indeed miss the case for observer mode. I will fix this, great catch!

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
JayShortway: >if there is a cached value for the given token we don't even need to call getAndRemovePresentedEvent above

True, but we should make sure the `paywallPresentedCache` doesn't contain stale data in any case. 

An additional thing we could do is check if the presented paywall's Offering corresponds to the PresentedOfferingContext we have? Not airtight by any means, but it does allow us to eliminate obvious missattributions.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
JayShortway: Yea I think we should avoid confusion, for us and for the backend. It could even lead to some subtly incorrect attributions. Maybe we leave the header as-is and introduce a new field in the body. (I'd also be okay removing the header if the backend isn't reading it).

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: > we should make sure the paywallPresentedCache doesn't contain stale data in any case.

Right, I was checking and we basically should be clearing the cached `paywallPresentedCache` when:
- Performing a successful `POST /receipt`.
- When closing the paywall (or at least asked to be dismissed). This includes:
   - When navigating back with the back button
   - When clicking on close buttons with the navigate back action
   - When navigation to a web checkout url.

I believe this shouldn't leave cached paywall information around.

> An additional thing we could do is check if the presented paywall's Offering corresponds to the PresentedOfferingContext we have? Not airtight by any means, but it does allow us to eliminate obvious missattributions.

I do think this is a valid check, but will likely do that in a separate PR, since it seems like a different thing than what we're trying to do in this PR

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: I was actually right the first time 😅 When a user purchases through observer mode, we receive the purchase immediately in the billing client listener, and post it with a `PURCHASE` initiation source. [See here](https://github.com/RevenueCat/purchases-android/blob/218609502d75711c3bf85cfd063ba6c3608addd1/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt#L1332-L1339).

So I believe it's safe to only cache in these scenarios. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
tonidero: Well... I don't think there should ever be a case where both have values IMO... I think the cached ones may have values with more information that includes what base plan and/or offer id was used to purchase. However, when recovering information from a restore/queue, only the productId would be available, and we should already have that information in the cache. So basically, I don't think it's needed

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/2989/commits/a7f7d229bfd075bb0b78b90e1ce6f00b8c965da7

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Changed the order of operations in https://github.com/RevenueCat/purchases-android/pull/2989/commits/a7f7d229bfd075bb0b78b90e1ce6f00b8c965da7.

Will take note and do try to figure a more robust solution for the cached paywall events in a follow-up PR

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
tonidero: Renamed and added docstring in https://github.com/RevenueCat/purchases-android/pull/2989/commits/a7f7d229bfd075bb0b78b90e1ce6f00b8c965da7

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: I split this into a separate field: https://github.com/RevenueCat/purchases-android/pull/2989/commits/3878eef8b80deb73ed2b8a1137318e9eb4ee3e02

@antoniobg we will need to accept this new field in the backend. It's meaning will be the value of this field at the time when the purchase was made (which could have potentially changed if the value changed before the data was successfully posted)

antoniobg: If it's a new parameter, can we change the name that matches better the configuration, like `purchase_completion_method` or `completion_method` ?

tonidero: Indeed... what do you think of `purchase_completed_by`, with possible values being `my_app` and `revenuecat`? Note that the public API for this is going to be `PurchasesAreCompletedBy.MY_APP` or `PurchasesAreCompletedBy.REVENUECAT`. cc @RevenueCat/coresdk @antoniobg

antoniobg: Sounds good to me

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -147,17 +141,24 @@ internal class PostReceiptHelper(
...
         onError: PostReceiptDataErrorCallback,
     ) {
-        val hasCachedTransactionMetadata = hasCachedTransactionMetadata(purchaseToken)
-        val shouldCacheTransactionMetadata = shouldCacheTransactionMetadata(purchaseToken, initiationSource)
+        val cachedTransactionMetadata = localTransactionMetadataCache.getLocalTransactionMetadata(purchaseToken)
+        val shouldCacheTransactionMetadata = shouldCacheTransactionMetadata(cachedTransactionMetadata, initiationSource)
 
-        val presentedPaywall = paywallPresentedCache.getAndRemovePresentedEvent()
+        val presentedPaywall = if (cachedTransactionMetadata == null) {
+            paywallPresentedCache.getAndRemovePresentedEvent()
```
rickvdl, line 155: Should we also remove the in-memory cached value if we don't use it (if we don't call `getAndRemovePresentedEvent()`) to prevent it from being used in the next (non-paywall?) purchase?

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -147,17 +141,24 @@ internal class PostReceiptHelper(
...
+        } else {
+            null
+        }
+        val effectivePaywallData = cachedTransactionMetadata?.paywallPostReceiptData
```
rickvdl, line 159: Did we decide on keying the in-memory cached `presentedPaywall` by product id / transaction ID or similar? In order to avoid mixing this data with an unrelated transaction? Not sure if it's a realistic scenario or only theoretical one, but what if a purchase comes in from the queue while this in-memory value is set?

rickvdl (commented):
Awesome work!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Ok renamed in https://github.com/RevenueCat/purchases-android/pull/2989/commits/95af91f685a74a30175fa4af6a750d1b705594df

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Note that the cachedTransactionMetadata's paywall is indeed keyed by the purchase token hash. As for the in memory event (that came from before these PRs), it's indeed a problem I'm thinking about how to solve... It normally shouldn't be an issue since we clear the in memory cache when closing the paywall, but there are edge cases that are not correctly handled, like when making a purchase when `PurchasesAreCompletedBy.MY_APP` or if for any reason presenting multiple paywalls at once.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: We shouldn't, since it's possible the paywall is still displayed while we're syncing purchases in the background. We should only clear it in the same situations we were clearing it before (upon a successful purchase and/or when closing the paywall)

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -147,25 +143,32 @@ internal class PostReceiptHelper(
...
+            ?: presentedPaywall?.toPaywallPostReceiptData()
+        val effectiveReceiptInfo = cachedTransactionMetadata?.receiptInfo?.let { receiptInfo.mergeWith(it) }
+            ?: receiptInfo
+        val originalPurchasesAreCompletedBy = cachedTransactionMetadata?.purchasesAreCompletedBy
```
ajpallares: Should we fall back here to the current value of `purchasesAreCompletedBy` (as the other fields) or is this only supposed to be sent when posting cached transaction metadata?

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
```diff
@@ -35,7 +36,7 @@ internal data class LocalTransactionMetadata(
         @SerialName("paywall_data")
         val paywallPostReceiptData: PaywallPostReceiptData? = null,
 
-        @SerialName("observer_mode")
-        val observerMode: Boolean? = null,
+        @SerialName("purchases_are_completed_by")
+        val purchasesAreCompletedBy: PurchasesAreCompletedBy? = null,
```
ajpallares: Question: does this need to be optional? I'm thinking we're always going to have this value set at the time of a purchase

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Honestly, open to do it either way... It might be good to fall back to the current value, in case there is no cached value, so the field basically means, the value at the time of the purchase.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadata.kt
tonidero: Right... That's a good point. Will change it, thanks!

tonidero: Applied changes for this here: https://github.com/RevenueCat/purchases-android/pull/2989/commits/5af7be11a87f9c3ef6f341c7774a7f51419eb056

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Applied changes so we always send it, falling back to the current value here: https://github.com/RevenueCat/purchases-android/pull/2989/commits/5af7be11a87f9c3ef6f341c7774a7f51419eb056

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
JayShortway: Makes sense, thanks for explaining!

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
JayShortway: Oh nice, yes in that case I think it's safe!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -276,11 +275,12 @@ internal class Backend(
                 return@let mapOf("revision" to it.revision, "rule_id" to it.ruleId)
             },
             "observer_mode" to !finishTransactions,
+            "purchase_completed_by" to purchasesAreCompletedBy.name.lowercase(),
```
JayShortway, line 279: ❤️

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -76,12 +79,51 @@ internal data class ReceiptInfo(
...
     }
 
     @IgnoredOnParcel
     val duration: String? = period?.iso8601?.takeUnless { it.isEmpty() }
+
+    /**
+     * Merges this [ReceiptInfo] with another [ReceiptInfo], giving precedence to the values in this
+     * instance when there are conflicts.
+     */
+    fun mergeWith(receiptInfo: ReceiptInfo): ReceiptInfo {
```
JayShortway, line 95: 🫶

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -140,32 +136,40 @@ internal class PostReceiptHelper(
...
 
-        val presentedPaywall = paywallPresentedCache.getAndRemovePresentedEvent()
+        val presentedPaywall = if (cachedTransactionMetadata == null) {
+            paywallPresentedCache.getAndRemovePresentedEvent()
+        } else {
+            null
+        }
+        val effectivePaywallData = cachedTransactionMetadata?.paywallPostReceiptData
+            ?: presentedPaywall?.toPaywallPostReceiptData()
+        val effectiveReceiptInfo = cachedTransactionMetadata?.receiptInfo?.let { receiptInfo.mergeWith(it) }
```
ajpallares, line 161: I wonder, if the result of the `mergeWith` actually results in changes, shouldn't we update the cached metadata with the merged information? Otherwise, a failure in the POST /receipt could result in data loss?

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -246,11 +247,10 @@ internal class Backend(
         finishTransactions: Boolean,
         subscriberAttributes: Map<String, Map<String, Any?>>,
         receiptInfo: ReceiptInfo,
-        storeAppUserID: String?,
-        @SuppressWarnings("UnusedPrivateMember")
-        marketplace: String? = null,
         initiationSource: PostReceiptInitiationSource,
         paywallPostReceiptData: PaywallPostReceiptData?,
+        // This reflects the value at the time of the purchase, which might come from the LocalTransactionMetadataCache
```
ajpallares:
```suggestion
        // This reflects the value at the time of the purchase, which might come from the LocalTransactionMetadataStore
```

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Well, I think the merge is more of a formality to be honest, but usually the cached information has everything we really need, and we could ignore anything from the "non-cached" request, since it won't have new information. So I think there is no need to update the cache after it's stored the first time. Wdyt?

PR #2991, opened 2026-01-01
feat: migrate ConfigureScreen to Jetpack Compose in purchase tester module

nimi0112:
Migrate ConfigureScreen to Jetpack Compose

- Add Compose theme system (Color, Type, Theme)
- Create ConfigureActivity as Compose entry point
- Add Compose dependencies to build config
- Add test cases for ConfigureScreenViewModel

### Checklist

### Motivation

This PR modernizes the purchase-tester UI to Jetpack Compose for better maintainability and consistency with the paywall-tester module. This is the first step in migrating from Fragments/XML to Compose, which reduces boilerplate and improves the developer experience.

### Description

 • Migrated ConfigureFragment to ConfigureScreen composable with ViewModel and state management
  • Created Compose theme system (Color, Type, Theme) following Material3 patterns
  • Added ConfigureActivity as the Compose entry point
  • Updated build.gradle.kts with Compose dependencies (BOM, UI, Material3, Navigation Compose, etc.)
  • Added test cases for ConfigureScreenViewModel
  • Maintained feature parity with the original Fragment implementation

I have verified the behaviour by comparing it with the previous version of the app, especially around input validation and navigation flows. The functionality matches the existing implementation based on my testing.

I’ve also attached a comparison video (new vs. old) for easier review. Please let me know if you'd like me to validate additional edge cases or specific scenarios.

I have also worked on migrating other screens as well, but I wanted to get feedback on this screen first before proceeding further.

https://github.com/user-attachments/assets/acb7cbbb-3392-4f8c-a18f-d4ff97280a9d

### DIFF
--- a/examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ui/screens/configure/ConfigureScreen.kt
+++ b/examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ui/screens/configure/ConfigureScreen.kt
@@ -30,7 +30,6 @@ import androidx.compose.material3.TextFieldColors
 import androidx.compose.material3.TextFieldDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -42,12 +41,14 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.TextUnit
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.viewmodel.compose.viewModel
 import com.revenuecat.purchases.EntitlementVerificationMode
 import com.revenuecat.purchases.PurchasesAreCompletedBy
@@ -98,7 +99,7 @@ private fun ConfigureScreenContent(
         factory = ConfigureScreenViewModelImpl.Factory
     ),
 ) {
-    val viewModelState by configureScreenViewModel.state.collectAsState()
+    val viewModelState by configureScreenViewModel.state.collectAsStateWithLifecycle()
 
     var apiKey by remember { mutableStateOf("") }
     var proxyUrl by remember { mutableStateOf("") }
@@ -395,6 +396,7 @@ fun EntitlementVerificationDropdownMenu(
             value = selected.name,
             onValueChange = {},
             readOnly = true,
+            textStyle = TextStyle(fontSize = 14.sp),
             trailingIcon = { ExposedDropdownMenuDefaults.TrailingIcon(expanded = expanded) },
             colors = TransparentTextFieldColors,
             modifier = Modifier


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureActivity.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
+import androidx.compose.ui.Modifier
+import com.revenuecat.purchasetester.ui.screens.configure.ConfigureScreen
+import com.revenuecat.purchasetester.ui.theme.PurchaseTesterTheme
+
+class ConfigureActivity : ComponentActivity() {
```
JayShortway, line 14: Instead of adding a `ConfigureActivity`, could you make `ConfigureFragment` host the new composable `ConfigureScreen`? That way we can migrate each fragment separately without having to add activities.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ui/screens/configure/ConfigureScreenViewModel.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    private val _state: MutableStateFlow<ConfigureScreenState> = MutableStateFlow(ConfigureScreenState.Loading)
+    private val _events = MutableSharedFlow<ConfigureUiEvent>(
+        replay = 0,
+        extraBufferCapacity = 1
+    )
+
+    init {
+        loadData()
+    }
```
JayShortway: Ideally constructors do not have side effects. Could you try kicking off the initial data load in the flow instead? See here for more info: https://proandroiddev.com/loading-initial-data-in-launchedeffect-vs-viewmodel-f1747c20ce62

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureActivity.kt
nimi0112: Hi @JayShortway, thanks a lot for the review.
 I can consider the Fragment wrapper approach, but here is the current scenerio:
  Current approach:
  • ConfigureFragment has been completely removed and replaced with a pure Compose implementation
  • ConfigureActivity serves as the entry point (LAUNCHER intent) while other screens remain in MainActivity

  If we go with the Fragment wrapper approach:
  Creating a ConfigureFragment that hosts ComposeView would require:
  1. Recreating the Fragment I just deleted
  2. Adding temporary scaffolding (Fragment lifecycle wrapping Compose)
  3. A second cleanup phase to remove the wrapper after all screens are migrated

  This leads to unnecessary rework - writing code we know will be deleted soon.
  Migration plan:
  I'm migrating all 6 remaining fragments (Login, Overview, Offering, Logs, etc.) to Compose following the same pattern. Once complete, I'll consolidate into a single
  ComponentActivity with Compose Navigation and remove MainActivity entirely.

Does this approach work, or is there a strong preference for the incremental wrapper pattern?

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ui/screens/configure/ConfigureScreenViewModel.kt
nimi0112: resolved

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureActivity.kt
JayShortway: So what's the plan with `ConfigureActivity`? Wouldn't that need to be deleted too? 

In this in-between phase we'd have 2 launcher activities, if I understand you correctly. That's going to be confusing for whoever is using this app.

nimi0112: Hi @JayShortway , I think there might be some confusion about the architecture which I might not be able to explain clearly, my bad. Apologies for that. Let me try to clarify the intent and migration plan more explicitly.

**Current state:**
  - `ConfigureActivity` is now the LAUNCHER
  - `MainActivity` is NOT a launcher anymore (only accessible via navigation)

  **Migration path:**
  - `ConfigureActivity` stays and grows (adds more Compose screens)
  - `MainActivity` shrinks and gets deleted (as Fragment screens are migrated)

  **End state:**
  - A single ConfigureActivity hosting all Compose screens
  - Navigation handled entirely via Compose Navigation
  - MainActivity deleted entirely
  - `nav_graph.xml` deleted (replaced by Compose Navigation)

  **Why this approach:**
  - ConfigureActivity is permanent, not temporary
  - No need to recreate ConfigureFragment (which I just deleted)
  - No intermediate Fragment wrappers that would need cleanup later
  - Clean migration: Fragment screens move from MainActivity → ConfigureActivity one by one.

That said, this is just my proposed approach based on how I see the migration evolving. I’m completely open to feedback, and of course the final decision is yours. Happy to adjust this if you feel there’s a better or simpler direction to take.

PR #2992, opened 2026-01-01
[Samsung] Implement product changes

MarkVillacampa:
Added support for Galaxy Store product changes by introducing the GalaxyReplacementMode enum and a new PurchaseParam config method .galaxyReplacementMode().


purchases/src/main/kotlin/com/revenuecat/purchases/models/GalaxyReplacementMode.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.samsung.android.sdk.iap.lib.constants.HelperDefine
+/**
+ * Enum of possible replacement modes to be passed to a Samsung Galaxy Store subscription change.
+ * Ignored for Google Play and Amazon purchases.
+ *
+ * See https://developer.samsung.com/iap/subscription-guide/manage-subscription-plan/proration-modes.html
+ * for more details.
+ */
+enum class GalaxyReplacementMode(
+    val samsungProrationMode: HelperDefine.ProrationMode,
```
JayShortway: Could we make this internal, to keep Samsung's types out of our public API where possible?

purchases/src/main/kotlin/com/revenuecat/purchases/models/GalaxyReplacementMode.kt
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        override fun createFromParcel(`in`: Parcel): GalaxyReplacementMode? {
+            return `in`.readString()?.let { GalaxyReplacementMode.valueOf(it) }
+        }
+
+        override fun newArray(size: Int): Array<GalaxyReplacementMode?> {
+            return arrayOfNulls(size)
+        }
+    }
```
JayShortway: You should be able to annotate `GalaxyReplacementMode` with `@Parcelize`, and then you can remove all the manual parcelling code.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -131,12 +138,22 @@ class PurchaseParams(val builder: Builder) {
...
 
+        /*
+         * The [GalaxyReplacementMode] to use when replacing the given oldProductId. Defaults to
+         * [GalaxyReplacementMode.IMMEDIATE_WITHOUT_PRORATION].
+         *
+         * Only applied for Galaxy Store product changes. Ignored for Google Play and Amazon Appstore purchases.
+         */
+        fun galaxyReplacementMode(galaxyReplacementMode: GalaxyReplacementMode) = apply {
+            this.galaxyReplacementMode = galaxyReplacementMode
+        }
```
JayShortway, line 157: This is quite a leaky abstraction. Any chance we can unify these into a single `ReplacementMode`? 

I understand we want to move fast, so maybe now's not the time to do it. But we could buy ourselves some time if we annotate this with `@ExperimentalPreviewRevenueCatPurchasesAPI`.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
```diff
@@ -0,0 +1,145 @@
...
+import com.revenuecat.purchases.strings.PurchaseStrings
+import com.samsung.android.sdk.iap.lib.vo.ErrorVo
+import com.samsung.android.sdk.iap.lib.vo.PurchaseVo
+
+internal class ChangeSubscriptionPlanHandler(
+    private val iapHelper: IAPHelperProvider,
+) : ChangeSubscriptionPlanResponseListener {
+
+    @get:Synchronized
+    private var inFlightRequest: Request? = null
```
JayShortway, line 26: This only synchronizes the getter, but this field is set too. Maybe synchronization is not needed since this is only executed sequentially? But if we _are_ synchronizing I think we should synchronize both reads and writes, if only to avoid confusion.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    private data class Request(
+        val newProductId: String,
+        val oldProductId: String,
+        val onSuccess: (PurchaseVo) -> Unit,
+        val onError: (PurchasesError) -> Unit,
+    )
+
+    @OptIn(GalaxySerialOperation::class)
+    override fun changeSubscriptionPlan(
```
JayShortway: Shouldn't this propagate `@GalaxySerialOperation`? Maybe I don't fully understand how it works haha.


```suggestion
    @GalaxySerialOperation
    override fun changeSubscriptionPlan(
```

api-tester/src/main/java/com/revenuecat/apitester/kotlin/GalaxyReplacementModeAPI.kt
```diff
@@ -0,0 +1,23 @@
+package com.revenuecat.apitester.kotlin
+
+import com.revenuecat.purchases.models.GalaxyReplacementMode
+import com.samsung.android.sdk.iap.lib.constants.HelperDefine
+
+@Suppress("unused", "UNUSED_VARIABLE")
+private class GalaxyReplacementModeAPI {
```
JayShortway, line 8: I know tests are (probably) not working on the `samsung-dev` branch, but I think it would be good to run `scripts/api-dump.sh` (and commit) so we can inspect the actual public API surface.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
```diff
@@ -0,0 +1,145 @@
...
+            log(LogIntent.GALAXY_ERROR) { GalaxyStrings.CHANGE_SUBSCRIPTION_PLAN_NO_OLD_PRODUCT_ID }
+            val error = PurchasesError(
+                code = PurchasesErrorCode.PurchaseInvalidError,
+                underlyingErrorMessage = GalaxyStrings.CHANGE_SUBSCRIPTION_PLAN_NO_OLD_PRODUCT_ID,
+            )
+            onError(error)
+            return
+        }
+
+        this.inFlightRequest = Request(
```
JayShortway: Nit, consistency:
```suggestion
       inFlightRequest = Request(
```

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
```diff
@@ -0,0 +1,145 @@
...
+            )
+            val onError = inFlightRequest?.onError
+            clearInFlightRequest()
+            onError?.invoke(purchasesError)
+            return
+        }
+
+        val onSuccess = inFlightRequest?.onSuccess
+        clearInFlightRequest()
+        onSuccess?.invoke(purchase)
```
JayShortway, line 128: Should we verify that the `purchase` matches the `inFlightRequest` before calling `onSuccess`? (Also applies to `handleUnsuccessfulResponse`.)

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
MarkVillacampa: Agreed, I just followed what we do for Google with `googleReplacementMode`. One argument against a unified ReplacementMode API could be that a replacement mode in one platform might not necessarily translate 1:1 to a similar replacement mode in a different platform, apart from some replacement modes simply not being available in some platforms. We'd need to rely on documentation and runtime checks/errors while making this a bit more leaky allow for more explicit choices by developers. Gonna leave it as it is for now and we can revisit.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
MarkVillacampa: Hmm agreed. gonna let @fire-at-will chime in when he's back as this is copied over from other handlers.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
MarkVillacampa: That sounds right, we're doing that in all other handlers *except* the purchase handler, changed in both places

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
tonidero: I do think we should try to unify them, even with that potential drawback... As mentioned above, we could use `@ExperimentalPreviewRevenueCatPurchasesAPI` temporarily and/or create some tickets to make sure we revisit this before merging this into main.

If we did want to go this route, we probably also would want to add some runtime checks to make sure the correct replacement mode is used depending on the configured store.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
fire-at-will: @JayShortway  Great callout!! 

I think that some form of synchronization at this level of the code is still appropriate since it's _technically_ possible that the callbacks that are executed by the Samsung SDK could be on a different thread from the original calling thread. I think that a better approach would be to use either `AtomicReference` on the `inFlightRequest` or a dedicated lock for the `inFlightRequest`, so that we can protect both gets and sets on the `inFlightRequest`.

If it's alright with y'all, my vote would be to merge this into the dev branch as-is, and then tackle the synchronization for the `inFlightRequest`s in all of the handlers in it's own dedicated PR, as there are several of these that use this existing pattern, and I think it would be easier to review if we migrate them over together.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
fire-at-will: When comparing `GoogleReplacementMode`/`GalaxyReplacementMode`, it looks like the galaxy store has no equivalent to `GoogleReplacementMode.CHARGE_FULL_PRICE`. If we were to go down the route of unifying these, I could see a developer who uses `GoogleReplacementMode.CHARGE_FULL_PRICE` creating a Galaxy version of their app, not test it the product change flow, release it to the app store, and then face unexpected errors when the SDK's runtime checks/errors state that there's no equivalent for the Galaxy Store.

If we go towards a unified approach, we'll probably want to deprecate the existing `PurchaseParams.googleReplacementMode()` function in favor of the new, unified one.

Since this is what we've already shipped to a customer, I'll go ahead and mark it as `@ExperimentalPreviewRevenueCatPurchasesAPI` for now.

fire-at-will: Marked it as experimental in 0e2850191c427742b4dd8319fe9e81e268f1d778 👍

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
JayShortway: Tackling later works for me. 

Coincidentally I wrote [a comment](https://github.com/RevenueCat/purchases-android/pull/2988#discussion_r2664849219) today on a totally different PR that applies here too haha:

>This [`AtomicReference`] needs [core library desugaring](https://developer.android.com/studio/write/java8-support-table) afaik 🙁
>
>Maybe we can just use a good ol' lock?

Core library desugaring enables "newer" Java 8 APIs to be used, but we shouldn't enable that as a library. That's more of an app-level decision. So yea, I'd say we use a lock.

purchases/src/main/kotlin/com/revenuecat/purchases/models/GalaxyReplacementMode.kt
```diff
@@ -0,0 +1,65 @@
+package com.revenuecat.purchases.models
+
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.ReplacementMode
+import com.samsung.android.sdk.iap.lib.constants.HelperDefine
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Enum of possible replacement modes to be passed to a Samsung Galaxy Store subscription change.
+ * Ignored for Google Play and Amazon purchases.
```
fire-at-will:
```suggestion
 * Not used for Google Play and Amazon purchases.
```

fire-at-will (commented):
Looks good so far, I'll add tests this afternoon

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ChangeSubscriptionPlanHandler.kt
fire-at-will: Using a lock works for me, thanks for the insight, TIL! :D

PR #2993, opened 2026-01-05
Improve accuracy of transactions origin Part 4: Post remaining transaction metadata when syncing purchases

tonidero:
### Description
Follow-up on https://github.com/RevenueCat/purchases-android/pull/2989

This handles a possible edge case where a transaction is not returned by the store anymore. In this scenario, we still want to post any remaining cached metadata. This handles that by syncing all remaining metadata after any remaining purchases from the queue.


purchases/src/main/kotlin/com/revenuecat/purchases/PostPendingTransactionsHelper.kt
```diff
@@ -104,6 +145,68 @@ internal class PostPendingTransactionsHelper(
...
+        onSuccess: ((CustomerInfo) -> Unit),
+    ) {
+        val results: MutableList<Result<CustomerInfo, PurchasesError>> = mutableListOf()
+        val transactionMetadataToSync = localTransactionMetadataCache.getAllLocalTransactionMetadata()
+        if (transactionMetadataToSync.isEmpty()) {
+            onNoTransactionsToSync()
+            return
+        }
+        transactionMetadataToSync.forEach { transactionMetadata ->
+            // Cached paywall data is retrieved from the cache when posting the receipt.
```
rickvdl: @tonidero Could we also make a helper method on the postReceiptHelper that would just post everything from the stored transaction? That way it may be a bit more clear from the callsite

rickvdl (commented):
Great job!

purchases/src/main/kotlin/com/revenuecat/purchases/PostPendingTransactionsHelper.kt
tonidero: These are actually 2 different things:
- In the PostReceiptHelper, we are already trying to post a receipt, and we add the cached information to that request there, if any cached information can be found
- Here, we check if there are any remaining cached data that hasn't been synced before, (possibly because the transaction can't be found in the store anymore), we initiate the POST receipt for those here.

So they kinda follow different intentions. Not sure if that's clear or if I understood you correctly. Happy to chat if you prefer!

tonidero: Discussed over Slack and I know what you meant now (basically, we are accessing cached data in 2 places and then merging without much need and this could be improved). Will take a look and see how we could make this cleaner 👍

purchases/src/main/kotlin/com/revenuecat/purchases/PostPendingTransactionsHelper.kt
```diff
@@ -104,6 +145,68 @@ internal class PostPendingTransactionsHelper(
         }
     }
 
+    private fun postNotFoundTransactionMetadata(
```
JayShortway: "NotFound" might not be super clear to our future selves. Maybe `postRemainingTransactionMetadata`? (As in, post whatever remains locally.)

tonidero: Renamed in https://github.com/RevenueCat/purchases-android/pull/2993/commits/bbda1dbc2548425d3a9a20a853e30916e958a9ab

purchases/src/main/kotlin/com/revenuecat/purchases/PostPendingTransactionsHelper.kt
tonidero: Did the refactor in https://github.com/RevenueCat/purchases-android/pull/2993/commits/be50f92ae51ddb7202a45285d1ff75928d454cd4, but ended up as pretty much a different approach 😅

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -136,6 +139,73 @@ internal class PostReceiptHelper(
...
+                isRestore = allowSharingPlayStoreAccount,
+                receiptInfo = transactionMetadata.receiptInfo,
+                initiationSource = PostReceiptInitiationSource.UNSYNCED_ACTIVE_PURCHASES,
+                paywallData = transactionMetadata.paywallPostReceiptData,
+                purchasesAreCompletedBy = transactionMetadata.purchasesAreCompletedBy,
+                hasCachedTransactionMetadata = true,
+                paywallEvent = null,
+                onSuccess = {
+                    // This is safe since all requests are performed in the same dispatcher serial queue.
+                    results.add(Result.Success(it.customerInfo))
```
JayShortway: This encodes an assumption about an implementation that is true now, but might not be true forever. Maybe we can use `ConcurrentLinkedQueue` or `CopyOnWriteArrayList`? Seems like the former should suffice because we're only looping. 

If only we had more coroutines usage 😅

tonidero: Agreed it's pretty brittle. So yeah, your suggestion could indeed work! I will give that a shot 🙏

JayShortway: Nice!

JayShortway (approved):

PR #2994, opened 2026-01-05
Add Solar Engine integration support

limdauto:
## Summary

Add support for capturing three Solar Engine identifiers to enable RevenueCat Solar Engine integration:
- `$solarEngineDistinctId`
- `$solarEngineAccountId`
- `$solarEngineVisitorId`

## Implementation Details

This implementation follows the exact same pattern as the existing Airbridge device ID integration:

### API Changes
- **Public API**: Added three new methods to `Purchases` class:
  - `setSolarEngineDistinctId(String?)`
  - `setSolarEngineAccountId(String?)`
  - `setSolarEngineVisitorId(String?)`

### Internal Changes
- Added three new values to `ReservedSubscriberAttribute` enum
- Created three new objects in `SubscriberAttributeKey.AttributionIds` sealed class
- Added three new methods to `PurchasesOrchestrator`
- Each method triggers automatic device identifier collection (if enabled)
- Attributes sync automatically via existing infrastructure

### Testing
- Added three unit tests in `SubscriberAttributesPurchasesTests.kt`
- Tests follow the existing `attributionIDTest` pattern
- All tests verify correct manager method invocation
- **Added API verification calls in api-tester module** (both Kotlin and Java)
- Detekt style checks pass ✅

## Files Modified
- `purchases/src/main/kotlin/com/revenuecat/purchases/common/subscriberattributes/SpecialSubscriberAttributes.kt`
- `purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt`
- `purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt`
- `purchases/src/testDefaults/kotlin/com/revenuecat/purchases/attributes/SubscriberAttributesPurchasesTests.kt`
- `api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/PurchasesAPI.kt`
- `api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java`

## Usage Example

```kotlin
// Get IDs from Solar Engine SDK
val distinctId = SolarEngineManager.getInstance().getDistinctId()
val visitorId = SolarEngineManager.getInstance().getVisitorID()
val accountId = SolarEngineManager.getInstance().getAccountID()

// Pass to RevenueCat
Purchases.sharedInstance.setSolarEngineDistinctId(distinctId)
Purchases.sharedInstance.setSolarEngineVisitorId(visitorId)
Purchases.sharedInstance.setSolarEngineAccountId(accountId)
```

## Manual Testing
Manual testing can be performed using the `purchase-tester` app in the examples folder.

## Breaking Changes
None - all changes are additive and backward compatible.

## Related
- iOS implementation: RevenueCat/purchases-ios#5992
- Backend implementation: RevenueCat/khepri#16785


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -818,6 +818,36 @@ class Purchases internal constructor(
         purchasesOrchestrator.setAirbridgeDeviceID(airbridgeDeviceID)
     }
 
+    /**
+     * Subscriber attribute associated with the Solar Engine Distinct ID for the user
+     * Recommended for the RevenueCat Solar Engine integration
+     *
+     * @param solarEngineDistinctId null or an empty string will delete the subscriber attribute.
+     */
+    fun setSolarEngineDistinctId(solarEngineDistinctId: String?) {
```
tonidero, line 827: A couple questions here:
- Do we need all 3 attributes for the integration? And if so,
- Would it make sense to create a single `setSolarEngineIds` public API method that receive all 3? 

Mostly want to make sure we can't unify it into a single call, instead of 3 different calls to set everything.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -818,6 +818,36 @@ class Purchases internal constructor(
...
+        purchasesOrchestrator.setSolarEngineDistinctId(solarEngineDistinctId)
+    }
+
+    /**
+     * Subscriber attribute associated with the Solar Engine Account ID for the user
+     * Recommended for the RevenueCat Solar Engine integration
+     *
+     * @param solarEngineAccountId null or an empty string will delete the subscriber attribute.
+     */
+    fun setSolarEngineAccountId(solarEngineAccountId: String?) {
```
tonidero, line 837: Whichever API we end up with, we should probably add calls to the `api-tester` module, which verifies the API usage to avoid breaking things.

tonidero (commented):
It's looking great! mostly just a question on whether it would make sense to try to unify the 3 APIs into one somehow. It feels slightly weird to have to call 3 different methods, specially if we can gather that data synchronously in one go.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
limdauto: Hi @tonidero yes I thought about having 1 API but decided against it. The main issue is the intention is not for end user to call 3 different methods. Solar Engine gives developers a choice to call any of these 3 methods depending on how the app / their attribution is set up. This clear separation makes sure that developers know what they are doing if they want to call multiple identification methods by calling them explicitly.

tonidero: Ah then this does make sense indeed. Thanks for explaining!!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
limdauto: Added the api-tester calls. Could you help me check?

PR #2996, opened 2026-01-07
Add retry mechanism to purchases integration tests

tonidero:
### Description
There is some flakyness in our purchases integration tests that we want to address... But for now until we have more time, we can do a couple retries before notifying anything is wrong.


rickvdl (approved):
Thanks for the quick workaround for now

PR #3002, opened 2026-01-09
Improve accuracy of transactions origin Part 5: Cleanup

tonidero:
### Description
In this PR we will group some other smaller changes left over from previous PRs:
- Split storage to a separate SharedPreferences file
- Store transaction metadata separately, instead of on a single key
- Refactor to avoid possible race conditions when retrieving and caching transaction metadata


purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -261,13 +249,59 @@ internal class PostReceiptHelper(
...
-            hasCachedTransactionMetadata = cachedTransactionMetadata != null || shouldCacheTransactionMetadata,
+            hasCachedTransactionMetadata = cachedTransactionMetadata != null || didCacheData,
             paywallEvent = presentedPaywall,
             onSuccess = onSuccess,
             onError = onError,
         )
     }
 
+    @OptIn(InternalRevenueCatAPI::class)
+    private data class CachedDataToPost(
```
tonidero, line 263: This is a really bad name... but I was running out of ideas trying to come up with a good name... Basically this has the different pieces of data we need from orchestrating between the fetching and caching of data...

purchases/src/test/java/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStoreTest.kt
```diff
@@ -182,27 +190,46 @@ class LocalTransactionMetadataStoreTest {
...
+        val apiKeySlot = slot<String>()
+
+        every {
+            context.getSharedPreferences(capture(apiKeySlot), any())
+        } returns sharedPreferences
+
+        LocalTransactionMetadataStore.initializeSharedPreferences(context, apiKey)
+
+        assertThat(apiKeySlot.captured).isEqualTo("com.revenuecat.purchases.transaction_metadata.$apiKey")
+    }
```
JayShortway: The SharedPreferences file name is an implementation detail imo. What we really want to test is that the store keeps metadata for different API keys separate, right? 

We can test that more directly by creating 2 stores for different API keys, writing something to both, and making sure they did not overwrite/modify each other. That avoids the need for `@VisibleForTesting` too.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
```diff
@@ -9,106 +13,111 @@ import kotlinx.serialization.SerializationException
...
+    private val sharedPreferences: SharedPreferences = initializeSharedPreferences(context, apiKey),
     private val json: Json = JsonTools.json,
 ) {
-    private companion object Companion {
-        const val CACHE_KEY = "local_transaction_metadata"
+    internal companion object {
+        private const val KEY_PREFIX = "local_transaction_metadata_"
+
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal fun initializeSharedPreferences(context: Context, apiKey: String): SharedPreferences =
```
JayShortway: Just attaching a comment here too because I think we should avoid `@VisibleForTesting`. Suggested a way to do so in my previous comment.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -261,13 +249,59 @@ internal class PostReceiptHelper(
...
+            )
+            localTransactionMetadataStore.cacheLocalTransactionMetadata(purchaseToken, dataToCache)
+        }
+
+        return CachedDataToPost(
+            localTransactionMetadata = cachedTransactionMetadata,
+            paywallEvent = presentedPaywall,
+            didCacheData = shouldCacheTransactionMetadata,
+        )
+    }
```
JayShortway, line 307: This one is a bit hard to understand. 

What's confusing is that it returns the previously `cachedTransactionMetadata`, even if we're caching new data (`shouldCacheTransactionMetadata` being true).  

I feel like it would help if we inlined `shouldCacheTransactionMetadata`, so just writing: 
```kotlin
cachedTransactionMetadata == null && source == PostReceiptInitiationSource.PURCHASE
```
Otherwise readers have to skip back and forth between this function and `shouldCacheTransactionMetadata`.

Also I think a better name would be `getOrPersistDataToPost()` ("Or" already indicates it might not persist, so we can omit "IfNeeded" imo.) Even clearer would be `LocalTransactionMetadataStore.getOrPutDataToPost()`. If `LocalTransactionMetadataStore` is the receiver, it's very clear _where_ we're putting that data or getting it from.

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
JayShortway: I think it's pretty clear already haha. Maybe `TransactionMetadataToPost`?

purchases/src/test/java/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStoreTest.kt
tonidero: This is indeed a good point... I think this test is pretty much useless. I based it on how we did it in diagnostics, but indeed, not needed.

Regarding your suggestion, in these tests, we're not actually creating shared preferences files (the shared preferences are mocked), so it would be more difficult to test that scenario I think. We could try to rewrite them to use actual shared preferences files considering we are using robolectric anyway, and we're mocking too much stuff already... but for now, I think I might just remove this test instead to make the initializeSharedPreferences method private. Lmk if you think it's important we test this scenario though!

tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3002/commits/ddb5411a062859277a4637d71141743d158fa9a9

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStore.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3002/commits/ddb5411a062859277a4637d71141743d158fa9a9

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: I do like that, thanks for the suggestions!! 🫶 Applied in https://github.com/RevenueCat/purchases-android/pull/3002/commits/2d5218f9093a4f88167eac9aa1dafd72f0a04ac3

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
tonidero: Hmm I kinda prefer the current name since it indicates it's the "cached data", not the current data we're trying to post... but honestly not a strong opinion. Will leave as it is for now, if it works for you 🙏

purchases/src/test/java/com/revenuecat/purchases/common/caching/LocalTransactionMetadataStoreTest.kt
JayShortway: Ahh got it. Not for this PR, but it would be useful to have a `FakeSharedPreferences` which keeps everything in memory. (`SharedPreferences` is an interface.)

Thanks for the cleanup!

purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
JayShortway: Awesomee

Issue #3001, opened 2026-01-09
Risk of non-consumables being consumed unintentionally


appetizermonster:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
I'm in the process of figuring out whether it's good to migrate my existing app, which has lifetime non-consumable products, to use RevenueCat.

I've seen a few issues related to RevenueCat consumed non-consumable products.
(https://github.com/RevenueCat/purchases-android/issues/139, https://github.com/RevenueCat/purchases-android/issues/2566)

So I looked into the SDK code, and I found that the RevenueCat SDK has some potentially disruptive behavior when migrating existing lifetime products. It might unintentionally consume non-consumables when a user launches a new version of the app that uses RevenueCat (via `syncPendingPurchaseQueue`, or when the user calls `restorePurchases`).

In [PostReceiptHelper.kr](https://github.com/RevenueCat/purchases-android/blob/d85feecd8e65dfddb74a9f3948e1b2e4fa685a64/purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt#L112), when the backend response is missing `purchased_products` or the product id doesn't match, shouldConsume defaults to **true**, causing any one-time product to be consumed.

I know the chance is low, but i don't think it's zero. For example, if the RevenueCat backend returns an incorrect response, such as missing the `purchased_products` key due to a backend or backend developer error, even when the product is configured as non-consumable in the RevenueCat dashboard.

Since consumption is irreversible, I think defaulting to false would be safer. I know this has been `true` for years, so changing it may not be trivial, but as a developer this still feels like a code smell.

1. Environment
   1. Platform:
   2. SDK version:
   3. OS version:
   4. Android Studio version:
   5. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

PR #3003, opened 2026-01-09
Avoids consuming purchases by default

JayShortway:
## Description
In case a backend bug would respond incomplete product data upon POST /receipt, we would default to consume. This PR switches that to default to not consume, which is less destructive. (To be clear, such a backend bug has never happened, but better safe than sorry.)

Resolves https://github.com/RevenueCat/purchases-android/issues/3001

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
@@ -114,7 +114,7 @@ internal class PostReceiptHelper(
                     ?.values
                     ?.firstOrNull()
                     ?.shouldConsume
-                    ?: true
+                    ?: false
                 billing.consumeAndSave(finishTransactions, purchase, shouldConsume, initiationSource)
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -114,7 +114,7 @@ internal class PostReceiptHelper(
                     ?.values
                     ?.firstOrNull()
                     ?.shouldConsume
-                    ?: true
+                    ?: false
```
tonidero, line 117: One thing that I think could be important... In this scenario, where it gets to this default, we shouldn't add the token to the cache of successfully posted tokens so we retry to post it on next app foreground. This way, if it was meant to be a consumable, we wouldn't have the issue of it not being able to be repurchased much longer. Wdyt?

JayShortway: Created [SDK-4270](https://linear.app/revenuecat/issue/SDK-4270) to track it, so we can merge this.

PR #3005, opened 2026-01-12
Refactor Module Structure + Get CI to Work

fire-at-will:
## Description
Alright, we're finally taking on fixing things in CI for the Galaxy Store! This is a whopper of a PR. Despite the large file count/diff size, we can distill what's going on in this PR into a few main categories:

### Moving the Galaxy Files to Their Own Module
To maintain the SDK's compatibility with the Samsung IAP's minSDK requirement of 23 and to ensure that SDK developers only need to supply the Samsung SDK when editing the Galaxy files, we moved almost all of the Galaxy files in the SDK to the `feature/galaxy` module. This module:
- Has a minSDK version of 23 (compared to the main SDK's requirement of 21)
- Is only compiled when the file `samsung-iap-{VERSION_NUMBER}.aar` is in the root directory's `libs` folder

To access certain classes in the main `purchases` module, like `BillingAbstract`, a new `@InternalRevenueCatStoreAPI` annotation is included to make the classes public, but not really usable to consumers of the SDK. The following classes are annotated with `@InternalRevenueCatStoreAPI`:
- StoreProductsCallback
- PurchasesStateProvider
- PurchasesState
- DeviceCache
- DateProvider
- DefaultDateProvider
- ProductEntitlementMapping
- HTTPResponseOriginalSource
- HTTPResult
- PurchaseStrings
- RestoreStrings
- BillingAbstract
- PostReceiptInitiationSource
- ReplaceProductInfo
- SerialRequestExecutor & SerialRequest
- PresentedOfferingContext.copy
- Config
- utils.sha1()
- utils.sha256()
- Emojis
The `logUtils.kt` and `logWrapper.kt` files are copy/pasted to the Galaxy module to avoid the annotation from spreading even more throughout the `purchases` module.

Then, to access the `GalaxyBillingWrapper` class from the `purchases` module, a new `GalaxyBillingWrapperFactory` object is introduced that uses reflection to instantiate GalaxyBillingWrapper.

### Moved the GalaxyReplacementMode -> Samsung Class Conversions out of the GalaxyReplacementMode class
Moved the GalaxyReplacementMode -> Samsung class conversion logic from the `GalaxyReplacementMode` class, which needed to stay in the `purchases` module, to standalone conversion classes in the `galaxy` module to remove the dependency on the Samsung SDK from `GalaxyReplacementMode`.

### Galaxy Module Conditional Compilation
When compiling the SDK, the galaxy module's compilation (and the compilation of the Galaxy API tester files) is now conditional based on the presence of the Samsung IAP SDK in the root `libs` directory. This prevents SDK contributors not working on the galaxy module from having to supply the Samsung IAP SDK to compile the purchases SDK

### CI Changes
- Added `getSamsungIapSdk` task to the main `build.gradle` to pull the Samsung IAP SDK into the CI environment. It pulls the SDK URL from a CI environment variable.
- Add a new `test-galaxy` CI job that get executed in the `build-test-deploy` pipeline. This allows us to run the Galaxy unit & API tests in these jobs with the Samsung IAP SDK, but exclude them in the other existing jobs. This allows us to confirm that the SDK still works as expected without the Samsung IAP SDK/galaxy module included

### API Tester Changes
Moved the Galaxy API tester files to a new directory in the api-tester module that only gets compiled when the Samsung IAP SDK is present in the root directory's `libs` folder

### Installation Instructions
For SDK contributors:
- If they want to work outside of the `feature:galaxy` module, no additional steps are required
- If they want to work in the `feature:galaxy` module, they need to include the Samsung IAP SDK AAR file in the project's `libs` directory

For app developers:
- If they don't want to use the Galaxy Store, they don't need to make any changes
- If they want to use the Galaxy store, they'll need to include the Samsung IAP SDK AAR and the `purchases-galaxy` modules in their dependenies:

```
// build.gradle
dependencies {
    implementation("com.revenuecat.purchases:purchases:{SDK_VERSION}")
    implementation("com.revenuecat.purchases:purchases-galaxy:{SDK_VERSION}")
    implementation(files("../libs/samsung-iap-6.5.0.aar"))
}
```

### Testing Performed
- Built the SDK without the Samsung IAP SDK in the `libs` folder, causing the galaxy module to not be compiled
- Built the SDK with the Samsung IAP SDK in the `libs` folder, causing the galaxy module to be compiled
- Installed the SDK to maven local, and in a separate, standalone app:
   - Built the app with the purchases SDK and no galaxy module or Samsung IAP SDK. The Galaxy classes in the galaxy module were inaccessible.
   -  Built the app with the purchases SDK, the galaxy module, and Samsung IAP SDK. The Galaxy classes in the galaxy module were accessible.

### TODO
- Write new installation instructions
- Move Samsung SDK location from Google Drive to S3, put it in a password-protected ZIP file, and unzip it in the `getSamsungIapSdk` gradle task

### DIFF
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
@@ -34,6 +34,7 @@ import com.revenuecat.purchases.galaxy.listener.PurchaseResponseListener
 import com.revenuecat.purchases.galaxy.logging.LogIntent
 import com.revenuecat.purchases.galaxy.logging.log
 import com.revenuecat.purchases.galaxy.utils.GalaxySerialOperation
+import com.revenuecat.purchases.galaxy.utils.SerialRequestExecutor
 import com.revenuecat.purchases.galaxy.utils.parseDateFromGalaxyDateString
 import com.revenuecat.purchases.models.GalaxyReplacementMode
 import com.revenuecat.purchases.models.InAppMessageType
@@ -42,7 +43,6 @@ import com.revenuecat.purchases.models.PurchasingData
 import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.strings.PurchaseStrings
 import com.revenuecat.purchases.strings.RestoreStrings
-import com.revenuecat.purchases.utils.SerialRequestExecutor
 import com.samsung.android.sdk.iap.lib.constants.HelperDefine
 import com.samsung.android.sdk.iap.lib.helper.IapHelper
 import com.samsung.android.sdk.iap.lib.vo.PurchaseVo
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/GalaxySerialOperation.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/GalaxySerialOperation.kt
@@ -11,7 +11,7 @@ package com.revenuecat.purchases.galaxy.utils
  * Functions marked with `@GalaxySerialOperation` **must** be executed one at a time,
  * in strict sequence, without concurrent execution. This is usually managed by ensuring the
  * execution path for these functions goes through a mechanism like
- * the [com.revenuecat.purchases.utils.SerialRequestExecutor]
+ * the [SerialRequestExecutor]
  * or a single-threaded dispatcher (e.g., a single-threaded Coroutine scope).
  *
  * Failing to execute these functions serially can lead to the `IapHelper` throwing
@@ -21,7 +21,7 @@ package com.revenuecat.purchases.galaxy.utils
  * Apply `@OptIn(GalaxySerialOperation::class)` to the calling function. This should only be done if the calling
  * scope *guarantees* serial execution for the call.
  *
- * @see [com.revenuecat.purchases.utils.SerialRequestExecutor] The underlying mechanism typically used to enforce
+ * @see [SerialRequestExecutor] The underlying mechanism typically used to enforce
  * this constraint.
  */
 @Retention(value = AnnotationRetention.BINARY)
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/SerialRequestExecutor.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/utils/SerialRequestExecutor.kt
@@ -1,4 +1,4 @@
-package com.revenuecat.purchases.utils
+package com.revenuecat.purchases.galaxy.utils
 
 import java.util.ArrayDeque
 import java.util.concurrent.atomic.AtomicBoolean
--- a/feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/utils/SerialRequestExecutorTest.kt
+++ b/feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/utils/SerialRequestExecutorTest.kt
@@ -1,6 +1,5 @@
 package com.revenuecat.purchases.galaxy.utils
 
-import com.revenuecat.purchases.utils.SerialRequestExecutor
 import org.assertj.core.api.Assertions
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.CountDownLatch


feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/LogUtils.kt
```diff
@@ -0,0 +1,90 @@
...
+        LogLevel.DEBUG
+    } else {
+        LogLevel.INFO
+    }
+}
+
+internal inline fun verboseLog(messageBuilder: () -> String) {
+    logIfEnabled(LogLevel.VERBOSE, currentLogHandler::v, messageBuilder)
+}
+
```
skydoves, line 28: Don't we need to add `@JvmSynthetic` as with our other internal functions?

.circleci/config.yml
```diff
@@ -310,6 +317,37 @@ jobs:
...
+          command: |
+            ./gradlew :getSamsungIapSdk --no-daemon
+      - run:
+          name: Compile API Tester with Galaxy Files
+          command: |
+            ./gradlew :api-tester:compileDefaultsDebugKotlin --no-daemon
+      - run:
+          name: Run Galaxy Unit Tests
+          command: |
+            ./gradlew testDefaultsBc8DebugUnitTest --parallel --no-daemon
```
tonidero, line 332: Hmm I guess not opposed to running all tests... but should we run only the tests for the `feature/galaxy` module?

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/README.md
```diff
@@ -0,0 +1,3 @@
+This directory contains copies of the logging utilities from the purchases module. They are copied here instead
+of imported to prevent the `@InternalRevenueCatStoreAPI` annotation from leaking to the vast majority of files/
+functions in the purchases module.
```
tonidero: Hmm the duplication is not great... but I guess we can do it to unblock... Might be worth moving the logging code to a common separate module soon after this PR though, so we can call it from our current modules, without actually duplicating the code.

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
```diff
@@ -10,7 +11,7 @@ import com.revenuecat.purchases.models.SubscriptionOptions
 import dev.drewhamilton.poko.Poko
 
 @Poko
-class GalaxyStoreProduct(
+internal class GalaxyStoreProduct(
```
tonidero: Hmm I'm a bit confused about these being made internal... I think these are expected to be kept public right? (same as GoogleStoreProduct/GoogleSubscriptionOption/GooglePurchasingData/...)

feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/constants/GalaxyConsumeOrAcknowledgeStatusCodeTest.kt
```diff
@@ -4,6 +4,7 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
 import org.junit.runner.RunWith
+import kotlin.code
```
tonidero: Hmm why this import? I see it in a few other places too

feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/utils/ErrorVoUtilsTest.kt
```diff
@@ -67,7 +69,7 @@ class ErrorVoUtilsTest {
             assertThat(purchasesError)
                 .describedAs("Galaxy code ${galaxyCode.name} should map to $purchasesErrorCode")
                 .isNotNull
-            assertThat(purchasesError!!.code).isEqualTo(purchasesErrorCode)
+            Assertions.assertThat(purchasesError!!.code).isEqualTo(purchasesErrorCode)
```
tonidero: Hmm might be nicer to revert this change? Same in some other changes in these test files.

feature/galaxy/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
-
 plugins {
     id("revenuecat-public-library")
+    alias(libs.plugins.kotlin.parcelize)
+    alias(libs.plugins.poko)
```
tonidero, line 4: We probably would like to enable metalava for this module, since there will be new public APIs exposed by it (like GalaxyStoreProduct/GalaxySubscriptionOption/...). This could help us make sure we only expose what we want.

purchases/src/main/kotlin/com/revenuecat/purchases/models/GalaxyReplacementMode.kt
```diff
@@ -53,13 +51,14 @@ enum class GalaxyReplacementMode(
...
+    DEFERRED,
     ;
 
     companion object {
         /**
          * The default replacement mode for Galaxy Store subscription changes.
          */
-        internal val default: GalaxyReplacementMode = INSTANT_NO_PRORATION
+        @InternalRevenueCatStoreAPI
+        val default: GalaxyReplacementMode = INSTANT_NO_PRORATION
```
tonidero, line 62: Hmm I guess I wouldn't be totally opposed to exposing this... But I'm ok keeping this internal for now

purchases/src/main/kotlin/com/revenuecat/purchases/strings/PurchaseStrings.kt
```diff
@@ -1,6 +1,9 @@
 package com.revenuecat.purchases.strings
 
-internal object PurchaseStrings {
+import com.revenuecat.purchases.InternalRevenueCatStoreAPI
+
+@InternalRevenueCatStoreAPI
+object PurchaseStrings {
```
tonidero, line 6: Hmm wonder if we should duplicate any common usages into `GalaxyStrings`, instead of exposing these... But well, we can revisit this later if needed.

purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
```diff
@@ -43,6 +44,7 @@ import kotlin.time.Duration.Companion.milliseconds
 
 @RunWith(AndroidJUnit4::class)
 @Config(manifest = Config.NONE)
+@OptIn(InternalRevenueCatStoreAPI::class)
```
tonidero: I'm thinking, could be worth adding a global opt in to this module for this new annotation. That would save us many of these OptIn annotations I think.

purchases/src/test/java/com/revenuecat/purchases/galaxy/BillingFactoryGalaxyTest.kt
```diff
@@ -16,52 +16,52 @@ import org.junit.runner.RunWith
...
-            pendingTransactionsForPrepaidPlansEnabled = true,
-            GalaxyBillingMode.TEST,
-            backend = mockBackend,
-        )
-        assertIs<GalaxyBillingWrapper>(galaxyBillingWrapper)
-        assertEquals(GalaxyBillingMode.TEST, galaxyBillingWrapper.billingMode)
-    }
-}
+//@RunWith(AndroidJUnit4::class)
+//class BillingFactoryGalaxyTest {
```
tonidero: Looks like we need to uncomment this?

settings.gradle.kts
```diff
@@ -26,14 +28,43 @@ pluginManagement {
...
-dependencyResolutionManagement {
-    repositoriesMode.set(RepositoriesMode.FAIL_ON_PROJECT_REPOS)
+val samsungIapSdkDir = file("$rootDir/libs")
+
+/**
+ * Returns true only when the expected Samsung IAP AAR (versioned from
+ * `gradle/libs.versions.toml`) is present in the SDK directory.
+ */
+@Suppress("ReturnCount")
+fun Settings.isSamsungIAPAARPresent(samsungIAPSDKDir: File): Boolean {
```
tonidero: This seems like it could be private?
```suggestion
private fun Settings.isSamsungIAPAARPresent(samsungIAPSDKDir: File): Boolean {
```

settings.gradle.kts
```diff
@@ -26,14 +28,43 @@ pluginManagement {
...
+    val samsungIapAar = samsungIAPSDKDir.resolve("samsung-iap-$samsungIapVersion.aar")
+    return samsungIapAar.isFile
+}
+
+/**
+ * Reads a version entry from `gradle/libs.versions.toml` without relying on the
+ * version catalog extension (not available during settings evaluation).
+ */
+@Suppress("ReturnCount")
+fun Settings.readVersionFromCatalog(versionKey: String): String? {
```
tonidero: Same here about being private
```suggestion
private fun Settings.readVersionFromCatalog(versionKey: String): String? {
```

build.gradle.kts
```diff
@@ -35,6 +38,50 @@ tasks.register<Delete>("clean") {
     delete(rootProject.buildDir)
 }
 
+val samsungIapVersion = libs.versions.samsungIap.get()
+val samsungIapFileName = "samsung-iap-$samsungIapVersion.aar"
+val samsungIapDestFile = rootProject.file("libs/$samsungIapFileName")
+
+tasks.register("getSamsungIapSdk") {
+    val downloadUrl = System.getenv("SAMSUNG_IAP_SDK_URL").orEmpty()
```
tonidero, line 46: Hmm could be worth obtaining this from `local.properties` so it's easier to work with locally, and then getting from an ENV could be a fallback. Not a blocker though.

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/LogUtils.kt
fire-at-will: I copy/pasted this file from the `logUtils.kt` file in the main `purchases` module, so I guess not? 😅 Is there any reason we'd want to hide this code from Java and keep it exclusive to Kotlin?

tonidero: The main problem is that `internal` kotlin visibility is actually visible in Java 😨 (they do get a linter warning, but they can compile. So we did mention to make all internal functions `@JvmSynthetic` to avoid this (right now these would be accesible by Java in the current location 😓. 

I guess since we are adding these as new, it might be good to add that annotation, but I don't feel too strongly about it.

.circleci/config.yml
fire-at-will: I thought it was the safe thing to do, just in case the presence of the galaxy module and/or samsung IAP SDK somehow interfered with how something else in the SDK was working, we'd want to know about it. I can't think of a concrete example, but it might help us catch something one day 🤷‍♂️

fire-at-will: If we want to only run the galaxy unit tests, we can do so with:

```suggestion
            ./gradlew :feature:galaxy:testDefaultsBc8DebugUnitTest --parallel --no-daemon
```

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/README.md
fire-at-will: Yeah, the duplication is not great. If we moved it to a common separate logging module, we would still need to use the `@InternalRevenueCatAPI` annotation, and it would then spread heavily throughout the main purchases module, which was the main reason to copy paste it here. 

If we want to do that or just make the existing one the purchases module public, I'm ok doing that in a follow up PR! I'm hesitant to make those changes in this PR to keep this PR's size from getting any larger

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
fire-at-will: Right...I made this internal because the `GalaxyPurchasingData` class was internal. However, looking at the other purchasing data classes, they're public too, so I made `GalaxyPurchasingData` and `GalaxyStoreProduct` public again in f4d204ba2849ef40319bbf6e69cb6cc19473bdb5 👍

feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/constants/GalaxyConsumeOrAcknowledgeStatusCodeTest.kt
fire-at-will: That's a great question. I'm not sure what that import does or why it was added, but they are all unused. Removed in 2a843be2348f7092e3315d8cf04a41fdb4d88a45

.circleci/config.yml
tonidero: Yeah, I think it's fine to run all of them... In general, I don't think the behavior in the other modules should change at all with the presence of the galaxy aar, only the feature/galaxy module should be affected ideally. But yeah, not a strong opinion if we want to run all the tests again.

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/README.md
tonidero: > If we moved it to a common separate logging module, we would still need to use the @InternalRevenueCatAPI annotation

Not really. This can be an internal module, that is not published to consumers, so it can just have everything as public. But again, this is probably a bigger refactor, so I'm ok leaving this for other PRs

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyStoreProduct.kt
tonidero: I think we're still missing making public the `GalaxySubscriptionOption`?

fire-at-will: Right, made it public in ac47b75886a245e90291a2136d22e323bcfdacab 👍

feature/galaxy/src/test/java/com/revenuecat/purchases/galaxy/utils/ErrorVoUtilsTest.kt
fire-at-will: Reverted in d19b9d38842d28f406058ba3b53c4c6c7a10874e 👍

feature/galaxy/build.gradle.kts
fire-at-will: I think that metalava is already enabled here transitively through the `revenuecat-public-libary` plugin. Check out `PublicLibraryConventionPlugin.kt`, which does the following:
```kotlin
override fun apply(target: Project) = with(target) {
        with(pluginManager) {
            apply(libs.plugins.android.library.get().pluginId)
            apply(libs.plugins.kotlin.android.get().pluginId)
            apply(libs.plugins.kover.get().pluginId)
            apply(libs.plugins.dokka.get().pluginId)
            apply(libs.plugins.baselineprofile.get().pluginId)
        }

        configureAndroidLibrary()
        configureApisFlavors()
        configureConditionalPublishing()
        configureMetalava()
    }
```

Additionally, it looks like `feature/galaxy/api.txt` is being updated. Is there anything else we should do to fully enable metalava? I'm definitely not an expert here, so could be missing something

purchases/src/main/kotlin/com/revenuecat/purchases/models/GalaxyReplacementMode.kt
fire-at-will: The reason we're making it public here is that `GalaxyReplacementMode` needs to live in the `purchases` module so it can be added to `PurchaseParams`. However, we refer to `GalaxyReplacementMode.default` in `GalaxyBillingWrapper` (in the galaxy module) on line 329 to determine which replacement mode to use in a product change if the developer doesn't provide a replacement mode

We could move the default value to the galaxy module if we wanted to, but I liked the idea of keeping it in the same place as the enum definition

purchases/src/test/java/com/revenuecat/purchases/common/diagnostics/DiagnosticsTrackerTest.kt
fire-at-will: TIL about `freeCompilerArgs` - that definitely helps simplify things! In 8ffa08cacd7c03252f0f8dbfd6886c188bec690c, I opted-in the entire `purchases` module (src + tests) into the annotation like this:
```kotlin
tasks.withType<KotlinCompilationTask<*>>().configureEach {
    compilerOptions {
        freeCompilerArgs.add("-Xjvm-default=all-compatibility")
        freeCompilerArgs.add("-opt-in=com.revenuecat.purchases.InternalRevenueCatStoreAPI")
    }

    if (name.contains("UnitTest") || name.contains("AndroidTest")) {
        compilerOptions {
            freeCompilerArgs.add("-opt-in=com.revenuecat.purchases.InternalRevenueCatAPI")
            freeCompilerArgs.add("-opt-in=com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI")
        }
    }
}
```
and removed the manual opt-ins throughout the purchases module.

This decreased the number of files touched in this PR by 65 :D

purchases/src/test/java/com/revenuecat/purchases/galaxy/BillingFactoryGalaxyTest.kt
fire-at-will: Right. I've removed it in 2c602367b8248e3b2c4322a3623191f940ef80e1 due to a conundrum:

The `BillingFactoryGalaxyTest` file needs to live in the purchases module to access `BillingFactory` without making it public (it really doesn't need to be public). However, in that module, it can't access the `GalaxyBillingWrapper` class, and even if it could through reflection, the tests would fail in CI since the galaxy module won't compile without the Samsung SDK AAR 🤪

We could keep the tests by making `BillingFactory` public/InternalRevenueCatStoreAPI`, and moving the tests to the Galaxy module, but that feels a little overkill. Let me know what you think!

build.gradle.kts
fire-at-will: So this task should only ever run in CI. Therefore, I feel like it's better and more lightweight to just get the value here from the CI environment instead of injecting the value into a local.properties file in CI. 

Developers working on the SDK should obtain a copy of the Samsung SDK on their own and provide it themselves instead of using this task, which will pull the Samsung AAR from a location that we host.

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/LogUtils.kt
fire-at-will: > The main problem is that internal kotlin visibility is actually visible in Java 😨

Oh...my. Are we trying to make all internal functions `@JvmSynthetic`, or just "dangling" functions like this that aren't a part of a class/object?

api-tester/src/main/java/com/revenuecat/apitester/kotlin/GalaxyReplacementModeAPI.kt
```diff
@@ -1,12 +1,12 @@
...
 
 import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.InternalRevenueCatStoreAPI
 import com.revenuecat.purchases.models.GalaxyReplacementMode
-import com.samsung.android.sdk.iap.lib.constants.HelperDefine
 
 @Suppress("unused", "UNUSED_VARIABLE")
 private class GalaxyReplacementModeAPI {
-    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatStoreAPI::class)
```
JayShortway: My 2 cents on the new annotation: I’m afraid using a separate `@InternalRevenueCatStoreAPI` makes things more messy instead of less. Chances are that this annotation lives longer than we’d like, and before you know it someone (accidentally) uses it for a new thing that it wasn’t meant for.

I would put the list of symbols in a Linear ticket and keep track of it that way. Not the biggest deal, but just the way I’m thinking about it.

fire-at-will:
@JayShortway 

> Maybe this should target main to avoid a nightmare merge down the line?

I actually think we're fine here for now, so long as another huge SDK refactor doesn't make it in soon. This PR is going into the `samsung-dev` branch, which I've been keeping up to date with main. In fact, commit 3108c48c3aaa5c2a7a837a1ca3fff28520526fd2 from a few minutes ago actually brings the latest main from today into this branch, so I _think_ we're fine for now

api-tester/src/main/java/com/revenuecat/apitester/kotlin/GalaxyReplacementModeAPI.kt
fire-at-will: I'm okay with that! The likelihood of this happening seems more likely than not:

> Chances are that this annotation lives longer than we’d like, and before you know it someone (accidentally) uses it for a new thing that it wasn’t meant for.

I'll replace the `@InternalRevenueCatStoreAPI` usages with `@InternalRevenueCatAPI` 👍

fire-at-will: Replaced in `@InternalRevenueCatStoreAPI` usages with `@InternalRevenueCatAPI` 524bd3b5bd00870e90507087163808887b0c8fb8 👍 

Note that since we automatically applied `@InternalRevenueCatStoreAPI` to the entire `purchases` module with `freeCompilerArgs`, we're doing that for `@InternalRevenueCatAPI` now :)

JayShortway: >Replaced in @InternalRevenueCatStoreAPI usages with @InternalRevenueCatAPI https://github.com/RevenueCat/purchases-android/commit/524bd3b5bd00870e90507087163808887b0c8fb8 👍

Nice!

>Note that since we automatically applied @InternalRevenueCatStoreAPI to the entire purchases module with freeCompilerArgs, we're doing that for @InternalRevenueCatAPI now :)

I _think_ that's fine. I intentionally didn't add that when introducing the annotation so we would be forced to decide between opting in and propagating every time we'd use an `@InternalRevenueCatAPI`. Now we're implicitly always opting in, while sometimes we might wanna propagate.

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperFactory.kt
```diff
@@ -0,0 +1,42 @@
...
+                context,
+                billingMode,
+                deviceCache,
+            ) as BillingAbstract
+        } catch (e: ClassNotFoundException) {
+            val error = NoClassDefFoundError(e.message)
+            error.initCause(e)
+            throw error
+        } catch (e: NoSuchMethodException) {
+            throw IllegalStateException("Failed to find GalaxyBillingWrapper constructor", e)
```
tonidero: Should we give instructions in this case that devs need to make sure they've added the galaxy library to their app?

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyDeviceIdentifiersFetcher.kt
```diff
@@ -1,4 +1,4 @@
-package com.revenuecat.purchases.galaxy.attribution
+package com.revenuecat.purchases.galaxy
```
tonidero: Nitpicky but could be worth leaving it in an `attribution` package, for consistency with the core SDK.

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/AcknowledgePurchaseUseCase.kt
```diff
@@ -10,7 +10,8 @@ import com.revenuecat.purchases.google.billingResponseToPurchasesError
 import com.revenuecat.purchases.google.toHumanReadableDescription
 import com.revenuecat.purchases.strings.PurchaseStrings
 
-internal class AcknowledgePurchaseUseCaseParams(
+internal class AcknowledgePurchaseUseCaseParams
+constructor(
```
tonidero, line 16: Very nitpicky but should we revert all these changes? Seem they are not needed anymore, and while it doesn't hurt, it can keep the history clean for these files, I think.

purchases/src/test/java/com/revenuecat/purchases/common/caching/DateExtensionsTest.kt
```diff
@@ -12,6 +13,7 @@ import kotlin.time.Duration.Companion.minutes
 
 @RunWith(AndroidJUnit4::class)
 @Config(manifest = Config.NONE)
+
```
tonidero: Many of these files are adding an empty space here which feels weird. We probably should revert those small changes as well... Usually these should be caught by detekt, but we're not running it on tests on the purchases module :(

purchases/src/test/java/com/revenuecat/purchases/common/offlineentitlements/productEntitlementMappingFactory.kt
```diff
@@ -1,7 +1,9 @@
 package com.revenuecat.purchases.common.offlineentitlements
 
+import com.revenuecat.purchases.InternalRevenueCatAPI
```
tonidero: Some of these also seem to have unused imports now?

fire-at-will: Hmmmm...it's quite odd that detekt didn't catch these. Will remove!

api-tester/src/main/java/com/revenuecat/apitester/kotlin/GalaxyReplacementModeAPI.kt
fire-at-will: Right, there could be instances where we want to propagate it. To keep that ability, I can keep the auto opt-in just for the module's tests, but still require manual opt-in in the main sources of the module

fire-at-will: Okay, modified things so that we require manual opt-in/propagation in the purchase module's `src`, but not in the tests in a7da1a06b0e2002df0a3029ac1597219d5b23e61

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapperFactory.kt
fire-at-will: Good idea! Extended the error message in 9b7abd5d5d4f2728492d6f6ff9813007a65ade4b

purchases/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyDeviceIdentifiersFetcher.kt
fire-at-will: Makes sense, moved to an `attribution` package in a05db018e96ab3dffeb126778cf9953fe99d3cfe 👍

purchases/src/main/kotlin/com/revenuecat/purchases/google/usecase/AcknowledgePurchaseUseCase.kt
fire-at-will: I figured out what was happening here and with those empty spaces...I recently made the purchases module automatically opt-in to the `@InternalRevenueCatAPI` annotation, and when I did so, Codex removed the annotations but didn't revert the formatting 🤦‍♂️ However, in a discussion with Joop, we decided to make opt-ing in to the annotation required in the module's sources so that developers must choose between opting in and propagating the annotation. In the tests, we just automatically opt-in for convenience.

I've made the following changes:
- Required opt-in or propagation of the `@InternalRevenueCatAPI` annotation in the `purchases` module's source files
- In the source files, gone through and opted-in to the `@InternalRevenueCatAPI` annotation
- Removed all of the unused imports & newlines in the test files

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/logging/LogUtils.kt
fire-at-will: Marked these functions as `@JvmSynthetic` in 059dd9c8af2eb54a8b3681b036f9e4b04d530580 👍

PR #3006, opened 2026-01-13
Fixes incorrect lifecycle transitions in `CompatComposeView`

JayShortway:
## Bug
Views can be detached and reattached to a Window multiple times. This means we should not set the `CompatComposeView`'s lifecycle to `DESTROYED` when it is detached, as that's a final state and does not allow us to move it back to `CREATED` once it gets reattached. 

## Fix
The `CompatComposeView`'s lifecycle is no longer set to `DESTROYED` in `onDetachedFromWindow`. Instead, it is destroyed when the Paywall or Customer Center is dismissed. It will also be destroyed if the Activity gets destroyed, in case that happens before dismissal. 

## Notes 
It's probably best to review by commit, as some things were moved around which makes the total diff a bit hard to read.

Fixes [b1542b35](https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/6507804763198446256/crashes/b1542b35/details) & [1b142dcb](https://play.google.com/sdk-console/accounts/3894912036640479868/sdks/6507804763198446256/crashes/1b142dcb/details).

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
@@ -94,6 +94,10 @@ abstract class CompatComposeView @JvmOverloads internal constructor(
             lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_START)
         }
 
+        fun onDetachedFromWindow() {
+            destroy()
+        }
+
         fun onWindowVisibilityChanged(visibility: Int) {
             if (visibility == VISIBLE) {
                 lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_START)
@@ -209,7 +213,7 @@ abstract class CompatComposeView @JvmOverloads internal constructor(
     }
 
     override fun onDetachedFromWindow() {
-        (lifecycleOwner as? ViewLifecycleOwner)?.destroy()
+        (lifecycleOwner as? ViewLifecycleOwner)?.onDetachedFromWindow()
         deinitViewTreeOwners()
         super.onDetachedFromWindow()
     }


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -172,6 +172,11 @@ class PaywallView : CompatComposeView {
         this.shouldDisplayDismissButton = shouldDisplayDismissButton
     }
 
+    private fun dismiss() {
+        dismissHandler?.invoke()
+        destroy()
```
tonidero: The case for destroying the view when onDismiss is called is interesting... As in, there is a chance (in a wrong integration), that the developer doesn't remove the view when onDismiss is called, right? Not sure if in that case we should just say it's a wrong integration... But in that case, it might be something we would need to log somewhere so it's easier to debug.

tonidero: I've been testing this scenario and it seems to work fine. I just [added a log](https://github.com/RevenueCat/purchases-android/pull/3006/commits/aed668cf0a9ff2a8ecb41917f667e22fdaf6a803) that could help debug in some situations.

JayShortway: Hmm you're right! For a hard paywall we shouldn't destroy in onDismiss. Maybe we should remove that functionality altogether to be on the safe side?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
```diff
@@ -148,18 +242,21 @@ abstract class CompatComposeView @JvmOverloads internal constructor(
         // null, so we set up all of them. In Expo 54+, LifecycleOwner is already set up, but SavedStateRegistryOwner
         // and ViewModelStoreOwner are not. We track each one separately to avoid performing operations on owners we
         // didn't set up.
-        if (windowRoot.findViewTreeLifecycleOwner() == null) {
+        val viewTreeLifecycleOwner = windowRoot.findViewTreeLifecycleOwner()
+        if (lifecycleOwner == null) {
```
vegaro, line 267: There's no way `viewTreeLifecycleOwner` has changed in a new call to `onAttachedToWindow` (where `lifecycleOwner` is not null)?

tonidero: Hmm I guess if they reused views accross different places, there could be a chance... Though TBH, that totally feels like the developer is doing something wrong to me 😅

tonidero:
I've made some changes here. So instead, we're again calling ON_DESTROY when the view is detached. Instead, the next time it's attached, if we already had an owner (thanks for that refactor!) we just reset the current state if needed and then create a new lifecycle owner. This feels safer, and would keep maintaining the hard paywall case. Lmk what you think @JayShortway @vegaro

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
```diff
@@ -2,7 +2,7 @@ package com.revenuecat.paywallstester
 
 object Constants {
     // PaywallsTester supports switching between 2 API keys on the App Info screen. It uses "A" by default.
-    const val GOOGLE_API_KEY_A = "API_KEY_A"
+    const val GOOGLE_API_KEY_A = "test_RVrPohspVACiBvDZJwoaBMAjhgs"
```
JayShortway: 👀

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
```diff
@@ -78,44 +166,42 @@ abstract class CompatComposeView @JvmOverloads internal constructor(
     }
 
     override fun onAttachedToWindow() {
+        // This is just a safety measure in case the View is re-attached after being detached without having
+        // been destroyed yet.
+        lifecycleOwner?.let { lifecycleOwner ->
+            if (lifecycleOwner is ViewLifecycleOwner) {
+                Logger.e("Attaching to a window a view that was already detached. Resetting state")
```
JayShortway:
```suggestion
                Logger.w("Attaching a previously-detached view to a window. Resetting state.")
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
```diff
@@ -49,16 +53,100 @@ abstract class CompatComposeView @JvmOverloads internal constructor(
...
+            this@CompatComposeView.onDestroy()
+        }
     }
     private val savedStateRegistryController: SavedStateRegistryController =
         SavedStateRegistryController.create(this)
 
-    override val lifecycle: Lifecycle = lifecycleRegistry
+    override val lifecycle: Lifecycle
+        get() = lifecycleOwner?.lifecycle
+            ?: error("Lifecycle accessed before view attached to window")
```
JayShortway: Oh I forgot this but I was thinking if we maybe could return `INITIALIZED` here instead of erroring? Not sure if that's possible.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
```diff
@@ -49,16 +53,100 @@ abstract class CompatComposeView @JvmOverloads internal constructor(
...
+            lifecycleRegistry.handleLifecycleEvent(Lifecycle.Event.ON_DESTROY)
+        }
+
+        fun destroy() {
+            if (lifecycleRegistry.currentState == Lifecycle.State.DESTROYED) return
+
+            activityLifecycleCallbacks?.let { callbacks ->
+                activity?.application?.unregisterActivityLifecycleCallbacks(callbacks)
+                activityLifecycleCallbacks = null
+            }
```
JayShortway, line 123: If we don't unregister this callback, we're leaking the `CompatComposeView`. We're only unregistering it in `destroy()`, which is called in `onDismiss()`, right? That should be fine then I think.

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
tonidero: 😬

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
tonidero: Ohh that's a great catch!! We were actually passing the lifecycle events before it had the chance to enter the `destroy` method, which handles the events already. I believe [this](https://github.com/RevenueCat/purchases-android/pull/3006/commits/5e44f56fe79b83ee21c566e69bab9210100a7a06) should fix it

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
tonidero: Is [this](https://github.com/RevenueCat/purchases-android/pull/3006/commits/0f77149cd8e5877c46058be37987e4e0249c687b) what you meant?

Feels dangerous, in case this is used... but I guess it should indeed never be used.

vegaro (approved):
Still looking good with the latest changes. I would wait for Joop's aproval though 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
JayShortway: Hmm, choosing the lesser evil. Indeed we are sure that _we_ will not add an observer to this initial lifecycle, but `lifecycle` is a public val, so who knows what other components do?

Maybe we should at the very least log an error in `addObserver` and `removeObserver`, or maybe even throw an `error()`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
JayShortway: Maybe we can keep the `ViewLifecycleOwner.onDetachedFromWindow()` method and just call `destroy()` in there? I quite like the pattern of `ViewLifecycleOwner` having identically-named methods that it cares about, that we should forward to it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3006/commits/708ecd54fdb002f29c6e3d9c0cbe446a3bcbd7ec

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/CompatComposeView.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3006/commits/14527c72878d0686032d89ed4f292854fc270ca5 👍

tonidero (approved):

PR #3017, opened 2026-01-15
Fix gradient overlay to cover full viewport instead of image bounds

facumenzella:
## Summary
Fixes gradient overlays on background images appearing more aggressive on Android compared to the web builder.

## Problem
When a paywall has a background image with a color overlay (gradient), the overlay was being applied via the `underlay()` modifier which draws within the image's bounds. Since images are often aspect-fit and smaller than the viewport, a gradient defined as "66% white from bottom" would compress into that smaller area, making much more of the screen appear white.

**Web builder:** Overlay covers 100% of viewport → gradient spread across full screen
**Android (before):** Overlay only covered image bounds → gradient compressed into smaller area

### `Background.kt`
- Removed `colorOverlay` application from the image background modifier
- Added `alignment = TopCenter` to position the image at the top

### `ViewWithVideoBackground.kt`
- Extended `WithOptionalVideoBackground` to also handle `BackgroundStyle.Image` with overlays
- Renders overlay as a separate `Box` layer using `.matchParentSize()` to fill the full container
- Overlay renders between background image and content (correct Z-order)

| Android | Dashboard |
|--------|--------|
| Screenshot 2026-01-15 at 10 37 51 | Screenshot 2026-01-15 at 10 46 56 |

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/BackgroundTests.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/BackgroundTests.kt
@@ -175,6 +175,7 @@ class BackgroundTests {
             // Image backgrounds with overlays are rendered in two layers:
             // 1. The image is applied via the background modifier
             // 2. The overlay is rendered as a separate Box layer (matching WithOptionalBackgroundOverlay)
+            val imageStyle = backgroundStyle as BackgroundStyle.Image
             Box(
                 modifier = Modifier
                     .requiredSize(sizeDp)
@@ -182,8 +183,8 @@ class BackgroundTests {
                     .semantics { testTag = "box" }
             ) {
                 // Overlay layer - rendered behind content but in front of image
-                backgroundStyle.colorOverlay?.let { overlay ->
-                    Box(modifier = Modifier.matchParentSize().background(overlay))
+                imageStyle.colorOverlay?.let { overlay ->
+                    Box(modifier = Modifier.matchParentSize().background(color = overlay))
                 }
                 // Content layer
                 Text(


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
```diff
@@ -72,16 +73,33 @@ internal fun WithOptionalVideoBackground(
...
+            ViewWithVideoBackground(
+                state = state,
+                background = background,
+                shape = shape,
+                modifier = modifier,
+            ) {
+                content()
+            }
+        }
+        background is BackgroundStyle.Image && background.colorOverlay != null -> {
```
vegaro: hmm this Composable is `WithOptionalVideoBackground`, does it make sense we are checking for a `BackgroundStyle.Image` background?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -34,12 +33,15 @@ internal fun Modifier.background(
     when (background) {
         is BackgroundStyle.Color -> this.background(color = background.color, shape = shape)
         is BackgroundStyle.Image ->
+            // Image backgrounds with color overlays are handled specially in WithOptionalOverlayBackground
```
vegaro: I can't find `WithOptionalOverlayBackground` in the code

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ViewWithVideoBackground.kt
facumenzella: What about renaming it to `WithOptionalBackgroundOverlay` or `WithOptionalOverlayBackground`?

vegaro: and maybe take it out of this class as well

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
facumenzella: I think this should be `WithOptionalVideoBackground` right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -34,12 +33,15 @@ internal fun Modifier.background(
...
         is BackgroundStyle.Color -> this.background(color = background.color, shape = shape)
         is BackgroundStyle.Image ->
+            // Image backgrounds with color overlays are handled specially in WithOptionalBackgroundOverlay
+            // to ensure the overlay covers the full container, not just the image bounds.
+            // This matches the web builder behavior where overlays cover 100% of the viewport.
             this.clip(shape)
                 .paint(
                     painter = background.painter,
                     contentScale = background.contentScale,
+                    alignment = androidx.compose.ui.Alignment.TopCenter,
```
vegaro: `Alignment` should be imported instead of being referenced like this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -34,12 +33,15 @@ internal fun Modifier.background(
     when (background) {
         is BackgroundStyle.Color -> this.background(color = background.color, shape = shape)
         is BackgroundStyle.Image ->
+            // Image backgrounds with color overlays are handled specially in WithOptionalBackgroundOverlay
+            // to ensure the overlay covers the full container, not just the image bounds.
+            // This matches the web builder behavior where overlays cover 100% of the viewport.
```
vegaro: I am a bit confused by this because doesn't this mean it will apply the background twice? Here and in `WithOptionalBackgroundOverlay`. I believe in this `when`, `BackgroundStyle.Video` is returning `this` because of that reason.

facumenzella: The background is not applied twice for images. Here's how it works:                                                  
                                                                                                                                       
  For `BackgroundStyle.Image`:                                                                                                           
  - `Background.kt` applies only the image via paint()                                                                                   
  - `WithOptionalBackgroundOverlay` renders only the color overlay (if present) as a separate layer                                      
                                                                                                                                       
  For `BackgroundStyle.Video`:                                                                                                           
  - `Background.kt` returns this (no-op)                                                                                                 
  - `WithOptionalBackgroundOverlay` renders both the video AND its overlay                                                               
                                                                                                                                       
  The difference is that videos need to be rendered as a composable (VideoComponentView), so everything is handled in `WithOptionalBackgroundOverlay. Images` can be applied via a modifier, so we only need `WithOptionalBackgroundOverlay` to add the overlay layer on top.                                                                                                                       
                                                                                                                                       
  I've updated the comment to make this clearer:                                                                                       
  // The image is applied here via paint(). Color overlays (if present) are rendered                                                   
  // separately in WithOptionalBackgroundOverlay to ensure the overlay covers the full                                                 
  // container, not just the image bounds. This matches the web builder behavior.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/modifier/Background.kt
```diff
@@ -34,12 +34,15 @@ internal fun Modifier.background(
...
         is BackgroundStyle.Color -> this.background(color = background.color, shape = shape)
         is BackgroundStyle.Image ->
+            // The image is applied here via paint(). Color overlays (if present) are rendered
+            // separately in WithOptionalBackgroundOverlay to ensure the overlay covers the full
+            // container, not just the image bounds. This matches the web builder behavior.
             this.clip(shape)
                 .paint(
                     painter = background.painter,
                     contentScale = background.contentScale,
+                    alignment = Alignment.TopCenter,
```
vegaro, line 44: is this the same in iOS and the builder?

facumenzella: Yes it's the same one 💅

PR #3018, opened 2026-01-15
Send `paywall_id` in paywall events

tonidero:
### Description
This PR parses the paywall ID sent by the backend, and sends it back as part of all paywall events and as part of the POST /receipt paywall data we attach.

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionBackendEventsTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionBackendEventsTest.kt
@@ -36,7 +36,7 @@ internal class ProductionBackendEventsTest: BaseBackendIntegrationTest() {
                             appUserID = UUID.randomUUID().toString(),
                             sessionID = UUID.randomUUID().toString(),
                             offeringID = UUID.randomUUID().toString(),
-                            paywallID = UUID.randomUUID().toString(),
+                            paywallID = "paywall_id",
                             paywallRevision = 1,
                             timestamp = Date().time,
                             displayMode = "FULL_SCREEN",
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEvent.kt
```diff
@@ -66,6 +66,8 @@ data class PaywallEvent(
 
     @Serializable
     data class Data(
+        // Nullable for legacy paywalls
+        val paywallIdentifier: String?,
```
ajpallares: > // Nullable for legacy paywalls

Can legacy paywalls not contain an id?

ajpallares: I ask because I'm seeing the responses of GET /offerings and I do see an `id` field in the `paywall` field 🤔

ajpallares, line 69: Also, to double check. The name of the field in `PaywallEvent` is `paywall_identifier` while in `PaywallPostReceiptData` is `paywall_id`. Is that correct?

tonidero: We were not adding it to events yet actually 😅 Added in https://github.com/RevenueCat/purchases-android/pull/3018/commits/a092ac9684a3f1a66bb92055c0cfe7c123781212 as `paywall_id`.

tonidero: Thanks for mentioning it. Now we should parse it correctly from https://github.com/RevenueCat/purchases-android/pull/3018/commits/a092ac9684a3f1a66bb92055c0cfe7c123781212

ajpallares (approved):
Great! Thanks for the latest changes! I think it's ready now 🚀

PR #3020, opened 2026-01-16
Add custom variables support for v2 paywalls

facumenzella:
## Summary

Adds support for custom paywall variables that can be defined in the RevenueCat dashboard and used as tokens in paywall text (e.g., `{{ custom.username }}`). Variables can be provided at runtime via SDK or fall back to dashboard-configured default values.

### Public API
- Added `CustomVariableValue` sealed class for type-safe custom variable values:
  - `CustomVariableValue.String(value)` - for string values (public)
  - `CustomVariableValue.Number(value)` - for numeric values (@InternalRevenueCatAPI)
  - `CustomVariableValue.Boolean(value)` - for boolean values (@InternalRevenueCatAPI)
  - `CustomVariableValue.from(value)` - factory method supporting String, Int, Long, Float, Double, Boolean
- Added `setCustomVariables(Map<String, CustomVariableValue>)` to `PaywallOptions.Builder`
- Added `customVariables` parameter to `PaywallActivityLauncher.launch()`

### Internal Implementation
- `UiConfig.CustomVariableDefinition` - deserializes dashboard-configured variables with types (`string`, `number`, `boolean`)
- `VariableProcessorV2` - processes custom variables in text templates with support for:
  - Standard syntax: `{{ custom.key }}`
  - Dollar syntax: `{{ $custom.key }}`
  - Functions: `{{ custom.key | uppercase }}`
- Resolution priority: SDK-provided value → Dashboard default → Empty string (with warning)
- `CustomVariableKeyValidator` - validates key format (must start with letter, only letters/numbers/underscores)
- Malformed variable detection - logs warnings for patterns like `{{ custom }}` or `{{ custom. }}`

### Example Usage
```kotlin
PaywallOptions.Builder { /* dismiss */ }
    .setCustomVariables(mapOf(
        "player_name" to CustomVariableValue.String("John"),
    ))
    .build()

// Or via activity launcher
paywallActivityLauncher.launch(
    offering = offering,
    customVariables = mapOf("player_name" to CustomVariableValue.String("John"))
)
```

In paywall text: `Hello {{ custom.player_name }}\!`

## Test plan
- [x] Unit tests for `CustomVariableValue` type conversions and factory method
- [x] Unit tests for key validation (valid/invalid keys)
- [x] Unit tests for malformed variable detection
- [x] Unit tests for `CustomVariableProcessingTests` covering:
  - Basic variable replacement
  - Resolution priority (SDK > Dashboard > empty)
  - Functions (uppercase, lowercase, capitalize)
  - Mixed with product variables
  - Edge cases (special characters, multiline, etc.)
  - All supported types (String, Number, Boolean)
- [x] API tests for `CustomVariableValue`, `PaywallOptions`, `PaywallActivityLauncher`

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform offerings backend request/response_002.json
@@ -2,7 +2,7 @@
   "statusCode": 200,
   "headers": {
     "Connection": "keep-alive",
-    "Content-Length": "92166",
+    "Content-Length": "92204",
     "Server": "cloudflare",
     "x-revenuecat-app-id": "appc735cd95ea"
   },
@@ -2111,7 +2111,8 @@
               "tWuZkHyp1p": "30"
             }
           },
-          "exit_offers": {}
+          "exit_offers": {},
+          "play_store_product_change_mode": {}
         },
         "packages": [
           {
--- a/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform verified offerings backend request/response_002.json
+++ b/purchases/src/test/resources/backend_integration_tests_golden/FallbackURLBackendIntegrationTest/can perform verified offerings backend request/response_002.json
@@ -2,7 +2,7 @@
   "statusCode": 200,
   "headers": {
     "Connection": "keep-alive",
-    "Content-Length": "92166",
+    "Content-Length": "92204",
     "Server": "cloudflare",
     "x-revenuecat-app-id": "appc735cd95ea"
   },
@@ -2111,7 +2111,8 @@
               "tWuZkHyp1p": "30"
             }
           },
-          "exit_offers": {}
+          "exit_offers": {},
+          "play_store_product_change_mode": {}
         },
         "packages": [
           {


purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
```diff
@@ -86,4 +89,15 @@ class UiConfig(
...
+
+    @InternalRevenueCatAPI
+    @Poko
+    @Serializable
+    class CustomVariableDefinition(
+        @get:JvmSynthetic
+        val type: String,
+        @SerialName("default_value")
+        @get:JvmSynthetic
+        val defaultValue: String,
```
vegaro: when we add support for non-strings, this will break. I believe you need a custom deserializer that deserializes depending on `type`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -757,4 +759,19 @@ internal class PaywallViewModelImpl(
             darkMode = isDarkMode,
         )
     }
+
+    /**
+     * Converts custom variables from Map<String, Any> to Map<String, String>.
+     * All values are converted to their string representation.
+     */
+    private fun convertCustomVariablesToStrings(customVariables: Map<String, Any>): Map<String, String> =
```
vegaro: I think the conversion to string should happen when processing in VariableProcessorV2 rather than here in the viewmodel

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
```diff
@@ -114,6 +120,9 @@ internal object VariableProcessorV2 {
         date: Date = Date(),
         countdownTime: CountdownTime? = null,
         countFrom: CountdownComponent.CountFrom = CountdownComponent.CountFrom.DAYS,
+        // Custom variables:
+        customVariables: Map<String, String> = emptyMap(),
```
vegaro: I believe this should get `Map<String, Any>`, since converting to string is part of the processing imo

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
```diff
@@ -205,6 +232,53 @@ internal object VariableProcessorV2 {
...
+        key: String,
+        customVariables: Map<String, String>,
+        defaultCustomVariables: Map<String, String>,
+        functions: List<Function>,
+        currencyLocale: Locale,
+    ): String {
+        val value = customVariables[key]
+            ?: defaultCustomVariables[key]
+            ?: run {
+                Logger.missingCustomVariable(key)
```
vegaro: I would probably inline this log instead of adding another function:

```
Logger.w(
        "Custom variable '$key' was not provided and has no default value. Defaulting to empty string.",
    )
```

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/CustomVariableProcessingTests.kt
```diff
@@ -0,0 +1,338 @@
...
+            variableDataProvider = variableDataProvider,
+            packageContext = PackageContext(
+                discountRelativeToMostExpensivePerMonth = null,
+                showZeroDecimalPlacePrices = false,
+            ),
+            rcPackage = packageYearlyUsd,
+            currencyLocale = Locale.US,
+            dateLocale = Locale.US,
+            date = Date(),
+            customVariables = customVariables,
```
vegaro, line 514: Since this function `VariableProcessorV2.processVariables` accepts `Map<String, String>`, we are only testing at the moment strings, not the whole `Map<String, Any>` that the `customVariables` option accepts. That's one of the reasons why I think we should make `VariableProcessorV2.processVariables` accept `Map<String, Any>` and remove that conversion to String frmo the view model

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
```diff
@@ -12,4 +12,10 @@ object Constants {
     // Optional: Set a preferred UI locale override (e.g., "en-US", "es-ES", "fr-FR")
     // Leave as empty string to use system default locale
     const val PREFERRED_UI_LOCALE_OVERRIDE: String = ""
+
+    // Custom variables to pass to paywalls. These will replace {{ custom.key }} placeholders.
+    val CUSTOM_VARIABLES: Map<String, Any> = mapOf(
+        "user_name" to "John",
+        "app_name" to "Paywall Tester",
```
vegaro: nit pick but maybe add an integer one

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
```diff
@@ -86,4 +102,72 @@ class UiConfig(
...
+        val type = jsonObject["type"]?.jsonPrimitive?.content ?: "string"
+        val defaultValueElement = jsonObject["default_value"]?.jsonPrimitive
+            ?: return UiConfig.CustomVariableDefinition(type = type, defaultValue = "")
+
+        val defaultValue: Any = when (type) {
+            "string" -> defaultValueElement.content
+            "integer" -> defaultValueElement.longOrNull ?: defaultValueElement.content.toLongOrNull()
+                ?: defaultValueElement.content
+            "number" -> defaultValueElement.doubleOrNull ?: defaultValueElement.content.toDoubleOrNull()
+                ?: defaultValueElement.content
```
vegaro: `integer` and `number` are a bit confusing, is this what the backend is sending? To me an integer is a number

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
```diff
@@ -86,4 +102,72 @@ class UiConfig(
...
+        require(decoder is JsonDecoder) { "CustomVariableDefinition can only be deserialized from JSON" }
+
+        val jsonObject = decoder.decodeJsonElement().jsonObject
+        val type = jsonObject["type"]?.jsonPrimitive?.content ?: "string"
+        val defaultValueElement = jsonObject["default_value"]?.jsonPrimitive
+            ?: return UiConfig.CustomVariableDefinition(type = type, defaultValue = "")
+
+        val defaultValue: Any = when (type) {
+            "string" -> defaultValueElement.content
+            "integer" -> defaultValueElement.longOrNull ?: defaultValueElement.content.toLongOrNull()
```
vegaro: I am also confused about this, is it an integer or a long

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/CustomVariableProcessingTests.kt
```diff
@@ -0,0 +1,468 @@
...
+        assertThat(result).isEqualTo("Hello, Jane!")
+    }
+
+    @Test
+    fun `multiple custom variables are replaced`() {
+        val template = "{{ custom.greeting }}, {{ custom.name }}! You have {{ custom.points }} points."
+        val customVariables = mapOf(
+            "greeting" to "Welcome",
+            "name" to "Alice",
+            "points" to "100",
```
vegaro: let's have this one as `100` instead:
```suggestion
            "points" to 100,
```

vegaro: we should also make sure there are tests for the other possible types

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/CustomVariableProcessingTests.kt
```diff
@@ -0,0 +1,468 @@
...
+        val defaultCustomVariables = mapOf<String, Any>("value" to "default")
+
+        val result = processTemplate(
+            template,
+            customVariables = customVariables,
+            defaultCustomVariables = defaultCustomVariables,
+        )
+
+        assertThat(result).isEqualTo("Value: 999")
+    }
```
vegaro, line 446: what about adding a test that passes a custom variable of an unsuported type. Maybe a random object. Or null?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -51,6 +51,11 @@ class PaywallOptions internal constructor(
     internal val mode: PaywallMode,
     val dismissRequest: () -> Unit,
     internal val dismissRequestWithExitOffering: ((exitOffering: Offering?) -> Unit)? = null,
+    /**
+     * Custom variables to be used in paywall text. These values will replace `{{ custom.key }}` or
+     * `{{ $custom.key }}` placeholders in the paywall configuration.
+     */
+    val customVariables: Map<String, Any> = emptyMap(),
```
vegaro: I am noticing when reviewing the iOS PR, that we are limiting the types in there using `CustomVariableValue` (`var customVariables: [String: CustomVariableValue]`) but here we are accepting `Any`, we should probably do something similar here, so it's more clear they can't just pass whatever object here.

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
facumenzella: Removed integer 👍 
We'll just have number

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
facumenzella: Done! Added `CustomVariableDefinitionSerializer` that deserializes based on `type`. Simplified to only support the 3 types the backend sends: `string`, `number`, `boolean` (removed the unused `integer` type).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
facumenzella: Done! Conversion now happens in `VariableProcessorV2.resolveCustomVariable()` using `value.stringValue`. The ViewModel just passes the `Map<String, CustomVariableValue>` directly.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
facumenzella: Done\! Now accepts `Map<String, CustomVariableValue>` which is even better - it's type-safe like iOS.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
facumenzella: Done! The log is now inlined directly in `resolveCustomVariable()`.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/CustomVariableProcessingTests.kt
facumenzella: Done! Tests now use `Map<String, CustomVariableValue>` and cover all supported types (String, Number, Boolean).

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
facumenzella: Done! Added `user_points` with `CustomVariableValue.Number(100)`.

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
facumenzella: You're right, it was confusing\! Checked the backend (khepri) - it only sends 3 types: `string`, `number`, `boolean`. Removed the unused `integer` type completely.

purchases/src/main/kotlin/com/revenuecat/purchases/UiConfig.kt
facumenzella: Removed! Backend only sends `number` (as Double), no `integer` type.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/CustomVariableProcessingTests.kt
facumenzella: Done! Now uses `CustomVariableValue.Number(100)`.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/CustomVariableProcessingTests.kt
facumenzella: Done! Added tests in `CustomVariableValueTests.kt` that verify unsupported types (custom objects, lists, maps, arrays) throw `IllegalArgumentException`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
facumenzella: Done! Created `CustomVariableValue` sealed class matching iOS implementation. Now uses `Map<String, CustomVariableValue>` throughout the entire stack - public API and internal processing.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomVariableValue.kt
```diff
@@ -0,0 +1,113 @@
...
+ *     ))
+ *     .build()
+ * ```
+ *
+ * In the paywall text (configured in the dashboard), use the `custom.` prefix:
+ * ```
+ * Hello {{ custom.player_name }}!
+ * ```
+ */
+sealed class CustomVariableValue {
```
vegaro: This can't be a sealed class becuase it will be a breaking change when we add a new case. We can use abstract classes instead

```
abstract class CustomVariableValue internal constructor() {

    class String(val value: kotlin.String) : CustomVariableValue() {
        override fun equals(other: Any?): kotlin.Boolean =
            other is String && value == other.value

        override fun hashCode(): Int = value.hashCode()

        override fun toString(): kotlin.String = "CustomVariableValue.String(value=$value)"
    }

    class Number(val value: kotlin.Double) : CustomVariableValue() {
        constructor(value: kotlin.Int) : this(value.toDouble())
        constructor(value: kotlin.Long) : this(value.toDouble())
        constructor(value: kotlin.Float) : this(value.toDouble())

        override fun equals(other: Any?): kotlin.Boolean =
            other is Number && value == other.value

        override fun hashCode(): Int = value.hashCode()

        override fun toString(): kotlin.String = "CustomVariableValue.Number(value=$value)"
    }

    class Boolean(val value: kotlin.Boolean) : CustomVariableValue() {
        override fun equals(other: Any?): kotlin.Boolean =
            other is Boolean && value == other.value

        override fun hashCode(): Int = value.hashCode()

        override fun toString(): kotlin.String = "CustomVariableValue.Boolean(value=$value)"
    }

    val stringValue: kotlin.String
        get() = when (this) {
            is String -> value
            is Number -> {
                if (value % 1.0 == 0.0) {
                    value.toLong().toString()
                } else {
                    value.toString()
                }
            }
            is Boolean -> value.toString()
            else -> throw IllegalStateException("Unknown CustomVariableValue type")
        }

    // ... companion object stays the same
}
```

facumenzella: Good catch! Changed to abstract class with internal constructor. Used `error()` instead of `throw IllegalStateException()` since detekt requires it, but functionally equivalent.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomVariableValue.kt
```diff
@@ -0,0 +1,154 @@
...
+
+        override fun hashCode(): Int = value.hashCode()
+
+        override fun toString(): kotlin.String = "CustomVariableValue.String(value=$value)"
+    }
+
+    /**
+     * A numeric value (integer or decimal).
+     */
+    @InternalRevenueCatAPI
```
vegaro: is this so we start exposing them later?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
```diff
@@ -171,11 +183,27 @@ internal object VariableProcessorV2 {
...
+                defaultCustomVariables = defaultCustomVariables,
+                functions = functions,
+                currencyLocale = currencyLocale,
+            )
+        }
+
+        val variable = findVariable(variableIdentifier, variableConfig.variableCompatibilityMap)
         return if (variable == null) {
-            Logger.unknownVariable(variableIdentifier)
+            // findVariable already logged the unsupported variable warning
```
vegaro: I think we can remove this comment

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomVariableValue.kt
facumenzella: That was the plan. Isn't this like `spi` in iOS?

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/ui/screens/paywallfooter/PaywallFooterScreen.kt
vegaro, line 1: This is PaywallFooterScreen which is used for paywalls v1 so we can remove custom variables from here

examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/Constants.kt
```diff
@@ -1,8 +1,12 @@
 package com.revenuecat.paywallstester
 
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.ui.revenuecatui.CustomVariableValue
+
+@OptIn(InternalRevenueCatAPI::class)
 object Constants {
     // PaywallsTester supports switching between 2 API keys on the App Info screen. It uses "A" by default.
-    const val GOOGLE_API_KEY_A = "API_KEY_A"
+    const val GOOGLE_API_KEY_A = "goog_TzjuPFWcFelhzMSOGvKoZWvrdcW"
```
vegaro: 😄

facumenzella:
```suggestion
    const val GOOGLE_API_KEY_A = "API_KEY_A"
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomVariableValue.kt
vegaro: it still can be used by developers if they add the `@InternalRevenueCatAPI` annotation. I am wondering if we should just make them internal instead?

vegaro:
We still need to replace the @RevenuecatInternal with internal so they are not exposed in the API. Like you did with spi

PR #3024, opened 2026-01-16
Add retry mechanism to purchases integration tests

tonidero:
### Description
There is some flakyness in our purchases integration tests that we want to address... But for now until we have more time, we can do a couple retries before notifying anything is wrong.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1000,6 +1000,7 @@ workflows:
       - run-firebase-tests-purchases-custom-entitlement-computation-integration-test:
           <<: *release-branches
       - run-firebase-tests-purchases-integration-tests:
+          <<: *release-branches
           context:
             - slack-secrets
           matrix:


fastlane/Fastfile
```diff
@@ -417,25 +417,44 @@ platform :android do
...
+        else
+          UI.error("All #{max_retries} attempts failed")
+        end
+      end
+    end
+
+    slack_backend_integration_test_results(environment: backend_environment, success: success)
+    ping_heartbeat_monitor(url: ENV['HEARTBEAT_MONITOR_URL_BACKEND_INTEGRATION_TESTS'])
+    if last_exception
+      raise last_exception
```
fire-at-will, line 458: I think that there's a bug here with how we handle `last_exception`. If the tests fail on the first execution, `last_exception` gets set in the `rescue` block. Then, if a subsequent execution passes, `last_exception` never gets reset to `nil`, and thus we raise the exception here even though we have a successful test execution.

To fix this, I think we could set `last_exception` to `nil` before the break in the `run_firebase_integration_tests()` block (before line 441)

fire-at-will (changes requested):
Looks good! I think I found one bug that we should address before shipping, but looks good otherwise!

fastlane/Fastfile
tonidero: Right... Note that we won't ping slack with an error in that case , it would just fail the job. I thought it might be good to keep the job failing, even if it passed in one of the retries, but yeah... I guess since we're retrying, we should just pass the job. Great catch!!

tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3024/commits/a3c5ceec2bf3211d4c99414655a3550ba5fd0cdb

PR #3027, opened 2026-01-19
Add `sdk_originated` to POST receipt

tonidero:
### Description
Adds a new convenience field to the body of the request so it's easier to know if the purchase was originated from the SDK, or at least while the SDK was configured even if it was performed outside while "PurchasesAreCompletedBy.MY_APP".

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -251,6 +251,7 @@ internal class Backend(
         paywallPostReceiptData: PaywallPostReceiptData?,
         // This reflects the value at the time of the purchase, which might come from the LocalTransactionMetadataStore
         purchasesAreCompletedBy: PurchasesAreCompletedBy,
+        sdkOriginated: Boolean,
         onSuccess: PostReceiptDataSuccessCallback,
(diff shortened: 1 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -277,6 +279,7 @@ internal class Backend(
             },
             "observer_mode" to !finishTransactions,
             "purchase_completed_by" to purchasesAreCompletedBy.name.lowercase(),
+            "sdk_originated" to sdkOriginated,
```
tonidero, line 282: A couple notes:
- We will need support for this new field in the backend before merging this cc @antoniobg 
- This will be true even if the purchase was performed outside our SDK in "PurchasesAreCompletedBy.MY_APP", at least in Android. The scenarios where it will be false are mostly to do with the SDK not being configured at the time of the purchase, or the purchase coming from a different device cc @RevenueCat/coresdk

ajpallares: > This will be true even if the purchase was performed outside our SDK in "PurchasesAreCompletedBy.MY_APP", at least in Android.

Thanks for the clarification Toni! Yep, the same applies to iOS 👍

PR #3030, opened 2026-01-19
Update CLAUDE code review workflow reference

vegaro:
It looks like the `author_association` is returning `CONTRIBUTOR` even for me. Changing the workflow to be more permissive in `author_association` but checking if the user has write access

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   claude-review:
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   claude-review:
-    uses: revenuecat/sdk-github-workflows/.github/workflows/claude-code-review.yml@v4
+    uses: revenuecat/sdk-github-workflows/.github/workflows/claude-code-review.yml@8c361bfe2c7753fa19a5a6380b7ce35f0ed2953f
```
vegaro: using this commit for now for tests

PR #3032, opened 2026-01-20
Fix paywall data misattributions

tonidero:
### Description
We noticed the current mechanism we use to attribute the paywall information to a transaction could potentially misattribute that information, since it basically attributed it to the next transaction posted to the servers. This could potentially be wrong in situations like restores with purchases, or if a purchase came from the store while the paywall was opened. This PR attempts a fix for this issue. Note that we can't rely on using our own purchase flow to attribute these purchases, since we also want to attribute purchases done through `PurchasesAreCompletedBy.MY_APP` while using our paywalls. The solution in this PR will support both of these scenarios

### Main changes
- We add 2 new paywall events `PURCHASE_INITIATED` and `PURCHASE_ERROR` and some new possible fields to paywall events `packageId`, `productId`, `errorCode` and `errorMessage`.
- Now, instead of storing in memory the impression event, we will store the `PURCHASE_INITIATED` event, and remove it when either a `CANCEL` or `PURCHASE_ERROR` is received, or in case of a successful purchase, it will get removed from the cache by the PostReceiptHelper, same way we did until now with the IMPRESSION event.
- When we do a POST /receipt, we check if there is a PURCHASE_INITIATED event with a matching product Id and timestamp values (event should have been generated before the actual purchase). In this scenario, we will attribute that information to the transaction.

### DIFF
--- a/purchases/api-defaults-bc7.txt
+++ b/purchases/api-defaults-bc7.txt
@@ -459,8 +459,8 @@ package com.revenuecat.purchases {
     method @kotlin.jvm.Synchronized public void setUpdatedCustomerInfoListener(com.revenuecat.purchases.interfaces.UpdatedCustomerInfoListener?);
     method public void showInAppMessagesIfNeeded(android.app.Activity activity);
     method public void showInAppMessagesIfNeeded(android.app.Activity activity, optional java.util.List<? extends com.revenuecat.purchases.models.InAppMessageType> inAppMessageTypes);
-    method public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price);
-    method public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price, optional Long? purchaseTime);
+    method @Deprecated public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price);
+    method public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price, long purchaseTime);
     method public void syncAttributesAndOfferingsIfNeeded(com.revenuecat.purchases.interfaces.SyncAttributesAndOfferingsCallback callback);
     method @Deprecated public void syncObserverModeAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price);
     method public void syncPurchases();
--- a/purchases/api-defauts.txt
+++ b/purchases/api-defauts.txt
@@ -459,8 +459,8 @@ package com.revenuecat.purchases {
     method @kotlin.jvm.Synchronized public void setUpdatedCustomerInfoListener(com.revenuecat.purchases.interfaces.UpdatedCustomerInfoListener?);
     method public void showInAppMessagesIfNeeded(android.app.Activity activity);
     method public void showInAppMessagesIfNeeded(android.app.Activity activity, optional java.util.List<? extends com.revenuecat.purchases.models.InAppMessageType> inAppMessageTypes);
-    method public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price);
-    method public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price, optional Long? purchaseTime);
+    method @Deprecated public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price);
+    method public void syncAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price, long purchaseTime);
     method public void syncAttributesAndOfferingsIfNeeded(com.revenuecat.purchases.interfaces.SyncAttributesAndOfferingsCallback callback);
     method @Deprecated public void syncObserverModeAmazonPurchase(String productID, String receiptID, String amazonUserID, String? isoCurrencyCode, Double? price);
     method public void syncPurchases();


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -275,20 +275,25 @@ class Purchases internal constructor(
...
      */
+    @Suppress("LongParameterList")
+    @JvmOverloads
     fun syncAmazonPurchase(
         productID: String,
         receiptID: String,
         amazonUserID: String,
         isoCurrencyCode: String?,
         price: Double?,
+        purchaseTime: Long? = null,
```
tonidero: This is needed for `PurchasesAreCompletedBy.MY_APP` amazon purchases to correctly attribute the purchase to the paywall. I was wondering if we should deprecate the existing method and make this mandatory in this case... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -99,6 +103,14 @@ internal sealed class BackendEvent : Event {
...
         @SerialName("exit_offering_id")
         val exitOfferingID: String? = null,
+        @SerialName("package_id")
+        val packageID: String? = null,
+        @SerialName("product_id")
+        val productID: String? = null,
+        @SerialName("error_code")
+        val errorCode: Int? = null,
+        @SerialName("error_message")
+        val errorMessage: String? = null,
```
tonidero, line 113: These will always be empty right now, until we start sending `PURCHASE_INITIATED` and `PURCHASE_ERROR` to the backend.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendStoredEvent.kt
```diff
@@ -69,7 +70,11 @@ internal fun BackendStoredEvent.toBackendEvent(): BackendEvent {
 @JvmSynthetic
 internal fun PaywallEvent.toBackendStoredEvent(
     appUserID: String,
-): BackendStoredEvent {
+): BackendStoredEvent? {
+    if (type == PaywallEventType.PURCHASE_INITIATED || type == PaywallEventType.PURCHASE_ERROR) {
+        // WIP: We should implement support for these paywalls in the backend.
+        return null
+    }
```
tonidero, line 77: Need to add support in the backend for these new paywall types and the new properties before sending these events to the backend. For now, this makes sure we won't send these events.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
ajpallares: I think so, yes. If we're going to add the new API with `purchaseTime` parameter anyway and we do want to have developers pass the `purchaseTime`, I think we should deprecate the version of this method that does not include the `purchaseTime` parameter 👍

rickvdl: Unless it's a lot of hassle I'd vote for deprecating and adding a new method that makes this required, since it's easy to never implement the parameter right now

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallPresentedCache.kt
```diff
@@ -9,30 +10,44 @@ import com.revenuecat.purchases.paywalls.events.PaywallEventType
...
 
     @Synchronized
-    fun cachePresentedPaywall(paywallEvent: PaywallEvent) {
-        lastPaywallImpressionEvent = paywallEvent
+    fun getAndRemovePurchaseInitiatedEventIfNeeded(
+        purchasedProductIDs: List<String>,
+        purchaseTimestamp: Long?,
+    ): PaywallEvent? {
+        val shouldAttributePaywallToPurchase: Boolean = lastPurchaseInitiatedEvent?.let { event ->
+            val wasPurchasePerformedAfterPaywallDisplayed = purchaseTimestamp?.let { timestamp ->
```
ajpallares: Nit: would `wasPurchasePerformedAfterPurchaseInitiated` perhaps be a better name for this variable?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
ajpallares: Question about this parameter. Is this `purchaseTime` something developers will receive or something they will have to create themselves? I ask because I wonder if there's something we can do to prevent developers from harcoding some values here (e.g. `.now()` or similar)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -443,6 +444,8 @@ internal class PaywallViewModelImpl(
         try {
             val customPurchaseHandler = purchaseLogic?.let { it::performPurchase }
 
+            trackPaywallPurchaseInitiated(packageToPurchase)
```
rickvdl, line 447: Should this be done before executing custom purchase logic? Could the actual purchase come in from the queue before reaching this line? Probably not because it's async, but still if there's a small change it might be worth it

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -631,6 +638,56 @@ internal class PaywallViewModelImpl(
...
+            product.productId
+        } else {
+            product.id
+        }
+        val purchaseInitiatedEventData = eventData.copy(
+            packageIdentifier = rcPackage.identifier,
+            productIdentifier = productId,
+        )
+        val event = PaywallEvent(
+            creationData = PaywallEvent.CreationData(UUID.randomUUID(), Date()),
```
rickvdl, line 658: Out of curiosity: should we inject the UUID and date creator and use it in the VM tests? Or is that something we usually don't do on Android (or have alternative ways of mocking these?)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Devs can obtain this from the Amazon SDK, through the `ProductResponse.Receipt.purchaseTime`. See: https://amzndevresources.com/iap/appstore-sdk-api-reference/3.0.8/com/amazon/device/iap/model/Receipt.html

tonidero: And regarding what we could do, we could create a method that accepts a `receipt` object instead, so we pick the data ourselves... But it's a bigger API change IMO, and would mean having amazon classes in our API by default, which I think we should avoid... We could try to extract amazon to a separate module, but that's also a quite big change...

Other than that, not much we can do to avoid devs just putting their own timestamp I'm afraid :(

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: We do that in a bunch of tests... But we didn't add it here 😅. I do think it could be good to make the tests more deterministics, but in this case, the tests are using relative dates (like `1.minutes.ago()` or `1.days.ago()`), so it should work... Still that's a valid improvement to these tests we could do, but I would prefer to do separately to minimize the changes. Wdyt?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
ajpallares: Right. I agree. The compromises aren't worth it in this case. Thank you for explaining!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: I don't think so. The trackPaywallPurchaseInitiated would do most work synchronously (except the last part, of adding the event to disk). So all the `PaywallPresentedCache` part should be performed before the custom purchase logic has a chance to even start, so I don't think there is a chance of that happening. Lmk if I misunderstood or you have any scenario in mind!!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
rickvdl: Sure! Thanks for the context on this. Totally fine if tests work without it for now I'd say

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
rickvdl: Ah sorry, I misread the diff. I thought `purchaseLogic?.let { it::performPurchase }` was executing the actual purchase logic (before tracking the purchase initiated). So that's why I was asking. But I see it's happening later down, so all good :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Deprecated and created new one with mandatory value in https://github.com/RevenueCat/purchases-android/pull/3032/commits/484ef223f57c1ad0c57ac3ad339432fb68da2ffd

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/PaywallPresentedCache.kt
tonidero: Fixed in https://github.com/RevenueCat/purchases-android/pull/3032/commits/9272c8614ff6a698f8ed63a53b9b04cc8596a110

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -276,19 +276,59 @@ class Purchases internal constructor(
...
+     *
+     * The receipt IDs are cached if successfully posted so they are not posted more than once.
+     *
+     * @param [productID] Product ID associated to the purchase.
+     * @param [receiptID] ReceiptId that represents the Amazon purchase.
+     * @param [amazonUserID] Amazon's userID. This parameter will be ignored when syncing a Google purchase.
+     * @param [isoCurrencyCode] Product's currency code in ISO 4217 format.
+     * @param [price] Product's price.
+     * @param [purchaseTime] Time the product was purchased, in milliseconds since the epoch.
+     * Can be obtained from the PurchaseResponse > Receipt > purchaseTime.
```
ajpallares, line 314: Nice!

PR #3034, opened 2026-01-20
Add `sdk_originated` to POST /receipt

tonidero:
### Description
This adds a new property `sdk_originated` to POST /receipt requests. This value will be true when the purchase was initiated using one of the `purchase` methods of the SDK (including purchases performed through our paywalls when `PurchasesAreCompletedBy.REVENUECAT`). It will be false in all other cases like purchases returned by the store listener or sync/restore purchases, or when we sync active purchases.

#### TODO
- [ ] Make sure it works once backend implementation is merged


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1317,8 +1318,14 @@ internal class PurchasesOrchestrator(
...
                 val callbackPair: Pair<SuccessfulPurchaseCallback, ErrorPurchaseCallback>
                 val deprecatedProductChangeListener: ProductChangeCallback?
+                val sdkOriginated: Boolean
 
                 synchronized(this@PurchasesOrchestrator) {
+                    sdkOriginated = purchases.all { purchase ->
+                        purchase.productIds.firstOrNull()?.let {
+                            state.purchaseCallbacksByProductId.contains(it)
+                        } ?: false
+                    }
```
tonidero, line 1328: This is the important bit to calculate whether the purchase was initiated by the SDK. Basically, we check if we have any pending callback for that product id in the listener before actually continuing.

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
```diff
@@ -33,6 +33,7 @@ internal data class ReceiptInfo(
     @Serializable(with = ReplacementModeSerializer::class)
     val replacementMode: ReplacementMode? = null,
     val platformProductIds: List<Map<String, String?>> = emptyList(),
+    val sdkOriginated: Boolean = false,
```
tonidero, line 36: Note I added this to the ReceiptInfo... I considered adding it to the LocalTransactionMetadata object directly... but this had a bit less wiring... TBH, Right now both data classes may be slightly redundant, but I think it's fine to add this here for now. Lmk if you think otherwise!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
skydoves: Just curiosity, for multi-line subscriptions (transactions with multiple product IDs), how does the logic handle determining `sdkOriginated`?

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
skydoves: So `ReceiptInfo` is serialized and cached, what happens to existing cached transactions after an app (SDK) update? They won't have `sdkOriginated` set. The default false might be incorrect for purchases that were actually SDK-originated. Is this an acceptable edge case or should there be migration logic somewhere.

tonidero: Right, unfortunately we won't be able to recover that information in that case. We could default to null in this case, but that indeed is an edge case and this information will only be used for tracking specific types of purchases for specific types of developers, which we can/should ask to update to a version with all these changes included. If any previous purchase gets posted with this set to false, it shouldn't be too bad IMO. cc @JayShortway

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: That's a very good question. So for multiline purchases performed by our SDK, there will always be a "base" product, and "add-on" products. The callback here will correspond to that base product. But it's true the current logic just picks the first for the transaction, but looking at it, there is no guarantee that first one will be the bae product. I will rethink this. Great catch!

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
JayShortway: Is my understanding correct that this edge case will only exist for transactions that were created after we introduced caching of transaction metadata (9.19.1), but before we added the `sdkOriginated` field?

tonidero: That's correct

ajpallares: > Note I added this to the ReceiptInfo... I considered adding it to the LocalTransactionMetadata object directly... but this had a bit less wiring...

In iOS, I did add it directly to `LocalTransactionMetadata` [here](https://github.com/RevenueCat/purchases-ios/pull/6091/changes#diff-090879da64deba32cccacfdf414ba4540a9b924ade84a4695997b1f4c92eb174R39-R42) precisely because I could define it later in the call tree (right before creating the `LocalTransactionMetadata`), so that I wouldn't need to pass it around.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: May be related to [my previous comment](https://github.com/RevenueCat/purchases-android/pull/3034/changes#r2711416394) but, in iOS, I basically defined `sdkOriginated` as
```swift
sdkOriginated = initiationSource == .purchase
```

Wouldn't the same be valid here? Or perhaps I'm missing some case where these two conditions are not equivalent

tonidero: No, that won't work in android, since purchases received through the listener (which may come from the developer's implementation, or from redemption codes for example) will also have initiationSource = purchase in android.

purchases/src/main/kotlin/com/revenuecat/purchases/common/ReceiptInfo.kt
tonidero: > I did add it directly to LocalTransactionMetadata [here](https://github.com/RevenueCat/purchases-ios/pull/6091/changes#diff-090879da64deba32cccacfdf414ba4540a9b924ade84a4695997b1f4c92eb174R39-R42) precisely because I could define it later in the call tree (right before creating the LocalTransactionMetadata), so that I wouldn't need to pass it around.

It's a bit the opposite in Android 😅, the ReceiptInfo object gets passed around as is, so after constructing it, we don't need to pass it more. The rest of the `LocalTransactionMetadata` object gets unpacked to be passed around, so it's a bit more wiring. TBH, I feel it's fine to have this difference between both platforms (Like `ReceiptInfo` doesn't exist in iOS for example). Happy to chat more about this though!

ajpallares: Totally agree 👍

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: Ohh, good to know. Thanks for explaining!
I will recheck it in iOS as well to make sure that `initiationSource == .purchase` is indeed equivalent to using a `purchase` API (either implicitly or explicitly).

ajpallares: One more question, regarding the current PR description:
> This value will be true when the purchase was initiated using one of the purchase methods of the SDK (including purchases performed through our paywalls when `PurchasesAreCompletedBy.MY_APP`)

Regarding the text in parenthesis. Wouldn't this "depend"?

Even when `PurchasesAreCompletedBy.MY_APP`, it can happen that the developer uses an SDK `purchase` API to make the purchase (I cannot think why, but, theoretically, it can happen I think). In this scenario, I believe the `sdkOriginated` should still be `true` though. I think that's considered in code, but want to make sure.

tonidero: Right right, that should be covered. The text in parenthesis was wrong, I mostly meant that this will be true only if using one of the `purchase` methods, which includes our paywalls purchases when using `PurchasesAreCompletedBy.REVENUECAT` (it was wrong before).

tonidero: Regarding multiline purchases, I think the proper solution would be to key the callbacks by all the product Ids in the purchase attempt, instead of just in the base product Id. I started doing that but it was becoming a bigger change... I would like to do this in a PR to avoid this from growing too much. cc @fire-at-will since I know you worked on that, in case you see any issues with this approach. I would 

For now, I wanted to minimize the changes so I've improved the current solution with this: https://github.com/RevenueCat/purchases-android/pull/3034/commits/5af7fbb781773c98662490a1551ce0189e5b135a. Unfortunately, there is no way of knowing which product is the "base" product from Google, since that's an RC system. So here I'm just checking if there is any callback waiting for any of the multiline purchases product ids. This could in theory fail if a user starts multiple purchases simultaneously, where one of them is part of a multiline purchase of another. Feels like a small enough edge case though.

ajpallares (approved):
Looks great!

PR #2058, opened 2025-01-15
[Paywalls V2] Handles intro offer eligibility overrides

JayShortway:
## Description
Intro offer eligibility was not properly hooked up yet to the override system, and it was using the iOS concept of at most being eligible for a single offer. This PR ensures the offer eligibility follows the Android model of potentially being eligible for multiple offers, and hooks it up to the override system. 

A bunch of the added lines are tests.

PR #3036, opened 2026-01-21
Make sure countdown component works when there's no package context

vegaro:
Makes the countdown component independent from package context, allowing countdown variables to be processed even when no package is available. It's an edge case but it can happen easily in unpublished paywalls.

Adds countdown previews to CountdownComponentView for both running and times-up states and also tests

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentViewTests.kt
@@ -994,10 +994,8 @@ class TextComponentViewTests {
         // Assert
         // The countdown variables should be processed, not shown as raw placeholders
         // If the bug exists, we'd see "{{ count_days_without_zero }}d {{ count_hours_without_zero }}h"
-        // With the fix, we'd see something like "1d 23h" (approximately 2 days from now)
         val node = onNodeWithTag("countdown_text").onChild()
 
-        // Get the actual text and verify it doesn't contain raw variable placeholders
         val actualText = node.fetchSemanticsNode().config
             .first { it.key.name == "Text" }
             .value
@@ -1005,7 +1003,6 @@ class TextComponentViewTests {
 
         assertThat(actualText).doesNotContain("{{")
         assertThat(actualText).doesNotContain("}}")
-        // Should contain something like "1d 23h" - verify it has numbers
         assertThat(actualText).containsPattern("\\d+d \\d+h")
     }
 


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/text/TextComponentView.kt
```diff
@@ -112,41 +110,34 @@ private fun rememberProcessedText(
...
-                val dateLocale = state.locale.toJavaLocale()
-                val currencyLocale = state.currencyLocale.toJavaLocale()
-
-                val introEligibility = packageToUse.introEligibility
-
-                when (introEligibility) {
-                    IntroOfferEligibility.INELIGIBLE -> textState.text
-                    IntroOfferEligibility.SINGLE_OFFER_ELIGIBLE -> textState.text
-                    IntroOfferEligibility.MULTIPLE_OFFERS_ELIGIBLE -> textState.text
-                }
```
vegaro, line 125: This was dead code. I honestly don't know what it was supposed to do. I think it was attempting to return the text without variables?

JayShortway: I see I added this in https://github.com/RevenueCat/purchases-android/pull/2058 a year ago. I really don't remember what the idea was here haha. Are we sure the text component view properly handles intro offer eligibility?

vegaro: yeah I need to confirm. I assume it is correctly handled otherwise we would have heard of it, but will double check

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
```diff
@@ -277,156 +275,138 @@ internal object VariableProcessorV2 {
...
+        Variable.PRODUCT_SECONDARY_OFFER_PERIOD ->
+            rcPackage?.secondIntroOffer?.productOfferPeriod(localizedVariableKeys)
 
-        Variable.PRODUCT_RELATIVE_DISCOUNT -> packageContext.relativeDiscount(localizedVariableKeys)
+        Variable.PRODUCT_SECONDARY_OFFER_PERIOD_ABBREVIATED ->
+            rcPackage?.secondIntroOffer?.productOfferPeriodAbbreviated(localizedVariableKeys)
 
-        Variable.PRODUCT_STORE_PRODUCT_NAME -> rcPackage.product.name
+        Variable.PRODUCT_RELATIVE_DISCOUNT -> packageContext?.relativeDiscount(localizedVariableKeys)
+        Variable.PRODUCT_STORE_PRODUCT_NAME -> rcPackage?.product?.name
```
tonidero, line 409: I wonder if we should split the variables with and without package support for cleanliness, since now most variables need to handle nullability... I'm assuming the backend already would do some of the validations to make sure we only use these variables when appropriate... But not a blocker, I would say.

tonidero (approved):
Makes sense, thanks for fixing this!! 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableProcessorV2.kt
vegaro: yeah, I tried that approach, but first checking the type of variable with a `when`, then running the replacement of the variables with another `when`. But thinking about it now... maybe we can have two sets, and we just need to check the variable is part of either set. That should be pretty efficient. Let me try that approach

vegaro: Actually I am not convinced. I think it can become fragile when someone adds a new variable...

I think I am just going to leave it as is.

tonidero: Sounds good to me! I do think it's nice to have all variables at the same level 👍

PR #3038, opened 2026-01-21
Support the Galaxy Store in the Purchase Tester App

fire-at-will:
### Description
This PR adds basic support for the Galaxy Store to the purchase tester app. Specifically, the PR:
- Bumps the purchase tester's minSDK version to 23
- Always declares a dependency on the `feature/galaxy` module
- Conditionally includes the Samsung IAP SDK in the classpath if the Samsung IAP SDK AAR is found in the repo root's `libs/` directory or the purchase tester app's `libs/` directory
- Adds a radio button to the configuration screen to allow the user to select the Galaxy Store. When the Galaxy radio option is selected, a message appears reminding the user to include the Samsung SDK and reminding them that Galaxy sandbox only works on physical Samsung devices

Screenshot 2026-01-21 at 4 49 39 PM

Product changes aren't supported yet - we'll add support for that in a future PR once the Galaxy product change APIs are solidified.

### DIFF
--- a/examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureFragment.kt
+++ b/examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureFragment.kt
@@ -241,7 +241,7 @@ class ConfigureFragment : Fragment() {
                     java.lang.Enum.valueOf(modeClass as Class<out Enum<*>>, "TEST")
                 val method = builderClass.getMethod("galaxyBillingMode", modeClass)
                 method.invoke(builder, testMode)
-            } catch (e: Exception) {
+            } catch (e: ReflectiveOperationException) {
                 Log.e("PurchaseTester", "Failed to set Galaxy billing mode via reflection.", e)
                 Toast.makeText(
                     context,


examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureFragment.kt
```diff
@@ -176,6 +194,50 @@ class ConfigureFragment : Fragment() {
...
+            // and settings.gradle.kts conditionally includes the module.
+            val builderClass =
+                Class.forName("com.revenuecat.purchases.galaxy.GalaxyConfiguration\$Builder")
+            val constructor = builderClass.getConstructor(Context::class.java, String::class.java)
+            val builder = constructor.newInstance(context, apiKey) as PurchasesConfiguration.Builder
+            try {
+                val modeClass = Class.forName("com.revenuecat.purchases.galaxy.GalaxyBillingMode")
+                @Suppress("UNCHECKED_CAST")
+                val testMode =
+                    java.lang.Enum.valueOf(modeClass as Class<out Enum<*>>, "TEST")
```
tonidero, line 241: We could import `java.lang.Enum` here :)

tonidero: I do wonder if we want to support the different modes in purchase tester... But I think it's ok to keep it like this for now.

tonidero (approved):
One thing that we are not supporting is Galaxy replacement modes, since we only pass Google's right now AFAIK... But I guess it can come in a separate PR.

examples/purchase-tester/src/main/java/com/revenuecat/purchasetester/ConfigureFragment.kt
fire-at-will: > We could import java.lang.Enum here :)
The reason we have to do some sort of import is that the default `Enum` value is `kotlin.Enum`, which we can't use because there's no equivalent to the `valueOf()` function for the Kotlin enum. We could import with a typealias, but I don't know if that's any clearer.

As for supporting the different modes...the only ones that we might want to use are `TEST` and `ALWAYS_FAIL`, since production doesn't work in sandbox. I'll keep it as is for now :)

PR #3040, opened 2026-01-22
Finish activities instead of crashing when not launched through SDK

ajpallares:
### Motivation

Google Play automated testing can launch our activities directly. Because these activities are only meant to be launched internally by the SDK, we implemented a crash in these situations (see #2974). However, these crashes show up in developer crash reporting tools, even though they don't affect real users.

Instead of crashing, we prefer to finish the activity with a log explaining the issue.

### Description

Changed `PaywallActivity`, `CustomerCenterActivity`, and `SimulatedStoreErrorDialogActivity` to gracefully `finish()` and log an error instead of throwing `IllegalStateException` when not launched through the SDK.

Each activity now logs a specific message indicating correct usage (except for `SimulatedStoreErrorDialogActivity`, which is for internal SDK use only)

Note: `SimulatedStoreErrorDialogActivity` still crashes when launched correctly through the SDK (for security reasons with test API keys).

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityTest.kt
@@ -1,0 +1,30 @@
+package com.revenuecat.purchases.ui.revenuecatui.activity
+
+import android.content.Context
+import android.content.Intent
+import androidx.lifecycle.Lifecycle
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.core.app.launchActivity
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class PaywallActivityTest {
+
+    @Test
+    fun `activity finishes gracefully when not launched through SDK`() {
+        // Arrange - launch without SDK extras (simulating Google automated testing)
+        val intent = Intent(
+            ApplicationProvider.getApplicationContext<Context>(),
+            PaywallActivity::class.java,
+        )
+
+        // Act - launch the activity (it should finish immediately in onCreate)
+        val scenario = launchActivity<PaywallActivity>(intent)
+
+        // Assert - activity should be destroyed (finished gracefully without crashing)
+        assertThat(scenario.state).isEqualTo(Lifecycle.State.DESTROYED)
+    }
+}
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
```diff
@@ -33,11 +34,13 @@ internal class CustomerCenterActivity : ComponentActivity() {
 
         val wasLaunchedThroughSDK = intent.getBooleanExtra(EXTRA_WAS_LAUNCHED_THROUGH_SDK, false)
         if (!wasLaunchedThroughSDK && !Purchases.isConfigured) {
-            throw IllegalStateException(
-                "CustomerCenterActivity was not launched through the SDK. " +
-                    "Please use the SDK methods to open the Customer Center. " +
-                    "This might happen on some Google automated testing, but shouldn't happen to users.",
+            Logger.e(
```
facumenzella, line 37: nit: why are we using `Logger.e` here, and `errorLog` in the other one?

purchases/src/test/java/com/revenuecat/purchases/SimulatedStoreErrorDialogActivityTest.kt
```diff
@@ -103,4 +103,22 @@ class SimulatedStoreErrorDialogActivityTest {
         )
     }
 
+    @Test
+    fun `activity finishes gracefully when not launched through SDK`() {
```
facumenzella, line 107: Can we add a test for the other two activities too?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/CustomerCenterActivity.kt
ajpallares: It seems like the SDK is using `errorLog` in the core SDK (`SimulatedStoreErrorDialogActivity`) and `Logger.e` in RevenueCatUI (`CustomerCenterActivity` and `PaywallActivity`), so I kept it the same

purchases/src/test/java/com/revenuecat/purchases/SimulatedStoreErrorDialogActivityTest.kt
ajpallares: Good point. The other two activities are part of RevenueCatUI, and I think it's not so trivial to add tests with the existing tests setup (at least, it's not so trivial for me 😅). Maybe @RevenueCat/coresdk have some ideas?

tonidero: Well, when dealing with activities, it's recommended to use instrumentation tests, rather than unit tests, since activities are part of the UI layer. But having said that there are some utilities like this one that can allow to do this.

I would say, we probably can create a new unit test class in the UI module (similar to `PaywallViewModelTest` for example) that does a similar thing? Not sure if I'm missing something though.

ajpallares: Thanks Toni! Yep, that worked. Added tests in https://github.com/RevenueCat/purchases-android/pull/3040/commits/45572ee8718ab6d801f4fd2e57c3c0c90cb7393f

PR #3041, opened 2026-01-22
Galaxy Store: Product Caching Improvements

fire-at-will:
### Description
This PR makes two adjustments to the caching logic that the SDK uses to cache Galaxy store products.

#### Bug Fix: Don't Cache Product Promotional Eligibilities
Before this PR, the SDK was caching promotional eligibilities for products, and would never invalidate the cached eligibilities. This could lead to us showing incorrect promotional eligibilities in certain circumstances when the eligibility changes but the products cache hasn't been updated.

This PR changes the SDK's product caching logic so that the SDK only caches the core product data (the `ProductVo`s received from the SDK), and then each time a product is requested, it recomputes the promotional eligibility for all requested products.

#### Optimization: Only Request Uncached Products
When a request was made for a mix of cached and uncached products, the SDK would ignore the cache and re-fetch all products from the Samsung SDK. This resulted in unnecessary overhead when requesting products. 

This PR updates the product fetching logic so that the SDK only requests the missing (uncached) products from the Samsung SDK. Once the uncached products are retrieved from the Samsung SDK, the code returns the fetched products along with the cached ones.

### DIFF
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
@@ -21,7 +21,6 @@ import com.samsung.android.sdk.iap.lib.vo.ErrorVo
 import com.samsung.android.sdk.iap.lib.vo.ProductVo
 import com.samsung.android.sdk.iap.lib.vo.PromotionEligibilityVo
 import java.util.ArrayList
-import java.util.LinkedHashMap
 
 @OptIn(InternalRevenueCatAPI::class)
 internal class ProductDataHandler(
@@ -36,7 +35,6 @@ internal class ProductDataHandler(
     private data class Request(
         val productIds: Set<String>,
         val productType: ProductType,
-        val cachedProducts: List<StoreProduct>,
         val onReceive: StoreProductsCallback,
         val onError: PurchasesErrorCallback,
     )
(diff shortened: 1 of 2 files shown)


feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -114,35 +112,19 @@ internal class ProductDataHandler(
     private fun handleSuccessfulProductsResponse(
         products: List<ProductVo?>,
     ) {
-        val nonNullProducts = products.mapNotNull { it }
-        // The serial execution of this call is an extension of the serial execution of the parent
-        // get products request
-        promotionEligibilityResponseListener.getPromotionEligibilities(
```
fire-at-will, line 120: Extracted this logic to a private `fetchPromotionEligibilityAndHandleStoreProducts()` helper so we can fetch eligibilities for cached products

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -40,7 +40,7 @@ internal class ProductDataHandler(
     )
 
     @get:Synchronized
-    internal val productsCache = mutableMapOf<String, StoreProduct>()
+    internal val productVoCache = mutableMapOf<String, ProductVo>()
```
tonidero, line 43: I guess that the products could potentially remain cached in the offerings cache for a bit, but it's indeed good to fetch the latest information. Related to this, it does bring to mind whether we should just not cache anything at this level, and always fetch all products, since we already have the offerings cache... But TBH, not sure how common it would be for product information to change (or if it will change at all). If we are fairly certain that the information shouldn't normally change, I'm ok keeping this cache I guess, but might be cleaner to just fetch always.

tonidero (approved):
Just a comment about something I'm unsure about... But I think this PR makes sense, so will approve it!

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
fire-at-will: Great point, the products should be cached at the offerings level. I took a look and we aren't doing any caching at this level for google in the `QueryProductDetailsUseCase`, but we do perform caching at this level for Amazon.

To cut down on the complexity, I think I'll close this PR and open a new one that removes the caching here for the Galaxy store so that we're consistent with what we do for Google

PR #3045, opened 2026-01-23
Add `payload_version` to POST /receipt

tonidero:
### Description
In order to make it easier to track changes in the POST /receipt body and how the backend should handle it, we're introducing a new hardcoded property, that should be bumped when there is any change to requests that might require different handling by the backend.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -119,6 +119,12 @@ internal class Backend(
         private const val FETCH_TOKEN = "fetch_token"
         private const val NEW_APP_USER_ID = "new_app_user_id"
 
+        /**
+         * This version should be bumped whenever there is a change to the payload of POST receipt that the backend
+         * may need to handle differently. It should be kept in sync with the iOS SDK.
+         */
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -287,6 +293,7 @@ internal class Backend(
             "initiation_source" to initiationSource.postReceiptFieldValue,
             "paywall" to paywallPostReceiptData?.toMap(),
             "sdk_originated" to receiptInfo.sdkOriginated,
+            "payload_version" to POST_RECEIPT_PAYLOAD_VERSION,
```
tonidero, line 296: Note that I'm not caching this in the LocalTransactionMetadata. I felt this belong to the way the POST /receipt is sent at the moment the receipt is currently sent. I think the LocalTransactionMetadata should have its own versioning, and if something changes in the meaning there, it should run through a migration on the `LocalTransactionMetadata` side. Wdyt? I doubted on this a bit TBH

ajpallares: I agree. I was adding it directly to POST /receipt as well, for the same reasons

PR #3047, opened 2026-01-23
Add Galaxy management url for offline entitlements

fire-at-will:
### Description
Adds the Galaxy Store's subscription list URL as the `management_url` for the Galaxy Store, which is used for offline entitlements.

### DIFF
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
@@ -39,9 +39,6 @@ internal class ProductDataHandler(
         val onError: PurchasesErrorCallback,
     )
 
-    @get:Synchronized
-    internal val productsCache = mutableMapOf<String, StoreProduct>()
-
     @GalaxySerialOperation
     override fun getProductDetails(
         productIds: Set<String>,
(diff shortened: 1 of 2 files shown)


purchases/src/test/java/com/revenuecat/purchases/common/offlineentitlements/OfflineCustomerInfoCalculatorTest.kt
```diff
@@ -660,6 +664,59 @@ class OfflineCustomerInfoCalculatorTest {
...
 
+    @Test
+    fun `management_url depends on store`() {
+        val purchasedProduct = mockActiveProducts().first()
+        val testCases = Store.values()
+            .associateWith { JSONObject.NULL }
+            .toMutableMap()
+            .apply {
+                this[Store.PLAY_STORE] = Constants.GOOGLE_PLAY_MANAGEMENT_URL
+                this[Store.GALAXY] = Constants.GALAXY_STORE_MANAGEMENT_URL
```
tonidero: Hmm total nitpick but feels a bit weird to create a map of nulls, and then add the values... Maybe we could create a computed property for the `Store` enum, `managementUrl` or something like that. Then, we can default it to null for other stores, and this can just use that computed property. Wdyt?

fire-at-will: Yeah, that approach works for me! Switched to this approach in 799bd5aac499e0b79a39ad7c61b31a059d498f79, let me know what you think 👍

PR #3050, opened 2026-01-26
[ADS-1] feat: add ad_format to AdTracker events

peterporfy:
### Checklist

### Motivation

This changes gives the possibility to attach ad format data to ad events. For currently unsupported format (or if it is not available) the user can give `other`. This makes it possible to provide another ad information during the creation of AdEvent.

### Description

- added `AdFormat` to represent the options
- an `AdFormat` must be given for each `AdEvent` instance
- adjusted existing unit tests and snapshots

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
@@ -126,7 +126,6 @@ internal sealed class BackendEvent : Event {
         @SerialName("mediator_name")
         val mediatorName: String,
         @SerialName("ad_format")
-        // support deserializing events before this field
         val adFormat: String? = null,
         val placement: String?,
         @SerialName("ad_unit_id")


purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/types/AdDisplayedData.kt
```diff
@@ -17,6 +18,7 @@ import dev.drewhamilton.poko.Poko
 class AdDisplayedData(
     val networkName: String,
     val mediatorName: AdMediatorName,
+    val adFormat: AdFormat,
```
tonidero, line 21: Just to confirm, this parameter will be mandatory correct? It will mean a breaking change for anyone currently using these APIs, but since it's experimental, I think it's ok if we need to add this.

peterporfy: @tonidero yes, as this API is still experimental I thought we can keep the interface cleaner, as it is low cost yet.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -125,6 +125,8 @@ internal sealed class BackendEvent : Event {
         val networkName: String,
         @SerialName("mediator_name")
         val mediatorName: String,
+        @SerialName("ad_format")
+        val adFormat: String,
```
skydoves: Small thing: should `adFormat` in `BackendEvent.Ad` be nullable with a default?

Events stored on disk before this update won't have the field, so they'd fail to deserialize on the next flush. Won't crash (handled as null in `EventsFileHelper.mapToEvent()`), but those events would be silently dropped.

peterporfy: That's a good catch, thank you! Even though we are willing to break the experimental API (https://github.com/RevenueCat/purchases-android/pull/3050#discussion_r2728513983), we shouldn't lose data. Fixed and added tests for it.

PR #3052, opened 2026-01-27
[Galaxy Store]: Log warning if Galaxy SDK Version Doesn't Match Expected Version

fire-at-will:
### Description
Logs a warning message if the version of the Samsung IAP SDK isn't the version expected by the RevenueCat SDK. This is helpful since developers will be responsible for providing their own copy of the Samsung IAP SDK.

### DIFF
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
@@ -11,7 +11,6 @@ import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesErrorCallback
 import com.revenuecat.purchases.PurchasesErrorCode
 import com.revenuecat.purchases.PurchasesStateProvider
-import com.revenuecat.purchases.api.BuildConfig as PurchasesBuildConfig
 import com.revenuecat.purchases.common.BillingAbstract
 import com.revenuecat.purchases.common.DateProvider
 import com.revenuecat.purchases.common.DefaultDateProvider
@@ -44,10 +43,11 @@ import com.revenuecat.purchases.models.PurchasingData
 import com.revenuecat.purchases.models.StoreTransaction
 import com.revenuecat.purchases.strings.PurchaseStrings
 import com.revenuecat.purchases.strings.RestoreStrings
-import com.samsung.android.sdk.iap.lib.BuildConfig as SamsungBuildConfig
 import com.samsung.android.sdk.iap.lib.constants.HelperDefine
 import com.samsung.android.sdk.iap.lib.helper.IapHelper
 import com.samsung.android.sdk.iap.lib.vo.PurchaseVo
+import com.revenuecat.purchases.api.BuildConfig as PurchasesBuildConfig
+import com.samsung.android.sdk.iap.lib.BuildConfig as SamsungBuildConfig
 
 @OptIn(InternalRevenueCatAPI::class)
 @Suppress("TooManyFunctions", "LongParameterList")


purchases/src/main/kotlin/com/revenuecat/purchases/common/Constants.kt
```diff
@@ -12,4 +12,5 @@ internal object Constants {
 object SharedConstants {
     const val RC_CUSTOMER_CENTER_TAG = "rc-customer-center"
     const val MICRO_MULTIPLIER = 1_000_000.0
+    const val EXPECTED_SAMSUNG_IAP_SDK_VERSION = "6.5.0.001"
```
fire-at-will: The `.001` is a bit unusual, but that's how they define their version numbers in the SDK:

```java
public class HelperConstants {
    public static final String HELPER_VERSION = "6.5.0." + String.format("%03d", 1);
    // ...
}
```

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -472,4 +475,16 @@ internal class GalaxyBillingWrapper(
...
         purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
     }
+
+    private fun logWarningIfUnexpectedSamsungIAPVersionFound() {
+        if (HelperConstants.HELPER_VERSION != SharedConstants.EXPECTED_SAMSUNG_IAP_SDK_VERSION) {
+            log(intent = LogIntent.GALAXY_WARNING) {
+                "Unexpected Samsung IAP SDK version found. Expected " +
+                    "${SharedConstants.EXPECTED_SAMSUNG_IAP_SDK_VERSION}, but found ${HelperConstants.HELPER_VERSION}" +
+                    ". Unexpected behavior related to the Galaxy Store in-app purchases may occur. Please obtain and" +
+                    " include version ${HelperConstants.HELPER_VERSION} of the Samsung IAP SDK from the Samsung's" +
```
skydoves: Do you mean below?

```suggestion
                    " include version ${SharedConstants.EXPECTED_SAMSUNG_IAP_SDK_VERSION} of the Samsung IAP SDK from the Samsung's" +
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/Constants.kt
tonidero: I wonder if we should get this from `BuildConfig` so we can use the `toml` file version directly... Though I see that one doesn't have the `.001` suffix... maybe we can add that? It's super weird though 😓

fire-at-will: I don't think we can add it directly to the `toml` file, because then the Gradle scripts will look for a `samsung-iap-sdk-6.5.0.001.aar` file, and when you download the SDK from Samsung, they don't mention the `.001` anywhere, so I think that that would cause a lot of confusion for developers. That said, I do like the idea of reading the version from the `toml` file so that we only need to bump our version number in one place when we update. Maybe we could compare versions ignoring the `.xxx` version number after the patch version?

fire-at-will: Update: I went and looked back on all 6.X releases of the Samsung IAP, and this value is correct and always uses `.001`, save for two exceptions:
- Version 6.0's `HelperDefine.HELPER_VERSION` is hardcoded to `5.5.5.001`
- Version 6.1's `HelperDefine.HELPER_VERSION` is hardcoded to `6.1.0.004`

That said, all versions since 6.1 (which was released in October of 2020), have all used the format `{SEMVER}.001`.

fire-at-will: Another update: it looks like we can use `com.samsung.android.sdk.iap.lib.BuildConfig.VERSION_NAME`, which never contains a `.xxx` suffix to the build number! 🙌 I'll see if I can wire it up so that we read our expected version number from the `toml` file

fire-at-will: @tonidero Updated! It now:
- Pulls the actual IAP SDK version from `com.samsung.android.sdk.iap.lib.BuildConfig.VERSION_NAME`, which doesn't contain the `.xxx` suffix
- Pulls the expected version number from `com.revenuecat.purchases.api.BuildConfig.SAMSUNG_IAP_SDK_VERSION`, which comes from the `toml` file

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -472,4 +473,19 @@ internal class GalaxyBillingWrapper(
     private fun onPurchaseError(error: PurchasesError) {
         purchasesUpdatedListener?.onPurchasesFailedToUpdate(error)
     }
+
+    private fun logWarningIfUnexpectedSamsungIAPVersionFound() {
+        val expectedSamsungIAPSDKVersion = com.revenuecat.purchases.api.BuildConfig.SAMSUNG_IAP_SDK_VERSION
```
tonidero: Hm mcan we do an import for these instead of the fully qualified name?

fire-at-will: Done in 94ca21cb6ab3d170144a1fed13c92c32f58d3f3b 👍

PR #3053, opened 2026-01-28
Add loading placeholder for the customer center

skydoves:
### Description

Add a loading placeholder for the customer center for better developer experiences. I would follow up on the error screen as well since it just shows a text.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/CustomerCenterLoadingView.kt
```diff
@@ -1,14 +1,158 @@
...
-    Text("Loading...")
+    val placeholderColor = MaterialTheme.colorScheme.onSurface.copy(alpha = LoadingViewConstants.PLACEHOLDER_ALPHA)
+    val fadeHighlight = Fade(
+        highlightColor = placeholderColor,
+        animationSpec = PlaceholderDefaults.fadeAnimationSpec,
+    )
+
+    Column(
+        modifier = Modifier
+            .fillMaxSize()
```
tonidero, line 55: I'm usually careful with these modifiers since it can affect the parent sizing... But looking at current usages, I think it's safe to do this 👍

PR #3054, opened 2026-01-28
Add internal API to debug tracked events

tonidero:
### Description
This provides some visibility on what events are tracked and when. This can help debug any issues with these events.

### DIFF
--- a/purchases/api-defaults-bc7.txt
+++ b/purchases/api-defaults-bc7.txt
@@ -457,7 +457,7 @@ package com.revenuecat.purchases {
     method public void setSolarEngineDistinctId(String? solarEngineDistinctId);
     method public void setSolarEngineVisitorId(String? solarEngineVisitorId);
     method public void setTenjinAnalyticsInstallationID(String? tenjinAnalyticsInstallationID);
-    method public void setTrackedEventListener(com.revenuecat.purchases.TrackedEventListener?);
+    method @kotlin.jvm.JvmSynthetic public void setTrackedEventListener(com.revenuecat.purchases.TrackedEventListener?);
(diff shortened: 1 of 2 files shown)


purchases/api-defaults-bc7.txt
```diff
@@ -397,6 +397,7 @@ package com.revenuecat.purchases {
     method public java.util.Locale? getStorefrontLocale();
     method @com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI public void getStorefrontLocale(com.revenuecat.purchases.interfaces.GetStorefrontLocaleCallback callback);
     method @Deprecated public void getSubscriptionSkus(java.util.List<java.lang.String> productIds, com.revenuecat.purchases.interfaces.GetStoreProductsCallback callback);
+    method @kotlin.jvm.JvmSynthetic public com.revenuecat.purchases.TrackedEventListener? getTrackedEventListener();
```
tonidero, line 400: I've checked and calling these APIs do require to opt in. But for some reason, metalava doesn't catch that in these delegated properties and it's adding it anyway to these (same thing happens with the CustomerCenterListener). 

Considering that people do need to opt in, I think this is ok, but lmk if you think otherwise!

JayShortway: Would it help to add `@InternalRevenueCatAPI` to the `PurchasesOrchestrator` field? (It would be pointless, but maybe it fools Metalava?)

JayShortway: If not, I think this is fine btw.

tonidero: Good thinking, but unfortunately it didn't work :( But yeah, will merge as is for now since it shouldn't be a problem

PR #3059, opened 2026-01-28
Log When Galaxy Products are Missing in Products Request's Response

fire-at-will:
### Description
Logs a warning message when a product that the RC SDK requests from the Galaxy SDK isn't returned by the Galaxy SDK.

We haven't seen this happen before, but just adding this to help out with debugging any issues in the future.

### DIFF
--- a/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
+++ b/feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
@@ -147,7 +147,7 @@ internal class ProductDataHandler(
         requestedProductIds: Set<String>?,
         products: List<ProductVo>,
     ) {
-        if (requestedProductIds == null) {
+        if (requestedProductIds == null || requestedProductIds.isEmpty()) {
             return
         }
 


feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/handler/ProductDataHandler.kt
```diff
@@ -139,6 +143,26 @@ internal class ProductDataHandler(
         onReceive?.invoke(storeProductsMatchingRequest)
     }
 
+    private fun logMissingProductIdsIfAny(
+        requestedProductIds: Set<String>?,
+        products: List<ProductVo>,
+    ) {
+        if (requestedProductIds == null) {
+            return
+        }
```
skydoves: We can return the empty set here as well. (`requestedProductIds.isEmpty()`)

fire-at-will: Good callout! Added that here: 0d1676e24d962db9aad120313a0ed9226bb1d4cb

tonidero (approved):
Nice!

PR #3061, opened 2026-01-30
CC-628: Refresh Customer Center UI after subscription cancellation

facumenzella:
## Summary
Fixes issue where Customer Center UI doesn't update after subscription cancellation.

## Changes
- Add lifecycle-aware refresh when activity resumes after being paused (when user returns from manage subscriptions screen)
- Add `isRefreshing` state to `CustomerCenterState.Success` to show loading indicator during refresh
- Add `refreshCustomerCenter()` method that keeps content visible while refreshing (shows subtle loading indicator instead of full loading screen)
- Match iOS behavior: refresh when returning from manage subscriptions screen
- Show subtle loading indicator and dim content during refresh (similar to iOS)

## Testing
- [ ] Test that Customer Center refreshes when returning from Google Play Store manage subscriptions screen
- [ ] Verify that cancelled subscriptions are shown correctly after cancellation
- [ ] Verify that loading indicator appears during refresh
- [ ] Verify that content is dimmed during refresh

Closes CC-628

### DIFF
--- a/CHANGELOG.latest.md
+++ b/CHANGELOG.latest.md
@@ -1,8 +1,10 @@
 ## RevenueCatUI SDK
 ### Paywallv2
-#### ✨ New Features
-* Add custom variables support for v2 paywalls (#3020) via Facundo Menzella (@facumenzella)
+#### 🐞 Bugfixes
+* Video's fallback image has double border (#3071) via Cesar de la Vega (@vegaro)
 
 ### 🔄 Other Changes
-* [AUTOMATIC] Update golden test files for backend integration tests (#3060) via RevenueCat Git Bot (@RCGitBot)
-* Add loading placeholder for the customer center (#3053) via Jaewoong Eum (@skydoves)
+* Rename `MultipleIntroOffers` override to `MultiplePhaseOffers` (#3069) via Cesar de la Vega (@vegaro)
+* Add LeakCanary (#3068) via Rick (@rickvdl)
+* Avoid posting stored metadata in pending transactions (#3065) via Toni Rico (@tonidero)
+* [AUTOMATIC] Update golden test files for backend integration tests (#3066) via RevenueCat Git Bot (@RCGitBot)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## 9.20.1
+## RevenueCatUI SDK
+### Paywallv2
+#### 🐞 Bugfixes
+* Video's fallback image has double border (#3071) via Cesar de la Vega (@vegaro)
+
+### 🔄 Other Changes
+* Rename `MultipleIntroOffers` override to `MultiplePhaseOffers` (#3069) via Cesar de la Vega (@vegaro)
+* Add LeakCanary (#3068) via Rick (@rickvdl)
+* Avoid posting stored metadata in pending transactions (#3065) via Toni Rico (@tonidero)
+* [AUTOMATIC] Update golden test files for backend integration tests (#3066) via RevenueCat Git Bot (@RCGitBot)
+
 ## 9.20.0
 ## RevenueCatUI SDK
 ### Paywallv2
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <meta http-equiv="refresh" content="0; url=https://sdk.revenuecat.com/android/9.20.0/index.html" />
+    <meta http-equiv="refresh" content="0; url=https://sdk.revenuecat.com/android/9.20.1/index.html" />
 </head>
 <body>
 </body>
--- a/purchases/api-defaults-bc7.txt
+++ b/purchases/api-defaults-bc7.txt
@@ -752,18 +752,16 @@ package com.revenuecat.purchases.ads.events.types {
   }
 
   @com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI @dev.drewhamilton.poko.Poko public final class AdFailedToLoadData {
-    ctor public AdFailedToLoadData(String networkName, String mediatorName, String adFormat, String? placement, String adUnitId, Integer? mediatorErrorCode);
+    ctor public AdFailedToLoadData(String mediatorName, String adFormat, String? placement, String adUnitId, Integer? mediatorErrorCode);
     method public String getAdFormat();
     method public String getAdUnitId();
     method public Integer? getMediatorErrorCode();
     method public String getMediatorName();
-    method public String getNetworkName();
     method public String? getPlacement();
     property public final String adFormat;
     property public final String adUnitId;
     property public final Integer? mediatorErrorCode;
     property public final String mediatorName;
-    property public final String networkName;
     property public final String? placement;
   }
 
--- a/purchases/api-defauts.txt
+++ b/purchases/api-defauts.txt
@@ -752,18 +752,16 @@ package com.revenuecat.purchases.ads.events.types {
   }
 
   @com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI @dev.drewhamilton.poko.Poko public final class AdFailedToLoadData {
-    ctor public AdFailedToLoadData(String networkName, String mediatorName, String adFormat, String? placement, String adUnitId, Integer? mediatorErrorCode);
+    ctor public AdFailedToLoadData(String mediatorName, String adFormat, String? placement, String adUnitId, Integer? mediatorErrorCode);
     method public String getAdFormat();
     method public String getAdUnitId();
     method public Integer? getMediatorErrorCode();
     method public String getMediatorName();
-    method public String getNetworkName();
     method public String? getPlacement();
     property public final String adFormat;
     property public final String adUnitId;
     property public final Integer? mediatorErrorCode;
     property public final String mediatorName;
-    property public final String networkName;
     property public final String? placement;
   }
 
--- a/purchases/api-entitlement.txt
+++ b/purchases/api-entitlement.txt
@@ -648,18 +648,16 @@ package com.revenuecat.purchases.ads.events.types {
   }
 
   @com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI @dev.drewhamilton.poko.Poko public final class AdFailedToLoadData {
-    ctor public AdFailedToLoadData(String networkName, String mediatorName, String adFormat, String? placement, String adUnitId, Integer? mediatorErrorCode);
+    ctor public AdFailedToLoadData(String mediatorName, String adFormat, String? placement, String adUnitId, Integer? mediatorErrorCode);
     method public String getAdFormat();
     method public String getAdUnitId();
     method public Integer? getMediatorErrorCode();
     method public String getMediatorName();
-    method public String getNetworkName();
     method public String? getPlacement();
     property public final String adFormat;
     property public final String adUnitId;
     property public final Integer? mediatorErrorCode;
     property public final String mediatorName;
-    property public final String networkName;
     property public final String? placement;
   }
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdEvent.kt
@@ -25,7 +25,7 @@ internal sealed interface AdEvent : FeatureEvent {
     val eventVersion: Int
     val type: AdEventType
     val timestamp: Long
-    val networkName: String
+    val networkName: String?
     val mediatorName: AdMediatorName
     val adFormat: AdFormat
     val placement: String?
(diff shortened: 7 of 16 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -105,6 +116,40 @@ internal fun InternalCustomerCenter(
...
+            when (event) {
+                Lifecycle.Event.ON_PAUSE -> {
+                    wasPaused = true
+                }
+                Lifecycle.Event.ON_RESUME -> {
+                    // Refresh when resuming after being paused (user returned from manage subscriptions screen)
+                    // Only refresh if we have a loaded state (to avoid refreshing during initial load)
+                    if (wasPaused && currentState is CustomerCenterState.Success) {
+                        // Reload Customer Center to reflect any subscription changes (e.g., cancellation)
+                        // that may have occurred while the user was in the Google Play Store
```
skydoves: What about adding the `!currentState.isRefreshing` condition for preventing a race condition here?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -5,13 +5,17 @@ package com.revenuecat.purchases.ui.revenuecatui.customercenter
...
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.core.tween
 import androidx.compose.foundation.background
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
```
skydoves: Unused one imported 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -388,12 +433,34 @@ private fun CustomerCenterLoaded(
...
+                onAction = onAction,
+            )
+        }
+
+        // Show loading indicator when refreshing (similar to iOS)
+        if (state.isRefreshing) {
+            CircularProgressIndicator(
+                modifier = Modifier
+                    .align(Alignment.Center)
+                    .padding(vertical = 16.dp),
```
skydoves: Seems not to have any effect (paddings) if this indicator is centralized.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -105,6 +116,40 @@ internal fun InternalCustomerCenter(
...
+    DisposableEffect(lifecycleOwner) {
+        val observer = LifecycleEventObserver { _, event ->
+            when (event) {
+                Lifecycle.Event.ON_PAUSE -> {
+                    wasPaused = true
+                }
+                Lifecycle.Event.ON_RESUME -> {
+                    // Refresh when resuming after being paused (user returned from manage subscriptions screen)
+                    // Only refresh if we have a loaded state (to avoid refreshing during initial load)
+                    if (wasPaused && currentState is CustomerCenterState.Success) {
```
skydoves: We can simplify this logic with the new Compose API, [LifecycleResumeEffect](https://developer.android.com/topic/libraries/architecture/compose#lifecycleresumeeffect). Also, the `androidx.compose.ui.platform.LocalLifecycleOwner` API has been deprecated.

```kotlin
  var wasPaused by remember { mutableStateOf(false) }
  val currentState by rememberUpdatedState(state)

  LifecycleResumeEffect(Unit) {
      // Called on ON_RESUME
      if (wasPaused && currentState is CustomerCenterState.Success) {
          coroutineScope.launch {
              viewModel.refreshCustomerCenter()
          }
      }
      wasPaused = false

      onPauseOrDispose {
          // Called on ON_PAUSE or when composable disposes
          wasPaused = true
      }
  }
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -937,15 +956,21 @@ internal class CustomerCenterViewModelImpl(
...
             }
         } catch (e: PurchasesException) {
-            _state.update {
-                CustomerCenterState.Error(e.error)
+            _state.update { currentState ->
+                if (isRefresh && currentState is CustomerCenterState.Success) {
+                    // On error during refresh, keep the existing state but clear isRefreshing
+                    currentState.copy(isRefreshing = false)
+                } else {
+                    CustomerCenterState.Error(e.error)
```
vegaro, line 986: we should probably log something here

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -105,6 +112,29 @@ internal fun InternalCustomerCenter(
         viewModel.trackImpressionIfNeeded()
     }
 
+    // Refresh Customer Center data when activity resumes after being paused
+    // This matches iOS behavior where we refresh when the manage subscriptions sheet is dismissed.
+    // When the user opens the manage subscriptions screen (Google Play Store), the activity pauses.
+    // When they return and dismiss it, the activity resumes, and we refresh to show updated subscription status.
+    var wasPaused by remember { mutableStateOf(false) }
```
vegaro: I believe this won't work since it doesn't survive configuration changes. Example:
- User opens Play Store
- User rotates while in Play Store
- `wasPaused` resets to `false`

You could use `rememberSaveable` but there's another problem with your implementation, `onPauseOrDispose` will trigger on configuration changes as well because the Activity is recreated, so this implementation is going to trigger false refreshes after any rotation.

The correct way of doing this is in the view model which does actually survive config changes. Something like this:

```kotlin
  private var wasBackgrounded = false
  fun onActivityStopped(isChangingConfigurations: Boolean) {
      if (!isChangingConfigurations) {
          wasBackgrounded = true
      }
  }
  fun onActivityStarted() {
      if (wasBackgrounded) {
          wasBackgrounded = false
          val currentState = _state.value
          if (currentState is CustomerCenterState.Success && !currentState.isRefreshing) {
              viewModelScope.launch {
                  refreshCustomerCenter()
              }
          }
      }
  }
```

  Then in `InternalCustomerCenter`, replace the `LifecycleResumeEffect` block with:
  
```kotlin
  val lifecycleOwner = LocalLifecycleOwner.current
  val activity = LocalContext.current as? Activity
  DisposableEffect(lifecycleOwner) {
      val observer = LifecycleEventObserver { _, event ->
          when (event) {
              Lifecycle.Event.ON_STOP -> {
                  viewModel.onActivityStopped(activity?.isChangingConfigurations == true)
              }
              Lifecycle.Event.ON_START -> {
                  viewModel.onActivityStarted()
              }
              else -> {}
          }
      }
      lifecycleOwner.lifecycle.addObserver(observer)
      onDispose {
          lifecycleOwner.lifecycle.removeObserver(observer)
      }
  }
```

Remember to check `isChangingConfigurations` to distinguish real backgrounding from rotation

facumenzella: I like this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
facumenzella: Added! 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
facumenzella: Great catch! Moved the logic to the ViewModel as you suggested. Now using ON_STOP/ON_START with isChangingConfigurations check.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -104,6 +104,13 @@ internal interface CustomerCenterViewModel {
 
     @InternalRevenueCatAPI
     suspend fun loadCustomerCenter()
+
+    /**
+     * Refreshes the Customer Center data while keeping the current Success state visible.
+     * Shows a subtle loading indicator instead of the full loading screen.
+     * Used when returning from external screens (e.g., manage subscriptions).
+     */
+    suspend fun refreshCustomerCenter()
```
vegaro, line 113: nit, add an empty line

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -34,10 +39,14 @@ import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalLifecycleOwner
```
vegaro: The correct import is "import androidx.lifecycle.compose.LocalLifecycleOwner"  
  
This other one is deprecated

vegaro: Or actually it might better to do @skydoves suggestion https://github.com/RevenueCat/purchases-android/pull/3061#discussion_r2756717093

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -105,6 +114,32 @@ internal fun InternalCustomerCenter(
...
     }
 
+    // Refresh Customer Center data when activity resumes after being backgrounded.
+    // This matches iOS behavior where we refresh when the manage subscriptions sheet is dismissed.
+    // When the user opens the manage subscriptions screen (Google Play Store), the activity stops.
+    // When they return, the activity starts again, and we refresh to show updated subscription status.
+    // Using ON_STOP/ON_START with isChangingConfigurations check to properly handle configuration changes
+    // (e.g., rotation) without triggering false refreshes.
+    val lifecycleOwner = LocalLifecycleOwner.current
+    val activity = context as? Activity
```
vegaro: This is a bit unsafe since `context` could be a `ContextWrapper` wrapping an `Activity`. This cast would return null even when there IS an underlying Activity. We have this in the debugview module of the project not sure if you can reuse it or maybe copy the extension in this module:

```kotlin
fun Context.findActivity(): Activity? {
      var context = this
      while (context is ContextWrapper) {
          if (context is Activity) return context
          context = context.baseContext
      }
      return null
  }
```

skydoves: Otherwise, there's a great official API that resolves this: [LocalActivity](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:activity/activity-compose/src/main/java/androidx/activity/compose/LocalActivity.kt;l=34?q=localactivity&ss=androidx%2Fplatform%2Fframeworks%2Fsupport) 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
facumenzella: Done

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
facumenzella: Fixed - using `androidx.lifecycle.compose.LocalLifecycleOwner`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
facumenzella: Using `context.getActivity()` which already exists in this module and handles ContextWrapper properly.

PR #3062, opened 2026-01-30
Add support for product changes in paywalls

vegaro:
Add support for product change (upgrade/downgrade) configuration in paywalls. When a user with an active Play Store subscription purchases a different product, the SDK now applies the appropriate Google replacement mode based on the paywall's ProductChangeConfig

## Product Change Logic
  1. Non-subscriptions: No product change (returns null)
  2. No active Play Store subscription: No product change (first subscription)
  3. Same product ID: No product change (Google handles base plan changes automatically)
  4. Different product: Compares normalized prices to detect upgrade vs downgrade
    • Production: Uses price per month for comparison
    • Sandbox: Uses price per sandbox minute (accounts for accelerated renewal times). I tested and this is how Google determines if it's an upgrade or a downgrade
    
## Price Comparison Limitations

  We use the current product price to determine whether a change is an upgrade or downgrade. This has some limitations with price changes. If product prices have changed since the original purchase, the comparison uses the new prices, not what the user originally paid.

  This is an acceptable trade-off because:
  1. Fetching the user's original purchase price requires a new API endpoint. We'll be working on it soon.
  2. Most product changes are between clearly different tiers (e.g., monthly → annual)
  3. The replacement mode defaults are conservative. Charge full price is the problematic one because it doesn't work for downgrades. We are not enabling setting that one yet in the frontend. We'll probably wait until we have that new endpoint we need. When that's enabled, the current code will work in most cases, unless there's a product change that changes the price so much that an upgrade becomes a downgrade (rare). The new code will correctly check the subscribed product price and we'll be able to correctly detect upgrades.


ui/revenuecatui/api.txt
```diff
@@ -126,7 +127,9 @@ package com.revenuecat.purchases.ui.revenuecatui {
 
   public abstract class PurchaseLogicWithCallback implements com.revenuecat.purchases.ui.revenuecatui.PurchaseLogic {
     ctor public PurchaseLogicWithCallback();
+    method public final suspend Object? performPurchase(android.app.Activity activity, com.revenuecat.purchases.Package rcPackage, com.revenuecat.purchases.paywalls.components.common.ProductChangeConfig? productChangeConfig, kotlin.coroutines.Continuation<? super com.revenuecat.purchases.ui.revenuecatui.PurchaseLogicResult>);
```
vegaro: since this is a new function, and the followup PR #3063 also adds a parameter, we should merge them both together

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ProductChangeConfig.kt
```diff
@@ -0,0 +1,95 @@
...
+    val downgradeReplacementMode: SerializableReplacementMode = SerializableReplacementMode.DEFERRED,
+)
+
+/**
+ * Serializable replacement mode enum for paywall configuration.
+ * Maps to [GoogleReplacementMode] for actual purchase operations.
+ */
+@InternalRevenueCatAPI
+@Serializable
+enum class SerializableReplacementMode {
```
tonidero: Hmm should we use the existing `GoogleReplacementMode` and use a custom serializer/deserializer instead? Not sure if there is any advantage to keeping it separate...

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ProductChangeConfig.kt
```diff
@@ -0,0 +1,95 @@
...
+    @SerialName("upgrade_replacement_mode")
+    val upgradeReplacementMode: SerializableReplacementMode = SerializableReplacementMode.CHARGE_PRORATED_PRICE,
+
+    /**
+     * Replacement mode to use for downgrades (moving to a lower price per unit time).
+     * Defaults to DEFERRED.
+     */
+    @get:JvmSynthetic
+    @SerialName("downgrade_replacement_mode")
+    val downgradeReplacementMode: SerializableReplacementMode = SerializableReplacementMode.DEFERRED,
```
tonidero: For enums I think we should use EnumDeserializerWithDefault so in case we add more types in the future, it doesn't break for existing clients, and just fallbacks to the default.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ProductChangeConfig.kt
tonidero: Also cc @fire-at-will since he's working on replacement modes for the galaxy store, so we might want to make sure we are in sync 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PaywallComponentsData.kt
```diff
@@ -40,4 +40,7 @@ class PaywallComponentsData(
     @get:JvmSynthetic
     @SerialName("exit_offers")
     val exitOffers: ExitOffers? = null,
+    @get:JvmSynthetic
+    @SerialName("play_store_product_change_mode")
+    val productChangeConfig: ProductChangeConfig? = null,
```
tonidero, line 45: Hmm I wonder, would we want to be more flexible and support different replacement modes per "old product id"? 

Just thinking, if we want to make it extensible, we could make this a list of rules with an optional "fallback" product change replacement mode. So, if the user is subscribed to any of the old product ids, the first product id from the list of rules would be applied. If none, use the fallback. Then, for now, if we don't want to make it too complex, we could just have a fallback replacement mode from the backend. 

This might be out of scope though, and I guess we can always move to a different property, so I won't block it if we prefer this simplified version for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt
```diff
@@ -0,0 +1,139 @@
...
+
+        if (oldSubscriptionId == newSubscriptionId) {
+            Logger.d("Same product ($newSubscriptionId), Google handles base plan change automatically")
+            return null
+        }
+
+        val oldProduct = purchases.awaitGetProduct(oldSubscriptionId, oldBasePlanId)
+        val isSandbox = activePlayStoreSubscription.isSandbox
+
+        val oldNormalizedPrice = oldProduct?.getNormalizedPrice(isSandbox)
```
tonidero, line 75: Would indeed be great to get this without having to fetch the current... Just to check, but the `SubscriptionInfo.price` from the `CustomerInfo` wouldn't work here? Not sure if that accounts for offers/free trials or not though, so it might be the reason we need a new endpoint?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt
```diff
@@ -0,0 +1,139 @@
...
+        val newNormalizedPrice = packageToPurchase.product.getNormalizedPrice(isSandbox)
+
+        val isUpgrade = oldNormalizedPrice != null &&
+            newNormalizedPrice != null &&
+            newNormalizedPrice > oldNormalizedPrice
```
tonidero, line 80: TBH, I'm unsure about this logic to differentiate between upgrades and downgrades... Feels like the difference shouldn't be based on the price, since those could potentially change over time. IMO, whether it's an upgrade or a downgrade should depend mostly on the developer... But if we have tested and google doesn't allow this, I think we just need to get as close to that logic as possible indeed...

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt
```diff
@@ -0,0 +1,139 @@
...
+    private suspend fun calculateProductChange(
+        activePlayStoreSubscription: SubscriptionInfo,
+        packageToPurchase: Package,
+        productChangeConfig: ProductChangeConfig,
+    ): ProductChangeInfo? {
+        val oldSubscriptionId = activePlayStoreSubscription.productIdentifier
+        val oldBasePlanId = activePlayStoreSubscription.productPlanIdentifier
+        val newProductIdentifier = packageToPurchase.product.id
+
+        val (newSubscriptionId, _) = parseProductIdentifier(newProductIdentifier)
```
tonidero: You could downcast the product to a `GoogleStoreProduct` and get the product ID from there instead... Though then you will need to handle amazon and soon galaxy products as well... Maybe this is something we need to add to the `StoreProduct` interface itself, since I've also needed to access this myself at times.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
```diff
@@ -29,6 +30,30 @@ interface PurchaseLogic {
...
+     *        the default option should be used.
+     * @param productChangeConfig Configuration for product changes (upgrades/downgrades), specifying replacement
+     *        modes. If null, no product change behavior is configured.
+     * @return A `PurchaseLogicResult` object containing the outcome of the purchase operation.
+     */
+    suspend fun performPurchase(
+        activity: Activity,
+        rcPackage: Package,
+        productChangeConfig: ProductChangeConfig?,
+    ): PurchaseLogicResult = performPurchase(activity, rcPackage)
```
tonidero: Hmm I'm not sure I fully like this... Like, devs now have the option to override this method, but they are still forced to implement the method without the `productChangeConfig` I think? 

While I also don't like it, maybe we need to follow a similar pattern to the `PurchaseLogicWithCallback` and create `PurchaseLogicWithProductChange` and `PurchaseLogicWithProductChangeAndCallback`... Or better not... 🤮 

Maybe we should create a class to be a single parameter with all that's needed and a builder-like pattern, so it's easier to add new fields. Then we would still need to add two new interfaces, but at least it would help moving forward?

tonidero (commented):
Not fully finished reviewing, but leaving a few comments so far. Looking great though!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
vegaro: hmm right. since it's an interface they will have to do implement it as well.   
  
With the builder you mean a class that we can use to hold the properties, and add more properties as needed right? Something like PurchaseParams.   
  
What's the two new interfaces? We would need a new one for the new PurchaseParams, but wha'ts the other one?  
  
Also, I am considering leaving this for another PR and just not supporting this case (completing product changes on your own) in this PR. Wdyt?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt
vegaro: Yeah the customer info will return 0 if the user is in a free trial at the moment. We need that new endpoint to know how much their base plan is

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt
vegaro: This is the logic that Google uses. Google uses the subscribed base plan to determine if it's an upgrade or a downgrade.   
  
Moving to a subscription that charges more per unit is an upgrade, the opposite is a downgrade.  
  
We are going to need the new endpoint to know the base plan of the user, but for now we use the product price, which will match in most cases.  
  
I see what you mean with letting the developer determine wether it's an upgrade or a downgrade. But to be able to do that we would need to let them define a map of all products. Or some way to set them in tiers. We explored that and it was more complex and probably better to move into the product catalog itself.   
  
Also `CHARGE_PRORATED_PRICE` is not allowed on downgrades, the developer could set it and Google still not allow it if it's a downgrade.  
  
This solution (matching as close as possible what Google does) was a middle ground.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt
vegaro: Oh that's better, that way the parsing logic is isolated

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
tonidero: > With the builder you mean a class that we can use to hold the properties, and add more properties as needed right? Something like PurchaseParams.

Yup indeed, that's what I'm thinking

> What's the two new interfaces? We would need a new one for the new PurchaseParams, but wha'ts the other one?

The same one but the "callbacks" version of it. So basically a `PurchaseLogicWithParams` and `PurchaseLogicWithParamsAndCallback` (again terrible naming)

> Also, I am considering leaving this for another PR and just not supporting this case (completing product changes on your own) in this PR. Wdyt?

I think that's ok for now yeah, it will probably get tricky, so we/should split it to a different PR indeed 👍

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PaywallComponentsData.kt
vegaro: We did discuss something like this but ended up with a simpler solution for simplicity since we believe that more flexible solution probably involves some revamp of the product catalog itself.

tonidero (approved):
This makes sense to me! Amazing job!!

PR #3063, opened 2026-01-30
Add support for promotional offers in paywalls

vegaro:
Adds the ability to configure Play Store promo offers on component-based paywalls. When a package has a playStoreOfferConfig with an offerId, we resolve that specific offer and use it for:
- Displaying offer-specific pricing variables ({{ product.offer_price }}, etc.)
- Conditional text/component overrides based on offer type
- Purchasing with the correct subscription option

The eligibility system now distinguishes between intro offers and promo offers, with single vs multiple phase variants for each. If a promo offer is misconfigured or the offer ID isn't found, we fall back to the intro offer eligibility (like in iOS), which is basically the defaultOption.

I refactored the component states to use a single `selectedPackageInfoProvider` instead of passing multiple provider lambdas around.

For testing, there's a feature flag that can be enabled in the frontend `PAYWALL_PLAY_STORE_OFFER_SETTINGS`

**⚠️ This PR is easier to review by commit**

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -259,7 +259,11 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
                     if (activity == null) {
                         Logger.e("Activity is null, not initiating package purchase")
                     } else {
-                        viewModel.handlePackagePurchase(activity, pkg = action.rcPackage)
+                        viewModel.handlePackagePurchase(
+                            activity,
+                            pkg = action.rcPackage,
+                            resolvedOffer = action.resolvedOffer,
+                        )
                     }
                 is PaywallAction.External.LaunchWebCheckout -> {
                     val url = viewModel.getWebCheckoutUrl(action)
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/LoadingPaywall.kt
@@ -33,6 +33,7 @@ import com.revenuecat.purchases.ui.revenuecatui.data.processed.PaywallTemplate
 import com.revenuecat.purchases.ui.revenuecatui.data.processed.TemplateConfiguration
 import com.revenuecat.purchases.ui.revenuecatui.data.processed.VariableDataProvider
 import com.revenuecat.purchases.ui.revenuecatui.extensions.createDefault
+import com.revenuecat.purchases.ui.revenuecatui.helpers.ResolvedOffer
 import com.revenuecat.purchases.ui.revenuecatui.helpers.ResourceProvider
 import com.revenuecat.purchases.ui.revenuecatui.helpers.isInPreviewMode
 import com.revenuecat.purchases.ui.revenuecatui.helpers.toLegacyPaywallState
@@ -213,7 +214,7 @@ private class LoadingViewModel(
         // no-op
     }
 
-    override suspend fun handlePackagePurchase(activity: Activity, pkg: Package?) {
+    override suspend fun handlePackagePurchase(activity: Activity, pkg: Package?, resolvedOffer: ResolvedOffer?) {
         // no-op
     }
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
@@ -3,6 +3,7 @@ package com.revenuecat.purchases.ui.revenuecatui.components
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.paywalls.components.ButtonComponent
 import com.revenuecat.purchases.ui.revenuecatui.components.style.ButtonComponentStyle
+import com.revenuecat.purchases.ui.revenuecatui.helpers.ResolvedOffer
 import dev.drewhamilton.poko.Poko
 
 /**
@@ -25,7 +26,10 @@ internal sealed interface PaywallAction {
 
         object RestorePurchases : External
         object NavigateBack : External
-        data class PurchasePackage(val rcPackage: Package?) : External
+        data class PurchasePackage(
+            val rcPackage: Package?,
+            val resolvedOffer: ResolvedOffer? = null,
+        ) : External
         data class LaunchWebCheckout(
             val customUrl: String?,
             val openMethod: ButtonComponent.UrlMethod,
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentState.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentState.kt
@@ -73,7 +73,10 @@ internal class ButtonComponentState(
             }
 
             is ButtonComponentStyle.Action.PurchasePackage ->
-                PaywallAction.External.PurchasePackage(rcPackage = rcPackage)
+                PaywallAction.External.PurchasePackage(
+                    rcPackage = rcPackage,
+                    resolvedOffer = resolvedOffer,
+                )
 
             is ButtonComponentStyle.Action.RestorePurchases -> PaywallAction.External.RestorePurchases
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/ButtonComponentStyle.kt
@@ -8,6 +8,7 @@ import com.revenuecat.purchases.paywalls.components.PaywallTransition
 import com.revenuecat.purchases.paywalls.components.common.LocaleId
 import com.revenuecat.purchases.paywalls.components.properties.Size
 import com.revenuecat.purchases.ui.revenuecatui.helpers.NonEmptyMap
+import com.revenuecat.purchases.ui.revenuecatui.helpers.ResolvedOffer
 import dev.drewhamilton.poko.Poko
 
 @Immutable
@@ -28,7 +29,10 @@ internal data class ButtonComponentStyle(
          * @param rcPackage The package that will be purchased by this button. Will purchase the globally-selected
          * package if this is null.
          */
-        data class PurchasePackage(val rcPackage: Package?) : Action
+        data class PurchasePackage(
+            val rcPackage: Package?,
+            val resolvedOffer: ResolvedOffer? = null,
+        ) : Action
         data class WebCheckout(
             val rcPackage: Package?,
             val autoDismiss: Boolean,
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
@@ -594,12 +594,18 @@ internal class StyleFactory(
     ): Result<ButtonComponentStyle.Action, NonEmptyList<PaywallValidationError>> {
         if (method == null) {
             return Result.Success(
-                ButtonComponentStyle.Action.PurchasePackage(rcPackage = rcPackage),
+                ButtonComponentStyle.Action.PurchasePackage(
+                    rcPackage = rcPackage,
+                    resolvedOffer = resolvedOffer,
+                ),
             )
         }
         return when (method) {
             is PurchaseButtonComponent.Method.InAppCheckout -> Result.Success(
-                ButtonComponentStyle.Action.PurchasePackage(rcPackage = rcPackage),
+                ButtonComponentStyle.Action.PurchasePackage(
+                    rcPackage = rcPackage,
+                    resolvedOffer = resolvedOffer,
+                ),
             )
 
             is PurchaseButtonComponent.Method.WebCheckout -> {
@@ -636,7 +642,10 @@ internal class StyleFactory(
             is PurchaseButtonComponent.Method.Unknown -> {
                 Logger.e("Unknown purchase button method. Defaulting to purchasing current/default package.")
                 Result.Success(
-                    ButtonComponentStyle.Action.PurchasePackage(rcPackage = rcPackage),
+                    ButtonComponentStyle.Action.PurchasePackage(
+                        rcPackage = rcPackage,
+                        resolvedOffer = resolvedOffer,
+                    ),
                 )
             }
         }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
@@ -35,8 +35,10 @@ import com.revenuecat.purchases.ui.revenuecatui.components.PaywallAction
 import com.revenuecat.purchases.ui.revenuecatui.data.processed.TemplateConfiguration
 import com.revenuecat.purchases.ui.revenuecatui.data.processed.VariableDataProvider
 import com.revenuecat.purchases.ui.revenuecatui.errors.PaywallValidationError
+import com.revenuecat.purchases.ui.revenuecatui.extensions.calculateOfferEligibility
 import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
 import com.revenuecat.purchases.ui.revenuecatui.helpers.PaywallValidationResult
+import com.revenuecat.purchases.ui.revenuecatui.helpers.ResolvedOffer
 import com.revenuecat.purchases.ui.revenuecatui.helpers.ResourceProvider
 import com.revenuecat.purchases.ui.revenuecatui.helpers.createLocaleFromString
 import com.revenuecat.purchases.ui.revenuecatui.helpers.fallbackPaywall
@@ -83,7 +85,7 @@ internal interface PaywallViewModel {
      * Note: This method requires the context to be an activity or to allow reaching an activity
      */
     fun purchaseSelectedPackage(activity: Activity?)
-    suspend fun handlePackagePurchase(activity: Activity, pkg: Package?)
+    suspend fun handlePackagePurchase(activity: Activity, pkg: Package?, resolvedOffer: ResolvedOffer? = null)
 
     fun restorePurchases()
     suspend fun handleRestorePurchases()
@@ -395,7 +397,7 @@ internal class PaywallViewModelImpl(
         finishAction()
     }
 
-    override suspend fun handlePackagePurchase(activity: Activity, pkg: Package?) {
+    override suspend fun handlePackagePurchase(activity: Activity, pkg: Package?, resolvedOffer: ResolvedOffer?) {
         if (verifyNoActionInProgressOrStartAction()) {
             return
         }
@@ -409,12 +411,12 @@ internal class PaywallViewModelImpl(
                 )
             }
             is PaywallState.Loaded.Components -> {
-                // Purchase the provided package if not null, otherwise purchase the selected package.
                 val selectedPackageInfo = pkg?.let {
                     PaywallState.Loaded.Components.SelectedPackageInfo(
                         rcPackage = it,
+                        resolvedOffer = resolvedOffer,
                         uniqueId = it.identifier,
-                        offerEligibility = currentState.selectedOfferEligibility,
+                        offerEligibility = calculateOfferEligibility(resolvedOffer, it),
                     )
                 } ?: currentState.selectedPackageInfo
                 val productChangeConfig = currentState.offering.paywallComponents?.data?.productChangeConfig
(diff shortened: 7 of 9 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PlayStoreOfferConfig.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * Configuration for a specific Play Store Offer to use for a package in a paywall.
+ * Similar to iOS's applePromoOfferProductCode.
+ */
+@InternalRevenueCatAPI
+@Poko
+@Serializable
+class PlayStoreOfferConfig(
```
tonidero: I wonder if we should make this more generic to consider other stores, even if for now we will only support the Play Store... Like `PromoOfferConfig` maybe?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PromoOfferResolver.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.Package
+import com.revenuecat.purchases.models.GoogleSubscriptionOption
+import com.revenuecat.purchases.models.SubscriptionOption
+import com.revenuecat.purchases.paywalls.components.common.PlayStoreOfferConfig
+
+/**
+ * Result of resolving a Play Store offer configuration.
+ */
+@InternalRevenueCatAPI
```
tonidero: Hmm since this is part of the UI library? Shouldn't it just be `internal` instead? Same for the resolver below.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PromoOfferResolver.kt
```diff
@@ -0,0 +1,105 @@
...
+     * @return The resolved offer result
+     */
+    fun resolve(
+        rcPackage: Package,
+        offerConfig: PlayStoreOfferConfig?,
+    ): ResolvedOffer {
+        val subscriptionOptions = rcPackage.product.subscriptionOptions
+        val defaultOption = rcPackage.product.defaultOption
+
+        if (subscriptionOptions == null || offerConfig == null) {
```
tonidero: Hmm shouldn't the case of `offeringConfig != null && subscriptionOptions == null` be a `ResolvedOffer.ConfigurationError`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PromoOfferResolver.kt
```diff
@@ -0,0 +1,105 @@
...
+
+        val configuredOption = findOfferById(subscriptionOptions, offerConfig.offerId)
+        return if (configuredOption != null) {
+            ResolvedOffer.ConfiguredOffer(configuredOption)
+        } else {
+            Logger.w(
+                "Configured Play Store offer '${offerConfig.offerId}' not found for package " +
+                    "'${rcPackage.identifier}'. Falling back to default option.",
+            )
+            ResolvedOffer.ConfigurationError(
```
tonidero: Just a note that this may happen if the user is not eligible for an offer as well. As in, if the developer configures a "normal" offer, as opposed to a "developer determined offer", a user may be eligible or not. 
Just a heads up since I'm not sure if we should consider this strictly a "ConfigurationError"

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/extensions/PackageExtensions.kt
```diff
@@ -3,25 +3,52 @@ package com.revenuecat.purchases.ui.revenuecatui.extensions
...
+
+        return when (phaseCount) {
+            1 -> OfferEligibility.IntroOfferSingle
+            2 -> OfferEligibility.IntroOfferMultiple
+            else -> OfferEligibility.Ineligible
+        }
+    }
+
+private fun ResolvedOffer.promoOfferEligibility(): OfferEligibility {
+    val phaseCount = (subscriptionOption?.pricingPhases?.size ?: 0) - 1
```
tonidero: Nitpick but maybe rename to make sure it's clear this refers to phases for offers? 
```suggestion
    val offerPhaseCount = (subscriptionOption?.pricingPhases?.size ?: 0) - 1
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PromoOfferResolver.kt
vegaro: yeah good point. I think it's better to rename this `OfferNotFound`. That way it means both that the user is not eligible and that there's a misconfiguration and we couldn't find the offer

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentState.kt
```diff
@@ -49,45 +47,49 @@ private fun rememberUpdatedCarouselComponentState(
...
-            style.rcPackage.identifier == selectedPackageProvider()?.identifier
-        } else if (style.tabIndex != null) {
-            style.tabIndex == selectedTabIndexProvider()
-        } else {
-            false
+        val selectedInfo = selectedPackageInfoProvider()
+        when {
+            style.packageUniqueId != null -> style.packageUniqueId == selectedInfo?.uniqueId
+            style.rcPackage != null -> style.rcPackage.identifier == selectedInfo?.rcPackage?.identifier
+            style.tabIndex != null -> style.tabIndex == selectedTabIndexProvider()
```
tonidero: Just in case you haven't, we should make sure to test tabs and packages inside and outside those packages with this new change.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
```diff
@@ -70,29 +66,36 @@ internal class IconComponentState(
...
     private var windowSize by mutableStateOf(initialWindowSize)
     private var layoutDirection by mutableStateOf(initialLayoutDirection)
+
     private val selected by derivedStateOf {
+        val selectedInfo = selectedPackageInfoProvider()
+        when {
+            style.rcPackage != null -> style.rcPackage.identifier == selectedInfo?.rcPackage?.identifier
+            style.tabIndex != null -> style.tabIndex == selectedTabIndexProvider()
+            else -> false
+        }
```
tonidero: Something I noticed is that for some component states, we check the `packageUniqueId` but not for others. Can you explain the difference? 

Separately, it might make sense to unify this logic in a single place? Since it seems a bit repeated... even if we should check only in some places, we could probably configure that with a parameter?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
```diff
@@ -70,29 +66,36 @@ internal class IconComponentState(
...
+    }
+
+    private val offerEligibility by derivedStateOf {
```
tonidero: The logic for this also seems a bit repeated, so it would be great to extract it somewhere.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/image/ImageComponentState.kt
```diff
@@ -38,31 +37,31 @@ import com.revenuecat.purchases.ui.revenuecatui.components.ktx.toShape
...
-        selectedTabIndexProvider = { paywallState.selectedTabIndex },
-    )
+): ImageComponentState = rememberUpdatedImageComponentState(
+    style = style,
+    localeProvider = { paywallState.locale },
+    selectedPackageInfoProvider = { paywallState.selectedPackageInfo },
+    selectedTabIndexProvider = { paywallState.selectedTabIndex },
+)
 
+@Suppress("LongParameterList")
```
tonidero: Nitpick, we didn't add more parameters here, so seems this is unnecessary now?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/IconComponentStyle.kt
```diff
@@ -52,6 +53,18 @@ internal class IconComponentStyle(
      */
     @get:JvmSynthetic
     val rcPackage: Package?,
+    /**
+     * Unique identifier for the package, combining package ID and offer ID.
+     * Used for selection comparison when multiple components reference the same package with different offers.
+     */
+    @get:JvmSynthetic
+    val packageUniqueId: String? = null,
```
tonidero: This seems currently unused, but might be because we're not adding it to the `selected` property calculation in the `State` class?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/IconComponentStyle.kt
```diff
@@ -52,6 +53,18 @@ internal class IconComponentStyle(
...
+    /**
+     * The resolved offer for this package, containing the subscription option and promo offer status.
+     * Used to determine offer eligibility and pricing phase information.
+     */
+    @get:JvmSynthetic
+    val resolvedOffer: ResolvedOffer? = null,
```
tonidero: I'm thinking whether it could be worth moving these properties to the common `ComponentStyle` interface? Not sure if there are any styles that don't actually need this, but seems like it's pretty common to all.

In this sense, it could be worth creating a base `ComponentState` class with some of the shared logic we have for these packages...

These are mostly cleanup refactors... I think they could be nice, but I wouldn't say it's a blocker.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/CarouselComponentStyle.kt
```diff
@@ -56,6 +57,18 @@ internal data class CarouselComponentStyle(
      */
     @get:JvmSynthetic
     val rcPackage: Package?,
+    /**
+     * Unique identifier for the package, combining package ID and offer ID.
+     * Used for selection comparison when multiple components reference the same package with different offers.
+     */
+    @get:JvmSynthetic
+    val packageUniqueId: String? = null,
```
tonidero: I'm thinking... since the unique ID is basically obtained from the package + offer and we have both on these already. Couldn't we just get this as a computed property instead of as a constructor parameter?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedPartial.kt
```diff
@@ -105,14 +104,17 @@ private fun <T : PresentedPartial<T>> PresentedOverride<T>.shouldApply(
...
             }
             ComponentOverride.Condition.IntroOffer -> {
-                if (introOfferEligibility == IntroOfferEligibility.INELIGIBLE) return false
+                if (!offerEligibility.isSinglePhaseIntroOffer) return false
             }
             ComponentOverride.Condition.Selected -> {
                 if (state != ComponentViewState.SELECTED) return false
             }
+            ComponentOverride.Condition.PromoOffer -> {
+                if (!offerEligibility.isSinglePhasePromoOffer) return false
```
tonidero: Hmm are we supporting multiple phase promo offers? If so, should this account for that?

TBH, I think it's a bit complex to split between "intro offers" and "promo offers" in these conditionals... But not sure if there are any use cases to splitting them up 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
```diff
@@ -29,6 +30,28 @@ interface PurchaseLogic {
...
+     * @param activity The current Android `Activity` triggering the purchase.
+     * @param rcPackage The package representing the in-app product that the user intends to purchase.
+     * @param subscriptionOption The specific subscription option to purchase (e.g., a promo offer). If null,
+     *        the default option should be used.
+     * @return A `PurchaseLogicResult` object containing the outcome of the purchase operation.
+     */
+    suspend fun performPurchase(
+        activity: Activity,
+        rcPackage: Package,
+        subscriptionOption: SubscriptionOption?,
```
tonidero: We talked about this in the other PR, but this will still force devs to implement the `performPurchase` without a `subscriptionOption` parameter...

Happy to leave to a separate PR if you'd like (might be better). But one alternative in this case could be to add the parameter as a nullable optional parameter? It might still break in Java though 😢

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/carousel/CarouselComponentState.kt
vegaro: I haven't tested tabs that deeply so thanks for the reminder 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PresentedPartial.kt
vegaro: yeah since it's a bit complex, we have decided to merge multiple phases promo and intro offers in the `MultiplePhaseOffers` override. `offerEligibility.hasMultipleDiscountedPhases` will return true for both `IntroOfferMultiple` and `PromoOfferMultiple`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
vegaro: removing it from here yes! I just removed it from the other one. That way we consolidate the change into one single PR

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
vegaro: It's a bug. Good catch!

vegaro: due to not extracitng this somewhere, so I will of ways of applying it everywhere 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
vegaro: Do you know why in the old code for `selected` was also repeated in each state instead of being passed to the view?  It feels a bit wasteful to recompute it in every component right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/IconComponentStyle.kt
vegaro: yes, I forgot. It was a bug 😄 same in `ImageComponentState`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
vegaro: This is what I am thinking....

```kotlin
internal interface PackageContext {
    /** The package this component is associated with, or null if outside a package scope. */
    val rcPackage: Package?
    /** Unique identifier combining package ID and offer ID, for distinguishing same-package components with different offers. */
    val packageUniqueId: String?
    /** The tab index this component is associated with, or null if not in a tab control. */
    val tabIndex: Int?
    /** The resolved offer for this package, containing subscription option and promo offer status. */
    val resolvedOffer: ResolvedOffer?
    /** Pre-computed offer eligibility for applying intro/promo offer conditional overrides. */
    val offerEligibility: OfferEligibility?
}
```

Then have all the styles that already have a package like `IconComponentStyle`, TextComponentStyle, etc. implement it. That's basically all styles but a couple of them. It's basically the ones that can change with the selection of a package.

Then I can move `computeIsSelected` and `resolveOfferEligibility` to the interface, and call from the states that need it like this:

```
private val isSelected by derivedStateOf {
        style.computeIsSelected(
            context = style,
            selectedPackageInfo = selectedPackageInfoProvider(),
            selectedTabIndex = selectedTabIndexProvider(),
        )
    }

    private val offerEligibility by derivedStateOf {
        style.resolveOfferEligibility(
            context = style,
            selectedOfferEligibility = selectedOfferEligibilityProvider(),
        )
    }
```

wdyt?

tonidero: That makes sense to me! I think that should indeed avoid all the repetition

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
```diff
@@ -70,29 +68,31 @@ internal class IconComponentState(
...
-    private val selected by derivedStateOf {
-        if (style.rcPackage != null) {
-            style.rcPackage.identifier == selectedPackageProvider()?.identifier
-        } else if (style.tabIndex != null) {
-            style.tabIndex == selectedTabIndexProvider()
-        } else {
-            false
-        }
+
+    private val isSelected by derivedStateOf {
```
tonidero: One idea I just had... Wdyt of having something like a `PackageAwareComponentState` interface like:
```
internal interface PackageAwareComponentState {
   // To be implemented by implementers
   val selectedPackageInfoProvider: () -> PaywallState.Loaded.Components.SelectedPackageInfo?
   val selectedOfferEligibilityProvider: () -> OfferEligibility

   // Already implemented
   val isSelected by derivedStateOf { 
       style.computeIsSelected(
            selectedPackageInfo = selectedPackageInfoProvider(),
            selectedTabIndex = selectedTabIndexProvider(),
        )
   }
   val offerEligibility by derivedStateOf {
        style.resolveOfferEligibility(
            selectedOfferEligibility = selectedOfferEligibilityProvider(),
        )
   }
}
```

Then, each state class should implement this interface, then just add `override`s to the `selectedPackageInfoProvider` and `selectedOfferEligibilityProvider`. This would allow to have this logic centralized in a single place... This is basically the same idea you applied for the styles, but for the `State` classes instead. Wdyt? Not a blocker I think though, just a way to avoid more repetition of this bit.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PromoOfferConfig.kt
```diff
@@ -0,0 +1,23 @@
...
+@Poko
+@Serializable
+class PromoOfferConfig(
+    /**
+     * The offer identifier to use for this package.
+     * On Android, this should match the offer ID configured in Google Play Console.
+     */
+    @get:JvmSynthetic
+    @SerialName("offer_id")
+    val offerId: String,
```
tonidero, line 30: I mentioned this before, but for Galaxy purchases (which are in a PR right now) we might want to consider if we can support those here, and if so how... but I think it's not a blocker for this PR.

tonidero (approved):
Just a possible small improvement, but not a blocker. I think this looks great. Thanks! So great to finally support this for Google 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
vegaro: amazing! this is exactly what I wanted to do

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PromoOfferConfig.kt
vegaro: hmm, right. I think when that time comes we can expand this one to include a store type maybe. @fire-at-will how do samsung offers look, are they just ids?

tonidero: Right, wonder if we should make errors when deserializing `PromoOfferConfig` not cause the overall paywall to fail deserializing and just default to null... Just in case we need to support other formats in the future... But NABD, we can probably figure out alternatives later as well.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/iconcomponent/IconComponentState.kt
vegaro: done in [81023c2](https://github.com/RevenueCat/purchases-android/pull/3063/commits/81023c2e21b560e4a8ba2e57f508c481e9933a51)

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/PromoOfferConfig.kt
vegaro: oh, adding a `type` here would break deserialization?

tonidero: Nope, but removing the `offer_id` (in case it doesn't exist, or it's not a `string`), could.

vegaro: I actually really like the of preventing the whole paywall to break if there's something not good here, so I will add a `ResilientPromoOfferConfigSerializer`. I think that's a good defensive thing to do

PR #3065, opened 2026-02-02
Avoid posting stored metadata in pending transactions

tonidero:
### Description
With recent tracking changes, we now store transaction metadata so we can post it again later in case something goes wrong. However, there was a path that could cause us to be posting metadata for a pending transaction. For pending transactions, we want to wait until they are final before actually posting them. This PR fixes that case so now we won't be posting transaction metadata for these pending transactions.


purchases/src/main/kotlin/com/revenuecat/purchases/PostReceiptHelper.kt
```diff
@@ -146,12 +146,14 @@ internal class PostReceiptHelper(
     fun postRemainingCachedTransactionMetadata(
         appUserID: String,
         allowSharingPlayStoreAccount: Boolean,
+        pendingTransactionsTokens: Set<String>,
         onNoTransactionsToSync: () -> Unit,
         onError: ((PurchasesError) -> Unit),
         onSuccess: ((CustomerInfo) -> Unit),
     ) {
         val results: ConcurrentLinkedQueue<Result<CustomerInfo, PurchasesError>> = ConcurrentLinkedQueue()
         val transactionMetadataToSync = localTransactionMetadataStore.getAllLocalTransactionMetadata()
```
rickvdl, line 155: If the spots in the file above are the only places we're calling `postRemainingCachedTransactionMetadata` we could also just pass the filtered `transactionMetadataToSync` to it rather than pass the tokens to ignore?

Not a strong opinion though :)

tonidero: We could... but then we would need to make the `localTransactionMetadataStore` a dependency of `PostPendingTransactionsHelper` so we can fetch and filter it there. 

The `PostPendingTransactionsHelper` for now is pretty agnostic to the changes we've done to store the transaction metadata, and most of these changes happen inside this `PostReceiptHelper` so I think I prefer to keep that dependency localized here. I think the current place is much easier but lmk if you have other idea!

PR #3068, opened 2026-02-02
Add LeakCanary

rickvdl:
Adds LeakCanary to test apps (Paywall Tester & PurchaseTester) and adds `assertNoLeaks()` to integration tests in order to catch memory leaks.


ajpallares (approved):
Nice!

PR #3069, opened 2026-02-02
Rename `MultipleIntroOffers` override to `MultiplePhaseOffers`

vegaro:
Internal change extracted from #3063 to reduce the amount of files changed in that PR

Renames to `MultiplePhaseOffers` since, `MultipleIntroOffers` doesn't actually make much sense. `MultiplePhaseOffers` is triggered when a `SubscriptionOption` has both a free trial and a discounted price phase.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
@@ -30,7 +30,7 @@ class ComponentOverride<T : PartialComponent>(
         object IntroOffer : Condition
 
         @Serializable
-        object MultipleIntroOffers : Condition
+        object MultiplePhaseOffers : Condition
 
         @Serializable
         object Selected : Condition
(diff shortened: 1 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/ComponentOverride.kt
```diff
@@ -30,7 +30,7 @@ class ComponentOverride<T : PartialComponent>(
         object IntroOffer : Condition
 
         @Serializable
-        object MultipleIntroOffers : Condition
+        object MultiplePhaseOffers : Condition
```
tonidero, line 33: TBH, I don't really feel strongly about one way or another... But I do agree it's closer to Google's nomenclature 👍

vegaro: Yeah, main reason is that this override is going to be used for both promo and intro offers. Which is also something we are probably going to move away from soon with Conditional Visibility since it doesn't really represent Google subscription system... But it felt less confusing

vegaro:
### Merge activity

* **Feb 3, 8:36 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/RevenueCat/purchases-android/3069).
* **Feb 3, 8:36 AM UTC**: @vegaro merged this pull request with [Graphite](https://app.graphite.com/github/pr/RevenueCat/purchases-android/3069).

Issue #3058, opened 2026-01-28
Very poor paywall UI performance when using multiple auto-playing carousel videos


jensck:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**

When displaying a Paywalls V2 paywall with a carousel containing 11 pages, each with an auto-playing looped video (muted), the entire UI becomes unresponsive on lower-end devices. Tapping the Close button takes several seconds to register. The main thread appears to be blocked.

1. **Environment**
   1. Platform: Android
   2. SDK version: 9.19.3
   3. OS version: Android 15 (issue reproduces), Android 13 & 16 (no issue)
   4. Android Studio version: (N/A)
   5. How widespread is the issue: Reproduced on budget device (Samsung Galaxy A14). Not reproduced on Pixel Fold or Pixel 4a. Likely affects lower-end hardware.

2. **Debug logs**
   
   No relevant output observed in logs during the freeze.

3. **Steps to reproduce**
   1. Create a Paywalls V2 paywall with an 11-page carousel
   2. Configure each page with a featured video set to auto-play, looped, muted
   3. Display the paywall on a budget device (tested: Samsung Galaxy A14)
   
   **Expected:** UI remains responsive; Close button responds immediately
   
   **Actual:** UI freezes; Close button takes 3-5+ seconds to respond

4. **Other information**
   
   **Workaround found:** Reducing the carousel to 3 pages produced acceptable performance on the A14.
   
   **Hypothesis:** Possibly all carousel videos are being initialized/decoded concurrently rather than lazily for only the visible page, causing memory pressure or main-thread blocking on resource-constrained devices.

   | Device | Android | Result |
   |--------|---------|--------|
   | Pixel Fold | 16 | ✅ Acceptable |
   | Pixel 4a | 13 | ✅ Acceptable |
   | Galaxy A14 | 15 | ❌ Unresponsive |

PR #3070, opened 2026-02-04
Improve carousel performance with videos

vegaro:
We got a report of videos performing very poorly in carousels -> #3058 

Befor --> https://drive.google.com/file/d/13re2sOIGsBcVLZrywRol8Qu4Cs8COdf_/view?usp=sharing
After --> https://drive.google.com/file/d/16uByiOEZSlcbnGINvJp72puWjB04BIj3/view?usp=sharing

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
@@ -55,7 +55,7 @@ internal fun VideoComponentView(
         val composeShape by remember(videoState.shape) { derivedStateOf { videoState.shape ?: RectangleShape } }
 
         var isVisible by remember { mutableStateOf(false) }
-        var videoReady by remember { mutableStateOf(false) }
+        var videoReady by remember(isVisible) { mutableStateOf(false) }
         val view = LocalView.current
 
         // Fallback style - always available so it shows while scrolling


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -74,40 +95,67 @@ internal fun VideoComponentView(
...
-                    ignoreTopWindowInsets = style.ignoreTopWindowInsets,
-                ),
+    // Fallback style without border/shadow/margin/padding - parent Box handles those
+    val fallbackImageViewStyle: ImageComponentStyle? = remember(style.fallbackSources) {
+        style.fallbackSources?.let { sources ->
+            ImageComponentStyle(
+                sources = sources,
+                visible = style.visible,
+                size = style.size,
+                padding = PaddingValues(0.dp),
```
vegaro, line 146: I extracted this fix to #3071

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -55,15 +72,19 @@ internal fun VideoComponentView(
                 .padding(videoState.margin)
                 .applyIfNotNull(shadowStyle) { shadow(it, composeShape) }
                 .clip(composeShape)
-                .applyIfNotNull(borderStyle) { border(it, composeShape).padding(it.width) },
+                .applyIfNotNull(borderStyle) { border(it, composeShape).padding(it.width) }
+                .onGloballyPositioned { coordinates ->
+                    isVisible = coordinates.isVisibleInWindow(view.width, view.height)
```
JayShortway: We might be able to use `coordinates.boundsInWindow()` to infer visibility without having to use the `LocalView`? Not a big deal though.

jensck:
As the original reporter of #3058, I've tested this fix locally and can confirm it resolves the issue. We're eager to see this included in an upcoming SDK release 🙌

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -76,41 +113,48 @@ internal fun VideoComponentView(
...
+internal fun Rect.isVisibleInViewport(viewportWidth: Int, viewportHeight: Int): Boolean {
+    return right > 0 && bottom > 0 && left < viewportWidth && top < viewportHeight
+}
+
 @Composable
 private fun rememberVideoContentState(
     style: VideoComponentStyle,
     videoUrls: VideoUrls,
     repository: FileRepository,
 ): Pair<URI?, ImageComponentStyle?> {
```
facumenzella, line 138: Are you still using the `ImageComponentStyle`?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -76,41 +113,48 @@ internal fun VideoComponentView(
...
+            if (fallbackStyle != null && !videoReady) {
+                ImageComponentView(fallbackStyle, state)
+            }
         }
     }
 }
 
+@JvmSynthetic
+internal fun Rect.isVisibleInViewport(viewportWidth: Int, viewportHeight: Int): Boolean {
+    return right > 0 && bottom > 0 && left < viewportWidth && top < viewportHeight
```
tonidero, line 130: Should we also consider partial visibility for this? Seems now we will only change to the video once it's totally visible, correct?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -43,12 +47,45 @@ internal fun VideoComponentView(
...
+
     if (videoState.visible) {
         val overlay = videoState.overlay?.forCurrentTheme
         val borderStyle = videoState.border?.let { rememberBorderStyle(border = it) }
         val shadowStyle = videoState.shadow?.let { rememberShadowStyle(shadow = it) }
         val composeShape by remember(videoState.shape) { derivedStateOf { videoState.shape ?: RectangleShape } }
-        val (videoUrl, fallbackImageViewStyle) = rememberVideoContentState(style, videoState.videoUrls, repository)
+
+        var isVisible by remember { mutableStateOf(false) }
+        var videoReady by remember { mutableStateOf(false) }
```
facumenzella: is this ever reset? I guess it's not a problem but  just in case 🤔

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
vegaro: yeah for other stuff like the size and the visibility

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
vegaro: This is actually considering it already. If it's partially visible from the right for example, the right will be `> 0`, left `< viewportWidth`. See this table with all the cases:

For a screen 1000x2000
Conditions are:
 ```
 right > 0
 bottom > 0
 left < 1000
 top < 2000  
```
 VISIBLE (returns `true`)

  | Situation                     | Bounds                                     | Check       | 
  |-------------------------------|--------------------------------------------|-------------|
  | Fully visible                 | left=100, top=100, right=500, bottom=500   | All pass ✓ | 
  | Partially visible from top    | left=100, top=-100, right=500, bottom=200  | All pass ✓ | 
  | Partially visible from bottom | left=100, top=1800, right=500, bottom=2200 | All pass ✓ | 
  | Partially visible from left   | left=-100, top=100, right=200, bottom=500  | All pass ✓ | 
  | Partially visible from right  | left=800, top=100, right=1200, bottom=500  | All pass ✓ | 


  ────────────────────────────────────────

  NOT VISIBLE (returns `false`)

  | Situation        | Bounds                                     | Failing check  | 
  |------------------|--------------------------------------------|----------------|
  | Off-screen above | left=100, top=-500, right=500, bottom=-100 | bottom > 0 ✗  | 
  | Off-screen below | left=100, top=2100, right=500, bottom=2500 | top < 2000 ✗  | 
  | Off-screen left  | left=-500, top=100, right=-100, bottom=500 | right > 0 ✗   | 
  | Off-screen right | left=1100, top=100, right=1500, bottom=500 | left < 1000 ✗ |

tonidero: Ohhh yup 🤦 yeah makes total sense. Thanks for explaining!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
vegaro: Good catch! this was actually causing it to show a black flash when coming back into the screen because it was not showing the fallback image. I fixed it by keying it on isVisible so it resets when that changes.

PR #3071, opened 2026-02-04
Video's fallback image has double border

vegaro:
While working on #3070 I noticed the fallback image applies the border as well

image

After the fix:

image

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.ui.revenuecatui.components.video
 
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
```diff
@@ -93,11 +95,12 @@ private fun rememberVideoContentState(
...
-                    shape = style.shape,
-                    border = style.border,
-                    shadow = style.shadow,
+                    // parent Box already handles padding, border, etc
+                    padding = PaddingValues(0.dp),
+                    margin = PaddingValues(0.dp),
+                    shape = null,
+                    border = null,
+                    shadow = null,
                     overlay = style.overlay,
```
tonidero, line 104: Would the overlay also have a chance to be applied twice? 

I guess the ideal fix would be to make the fallback behave like a StackComponent, and have it configured like that from the backend... But I think this might indeed be the best short term fix for now 👍

tonidero (approved):
Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoComponentView.kt
vegaro: I don't think so, the `Box` doesn't have the `overlay` and it's only applied directly to the video view. Not sure why, but I'd prefer not to touch that just in case it was applied directly to the video for any reason 🤔

PR #3075, opened 2026-02-06
Add customVariables support to all PaywallActivityLauncher methods and PaywallView

facumenzella:
## Summary

Adds `customVariables` support to all paywall launch methods and views:

### PaywallActivityLauncher
- `launch(offering, ...)` - already had support
- `launch(offeringIdentifier, presentedOfferingContext, ...)` - internal API
- `launchIfNeeded(requiredEntitlementIdentifier, offering, ...)` - public API
- `launchIfNeeded(requiredEntitlementIdentifier, offeringIdentifier, presentedOfferingContext, ...)` - internal API
- `launchIfNeeded(offering, shouldDisplayBlock)` - public API

### PaywallView
- Added `customVariables` parameter to the programmatic constructor
- Added `setCustomVariables()` method for updating variables after initialization

This enables hybrid SDKs (Flutter, React Native, etc.) and native users to pass custom variables through all launch paths.

## Test plan

- [x] Verify existing API tester compiles for both Kotlin and Java
- [ ] Test custom variables are passed through to the paywall in all launch methods
- [ ] Test custom variables work with PaywallView

### DIFF
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallActivityLauncherAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallActivityLauncherAPI.java
@@ -79,20 +79,27 @@ final class PaywallActivityLauncherAPI {
                 .build();
         launcher.launchWithOptions(options);
 
-        // LaunchIfNeeded with requiredEntitlementIdentifier
+        // LaunchIfNeeded with requiredEntitlementIdentifier (all builder methods)
         PaywallActivityLaunchIfNeededOptions optionsWithEntitlement = new PaywallActivityLaunchIfNeededOptions.Builder()
                 .setRequiredEntitlementIdentifier("premium")
                 .setOffering(offering)
+                .setFontProvider(fontProvider)
+                .setShouldDisplayDismissButton(true)
+                .setEdgeToEdge(true)
                 .setCustomVariables(customVariables)
                 .setPaywallDisplayCallback(paywallDisplayCallback)
                 .build();
         launcher.launchIfNeededWithOptions(optionsWithEntitlement);
 
-        // LaunchIfNeeded with shouldDisplayBlock
+        // LaunchIfNeeded with shouldDisplayBlock (all builder methods)
         PaywallActivityLaunchIfNeededOptions optionsWithBlock = new PaywallActivityLaunchIfNeededOptions.Builder()
                 .setShouldDisplayBlock(customerInfo -> customerInfo.getEntitlements().getActive().isEmpty())
                 .setOffering(offering)
+                .setFontProvider(fontProvider)
+                .setShouldDisplayDismissButton(true)
+                .setEdgeToEdge(true)
                 .setCustomVariables(customVariables)
+                .setPaywallDisplayCallback(paywallDisplayCallback)
                 .build();
         launcher.launchIfNeededWithOptions(optionsWithBlock);
     }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLaunchOptions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLaunchOptions.kt
@@ -6,6 +6,7 @@ import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.ui.revenuecatui.CustomVariableKeyValidator
 import com.revenuecat.purchases.ui.revenuecatui.CustomVariableValue
+import com.revenuecat.purchases.ui.revenuecatui.OfferingSelection
 import com.revenuecat.purchases.ui.revenuecatui.fonts.ParcelizableFontProvider
 
 /**
@@ -36,6 +37,20 @@ class PaywallActivityLaunchOptions private constructor(
     internal val offeringIdentifier: String?,
     internal val presentedOfferingContext: PresentedOfferingContext?,
 ) {
+    internal fun toOfferingSelection(): OfferingSelection.IdAndPresentedOfferingContext? {
+        return when {
+            offeringIdentifier != null -> OfferingSelection.IdAndPresentedOfferingContext(
+                offeringId = offeringIdentifier,
+                presentedOfferingContext = presentedOfferingContext,
+            )
+            offering != null -> OfferingSelection.IdAndPresentedOfferingContext(
+                offeringId = offering.identifier,
+                presentedOfferingContext = offering.availablePackages.firstOrNull()?.presentedOfferingContext,
+            )
+            else -> null
+        }
+    }
+
     /**
      * Builder for creating [PaywallActivityLaunchOptions].
      */
@@ -176,6 +191,20 @@ class PaywallActivityLaunchIfNeededOptions private constructor(
     internal val offeringIdentifier: String?,
     internal val presentedOfferingContext: PresentedOfferingContext?,
 ) {
+    internal fun toOfferingSelection(): OfferingSelection.IdAndPresentedOfferingContext? {
+        return when {
+            offeringIdentifier != null -> OfferingSelection.IdAndPresentedOfferingContext(
+                offeringId = offeringIdentifier,
+                presentedOfferingContext = presentedOfferingContext,
+            )
+            offering != null -> OfferingSelection.IdAndPresentedOfferingContext(
+                offeringId = offering.identifier,
+                presentedOfferingContext = offering.availablePackages.firstOrNull()?.presentedOfferingContext,
+            )
+            else -> null
+        }
+    }
+
     /**
      * Builder for creating [PaywallActivityLaunchIfNeededOptions].
      *
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
@@ -333,14 +333,12 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
     /**
      * Launch the paywall activity with the specified options.
      *
-     * This method provides a builder-based API for launching paywalls with custom configuration
-     * including custom variables.
+     * This method provides a builder-based API for launching paywalls with custom configuration.
      *
      * Example:
      * ```kotlin
      * val options = PaywallActivityLaunchOptions.Builder()
      *     .setOffering(offering)
-     *     .setCustomVariables(mapOf("user_name" to CustomVariableValue.String("John")))
      *     .build()
      *
      * launcher.launchWithOptions(options)
@@ -349,22 +347,9 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
      * @param options The launch options configured via [PaywallActivityLaunchOptions.Builder]
      */
     fun launchWithOptions(options: PaywallActivityLaunchOptions) {
-        // Support both public Offering API and internal offeringIdentifier API
-        val offeringIdAndContext = when {
-            options.offeringIdentifier != null -> OfferingSelection.IdAndPresentedOfferingContext(
-                offeringId = options.offeringIdentifier,
-                presentedOfferingContext = options.presentedOfferingContext,
-            )
-            options.offering != null -> OfferingSelection.IdAndPresentedOfferingContext(
-                offeringId = options.offering.identifier,
-                presentedOfferingContext = options.offering.availablePackages
-                    .firstOrNull()?.presentedOfferingContext,
-            )
-            else -> null
-        }
         activityResultLauncher.launch(
             PaywallActivityArgs(
-                offeringIdAndPresentedOfferingContext = offeringIdAndContext,
+                offeringIdAndPresentedOfferingContext = options.toOfferingSelection(),
                 fontProvider = options.fontProvider,
                 shouldDisplayDismissButton = options.shouldDisplayDismissButton,
                 edgeToEdge = options.edgeToEdge,
@@ -423,23 +408,10 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         shouldDisplayPaywall(shouldDisplayBlock) { shouldDisplay ->
             options.paywallDisplayCallback?.onPaywallDisplayResult(shouldDisplay)
             if (shouldDisplay) {
-                // Support both public Offering API and internal offeringIdentifier API
-                val offeringIdAndContext = when {
-                    options.offeringIdentifier != null -> OfferingSelection.IdAndPresentedOfferingContext(
-                        offeringId = options.offeringIdentifier,
-                        presentedOfferingContext = options.presentedOfferingContext,
-                    )
-                    options.offering != null -> OfferingSelection.IdAndPresentedOfferingContext(
-                        offeringId = options.offering.identifier,
-                        presentedOfferingContext = options.offering.availablePackages
-                            .firstOrNull()?.presentedOfferingContext,
-                    )
-                    else -> null
-                }
                 launchPaywallWithArgs(
                     PaywallActivityArgs(
                         requiredEntitlementIdentifier = options.requiredEntitlementIdentifier,
-                        offeringIdAndPresentedOfferingContext = offeringIdAndContext,
+                        offeringIdAndPresentedOfferingContext = options.toOfferingSelection(),
                         fontProvider = options.fontProvider,
                         shouldDisplayDismissButton = options.shouldDisplayDismissButton,
                         edgeToEdge = options.edgeToEdge,


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -171,6 +176,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         shouldDisplayDismissButton: Boolean = DEFAULT_DISPLAY_DISMISS_BUTTON,
         edgeToEdge: Boolean = defaultEdgeToEdge,
         paywallDisplayCallback: PaywallDisplayCallback? = null,
```
vegaro, line 181: the callback should be the last. See this PR https://github.com/RevenueCat/purchases-android/pull/2530 where Toni added the `edgeToEdge` parameter. But I think we introduced a Java breaking change there by mistake

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -263,6 +270,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
+        customVariables: Map<String, CustomVariableValue> = emptyMap(),
```
vegaro: same here, paywallDisplayCallback should be last

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
vegaro: Not sure how to not cause a breaking change in Java here... we might need a new function

facumenzella: Fixed in c1508fe - moved `customVariables` before `paywallDisplayCallback` so the callback stays last.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Fixed in c1508fe - `paywallDisplayCallback` is now last.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: By keeping `paywallDisplayCallback` as the last parameter (after `customVariables`), we maintain backward compatibility with existing Java code that was already using this method with the callback. The `@JvmOverloads` annotation generates the appropriate overloads.

vegaro: I don't think this is correct. If you look at the Java API https://github.com/RevenueCat/purchases-android/pull/3075/files#diff-a5f257aaa19bec782b952c369c2f463f7bd00d076084aade9833a2cb8afda210L49, there was a breaking change. This is very unfortunate because the only way is creating a new function to keep support for the one that changes.

vegaro: We have to consider moving into a Builder pattern to prevent this sort of thing

facumenzella: You're right, I see the issue now. The 6-argument Java overload changed from expecting `PaywallDisplayCallback` to expecting `Map<String, CustomVariableValue>` as the 6th parameter. This is a breaking change for Java callers using positional arguments.

What's your preferred approach?
1. Create a separate method (e.g., `launchIfNeededWithCustomVariables`)
2. Move toward a Builder pattern for future additions
3. Something else?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -44,6 +48,7 @@ class PaywallView : CompatComposeView {
         fontProvider: FontProvider? = null,
         shouldDisplayDismissButton: Boolean? = null,
         dismissHandler: (() -> Unit)? = null,
+        customVariables: Map<String, CustomVariableValue>? = null,
```
tonidero: This is a breaking change in Java 😢 . Note that since the `dismissHandler` was the last parameter, in Kotlin you could do `PaywallView(....) { // dismiss handler }`, since that's the last parameter. Since you added one after, that basically breaks that API.

I think we would need to either:
- Add another deprecated constructor that matches the current one, and then, add a new constructor with the new `customVariables` BEFORE the `dismissHandler` parameter.
- Create a new builder-like class to pass all these parameters, and deprecate the current constructor based APIs.

I think option 1 might be faster for now, but we should really consider moving to a builder-like pattern.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -143,6 +150,17 @@ class PaywallView : CompatComposeView {
...
     }
 
+    /**
+     * Sets custom variables to be used in paywall text. These values will replace
+     * `{{ custom.key }}` or `{{ $custom.key }}` placeholders in the paywall configuration.
+     *
+     * @param variables A map of variable names to their [CustomVariableValue] values.
+     */
+    fun setCustomVariables(variables: Map<String, CustomVariableValue>) {
+        CustomVariableKeyValidator.validate(variables)
```
tonidero: I see this just logs, but still allows to move forward with invalid variables... Should we filter out invalid ones here instead?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -170,6 +175,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         fontProvider: ParcelizableFontProvider? = null,
         shouldDisplayDismissButton: Boolean = DEFAULT_DISPLAY_DISMISS_BUTTON,
         edgeToEdge: Boolean = defaultEdgeToEdge,
+        customVariables: Map<String, CustomVariableValue> = emptyMap(),
```
tonidero: Unfortunately, we're in the same situation with all of these I think, where this could be a breaking change in Java. Same as before, to support this we would need to either, create separate functions with the new parameter and deprecate the current ones, or create a builder-like method to do this. I think it might be worth moving to the builder-like pattern for these TBH... We might want to think about how the API should look like in that case though, to try to minimize the API surface moving forward as much as possible.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
tonidero: Oh I just saw César's comment above... I think he's right. Also, lol at that AI comment 🥴

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
facumenzella: Fixed in 8e5fae988 - invalid keys are now filtered out instead of just logged. The `validateAndFilter()` method returns only valid entries.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
facumenzella: Fixed in 81481cd47 - removed `customVariables` from the constructor. Users can use `setCustomVariables()` method instead, which preserves trailing lambda syntax for `dismissHandler`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Fixed in 81481cd47 - introduced `PaywallActivityLaunchOptions.Builder` pattern. Removed `customVariables` from methods with callbacks/lambdas at the end. Users can now use:

```kotlin
val options = PaywallActivityLaunchOptions.Builder()
    .setOffering(offering)
    .setCustomVariables(mapOf("key" to CustomVariableValue.String("value")))
    .setRequiredEntitlementIdentifier("premium")
    .build()
launcher.launchIfNeeded(options)
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -262,6 +266,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
```
tonidero: We should remove this parameter, deprecate this method and redirect to the new method instead :) 

We also should update the hybrids to use the new API and the docs for any usages of this API to point to the new one (but that would be in other repos).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -297,6 +301,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
      * Default is true for Android 15+, false otherwise.
      * @param shouldDisplayBlock the paywall will be displayed only if this returns true.
      */
+    @Suppress("LongParameterList")
```
tonidero: I think this isn't needed anymore?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -324,6 +329,113 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
+     *     }
+     *     .setCustomVariables(mapOf("user_name" to CustomVariableValue.String("John")))
+     *     .build()
+     *
+     * launcher.launchIfNeeded(options)
+     * ```
+     *
+     * @param options The launch options configured via [PaywallActivityLaunchOptions.Builder]
+     */
+    fun launchIfNeeded(options: PaywallActivityLaunchOptions) {
```
tonidero: Hmm it's annoying because nothing is forcing the developer to pass in the `requiredEntitlementIdentifier` or the `shouldDisplayBlock` parameters in here... We might want to either have a subclass of `PaywallActivityLaunchOptions` for this method that makes it mandatory, or maybe even unify into a single `launch` method, and just let the `requiredEntitlementIdentifier`/`shouldDisplayBlock` be possible parameters... In this second option, we would be losing  API visibility though, so not sure if we should go with the first one then?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Fixed in 275f8b574 - deprecated this method and removed `customVariables` parameter. The deprecation message points to `launchIfNeeded(PaywallActivityLaunchOptions)` instead.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Fixed in 275f8b574 - removed the `@Suppress("LongParameterList")` annotation.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Ok sounds good. let's do the first one

api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallActivityLauncherAPI.java
```diff
@@ -28,7 +34,7 @@ static void check(
         PaywallActivityLauncher launcher3 = new PaywallActivityLauncher(resultCaller, resultHandler);
         launcher.launch();
         launcher.launch(offering);
-        launcher.launch(null);
+        launcher.launch((Offering) null);
```
vegaro: hmmm if this requires a cast, it's because of a breaking change. We might need to rename the new function to `launchWithOptions` or something like that

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallViewAPI.kt
```diff
@@ -34,14 +34,18 @@ private class PaywallViewAPI {
         PaywallView(context, offering, listener)
         PaywallView(context, offering, listener, fontProvider)
         PaywallView(context, offering, listener, fontProvider, shouldDisplayDismissButton)
-        PaywallView(context, offering, listener, fontProvider, shouldDisplayDismissButton, dismissHandler)
```
vegaro, line 37: I would have both versions (with trailing and passing the paraemter) here to confirm they both work

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
vegaro: this function should build the builder options and call the new function, that way we don't have duplicated code

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
```diff
@@ -36,6 +39,7 @@ class PaywallView : CompatComposeView {
     /**
      * Constructor when creating the view programmatically.
      */
+    @Suppress("LongParameterList")
```
vegaro: you're not adding any parameter so this is probably safe to remove?

vegaro: you're not adding any parameter so this is probably safe to remove?

facumenzella:
All review comments have been addressed:

1. **Java cast issue** - Renamed `launch(options)` to `launchWithOptions(options)` and `launchIfNeeded(options)` to `launchIfNeededWithOptions(options)` to avoid ambiguity (905883889)

2. **Duplicated code in deprecated internal method** - Refactored to use builder pattern with new internal `setOfferingIdentifier()` method (905883889)

3. **PaywallViewAPI test** - Added both named parameter and trailing lambda syntax tests (c22f15c48)

4. **PaywallView @Suppress** - Removed `@Suppress("LongParameterList")` since only 6 params now (c22f15c48)

5. **Type-safe options classes** - Split into `PaywallActivityLaunchOptions` (for `launch`) and `PaywallActivityLaunchIfNeededOptions` (for `launchIfNeeded` with required condition) (0e3ac029c)

6. **Filter invalid custom variable keys** - Changed to `validateAndFilter()` method

7. **Builder pattern** - Implemented for both launch and launchIfNeeded

8. **PaywallView customVariables** - Removed from constructor, using `setCustomVariables()` method instead to preserve trailing lambda syntax

api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallActivityLauncherAPI.java
facumenzella: Fixed in 905883889 - renamed to `launchWithOptions(options)` and `launchIfNeededWithOptions(options)`. The cast is no longer needed.

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallViewAPI.kt
facumenzella: Fixed in c22f15c48 - added both named parameter (`dismissHandler = {}`) and trailing lambda syntax tests.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Fixed in 905883889 - refactored to use builder pattern. Added internal `setOfferingIdentifier()` method to support hybrid SDK use case, then the deprecated method builds the options and calls `launchIfNeededWithOptions()`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
facumenzella: Fixed in c22f15c48 - removed `@Suppress("LongParameterList")` since we now have only 6 parameters (under the threshold).

api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallActivityLauncherAPI.java
```diff
@@ -51,6 +57,44 @@ static void check(
         launcher.launchIfNeeded(offering, null, true, customerInfo -> null);
         launcher.launchIfNeeded(null, fontProvider, true, customerInfo -> null);
         launcher.launchIfNeeded(null, null, true, customerInfo -> null);
+        launcher.launchIfNeeded(null, null, true, true, customerInfo -> null);
```
vegaro, line 60: which one is this?

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
```diff
@@ -117,6 +120,53 @@ private class PaywallActivityLauncherAPI {
             val customerInfo: CustomerInfo = it
             true
         }
+        activityLauncher.launchIfNeeded(
+            offering = offering,
+            fontProvider = fontProvider,
+            shouldDisplayDismissButton = true,
+            edgeToEdge = true,
```
vegaro, line 127: oh we were missing this? nice

api-tester/src/defaults/java/com/revenuecat/apitester/java/revenuecatui/PaywallActivityLauncherAPI.java
vegaro: oh I see it's the edge to edge one we were missing

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
```diff
@@ -117,6 +120,53 @@ private class PaywallActivityLauncherAPI {
...
+            .build()
+        activityLauncher.launchWithOptions(options)
+
+        // LaunchIfNeeded with requiredEntitlementIdentifier
+        val optionsWithEntitlement = PaywallActivityLaunchIfNeededOptions.Builder()
+            .setRequiredEntitlementIdentifier("premium")
+            .setOffering(offering)
+            .setCustomVariables(customVariables)
+            .setPaywallDisplayCallback(paywallDisplayCallback)
+            .build()
```
vegaro, line 161: I am surprised there's no `setFontProvider`, `setShouldDisplayDismissButton`, and `setEdgeToEdge` here

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
```diff
@@ -117,6 +120,53 @@ private class PaywallActivityLauncherAPI {
...
+        activityLauncher.launchIfNeededWithOptions(optionsWithEntitlement)
+
+        // LaunchIfNeeded with shouldDisplayBlock
+        val optionsWithBlock = PaywallActivityLaunchIfNeededOptions.Builder()
```
vegaro, line 165: Can we join it with `optionsWithEntitlement`? There's no difference right?They are both PaywallActivityLaunchIfNeededOptions.Builder

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLaunchOptions.kt
```diff
@@ -0,0 +1,292 @@
...
+                customVariables = customVariables,
+                requiredEntitlementIdentifier = requiredEntitlementIdentifier,
+                shouldDisplayBlock = shouldDisplayBlock,
+                paywallDisplayCallback = paywallDisplayCallback,
+                offeringIdentifier = offeringIdentifier,
+                presentedOfferingContext = presentedOfferingContext,
+            )
+        }
+    }
+}
```
vegaro, line 349: This will crash at runtime right? Do we want that? Maybe we could call `PaywallActivityLaunchOptions` ourselves here?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -135,6 +137,7 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         fontProvider: ParcelizableFontProvider? = null,
         shouldDisplayDismissButton: Boolean = DEFAULT_DISPLAY_DISMISS_BUTTON,
         edgeToEdge: Boolean = defaultEdgeToEdge,
+        customVariables: Map<String, CustomVariableValue> = emptyMap(),
```
vegaro, line 147: I think we shoul deprecate this functoin in favor of the one with the builder as well. For parity with `launchIfNeeded` and for future proof

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -324,6 +330,126 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
         }
     }
 
+    /**
+     * Launch the paywall activity with the specified options.
+     *
+     * This method provides a builder-based API for launching paywalls with custom configuration
+     * including custom variables.
```
tonidero: I don't think we need to explicitedly mention custom variables in the docs here. That should be an option as any other :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -324,6 +330,126 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
+                offeringId = options.offeringIdentifier,
+                presentedOfferingContext = options.presentedOfferingContext,
+            )
+            options.offering != null -> OfferingSelection.IdAndPresentedOfferingContext(
+                offeringId = options.offering.identifier,
+                presentedOfferingContext = options.offering.availablePackages
+                    .firstOrNull()?.presentedOfferingContext,
+            )
+            else -> null
+        }
```
vegaro: we might be able to extract this into an internal function of the `Options` maybe

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -324,6 +330,126 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
+                    options.offering != null -> OfferingSelection.IdAndPresentedOfferingContext(
+                        offeringId = options.offering.identifier,
+                        presentedOfferingContext = options.offering.availablePackages
+                            .firstOrNull()?.presentedOfferingContext,
+                    )
+                    else -> null
+                }
+                launchPaywallWithArgs(
+                    PaywallActivityArgs(
+                        requiredEntitlementIdentifier = options.requiredEntitlementIdentifier,
```
vegaro, line 413: This is making me wonder if we should just have a `PaywallActivityLaunchOptions` that has a `requiredEntitlementIdentifier` and `shouldDisplayBlock` and we check for those properties instead of having a `PaywallActivityLaunchIfNeededOptions` and `PaywallActivityLaunchOptions`. That way we can just have a single `launchWithOptions`. I think it might be cleaner. Sorry for not thinking of this earlier. What do you think @tonidero ?

tonidero: Right... I thought about [this](https://github.com/RevenueCat/purchases-android/pull/3075#discussion_r2783361697), and ended up thinking that keeping it separate might be better for API discoverability... But I also agree it would help keeping the API surface cleaner so I'm not too opposed... 

Another approach could be to have 2 "static" builder constructors with the different parameters for each.. But I'm not too in love with that option, I would prefer keeping it simple and just have it as "normal" setters (the option you mentioned).

So yeah... I'm happy if we want to unify it, and then mention this in the docs 👍 Sorry for the back and forth @facumenzella 🙏

vegaro: hmmm. you know what... API discoverability is a good point. It also helps migrating. Maybe we can unify in the next major or something instead

facumenzella: thank you both!!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: 👍  done

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/views/PaywallView.kt
facumenzella: Same fix as the other thread - removed in c22f15c48

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
facumenzella: Added in d26dc100d - both test cases now have all builder methods (`setFontProvider`, `setShouldDisplayDismissButton`, `setEdgeToEdge`, `setPaywallDisplayCallback`)

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallActivityLauncherAPI.kt
facumenzella: Kept them separate since they test different conditions (`requiredEntitlementIdentifier` vs `shouldDisplayBlock`), but both now have all builder methods in d26dc100d

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLaunchOptions.kt
facumenzella: The runtime crash is intentional - it provides a clear error message guiding developers to either:
1. Set a condition (`requiredEntitlementIdentifier` or `shouldDisplayBlock`) for conditional launches
2. Use `PaywallActivityLaunchOptions` with `launchWithOptions()` for unconditional launches

Silently converting would change the API semantics from conditional to unconditional.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
facumenzella: Done in 7d4e22557 - deprecated the internal `launch(offeringIdentifier, presentedOfferingContext, ...)` in favor of `launchWithOptions()`. Also added internal `setOfferingIdentifier()` to `PaywallActivityLaunchOptions.Builder`.

PR #3078, opened 2026-02-06
Send `PresentedOfferingContext` info when `PurchasesAreCompletedBy.MY_APP` and using paywalls

tonidero:
### Description
Right now, for devs using `PurchasesAreCompletedBy.MY_APP` and paywalls, the SDK will send the paywall information to our backend, but won't send the presented offering context information along. This PR will fix that, so we send this information as well in this case

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/PostReceiptHelperTest.kt
@@ -2159,6 +2159,151 @@ class PostReceiptHelperTest {
     }
 
     @Test
+    fun `postTransactionAndConsumeIfNeeded uses paywall presentedOfferingContext when receiptInfo context is null`() {
+        // Create a transaction without presentedOfferingContext
+        val transactionWithoutContext = mockGooglePurchase.toStoreTransaction(
+            ProductType.SUBS,
+            null, // No presentedOfferingContext
+            subscriptionOptionId,
+            replacementMode = GoogleReplacementMode.CHARGE_FULL_PRICE,
+        )
+
+        // Create a paywall event with presentedOfferingContext
+        val paywallOfferingContext = PresentedOfferingContext("paywall_offering_id")
+        val paywallEvent = PaywallEvent(
+            creationData = PaywallEvent.CreationData(UUID.randomUUID(), 1.hours.ago()),
+            data = PaywallEvent.Data(
+                paywallIdentifier = "paywall_id",
+                paywallOfferingContext,
+                10,
+                UUID.randomUUID(),
+                "footer",
+                "es_ES",
+                false,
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEvent.kt
```diff
@@ -100,7 +116,166 @@ data class PaywallEvent(
...
+        )
+    }
+}
+
+/**
+ * Custom serializer for PaywallEvent.Data that supports parsing both the old
+ * `offeringIdentifier` and the new `presentedOfferingContext`.
+ */
+@OptIn(InternalRevenueCatAPI::class)
+internal object PaywallEventDataSerializer : KSerializer<PaywallEvent.Data> {
```
tonidero, line 129: Need a custom serializer here because legacy paywall events are actually serialized (for components paywalls, we don't really need to serialize this, since backendEvent is serialized instead) otherwise, we could just make `PaywallEvent` not serializable 😢.

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/PaywallEvent.kt
```diff
@@ -74,10 +90,10 @@ data class PaywallEvent(
         val date: Date,
     )
 
-    @Serializable
+    @Serializable(with = PaywallEventDataSerializer::class)
     data class Data(
         val paywallIdentifier: String?,
-        val offeringIdentifier: String,
+        val presentedOfferingContext: PresentedOfferingContext,
```
tonidero, line 96: Basically now the event will hold the entire information of the presented offering context. This will only be used for paywalls if we don't already have presentedOfferingContext information from the current system though. This way we make sure we cover the case of PurchasesAreCompletedBy.MY_APP + Paywalls.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -768,4 +769,7 @@ internal class PaywallViewModelImpl(
         offering.paywallComponents?.uiConfig?.customVariables
             ?.mapValues { (_, definition) -> CustomVariableValue.from(definition.defaultValue) }
             ?: emptyMap()
+
+    private val Offering.presentedOfferingContext: PresentedOfferingContext
+        get() = availablePackages.firstOrNull()?.presentedOfferingContext ?: PresentedOfferingContext(identifier)
```
tonidero, line 774: There are previous validations so offerings should have at least one package, so it should never fallback, but this seems like an ok compromise for now since we don't have the `PresentedOfferingContext` as part of the `Offering` right now 🥲

PR #3081, opened 2026-02-10
[EXPERIMENTAL] Support custom paywall logic and event listening when using PaywallActivity

tonidero:
### Description
Currently, if devs need to set a custom `PurchaseLogic` in order to support paywalls in `PurchasesAreCompletedBy.MY_APP`, they are forced to use the `Paywall` composable directly, but won't work for the `PaywallView` nor `PaywallActivityLauncher` APIs. Similar story for setting the `PaywallListener` to listen to events when using these APIs.

This PR:
- Adds a parameter to the `PurchasesConfiguration` when configuring the SDK, `CustomPaywallHandlerFactory`. This API allows developers to create custom handlers including both `PaywallLogic` and `PaywallListener` that will work through any supported API. These new APIs are currently experimental
- In order to support this, and to reuse the existing types, I was forced to move `PurchaseLogic`, `PaywallListener` and `Resumable` from the `ui` module to the `purchases` module, so it can be set as part of the configuration.

Some notes:
- The new APIs are currently experimental
- If using the both the existing APIs for the `Paywall` composable , and the new system, the `Paywall` composable APIs will have preference over the new APIs.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
@@ -83,7 +83,7 @@ open class PurchasesConfiguration(builder: Builder) {
                 automaticDeviceIdentifierCollectionEnabled,
             )
             .preferredUILocaleOverride(preferredUILocaleOverride)
-            .setCustomPaywallHandlerFactoryInternal(customPaywallHandlerFactory)
+            .setCustomPaywallHandlerFactory(customPaywallHandlerFactory)
         if (service != null) {
             builder = builder.service(service)
         }
@@ -326,21 +326,6 @@ open class PurchasesConfiguration(builder: Builder) {
         ) = apply {
             this.customPaywallHandlerFactory = factory
         }
-
-        /**
-         * Internal method for setting the custom paywall handler factory.
-         * Use [setCustomPaywallHandlerFactory] instead.
-         *
-         * @param factory The factory instance
-         */
-        @Deprecated(
-            "Use setCustomPaywallHandlerFactory instead",
-            ReplaceWith("setCustomPaywallHandlerFactory(factory)"),
-        )
-        @InternalRevenueCatAPI
-        fun setCustomPaywallHandlerFactoryInternal(
-            factory: CustomPaywallHandlerFactory?,
-        ) = setCustomPaywallHandlerFactory(factory)
 
         /**
          * Creates a [PurchasesConfiguration] instance with the specified properties.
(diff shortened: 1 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomPaywallHandlerParams.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Parameters passed to [CustomPaywallHandlerFactory] when creating a custom paywall handler.
+ *
+ * This class encapsulates all context needed to create an appropriate handler for a paywall.
+ * Currently contains the offering being displayed, but may be extended with additional
+ * parameters in the future.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@Poko
+class CustomPaywallHandlerParams @InternalRevenueCatAPI constructor(
```
tonidero, line 17: I tried to make it easier to add other parameters later

purchases/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomPaywallHandler.kt
```diff
@@ -0,0 +1,62 @@
+package com.revenuecat.purchases.ui.revenuecatui
```
tonidero, line 1: Do note that I added this package in the `purchases` module. It's leaking types into the `purcahses` module, but it felt the cleanest option to me (see Linear for more details on the discussion)

purchases/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomPaywallHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+     * The [PurchaseLogic] to use for this paywall.
+     * Return null if you don't want to provide custom purchase logic for this paywall.
+     *
+     * Note: [PurchaseLogic] is only used when [com.revenuecat.purchases.Purchases.purchasesAreCompletedBy] is set to
+     * [com.revenuecat.purchases.PurchasesAreCompletedBy.MY_APP]. If you provide this when using
+     * [com.revenuecat.purchases.PurchasesAreCompletedBy.REVENUECAT], it will be ignored.
+     */
+    val purchaseLogic: PurchaseLogic?
+        get() = null
+}
```
JayShortway, line 62: Do we need a new interface grouping the 2 existing interfaces? Why not keep them separate?

tonidero: I tried to make this a more generic class that allows to set both a PaywallListener and a PurchaseLogic. We could indeed split it into 2 more focused factories, but this seemed easier to extend, so you just need to implement one factory to handle everything... Not a strong opinion though, I would be ok if we think having 2 factories could be simpler...

PR #3082, opened 2026-02-10
Add CLAUDE.md and AGENTS.md

vegaro:
Reverts RevenueCat/purchases-android#2648


AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+
+# Run unit tests
+./gradlew test
+
+# Run unit tests with backend integration tests
+./gradlew test -PRUN_INTEGRATION_TESTS=true
+
+# Run unit tests for specific modules
+./gradlew :purchases:testDefaultsDebugUnitTest
+./gradlew :ui:revenuecatui:testDefaultsDebugUnitTest
```
tonidero: These need to be updated to account the newer `Bc7/Bc8` flavor dimension. Possibly some others below as well.

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+./gradlew detektAll --auto-correct
+
+# Create detekt baseline
+./gradlew detektAllBaseline
+
+# API compatibility check
+./scripts/api-check.sh
+
+# Generate API signatures
+./scripts/api-dump.sh
```
tonidero: Could be worth mentioning here the tool we use? 
```suggestion
# API compatibility check using Metalava
./scripts/api-check.sh

# Generate API signatures using Metalava
./scripts/api-dump.sh
```

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+bundle exec fastlane run_backend_integration_tests
+```
+
+## Project Architecture
+
+### Module Structure
+This is a multi-module Android project with clear separation of concerns:
+
+- **`:purchases`** - Core SDK module containing main API, business logic, networking, billing abstractions
+- **`:ui:revenuecatui`** - Jetpack Compose UI module for paywalls and customer center (min SDK 24)
```
tonidero: Could be worth mentioning that these modules have dependencies with the `purchases` module? I guess it can probably figure it out on its own, but might make it simpler 🤷

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+├── interfaces/             # Public API contracts
+└── utils/                  # Utility functions
+```
+
+### Product Flavors
+The `purchases` module has 2 flavors:
+- **`defaults`** - Standard Google Play implementation
+- **`customEntitlementComputation`** - Custom entitlement computation variant
+
+## Testing Framework
```
tonidero, line 115: [Unrelated to this line] Could be worth adding a section about the annotations we use and their usages:
- `@InternalRevenueCatAPI`, meant for APIs that are public to be accessible by other modules or hybrid SDKs that use this, but not for the public
- `@ExperimentalPreviewRevenueCatAPI`, meant for APIs that are meant to be public and used by developers but may be subject to change before being made stable
- `@ExperimentalPreviewRevenueCatUIAPI`, same but for the `ui:revenuecatui` module.

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+- **Orchestrator Pattern**: `PurchasesOrchestrator` as central coordinator
+- **Abstract Factory**: `BillingAbstract` for different store implementations
+- **Repository Pattern**: `Backend` for networking, `DeviceCache` for local storage
+- **Manager Pattern**: `IdentityManager`, `SubscriberAttributesManager`, `EventsManager`
+
+#### UI Modules
+- **MVVM Pattern**: ViewModels with Jetpack Compose UI
+- **Main Components**: `PaywallViewModel`, `CustomerCenterViewModel`
+
+### Package Organization
```
tonidero: I wonder if we should remove this... I feel it's very easy for this to become outdated... Alternatively, maybe we could have a job that checks this and updates it periodically... not sure how that would work 😅

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+- **`:baselineprofile`** - Performance optimization profiles
+- **`:integration-tests`** - Integration test suite
+- **`:examples/*`** - Sample applications and testers
+
+### Key Architectural Patterns
+
+#### Core Purchases Module
+- **Orchestrator Pattern**: `PurchasesOrchestrator` as central coordinator
+- **Abstract Factory**: `BillingAbstract` for different store implementations
+- **Repository Pattern**: `Backend` for networking, `DeviceCache` for local storage
```
tonidero: I don't know if I would call these a "Repository" pattern 😅, but I guess I don't have a better name :P

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+├── subscriberattributes/    # Subscriber attributes system
+├── paywalls/               # Paywall data and components
+├── customercenter/         # Customer center functionality
+├── models/                 # Data models and DTOs
+├── interfaces/             # Public API contracts
+└── utils/                  # Utility functions
+```
+
+### Product Flavors
+The `purchases` module has 2 flavors:
```
tonidero: Mentioned it above, but this is outdated since we added a new flavor dimension

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Development Workflow
+
+### Environment Setup
+1. Install sdkman and run `sdk env install` in project root
+2. Run `bundle exec fastlane setup_dev` to link pre-commit hooks
+3. Pre-commit hooks automatically run detekt on commits
+
+### Code Quality
+- **Detekt**: Static code analysis with auto-correction
```
tonidero: Could be worth also mention `./gradlew lint` in this list?

AGENTS.md
```diff
@@ -0,0 +1,192 @@
...
+### Backend Integration Testing
+- Enable with `-PRUN_INTEGRATION_TESTS=true`
+- Uses real API keys and tokens from environment variables
+- Located in `backend_integration_tests` packages
+
+### Sample Applications
+- **MagicWeather**: Standard sample app
+- **MagicWeatherCompose**: Compose-based sample
+- **purchase-tester**: Testing app for purchase flows
+- **paywall-tester**: Testing app for paywall UI
```
tonidero, line 184: This list is not complete... but I guess it's NABD.

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+- **Lint**: Android lint for static code analysis (`./gradlew lint`)
+- **Detekt**: Static code analysis with auto-correction (`./gradlew detektAll`)
+- **Pre-commit Hook**: Runs detekt before each commit
+- **API Compatibility**: Metalava for API checking and signature generation
+- **Baseline Profiles**: Performance optimization for improved startup
+
+### Main Entry Points
+- **`Purchases`** class: Primary SDK entry point
+- **`PurchasesConfiguration`**: Configuration builder
+- **UI Components**: `Paywall()`, `CustomerCenter()` composables
```
tonidero: Maybe worth adding `PaywallActivityLauncher` and `ShowCustomerCenter`?

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+- **Kotlinx Serialization**: JSON handling
+- **Google Tink**: Cryptography
+- **OkHttp**: HTTP networking foundation
+
+## Build Configuration
+
+### Gradle Structure
+- **Multi-module**: Uses `include()` statements in `settings.gradle.kts`
+- **Version Catalogs**: Centralized dependency management in `gradle/libs.versions.toml`
+- **Flavors**: Support for different store implementations
```
tonidero: I guess this could be a bit confusing, since we don't use flavors for store implementations, but more for API and billing client versions right now... (we probably can remove the billing client dimension soon though)

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+- **Build Types**: Debug and release configurations
+
+### Target Specifications
+- **Compile SDK**: 34
+- **Min SDK**: 21 (24 for UI modules)
+- **Java**: 8+
+- **Kotlin**: 1.8.0+
```
tonidero: Looks like Claude had some comments that I agree with in their review here :) Mostly need to update the Compile SDK version, and make clear what the kotlin version means.

PR #3083, opened 2026-02-10
RCT Tester app: Add initial app

rickvdl:
### Motivation
This is the first of multiple PRs related to implementing the RCT Tester app. This internal test tool is used for verifying that the right attribution data is sent in the `POST /receipt` endpoint, across configurations and scenarios.

### Description
Adds the initial barebones app with support for configuring the SDK with:
- API Key
- Purchases completed by
- User ID

This PR also includes the offering screen with the option of buying the specific packages as well as showing the configured paywall for the given offering.

| Setup screen | Main screen | Offerings screen |
|--------------|-------------|------------------|
| Screenshot_20260210_142757 | Screenshot_20260210_142724 | Screenshot_20260210_142729 |

### DIFF
--- a/examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/MainScreen.kt
+++ b/examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/MainScreen.kt
@@ -3,9 +3,9 @@ package com.revenuecat.rcttester.ui.main
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
--- a/examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
+++ b/examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
@@ -539,7 +539,6 @@ private fun OfferingCardPackages(
             PackageRow(
                 packageItem = packageItem,
                 isPurchasing = purchasingPackageId == packageItem.identifier,
-                isPurchaseEnabled = purchasingPackageId == null,
                 onPurchase = { onPurchasePackage(packageItem) },
             )
         }
@@ -556,7 +555,6 @@ private fun OfferingCardPackages(
 private fun PackageRow(
     packageItem: Package,
     isPurchasing: Boolean,
-    isPurchaseEnabled: Boolean,
     onPurchase: () -> Unit,
 ) {
     Row(
@@ -579,7 +577,7 @@ private fun PackageRow(
         }
         Button(
             onClick = onPurchase,
-            enabled = isPurchaseEnabled,
+            enabled = !isPurchasing,
         ) {
             if (isPurchasing) {
                 CircularProgressIndicator(


examples/rcttester/src/androidTest/java/com/example/rcttester/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
```
tonidero: Can potentially remove these default tests until we need them :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/config/SDKConfiguration.kt
```diff
@@ -0,0 +1,80 @@
...
+        /**
+         * Loads the configuration from SharedPreferences, or returns null if none exists
+         */
+        fun load(context: Context): SDKConfiguration? {
+            val prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE)
+            val jsonString = prefs.getString(KEY_CONFIG, null) ?: return null
+            return try {
+                json.decodeFromString<SDKConfiguration>(jsonString)
+            } catch (e: IllegalArgumentException) {
+                android.util.Log.w("SDKConfiguration", "Failed to decode configuration", e)
```
tonidero: We should import instead of using the fully qualified name.

examples/rcttester/src/main/java/com/revenuecat/rcttester/config/SDKConfiguration.kt
```diff
@@ -0,0 +1,80 @@
...
+
+@Serializable
+enum class PurchasesCompletedByType {
+    REVENUECAT,
+    MY_APP, ;
+
+    val displayName: String
+        get() = when (this) {
+            REVENUECAT -> ".revenueCat"
+            MY_APP -> ".myApp"
```
tonidero: Hmm not a strong opinion, but wondering if we should use the Android names for this and it would be simpler... As in, use `REVENUECAT` and `MY_APP`.

tonidero: That way I think we can just use `PurchasesAreCompletedBy` directly and not create `PurchasesCompletedByType`? Though it's true that's not serializable, so we would need to write our own serializer... so I'm ok keeping this enum :)

examples/rcttester/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.revenuecat.rcttester"
+    android:versionCode="2"
+    android:versionName="1.0">
+
+    <uses-permission android:name="com.google.android.gms.permission.AD_ID"/>
```
tonidero, line 7: Hmm we probably don't need the ad permission... But NABD to leave it though.

examples/rcttester/build.gradle.kts
```diff
@@ -0,0 +1,96 @@
...
+    implementation(libs.androidx.datastore.preferences)
+
+    // Coroutines (included transitively via lifecycle-runtime-ktx, but explicit for clarity)
+    implementation(libs.coroutines.core)
+
+    // Serialization for configuration persistence
+    implementation(libs.kotlinx.serialization.json)
+
+    // Testing
+    testImplementation(libs.junit)
```
tonidero: Just a note that if we do end up removing the default tests and androidTests, we could remove these dependencies as well

gradle/libs.versions.toml
```diff
@@ -168,6 +169,8 @@ androidx-benchmark-macro-junit4 = { group = "androidx.benchmark", name = "benchm
 androidx-profileinstaller = { group = "androidx.profileinstaller", name = "profileinstaller", version.ref = "profileinstaller" }
 leakcanary-android = { module = "com.squareup.leakcanary:leakcanary-android", version.ref = "leakcanary" }
 leakcanary-android-instrumentation = { module = "com.squareup.leakcanary:leakcanary-android-instrumentation", version.ref = "leakcanary" }
+junit = { group = "junit", name = "junit", version.ref = "junitVersion" }
```
tonidero: It could be good to use a different junit that we might already be using in other modules instead of adding a dependency with base junit I think (this is only if we want to keep unit tests, but might be easier to just remove those and also remove the test dependencies)

examples/rcttester/src/main/java/com/revenuecat/rcttester/MainApplication.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    fun configureSDK(configuration: SDKConfiguration) {
+        // Sanitize API key before saving
+        val sanitizedApiKey = configuration.apiKey.trim().replace("\n", "").replace("\r", "")
+        val sanitizedConfig = configuration.copy(apiKey = sanitizedApiKey)
+        sanitizedConfig.save(this)
+
+        // Don't reconfigure if already configured with the same config
+        if (Purchases.isConfigured) {
```
tonidero: I guess it should never get to this point right? Since we do the check for this before getting to this point.

examples/rcttester/src/main/java/com/revenuecat/rcttester/MainApplication.kt
```diff
@@ -0,0 +1,98 @@
...
+        val purchasesAreCompletedBy = sanitizedConfig.toPurchasesAreCompletedBy()
+        val builder = PurchasesConfiguration.Builder(this, sanitizedApiKey)
+            .purchasesAreCompletedBy(purchasesAreCompletedBy)
+
+        // Set app user ID if provided, otherwise SDK will generate anonymous ID
+        if (sanitizedConfig.appUserID.isNotBlank()) {
+            builder.appUserID(sanitizedConfig.appUserID.trim())
+        } else {
+            builder.appUserID(null)
+        }
```
tonidero: Super nitpick, I think you don't need the `else`, since it should do the same thing. NABD though :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/MainActivity.kt
```diff
@@ -0,0 +1,93 @@
...
+        )
+    }
+    var configuration by remember {
+        mutableStateOf(
+            SDKConfiguration.load(application) ?: SDKConfiguration.default(),
+        )
+    }
+
+    // Navigate to main screen if SDK becomes configured
+    LaunchedEffect(application.isSDKConfigured) {
```
tonidero: This won't trigger automatically since `isSDKConfigured` is not a state variable I think. We would need to make it a `mutableStateOf` there, and update it as needed if we want this to listen to changes there?

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/MainScreen.kt
```diff
@@ -0,0 +1,151 @@
...
+}
+
+@Composable
+private fun MainScreenContent(
+    configuration: SDKConfiguration,
+    onConfigurationUpdate: (SDKConfiguration) -> Unit,
+    onNavigateToOfferings: () -> Unit,
+    paddingValues: androidx.compose.foundation.layout.PaddingValues,
+) {
+    LazyColumn(
```
tonidero: Hmm I think there is no need to use a `LazyColumn` here, considering the items here are not dynamic nor too many? A normal `Column` and making it scrollable might suffice.

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/MainScreen.kt
```diff
@@ -0,0 +1,151 @@
...
+        )
+    }
+}
+
+@Composable
+private fun MainScreenContent(
+    configuration: SDKConfiguration,
+    onConfigurationUpdate: (SDKConfiguration) -> Unit,
+    onNavigateToOfferings: () -> Unit,
+    paddingValues: androidx.compose.foundation.layout.PaddingValues,
```
tonidero: Should import :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
```diff
@@ -0,0 +1,757 @@
...
+            purchaseResult = purchaseResult,
+        ),
+        callbacks = OfferingsScreenCallbacks(
+            onNavigateBack = onNavigateBack,
+            onRetry = { refreshKey++ },
+            onShowMetadata = { selectedOfferingForMetadata = it },
+            onPresentPaywall = { selectedOfferingForPaywall = it },
+            onDismissMetadata = { selectedOfferingForMetadata = null },
+            onDismissPaywall = { selectedOfferingForPaywall = null },
+            onDismissPurchaseResult = { purchaseResult = null },
```
tonidero, line 119: I wonder... would it be worth navigating back to the main screen when a purchase is successful?

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
```diff
@@ -0,0 +1,757 @@
...
+    var context = this
+    while (context is ContextWrapper) {
+        if (context is Activity) return context
+        context = context.baseContext
+    }
+    error("No activity found")
+}
+
+private sealed class PurchaseResult {
+    data class Success(val orderId: String, val customerInfo: com.revenuecat.purchases.CustomerInfo) : PurchaseResult()
```
tonidero: For a moment, I got confused with the SDK's `PurchaseResult` 😅, but I think it's probably ok, considering this is very scoped to this file 👍 

Also, we should import :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/UserSummarySection.kt
```diff
@@ -0,0 +1,592 @@
...
+            coroutineScope = coroutineScope,
+            onConfigurationUpdate = onConfigurationUpdate,
+        ),
+    )
+}
+
+@Composable
+private fun rememberUserSummaryState(): MutableUserSummaryState {
+    return remember {
+        MutableUserSummaryState(
```
tonidero: As claude said in its comment, I don't think this is a proper "state" that would get automatic observed updates... We might want to change that, and then maybe rethink the name of this class since it could be confusing?

examples/rcttester/src/main/AndroidManifest.xml
rickvdl: This was necessary to get the first build submitted (used the same permission declaration as the other test apps) and then the play console complained that the permission was not configured by the app, so I just added it. We could probably remove it but we might need it when submitting a build again.

tonidero: Ahhh ok, it's fine to leave it then :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/MainApplication.kt
rickvdl: I think this should no longer be necessary now indeed :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/MainActivity.kt
rickvdl: Good point, updated :)

gradle/libs.versions.toml
rickvdl: Removed :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/MainScreen.kt
rickvdl: Good catch, updated :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
rickvdl: Good idea, added 👍

examples/rcttester/src/main/java/com/revenuecat/rcttester/config/SDKConfiguration.kt
rickvdl: Good point, I've updated the display name but did not add (de)serialization to the SDK type to keep it simple :) LMK if you disagree

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/UserSummarySection.kt
rickvdl: Thanks for flagging this! This should now properly work :)

gradle/libs.versions.toml
tonidero: We can remove this since it's unused now I think?

examples/rcttester/build.gradle.kts
```diff
@@ -0,0 +1,87 @@
+plugins {
+    alias(libs.plugins.android.application)
+    alias(libs.plugins.kotlin.android)
+    alias(libs.plugins.compose.compiler)
+    alias(libs.plugins.kotlin.serialization)
+}
+
+android {
+    namespace = "com.revenuecat.rcttester"
+    compileSdk {
```
tonidero: I will say I haven't seen this syntax, instead it used to be just `compileSdk = 36`... But I'm ok if it works :)

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/UserSummarySection.kt
```diff
@@ -0,0 +1,532 @@
...
+    val coroutineScope = rememberCoroutineScope()
+
+    var userSummaryState by remember {
+        mutableStateOf(
+            UserSummaryState(currentAppUserID = "Not configured"),
+        )
+    }
+    var dialogsData by remember { mutableStateOf(DialogsData()) }
+
+    LaunchedEffect(Purchases.isConfigured) {
```
tonidero: This won't do anything either and right now it could be removed and just leave the `if (Purchases.isConfigured)` inside. The alternative is to expose a mutable state with the value of this, but not sure if it's needed...

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
```diff
@@ -0,0 +1,763 @@
...
+        Text(
+            text = "Packages:",
+            style = MaterialTheme.typography.labelMedium,
+            fontWeight = FontWeight.Bold,
+        )
+        packages.forEach { packageItem ->
+            PackageRow(
+                packageItem = packageItem,
+                isPurchasing = purchasingPackageId == packageItem.identifier,
+                isPurchaseEnabled = purchasingPackageId == null,
```
tonidero: I'm thinking couldn't we just unify `isPurchasing` and `isPurchaseEnabled`? Seems like it's basically the same data (or the negative in this case 😅

tonidero (approved):
A few smallish comments, but I think it's looking great! 🚢

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/offerings/OfferingsScreen.kt
rickvdl: Well sort of I think. The original idea was to disable all (other) buttons when purchasing a package. Just having the isPurchasing would only disable that specific button right?

examples/rcttester/src/main/java/com/revenuecat/rcttester/ui/main/UserSummarySection.kt
rickvdl: Good catch, updated!

ajpallares (approved):
Great work!

PR #3084, opened 2026-02-10
[EXPERIMENTAL] Support `PaywallListener` and `PurchaseLogic` when using PaywallActivity

tonidero:
### Description
This is based on #3075 (because of the launcher builder changes.
This is a different approach than #3081 that makes it simpler

This adds a couple new setter functions to the `PaywallActivityLaunchOptions` to be able to pass in a `PaywallListener` and/or a `PurchaseLogic` types. Previously, this functionality was limited to the `Purchase` composable, and this expands these functionalities to this API. Currently behind an experimental flag.


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityNonSerializableArgsStore.kt
```diff
@@ -0,0 +1,26 @@
+package com.revenuecat.purchases.ui.revenuecatui.activity
+
+import com.revenuecat.purchases.ui.revenuecatui.PaywallListener
+import com.revenuecat.purchases.ui.revenuecatui.PurchaseLogic
+
+internal data class PaywallActivityNonSerializableArgs(
+    val purchaseLogic: PurchaseLogic? = null,
+    val listener: PaywallListener? = null,
+)
```
JayShortway, line 9: Maybe we should keep these by `System.identityHashCode()`, so we can support multiple paywalls visible at the same time with (potentially) different listeners/logic? (We should then also keep track of a count probably.) 

I'm thinking of large displays, multiple windows, those kinds of setups.
https://developer.android.com/develop/ui/views/layout/support-multi-window-mode#lifecycle

tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3084/commits/f0e939cd7926cf51d1e4b63d882a5f2e3284e6d2

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLauncher.kt
```diff
@@ -416,12 +437,27 @@ class PaywallActivityLauncher(resultCaller: ActivityResultCaller, resultHandler:
...
+        purchaseLogic: PurchaseLogic?,
+        listener: PaywallListener?,
+    ): Int? {
+        if (purchaseLogic == null && listener == null) return null
+        val args = PaywallActivityNonSerializableArgs(
+            purchaseLogic = purchaseLogic,
+            listener = listener,
+        )
+        val key = PaywallActivityNonSerializableArgsStore.store(args)
+        currentNonSerializableArgsKey = key
```
JayShortway, line 457: I like this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -112,6 +115,74 @@ internal class PaywallActivity : ComponentActivity(), PaywallListener {
             return
         }
 
+        val nonSerializableArgs = args?.nonSerializableArgsKey?.let {
+            PaywallActivityNonSerializableArgsStore.get(it)
+        }
+        if (args?.nonSerializableArgsKey != null && nonSerializableArgs == null) {
+            Logger.w(
+                "PaywallActivity was recreated after process death but non-serializable args " +
+                    "(PurchaseLogic/PaywallListener) were lost. Finishing activity.",
```
JayShortway:
```suggestion
                "PaywallActivity was recreated after process death causing " +
                    "PurchaseLogic and/or PaywallListener to be lost. Finishing activity.",
```

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLaunchOptions.kt
```diff
@@ -128,6 +136,24 @@ class PaywallActivityLaunchOptions private constructor(
...
+        }
+
+        /**
+         * Sets a listener to receive paywall events such as purchase completion,
+         * restoration, and errors.
+         */
+        @ExperimentalPreviewRevenueCatUIPurchasesAPI
+        fun setListener(listener: PaywallListener?) = apply {
+            this.listener = listener
+        }
```
JayShortway, line 155: Awesome to finally have this in the PaywallActivity!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivityLaunchOptions.kt
```diff
@@ -128,6 +136,24 @@ class PaywallActivityLaunchOptions private constructor(
             this.customVariables = CustomVariableKeyValidator.validateAndFilter(customVariables)
         }
 
+        /**
+         * Sets custom purchase logic to handle purchases directly by the application
+         * rather than by RevenueCat.
+         */
+        @ExperimentalPreviewRevenueCatUIPurchasesAPI
```
tonidero, line 143: I'm unsure if still want to keep these as experimental with the new approach. Feels much cleaner so I think it might be ok to release as stable... Wdyt @JayShortway @RevenueCat/coresdk ?

JayShortway: I don't have a strong opinion on this. _If_ we were keeping it as experimental, I don't think it needs to be experimental for long.

tonidero: Yeah... Let's keep it as experimental for this PR, but will likely revisit after having the view API done as well

PR #3145, opened 2026-02-26
FIX ANR - Removed player.stop() + player.reset() from teardown.

JZDesign:
### Checklist
- [ ] If applicable, unit tests (module compile checks passed; full unit suite has unrelated existing failures)

### Motivation
Main-thread teardown could stall (ANR-style) when dismissing a paywall video backed by HTTP media. The reported stack shows:
`TextureVideoView.release()` -> `MediaPlayer.reset()` -> `MediaHTTPConnection.disconnect()` on the main thread.

This blocks fragment transition / view detach and can surface as a crash report.

### Description
This PR updates `TextureVideoView.release()` in `VideoView.kt` to avoid synchronous disconnect work on the UI thread during detach:

- Remove `player.stop()` and `player.reset()` from teardown.
- Set `released = true` immediately to make teardown idempotent and guard re-entrant callbacks.
- Hide and clear `MediaController`.
- Detach the surface (`player.setSurface(null)`) and directly call `player.release()`.

Why not just dispatch `reset()` to a background thread?

- `reset()` is for reuse; this code path is disposal, so `reset()` is unnecessary.
- The player lifecycle is currently main-thread-owned. Dispatching only part of teardown to a worker introduces cross-thread state races (`pause` / `setSurface` / `start` callbacks vs `reset` / `release`).
- Moving only `reset()` may still leave UI stalls due to internal/native locking.
- A robust async approach would require moving the entire MediaPlayer lifecycle to a dedicated thread, which is a larger refactor.

This change is the smallest safe fix aligned with the observed stack trace.

### Validation
- `./gradlew :ui:revenuecatui:compileDefaultsBc8DebugKotlin`
- `./gradlew :ui:revenuecatui:compileDefaultsBc7DebugKotlin`
- `./gradlew :ui:revenuecatui:testDefaultsBc8DebugUnitTest` (fails due to pre-existing unrelated tests: `TextComponentViewVariablesTests`, `PaywallComponentsTemplatePreviewRecorder`)

---

> [!NOTE]
> **Medium Risk**
> Touches Android `MediaPlayer` lifecycle/teardown behavior, which can surface as playback edge cases if release ordering differs across devices/streams. Scope is small and localized to `VideoView` disposal.
> 
> **Overview**
> Prevents UI-thread stalls when dismissing paywall videos by changing `TextureVideoView.release()` to be idempotent earlier (`released = true` up front), hide/clear the `MediaController`, detach the surface (`setSurface(null)`), and then `release()` the `MediaPlayer`.
> 
> Removes teardown calls to `player.stop()` and `player.reset()`, which can block for network-backed streams, and adds a guarded error log message specifically for `MediaPlayer.release()` failures.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3be34e4a549a17329f1136f7824bd625c75c4a19. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
@@ -237,15 +237,18 @@ private class TextureVideoView @JvmOverloads constructor(
 
     fun release() {
         if (released) return
-        controller?.hide()
-        safely(execute = {
-            player.stop()
-        })
-        safely(execute = {
-            player.reset()
-        })
-        player.release()
         released = true
+        controller?.hide()
+        controller = null
+        // Detach surface and release directly. reset()/stop() can block for network streams.
+        safely(execute = {
+            player.setSurface(null)
+        })
+        safely(execute = {
+            player.release()
+        }, failureMessage = { e ->
+            "Could not release media player: ${e.message}"
+        })
         if (viewTreeObserverListening) {
             viewTreeObserver.removeOnGlobalLayoutListener(layoutListener)
             viewTreeObserverListening = false


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
```diff
@@ -237,15 +237,18 @@ private class TextureVideoView @JvmOverloads constructor(
...
         controller?.hide()
+        controller = null
+        // Detach surface and release directly. reset()/stop() can block for network streams.
         safely(execute = {
-            player.stop()
+            player.setSurface(null)
         })
         safely(execute = {
-            player.reset()
+            player.release()
```
tonidero, line 248: So just to check, this won't need to perform any heavy work? Just to make sure :)

tonidero (dismissed):
Just a question, but if we've checked that, I think this should be ok. Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/video/VideoView.kt
JZDesign: I am not 100% sure so I'm asking a model. 

I asked Codex 5.3 Extra High `if thread blocking is a risk and we're not reusing the media player ever, should we do it on a background thread`. It responded:

> Short answer: not as an isolated change.
>  • reset() was the proven blocker, and we removed it; that addresses the reported ANR path.
>  • release() can still block, but usually much less often/less severely than reset() for HTTP teardown.
>  • Running only release() on a background thread while everything else stays on main introduces race risk (main-thread surface/lifecycle callbacks vs worker-thread release() on the same MediaPlayer instance).
>  • Because this class currently “owns” MediaPlayer on main, partial offloading is the unsafe middle ground.
>
>  If we want background teardown done correctly, the right approach is:
>  • create/use/release MediaPlayer on a dedicated HandlerThread,
>  • serialize all player calls on that thread,
>  • marshal UI updates back to main.
>
>  So: yes in principle, but only with full thread-ownership refactor; no for just “dispatch release to background” in the current architecture.

JZDesign: This is an option too https://github.com/RevenueCat/purchases-android/pull/3148

JZDesign: After lots of digging… yes it can still block. It's more rare

tonidero: Hmm right... Then I guess that ideally we can do the proper fix and offload video operations to a background thread as you have in that other PR yeah... We should also see how that works with multiple videos.

JZDesign: Yea, I could remove the release and we can ship this as a hotfix then make the next fix in a less rushed state. Downside of not releasing is we rely on system cleanup of the media player processes

tonidero: Hmm I'm just concerned with making a "halfway" fix... If we think this could provide a noticeable improvement while we fully fix it, I wouldn't be too opposed, just scared of introducing other issues... So for something that I don't think it's completely urgent, it might be worth waiting until we have bandwidth to do the "correct" fix?

PR #3217, opened 2026-03-11
Pin GCP CLI version to fix run-firebase-tests

fire-at-will:
### Description
The `run-firebase-tests` job in the native Android release pipeline is repeatedly failing today on the `Install latest gcloud CLI version if not available step` with the following failure message:

```bash
Failure:
/bin/bash: line 141: python: command not found
Detected platform: linux ()
The version installed (559.0.0) differs from the version requested (560.0.0).
Uninstalling v559.0.0...
sudo is required to uninstall the Google Cloud SDK.
Please install it and try again.
Failed to uninstall the current version.

Exited with code exit status 1
```

This address the failure by pinning the GCP CLI version used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,9 @@ commands:
         type: string
     steps:
       - gcp-cli/setup:
+          # Pin to the version preinstalled in the gcp-cli executor image.
+          # Using "latest" can trigger an uninstall/reinstall path that fails because sudo is unavailable.
+          version: 559.0.0
           gcloud_service_key: GCLOUD_SERVICE_KEY
           google_compute_zone: GOOGLE_COMPUTE_ZONE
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1115,7 +1124,7 @@ workflows:
       - run-maestro-e2e-tests:
           requires:
             - prepare-tests
-      - integration-tests-build: *release-branches
+      - integration-tests-build
```
fire-at-will, line 1127: TODO: Revert this when we see that the job passes

ajpallares (commented):
I worry what would happen if the executor version ever changes?

I see [here](https://circleci.com/developer/orbs/orb/circleci/gcp-cli) that there's `skip_installation` which seems like does what we need?

PR #3234, opened 2026-03-13
Unified StoreReplacementMode API

fire-at-will:
### Motivation
With the introduction of the Galaxy Store, we needed to find a way to gracefully model our replacement mode APIs for product changes, now that we have two stores in the Android SDK that support product changes with replacement modes.

Our current replacement mode APIs are store-specific. This allows you to tailor the replacement modes to each store, but puts the onus on the developer to maintain different logic for each store to call the appropriate `PurchaseParams.Builder` function to set the replacement mode for that store.

### Description
This PR introduces a new unified `StoreReplacementMode` abstract class that can be used across both the Play Store and the Galaxy Store. It replaces the existing `GoogleReplacementMode` and `GalaxyReplacementMode` enums. It has the same fields as the existing `GoogleReplacementMode`:
```kotlin
public abstract class StoreReplacementMode internal constructor(
    override val name: String,
) : ReplacementMode {
public companion object {
        @JvmField public val WITHOUT_PRORATION: StoreReplacementMode = WithoutProration
        @JvmField public val WITH_TIME_PRORATION: StoreReplacementMode = WithTimeProration
        @JvmField public val CHARGE_FULL_PRICE: StoreReplacementMode = ChargeFullPrice
        @JvmField public val CHARGE_PRORATED_PRICE: StoreReplacementMode = ChargeProratedPrice
        @JvmField public val DEFERRED: StoreReplacementMode = Deferred
}
```

The following table documents how each `StoreReplacementMode` maps to each store:
| StoreReplacementMode  | Play Store Replacement Mode | Galaxy Replacement Mode                        |
|-----------------------|-----------------------------|------------------------------------------------|
| WITHOUT_PRORATION     | WITHOUT_PRORATION           | INSTANT_NO_PRORATION                           |
| WITH_TIME_PRORATION   | WITH_TIME_PRORATION         | INSTANT_PRORATED_DATE                          |
| CHARGE_FULL_PRICE     | CHARGE_FULL_PRICE           | N/A - error returned if provided to purchase() |
| CHARGE_PRORATED_PRICE | CHARGE_PRORATED_PRICE       | INSTANT_PRORATED_CHARGE                        |
| DEFERRED              | DEFERRED                    | DEFERRED                                       |

Additionally, this PR **deprecates** the existing `GoogleReplacementMode` enum and associated public APIs and hard-removes the experimental `GalaxyReplacementMode` enum and its APIs.

Under the hood, it also updates the product change flows for both the Galaxy and Play Stores to use the `StoreReplacementMode` class instead of the store-specific enums.

#### Additions
- `StoreReplacementMode` abstract class
- `PurchaseParams.replacementMode: StoreReplacementMode` (defaults to WITHOUT_PRORATION)
- `PurchaseParams.Builder.replacementMode(replacementMode: StoreReplacementMode)`

#### Deprecations
- `GoogleReplacementMode` - use `StoreReplacementMode` instead
- `PurchaseParams.googleReplacementMode` - use `PurchaseParams.replacementMode` instead
- `PurchaseParams.Builder.googleReplacementMode` - use `PurchaseParams.Builder.replacementMode` instead
- `PurchaseParams.Builder.googleReplacementMode(googleReplacementMode: GoogleReplacementMode)` - use `PurchaseParams.Builder.replacementMode(replacementMode: StoreReplacementMode)` instead

#### Hard Removals
These are all currently marked as `@ExperimentalPreviewRevenueCatPurchasesAPI`:
- `GalaxyReplacementMode` - use `StoreReplacementMode` instead
- `PurchaseParams.galaxyReplacementMode` - use `PurchaseParams.replacementMode` instead
- `PurchaseParams.Builder.galaxyReplacementMode` - use `PurchaseParams.Builder.replacementMode` instead
- `PurchaseParams.Builder.galaxyReplacementMode(galaxyReplacementMode: GalaxyReplacementMode)` - use `PurchaseParams.Builder.replacementMode(replacementMode: StoreReplacementMode)` instead

### Paywalls Changes
This PR includes minor changes to the paywalls product to account for this change, primarily in how product changes and custom purchase logic are handled.

### Testing
- Lots of unit tests
- Manually tested:
   - Product changes on the Galaxy Store using the new `StoreReplacementAPI` work
   - Confirmed that making a product change on the Galaxy Store with `StoreReplacementAPI.CHARGE_FULL_PRICE` returns an unsupported error
   - Product changes on the Play Store using the deprecated `GoogleReplacementAPI` still work
   - Product changes on the Play Store using the new `StoreReplacementAPI` work

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes public purchase/product-change APIs and the proration/replacement-mode handling across Play Store and Galaxy flows, which can affect upgrade/downgrade behavior and backend payloads.
> 
> **Overview**
> Introduces a unified `StoreReplacementMode` and wires it through `PurchaseParams`/`PurchaseParams.Builder` (`replacementMode`), while **deprecating** `GoogleReplacementMode` APIs and **removing** `GalaxyReplacementMode`.
> 
> Updates Play and Galaxy product-change flows to consume `StoreReplacementMode` end-to-end (including backend `proration_mode` mapping and Google Billing upgrade params), adds store-specific conversion/mapping helpers, and tightens Galaxy behavior by rejecting unsupported modes (e.g. `CHARGE_FULL_PRICE`) with explicit errors and fallbacks when an unknown mode is provided.
> 
> Refreshes examples and API tester code to use the new API, and updates/expands unit tests and ABI snapshots to validate mappings, parcelization, and deferred product-change behavior on both stores.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1930d366748be15712ff9aa694ad4d78f3360299. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/ReplacementMode.kt
```diff
@@ -44,33 +46,69 @@ private val GoogleReplacementMode.asLegacyProrationMode: LegacyProrationMode
...
+        Store.GALAXY -> when (this) {
+            is GoogleReplacementMode -> this.storeReplacementMode.galaxyName
+            is StoreReplacementMode -> this.galaxyName
+            else -> null
+        }
+        else -> null
+    }
+}
+
 internal object ReplacementModeSerializer : KSerializer<ReplacementMode> {
```
fire-at-will, line 50: This PR updates `ReplacementModeSerializer` to be asymmetric to account for the migration from `GoogleReplacementMode` -> `StoreReplacementMode`. 

For serialization, it always serializes to `StoreReplacementMode`, even if the object is a `GoogleReplacementMode`.

For deserialization, it supports both `GoogleReplacementMode` and `StoreReplacementMode` as inputs to maintain backwards compatibility, but always deserializes to a `StoreReplacementMode`.

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/conversions/StoreReplacementModeConversions.kt
```diff
@@ -0,0 +1,30 @@
...
+
+@Throws(PurchasesException::class)
+internal fun StoreReplacementMode.toGalaxyReplacementMode(): HelperDefine.ProrationMode {
+    return when (this) {
+        StoreReplacementMode.WITHOUT_PRORATION -> HelperDefine.ProrationMode.INSTANT_NO_PRORATION
+        StoreReplacementMode.WITH_TIME_PRORATION -> HelperDefine.ProrationMode.INSTANT_PRORATED_DATE
+        StoreReplacementMode.CHARGE_PRORATED_PRICE -> HelperDefine.ProrationMode.INSTANT_PRORATED_CHARGE
+        StoreReplacementMode.DEFERRED -> HelperDefine.ProrationMode.DEFERRED
+        StoreReplacementMode.CHARGE_FULL_PRICE -> {
+            val message = GalaxyStrings.CHARGE_FULL_PRICE_NOT_SUPPORTED
```
fire-at-will: If `StoreReplacementMode.CHARGE_FULL_PRICE` is provided to the Galaxy store for a product change, we return an error, since there is no equivalent of this mode for the Galaxy Store.

We could gracefully degrade to a default here, but I thought it might be best to fail since the request can't be executed as specified. What do y'all think?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchaseLogic.kt
```diff
@@ -137,7 +137,7 @@ public class PaywallPurchaseLogicParams internal constructor(
 
         /**
          * Sets the replacement mode for this product change.
-         * For Google Play, use [com.revenuecat.purchases.models.GoogleReplacementMode].
+         * Use [com.revenuecat.purchases.models.StoreReplacementMode].
          */
         public fun replacementMode(replacementMode: ReplacementMode?): Builder = apply {
```
fire-at-will, line 142: I don't believe this is technically a breaking change since `.replacementMode()` uses `ReplacementMode`

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementModeConversions.kt
```diff
@@ -0,0 +1,90 @@
...
+@BillingFlowParams.SubscriptionUpdateParams.ReplacementMode
+internal fun StoreReplacementMode.toPlayBillingClientMode(): Int {
+    return mapping.playBillingClientMode
+}
+
+internal fun StoreReplacementMode.legacyPlayBackendName(): String {
+    return mapping.legacyPlayBackendName
+}
+
+internal fun StoreReplacementMode.storeBackendName(store: Store): String? {
```
tonidero, line 64: Could have been nice to try to consolidate these in the backend... But that should be a different PR.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
```diff
@@ -0,0 +1,46 @@
+package com.revenuecat.purchases.models
+
+import com.revenuecat.purchases.ReplacementMode
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Enum of possible replacement modes to be used when performing a subscription product change.
+ */
+@Parcelize
+public enum class StoreReplacementMode : ReplacementMode {
```
tonidero: I'm thinking... should we take the chance to make this not an enum? Since we want to avoid enums/sealed classes in public APIs when possible.... The approach we've been folllowing can be seen for example with the [CustomVariableValue](https://github.com/RevenueCat/purchases-android/blob/3a6b6247fd630af5df8b7244bccd3749c5f52db5/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/CustomVariableValue.kt#L29) type. It would mean we can't have exhaustive `when-else` statements... but I think that's exactly the point, for public APIs 😅 . Wdyt?

tonidero (commented):
Great work! My only comment is whether we should try to avoid using enums for the new type now that we have the chance. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
fire-at-will: If this is the direction we're trying to take our APIs, I'm more than happy to go with the pattern! Will update the PR to use an abstract class like `CustomVariableValue` does :)

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
```diff
@@ -0,0 +1,78 @@
...
+    private object Deferred : StoreReplacementMode("DEFERRED")
+
+    public companion object {
+        /**
+         * Old subscription is cancelled, and new subscription takes effect immediately.
+         * User is charged for the full price of the new subscription on the old subscription's expiration date.
+         *
+         * This is the default behavior for the Galaxy and Play stores.
+         */
+        @JvmField public val WITHOUT_PRORATION: StoreReplacementMode = WithoutProration
```
fire-at-will: We're using these so that the API in Java can be `StoreReplacementMode.WITHOUT_PRORATION` instead of `StoreReplacementMode.WITHOUT_PRORATION.Instance`

feature/galaxy/src/main/kotlin/com/revenuecat/purchases/galaxy/GalaxyBillingWrapper.kt
```diff
@@ -407,22 +407,24 @@ internal class GalaxyBillingWrapper(
...
         val productId = galaxyPurchaseInfo.productId
 
-        if (replaceProductInfo != null) {
-            val galaxyReplacementMode = replaceProductInfo.replacementMode as? GalaxyReplacementMode
-                ?: GalaxyReplacementMode.default
+        replaceProductInfo?.let { replaceInfo ->
+            val replacementMode = when (val mode = replaceInfo.replacementMode) {
+                is StoreReplacementMode -> mode
+                else -> StoreReplacementMode.WITHOUT_PRORATION
+            }
```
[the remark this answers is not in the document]
fire-at-will: Addressed in 0d7e55dc34df44ab54f6a694c4a9ea4a5649dcfa

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import com.revenuecat.purchases.ReplacementMode
+import kotlinx.parcelize.Parcelize
+
+/**
+ * Enum of possible replacement modes to be used when performing a subscription product change.
+ */
+public abstract class StoreReplacementMode internal constructor(
+    override val name: String,
+) : ReplacementMode {
```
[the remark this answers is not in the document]
fire-at-will: added in 0448946a146d4239dbc12f19d417a88e2d8f54e8

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementModeConversions.kt
```diff
@@ -0,0 +1,90 @@
...
+    return when (store) {
+        Store.PLAY_STORE -> legacyPlayBackendName()
+        Store.GALAXY -> mapping.galaxyBackendName
+        else -> null
+    }
```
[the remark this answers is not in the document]
fire-at-will: There is no `CHARGE_FULL_PRICE` equivalent for the galaxy store, so this should be okay.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
fire-at-will: Updated to use an abstract class in ce48c17b83d713547343abe63be2b2634ea9dee2. Let me know what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1594,7 +1591,7 @@ internal class PurchasesOrchestrator(
                 // We also need to normalize oldProductId by stripping any basePlanId suffix
                 // (e.g., "productId:basePlanId" becomes "productId") to ensure the callback key matches the productId
                 // in the transaction returned by Google Play, which only contains the product ID without the base plan.
-                val productId = if (googleReplacementMode == GoogleReplacementMode.DEFERRED) {
+                val productId = if (replacementMode == StoreReplacementMode.DEFERRED && store == Store.PLAY_STORE) {
```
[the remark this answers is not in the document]
fire-at-will: I'll check and see if the galaxy store also exhibits this behavior or not. Not sure why it would, but you never know

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
```diff
@@ -17,10 +19,11 @@ public class PurchaseParams(public val builder: Builder) {
 
     public val isPersonalizedPrice: Boolean?
     public val oldProductId: String?
+
+    @Deprecated("Use replacementMode instead")
     public val googleReplacementMode: GoogleReplacementMode
```
tonidero, line 24: Should we just make this a computed property based on `replacementMode` to avoid the breaking change, but avoid having the same state stored twice?

tonidero (approved):
I think this looks good! Just a small suggestion, but nothing blocking

purchases/src/main/kotlin/com/revenuecat/purchases/PurchaseParams.kt
fire-at-will: Good idea! Implemented this in 7d343118ed440505a0916aef62aec728e68377b3. I was able to make `PurchaseParams.googleReplacementMode` computed, and remove the internal `PurchaseParams.Builder.googleReplacementMode` property entirely.

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Type of possible replacement modes to be used when performing a subscription product change.
+ */
+public abstract class StoreReplacementMode internal constructor(
+    override val name: String,
+) : ReplacementMode {
+
+    override fun toString(): String = name
+
+    @Parcelize
```
tonidero: Could we remove these from here and make `StoreReplacementMode` parcelize instead? We might need to make the constructor public, but we can make the constructor `@InternalRevenueCatAPI` I think?

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
tonidero: I'm wondering, could we just remove the ones above (outside the `companion object`) and just create the corresponding `StoreReplacementMode`s here directly? Not sure if we need that level of indirection...

purchases/src/main/kotlin/com/revenuecat/purchases/models/StoreReplacementMode.kt
fire-at-will: Updated to do that in 1c00c222c8e632136f1d066cc48946ebb8890f59 - doesn't look like we needed to make the constructor public

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/common/serializers/ReplacementModeDeserializers.kt
```diff
@@ -1,14 +1,44 @@
...
+        PrimitiveSerialDescriptor("StoreReplacementMode", PrimitiveKind.STRING)
+
+    override fun deserialize(decoder: Decoder): StoreReplacementMode {
+        val key = decoder.decodeString()
+        return valuesByType[key] ?: defaultValue
+    }
+
+    override fun serialize(encoder: Encoder, value: StoreReplacementMode) {
+        throw NotImplementedError("Serialization is not implemented because it is not needed.")
+    }
```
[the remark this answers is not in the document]
fire-at-will: `EnumDeserializerWithDefault.serialize` also threw `NotImplementedError`, so this isn't new.

tonidero (approved):
🚢

PR #3253, opened 2026-03-17
Add placement and targeting context to paywall events

dpannasch:
Include presentedOfferingContext (placementIdentifier, targetingRevision, targetingRuleId) in paywall event serialization so that events & charts can utilize those dimensions.

### Checklist

### Description

---

> [!NOTE]
> **Medium Risk**
> Changes the schema of emitted paywall events (new optional field), which could impact backend analytics/parsing if assumptions are strict, though the field is nullable and covered by compatibility tests.
> 
> **Overview**
> Paywall backend event payloads now optionally include a nested `presented_offering_context` object (placement + targeting metadata) so downstream processing can segment paywall events by placement/targeting.
> 
> Conversion paths (`PaywallEvent.toBackendStoredEvent` and `PaywallStoredEvent.toBackendEvent`) populate this new field via `PresentedOfferingContextData.fromContext`, which omits the object entirely when no placement/targeting is present, and tests were added to validate request JSON, round-trip serialization, and backward-compatible deserialization of older stored events without the new field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e64fda053deb6e92584c02849fe045b51bc4d1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
@@ -100,7 +100,7 @@ internal sealed class BackendEvent : Event {
         @SerialName("locale")
         val localeIdentifier: String,
         @SerialName("presented_offering_context")
-        val presentedOfferingContext: PresentedOfferingContextBackend? = null,
+        val presentedOfferingContext: PresentedOfferingContextData? = null,
         @SerialName("exit_offer_type")
         val exitOfferType: String? = null,
         @SerialName("exit_offering_id")
@@ -154,7 +154,7 @@ internal sealed class BackendEvent : Event {
     ) : BackendEvent()
 
     @Serializable
-    data class PresentedOfferingContextBackend(
+    data class PresentedOfferingContextData(
         @SerialName("placement_identifier")
         val placementIdentifier: String? = null,
         @SerialName("targeting_revision")
@@ -165,11 +165,11 @@ internal sealed class BackendEvent : Event {
         companion object {
             fun fromContext(
                 context: com.revenuecat.purchases.PresentedOfferingContext,
-            ): PresentedOfferingContextBackend? {
+            ): PresentedOfferingContextData? {
                 if (context.placementIdentifier == null && context.targetingContext == null) {
                     return null
                 }
-                return PresentedOfferingContextBackend(
+                return PresentedOfferingContextData(
                     placementIdentifier = context.placementIdentifier,
                     targetingRevision = context.targetingContext?.revision,
                     targetingRuleId = context.targetingContext?.ruleId,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.common.events
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.PresentedOfferingContext
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
 import com.revenuecat.purchases.customercenter.events.CustomerCenterDisplayMode
 import com.revenuecat.purchases.customercenter.events.CustomerCenterEventType
@@ -164,7 +165,7 @@ internal sealed class BackendEvent : Event {
     ) {
         companion object {
             fun fromContext(
-                context: com.revenuecat.purchases.PresentedOfferingContext,
+                context: PresentedOfferingContext,
             ): PresentedOfferingContextData? {
                 if (context.placementIdentifier == null && context.targetingContext == null) {
                     return null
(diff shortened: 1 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -113,6 +115,31 @@ internal sealed class BackendEvent : Event {
...
+        @SerialName("placement_identifier")
+        val placementIdentifier: String? = null,
+        @SerialName("targeting_revision")
+        val targetingRevision: Int? = null,
+        @SerialName("targeting_rule_id")
+        val targetingRuleId: String? = null,
+    ) {
+        companion object {
+            fun fromContext(
+                context: com.revenuecat.purchases.PresentedOfferingContext,
```
tonidero: Should import instead of using the fully-qualified name :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -113,6 +115,31 @@ internal sealed class BackendEvent : Event {
         val errorMessage: String? = null,
     ) : BackendEvent()
 
+    @Serializable
+    data class PresentedOfferingContextData(
```
tonidero, line 158: Just to check... but would it be worth to move the `offering_identifier` inside this as well to mimic the `PresentedOfferingContext` structure? (If were did that, we could just use a custom serializer for the `PresentedOfferingContext` instead of creating a new type).

It's true the backend would need to keep listening to the offering id in the current location due to existing clients though... So I'm ok leaving that for later.

purchases/src/test/java/com/revenuecat/purchases/paywalls/events/PaywallEventSerializationTests.kt
```diff
@@ -137,6 +137,47 @@ class PaywallEventSerializationTests {
...
+    fun `round trip serialization preserves placement and targeting in backend event`() {
+        val eventString = PaywallStoredEvent.json.encodeToString(impressionEvent)
+        val decodedEvent = PaywallStoredEvent.fromString(eventString)
+        val backendEvent = decodedEvent.toBackendEvent()
+
+        val context = backendEvent.presentedOfferingContext
+        assertThat(context).isNotNull
+        assertThat(context?.placementIdentifier).isEqualTo("placementID")
+        assertThat(context?.targetingRevision).isEqualTo(5)
+        assertThat(context?.targetingRuleId).isEqualTo("ruleID")
```
tonidero: Nitpick. I usually try to assert the full object instead of the individual fields for something like this so if new fields are added, we can more easily add them here.

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
```diff
@@ -99,6 +99,8 @@ internal sealed class BackendEvent : Event {
         val darkMode: Boolean,
         @SerialName("locale")
         val localeIdentifier: String,
+        @SerialName("presented_offering_context")
+        val presentedOfferingContext: PresentedOfferingContextData? = null,
```
tonidero, line 104: I believe it would work but just to confirm. If we have previous events stored without this field, it can still deserialize them correct? (I didn't see a test for this, so making sure)

tonidero (approved):
Looking great! If we've tested this and the backend is already accepting the new propert, let's 🚢

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
rickvdl: Good question! I think indeed the backend will currently not read the value, we could have both of course. However I think I'm actually kind of liking having these as separate types. It's true that they're quite similar, but it does keep us flexible in what we expose publicly vs what we send to the backend. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
rickvdl: Great callout, added a test for that!

purchases/src/test/java/com/revenuecat/purchases/paywalls/events/PaywallEventSerializationTests.kt
rickvdl: Right, agree that's better. Done :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/events/BackendEvent.kt
tonidero: Right... I think if there is no explicit reason currently to split it, it could make sense to have only one, and add a separate model later when/if needed. Mostly to keep things simple. Having said that, I don't really have a strong opinion, and this is all internal so we can keep it as is :)

tonidero (approved):
🚢

PR #3254, opened 2026-03-18
Fix null Placements when offering_ids_by_placement is absent

dpannasch:
When the backend sends a placements object with only fallback_offering_id and no offering_ids_by_placement key (e.g., "all placements" targeting or no targeting match), the Android SDK was discarding the entire Placements object — including the fallback. This caused getCurrentOfferingForPlacement to return null instead of the fallback offering, diverging from iOS behavior.

Default offeringIdsByPlacement to an empty map when the key is missing, matching iOS's @DefaultDecodable.EmptyDictionary behavior.

### Checklist

### Description

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how backend JSON is interpreted for placements, which can affect which offering is returned for a placement. Scope is small and covered by new unit tests for the missing-key cases.
> 
> **Overview**
> Ensures `OfferingParser.createOfferings` always constructs an `Offerings.Placements` object when `placements` is present, defaulting `offeringIdsByPlacement` to `emptyMap()` if `offering_ids_by_placement` is absent so `fallback_offering_id` still applies.
> 
> Adds unit tests covering placements responses with only `fallback_offering_id` (including explicit `null`) to verify `getCurrentOfferingForPlacement` correctly falls back or returns `null` as expected.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddae70bf09123405fdff7d3afa13af828a277a15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
@@ -87,15 +87,12 @@ internal abstract class OfferingParser {
             val offeringIdsByPlacement = it.optJSONObject("offering_ids_by_placement")
                 ?.toMap<String?>()
                 ?.replaceJsonNullWithKotlinNull()
+                ?: emptyMap()
 
-            return@let offeringIdsByPlacement?.let {
-                Offerings.Placements(
-                    fallbackOfferingId = fallbackOfferingId,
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Thanks for the fix!!

ajpallares (approved):
Nice catch!

PR #3300, opened 2026-04-06
Add Workflows network layer

vegaro:
### Motivation

This is the first milestone toward multipage paywalls on Android. The backend exposes two endpoints for workflows: one to list available workflows for a subscriber and one to fetch the full published workflow document. The detail endpoint uses an envelope pattern where the server decides whether to return the workflow payload inline or redirect to a CDN URL. This PR adds the networking layer for the **detail** endpoint so the SDK can fetch and parse workflows end-to-end.

#### Endpoints

One new `Endpoint` subtype: `GetWorkflow` (detail). The endpoint participates in signature verification (same as offerings). The CDN response is hashed and the hash compared when checking signature verification.

#### Response envelope processing

`WorkflowDetailResolver` handles the detail endpoint's envelope outside of `Backend`:
- **`inline`**: unwraps the `data` object as the workflow payload.
- **`use_cdn`**: delegates to an injected `WorkflowCdnFetcher` to download the compiled JSON from the CDN URL.

`WorkflowDetailResolver` is its own class, keeping `WorkflowManager` thin (just orchestration + error mapping) and making envelope logic independently testable.

#### CDN fetching

`WorkflowCdnFetcher` is a `fun interface` so tests can substitute a lambda. The production implementation, `FileCachedWorkflowCdnFetcher`, uses `FileRepository` for disk caching when available and falls back to a direct URL read otherwise. Injected into `Backend` via `PurchasesFactory`.

---

> [!NOTE]
> **Medium Risk**
> Adds a new backend endpoint and end-to-end fetch path (including optional CDN download and hash verification), which can affect networking, caching, and signature-verification behavior. While gated behind `@InternalRevenueCatAPI`, failures here could surface as new `PurchasesError` paths at runtime.
> 
> **Overview**
> Adds an **internal workflows fetch path** from `Purchases.getWorkflowWith`/`awaitGetWorkflow` through `PurchasesOrchestrator` to a new `Backend.getWorkflow` call and `Endpoint.GetWorkflow`.
> 
> Implements workflow envelope handling via new `common.workflows` components (`WorkflowManager`, `WorkflowDetailResolver`, `WorkflowCdnFetcher` with `FileCachedWorkflowCdnFetcher`) to support *inline* vs *CDN* workflow payloads, plus SHA-256 content-hash verification tied to `SignatureVerificationMode`.
> 
> Introduces new workflow models (`PublishedWorkflow`, `WorkflowFetchResult`, response envelope types) and a `JsonObjectToMapSerializer` for decoding dynamic `metadata`, with accompanying unit tests and updated test wiring/baselines.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6e0a7b7090cc2f13a0e526b037d3f2d150141dd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

> [!NOTE]
> **Medium Risk**
> Adds a new backend endpoint and async fetch flow that can introduce new network/caching and error-mapping behavior (including optional CDN reads) even though it’s gated behind `@InternalRevenueCatAPI`.
> 
> **Overview**
> Adds an internal end-to-end fetch path for *published workflows* via a new `GET /v1/subscribers/{userId}/workflows/{workflowId}` endpoint (`Endpoint.GetWorkflow`), wired through `Backend.getWorkflow` and exposed from `PurchasesOrchestrator.getWorkflow`.
> 
> Introduces a new `common.workflows` layer (`WorkflowManager`, `WorkflowDetailResolver`, `WorkflowCdnFetcher`) to handle the server “envelope” response (inline payload vs CDN URL) including disk-cached CDN downloads via `DefaultFileRepository` (now supports configurable cache subdirectories).
> 
> Adds workflow serialization models (`PublishedWorkflow`, envelope types, `WorkflowFetchResult`) plus a `JsonObjectToMapSerializer` for decoding dynamic `metadata`, with accompanying unit tests; also updates detekt baseline and test wiring to account for the new orchestrator dependency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19d4bcca515dde3d6d98cd800c5cc5e2dc1a7469. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowJsonParser.kt
```diff
@@ -0,0 +1,21 @@
+@file:OptIn(InternalRevenueCatAPI::class)
+
+package com.revenuecat.purchases.common.workflows
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import kotlinx.serialization.json.Json
+
+internal object WorkflowJsonParser {
```
facumenzella, line 8: This feels like an unnecessary abstraction to me. Or what am I missing? Can't we reuse an existing parser?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailHttpProcessor.kt
```diff
@@ -0,0 +1,54 @@
...
+            enrolledVariants = enrolledVariants,
+        )
+    }
+
+    private fun JSONObject.toEnrolledVariantsMap(): Map<String, String>? {
+        val map = mutableMapOf<String, String>()
+        val keyIterator = keys()
+        while (keyIterator.hasNext()) {
+            val key = keyIterator.next()
+            map[key] = optString(key)
```
facumenzella: `optString` will convert anything to string? How does that work? Can we add tests for non-string keys?

facumenzella (approved):
Just a couple of questions. Looks good

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -115,7 +128,9 @@ internal class Backend(
     private val eventsDispatcher: Dispatcher,
     private val httpClient: HTTPClient,
     private val backendHelper: BackendHelper,
+    workflowCdnFetcher: WorkflowCdnFetcher = FileCachedWorkflowCdnFetcher(fileRepository = null),
```
tonidero: Hmm why default here to a different one since we're creating the `Backend` instance in the `PurchasesFactory` with a value? If this is for tests, I wonder if we should just make this explicit there... Also, it would make `fileRespository` not nullable, which feels better?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -0,0 +1,104 @@
...
+@Serializable
+internal data class WorkflowSummary(
+    val id: String,
+    @SerialName("display_name") val displayName: String,
+)
+
+@Serializable
+internal data class WorkflowsListResponse(
+    val workflows: List<WorkflowSummary> = emptyList(),
+    @SerialName("ui_config") val uiConfig: UiConfig,
```
tonidero: Just curious... why is the `uiConfig` needed here if we also have it in the individual workflow's model?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -0,0 +1,104 @@
...
+@Serializable
+internal data class PublishedWorkflow(
+    val id: String,
+    @SerialName("display_name") val displayName: String,
+    @SerialName("initial_step_id") val initialStepId: String,
+    val steps: Map<String, WorkflowStep>,
+    val screens: Map<String, WorkflowScreen>,
+    @SerialName("ui_config") val uiConfig: UiConfig,
+    @SerialName("content_max_width") val contentMaxWidth: Int? = null,
+    val metadata: JsonObject? = null,
```
tonidero: Hmm I guess this could have nested data? Otherwise, we could make this a `Map<String, Any>` for ease of use.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailHttpProcessor.kt
```diff
@@ -0,0 +1,54 @@
...
+            return WorkflowDetailHttpProcessingResult(httpResult = result, enrolledVariants = null)
+        }
+        val root = JSONObject(result.payload)
+        val enrolledVariants = root.optJSONObject("enrolled_variants")?.toEnrolledVariantsMap()
+        val rawAction = root.getString("action")
+        val action = WorkflowResponseAction.fromValue(rawAction)
+            ?: throw JSONException("Unknown workflow response action: $rawAction")
+        val normalizedPayload = when (action) {
+            WorkflowResponseAction.INLINE -> root.getJSONObject("data").toString()
+            WorkflowResponseAction.USE_CDN -> workflowCdnFetcher.fetchCompiledWorkflowJson(root.getString("url"))
```
tonidero: I wonder if we should just stick to plain Json serialization here, and avoid the custom parsing we are doing here. Then, do the fetching of the compiled workflow as a higher level later on top of the network call? Feels the network call and the logic are a bit mixed right now...

tonidero (commented):
Not a lot of context and just some thoughts. But happy to iterate on this in future PRs since it's still not used

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -27,6 +27,21 @@ internal sealed class Endpoint(
             }
         }
     }
+
+    data class GetWorkflows(val userId: String) : Endpoint(
+        "/v1/subscribers/%s/workflows",
+        "get_workflows",
+    ) {
+        override fun getPath(useFallback: Boolean) = pathTemplate.format(Uri.encode(userId))
+    }
```
RosieWatson, line 36: Happy to chat more, but we are adding some `type` filtering to define between a web funnel and a paywall type of workflow to refine what is returned (and to mirror how we are adapting workflow endpoints for the dashboard)

Would you want both paywall and web-funnel types returned? Otherwise, sending the param `type=paywall` would reduce the response down to just paywalls

vegaro: Thanks @RosieWatson adding that to this PR now

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
vegaro: I am going to remove it becuase the SDK will most like not use it

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
vegaro: Added a `JsonObjectToMapSerializer` for this. Not sure how it will be used, but I think we should at least expose it same way we expose offering metadata

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -102,6 +117,8 @@ internal sealed class Endpoint(
             LogIn,
             PostReceipt,
             is GetOfferings,
+            is GetWorkflows,
+            is GetWorkflow,
```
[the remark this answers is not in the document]
vegaro: this has been updated

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
```diff
@@ -0,0 +1,134 @@
...
+         * Mirrors the backend's `compute_workflow_json_hash` which uses
+         * `json.dumps(sort_keys=True, separators=(',', ':'))` on the payload without the hash key.
+         */
+        internal fun computeCanonicalHash(json: String): String {
+            val parsed = JSONObject(json)
+            parsed.remove("hash")
+            val canonical = toCanonicalJson(parsed)
+            val digest = MessageDigest.getInstance("SHA-256").digest(canonical.toByteArray())
+            return digest.joinToString("") { "%02x".format(it) }
+        }
```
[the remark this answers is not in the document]
vegaro: https://github.com/RevenueCat/khepri/pull/19460

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -0,0 +1,54 @@
...
+        appInBackground: Boolean,
+        onSuccess: (WorkflowFetchResult) -> Unit,
+        onError: (PurchasesError) -> Unit,
+    ) {
+        backend.getWorkflow(
+            appUserID = appUserID,
+            workflowId = workflowId,
+            appInBackground = appInBackground,
+            onSuccess = { response ->
+                scope.launch {
```
vegaro, line 39: `workflowDetailResolver` internally uses `DefaultFileRepository` which uses its own uses its own `CoroutineScope(Dispatchers.IO + NonCancellable)`

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -394,6 +395,16 @@ public class Purchases internal constructor(
         purchasesOrchestrator.getOfferings(listener)
     }
 
+    @InternalRevenueCatAPI
+    @JvmSynthetic
+    public fun getWorkflowWith(
```
tonidero: Since this is an `InternalRevenueCatAPI` API, no need to create all the different variants? If we need the coroutine version only, no need to expose this one (or the other way around).

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -394,6 +395,16 @@ public class Purchases internal constructor(
...
+        workflowId: String,
+        onError: (PurchasesError) -> Unit,
+        onSuccess: (WorkflowFetchResult) -> Unit,
```
tonidero: Total nitpick.
I wonder if this API should return this type? As in, feels like returning a "Fetch" object indicates it comes directly from the backend, but it can come from cache as well... 

Maybe the object we return should be a `WorkflowResult` or `WorkflowData` or something along those lines, which would not indicate that it comes from network directly?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowCdnFetcher.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Uses [FileRepository] for disk caching.
+ */
+internal class FileCachedWorkflowCdnFetcher(
+    private val fileRepository: FileRepository,
+) : WorkflowCdnFetcher {
+
+    @Throws(IOException::class)
+    override suspend fun fetchCompiledWorkflowJson(cdnUrl: String): String {
+        val url = URL(cdnUrl)
+        val uri = fileRepository.generateOrGetCachedFileURL(url)
```
facumenzella: I wonder if we should use the hash here too 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
```diff
@@ -0,0 +1,76 @@
...
+            WorkflowResponseAction.USE_CDN -> {
+                val url = response.url
+                    ?: error("CDN workflow response missing url")
+                val json = workflowCdnFetcher.fetchCompiledWorkflowJson(url)
+                val workflow = WorkflowJsonParser.parsePublishedWorkflow(json)
+                if (signatureVerificationMode.shouldVerify) {
+                    val expectedHash = response.hash
+                    if (expectedHash != null) {
+                        verifyContentHash(json, url, expectedHash)
+                    }
```
facumenzella: If `shouldVerify = true` and `expectedHash = null` shouldn't we throw an error?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -198,6 +202,16 @@ internal class PurchasesFactory(
                 backendHelper,
             )
 
+            val workflowManager = WorkflowManager(
+                backend = backend,
+                workflowDetailResolver = WorkflowDetailResolver(
+                    workflowCdnFetcher = FileCachedWorkflowCdnFetcher(
+                        fileRepository = DefaultFileRepository(contextForStorage),
```
tonidero: One thing I'm thinking... could we split the downloaded files into a different folder? So the files do not get mixed with paywalls assets or something, which would provide better organization IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
```diff
@@ -0,0 +1,76 @@
...
+            WorkflowResponseAction.INLINE -> {
+                response.data
+                    ?: error("Inline workflow response missing data")
+            }
+            WorkflowResponseAction.USE_CDN -> {
+                val url = response.url
+                    ?: error("CDN workflow response missing url")
+                val json = workflowCdnFetcher.fetchCompiledWorkflowJson(url)
+                val workflow = WorkflowJsonParser.parsePublishedWorkflow(json)
+                if (signatureVerificationMode.shouldVerify) {
```
tonidero: I think we're using `signatureVerificationMode` incorrectly here... `signatureVerificationMode` is used for trusted entitlements, but here we seem to be using it to verify that the downloaded hash matches the contents, which I think it's something we should ALWAYS do. 

It's true this should also partially make sure we're protected against MITM, as long as the workflows list endpoint has Trusted Entitlements enabled. But I think the hash verification we still need to do it always, to make sure we downloaded the correct file.

tonidero: Thinking about it, if there is a mismatch here, we might want to trigger another workflow list operation, in case there is a race condition where:
- We fetch the list of workflows, and have hash ABC
- We start fetching the detail workflow but in the meantime, the workflow is modified
- We get hash XYZ, which does not match.

tonidero (commented):
Some more questions but I think we're close!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
vegaro: yeah I was actually just looking into this in iOS because I think I am mixing two things...

in trusted entitlements mode, we do check the integrity of `/workflows/<identifier>` response, that would be the equivalent of checking the integrity of the `/offerings` endpoint that we do right now. Then the CDN response has a hash associated to it, and we should check the integrity of it at all times, like we do for fonts, or videos in paywalls.

right?

vegaro: > Thinking about it, if there is a mismatch here, we might want to trigger another workflow list operation, in case there is a race condition where:

I am not sure I follow. If the workflow has been modified, the CDN url will be different and the hash as well. 

Example
`GET https://api.revenuecat.com/v1/subscribers/hfioerwpfpeqw/workflows/wfca0fb59a473d4dba`

returns

```
{
    "action": "use_cdn",
    "enrolled_variants": {},
    "hash": "81b73cc0cb60c5e87a6efe5bad8655c8331c89811f49a5e3cbe23b795313ddb0",
    "url": "https://workflowscdn.revenuecat.com/workflows/2403eb57/wfca0fb59a473d4dba/4.json"
}
```

if it changes, the `hash` and `url` are returned together

tonidero: > in trusted entitlements mode, we do check the integrity of /workflows/<identifier> response, that would be the equivalent of checking the integrity of the /offerings endpoint that we do right now. Then the CDN response has a hash associated to it, and we should check the integrity of it at all times, like we do for fonts, or videos in paywalls.

That sounds correct!

> I am not sure I follow. If the workflow has been modified, the CDN url will be different and the hash as well.

Right... sorry my bad. I confused the list workflows with the workflow detail endpoint there. The hash is only part of the detail workflow endpoint, which is validated using trusted entitlements. So we should indeed be ok I think. Please ignore that!

tonidero: One point however we might want to add to the public API we end up using for workflows, is to expose the trusted entitlements verification result in that API somehow... so for `informational` mode, the developer can handle that appropriately (note that this is already an issue for paywalls/offerings... so I wouldn't consider it a blocker, but something we should also think about as a follow-up)

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
```diff
@@ -0,0 +1,61 @@
...
+            workflow = workflow,
+            enrolledVariants = response.enrolledVariants,
+        )
+    }
+
+    @Throws(SignatureVerificationException::class)
+    private fun verifyContentHash(json: String, url: String, expectedHash: String) {
+        val actualHash = computeCanonicalHash(json)
+        if (actualHash != expectedHash) {
+            throw SignatureVerificationException(url)
```
tonidero: Hmm I don't think we should use this exception here since it's meant for trusted entitlements... Maybe worth using a different one?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
```diff
@@ -0,0 +1,61 @@
...
+ */
+internal class WorkflowDetailResolver(
+    private val workflowCdnFetcher: WorkflowCdnFetcher,
+) {
+
+    @Throws(IllegalStateException::class, SignatureVerificationException::class)
+    suspend fun resolve(response: WorkflowDetailResponse): WorkflowFetchResult {
+        val workflow = when (response.action) {
+            WorkflowResponseAction.INLINE -> {
+                response.data
```
tonidero, line 21: I've noticed that if the workflow comes inline, we're not really caching it... is that something we should add?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
vegaro: turns out `generateOrGetCachedFileURL` has a checksum parameter that performs the hash check already so I removed this code

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
vegaro: Going to remove any public exposure from this PR, and consider that in the followups

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
vegaro: Leaving for future PR since I am removing the public API from this one

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowDetailResolver.kt
vegaro: Good eye and that's a good question. For now the only case when we are going to be returning `inline` is the first time, if there's no CDN url yet to return becuase it's been uploaded. In that case we don't want to cache as we want to force routing through the CDN next time. I will bring it up though, but this will come as a followup if we do.

vegaro: I also wanted to bring up if we should be caching the response of `/workflows/< idenfitier>` at all (similar to how we cache offerings)

vegaro: For reference https://linear.app/revenuecat/issue/WFL-99/add-workflow-caching-to-workflowmanager-android

tonidero (approved):
Looks great! Thanks for doing this!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
vegaro: did it in [`1f9a000` (#3350)](https://github.com/RevenueCat/purchases-android/pull/3350/commits/1f9a000b63b0584f4d26f21c380539c3e3ab5d92)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
vegaro: did it in [`1f9a000` (#3350)](https://github.com/RevenueCat/purchases-android/pull/3350/commits/1f9a000b63b0584f4d26f21c380539c3e3ab5d92)

PR #3325, opened 2026-04-10
feat(ads): add possibility to update fullscreen content callback

peterporfy:
### Checklist
- [x] Manual testing with sample app overrides 

### Motivation

When ad loading and displaying are separate actions - for example preloading/caching ads and then deciding where to show it later - it can come in handy to change the delegate later to match the use case.

### Description

This PR provides a setTrackingFullScreenContentDelegate method for all the 4 relevant ad types, which can (potentially) replace the underlying user callback, without breaking RC tracking.

The user still needs to be aware of this - but the call is semantically equivalent to the assignment. It can be used even without initializing the ad with RC, which is useful when that might not happen (for example depending on a feature flag).

---

> [!NOTE]
> **Medium Risk**
> Adds new public extension APIs that alter how full-screen ad callbacks are reassigned, which could affect ad lifecycle behavior and RevenueCat tracking if misused. Includes tests, but changes touch analytics/tracking wrappers used at runtime.
> 
> **Overview**
> Adds `setTrackingFullScreenContentCallback` extensions for AdMob `InterstitialAd`, `AppOpenAd`, `RewardedAd`, and `RewardedInterstitialAd` to let apps replace the user `FullScreenContentCallback` **without stripping RevenueCat’s tracking wrapper** (and falls back to direct assignment when no wrapper is present).
> 
> Updates `TrackingFullScreenContentCallback` to allow its delegated callback to be swapped after load, and expands unit coverage to verify delegate swapping, null handling, and fallback assignment behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c8ce9f0252bce839349822d59bdda2ffc3fd527. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


polmiro (approved):

ajpallares (commented):
Makes sense! ~I have one suggestion related to #3318~ My mistake, I got confused 🙏 . All good

PR #3324, opened 2026-04-10
Fix ripple effect ignoring component shape and extending into margins

bahdaniec:
### Motivation

Resolves #3321

The ripple effect in Paywalls V2 ignores rounded corners and extends into the component's margin area. This happens because `Modifier.clickable()` is applied before `StackComponentView` applies `.clip(shape)` and margin internally — so the ripple fills the entire rectangular area including margins.

### Description

Introduces an `interactionModifier` parameter on `StackComponentView` that gets threaded down to `MainStackComponent` and applied **after** `.clip(composeShape)` in all 4 code paths (standard, video background, nested badge, overlay). This ensures the ripple is bounded by the component's actual shape.

Updated all callers:
- **ButtonComponentView** — moves `clickable` into `interactionModifier`
- **PackageComponentView** — moves conditional `clickable` into `interactionModifier`, removes unused `conditional` import
- **TabControlButtonView** — moves `clickable` into `interactionModifier`

Also adds a `ButtonComponent` to the bless sample paywall (#8) to make it easier to reproduce and verify the fix visually.

**Testing:** All existing unit tests pass — StackComponentViewTests, StackComponentViewWindowTests, ButtonComponentViewTests, PackageComponentViewTests, TabsComponentViewTests. Verified visually via the updated sample paywall.

---

> [!NOTE]
> **Medium Risk**
> Touches the core `StackComponentView` container and changes where click/interaction modifiers are applied, which could subtly affect touch targets and interaction behavior across multiple paywall components. Functionality is UI-scoped (ripple/press feedback) with no data/auth implications.
> 
> **Overview**
> Fixes Compose ripple/press feedback on paywall components so it respects rounded corners and does not extend into margins by adding an `interactionModifier` to `StackComponentView` and applying it **after** `clip(shape)` across all rendering paths (standard, video background, nested badge, overlay/badge variants).
> 
> Updates callers (`ButtonComponentView`, `PackageComponentView`, `TabControlButtonView`) to pass their `Modifier.clickable` via `interactionModifier` (instead of on the outer `modifier`), and tweaks the paywall tester sample (#8) to use a `ButtonComponent` to make the issue easy to reproduce/verify.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7368b07999c1c11eb3e49404740709a91cfeebb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3345, opened 2026-04-15
[EXTERNAL] Fix ripple effect ignoring component shape and extending into margins (#3324) contributed by @bahdaniec

### LINKED ISSUE #3321: Ripple effect ignores component shape and extends into margins

## Describe the bug

When tapping interactive elements (buttons, package selectors, tab controls) in Paywalls V2, the ripple effect has two problems:

1. **Ignores rounded corners** -- the ripple renders as a rectangle instead of following the component's shape (pill, rounded rectangle, etc.).
2. **Extends beyond visible bounds** -- the ripple area includes the component's margin, making it visually larger than the actual button or selector.

Both issues stem from the same root cause: `Modifier.clickable()` is applied on the outer modifier, before `StackComponentView` applies its margin and shape internally. In Jetpack Compose, the ripple is bounded by the nearest clip ancestor. Without `.clip(shape)` positioned after the margin and before `.clickable()`, the ripple fills the full rectangular bounds including margins.

## Environment

- Platform: Android (Jetpack Compose)
- SDK version: 10.0.2 (purchases + purchases-ui)
- OS version: Android 14+
- Affects all devices using Paywalls V2 with rounded components

## Steps to reproduce

1. Create a Paywall V2 with buttons or package selectors that have rounded corners (e.g. `Shape.Pill`).
2. Display the paywall using the `Paywall` composable.
3. Tap any button or package selector.
4. **Expected:** Ripple follows the rounded shape of the component and stays within its visible bounds.
5. **Actual:** Ripple fills a rectangular area that extends into the component's margin, ignoring both corner radius and visible bounds.

## Reproducing in paywall-tester

The sample paywall "bless" (#8) uses a plain `StackComponent` for its CTA button, which has no click handler and therefore no ripple. To reproduce the issue, replace the CTA `StackComponent` with a `ButtonComponent` (e.g., `action = ButtonComponent.Action.RestorePurchases`) wrapping the same stack. The attached screenshot was captured after this change.

Image

tonidero:
### Motivation

Resolves #3321

The ripple effect in Paywalls V2 ignores rounded corners and extends into the component's margin area. This happens because `Modifier.clickable()` is applied before `StackComponentView` applies `.clip(shape)` and margin internally — so the ripple fills the entire rectangular area including margins.

### Description

Introduces an `interactionModifier` parameter on `StackComponentView` that gets threaded down to `MainStackComponent` and applied **after** `.clip(composeShape)` in all 4 code paths (standard, video background, nested badge, overlay). This ensures the ripple is bounded by the component's actual shape.

Updated all callers:
- **ButtonComponentView** — moves `clickable` into `interactionModifier`
- **PackageComponentView** — moves conditional `clickable` into `interactionModifier`, removes unused `conditional` import
- **TabControlButtonView** — moves `clickable` into `interactionModifier`

Also adds a `ButtonComponent` to the bless sample paywall (#8) to make it easier to reproduce and verify the fix visually.

**Testing:** All existing unit tests pass — StackComponentViewTests, StackComponentViewWindowTests, ButtonComponentViewTests, PackageComponentViewTests, TabsComponentViewTests. Verified visually via the updated sample paywall.

Contributed by @bahdaniec in #3324

---

> [!NOTE]
> **Low Risk**
> Low risk UI behavior change: reorders where `clickable` is applied so ripple/press effects respect clipping; could subtly affect touch target/ripple rendering across paywall components.
> 
> **Overview**
> Fixes Paywalls V2 press/ripple feedback overflowing rounded corners and margins by introducing `interactionModifier` on `StackComponentView` and applying it *after* shape clipping across all render paths (standard, video background, nested badge, overlay).
> 
> Updates button-like callers (`ButtonComponentView`, `PackageComponentView`, `TabControlButtonView`) to pass their `clickable` modifier via `interactionModifier` instead of the root `modifier`, and tweaks the paywall tester sample to use a real `ButtonComponent` for easier visual verification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b996696e3c406fe5ac6967f0b5130bb033fe1f8f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -645,6 +683,8 @@ private fun MainStackComponent(
         } else {
             stack(
                 outerShapeModifier
+                    .conditional(interactionModifier !== Modifier) { clip(composeShape) }
```
tonidero, line 686: Note this clipping is a difference in behavior from before. Now, if the stack is clickable, we will be clipping the content, which, depending on the paywall, could potentially cause some paywalls to change and be cutoff. For example, if you have a clickable stack, and inside another stack with a badge. The badge could potentially go outside of the bounds of the parent stack before, but now it would be clipped. 

So I'm actually a bit unsure about this fix. I think it makes sense but any thoughts?@RevenueCat/paywallsengine

JayShortway: Compose has [Indication](https://developer.android.com/develop/ui/compose/touch-input/user-interactions/handling-interactions#reusable-custom-effect) APIs to customize the visual feedback from clicks. We might need to look into that?

tonidero:
@bahdaniec after testing this fix more thoroughly, we detected it could cause some unintended consequences in some cases where some content could be cutoff. Instead, we're applying a different fix: https://github.com/RevenueCat/purchases-android/pull/3395. In any case, thank you for letting us know and your contribution!! Will close this for now.

PR #3353, opened 2026-04-16
Fix: dismiss was called before onPurchaseComplete callback invocation

JZDesign:
### Checklist

### Motivation

Apps were missing the on purchase complete event

### Description

Moves finish out of it's handler functions and relies on the on dismiss request. Also uses yield to ensure correct ordering of events

---

> [!NOTE]
> **Medium Risk**
> Changes the paywall activity’s lifecycle/auto-dismiss behavior after purchase/restore, which can affect callback ordering and whether the activity closes in edge cases (e.g., errors, exit offers, entitlement-gated display). Scope is contained to UI paywall flow and related tests.
> 
> **Overview**
> Ensures `PaywallActivity` no longer calls `finish()` directly from purchase/restore callbacks; it now relies on the configured `dismissRequest`/`onDismissRequest` flow to close (or launch an exit offer) after results are set, preventing premature activity termination that could drop `onPurchaseCompleted` events.
> 
> Also wires `requiredEntitlementIdentifier` into the paywall view model via `shouldDisplayBlock` when creating it in the activity, and updates `PaywallViewModelTest`’s factory helper to allow injecting a custom `dismissRequest` for more precise dismissal assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd8c30c39e184e389d92b67ba6eefe4642e7d425. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -180,10 +180,6 @@ internal class PaywallActivity : ComponentActivity() {
             override fun onRestoreCompleted(customerInfo: CustomerInfo) {
                 userListener?.onRestoreCompleted(customerInfo)
                 setResult(RESULT_OK, createResultIntent(PaywallResult.Restored(customerInfo)))
-                val requiredEntitlementIdentifier = args?.requiredEntitlementIdentifier ?: return
-                if (customerInfo.entitlements.active.containsKey(requiredEntitlementIdentifier)) {
-                    finish()
-                }
```
vegaro, line 186: hmm not sure if I am reading this correctly but this is an important change in behavior. It means apps that didn't implement dismissRequest will not dismiss their paywall, right?

JZDesign: Nope, look a little further down the file

JZDesign: We pass it in ourselves

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -396,6 +397,11 @@ internal class PaywallViewModelImpl(
                     shouldDisplayBlock?.let {
                         if (!it(customerInfo)) {
                             _purchaseCompleted.value = true
+                            if (listener != null) {
+                                // Give listener implementations a chance to flush any posted main-thread work
+                                // before the paywall dismisses itself.
+                                yield()
```
vegaro, line 403: I don't think this will always work. It only runs the dispatcher clock by one frame, I don't think it can guarantee it will flush everything that you need

JZDesign: Open to ideas. I was thinking this would work like Swift's Task.yield allowing anything else waiting in the structured concurrency queue to step in and the coroutine would pick this back up when that was done

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
vegaro: Ah `options.dismissRequest()` calls the `dismissRequest` set in line 211, which calls `onDismissRequest`, which finishes

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
```diff
@@ -209,9 +207,8 @@ internal class PaywallActivity : ComponentActivity() {
                     Box(
                         Modifier.fillMaxSize(),
                     ) {
-                        // Empty dismissRequest is overridden below by setDismissRequestWithExitOffering
                         val paywallOptions = PaywallOptions.Builder(
-                            dismissRequest = {},
+                            dismissRequest = { onDismissRequest(exitOffering = null, result = null) },
```
tonidero, line 211: So this is also passed in the `setDismissRequestWithExitOffering` below. TBH, it's confusing to have these both parameters 🥴 . But just to make sure, have you tested with an exit offer as well to make sure we're not dismissing twice?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -557,6 +563,11 @@ internal class PaywallViewModelImpl(
                     val purchaseResult = purchases.awaitPurchase(purchaseParamsBuilder)
                     _purchaseCompleted.value = true
                     listener?.onPurchaseCompleted(purchaseResult.customerInfo, purchaseResult.storeTransaction)
+                    if (listener != null) {
+                        // Give listener implementations a chance to flush any posted main-thread work
+                        // before the paywall dismisses itself.
+                        yield()
```
tonidero, line 569: Hmm I'm confused as for why these are needed... Seems that the call to the listener `onPurchaseCompleted` happens simultaneously and in the same thread? So there shouldn't be a need to wait IMO?

In the tests, I see that we want to handle cases where the consumer of this API changes to the main thread in a non-sync wait. Does this happen somewhere in our code? (I was searching but didn't find it). If not, this seems like an issue that the developer should handle (they shouldn't change threads within these callback functions).

tonidero (commented):
Just a question about the need for the `yield` functions... PaywallActivity changes make sense but just a question.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/activity/PaywallActivity.kt
JZDesign: There is logic inside the paywall that will invoke the exit offering one if there is an exit offering. Then after the exit offering dismisses, this one would trigger. 

But I will test with an exit offering to be sure

JZDesign: This works as expected

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
JZDesign: I understand that, but IMO this is a simple way to eat the pain of the developer.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
JZDesign: There is the fix in the activity, but this is just a nice helper for consumers attempting to switch threads in the callback. If I remember correctly after 2 weeks 🫠

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
vegaro: if I understand correctly, this `yield()` is a way to execute whatever the developer might have scheduled in the main thread in `onPurchaseCompleted`? Am I correct?

I think that even if that's what this is trying to do, that post to main could have posted to main itself (think `Handler.post { Handler.post { ... } }`), so it would require two yields. I don't think this is a reliable way to do what you're trying to do, as you would need a complete flush of the Main thread, and I don't think we should be doing that imho.

vegaro (commented):
The changes in `PaywallActivity` make sense, but looking at the original report, they are not using the `PaywallActivity`...

This is a simplified version of their code:

```
var purchaseCompleted by remember { mutableStateOf(false) }

Paywall(
    options = PaywallOptions.Builder(
        dismissRequest = {
            // Suppress auto-dismiss after purchase so we can run
            // post-purchase logic (e.g. show a link-account sheet)
            // and dismiss manually when ready.
            if (!purchaseCompleted) onCloseClick()
        },
    )
        .setOffering(offering)
        .setListener(object : PaywallListener {
            override fun onPurchaseCompleted(
                customerInfo: CustomerInfo,
                storeTransaction: StoreTransaction,
            ) {
                purchaseCompleted = true
                viewModel.onPurchaseCompleted(customerInfo)
            }

            override fun onRestoreCompleted(customerInfo: CustomerInfo) {
                purchaseCompleted = true
                viewModel.onRestoreCompleted(customerInfo)
            }
        })
        .build(),
)
```

The `yield` changes I am a bit more hesitant about, specially after looking at their code. It’s true that if they are doing any async work in `viewModel.onPurchaseCompleted`, the `yield` would let that work run on the main thread before we call `options.dismissRequest()`. But notice that they set `purchaseCompleted = true` synchronously in the listener, before doing anything else. So by the time `dismissRequest()` fires, the guard already evaluates to `false` and `onCloseClick()` is suppressed yield or no yield, their pattern would work the same independently of the yields.

The other concern is the one I raised earlier: yield only flushes one level of `Handler.post`. If the listener’s posted work itself posts more work (or hops to a different dispatcher), we’re back to racing the dismiss. If we do want a first-class story for "finish my post-purchase flow, then dismiss," I think that deserves an explicit API (e.g. an opt-out flag or a completion signal to the listener) rather than relying on yiel based timing.

Stepping back a bit, the customer says: "the paywall should stay presented (or at least not report 'dismissed') until we finish follow-up steps and then dismiss it." But the SDK doesn’t actually dismiss anything itself, the only thing it does after purchase is call `options.dismissRequest()`, and what happens from there is entirely up to the developer's lambda. With their guard in place (the check for `purchaseCompleted`), the lambda is a no-op and the paywall should stay presented. The "not report dismissed" wording makes me think they’re misinterpreting the callback firing as the SDK dismissing them, when it isn’t.

My suggestion: keep the structural change in PaywallActivity because I think it makes sense and drop the yields (since I think it won't fix their probelm anyways). The actual issue is more likely either (a) something in their `viewModel.onPurchaseCompleted` is unmounting the paywall (navigation, parent state change), or (b) they’re misreading the `dismissRequest` callback as the dismissal itself.

Not sure how easy it would be, would it be worth trying to write a failing test on main using `Paywall()` composable that reproduces the report? If we can’t reproduce it, that strengthens the case that the issue is in their app code, not ours.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
tonidero: Right... I agree with that. Also, conceptually to me, it feels that we're fighting the listener pattern... If within a callback, you try to execute code asynchronously (even if within the main thread), the flow will continue. Trying to handle those async calls are out of the scope of this pattern. 

We could think about changing the pattern so developers can optionally call a "resume-like" function to continue... but that seems overkill and unnecessary here IMO.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
```diff
@@ -1116,6 +1117,7 @@ class PaywallViewModelTest {
         } returns customerInfo
 
         model.restorePurchases()
+        InstrumentationRegistry.getInstrumentation().waitForIdleSync()
```
vegaro: does this work/ is it needed? looks like it's the only thing added to this test?

vegaro (approved):

tonidero (approved):

PR #3365, opened 2026-04-19
Migrate all suspendCoroutine usages to suspendCancellableCoroutine

skydoves:
Migrate all `suspendCoroutine` usages to `suspendCancellableCoroutine` with `isActive` guards.

Why: suspendCoroutine is not cancellation-aware, which means if the calling coroutine scope is cancelled (e.g., ViewModel cleared), the coroutine hangs forever instead of being cleaned up. This causes silent resource leaks and violates structured concurrency.

What changed:
- Replaced suspendCoroutine → suspendCancellableCoroutine across all await* extensions, internal helpers, and UI modules.
- Added `continuation.isActive` checks before every `resume`/`resumeWithException` to prevent `IllegalStateException` when callbacks fire after cancellation
- Added cancellation-specific unit tests.

---

> [!NOTE]
> **Medium Risk**
> Touches many public coroutine `await*` APIs and purchase/UI flows; cancellation now propagates (instead of potentially hanging), which is a behavior change that could affect callers with broad exception handling.
> 
> **Overview**
> Migrates coroutine wrappers across the SDK and RevenueCatUI from `suspendCoroutine` to `suspendCancellableCoroutine` and adds guarded resumption (`safeResume` / `safeResumeWithException` or `isActive` checks) to prevent `IllegalStateException` when callbacks fire multiple times or after cancellation.
> 
> Updates internal helpers (e.g., `IdentityManager.aliasCurrentUserIdTo`, Blockstore calls) and example BillingClient purchase/ack/consume flows to be cancellation-aware, and adds new cancellation-focused tests for key `await*` functions. The changelog notes the **behavior change** that `CancellationException` now propagates from public `await*` APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04f97e636511619f2c0530f58b6d04e4680ec71f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


examples/rcttester/src/main/java/com/revenuecat/rcttester/purchasing/BillingClientAcknowledgeHelper.kt
```diff
@@ -29,11 +29,12 @@ class BillingClientAcknowledgeHelper(private val billingClient: BillingClient) {
             Log.e(TAG, "Failed to connect BillingClient for acknowledge")
             return false
         }
-        return suspendCoroutine { continuation ->
+        return suspendCancellableCoroutine { continuation ->
```
tonidero, line 32: I think the main reason we didn't do this originally is that this is part of the coroutines library, which we didn't originally add. However, we have since added it so it makes sense to do this 🙌 Thank you!!

examples/rcttester/src/main/java/com/revenuecat/rcttester/purchasing/BillingClientAcknowledgeHelper.kt
```diff
@@ -58,11 +59,12 @@ class BillingClientAcknowledgeHelper(private val billingClient: BillingClient) {
             Log.e(TAG, "Failed to connect BillingClient for consume")
             return false
         }
-        return suspendCoroutine { continuation ->
+        return suspendCancellableCoroutine { continuation ->
```
tonidero, line 61: Another point to note is that these are not really cancellable... In fact, we're just ignoring the result value but not cancelling the request. I'm not sure if it would be better to keep it as `suspendCoroutine` to better reflect what's happening... Having said that, I do understand using this has some benefits so not opposed either. cc @JayShortway for thoughts 🙏

examples/rcttester/src/main/java/com/revenuecat/rcttester/purchasing/PurchasesAreCompletedByMyAppUsingBillingClientPurchaseManager.kt
```diff
@@ -198,12 +197,14 @@ class PurchasesAreCompletedByMyAppUsingBillingClientPurchaseManager(
...
         if (billingClient.isReady) return true
-        return suspendCoroutine { continuation ->
+        return suspendCancellableCoroutine { continuation ->
             billingClient.startConnection(object : BillingClientStateListener {
                 override fun onBillingSetupFinished(billingResult: BillingResult) {
-                    continuation.resume(
-                        billingResult.responseCode == BillingClient.BillingResponseCode.OK,
-                    )
+                    if (continuation.isActive) {
+                        continuation.resume(
```
tonidero, line 204: Since we're doing this so many times, I wonder if it would be worth to create a `safeResume` and `safeResumeWithException` functions that check whether the coroutine is active... Then, we could write a custom lint rule to disallow the "non-safe" versions, at least for now. That way, it would be more difficult to forget to add this. Wdyt?

tonidero (commented):
Looking great! Just some thoughts but it's great to have this. Thank you!!

examples/rcttester/src/main/java/com/revenuecat/rcttester/purchasing/BillingClientAcknowledgeHelper.kt
JayShortway: Yea, we've had this contributed in the past, but indeed the underlying operation is not cancellable. We're not using `continuation.invokeOnCancellation` now either, which is what `suspendCancellableCoroutine` would enable. 

Happy to hear other benefits this might bring if I'm missing something!

skydoves: Yea, you're right that the underlying operations aren't truly cancellable we're not cancelling the actual requests, just ignoring the result.

So the main motivation is coroutine lifecycle management (like more higher-level usages): with `suspendCoroutine`, if the calling scope is cancelled (e.g., ViewModel cleared), the coroutine hangs forever. `suspendCancellableCoroutine` lets it unblock immediately and clean up captured references. So callers can rely on standard cancellation (`viewModelScope`, `withTimeout`, etc.) working correctly with our await* APIs.

 Adding `invokeOnCancellation` to actually cancel underlying requests could be a nice follow-up!

examples/rcttester/src/main/java/com/revenuecat/rcttester/purchasing/PurchasesAreCompletedByMyAppUsingBillingClientPurchaseManager.kt
skydoves: Thanks for your suggestion! 

I implemented safeResume / safeResumeWithException and used at every suspendCancellableCoroutine site in this PR.

One note: the helpers are internal in :purchases, which forced a near-duplicate copy in :ui:revenuecatui. If we're adding a lint rule anyway, worth promoting to @InternalRevenueCatAPI and deduping at the same time. I can do it here or in the follow-up PR.

tonidero: Great point... I think that can be part of another PR indeed. Thanks for adding this!

tonidero (approved):
While I think the benefits are small considering the operation itself won't actually cancel, I think in the end it's a net benefit and shouldn't have any drawbacks that I could think of, so let's go with this. Thank you!!

PR #3369, opened 2026-04-20
fix: move Google BillingClient connection off the main thread

tonidero:
## Summary

- Moves Google's `BillingClient.startConnection()` and `endConnection()` to a dedicated `HandlerThread("revenuecat-billing")` to address reported ANRs when the connection work blocks the main thread during `bindService`.
- Play Billing Library supports calling `startConnection()` from a background thread. Pending-request draining and consumer callbacks still run on the main `Handler`, so observable threading for SDK consumers is unchanged.
- Amazon keeps its main-thread requirement (Amazon's SDK needs `registerListener` to run synchronously during `Activity.onCreate` to detect the foreground activity).
- API refactor: collapses `BillingAbstract.startConnectionOnMainThread(delay)` and the parameterless `startConnection()` into a single `startConnection(delayMilliseconds: Long = 0)` entry point. `BillingAbstract` is `@InternalRevenueCatAPI`, so the rename only affects hybrid SDKs.

## Changes

- **`BillingAbstract`**: single `startConnection(delay)` entry point, with KDoc noting that impls choose the thread.
- **`BillingWrapper`** (Google):
  - New constructor-injected `backgroundHandler: Handler`.
  - `startConnection(delay)` now posts to `backgroundHandler`; the actual work lives in a new private `performStartConnection()` helper.
  - `endConnection()` also posts to `backgroundHandler` for symmetry.
  - Removed the self-imposed `@UiThread` annotation on `ClientFactory.buildClient` — `BillingClient.newBuilder()` doesn't require the UI thread.
  - Pending-request queue draining (`executePendingRequests`) still posts to `mainHandler`, preserving callback-thread semantics for callers.
- **`BillingFactory`**: creates a `HandlerThread("revenuecat-billing")` for the `Store.PLAY_STORE` branch and passes its looper into `BillingWrapper`.
- **`AmazonBilling`**: override renamed; retains the existing `runOnUIThread { … }` and the comment explaining the Amazon main-thread requirement. Connection logic extracted to `performStartConnection()` (internal `@VisibleForTesting`).
- **`GalaxyBillingWrapper`**, **`SimulatedStoreBillingWrapper`**: override renamed; no behavior change.
- **`PurchasesOrchestrator`**: call site renamed.
- **Tests** (`BillingWrapperTest`, `BaseBillingUseCaseTest`, `BasePurchasesTest`, `SimulatedStoreBillingWrapperTest`): pass through to the new signature; the existing mocked `Handler` is reused for both `mainHandler` and `backgroundHandler` so `post`/`postDelayed` keep running synchronously in tests.
- **`baseline-prof.txt`**: symbol renames updated.

## Test plan

- [x] `./gradlew :purchases:testDefaultsBc8DebugUnitTest`
- [x] `./gradlew :purchases:testDefaultsBc7DebugUnitTest`
- [x] `./gradlew detektAll`
- [x] `./gradlew :ui:revenuecatui:compileDefaultsBc8DebugKotlin` + `:feature:amazon:compileDefaultsBc8DebugKotlin`
- [x] Manual: cold-start `purchase-tester` with StrictMode enabled; confirm no main-thread `bindService` violation from `BillingClient.startConnection`.
- [x] Manual: verify product fetch, purchase, and restore still work on a real device.
- [x] Integration tests on CI

---

> [!NOTE]
> **Medium Risk**
> Changes threading and lifecycle behavior around Play Billing service connection, which can affect stability (race conditions, leaked threads, missed callbacks) if not handled correctly, though the change is localized to billing wrappers and covered by updated tests.
> 
> **Overview**
> Moves Google Play Billing connection setup/teardown off the main thread by introducing a background `Handler`/owned `HandlerThread` in `BillingWrapper`, posting `startConnection(...)` and `endConnection()` work there while keeping pending-request callbacks on the main `Handler`.
> 
> Refactors the billing abstraction to a single `BillingAbstract.startConnection(delayMilliseconds)` entry point (removing `startConnectionOnMainThread`/parameterless `startConnection`) and updates Amazon, Galaxy, Simulated Store, orchestrator call sites, tests, and `baseline-prof.txt` to the new API; `BillingWrapper.close()` now also shuts down the owned background thread.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b39d21ea04070b98548d09fd883348a9137da600. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt
```diff
@@ -38,13 +39,17 @@ internal object BillingFactory {
...
             )
-            Store.PLAY_STORE -> BillingWrapper(
-                BillingWrapper.ClientFactory(application, pendingTransactionsForPrepaidPlansEnabled),
-                Handler(application.mainLooper),
-                cache,
-                diagnosticsTrackerIfEnabled,
-                stateProvider,
-            )
+            Store.PLAY_STORE -> {
+                val backgroundThread = HandlerThread("revenuecat-billing").apply { start() }
```
tonidero, line 43: Note I decided to use a separate thread from the one we use for network connections and other things, so we avoid overloading that single thread with many operations. This thread will be dedicated to just connecting/disconnecting from the billing client, so won't have much work normally.

rickvdl: Do we need to tear down this tread at some point? Reading the docs we shouldn't call [`stop()`](https://developer.android.com/reference/java/lang/Thread#stop()), but do you know what will happen if we reconfigure the SDK and create a new `HandlerThread` with the same identifier again?

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
```diff
@@ -105,7 +106,18 @@ internal class AmazonBilling(
 
     private var connected = false
 
-    override fun startConnection() {
+    override fun startConnection(delayMilliseconds: Long) {
+        // Start connection has to be called on onCreate, otherwise Amazon fails to detect the foregrounded Activity
+        // Be careful with doing mainHandler.post since that will not guarantee that it's called in onCreate
+        // runOnUIThread checks if the function is called in the UI thread and doesn't do post, so we are good since
+        // startConnection is called on the main thread
```
rickvdl, line 112: This is probably still true, but it's no longer in the method signature (`onMainThread`), could we add an assertion in this method or similar?

purchases/src/test/java/com/revenuecat/purchases/simulatedstore/SimulatedStoreBillingWrapperTest.kt
```diff
@@ -83,34 +83,34 @@ class SimulatedStoreBillingWrapperTest {
...
-        
-        testStoreBilling.startConnection()
+
+        testStoreBilling.performStartConnection()
         assertThat(testStoreBilling.isConnected()).isTrue()
-        
+
         testStoreBilling.close()
         assertThat(testStoreBilling.isConnected()).isFalse()
     }
```
rickvdl, line 116: Are we able to reproduce the ANR in a test by any chance? If so we could verify the fix as well. But I assume it's tricky..

rickvdl (commented):
Nice work 🙌 Few small questions

purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt
tonidero: Yup, seeing the comment from cursor, I'm now actually handling the disconnection right now, so we stop the thread when disconnecting. Regarding having multiple `HandlerThread` with the same name. That's allowed, since the name is mostly meant as a debugging tool, but each thread created is actually unique.

vegaro: Can we inject this as a dependency? Maybe it's not worth it? I was thinking it would be helpful for tests, maybe we can test we are quitting it (when adding Cursor's suggestion https://github.com/RevenueCat/purchases-android/pull/3369/files#r3115790059)

purchases/src/test/java/com/revenuecat/purchases/simulatedstore/SimulatedStoreBillingWrapperTest.kt
tonidero: In a test it's indeed tricky I'm afraid... But I added some tests to make sure we keep this in the main thread 🙏 Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
tonidero: Actually, it doesn't matter which thread this is called on for Amazon, since we do runOnUIThread, to actually start the connection. So we should be safe here. I think this comment mostly refers to the need of wrapping that connection in that function. So I don't think we need any assertion here. Lmk if you think otherwise!

purchases/src/main/kotlin/com/revenuecat/purchases/BillingFactory.kt
rickvdl: Nice!

tonidero: I did something like that in the follow-up commits, including adding some tests. However, what we inject is a `Handler` so it's easier to test. Otherwise, if we inject a `HandlerThread` we would need to mock the internal behavior anyway... Lmk what you think! 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -124,7 +137,6 @@ internal class BillingWrapper(
         private val context: Context,
         private val pendingTransactionsForPrepaidPlansEnabled: Boolean,
     ) {
-        @UiThread
```
vegaro, line 127: why did we have this here? do you know?

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -152,14 +164,14 @@ internal class BillingWrapper(
...
+    override fun startConnection(delayMilliseconds: Long) {
+        backgroundHandler.postDelayed(
+            { performStartConnection() },
             delayMilliseconds,
         )
     }
 
-    override fun startConnection() {
+    private fun performStartConnection() {
         synchronized(this@BillingWrapper) {
```
vegaro: I think this can still lock the main thread since the lock is being held during the whole function, and the main thread can be trying to acquire it as well.

I think we can close the synchronized block by doing somehting like this:

```
private fun performStartConnection() {
    val clientToStart = synchronized(this@BillingWrapper) {
        if (billingClient == null) billingClient = clientFactory.buildClient(this)
        reconnectionAlreadyScheduled = false
        billingClient?.takeIf { !it.isReady }
    } ?: return

    try {
        clientToStart.startConnection(this)
    } catch (e: IllegalStateException) {
        ...
    } catch (e: SecurityException) {
        ...
    }
}
```

vegaro: we can do the same in `sendErrorsToAllPendingRequests` so it doesn't hold the lock for that long

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -163,13 +164,14 @@ class BillingWrapperTest {
         wrapper = BillingWrapper(
             mockClientFactory,
             handler,
+            handler,
```
vegaro: do you mind adding named parameters here 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Good question. TBH, don't know. But none of the code we're executing here requires main thread. It's actually annotated with `AnyThread`.

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
tonidero: Great catch! I think what you mentioned for the `performStartOperation` makes sense. However, I'm not sure about the change for `sendErrorsToAllPendingRequests`... Seems that should be quick enough, since it's just posting some work to be executed in the main thread, but not actually executing anything long I think?

tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3369/commits/cdfb09b29d0d0022b2995f307dda8d8352c5c540

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3369/commits/cc2462ad6618df2db33d85a437f8874e63c1a8ba

vegaro (approved):
Love the changes! Very confident this will fix a lot of the ANRs

purchases/src/main/kotlin/com/revenuecat/purchases/amazon/AmazonBilling.kt
```diff
@@ -105,7 +106,18 @@ internal class AmazonBilling(
...
+        runOnUIThread {
+            performStartConnection()
+        }
+    }
+
+    @VisibleForTesting
+    internal fun performStartConnection() {
```
JayShortway: Does the annotation do anything without an `otherwise`? Also, `internal` visibility is already visible from tests right? Same in `BillingWrapper` below.

tonidero: > Does the annotation do anything without an otherwise?

The default is `PRIVATE`, so this means trying to access it anywhere in production code would give a warning.
image

I do agree that it's not a great practice to make things visible for tests only, since we should only test the public API... But it helps to simplify the writing tests sometimes (though it's definitely a code smell). 

Having said that, I think this one in AmazonBilling can actually be private... So I will change that. 

Regarding the one in BillingWrapper, it helps to make sure we close the thread when closing the purchases instance... We can probably refactor things to be able to test this by passing the `HandlerThread` as a parameter to the class... but then, we would need to use a real one in tests, which might make them more brittle, so I'm considering leaving that as is.

tonidero: Removed this instance in https://github.com/RevenueCat/purchases-android/pull/3369/commits/960353f89c160dbd29f1beb08f053fc43ac88531

PR #3376, opened 2026-04-23
Fix `WorkflowTriggerAction.stepId` JSON field name

vegaro:
## Summary
- The backend serializes the target step id for a workflow trigger action under the `value` key, not `step_id`.
- Deserialization was silently failing and leaving `stepId` null, which blocks any multipage workflow navigation.

## Test plan
- [ ] Unit tests pass: `./gradlew :purchases:testDefaultsBc8DebugUnitTest`
- [ ] Manually verified a workflow response parses with a non-null `stepId`


vegaro:
Closing — verified with khepri backend (`StepTriggerAction.step_id` in `services/workflows/domain/workflow_graph.py`) that the correct JSON key is `step_id`. The branch change to `value` was a bug, not a fix.

PR #3379, opened 2026-04-23
Add `WorkflowNavigator` for multipage workflow step navigation

vegaro:
## Summary
- Introduces a standalone `WorkflowNavigator` class that tracks the current step in a `PublishedWorkflow` and supports forward navigation via component triggers + back navigation via a back stack.
- Not wired into any UI. A follow-up PR will consume this from `PaywallViewModel` to drive multipage paywall navigation.

---

> [!NOTE]
> **Low Risk**
> Low risk: new, internal navigation helper and tests only; no existing UI flow is modified, though future integration will rely on correct trigger/step ID handling.
> 
> **Overview**
> Introduces `WorkflowNavigator`, an internal helper that tracks the current `PublishedWorkflow` step via a `StateFlow`, advances to the next step when a component trigger maps to a `step` action, and maintains a back stack for `navigateBack`/`canNavigateBack`.
> 
> Adds comprehensive tests validating forward navigation, back navigation behavior, and non-navigation on invalid component/action/step references.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 22c1559d2ee3a89a3587ad8dcef5a003107c6751. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/workflow/WorkflowNavigator.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/workflow/WorkflowNavigator.kt
@@ -1,8 +1,5 @@
-@file:OptIn(InternalRevenueCatAPI::class)
-
 package com.revenuecat.purchases.ui.revenuecatui.workflow
 
-import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.common.workflows.PublishedWorkflow
 import com.revenuecat.purchases.common.workflows.WorkflowStep
 import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/workflow/WorkflowNavigator.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/workflow/WorkflowNavigator.kt
@@ -51,5 +51,6 @@ internal class WorkflowNavigator(private val workflow: PublishedWorkflow) {
         return workflow.steps[prevStepId]
     }
 
-    fun canNavigateBack(): Boolean = backStack.isNotEmpty()
+    val canNavigateBack: Boolean
+        get() = backStack.isNotEmpty()
 }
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/workflow/WorkflowNavigator.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    fun navigateBack(): WorkflowStep? {
+        if (backStack.isEmpty()) return null
+        val prevStepId = backStack.removeLast()
+        _currentStepId.value = prevStepId
+        return workflow.steps[prevStepId]
+    }
+
+    fun canNavigateBack(): Boolean = backStack.isNotEmpty()
```
facumenzella: curious, why `fun` and not val?

facumenzella (commented):
Comparing with the iOS equivalent ([purchases-ios#6680](https://github.com/RevenueCat/purchases-ios/pull/6680)), there are three test cases present on iOS that are missing here:

- **Multi-step back navigation** — iOS has [`testMultipleForwardAndBackNavigationsWorkCorrectly`](https://github.com/RevenueCat/purchases-ios/pull/6680/files#diff-a4b0e89e37b2f2f79f9bb2bccfac413fa7c57f91b02f3c81b01ce1a30a697ad6R195) which does `step1→step2→step3→back→step2→back→step1`. The Android tests only go back one step, so the `ArrayDeque` isn't fully exercised as a stack.

- **Trigger with a nil/missing `actionId`** — iOS has [`testTriggerActionWithMissingActionIdReturnsNil`](https://github.com/RevenueCat/purchases-ios/pull/6680/files#diff-a4b0e89e37b2f2f79f9bb2bccfac413fa7c57f91b02f3c81b01ce1a30a697ad6R111) for the case where a trigger matches by `componentId` but has no `actionId`. Not covered here.

- **Target step not in `workflow.steps`** — iOS has [`testTriggerActionWithTargetStepNotInWorkflowReturnsNil`](https://github.com/RevenueCat/purchases-ios/pull/6680/files#diff-a4b0e89e37b2f2f79f9bb2bccfac413fa7c57f91b02f3c81b01ce1a30a697ad6R139) for when `triggerAction.stepId` points to a step that doesn't exist in the workflow. The Android code handles this (the `workflow.steps[stepId] ?: run { ... }` guard), but there's no test for it.

None of these are blocking, but worth adding for parity.

facumenzella (approved):
solid

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/workflow/WorkflowNavigator.kt
vegaro: changing to val to match CustomerCenterNavigationState

PR #3377, opened 2026-04-23
Add `id` field to `ButtonComponent`

vegaro:
## Summary
- Adds an optional `id` field to `ButtonComponent` that mirrors the `id` key the backend serializes for button components in `components_config` ([khepri `Button.id`](https://github.com/RevenueCat/khepri/blob/main/khepri/services/paywalls/domain/template_schemas/v2/components/button.py)).
- This id is how a paywall button is referenced from a workflow step's `triggers[*].component_id`, and will be used in follow-up PRs to dispatch multipage workflow navigation when a button is pressed.
- No behavior change — the field is parsed but not consumed yet.

## Test plan
- [x] New deserialization test covers `ButtonComponent` JSON with an `id` field
- [x] `./gradlew :purchases:testDefaultsBc8DebugUnitTest --tests "com.revenuecat.purchases.paywalls.components.ButtonComponentTests"`
- [x] `./scripts/api-check.sh` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Low risk: adds a new nullable field to `ButtonComponent` and a deserialization test, with no behavior changes beyond capturing additional backend JSON.
> 
> **Overview**
> Paywall `ButtonComponent` now includes an optional `id` field so backend-provided button identifiers are preserved during serialization/deserialization.
> 
> Tests add coverage to ensure `ButtonComponent` JSON containing `id` is decoded correctly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d277380e09b91defb1dd246b8195f75a1a219e00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3380, opened 2026-04-23
Plumb `componentId` through buttons on workflow interactions

vegaro:
> [!NOTE]
> Stacked on top of #3377 (`ButtonComponent.id`). Base will update to `main` once that PR merges.

## Summary
- Plumbs `componentId` through `ButtonComponentStyle`, populated from `ButtonComponent.id` in `StyleFactory`.
- Tracks a `"workflow_trigger"` value as the component interaction event.
- `InternalPaywall`'s action handler currently just logs. The actual dispatch to `PaywallViewModel.handleWorkflowAction` lands in the multipage-workflow wiring PR that depends on this one

---

> [!NOTE]
> **Medium Risk**
> Touches paywall button action mapping and analytics tracking paths; misconfigured buttons (missing `componentId`) now fall back to `NavigateBack`, so behavior changes should be validated on real paywalls.
> 
> **Overview**
> Adds explicit **workflow trigger** support for V2 paywall buttons by renaming the serialized `ButtonComponent.Action.Workflow` to `WorkflowTrigger` while keeping the backend JSON `"type": "workflow"` mapping.
> 
> Plumbs the button `id` into `ButtonComponentStyle.componentId` via `StyleFactory`, emits a new `PaywallAction.External.WorkflowTrigger(componentId)`, and tracks a `component_value` of `"workflow_trigger"` for component-interaction analytics. `InternalPaywall` now recognizes this action (currently logs only), and tests are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 186f654f10b3aefe96f63051c2b24997e233bc62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/ButtonComponent.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/ButtonComponent.kt
@@ -46,7 +46,7 @@ public class ButtonComponent(
         public object NavigateBack : Action
 
         @Serializable
-        public object Workflow : Action
+        public object WorkflowTrigger : Action
 
         @Serializable
         @Immutable
@@ -148,13 +148,13 @@ private class ActionSurrogate(
             is Action.NavigateBack -> ActionTypeSurrogate.navigate_back
             is Action.NavigateTo -> ActionTypeSurrogate.navigate_to
             is Action.RestorePurchases -> ActionTypeSurrogate.restore_purchases
-            is Action.Workflow -> ActionTypeSurrogate.workflow
+            is Action.WorkflowTrigger -> ActionTypeSurrogate.workflow
         },
         destination = when (action) {
             is Action.Unknown,
             is Action.NavigateBack,
             is Action.RestorePurchases,
-            is Action.Workflow,
+            is Action.WorkflowTrigger,
             -> null
 
             is Action.NavigateTo -> when (action.destination) {
@@ -170,7 +170,7 @@ private class ActionSurrogate(
             is Action.Unknown,
             is Action.NavigateBack,
             is Action.RestorePurchases,
-            is Action.Workflow,
+            is Action.WorkflowTrigger,
             -> null
 
             is Action.NavigateTo -> when (action.destination) {
@@ -198,7 +198,7 @@ private class ActionSurrogate(
             is Action.Unknown,
             is Action.NavigateBack,
             is Action.RestorePurchases,
-            is Action.Workflow,
+            is Action.WorkflowTrigger,
             -> null
 
             is Action.NavigateTo -> when (action.destination) {
@@ -218,7 +218,7 @@ private class ActionSurrogate(
             ActionTypeSurrogate.unknown -> Action.Unknown
             ActionTypeSurrogate.restore_purchases -> Action.RestorePurchases
             ActionTypeSurrogate.navigate_back -> Action.NavigateBack
-            ActionTypeSurrogate.workflow -> Action.Workflow
+            ActionTypeSurrogate.workflow -> Action.WorkflowTrigger
             ActionTypeSurrogate.navigate_to -> Action.NavigateTo(
                 destination = when (destination) {
                     DestinationSurrogate.customer_center -> Destination.CustomerCenter
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -361,6 +361,8 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
                 }
 
                 is PaywallAction.External.NavigateBack -> viewModel.closePaywall()
+                is PaywallAction.External.Workflow ->
+                    Logger.d("Workflow received for componentId=${action.componentId}")
                 is PaywallAction.External.NavigateTo -> when (val destination = action.destination) {
                     is PaywallAction.External.NavigateTo.Destination.CustomerCenter ->
                         Logger.w("Customer Center is not yet implemented on Android.")
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
@@ -361,7 +361,7 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
                 }
 
                 is PaywallAction.External.NavigateBack -> viewModel.closePaywall()
-                is PaywallAction.External.Workflow ->
+                is PaywallAction.External.WorkflowTrigger ->
                     Logger.d("Workflow received for componentId=${action.componentId}")
                 is PaywallAction.External.NavigateTo -> when (val destination = action.destination) {
                     is PaywallAction.External.NavigateTo.Destination.CustomerCenter ->
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
@@ -26,7 +26,7 @@ internal sealed interface PaywallAction {
 
         object RestorePurchases : External
         object NavigateBack : External
-        data class Workflow(val componentId: String) : External
+        data class WorkflowTrigger(val componentId: String) : External
         data class PurchasePackage(
             val rcPackage: Package?,
             val resolvedOffer: ResolvedOffer? = null,
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
@@ -26,6 +26,7 @@ internal sealed interface PaywallAction {
 
         object RestorePurchases : External
         object NavigateBack : External
+        data class Workflow(val componentId: String) : External
         data class PurchasePackage(
             val rcPackage: Package?,
             val resolvedOffer: ResolvedOffer? = null,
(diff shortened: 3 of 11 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
```diff
@@ -26,6 +26,7 @@ internal sealed interface PaywallAction {
 
         object RestorePurchases : External
         object NavigateBack : External
+        data class Workflow(val componentId: String) : External
```
tonidero: TBH, having `Workflow` as an "Action" feels weird 😅. Maybe `ContinueWorkflow` or `WorkflowNextStep`? I'm also wondering if it should be a `Destination` instead of an action itself... But TBH, I don't have all the context, so just some thoughts from looking at this code.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentInteraction.kt
```diff
@@ -40,6 +41,13 @@ private fun ButtonComponentStyle.Action.NavigateTo.Destination.componentInteract
             ButtonComponentInteraction(value = "navigate_to_sheet")
     }
 
+internal fun PaywallAction.workflowInteraction(): ButtonComponentInteraction? =
+    if (this is PaywallAction.External.Workflow) {
+        ButtonComponentInteraction(value = "workflow")
```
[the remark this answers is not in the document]
tonidero: this does have a point I think

tonidero (commented):
I'm lacking some context so just left some questions. Code looks good in general though, but thinking about leaving to others to approve.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
vegaro: right, the thing is that this can be anything in the workflow, for now it's linear navigation, but in the future it could be other stuff. This is the flow:

Button has action type workflow:

```
{
    "action": {
        "type": "workflow"
    },
    "id": "4xeP-v4go0",
    "name": "Continue",
    "stack": {
    },
    "type": "button"
} 
```

When pressed, we check in that step's triggers what the trigger of the component should do by `id = component_id` and check in the `trigger_actions` by `action_id`:

```
"steps": {
        "BmwFBdp": {
            "id": "BmwFBdp",
            "metadata": null,
            "outputs": {},
            "param_values": {
                "is_last_step": false
            },
            "screen_id": "pw7553a2d8d98a4d22",
            "trigger_actions": {
                "btn_4xeP-v4go0": {
                    "step_id": "RjPCgjH",
                    "type": "step"
                }
            },
            "triggers": [
                {
                    "action_id": "btn_4xeP-v4go0",
                    "component_id": "4xeP-v4go0",
                    "name": "Continue",
                    "type": "on_press"
                }
            ],
            "type": "screen"
        },
        "RjPCgjH": {
            "id": "RjPCgjH",
            "metadata": null,
```

I considered `WorkflowTrigger`, `WorkflowTriggerAction`, but then ended up with just `Workflow` to match the JSON...

I do agree it's confusing and went arouond a bit. I think the one I preffer is `WorkflowTrigger`. What do you think?

vegaro: To add more context... `trigger_actions` can be:

 1. step -Navigate directly to a specific step.                                                                                                          
`{ type: 'step', step_id: string }`
  2. conditions — Evaluate JSON Logic conditions (using the jsonLogic library) to decide the next step dynamically.                                        
 `{ type: 'conditions', conditions: Record<string, unknown> }`                                                                                            
  
  so a press of a button can eventually trigger some conditions. That's why it's not really a destination

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentInteraction.kt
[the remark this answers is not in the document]
facumenzella: agree

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
tonidero: Ahh ok makes sense... I think `WorkflowTrigger` makes sense... Could also be `WorkflowAction` in a sense of a "generic" workflow action. But happy with trigger as well.

vegaro: just noticed we also have it in purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/ButtonComponent.kt

which got merged in https://github.com/RevenueCat/purchases-android/pull/3385/, would you rename there as well? or keep it as `workflow` since that's what the backend sends. Happy to rename that, it's public but just got merged 😄

tonidero: Oh indeed... Hmm I think it might be better to have better name, otherwise `Workflow` seems very generic and probably difficult to find. It's true that in this case, it's encapsulated within a class, so not a huge deal though... I would say to rename it if possible, but not a huge deal!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentInteraction.kt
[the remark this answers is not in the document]
vegaro: yeah the description is outdated, and I am defining that now. I am going to set it to "workflow_trigger". @MonikaMateska since you worked on these `ButtonComponentInteraction`. Does this `value = workflow_trigger` make sense? Do we need to define these in the backend?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/PaywallAction.kt
vegaro: renamed everything to `WorkflowTrigger`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -361,6 +361,8 @@ private fun rememberPaywallActionHandler(viewModel: PaywallViewModel): suspend (
                 }
 
                 is PaywallAction.External.NavigateBack -> viewModel.closePaywall()
+                is PaywallAction.External.Workflow ->
+                    Logger.d("Workflow received for componentId=${action.componentId}")
```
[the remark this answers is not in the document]
vegaro: This is ok for now, will be wired in the future. No WorfklowTriggers in any paywall yet anyways 😄

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/ButtonComponent.kt
```diff
@@ -46,7 +46,7 @@ public class ButtonComponent(
         public object NavigateBack : Action
 
         @Serializable
-        public object Workflow : Action
+        public object WorkflowTrigger : Action
```
facumenzella, line 49: like it

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentInteraction.kt
```diff
@@ -40,6 +41,13 @@ private fun ButtonComponentStyle.Action.NavigateTo.Destination.componentInteract
             ButtonComponentInteraction(value = "navigate_to_sheet")
     }
 
+internal fun PaywallAction.workflowInteraction(): ButtonComponentInteraction? =
+    if (this is PaywallAction.External.WorkflowTrigger) {
+        ButtonComponentInteraction(value = "workflow_trigger")
+    } else {
+        null
+    }
```
facumenzella: let's move it to a `PaywallActionExtensions.kt` alongside `navigationUrlForComponentInteraction()`, or pull it inline into `ButtonComponentView` where it's used.

facumenzella (commented):
The PR description had a mismatch with the implementation: the Summary bullet said the interaction event was `"trigger_workflow_action"`, but the code emits `"workflow_trigger"` (matching the Cursor-generated section below the fold). Updated the description to say `"workflow_trigger"` so it stays in sync with the code.

facumenzella (approved):
looks good to me

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/button/ButtonComponentState.kt
```diff
@@ -110,9 +116,9 @@ internal class ButtonComponentState(
                     ),
                 )
             }
-            // Workflow buttons are normally intercepted before toPaywallAction() is called.
-            // Reaching this is unexpected and indicates a misconfigured button (e.g. null componentId).
-            is ButtonComponentStyle.Action.Workflow -> {
+            // Should not reach here: Action.WorkflowTrigger is handled before calling toPaywallAction.
+            // Reached only if componentId is null (misconfigured button).
+            is ButtonComponentStyle.Action.WorkflowTrigger -> {
                 Logger.e("ButtonComponentState: reached toPaywallAction for Workflow action — componentId may be null")
```
facumenzella, line 122: What do you think about adding something like 
`if (BuildConfig.DEBUG) error("WorkflowTrigger reached toPaywallAction — componentId is null")`

tonidero: Note that `BuildConfig.DEBUG` will only work for us, not on developer's apps. If we wanted to do that, we could use `appConfig.isDebugBuild`, but maybe we want to keep that behavior internal...

vegaro: I think I can get rid of this. It's in draft but https://github.com/RevenueCat/purchases-android/pull/3381/files#diff-da3cac32a5cf63c0e97c308a03655c8d66360bd41c88d3aa05e8ae68c434a23dR58 here's the idea

PR #3386, opened 2026-04-24
Create `WorkflowTriggerType` and `WorkflowTriggerActionType`

vegaro:
- Added `WorkflowTriggerActionType` and `WorkflowTriggerType`. Even though they only have one type now, it makes things cleaner. 
- Added the type to `triggerAction` so the Button can be the one passing the trigger type to the navigator and allow for different types of triggers. For now it's "on_press", but we could add other types, imagine "on_long_press"

---

> [!NOTE]
> **Medium Risk**
> Changes workflow JSON deserialization shapes and the `triggerAction` API, which could affect navigation if callers or backend payloads don’t match the new typed expectations. Risk is mitigated by explicit `UNKNOWN` fallbacks and added unit tests for forward-compatibility.
> 
> **Overview**
> **Workflow triggers/actions are now strongly typed and forward-compatible.** `WorkflowTrigger.type` is changed from a `String` to a `WorkflowTriggerType` enum (currently `ON_PRESS` + `UNKNOWN`) with a defaulting deserializer for unknown values.
> 
> `WorkflowTriggerAction` is refactored from a plain data class into a sealed type (`Step` + `Unknown`) using a deserializer that safely falls back when the JSON is missing/invalid or the action type is unrecognized. `WorkflowNavigator.triggerAction` now requires a `triggerType` and only navigates when the resolved action is a `Step`; tests were updated and new deserialization tests were added to lock in the unknown-value behavior (plus a small assertion-style cleanup in `FlexDistributionTests`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 710e4a74d2f272e14cbdbc197c545670f1cbed87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -17,18 +17,32 @@ import kotlinx.serialization.json.JsonElement
...
+@InternalRevenueCatAPI
+@Serializable
+public enum class WorkflowTriggerActionType {
+    @SerialName("step")
+    STEP,
+}
+
 @InternalRevenueCatAPI
 @Serializable
 public data class WorkflowTriggerAction(
```
tonidero: Considering this is internal, I wonder if we should represent this as a sealed class, since `stepId` should probably only exist when action type is `step`... Maybe the same thing could be applied to `WorkflowTrigger`, not sure?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -17,18 +17,32 @@ import kotlinx.serialization.json.JsonElement
...
+    @SerialName("on_press")
+    ON_PRESS,
+}
+
+@InternalRevenueCatAPI
+@Serializable
+public enum class WorkflowTriggerActionType {
+    @SerialName("step")
+    STEP,
+}
```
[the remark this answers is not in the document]
vegaro: Thank you Cursor

tonidero (commented):
Just a suggestion in the modeling, considering this is mostly internal. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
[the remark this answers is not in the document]
tonidero: Oh right, good one 😅

purchases/src/test/java/com/revenuecat/purchases/common/workflows/WorkflowEnumDeserializationTests.kt
```diff
@@ -0,0 +1,55 @@
...
+            arrayOf("\"on_press\"", WorkflowTriggerType.ON_PRESS),
+            arrayOf("\"on_long_press\"", WorkflowTriggerType.UNKNOWN),
+            arrayOf("\"some_future_unknown_trigger_type\"", WorkflowTriggerType.UNKNOWN),
+        )
+    }
+
+    @Test
+    fun `Should properly deserialize WorkflowTriggerType`() {
+        val actual = JsonTools.json.decodeFromString<WorkflowTriggerType>(serialized)
+        assert(actual == expected)
```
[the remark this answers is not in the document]
vegaro: yeah thanks!!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
vegaro: I like it, I need to change the deserializer but I think it's worth it

vegaro: For `WorkflowTrigger`, every trigger type carries the same fields (`name`, `actionId`, `componentId`), so a sealed class would look like:

```
sealed class WorkflowTrigger {
    data class OnPress(val name: String, val actionId: String, val componentId: String) : WorkflowTrigger()
    object Unknown : WorkflowTrigger()
}
```

The only win is that `Unknown` lives in the hierarchy rather than as an enum value. But no nullability removed in this case, so not sure if it's worth it?

tonidero: I guess it indeed depends on whether we think other parameters might be added in the future that are specific to only one type... if not, it can indeed be kept as is IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/PolymorphicSerializerWithDefault.kt
```diff
@@ -0,0 +1,48 @@
...
+ * a subclass whose fields couldn't be parsed — indicated by [selectByType] returning `null`.
+ *
+ * For the common case where each type maps to a single required field, pass a [serializers] map
+ * of `type -> (requiredField to serializer)` instead of overriding [selectByType].
+ */
+internal abstract class PolymorphicSerializerWithDefault<T : Any>(
+    baseClass: KClass<T>,
+    private val unknownSerializer: DeserializationStrategy<T>,
+    private val typeField: String = "type",
+    private val serializers: Map<String, Pair<String, DeserializationStrategy<T>>> = emptyMap(),
```
tonidero: I see we have the `selectByType` available for this cases, but I'm wondering if we should use something like `Map<String, Pair<List<String>, DeserializationStrategy<T>>>` so we support multiple required parameters without having to override the other method?

I wonder if we should just not check required parameters at all, and just have the serializer try to perform the deserialization, and add a try-catch over serialization exceptions, so if it fails, we fallback to the unknownSerializer? This would be much simpler IMO... 

I'm not sure if we want to have more customizable serializers in follow-up PRs that would explain this though, so lmk in that case :)

tonidero (commented):
Looking good! Just one question on the new serializer utility, in case it can be simplified.

purchases/src/main/kotlin/com/revenuecat/purchases/utils/serializers/PolymorphicSerializerWithDefault.kt
vegaro: I might be overengineering this tbh. Went a bit in circles trying to make it generic so we could reuse it in the future but all versions are a bit meh. Are you fine with a custom serializer for `WorkflowTriggerAction`? that will be cleaner imho

vegaro: I've removed this in b0462e0c1618dc7e6ccec4d08c7f691955139748

Wdyt? I think it's more readable and feels what we should be doing for now

tonidero: Right... I think this is something that could be useful down the line... but until we have other cases where this is needed, I'm ok keeping it focused to what we need right now.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -8,27 +8,65 @@ import com.revenuecat.purchases.paywalls.components.common.ComponentsConfig
...
+
+internal object WorkflowTriggerActionSerializer : JsonContentPolymorphicSerializer<WorkflowTriggerAction>(
+    WorkflowTriggerAction::class,
+) {
+    override fun selectDeserializer(element: JsonElement): DeserializationStrategy<WorkflowTriggerAction> {
+        val obj = element.jsonObject
+        val type = (obj["type"] as? JsonPrimitive)?.contentOrNull
+        val stepId = (obj["step_id"] as? JsonPrimitive)?.contentOrNull
+        return if (type == "step" && stepId != null) {
```
tonidero: I guess this only checks for the existence of `stepId` but doesn't check for example that it's of the correct type (like, it could be a boolean and it would fail to parse it later)... I wonder if we should override the `deserialize method directly to just catch any errors deserializing... but well, TBH, this shouldn't really happen ever, so I won't block this PR because of that.

tonidero (approved):
Looks great!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
vegaro: not a bad idea to handle that. Added it in [bad69f4](https://github.com/RevenueCat/purchases-android/pull/3386/commits/bad69f40a90585d9eca4eb74604917ab7a70e018)

vegaro: lol, can you believe we already had a `SealedDeserializerWithDefault` 😮‍💨

tonidero: Ohhh I had totally forgotten about that 😅. Oh well, that's good then 🙌

PR #3395, opened 2026-04-28
Apply ripple shape clip on a sibling Box to avoid clipping content

tonidero:
## Summary
- Follow-up to #3324, which fixed the original ripple-vs-margin issue (#3321) by adding `.clip(composeShape)` on the click target so the ripple respects rounded corners. That fix uses a graphics-layer clip, which also crops descendants — stacks with nested children that intentionally extend outside the parent (badges with offsets, drop shadows, decorative elements) get visibly clipped during press.
- Introduces declarative `onStackClick: (() -> Unit)?`, `enabled: Boolean`, `interactionSource: MutableInteractionSource?` parameters on `StackComponentView`, threaded through `StackWithOverlaidBadge`, `StackWithLongEdgeToEdgeBadge`, `StackWithShortEdgeToEdgeBadge`, and `MainStackComponent`. This lets `MainStackComponent` split the click gesture from the indication.
- **Plain stack path:** wraps in a `Box {}` containing the stack with `clickable(indication = null)` (gesture only, no clip on the stack) and a sibling `Box(Modifier.matchParentSize().padding(margin).clip(composeShape).indication(source, LocalIndication.current))` that draws the shape-clipped ripple. Children render unclipped.
- **Other render paths** (video background, nested badge, overlay) keep their existing `.clip(composeShape)` on the click target — only the click wiring is swapped over to the new `onStackClick` parameter. The overflow-clip fix therefore only applies to the plain render path; this limitation is documented in the `onStackClick` kdoc so future callers know what to expect.
- Updates the three call sites (`ButtonComponentView`, `PackageComponentView`, `TabControlButtonView`) to use the new declarative API.
- Adds a Compose preview (`StackComponentView_Preview_Clickable_With_Overflowing_Child_Shadow`) as a regression guard for the overflowing-child shadow case.

## Test plan
- [x] `./gradlew :ui:revenuecatui:testDefaultsBc8DebugUnitTest` (full suite + Paparazzi `verify`) — green
- [x] `./gradlew detektAll` — green
- [x] Manual: build a paywall whose plain stack has a nested child positioned to extend outside the parent (offset, negative padding, or drop shadow); tap the parent — overflowing child must remain fully visible during the press, and the ripple must follow the parent's rounded corners

---

> [!NOTE]
> **Medium Risk**
> Touches core Compose layout/click/semantics for paywall components, so regressions could affect tap handling, accessibility, or visual clipping across many paywalls despite being UI-only.
> 
> **Overview**
> Fixes clickable `StackComponentView` ripple rendering so the ripple remains clipped to the stack shape **without clipping overflowing children** (e.g. shadows/badges) by splitting gesture handling and indication onto separate sibling nodes.
> 
> Introduces a new declarative click API on `StackComponentView` (`onStackClick`, `enabled`, `interactionSource`) and updates existing button/tab/package components to use it instead of applying `Modifier.clickable` externally; adds Compose previews as regression guards. The paywall-tester sample paywall is also updated to use a `ButtonComponent` for its CTA.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2471d578a4b5046066755d973d8585d7f0cc091a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -695,9 +774,11 @@ private fun MainStackComponent(
                     .size(stackState.size)
                     .then(outerShapeModifier)
                     .clip(composeShape)
+                    .then(clickModifier)
                     .then(borderModifier),
             ) {
                 stack(
+                    modifier,
```
vegaro, line 781: Should this be `Modifier`? the `modifier` is already being applied to `WithOptionalBackgroundOverlay`. Like 813 has a similar case and applies `Modifier`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -726,10 +848,11 @@ private fun MainStackComponent(
             modifier = modifier
                 .then(outerShapeModifier)
                 .clip(composeShape)
+                .then(clickModifier)
                 .then(borderModifier),
         ) {
             WithOptionalBackgroundOverlay(state, background = backgroundStyle) {
-                stack(Modifier.then(innerShapeModifier))
+                stack(modifier, Modifier.then(innerShapeModifier))
```
vegaro, line 855: Same `Modifier` question.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
```diff
@@ -745,10 +868,11 @@ private fun MainStackComponent(
         Box(
             modifier = modifier
                 .then(outerShapeModifier)
-                .clip(composeShape),
+                .clip(composeShape)
+                .then(clickModifier),
         ) {
             WithOptionalBackgroundOverlay(state, background = backgroundStyle) {
-                stack(borderModifier.then(innerShapeModifier))
+                stack(modifier, borderModifier.then(innerShapeModifier))
```
vegaro, line 875: and here as well

vegaro (commented):
I tested it and it works great. Just one question on `modifier` vs `Modifier`

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
tonidero: So... I think you're right, however, I think we were doing that before the changes in this PR... But not sure why 😕. Going to test that change anyway with Emerge snapshots and see if there are any changes and think in case there might be any unintended consequences.

tonidero: I can confirm before we were passing it again and applying the modifier twice 🥴 So... I guess now we need to decide whether to apply the change in behavior or not... If we want to, I would suggest doing it on a different PR. For context, I added some new previews to track this. This is how those looks before this PR:
image

And this is how it looks if I change this to not apply the modifier twice:
image

cc @RevenueCat/paywallsengine in case we want to fix this. For now, I'm planning to leave as is in this PR. Lmk what you think!!

vegaro: oh damn, I wasn't expecting it to affect that much!

I would leave it as is, if we were already doing it, and fix it in another moment

vegaro (approved):

PR #3398, opened 2026-04-28
feat(Customer Center): Add Purchase History screen

facumenzella:
## Summary

We want feature parity with iOS, so adding purchase history 👍 

---

> [!NOTE]
> **Medium Risk**
> Touches Customer Center navigation, customer-info loading paths, and purchase modeling; gated by config flag but expands data shown in UI and adds parallel loadAllPurchases work when enabled.
> 
> **Overview**
> Adds **Purchase History** to Android Customer Center when `display_purchase_history_link` is enabled in support config, aligned with iOS.
> 
> A **See all purchases** entry appears on management, no-active, and selected-purchase screens only when purchase counts justify a full list (e.g. active + inactive subs, multiple inactive subs, or more than two one-time purchases). The view model loads **`allPurchases`** from full customer info (not just active/relevant items) when the flag is on, and maps richer fields into **`PurchaseInformation`** and **`TransactionDetails`** (dates, billing issues, sandbox, transaction IDs, stable **`purchaseHistoryEntryId`** for detail navigation after refresh).
> 
> New **`PurchaseHistoryView`** groups active, inactive, and one-time purchases; **`PurchaseHistoryDetailView`** shows a labeled detail sheet. Config adds **`displayPurchaseHistoryLink`** and localization keys for history UI strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67cbfd0dceab20f61738d0f919b88f1ebf85f003. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/navigation/CustomerCenterDestination.kt
```diff
@@ -45,4 +45,16 @@ internal sealed class CustomerCenterDestination {
         val data: CreateSupportTicketData,
         override val title: String,
     ) : CustomerCenterDestination()
+
+    data class PurchaseHistory(
```
vegaro, line 49: Storing these lists in the destination means the data is frozen when the screen starts. When a background refresh runs (e.g. after the app is foregrounded after a cancellation), these lists will be stale, while allPurchases in the new Success state is fresh. The user ends up viewing outdated history until they back out and re-enter.

I believe the fix is to make this an object and have PurchaseHistoryView read from `CustomerCenterState.Success.allPurchases` directly (same pattern as `MainScreenContent` with `state.purchases`). `InternalCustomerCenter` stops reading from `destination.activeSubscriptions` etc. and instead filters from `customerCenterState.allPurchases` inline

purchases/src/main/kotlin/com/revenuecat/purchases/customercenter/CustomerCenterConfigData.kt
```diff
@@ -315,6 +315,78 @@ public data class CustomerCenterConfigData(
...
+            SCREEN_MANAGEMENT_SEE_ALL_PURCHASES,
+
+            @SerialName("screen_purchase_history_title")
+            SCREEN_PURCHASE_HISTORY_TITLE,
+
+            @SerialName("subscriptions_section_title")
+            SUBSCRIPTIONS_SECTION_TITLE,
+
+            @SerialName("inactive")
+            INACTIVE,
```
vegaro, line 329: this is used in two places, as both a section header title (“Inactive” subscriptions) and as a status badge. These are semantically different strings and need separate keys.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/viewmodel/CustomerCenterViewModel.kt
```diff
@@ -751,6 +823,69 @@ internal class CustomerCenterViewModelImpl(
...
+            val entitlement = customerInfo.entitlements.active.values
+                .firstOrNull { it.productIdentifier == transaction.productIdentifier }
+                ?: customerInfo.entitlements.all.values
+                    .firstOrNull { it.productIdentifier == transaction.productIdentifier }
+
+            val isActivePlayStoreSubscription = transaction.store == Store.PLAY_STORE &&
+                transaction is TransactionDetails.Subscription &&
+                transaction.isActive
+            val product = if (isActivePlayStoreSubscription) {
+                purchases.awaitGetProduct(
```
vegaro: this is serial and in a loop...

You have two alternatives, awaiting all using coroutines:

```
return coroutineScope {
    transactions.map { transaction ->
        async {
            val isActivePlayStoreSubscription = transaction.store == Store.PLAY_STORE &&
                transaction is TransactionDetails.Subscription &&
                transaction.isActive
            val product = if (isActivePlayStoreSubscription) {
                purchases.awaitGetProduct(
                    transaction.productIdentifier,
                    (transaction as TransactionDetails.Subscription).productPlanIdentifier,
                )
            } else {
                null
            }
            // build PurchaseInformation...
        }
    }.awaitAll()
}
```

or `PurchasesOrchestrator.getProducts` already accepts a `List<String>` of product IDs and fetches them all in one Play Store query. You could extract all active Play Store subscription IDs upfront, call `getProducts` once, build a map, then map over transactions. This is the most efficient but requires wrapping the callback API in a `suspendCoroutine`.

The number of active Play Store subscriptions per user is almost always 1–2 so the parallel vs batch difference is minimal, so your call.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/views/PurchaseHistoryDetailView.kt
```diff
@@ -0,0 +1,247 @@
...
+    val locale = remember { Locale.getDefault() }
+
+    fun Date.fmt() = dateFormatter.format(this, locale)
+
+    Column(
+        modifier = modifier
+            .fillMaxSize()
+            .verticalScroll(rememberScrollState()),
+    ) {
+        DetailSection(modifier = Modifier.padding(ManagementViewHorizontalPadding)) {
```
vegaro: I think this should be 

```
DetailSection(modifier = Modifier.padding(horizontal = ManagementViewHorizontalPadding))
```

Otherwise you'd have double top padding. `DetailScreen` already has top padding.

vegaro (changes requested):
This PR could use some tests 😄

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -530,6 +537,24 @@ private fun CustomerCenterNavHost(
...
+            is CustomerCenterDestination.PurchaseHistory -> {
+                val allPurchases = customerCenterState.allPurchases
+                PurchaseHistoryView(
+                    activeSubscriptions = allPurchases.filter { it.isSubscription && !it.isExpired },
+                    inactiveSubscriptions = allPurchases.filter { it.isSubscription && it.isExpired },
+                    nonSubscriptions = allPurchases.filter { !it.isSubscription },
+                    localization = customerCenterState.customerCenterConfigData.localization,
+                    onAction = onAction,
+                )
+            }
```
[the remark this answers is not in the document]
facumenzella: Intentional — the two sites serve different purposes (button visibility vs. section rendering), and the filter expressions are simple enough that extracting computed properties would add complexity without preventing any real divergence risk.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/data/CustomerCenterState.kt
```diff
@@ -43,6 +44,17 @@ internal sealed class CustomerCenterState(
...
+        val shouldShowPurchaseHistory: Boolean
+            get() {
+                if (customerCenterConfigData.support.displayPurchaseHistoryLink != true) return false
+                val activeSubscriptions = allPurchases.filter { it.isSubscription && !it.isExpired }
+                val inactiveSubscriptions = allPurchases.filter { it.isSubscription && it.isExpired }
+                val nonSubscriptions = allPurchases.filter { !it.isSubscription }
+                return (activeSubscriptions.isNotEmpty() && inactiveSubscriptions.isNotEmpty()) ||
+                    (activeSubscriptions.isEmpty() && inactiveSubscriptions.size > 1) ||
+                    nonSubscriptions.size > 2
+            }
```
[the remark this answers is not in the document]
facumenzella: By design — mirrors iOS behavior exactly. The per-category thresholds are a product decision, not a code issue.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/customercenter/InternalCustomerCenter.kt
```diff
@@ -530,6 +537,28 @@ private fun CustomerCenterNavHost(
...
+            is CustomerCenterDestination.PurchaseHistoryDetail -> {
+                customerCenterState.allPurchases
+                    .firstOrNull { it.purchaseHistoryEntryId == destination.purchaseHistoryEntryId }
+                    ?.let { purchase ->
+                        PurchaseHistoryDetailView(
+                            purchase = purchase,
+                            localization = customerCenterState.customerCenterConfigData.localization,
+                        )
+                    }
+            }
```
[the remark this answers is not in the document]
facumenzella: Not addressing this — purchases are never dropped from `CustomerInfo` (subscriptions stay in `subscriptionsByProductIdentifier`, non-subscriptions in `nonSubscriptionTransactions`), so the null case is unreachable in practice. If that invariant ever breaks, the blank screen will make the bug obvious.

facumenzella:
Superseded by a 4-PR stack, so each piece fits under the 300-line production churn limit that `revenuecat/danger` now enforces (this PR measured 863 lines, which is why danger started failing here once it was brought up to date with main).

Merge in order:

1. #3997 config + model fields (175 lines)
2. #3998 loading + state + gate (99 lines)
3. #3999 history list screen (284 lines)
4. #4000 detail screen (295 lines)

The stack was carved from this PR's verified merged state, and its tip reconstructs the production code here byte-for-byte in 12 of 13 files. The 13th, `PurchaseHistoryDetailView.kt`, differs only in the expired `@Preview`, which now reuses the existing `purchaseInformationYearlyExpired` fixture instead of rebuilding one inline (this also drops a fully qualified `Store` reference that the repo style rules disallow). All 8 purchase history view model tests from here are carried across the stack, each exactly once, plus 4 new `PurchaseInformation` field-propagation tests so #3997 covers its own layer.

CI on the stack: #3997 fully green, #3998 green with `codecov/project` pending, #3999 and #4000 green except `Snapshot Test | Emerge`, which is red only because Emerge gates newly added snapshots on approval (0 removed, 0 modified, 0 errored on both). Those need an approval click, not a code change.

One thing flagged rather than changed, documented in #4000: iOS gates `store`, `productID`, `sandbox`, `transactionID` and `originalPurchaseDate` behind `#if DEBUG`, and four of those five render inline and unconditionally here. `#if DEBUG` has no 1:1 Android equivalent, since a prebuilt AAR's `BuildConfig.DEBUG` is the library's build type rather than the consuming app's, so matching the intent would need something like `ApplicationInfo.FLAG_DEBUGGABLE`. Worth a product decision.

Correction to an earlier draft of this comment: I had said there was no snapshot coverage for the new screens. That is wrong as stated. There is no Paparazzi coverage in-repo, but Emerge does snapshot the new Compose previews in CI, which is exactly why the checks above fired.

Closing in favour of the stack. Review discussion stays readable here.

PR #3404, opened 2026-04-29
Fix component paywall crash with root vertical overflow

MonikaMateska:
### Checklist

### Motivation
Bug: https://github.com/RevenueCat/react-native-purchases/issues/1742

Some Paywalls V2 layouts crash during rendering when the root stack declares vertical overflow. The crash happens because `LoadedPaywallComponents` always wraps the main content in `verticalScroll`, while the root stack can also apply its own vertical scroll modifier. Compose rejects this nested vertical scroll measurement with infinite height constraints.

### Description
- Avoid wrapping the main content in an outer `verticalScroll` when the root stack already owns vertical scrolling.
- Add a regression test that renders a root stack with `overflow = SCROLL` through `LoadedPaywallComponents`.

---

> [!NOTE]
> **Low Risk**
> Small UI behavior change limited to scroll modifier composition plus a new regression test; main risk is unintended scroll behavior differences in edge cases.
> 
> **Overview**
> Fixes a rendering crash in `LoadedPaywallComponents` by conditionally applying the outer `verticalScroll`: when the root `StackComponentStyle` indicates vertical scrolling (`overflow = SCROLL`), the wrapper scroll modifier is skipped to avoid nested vertical scroll on the same axis.
> 
> Adds a Compose regression test that builds a paywall with a root stack configured for vertical scrolling and verifies it renders without crashing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c58d08008e1c7a770f9826ee2e9e71224d4cb737. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedPaywallComponentsLocaleTests.kt
```diff
@@ -154,5 +154,50 @@ internal class LoadedPaywallComponentsLocaleTests {
             )
 
         }
+
+        @Test
+        fun `Should render root stack with vertical overflow without nested scroll crash`(): Unit = with(composeTestRule) {
```
JayShortway, line 159: ❤️

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedPaywallComponents.kt
```diff
@@ -85,6 +87,9 @@ internal fun LoadedPaywallComponents(
     val headerComponentStyle = state.header
     val footerComponentStyle = state.stickyFooter
     val background = rememberBackgroundStyle(state.background)
+    val mainScrollState = rememberScrollState()
+    val shouldWrapMainContentInVerticalScroll =
```
vegaro, line 93: Thought it would be more complex to solve, so I love it's just checking for this 😄 

I would add a comment for future reference:

```
// If the root stack already scrolls vertically (overflow = SCROLL on a vertical dimension),
// skip the outer verticalScroll — two vertical scroll modifiers on the same axis crashes.
```

vegaro (approved):
Tested it and it looks good 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedPaywallComponents.kt
MonikaMateska: Good point, adding it!

PR #3424, opened 2026-04-30
Add presented offering context to custom paywall events

rickvdl:
### Description

Adds `placement_identifier`, `targeting_revision`, and `targeting_rule_id` to the custom paywall impression event.

Mirrors the event payload work in iOS PR https://github.com/RevenueCat/purchases-ios/pull/6707 for Android. The internal paywall events already received this context via #3253; this PR brings the same placement/targeting attribution to custom paywall events.

### What's new

- `CustomPaywallImpressionParams` gains an `Offering`-based constructor. Passing an `Offering` lets the SDK derive the offering identifier and `PresentedOfferingContext` from the offering's first available package.
- The params object stores `paywallId`, `offeringId`, and `presentedOfferingContext`; it does not store or expose the `Offering` object. This keeps the public params surface focused on the values needed for event tracking.
- The string `offeringId` constructor is deprecated in favor of passing an `Offering`, because an offering identifier alone cannot reliably carry placement and targeting context.
- An `@InternalRevenueCatAPI` constructor accepts `paywallId`, `offeringId`, and `presentedOfferingContext` directly, so hybrid SDKs can pass the context through PHC to native without exposing `Offering` on the public params API.
- `Purchases.trackCustomPaywallImpression(params)` resolves event data with this fallback chain:

  | Developer passes | `offering_id` sent | context sent |
  | --- | --- | --- |
  | Nothing (or only `paywallId`) | from `cachedOfferings.current` | from current offering |
  | Deprecated `offeringId` matching cache | passed string | from cached match |
  | Deprecated `offeringId` not in cache | passed string | nil |
  | `Offering` object | from offering | derived from offering |
  | Internal `presentedOfferingContext` | passed `offeringId` | passed context |

- `CustomPaywallEvent.Impression.Data` carries flat `placementIdentifier` / `targetingRevision` / `targetingRuleId` fields, mirroring the existing internal `PaywallEvent` Android shape.
- The wire payload nests these under `presented_offering_context`. The custom paywall event has its own `BackendEvent.CustomPaywallPresentedOfferingContextData` rather than reusing the internal paywall event's struct, matching the iOS approach where each event owns its own.

---

> [!NOTE]
> **Medium Risk**
> Public API surface changes (new constructors, deprecation) and analytics event schema updates; behavior is localized to custom paywall tracking rather than purchases or auth.
> 
> **Overview**
> Custom paywall impression tracking now sends **placement and targeting context** (placement identifier, targeting revision, rule id) on the backend event, nested as `presented_offering_context`, aligning custom paywalls with internal paywall analytics.
> 
> **`CustomPaywallImpressionParams`** adds constructors that take an **`Offering`** (or offering-only) so the SDK can set `offeringId` and `presentedOfferingContext` from the offering’s first package. The **`offeringId` string constructor is deprecated** in favor of `Offering`. Params expose `presentedOfferingContext` but not the offering object itself.
> 
> **`Purchases.trackCustomPaywallImpression`** resolves what gets tracked: explicit context from params, lookup from **cached offerings** when only an offering id or nothing is passed, or values taken directly from a passed **`Offering`**.
> 
> Wire and storage paths extend **`CustomPaywallEvent.Impression.Data`** and **`BackendEvent.CustomPaywall`** with the new fields; **`cachedOfferings`** is exposed on the orchestrator/offerings manager to support resolution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ea3369c13d4aa1fa9b1904c4ca8d272c0c57c73d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
@@ -38,6 +38,7 @@ import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI;
 
 import org.jetbrains.annotations.NotNull;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
@@ -144,7 +145,7 @@ final class PurchasesAPI {
                 "my-offering",
                 "",
                 new HashMap<>(),
-                new java.util.ArrayList<>()
+                new ArrayList<>()
         );
         CustomPaywallImpressionParams paramsWithOfferingObject =
                 new CustomPaywallImpressionParams("my-paywall", offering);
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
@@ -10,6 +10,7 @@ import com.revenuecat.purchases.AmazonLWAConsentStatus;
 import com.revenuecat.purchases.CacheFetchPolicy;
 import com.revenuecat.purchases.CustomerInfo;
 import com.revenuecat.purchases.EntitlementVerificationMode;
+import com.revenuecat.purchases.Offering;
 import com.revenuecat.purchases.Offerings;
 import com.revenuecat.purchases.Purchases;
 import com.revenuecat.purchases.PurchasesAreCompletedBy;
@@ -139,10 +140,21 @@ final class PurchasesAPI {
         VirtualCurrencies cachedVirtualCurrencies = purchases.getCachedVirtualCurrencies();
 
         // trackCustomPaywallImpression API
+        Offering offering = new Offering(
+                "my-offering",
+                "",
+                new HashMap<>(),
+                new java.util.ArrayList<>()
+        );
+        CustomPaywallImpressionParams paramsWithOfferingObject =
+                new CustomPaywallImpressionParams("my-paywall", offering);
+        Offering offeringObject = paramsWithOfferingObject.getOffering();
         purchases.trackCustomPaywallImpression();
         purchases.trackCustomPaywallImpression(new CustomPaywallImpressionParams());
         purchases.trackCustomPaywallImpression(new CustomPaywallImpressionParams("my-paywall"));
         purchases.trackCustomPaywallImpression(new CustomPaywallImpressionParams("my-paywall", "my-offering"));
+        purchases.trackCustomPaywallImpression(paramsWithOfferingObject);
+        purchases.trackCustomPaywallImpression(new CustomPaywallImpressionParams(offering));
     }
 
     static void checkSyncAmazonPurchase(final Purchases purchases,
--- a/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
+++ b/api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
@@ -12,6 +12,7 @@ import com.revenuecat.purchases.CustomerInfo;
 import com.revenuecat.purchases.EntitlementVerificationMode;
 import com.revenuecat.purchases.Offering;
 import com.revenuecat.purchases.Offerings;
+import com.revenuecat.purchases.PresentedOfferingContext;
 import com.revenuecat.purchases.Purchases;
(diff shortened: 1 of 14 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/CustomPaywallImpressionParams.kt
```diff
@@ -1,16 +1,40 @@
...
+ * If neither [offeringId] nor [offering] is provided, the SDK will use the current offering
+ * identifier from the cache.
+ * @property offering An optional [Offering] associated with the custom paywall. When provided, the
+ * SDK will derive both the offering identifier and the presented offering context (placement and
+ * targeting information) from this offering.
  */
 @Poko
 public class CustomPaywallImpressionParams @JvmOverloads constructor(
     public val paywallId: String? = null,
     public val offeringId: String? = null,
```
tonidero: I think we should deprecate the constructor with the `offeringId` and just leave one with the optional `offering` as the preferred way to do this, so there are less likely chances of folks passing us the offering id directly

Also, doesn't make sense to have this constructor with both the offeringId + offering. Instead, I would say to make the main constructor private or internal, and make it store just the paywallId + offeringId + presentedOfferingContext.

tonidero: I would also add a `@InternalRevenueCatAPI` constructor that received both an optional offeringId + presentedOfferingContext. This will be meant to be used by hybrids

tonidero (commented):
Just some suggestions in the public API but in general makes sense. Thank you for adding this!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/events/CustomPaywallImpressionParams.kt
rickvdl: Thanks @tonidero, great callouts. Refactored it a bit and implemented both suggestions :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -677,11 +677,41 @@ public class Purchases internal constructor(
...
+                resolvedOfferingId = params.offeringId
+                resolvedPresentedOfferingContext = params.presentedOfferingContext
+            }
+            params.offeringId != null -> {
+                val resolvedOffering = cachedOfferings?.get(params.offeringId)
+                resolvedOfferingId = params.offeringId
+                resolvedPresentedOfferingContext = resolvedOffering
+                    ?.availablePackages
+                    ?.firstOrNull()
+                    ?.presentedOfferingContext
```
tonidero, line 695: Not related to this PR but I'm seeing we have this code (getting the presented offering context from the first package of an offering) in a ton of places. This could/should be an internal function on the `Offering` object and uses that instead... But I think that should be on a separate PR.

api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
```diff
@@ -139,10 +142,21 @@ public void onError(@NonNull PurchasesError error) {}
         VirtualCurrencies cachedVirtualCurrencies = purchases.getCachedVirtualCurrencies();
 
         // trackCustomPaywallImpression API
+        Offering offering = new Offering(
```
tonidero: Hmm we could potentially just have this as a parameter to avoid having to create a new instance here, same for the kotlin API tests... NABD though.

tonidero (approved):
I think this makes a lot of sense! Thank you!!

api-tester/src/defaults/java/com/revenuecat/apitester/java/PurchasesAPI.java
rickvdl: Good idea, just updated it :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
rickvdl: Agree! Was thinking about that as well, will do so in a follow-up

rickvdl: Added [here](https://github.com/RevenueCat/purchases-android/pull/3513)

PR #3431, opened 2026-05-04
Propagate default package across workflow steps

vegaro:
## Motivation

In a multipage workflow paywall, `PackageComponent`s only live on the terminal step. Early "info" screens have no package components, so `selectedPackageInfo` is `null` on them. Price/period template variables like `{{ price }}` and `{{ price_per_year }}` can't resolve on those screens.

## Description

Adds a default-package fallback. The terminal step is identified by `singleStepFallbackId` on the workflow (added in #3436). When any step state is built, the terminal step's `selectedPackageInfo` is applied as the default. Steps with their own `PackageComponent`s ignore the default; steps without one resolve variables against it.

The default is set once per step and never overwritten, so back navigation always shows the same content as the initial render, even if the user changed the selection on the terminal step.

Changes:

- `PaywallState.Loaded.Components`: add `defaultPackageInfo` and `setDefaultPackage(info)`. `setDefaultPackage` is idempotent (set once, never overwritten). `selectedPackageInfo` is now `ownSelection ?: defaultPackageInfo`.
- `PaywallViewModel`: in `updateStateFromWorkflow`, pre-compute the `singleStepFallbackId` step first so its default is in cache before any other step loads. On every step build, apply the fallback step's `selectedPackageInfo` (idempotency means it only takes effect on the first visit). Log a warning if `singleStepFallbackId` points to a missing step.
- Tests: default-package application on packageless steps, back-navigation preserving the initial default, own selection winning over the default, graceful handling when `singleStepFallbackId` is missing. Plus extra `WorkflowNavigator` cases (deep backstack, non-linear navigation).

---

> [!NOTE]
> **Medium Risk**
> Touches workflow state construction/caching and selection precedence, which can subtly affect navigation/backstack behavior and displayed pricing context across steps.
> 
> **Overview**
> Adds a **default-package fallback** to multi-step workflow paywalls: `PaywallState.Loaded.Components` now stores an idempotent `defaultPackageInfo`, and `selectedPackageInfo` resolves as *own selection* falling back to that default.
> 
> `PaywallViewModel` now precomputes the `singleStepFallbackId` step to seed the cache, applies the fallback step’s `selectedPackageInfo` as a default on every step build/prewarm (without overwriting on revisits), and logs when the fallback step ID is missing.
> 
> Expands tests to cover context propagation on packageless steps, back-navigation stability, selection precedence, missing-fallback safety, plus additional `WorkflowNavigator` backstack/loop scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a9a862ddc922d3bf35d7f6297d4230ba7f477bae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -753,6 +753,17 @@ internal class PaywallViewModelImpl(
         workflowStepStateCache.clear()
         _workflowState.value = null
 
+        // Pre-compute the package step so its default package is available in cache
+        // for early packageless steps to use as context.
```
facumenzella, line 757: I think we should do this backend side. The default package should be the same for every paywall

facumenzella: Forget about this, we must do this SDK side

facumenzella (commented):
.

facumenzella (approved):

facumenzella:
I had a few comments that gh does not let me make 😢

facumenzella (changes requested):
rejecting till GitHub lets me put my comments 🙏

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -770,6 +781,16 @@ internal class PaywallViewModelImpl(
...
+        // Apply the workflow's default package to all steps. setDefaultPackage is idempotent
+        // so it is safe to call on every visit — it will only take effect the first time.
+        // On steps with their own packages, ownSelection takes precedence over defaultPackageInfo.
+        if (newState is PaywallState.Loaded.Components) {
+            val defaultPackage = workflow.singleStepFallbackId
+                ?.let { workflowStepStateCache[it]?.selectedPackageInfo }
+            if (defaultPackage != null) {
+                newState.setDefaultPackage(defaultPackage)
+            }
+        }
```
facumenzella, line 793: I wonder if we could do this at the prewarm stage instead? 🤔

facumenzella (approved):
I think it looks good. Just one nit open for debate 👍

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
vegaro: Agreed, but it needs to be here as well for the first step. So the prewarm of the following steps have this (it's idempotent only happens once) so when they load it's ready. Basically:

Prewarmed steps don’t have their `defaultPackageInfo` set until the user navigates to them. If `stepStates` is observed by the UI between prewarm completing and actual navigation, those states would have `defaultPackageInfo = null`. The initial step is built synchronously before prewarm even starts, so the initial step would lose its setDefaultPackage call. So we need to add the `setDefaultPackage` call inside `preWarmWorkflowStepCache`  but keep it in `buildStateFromStep` for the initial step.

PR #3435, opened 2026-05-05
Add network scaffolding for remote config endpoint

tonidero:
### Motivation

First step in landing remote-config support. The backend exposes `GET /v2/config`, which returns a manifest describing remote-controlled topics (e.g. product-entitlement mapping) plus the asset sources for downloading them. Wiring up the endpoint and its response models is a prerequisite for the orchestration and on-disk cache work in the rest of the stack.

### Description

- Adds `Endpoint.GetRemoteConfig` to the sealed endpoint hierarchy and `Backend.getRemoteConfig(appUserID, appInBackground, onSuccess, onError)` to dispatch the call. Reuses the existing `BackgroundAwareCallbackCacheKey` / `RemoteConfigCallback` callback-coalescing pattern.
- Adds the response models in a new feature package `com.revenuecat.purchases.common.remoteconfig`:
  - `RemoteConfigResponse(apiSources, blobSources, manifest)`
  - `ApiSource(id, url, priority, weight)`
  - `BlobSource(id, urlFormat, priority, weight)`
  - `Manifest(topics)`
  - `Topic` enum (`PRODUCT_ENTITLEMENT_MAPPING`) with `fromKey(key)` for wire-key lookups
  - `TopicEntry(blobRef)`
  - `TopicsMapSerializer` — custom kotlinx-serialization `KSerializer` that drops unknown topic wire keys at decode time and re-emits the enum's wire key on encode.
- Test coverage for the serializer (unknown-key filtering, default decoding, round-trip), plus `BackendGetRemoteConfigTest` for the endpoint plumbing and error paths.
- Adds a developer-only build-time override: setting `REMOTE_CONFIG_BASE_URL` in `local.properties` (e.g. `http://localhost:8080/`) redirects only `getRemoteConfig` to that host. Empty by default and gated to debug builds, so CI / release traffic is unaffected. Plumbed via a new `BuildConfig.REMOTE_CONFIG_BASE_URL` field, mirroring the existing `ENABLE_EXTRA_REQUEST_LOGGING` pattern.

All new types are `internal`; no public API surface change.

### Stack

- This PR (base): #3435
- ⬆ `Add RemoteConfigManager and TopicFetcher`: #3437
- ⬆ `Clean up unreferenced topic files after successful remote-config refresh`: #3439

### Checklist

- [x] Unit tests
- [ ] Follow-up issues for `purchases-ios` / hybrids (deferred — feature is not yet wired to any consumer)

---

> [!NOTE]
> **Medium Risk**
[...]

PR #3437, opened 2026-05-05
Add RemoteConfigManager and TopicFetcher

tonidero:
### Motivation

Builds on #3435 by adding the orchestration layer that turns a remote-config response into a populated on-disk cache of topic blobs. With this PR, calling `RemoteConfigManager.updateRemoteConfigIfNeeded` will (a) hit the backend for the latest manifest and (b) download every referenced topic blob, verify its SHA-256, and persist it under `noBackupFilesDir`. Subsequent stack PRs build on this orchestration.

### Description

Adds two classes in the `remoteconfig` package:

- **`TopicFetcher`** — exposes `suspend fun fetchTopicIfNeeded(topic, variant, topicEntry, source): PurchasesError?`. Given the inputs, downloads the blob from `source.urlFormat` (with `{blob_ref}` substituted), writes to a temp file, verifies SHA-256 against `topicEntry.blobRef`, and atomically renames into `noBackupFilesDir/RevenueCat/topics/<topic.key>/<blobRef>`. The blocking I/O is wrapped in `withContext(Dispatchers.IO)`. Skips the download entirely when the target file already exists. Errors surface as `PurchasesError(NetworkError, ...)` for checksum mismatch or `IOException → toPurchasesError()` for transport failures.
- **`RemoteConfigManager`** — keeps a callback boundary externally (`fun updateRemoteConfigIfNeeded(appInBackground, completion)`) since its caller (`PurchasesOrchestrator`) is callback-shaped. Internally launches a `private suspend fun refresh(...)` on a private `CoroutineScope(SupervisorJob() + dispatcher)` (`dispatcher: CoroutineDispatcher = Dispatchers.IO`, overridable for tests). The refresh:
  - Bridges the legacy callback-based `Backend.getRemoteConfig` via `suspendCancellableCoroutine` + `safeResume`/`safeResumeWithException(PurchasesException(error))`. The backend network call stays callback-based intentionally.
  - Fans the topic downloads out with `coroutineScope { tasks.map { async { topicFetcher.fetchTopicIfNeeded(...) } }.awaitAll().firstNotNullOfOrNull { it } }`. The first error wins; structured concurrency makes this much smaller than a hand-rolled `CompletionTracker`.

Currently only the `DEFAULT` variant of each known topic is downloaded.

### Tests

`runTest` + `UnconfinedTestDispatcher(testScheduler)` for both. `coEvery` / `coVerify` for the suspend `TopicFetcher`. Coverage:

- **`TopicFetcherTest`** — cache hit, successful download + verify + persist, URL-format substitution, HTTP failure, IO failure, SHA mismatch, multi-variant filename isolation.
- **`RemoteConfigManagerTest`** — empty-source / empty-topics / no-DEFAULT skip paths, first-error-wins fan-out, backend-error short-circuit, background flag forwarded, null completion.

All new types are `internal`. No public API surface change.

### Stack

- This PR: #3437
- ⬆ `Clean up unreferenced topic files after successful remote-config refresh`: #3439

### Checklist

- [x] Unit tests
- [ ] Follow-up issues for `purchases-ios` / hybrids (deferred)

---

> [!NOTE]
> **Medium Risk**
> Adds new remote-config orchestration that performs concurrent network downloads and filesystem writes under `noBackupFilesDir`, including checksum verification and new error-handling paths. While internal-only, bugs here could impact startup/network behavior and caching correctness.
> 
> **Overview**
> Adds a new remote-config orchestration layer: `RemoteConfigManager.updateRemoteConfigIfNeeded` fetches the remote-config manifest from `Backend.getRemoteConfig` (bridged into coroutines) and concurrently downloads each topic’s `default` entry via a new `TopicFetcher`, returning the first download error (if any).
> 
> Introduces `TopicFetcher` to cache topic blobs on disk under `noBackupFilesDir/RevenueCat/topics/<topic>/<blob_ref>`, with blob-ref input validation, limited parallel downloads, temp-file writes, and SHA-256 verification.
> 
> Refactors `FontLoader` to reuse new `UrlConnectionFactory.downloadToFile` helpers, and extends `UrlConnectionFactory` with shared download + checksum-verification utilities. Updates/expands tests to cover the new manager/fetcher behavior and changes the manifest entry key from `DEFAULT` to `default` in `RemoteConfigResponseTest`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8cdae44fe4a5f07af6e03e3e2fc8d73f0ba02e33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -46,9 +46,9 @@ internal class RemoteConfigManager(
         }
         // WIP: We should have some logic to pick the correct source for this. Right now, hardcoded to the first source.
         val source = response.blobSources.firstOrNull()
-        val tasks = response.manifest.topics.mapNotNull { (topic, variants) ->
-            val entry = variants[DEFAULT_VARIANT] ?: return@mapNotNull null
-            TopicTask(topic, DEFAULT_VARIANT, entry)
+        val tasks = response.manifest.topics.mapNotNull { (topic, entries) ->
+            val entry = entries[DEFAULT_ENTRY_ID] ?: return@mapNotNull null
+            TopicTask(topic, DEFAULT_ENTRY_ID, entry)
         }
         return if (source == null || tasks.isEmpty()) {
             null
@@ -58,7 +58,7 @@ internal class RemoteConfigManager(
                     async(downloadDispatcher) {
                         topicFetcher.fetchTopicIfNeeded(
                             topic = task.topic,
-                            variant = task.variant,
+                            entryId = task.entryId,
                             topicEntry = task.entry,
                             source = source,
                         )
@@ -77,10 +77,10 @@ internal class RemoteConfigManager(
             )
         }
 
-    private data class TopicTask(val topic: Topic, val variant: String, val entry: TopicEntry)
+    private data class TopicTask(val topic: Topic, val entryId: String, val entry: TopicEntry)
 
     private companion object {
-        const val DEFAULT_VARIANT = "default"
+        const val DEFAULT_ENTRY_ID = "default"
         const val MAX_PARALLEL_TOPIC_DOWNLOADS = 4
     }
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -1,79 +1,74 @@
 package com.revenuecat.purchases.common.remoteconfig
 
 import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesException
 import com.revenuecat.purchases.common.Backend
 import com.revenuecat.purchases.common.errorLog
+import com.revenuecat.purchases.common.safeResume
+import com.revenuecat.purchases.common.safeResumeWithException
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
 
 internal class RemoteConfigManager(
     private val backend: Backend,
     private val topicFetcher: TopicFetcher,
+    dispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) {
+    private val scope = CoroutineScope(SupervisorJob() + dispatcher)
+
     fun updateRemoteConfigIfNeeded(
-        appUserID: String,
         appInBackground: Boolean,
         completion: ((PurchasesError?) -> Unit)? = null,
     ) {
-        backend.getRemoteConfig(
-            appUserID = appUserID,
-            appInBackground = appInBackground,
-            onSuccess = { response -> fetchTopicsForResponse(response, completion) },
-            onError = { error ->
-                errorLog { "Failed to fetch remote config: $error" }
-                completion?.invoke(error)
-            },
-        )
+        scope.launch {
+            completion?.invoke(refresh(appInBackground))
+        }
     }
 
-    private fun fetchTopicsForResponse(
-        response: RemoteConfigResponse,
-        completion: ((PurchasesError?) -> Unit)?,
-    ) {
-        val source = response.sources.firstOrNull()
+    private suspend fun refresh(appInBackground: Boolean): PurchasesError? {
+        val response = try {
+            getRemoteConfig(appInBackground)
+        } catch (e: PurchasesException) {
+            errorLog { "Failed to fetch remote config: ${e.error}" }
+            return e.error
+        }
+        val source = response.blobSources.firstOrNull()
         val tasks = response.manifest.topics.mapNotNull { (topic, variants) ->
             val entry = variants[DEFAULT_VARIANT] ?: return@mapNotNull null
             TopicTask(topic, DEFAULT_VARIANT, entry)
         }
-        if (source == null || tasks.isEmpty()) {
-            completion?.invoke(null)
-            return
-        }
-        val tracker = CompletionTracker(tasks.size, completion)
-        tasks.forEach { task ->
-            topicFetcher.fetchTopicIfNeeded(
-                topic = task.topic,
-                variant = task.variant,
-                topicEntry = task.entry,
-                source = source,
-            ) { error -> tracker.recordCompletion(error) }
+        if (source == null || tasks.isEmpty()) return null
+        return coroutineScope {
+            tasks.map { task ->
+                async {
+                    topicFetcher.fetchTopicIfNeeded(
+                        topic = task.topic,
+                        variant = task.variant,
+                        topicEntry = task.entry,
+                        source = source,
+                    )
+                }
+            }.awaitAll().firstNotNullOfOrNull { it }
         }
     }
+
+    private suspend fun getRemoteConfig(appInBackground: Boolean): RemoteConfigResponse =
+        suspendCancellableCoroutine { cont ->
+            backend.getRemoteConfig(
+                appInBackground = appInBackground,
+                onSuccess = { cont.safeResume(it) },
+                onError = { cont.safeResumeWithException(PurchasesException(it)) },
+            )
+        }
 
     private data class TopicTask(val topic: Topic, val variant: String, val entry: TopicEntry)
-
-    private class CompletionTracker(
-        totalTasks: Int,
-        private val completion: ((PurchasesError?) -> Unit)?,
-    ) {
-        private val lock = Any()
-        private var remaining = totalTasks
-        private var firstError: PurchasesError? = null
-
-        fun recordCompletion(error: PurchasesError?) {
-            val shouldInvoke: Boolean
-            val errorToReport: PurchasesError?
-            synchronized(lock) {
-                if (error != null && firstError == null) {
-                    firstError = error
-                }
-                remaining -= 1
-                shouldInvoke = remaining == 0
-                errorToReport = firstError
-            }
-            if (shouldInvoke) {
-                completion?.invoke(errorToReport)
-            }
-        }
-    }
 
     private companion object {
         const val DEFAULT_VARIANT = "DEFAULT"
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -71,6 +71,6 @@ internal class RemoteConfigManager(
     private data class TopicTask(val topic: Topic, val variant: String, val entry: TopicEntry)
 
     private companion object {
-        const val DEFAULT_VARIANT = "DEFAULT"
+        const val DEFAULT_VARIANT = "default"
     }
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -44,18 +44,21 @@ internal class RemoteConfigManager(
             val entry = variants[DEFAULT_VARIANT] ?: return@mapNotNull null
             TopicTask(topic, DEFAULT_VARIANT, entry)
         }
-        if (source == null || tasks.isEmpty()) return null
-        return coroutineScope {
-            tasks.map { task ->
-                async {
-                    topicFetcher.fetchTopicIfNeeded(
-                        topic = task.topic,
-                        variant = task.variant,
-                        topicEntry = task.entry,
-                        source = source,
-                    )
-                }
-            }.awaitAll().firstNotNullOfOrNull { it }
+        return if (source == null || tasks.isEmpty()) {
+            null
+        } else {
+            coroutineScope {
+                tasks.map { task ->
+                    async {
+                        topicFetcher.fetchTopicIfNeeded(
+                            topic = task.topic,
+                            variant = task.variant,
+                            topicEntry = task.entry,
+                            source = source,
+                        )
+                    }
+                }.awaitAll().firstNotNullOfOrNull { it }
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


tonidero:
* **#3437** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3437?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -0,0 +1,79 @@
...
+        )
+    }
+
+    private fun fetchTopicsForResponse(
+        response: RemoteConfigResponse,
+        completion: ((PurchasesError?) -> Unit)?,
+    ) {
+        val source = response.blobSources.firstOrNull()
+        val tasks = response.manifest.topics.mapNotNull { (topic, variants) ->
+            val entry = variants[DEFAULT_VARIANT] ?: return@mapNotNull null
```
tonidero: This is still todo, right now only product entitlement mapping is fetched, which only has a a single entry in the topic, which is the default.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -0,0 +1,80 @@
...
+    private val scope = CoroutineScope(SupervisorJob() + dispatcher)
+
+    fun updateRemoteConfigIfNeeded(
+        appInBackground: Boolean,
+        completion: ((PurchasesError?) -> Unit)? = null,
+    ) {
+        scope.launch {
+            val error = refresh(appInBackground)
+            completion?.invoke(error)
+        }
```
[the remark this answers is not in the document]
tonidero: We should never throw from the refresh function, so this would be an unexpected situation... I think it's safe to ignore for now.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -0,0 +1,76 @@
...
+                appInBackground = appInBackground,
+                onSuccess = { cont.safeResume(it) },
+                onError = { cont.safeResumeWithException(PurchasesException(it)) },
+            )
+        }
+
+    private data class TopicTask(val topic: Topic, val variant: String, val entry: TopicEntry)
+
+    private companion object {
+        const val DEFAULT_VARIANT = "default"
```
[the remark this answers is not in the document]
ajpallares: This seems like a real thing?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -0,0 +1,80 @@
...
+                tasks.map { task ->
+                    async {
+                        topicFetcher.fetchTopicIfNeeded(
+                            topic = task.topic,
+                            variant = task.variant,
+                            topicEntry = task.entry,
+                            source = source,
+                        )
+                    }
+                }.awaitAll().firstNotNullOfOrNull { it }
```
ajpallares, line 62: I guess it's deliberate that we're downloading everything in parallel. Should we limit this somehow or we can expect this will be handled by the coroutines runtime correctly?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
```diff
@@ -0,0 +1,121 @@
...
+                    PurchasesErrorCode.NetworkError,
+                    "Downloaded topic $topic ($variant) failed SHA-256 verification.",
+                )
+            } catch (e: IOException) {
+                e.toPurchasesError().also { errorLog(it) }
+            }
+        }
+    }
+
+    private fun topicFile(topic: Topic, blobRef: String): File {
```
ajpallares, line 69: Should this file dir depend on the variant?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
```diff
@@ -0,0 +1,121 @@
...
+        return File(dir, blobRef)
+    }
+
+    @Throws(IOException::class, Checksum.ChecksumValidationException::class)
+    private fun downloadVerifyAndStore(url: String, expectedSha256: String, target: File) {
+        val parent = target.parentFile ?: throw IOException("Topic target file has no parent: $target")
+        val tempFile = File.createTempFile("rc_topic_", ".tmp", parent)
+        try {
+            downloadToFile(url, tempFile)
+            val actual = Checksum.generate(tempFile.readBytes(), Checksum.Algorithm.SHA256).value
```
ajpallares: Here we're loading the whole file into memory just to calculate its checksum. Would it be worth doing something like https://github.com/RevenueCat/purchases-android/blob/315b7706f0445126bb0457d748151f9cfc1ccca2/purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt#L281 to prevent that?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private suspend fun refresh(appInBackground: Boolean): PurchasesError? {
+        val response = try {
+            getRemoteConfig(appInBackground)
+        } catch (e: PurchasesException) {
+            errorLog { "Failed to fetch remote config: ${e.error}" }
+            return e.error
+        }
+        val source = response.blobSources.firstOrNull()
```
ajpallares, line 44: We're picking `blobSources.firstOrNull()` and ignoring `priority` / `weight`. Is that correct?

ajpallares (commented):
Some initial comments. But this is looking good!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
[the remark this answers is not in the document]
tonidero: right I fixed this in a follow-up branch and when importing it to this one, I only patched part of it initially, and fixed later. Should be fixed already :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
tonidero: No, the variant name is not great TBH... It references different files need to be downloaded. For example, once we have an offerings topic, each one would be a different "variant"... Since it's kinda generic, it was hard to come up with a name... Open to thoughts!

tonidero: It could potentially be more of a list TBH... since we would need to download all of them...

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Right now yeah. We need to think of the logic that will go with this TBH.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
tonidero: It's a good question... One of the reasons I did it this way was because that way we can verify that there weren't any issues while writing on disk either... But I can see your point, so I will change it to a stream based checksum. Thanks!

tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3437/commits/ba8adae8c8c1f036cec1a4a90df53b3676189049

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Good point... I was going with the simple case, which is that we only have a single topic/file right now and was planning to leave that for later, but it's true that with this, we might be downloading more once we add more topics in the future so we should future-proof it. Added limited parallelism in https://github.com/RevenueCat/purchases-android/pull/3437/commits/ba8adae8c8c1f036cec1a4a90df53b3676189049

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Added a comment for now regarding this

tonidero: Actually we do have the logic but I will implement this in a follow-up PR to avoid making this even bigger 🥴

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
```diff
@@ -0,0 +1,86 @@
...
+            } catch (e: Checksum.ChecksumValidationException) {
+                errorLog(e) { "Downloaded topic $topic ($variant) failed SHA-256 verification." }
+                PurchasesError(
+                    PurchasesErrorCode.NetworkError,
+                    "Downloaded topic $topic ($variant) failed SHA-256 verification.",
+                )
+            } catch (e: IOException) {
+                e.toPurchasesError().also { errorLog(it) }
+            }
+        }
```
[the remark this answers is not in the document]
ajpallares: Got this same comment from Cursor 🙃

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
ajpallares: Oh, I see. So the `blobRef` is what makes the file name already unique, right?

Regarding the "variant" naming, maybe "entryId" is better, since that's how it's referred to in the intetnal doc?

> It could potentially be more of a list TBH... since we would need to download all of them...

Hmm that could make sense since we're going to have to download them all anyway. We could have the list contain an `id` property instead of having the key. But honestly I don't have the full context yet to know what would be best. And also, in terms of quick lookup, if the `"DEFAULT"` topic is going to play an important role in certain topics (according to the specs), having it be a map/dictionary could make more sense

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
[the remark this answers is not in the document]
ajpallares: Oh I see you changed it all to "default" in lower case. But according to the specification document, isn't it expected to get `"DEFAULT"` (in uppercase)?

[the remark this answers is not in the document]
tonidero: Right, the mock server we've been using is actually using lowercase so I had to change to that (I originally also had it uppercased). I would say let's keep it like this and we can change it later if needed (this is not going to be used yet in any case)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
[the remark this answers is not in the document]
tonidero: Indeed. Fixed in https://github.com/RevenueCat/purchases-android/pull/3437/commits/8cdae44fe4a5f07af6e03e3e2fc8d73f0ba02e33

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
tonidero: Good one! Indeed, renamed in https://github.com/RevenueCat/purchases-android/pull/3437/commits/8cdae44fe4a5f07af6e03e3e2fc8d73f0ba02e33

tonidero: And yes, the blobRef is what makes the file unique within a topic.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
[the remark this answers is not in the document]
ajpallares: Sure! Makes sense! Thank you for clarifying!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
```diff
@@ -0,0 +1,110 @@
...
+                tempFile.delete()
+            }
+        }
+    }
+
+    private companion object {
+        const val TOPICS_ROOT = "RevenueCat/topics"
+        const val BLOB_REF_PLACEHOLDER = "{blob_ref}"
+        const val MAX_PARALLEL_TOPIC_DOWNLOADS = 4
+        val BLOB_REF_PATTERN = Regex("^[a-zA-Z0-9]+$")
```
ajpallares, line 108: Should we tighten `BLOB_REF_PATTERN` to exactly `^[a-fA-F0-9]{64}$`? Right now any alphanumeric string of any length (including non-hex chars like `G–Z`) passes validation and we only catch obviously-bad values after downloading and recomputing SHA-256. Since `blobRef` is the SHA-256 hex digest itself, enforcing the exact shape feels like cheap insurance and would make the "accepts mixed-case 64-char hex blobRef" test case actually exercise the real constraint.

ajpallares (approved):
I think it looks good! Just one minor comment, not a blocker though

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
tonidero: Indeed, I started with that, and then thought to futureproof it by relaxing the constraints... Though it's true that this just means it will fail later when verifying the checksum... So yeah, can change it back 👍

tonidero: Will do this in the follow-up PR, so we can merge this and avoid the extra iterations. Thanks again!

PR #3439, opened 2026-05-06
Clean up unreferenced topic files after successful remote-config refresh

tonidero:
### Motivation

`TopicFetcher` writes a new file every time the manifest's `blob_ref` rotates, but never deletes the previous one. Without cleanup, the on-disk cache at `noBackupFilesDir/RevenueCat/topics/...` grows unboundedly across backend rotations. This PR adds a one-shot disk-hygiene pass that runs at the end of each fully successful remote-config refresh.

### Description

- **`suspend fun TopicFetcher.cleanupUnreferencedTopics(referenced: Map<Topic, Set<String>>)`** — wraps the disk walk in `withContext(Dispatchers.IO)`, iterates `Topic.values()`, lists each topic's directory, and deletes files whose name isn't in the keep-set for that topic. Skips files prefixed with `rc_topic_` (the `createTempFile` prefix, now extracted as a `TEMP_PREFIX` constant) so a concurrent in-flight download isn't clobbered.
- **`RemoteConfigManager.refresh`** — builds the reference set as the union of every blob ref in the new manifest (across all variants, not just `DEFAULT`) and triggers cleanup at the right moment, all within the existing suspend coroutine flow:
  - When `sources` is empty or no DEFAULT-variant tasks remain → cleanup runs immediately (`fetchError = null` short-circuit).
  - When downloads dispatch → after the `async`/`awaitAll` fan-out, cleanup is awaited only if every fetch returned `null` (no error). Partial-failure refreshes leave the cache untouched.

Confirmed semantics:
- Reference set spans **all** variants in the new manifest (forward-compatible if non-DEFAULT variants ever start being cached).
- Topics in the SDK's `Topic` enum but absent from the new manifest get **all** their cached files wiped (reference-set = ∅).
- Cleanup runs **only** when every download in the refresh succeeded.

### Tests

`runTest` for both:

- 5 new `TopicFetcherTest` cases: delete unreferenced / wipe topic absent from set / no-op when root missing / preserve `rc_topic_*.tmp` / mixed keep+delete.
- 5 new `RemoteConfigManagerTest` cases (with `coEvery { topicFetcher.cleanupUnreferencedTopics(any()) } returns Unit` + `coVerify`): cleanup with all-variant refs / cleanup on empty sources & empty topics / no cleanup on fetcher error / no cleanup on backend error.

All changes are `internal`; no public API surface change.

### Stack

- ⬇ `Add RemoteConfigManager and TopicFetcher`: #3437
- This PR: #3439

### Checklist

- [x] Unit tests
- [ ] Follow-up issues for `purchases-ios` / hybrids (deferred)

---

> [!NOTE]
> **Medium Risk**
> Adds automatic deletion of on-disk topic files after successful remote-config refresh; mistakes in reference calculation or deletion logic could remove still-needed cached data, though it is guarded to run only on fully successful refreshes and has test coverage.
> 
> **Overview**
> After a fully successful remote-config refresh, the SDK now performs a one-shot cleanup pass to delete **unreferenced cached topic blob files** under `noBackupFilesDir/RevenueCat/topics`, preventing unbounded growth when `blob_ref`s rotate.
> 
> `RemoteConfigManager` builds a per-topic keep-set from *all* manifest entryIds’ `blob_ref`s and calls `TopicFetcher.cleanupUnreferencedTopics()` only when refresh completes with no errors (and caching conditions are met). `TopicFetcher` implements the disk walk/deletion, skips in-flight temp files via a shared `rc_topic_` prefix constant, and adds unit tests covering the new cleanup behavior and its failure/no-op cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9503c3688c647908aa63f955c52f75ecd12db710. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -79,11 +79,11 @@ internal class RemoteConfigManager(
             }
         }
         if (firstError == null) {
-            topicFetcher.cleanupUnreferencedTopics(referenced)
             // Only cache when at least one topic was actually fetched — empty sources, empty topics,
             // and missing default entryIds are treated as no-op refreshes that don't populate the cache.
             if (source != null && tasks.isNotEmpty()) {
                 val previousResponse = cache?.response
+                topicFetcher.cleanupUnreferencedTopics(referenced)
                 cache = CacheEntry(response, dateProvider.now)
                 if (previousResponse != response) {
                     diskCache.write(response)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -57,6 +57,7 @@ internal class RemoteConfigManager(
         }
         // WIP: We should have some logic to pick the correct source for this. Right now, hardcoded to the first source.
         val source = response.blobSources.firstOrNull()
+        val referenced = buildReferenceSet(response.manifest)
         val tasks = response.manifest.topics.mapNotNull { (topic, entries) ->
             val entry = entries[DEFAULT_ENTRY_ID] ?: return@mapNotNull null
             TopicTask(topic, DEFAULT_ENTRY_ID, entry)
@@ -77,13 +78,16 @@ internal class RemoteConfigManager(
                 }.awaitAll().firstNotNullOfOrNull { it }
             }
         }
-        // Only cache when at least one topic was actually fetched — empty sources, empty topics,
-        // and missing default entryIds are treated as no-op refreshes that don't populate the cache.
-        if (firstError == null && source != null && tasks.isNotEmpty()) {
-            val previousResponse = cache?.response
-            cache = CacheEntry(response, dateProvider.now)
-            if (previousResponse != response) {
-                diskCache.write(response)
+        if (firstError == null) {
+            topicFetcher.cleanupUnreferencedTopics(referenced)
+            // Only cache when at least one topic was actually fetched — empty sources, empty topics,
+            // and missing default entryIds are treated as no-op refreshes that don't populate the cache.
+            if (source != null && tasks.isNotEmpty()) {
+                val previousResponse = cache?.response
+                cache = CacheEntry(response, dateProvider.now)
+                if (previousResponse != response) {
+                    diskCache.write(response)
+                }
             }
         }
         return firstError
@@ -96,6 +100,11 @@ internal class RemoteConfigManager(
                 onSuccess = { cont.safeResume(it) },
                 onError = { cont.safeResumeWithException(PurchasesException(it)) },
             )
+        }
+
+    private fun buildReferenceSet(manifest: Manifest): Map<Topic, Set<String>> =
+        manifest.topics.mapValues { (_, entries) ->
+            entries.values.map { it.blobRef }.toSet()
         }
 
     private data class TopicTask(val topic: Topic, val entryId: String, val entry: TopicEntry)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
@@ -66,6 +66,34 @@ internal class TopicFetcher(
         }
     }
 
+    suspend fun cleanupUnreferencedTopics(referenced: Map<Topic, Set<String>>) {
+        withContext(Dispatchers.IO) {
+            runCatching { deleteUnreferencedTopicFiles(referenced) }
+                .onFailure { e ->
+                    errorLog(e) { "Failed to clean up unreferenced topic files." }
+                }
+        }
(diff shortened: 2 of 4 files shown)


tonidero:
* **#3439** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3439?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3439?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -77,13 +78,16 @@ internal class RemoteConfigManager(
...
-        if (firstError == null && source != null && tasks.isNotEmpty()) {
-            val previousResponse = cache?.response
-            cache = CacheEntry(response, dateProvider.now)
-            if (previousResponse != response) {
-                diskCache.write(response)
+        if (firstError == null) {
+            topicFetcher.cleanupUnreferencedTopics(referenced)
+            // Only cache when at least one topic was actually fetched — empty sources, empty topics,
+            // and missing default entryIds are treated as no-op refreshes that don't populate the cache.
+            if (source != null && tasks.isNotEmpty()) {
```
ajpallares, line 84: Hmm I wonder when exactly we should call
```
topicFetcher.cleanupUnreferencedTopics(referenced)
```

As of now, this we could end up cleaning too much after receiving an invalid state (e.g.  if `source == null` or `tasks.isEmpty()`). That tells me we should call `cleanupUnreferencedTopics` inside `if (source != null && tasks.isNotEmpty())`.

Now, what if `(previousResponse == response)`? Should we even call `cleanupUnreferencedTopics` then?

ajpallares (commented):
Looking good! I do have one concern that we may want to address before merging

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Good one! I had not tested this after the changes in the base branches 🥴. But yeah, definitely inside the `if (source != null && tasks.isNotEmpty())`. Regarding the other... I think it's probably fine to keep out, so in case something fails during a previous cleanup, it retries later.

tonidero: Fixed in https://github.com/RevenueCat/purchases-android/pull/3439/commits/9503c3688c647908aa63f955c52f75ecd12db710

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
```diff
@@ -66,6 +66,34 @@ internal class TopicFetcher(
...
+                .onFailure { e ->
+                    errorLog(e) { "Failed to clean up unreferenced topic files." }
+                }
+        }
+    }
+
+    private fun deleteUnreferencedTopicFiles(referenced: Map<Topic, Set<String>>) {
+        val topicsRoot = File(applicationContext.noBackupFilesDir, TOPICS_ROOT)
+        if (!topicsRoot.exists()) return
+        Topic.values().forEach topics@{ topic ->
```
ajpallares, line 81: Seems like since Kotlin 1.9, `entries` is the recommended replacement for `values()`
```suggestion
        Topic.entries.forEach topics@{ topic ->
```

ajpallares, line 81: If we ever remove a case from the `Topic` enum, we would end up with an unused directory here. I guess that's fine for now

ajpallares (approved):
Great progress!! A couple of nits, but no blockers. Good job!

tonidero:
### Merge activity

* **May 12, 2:06 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/RevenueCat/purchases-android/3439).
* **May 12, 2:07 PM UTC**: [Graphite](https://app.graphite.com/github/pr/RevenueCat/purchases-android/3439) couldn't merge this PR because it failed for an unknown reason (This repository has GitHub's merge queue enabled. Please configure the GitHub merge queue integration in Graphite settings.).

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
tonidero: Indeed... but that requires opting in to an experimental flag because we are supporting older kotlin versions... So I think it's better to keep as is :)

tonidero: That's a valid point... Could be a further cleanup we do but I won't block the PR because of this. Good catch though!

ajpallares: Agreed!

PR #3430, opened 2026-05-04
Abstract workflow page transition animation behind sealed class 

vegaro:
Refactor creating a `WorkflowTransitionAnimation` that can be easily extended in the future. This PR doesn't change any behavior

---

> [!NOTE]
> **Low Risk**
> Low risk refactor that preserves existing slide behavior but rewires workflow-step transition plumbing; main risk is subtle regressions in step/header selection or animation timing.
> 
> **Overview**
> Refactors workflow paywall step transitions to be driven by a new `WorkflowTransitionAnimation`/`WorkflowTransitionState` sealed-class abstraction, keeping the current Slide-In/Out behavior as the default.
> 
> `LoadedWorkflowPaywall` and related composables now use `rememberWorkflowTransitionState` and `Modifier.workflowTransition` (instead of the slide-specific state/modifier), and header selection logic is updated to derive any pending transition from the active transition-state subtype.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27b67503cb1b25ddd5a580ca0b01fe713870b745. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3441, opened 2026-05-06
Simplify `WorkflowTransitionState` with explicit from/to step fields

vegaro:
Cleanup from a review comment on #3430.

`visibleStepIds: Set` was redundant with `animatingFromStepId: String?`: the set was always either `{from, to}` or `{to}`, derivable from the from/to pair. Replaced both with explicit `animatingFromStepId` + `animatingToStepId`. This was there becuase I was attempting some performance improvements in other branch that used a set, but we are going with just from and to for now.

Also dropped some dead code:
- the unreachable parked-step branch in `applyWorkflowTransition` (vestige of an old "render all steps" design)
- the `currentStepId` param of `workflowTransition`, now read off `state.animatingToStepId`

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes which workflow steps are kept in composition and how slide translations are computed, which could cause visual/animation regressions if edge cases exist.
> 
> **Overview**
> Simplifies workflow paywall transitions by removing `visibleStepIds` and making `WorkflowTransitionState` explicitly track `animatingFromStepId` and `animatingToStepId`.
> 
> Rendering now only composes the outgoing + incoming steps during a transition (via `listOfNotNull(from,to)`), and `workflowTransition` no longer requires `currentStepId`, instead deriving positioning from `state.animatingToStepId`. Dead/unused “parked step” translation logic is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67fadd3be53f6361148cb5613f411f981c9aef34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
@@ -136,19 +136,20 @@ private fun WorkflowStepsContent(
     // a top clip causes the hero image (which renders behind the status bar) to get cropped
     // during the slide transition.
     Box(modifier = Modifier.fillMaxSize()) {
-        for (stepId in transitionState.visibleStepIds) {
-            val stepState = stepStates[stepId] ?: continue
-            key(stepId) {
-                WorkflowStepContent(
-                    stepId = stepId,
-                    stepState = stepState,
-                    currentStepId = currentStepId,
-                    transitionState = transitionState,
-                    clickHandler = clickHandler,
-                    componentInteractionTracker = componentInteractionTracker,
-                )
+        listOfNotNull(transitionState.animatingFromStepId, transitionState.animatingToStepId)
+            .forEach { stepId ->
+                val stepState = stepStates[stepId] ?: return@forEach
+                key(stepId) {
+                    WorkflowStepContent(
+                        stepId = stepId,
+                        stepState = stepState,
+                        currentStepId = currentStepId,
+                        transitionState = transitionState,
+                        clickHandler = clickHandler,
+                        componentInteractionTracker = componentInteractionTracker,
+                    )
+                }
             }
-        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
```diff
@@ -136,19 +136,20 @@ private fun WorkflowStepsContent(
...
-                WorkflowStepContent(
-                    stepId = stepId,
-                    stepState = stepState,
-                    currentStepId = currentStepId,
-                    transitionState = transitionState,
-                    clickHandler = clickHandler,
-                    componentInteractionTracker = componentInteractionTracker,
-                )
+        listOfNotNull(transitionState.animatingFromStepId, transitionState.animatingToStepId)
+            .forEach { stepId ->
```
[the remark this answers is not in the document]
vegaro: not worried about this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
```diff
@@ -136,19 +136,20 @@ private fun WorkflowStepsContent(
...
-            key(stepId) {
-                WorkflowStepContent(
-                    stepId = stepId,
-                    stepState = stepState,
-                    currentStepId = currentStepId,
-                    transitionState = transitionState,
-                    clickHandler = clickHandler,
-                    componentInteractionTracker = componentInteractionTracker,
-                )
+        listOfNotNull(transitionState.animatingFromStepId, transitionState.animatingToStepId)
```
tonidero, line 139: I guess this does mean we won't be supporting "peeking" into other pages... But that's probably ok for now.

tonidero (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
vegaro: yeah, that's fine. I think at that point we should probably explore horizontal pagers

PR #3451, opened 2026-05-07
fix: url encode query prameters

JZDesign:
### Checklist

### Motivation

URL query parameters with spaces caused crashes

### Description

url encode query parameters to prevent crashes

---

> [!NOTE]
> **Medium Risk**
> Changes how web checkout/custom URLs are constructed by encoding query components and inserting parameters before fragments, which could subtly affect downstream URL parsing and expectations in integrations.
> 
> **Overview**
> Fixes URL construction when appending query params by **URL-encoding parameter names/values** and ensuring new params are inserted *before* any `#fragment` in `URI.appendQueryParameter`.
> 
> `URL.appendQueryParameter` is simplified to delegate through the `URI` implementation, and tests are updated/added to assert encoding behavior (including whitespace) and fragment handling; web-checkout URL expectations now reflect encoded values (e.g. `%24rc_monthly`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4a604b8063ebba8b6a2f57e5ba85d019ce09452. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
@@ -2,6 +2,7 @@
 
 package com.revenuecat.purchases.ui.revenuecatui.utils
 
+import android.net.Uri
 import java.net.URI
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets
@@ -18,5 +19,4 @@ internal fun URI.appendQueryParameter(name: String, value: String): URI {
     return URI("$uriWithoutFragment$separator$encodedParameter$fragment")
 }
 
-private fun String.encodeQueryParameterComponent(): String =
-    URLEncoder.encode(this, StandardCharsets.UTF_8.name()).replace("+", "%20")
+private fun String.encodeQueryParameterComponent(): String = Uri.encode(this)
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
@@ -3,9 +3,20 @@
 package com.revenuecat.purchases.ui.revenuecatui.utils
 
 import java.net.URI
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets
 
 @JvmSynthetic
 internal fun URI.appendQueryParameter(name: String, value: String): URI {
-    val separator = if (this.query == null) "?" else "&"
-    return URI("$this$separator$name=$value")
+    val encodedParameter = "${name.encodeQueryParameterComponent()}=${value.encodeQueryParameterComponent()}"
+    val uriString = toString()
+    val fragmentIndex = uriString.indexOf('#')
+    val uriWithoutFragment = if (fragmentIndex == -1) uriString else uriString.substring(0, fragmentIndex)
+    val fragment = if (fragmentIndex == -1) "" else uriString.substring(fragmentIndex)
+    val separator = if (this.rawQuery == null) "?" else "&"
+
+    return URI("$uriWithoutFragment$separator$encodedParameter$fragment")
 }
+
+private fun String.encodeQueryParameterComponent(): String =
+    URLEncoder.encode(this, StandardCharsets.UTF_8.name()).replace("+", "%20")
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
@@ -4,8 +4,6 @@ package com.revenuecat.purchases.ui.revenuecatui.utils
 
 import android.net.Uri
 import java.net.URI
-import java.net.URLEncoder
-import java.nio.charset.StandardCharsets
 
 @JvmSynthetic
 internal fun URI.appendQueryParameter(name: String, value: String): URI {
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
```diff
@@ -3,9 +3,20 @@
...
+    val fragmentIndex = uriString.indexOf('#')
+    val uriWithoutFragment = if (fragmentIndex == -1) uriString else uriString.substring(0, fragmentIndex)
+    val fragment = if (fragmentIndex == -1) "" else uriString.substring(fragmentIndex)
+    val separator = if (this.rawQuery == null) "?" else "&"
+
+    return URI("$uriWithoutFragment$separator$encodedParameter$fragment")
 }
+
+private fun String.encodeQueryParameterComponent(): String =
+    URLEncoder.encode(this, StandardCharsets.UTF_8.name()).replace("+", "%20")
```
JayShortway: Just curious, why was the manual `.replace("+", "%20")` added? A `+` is allowed in URLs afaik.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
```diff
@@ -2006,7 +2006,7 @@ class PaywallViewModelTest {
         val model = create(offering = offeringWithWPL)
         assertThat(
             model.getWebCheckoutUrl(launchWebCheckoutWithCustomUrlAndPackage),
-        ).isEqualTo("https://revenuecat.com?rc_package=\$rc_monthly")
+        ).isEqualTo("https://revenuecat.com?rc_package=%24rc_monthly")
```
tonidero, line 2009: We might want to confirm the web checkout urls work with these encoded versions... I assume so, but just to check.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
```diff
@@ -3,9 +3,20 @@
...
+import java.net.URLEncoder
+import java.nio.charset.StandardCharsets
 
 @JvmSynthetic
 internal fun URI.appendQueryParameter(name: String, value: String): URI {
-    val separator = if (this.query == null) "?" else "&"
-    return URI("$this$separator$name=$value")
+    val encodedParameter = "${name.encodeQueryParameterComponent()}=${value.encodeQueryParameterComponent()}"
+    val uriString = toString()
+    val fragmentIndex = uriString.indexOf('#')
```
tonidero, line 12: Hmm I'm thinking, do we support fragment in these urls? Just wondering if this is needed at all... If we have it in iOS though, I think it might make sense to support it. Otherwise, might be better to reduce complexity and leave it out? NABD.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
tonidero: Hmm in other places we just use `Uri.encode`... Does that have any issues? Just wondering if we should try to reuse that instead.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URLExtensions.kt
```diff
@@ -5,7 +5,5 @@ package com.revenuecat.purchases.ui.revenuecatui.utils
 import java.net.URL
 
 @JvmSynthetic
-internal fun URL.appendQueryParameter(name: String, value: String): URL {
-    val separator = if (this.query == null) "?" else "&"
-    return URL("$this$separator$name=$value")
-}
+internal fun URL.appendQueryParameter(name: String, value: String): URL =
+    toURI().appendQueryParameter(name, value).toURL()
```
tonidero, line 9: Good to avoid repeating the logic 👍 Thanks!

tonidero (approved):
Just some thoughts... But not blockers I would say. Thanks for fixing this!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
JZDesign: This was a GPT 5.5 change. The reasoning made sense to me. It is:

> URLEncoder is for application/x-www-form-urlencoded, where spaces become +. In a URI query string, + is legal but not universally equivalent to a space unless the receiver applies form decoding. %20 is the safer URI encoding and avoids ambiguity.

> This replacement is safe for literal plus signs too: URLEncoder turns + into %2B before our .replace("+", "%20"), so:

>  "A+B C" -> "A%2BB%20C"

>  So the current .replace("+", "%20") is intentional.

JZDesign: I will look into that!

JZDesign: but @tonidero had good feedback and I changed this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/utils/URIExtensions.kt
JZDesign: I believe this to be how URLComponents handles things on iOS actually

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModelTest.kt
JZDesign: It works

PR #3450, opened 2026-05-07
Wire RemoteConfigManager refresh on first foreground and identity changes (off by default)

tonidero:
### Motivation

Final piece of the initial remote-config wiring: hook `RemoteConfigManager` into the SDK's actual lifecycle so the topic-fetch pipeline runs on real devices. The earlier PRs in the stack added the network scaffolding (#3435), the manager + topic fetcher (#3437), and the unreferenced-topic cleanup (#3439). None of them invoke the manager — this PR does, behind a developer-only build flag so production traffic is unaffected.

### Description

- Constructs `RemoteConfigManager` (with a `TopicFetcher`) inside `PurchasesFactory` and passes it into `PurchasesOrchestrator`.
- Adds a private `refreshRemoteConfigIfEnabled(appUserID)` helper on `PurchasesOrchestrator` that calls `remoteConfigManager.updateRemoteConfigIfNeeded(...)`. The helper is invoked from four places, each of which already runs only when something meaningful changed:
  - `onAppForegrounded()` — only when `state.firstTimeInForeground` is true (i.e. first foreground after `Purchases.configure(...)`, exactly once per `Purchases` instance).
  - `logIn(...)` `onSuccess` — only reached when the new `appUserID` differs from the current one.
  - `logOut(...)` success branch — the `appUserID` has just been replaced with a fresh anonymous id.
  - `switchUser(...)` — the early-return on the same-id path filters out no-ops before we get here.
- Every call site is gated on the new compile-time flag `BuildConfig.ENABLE_REMOTE_CONFIG`, fed from `local.properties` (`ENABLE_REMOTE_CONFIG=true`). Default: `false`, so disabling the flag is a no-op at every call site. Mirrors the existing `ENABLE_EXTRA_REQUEST_LOGGING` pattern.
- Existing orchestrator tests are updated to pass a relaxed mock for the new constructor parameter. No tests are added for the flag-on path: `BuildConfig.ENABLE_REMOTE_CONFIG` is `false` in CI unit-test builds, so that branch is unreachable from JUnit. Same precedent as `ENABLE_EXTRA_REQUEST_LOGGING`. The on-path is exercised manually via `purchase-tester` against a localhost backend (combine with `REMOTE_CONFIG_BASE_URL` from #3435).

All new types remain `internal`. No public API surface change.

### Stack

- ⬇ `Add remote-config network scaffolding`: #3435
- ⬇ `Add RemoteConfigManager and TopicFetcher`: #3437
[...]

PR #3455, opened 2026-05-07
Use product_entitlement_mapping topic blob for offline entitlements when remote config enabled

tonidero:
### Motivation

When `BuildConfig.ENABLE_REMOTE_CONFIG` is on, the remote-config pipeline (#3435/#3437/#3439/#3450) downloads the `product_entitlement_mapping` topic blob to disk. The SDK should consume that file as the source-of-truth for offline entitlements instead of issuing the legacy `GET /v1/product_entitlement_mapping` call. When the flag is off, behavior must be unchanged.

### Description

Behavior gated on `BuildConfig.ENABLE_REMOTE_CONFIG`, surfaced into the SDK via a single `useRemoteConfigForProductEntitlementMapping` value computed once in `PurchasesFactory`:

- **Stop the direct backend fetch.** `OfflineEntitlementsManager.updateProductEntitlementMappingCacheIfStale` becomes a no-op (`completion?.invoke(null)` and return) when the flag is on, so `Backend.getProductEntitlementMapping` is never issued.
- **Read the topic file at consumer-call time, off-main, with in-memory caching.** New `ProductEntitlementMappingTopicReader`:
  - `suspend fun read(): ProductEntitlementMapping?` — reads the single non-`rc_topic_` file in `noBackupFilesDir/RevenueCat/topics/product_entitlement_mapping/` and parses it via `ProductEntitlementMapping.fromJson(JSONObject(...), loadedFromCache = true)`.
  - Coalesces concurrent callers via a shared `Deferred<ProductEntitlementMapping?>` held in `inFlight`, guarded by `synchronized(lock)`. Started with `CoroutineStart.LAZY` so `inFlight` is assigned **before** the body runs (otherwise an eager dispatcher would let the body's `inFlight = null` execute and `.also { inFlight = it }` would overwrite it back with a completed deferred, leaking it into the next read).
  - Owns its own `CoroutineScope(SupervisorJob() + dispatcher)` (`dispatcher: CoroutineDispatcher = Dispatchers.IO`, overridable for tests) so cancellation of one caller doesn't cancel the load others are awaiting.
  - `fun invalidate()` is non-suspend (it's called from a synchronous topic-update listener); just clears `cached` under the lock.
- **Fall back to the existing `SharedPreferences` cache** if the topic file is absent (first launch, download in-flight). New `ProductEntitlementMappingSource` interface (kept callback-shaped because its consumer `PurchasedProductsFetcher.queryActiveProducts` is pre-existing callback API):
  - `DeviceCacheProductEntitlementMappingSource` — synchronous, used when the flag is off.
  - `RemoteConfigProductEntitlementMappingSource` — owns a small `CoroutineScope`, launches `reader.read()` (suspend) and forwards `topicMapping ?: deviceCache.getProductEntitlementMapping()` to the callback.
- **Invalidate the in-memory cache when a fresh blob is downloaded.** `TopicFetcher` gains a `topicUpdatedListener: ((Topic) -> Unit)? = null` parameter, invoked after a successful `downloadVerifyAndStore`. `PurchasesFactory` registers a listener that calls `productEntitlementMappingTopicReader.invalidate()` when the `PRODUCT_ENTITLEMENT_MAPPING` topic is rewritten, so a new manifest takes effect within the same session.
- **Wiring (`PurchasesFactory`).** Hoists `TopicFetcher` out of `RemoteConfigManager`'s inline construction so the same instance is shared between the manager and the reader's invalidation hook. The flag funnels in at one place, matching the existing `PurchasesOrchestrator.refreshRemoteConfigIfEnabled` gating style.

### Tests

`runTest` for everything; `UnconfinedTestDispatcher` for the simple paths and `StandardTestDispatcher` for the coalescing test (so two callers can interleave before the body runs):

- `ProductEntitlementMappingTopicReaderTest` (new) — missing dir, empty dir, valid blob parsed, second read returns same instance with no extra disk access (counted via `noBackupFilesDir` mock answer), ignores `rc_topic_` prefix files, corrupt JSON returns null, `invalidate()` clears the cache, concurrent reads coalesce into a single disk access.
- `OfflineEntitlementsManagerTest` — adds a "skips backend fetch when `useRemoteConfigForProductEntitlementMapping = true`" case and confirms the existing 25h staleness logic still runs when the flag is off.
- `PurchasedProductsFetcherTest` — already callback-shaped via `ProductEntitlementMappingSource`; reused unchanged.
- `TopicFetcherTest` — adds three listener cases (fires after successful download, doesn't fire on cache hit, doesn't fire on download error).

### Trade-offs

- The topic blob is read into a `ByteArray` and decoded as UTF-8 JSON. Realistic mapping size is well under 1 MB; revisit if it grows materially.
- We do not bridge a topic-file mapping into `SharedPreferences`. The two stores stay independent — SharedPreferences exists only as a fallback. With the flag on and the topic system healthy, the SharedPreferences cache will go stale; that's intentional.
- `Backend.getProductEntitlementMapping` stays callback-based — it's the legacy network path.

All changes are `internal`; no public API surface change.

### Stack

- ⬇ `Add RemoteConfigManager and TopicFetcher`: #3437
- ⬇ `Clean up unreferenced topic files after successful remote-config refresh`: #3439
- ⬇ `Wire RemoteConfigManager refresh on first foreground and identity changes`: #3450
- This PR: #3455

### Checklist

- [x] Unit tests
- [ ] Follow-up issues for `purchases-ios` / hybrids (deferred — feature is not yet wired to any consumer)

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -6,6 +6,7 @@ import android.content.Context
 import android.content.pm.PackageManager
 import androidx.annotation.VisibleForTesting
 import androidx.core.os.UserManagerCompat
+import com.revenuecat.purchases.api.BuildConfig
 import com.revenuecat.purchases.common.AppConfig
 import com.revenuecat.purchases.common.Backend
 import com.revenuecat.purchases.common.BackendHelper
@@ -33,11 +34,16 @@ import com.revenuecat.purchases.common.networking.ETagManager
 import com.revenuecat.purchases.common.offerings.OfferingsCache
 import com.revenuecat.purchases.common.offerings.OfferingsFactory
 import com.revenuecat.purchases.common.offerings.OfferingsManager
+import com.revenuecat.purchases.common.offlineentitlements.DeviceCacheProductEntitlementMappingSource
 import com.revenuecat.purchases.common.offlineentitlements.OfflineCustomerInfoCalculator
 import com.revenuecat.purchases.common.offlineentitlements.OfflineEntitlementsManager
+import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMappingSource
+import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMappingTopicReader
 import com.revenuecat.purchases.common.offlineentitlements.PurchasedProductsFetcher
+import com.revenuecat.purchases.common.offlineentitlements.RemoteConfigProductEntitlementMappingSource
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigDiskCache
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigManager
+import com.revenuecat.purchases.common.remoteconfig.Topic
 import com.revenuecat.purchases.common.remoteconfig.TopicFetcher
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.common.verification.SigningManager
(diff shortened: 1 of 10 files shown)


tonidero:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3455?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3455** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3455?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tonidero:
Going to close this PR since I'm working off a different stack of PRs (see https://github.com/RevenueCat/purchases-android/pull/3534) and it's easier to start over from there.

PR #3457, opened 2026-05-08
Cache remote config response in memory with TTL and persist to disk

tonidero:
### Description
This caches the remote config response both on disk and in memory. Right now the in memory cache is only used to avoid fetching the config multiple times without need. The disk cache is currently unused but will be used in future PRs.

The idea is to use this cache for the api config to be used to perform other SDK requests.

---

> [!NOTE]
> **Medium Risk**
> Introduces new caching and disk I/O paths in remote config fetching, which can affect refresh behavior and persistence semantics if TTL logic or write conditions are incorrect.
> 
> **Overview**
> Adds an in-memory TTL cache to `RemoteConfigManager` so repeated `updateRemoteConfigIfNeeded` calls can short-circuit without hitting the backend when the cached response is still fresh (foreground vs background TTL via `isCacheStale`).
> 
> Introduces `RemoteConfigDiskCache` to atomically persist the latest *successful* `RemoteConfigResponse` to `noBackupFilesDir/RevenueCat/remote_config.json`, writing only when at least one topic fetch occurred and the response changed.
> 
> Tightens `TopicFetcher` blob ref validation from “alphanumeric” to a strict 64-char hex SHA-256 format, with expanded tests covering acceptance/rejection cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f91f3d8ebf0d4d8a9f9200218ca9dc4e4909bf2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
@@ -1,0 +1,51 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+import android.content.Context
+import com.revenuecat.purchases.common.errorLog
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import java.io.File
+import java.io.IOException
+
+/**
+ * Persists the latest successful [RemoteConfigResponse] to disk under
+ * `noBackupFilesDir/RevenueCat/remote_config.json`.
+ *
+ * This is currently write-only. A subsequent PR will read this file at startup to seed the
+ * in-memory cache before the first network refresh.
+ */
+internal class RemoteConfigDiskCache(
+    private val applicationContext: Context,
+    private val json: Json,
+) {
+    fun write(response: RemoteConfigResponse) {
+        val parent = File(applicationContext.noBackupFilesDir, REMOTE_CONFIG_ROOT)
+        try {
+            if (!parent.exists()) {
+                parent.mkdirs()
+            }
+            val target = File(parent, REMOTE_CONFIG_FILE_NAME)
+            val tempFile = File.createTempFile("rc_remote_config_", ".tmp", parent)
+            try {
+                tempFile.writeText(json.encodeToString(response))
+                if (!tempFile.renameTo(target)) {
+                    tempFile.copyTo(target, overwrite = true)
+                }
+            } finally {
+                if (tempFile.exists()) {
+                    tempFile.delete()
+                }
+            }
+        } catch (e: IOException) {
+            errorLog(e) { "Failed to persist remote config to disk." }
+        } catch (e: SerializationException) {
+            errorLog(e) { "Failed to serialize remote config for disk persistence." }
+        }
+    }
+
+    private companion object {
+        const val REMOTE_CONFIG_ROOT = "RevenueCat"
+        const val REMOTE_CONFIG_FILE_NAME = "remote_config.json"
+    }
+}
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
@@ -19,6 +19,7 @@ internal class RemoteConfigDiskCache(
     private val applicationContext: Context,
     private val json: Json,
 ) {
+    /** Performs blocking disk I/O — call from a background dispatcher (e.g. [kotlinx.coroutines.Dispatchers.IO]). */
     fun write(response: RemoteConfigResponse) {
         val parent = File(applicationContext.noBackupFilesDir, REMOTE_CONFIG_ROOT)
         try {
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -1,8 +1,12 @@
 package com.revenuecat.purchases.common.remoteconfig
 
+import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.PurchasesException
 import com.revenuecat.purchases.common.Backend
+import com.revenuecat.purchases.common.DateProvider
+import com.revenuecat.purchases.common.DefaultDateProvider
+import com.revenuecat.purchases.common.caching.isCacheStale
 import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.safeResume
 import com.revenuecat.purchases.common.safeResumeWithException
@@ -15,18 +19,29 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
+import java.util.Date
 
+@OptIn(InternalRevenueCatAPI::class)
 internal class RemoteConfigManager(
     private val backend: Backend,
     private val topicFetcher: TopicFetcher,
+    private val diskCache: RemoteConfigDiskCache,
+    private val dateProvider: DateProvider = DefaultDateProvider(),
     dispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) {
     private val scope = CoroutineScope(SupervisorJob() + dispatcher)
+
+    @Volatile
+    private var cache: CacheEntry? = null
 
     fun updateRemoteConfigIfNeeded(
         appInBackground: Boolean,
         completion: ((PurchasesError?) -> Unit)? = null,
     ) {
+        if (!cache?.cachedAt.isCacheStale(appInBackground, dateProvider)) {
+            completion?.invoke(null)
+            return
+        }
         scope.launch {
             val error = refresh(appInBackground)
             completion?.invoke(error)
@@ -46,7 +61,7 @@ internal class RemoteConfigManager(
             val entry = entries[DEFAULT_ENTRY_ID] ?: return@mapNotNull null
             TopicTask(topic, DEFAULT_ENTRY_ID, entry)
         }
-        return if (source == null || tasks.isEmpty()) {
+        val firstError: PurchasesError? = if (source == null || tasks.isEmpty()) {
             null
         } else {
             coroutineScope {
@@ -62,6 +77,11 @@ internal class RemoteConfigManager(
                 }.awaitAll().firstNotNullOfOrNull { it }
             }
         }
+        if (firstError == null) {
+            cache = CacheEntry(response, dateProvider.now)
+            diskCache.write(response)
+        }
+        return firstError
     }
 
     private suspend fun getRemoteConfig(appInBackground: Boolean): RemoteConfigResponse =
@@ -74,6 +94,8 @@ internal class RemoteConfigManager(
         }
 
     private data class TopicTask(val topic: Topic, val entryId: String, val entry: TopicEntry)
+
+    private data class CacheEntry(val response: RemoteConfigResponse, val cachedAt: Date)
 
     private companion object {
         const val DEFAULT_ENTRY_ID = "default"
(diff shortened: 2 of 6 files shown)


tonidero:
* **#3457** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3457?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3457?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ajpallares (commented):
I didn't take a deep look yet, but an idea came to mind as I was starting to review this. Would it make sense to try to reuse, somehow, the response that the ETagManager already caches? (maybe based on `ETagData.lastRefreshTime` for the TTL gate). In fact, with the current changes in this PR, wouldn't we be caching the response of the Config endpoint twice (new custom cache + ETagManager's cache)?

Perhaps it's a crazy idea, but may be worth considering

tonidero:
Not a crazy idea at all! We are indeed caching the offerings response twice right now for example 😅 (etag + offerings cache).

I think you have a point and we could totally do that... But it might be a bigger refactor/behavior change than this approach... As in, the etag cache will become used by things other than networking so we would need to rewire our dependencies... feels cleaner to me to do it separately. Especially considering this response should never actually be too big. Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -16,21 +20,31 @@ import kotlinx.coroutines.awaitAll
...
+    private var cache: CacheEntry? = null
+
     fun updateRemoteConfigIfNeeded(
         appInBackground: Boolean,
         completion: ((PurchasesError?) -> Unit)? = null,
     ) {
+        if (!cache?.cachedAt.isCacheStale(appInBackground, dateProvider)) {
+            completion?.invoke(null)
+            return
+        }
```
ajpallares, line 44: If two callers hit `updateRemoteConfigIfNeeded` while the cache is empty (or stale), both would pass this staleness check and both would enter `scope.launch { refresh(...) }`.

Should we somehow keep track of the current `refresh` task to avoid this kind of race conditions?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
```diff
@@ -0,0 +1,51 @@
...
+ * `noBackupFilesDir/RevenueCat/remote_config.json`.
+ *
+ * This is currently write-only. A subsequent PR will read this file at startup to seed the
+ * in-memory cache before the first network refresh.
+ */
+internal class RemoteConfigDiskCache(
+    private val applicationContext: Context,
+    private val json: Json,
+) {
+    fun write(response: RemoteConfigResponse) {
```
ajpallares, line 23: Not a problem now, but should we add a doc saying something like
> this performs blocking I/O; call from a background dispatcher

?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: We could also write a test for this edge case

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
```diff
@@ -0,0 +1,51 @@
...
+        val parent = File(applicationContext.noBackupFilesDir, REMOTE_CONFIG_ROOT)
+        try {
+            if (!parent.exists()) {
+                parent.mkdirs()
+            }
+            val target = File(parent, REMOTE_CONFIG_FILE_NAME)
```
ajpallares, line 29: Perhaps it's overkill, but should we skip `diskCache.write(response)` when `cache?.response == response`?

Since there's a 5 min foreground TTL, in steady state we'll write `remote_config.json` ~12 times per hour even when the response is byte-identical.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -50,7 +64,7 @@ internal class RemoteConfigManager(
             val entry = entries[DEFAULT_ENTRY_ID] ?: return@mapNotNull null
             TopicTask(topic, DEFAULT_ENTRY_ID, entry)
         }
-        return if (source == null || tasks.isEmpty()) {
+        val firstError: PurchasesError? = if (source == null || tasks.isEmpty()) {
```
ajpallares, line 64: Maybe this could be addressed in a follow-up PR, but I wonder if we should become more restrictive here about when we do cache the response (i.e. when we return an error). I fear we could be caching invalid/incomplete data.

ajpallares (commented):
I think this looks great! I have one small concern about concurrency calls in `updateRemoteConfigIfNeeded`. And some other minor comments

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3457/commits/79b773f894a997ce04fb0bc278c36449824902c5

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDiskCache.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3457/commits/79b773f894a997ce04fb0bc278c36449824902c5

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: I added some [checks](https://github.com/RevenueCat/purchases-android/pull/3457/commits/79b773f894a997ce04fb0bc278c36449824902c5) so we don't cache when the config is empty... But yeah, we can revisit this in follow-up PRs. Let me know if you have other ideas!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -1,35 +1,68 @@
...
+                } finally {
+                    synchronized(refreshLock) { inFlightRefresh = null }
+                }
+            }.also { inFlightRefresh = it }
+        }
+        // LAZY + explicit start: prevents the body's `finally` from clearing `inFlightRefresh`
+        // before the `.also` assignment has stored it (possible with a synchronous dispatcher).
+        deferred.start()
+        if (completion != null) {
+            scope.launch { completion.invoke(deferred.await()) }
```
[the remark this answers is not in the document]
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3457/commits/02d493a0919ebde2370c28cf27161447117cb85f. It's not a real solution though since it's not the caller thread, but the main thread... But I think it works for now. We can revisit once we're actually using this. I don't think it should matter for our uses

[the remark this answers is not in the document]
tonidero: Ended up reverting it... I had to fix some other things to make this work, and I don't think it's worth it until we actually end up using this. We can change this later if needed.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -1,35 +1,70 @@
...
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
+import java.util.Date
 
+@OptIn(InternalRevenueCatAPI::class)
```
ajpallares, line 24: Is this needed?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -62,6 +97,16 @@ internal class RemoteConfigManager(
...
         }
+        // Only cache when at least one topic was actually fetched — empty sources, empty topics,
+        // and missing default entryIds are treated as no-op refreshes that don't populate the cache.
+        if (firstError == null && source != null && tasks.isNotEmpty()) {
+            val previousResponse = cache?.response
+            cache = CacheEntry(response, dateProvider.now)
+            if (previousResponse != response) {
+                diskCache.write(response)
+            }
+        }
```
ajpallares, line 88: If we get a null source of empty tasks, we would be keeping the existing cache. Is that correct? (I'm guessing it's better to have a stale cached rather than no cache, but want to be sure)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -1,35 +1,70 @@
...
+                try {
+                    refresh(appInBackground)
+                } finally {
+                    synchronized(refreshLock) { inFlightRefresh = null }
+                }
+            }.also { inFlightRefresh = it }
+        }
+        // LAZY + explicit start: prevents the body's `finally` from clearing `inFlightRefresh`
+        // before the `.also` assignment has stored it (possible with a synchronous dispatcher).
+        deferred.start()
```
ajpallares: If I'm not mistaken, when `appInBackground == true`, there's some jitter to the call. Could this deduplication logic mean that a subsequent foreground call to the config may get stalled by the same call triggered while `appInBackground`? 🤔 

Perhaps not very likely though

ajpallares (commented):
I have some more comments. 

Also, regarding the PR description:
> This caches the remote config response both on disk and in memory. Right now the in memory cache is only used to avoid fetching the config multiple times without need.

I believe the memory cache is not used yet in this PR either (since the `RemoteConfigManager` is not yet used anywhere)?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: It is since we use some internal APIs, for example `DateProvider`.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: I believe so yeah... I would say, a config without tasks feels like an error, so we might as well ignore it and keep what we have if any.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: That's indeed true, good catch!

tonidero: Ended up [removing](https://github.com/RevenueCat/purchases-android/pull/3457/commits/8f91f3d8ebf0d4d8a9f9200218ca9dc4e4909bf2) the dedupe logic here and relying on the backend dedupe itself

tonidero:
> I believe the memory cache is not used yet in this PR either (since the RemoteConfigManager is not yet used anywhere)?

This is indeed correct in this PR, but it's used in the final PRs in this stack which actually use the RemoteConfigManager

ajpallares (approved):
Great job! I think this looks good!

PR #3458, opened 2026-05-08
Select blob source by priority and weighted random

tonidero:
### Motivation
`RemoteConfigManager.refresh` was hardcoded to pick `blobSources.firstOrNull()` with a `WIP:` comment. The backend already returns `priority` and `weight` on each source, so this PR wires up the real selection logic.

### Description
- Highest `priority` value wins.
- Among sources tied at the top priority, pick one with weighted random using `weight`:
  ```
  total = sum(weights)
  r = random(0, total)
  cumulative = 0
  for source in candidates:
      cumulative += source.weight
      if r < cumulative: return source
  ```
- Fallback to uniform random when all top-priority weights are zero (or any of them are negative, defensively).
- Selection extracted into a `WeightedSource` interface + `List<T>.selectWeighted(Random)` extension so the same algorithm can later be reused for `ApiSource` (which has the same `priority`/`weight` shape).
- `RemoteConfigManager` now takes an injectable `Random` (defaulted to `Random.Default`) for deterministic unit tests.

### Test plan
- [x] New `WeightedSourceSelectionTest` covers: empty list, single source with zero weight, priority dominance, weighted-pick boundary cases (`r=0`, `r=29`, `r=30`, `r=99` for weights `[30, 70]`), zero-weight uniform fallback, negative-weight uniform fallback.
- [x] `RemoteConfigManagerTest` updated: existing multi-source test relaxed to assert delegation; new test asserts the highest-priority source is selected through the manager.
- [x] `./gradlew :purchases:testDefaultsBc8DebugUnitTest --tests "com.revenuecat.purchases.common.remoteconfig.*"` passes.
- [x] `./gradlew detektAll` passes.

---

> [!NOTE]
> **Medium Risk**
> Changes remote-config blob source selection from deterministic first-item to priority/weight-based randomness, which can affect which CDN/source is used in production. Risk is moderate but bounded to remote config fetching and covered by new unit tests and injectable RNG for determinism.
> 
> **Overview**
> Remote config blob fetching now selects a `BlobSource` by **highest `priority`**, and breaks ties with **weighted random** using `weight` (with a uniform-random fallback for zero/negative weights) rather than always using `blobSources.firstOrNull()`.
> 
> This extracts the algorithm into a reusable `WeightedSource` + `selectWeighted(...)` helper, updates `ApiSource`/`BlobSource` to implement it, injects `Random` into `RemoteConfigManager` for deterministic tests, and expands unit coverage for priority/weight edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98ba25479bee1cd7355d70dedb90274c77c9a7ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -20,13 +20,16 @@ import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import java.util.Date
+import kotlin.random.Random
 
 @OptIn(InternalRevenueCatAPI::class)
+@Suppress("LongParameterList")
 internal class RemoteConfigManager(
     private val backend: Backend,
     private val topicFetcher: TopicFetcher,
     private val diskCache: RemoteConfigDiskCache,
     private val dateProvider: DateProvider = DefaultDateProvider(),
+    private val random: Random = Random.Default,
     dispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) {
     private val scope = CoroutineScope(SupervisorJob() + dispatcher)
@@ -55,8 +58,7 @@ internal class RemoteConfigManager(
             errorLog { "Failed to fetch remote config: ${e.error}" }
             return e.error
         }
-        // WIP: We should have some logic to pick the correct source for this. Right now, hardcoded to the first source.
-        val source = response.blobSources.firstOrNull()
+        val source = response.blobSources.selectWeighted(random)
         val referenced = buildReferenceSet(response.manifest)
         val tasks = response.manifest.topics.mapNotNull { (topic, entries) ->
             val entry = entries[DEFAULT_ENTRY_ID] ?: return@mapNotNull null
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigResponse.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigResponse.kt
@@ -20,17 +20,17 @@ internal data class RemoteConfigResponse(
 internal data class ApiSource(
     val id: String,
     val url: String,
-    val priority: Int,
-    val weight: Int,
-)
+    override val priority: Int,
+    override val weight: Int,
+) : WeightedSource
 
 @Serializable
 internal data class BlobSource(
     val id: String,
     @SerialName("url_format") val urlFormat: String,
-    val priority: Int,
-    val weight: Int,
-)
+    override val priority: Int,
+    override val weight: Int,
+) : WeightedSource
 
 @Serializable
 internal data class Manifest(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
@@ -1,0 +1,31 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+import kotlin.random.Random
+
+internal interface WeightedSource {
+    val priority: Int
+    val weight: Int
+}
+
+/**
+ * Helper function to select a source from a list of [WeightedSource]s
+ * - Highest priority
+ * - Tied priority uses weights to choose randomly.
+ * - If weights add up to 0 or any negative weight, uniform random selection.
+ */
+internal fun <T : WeightedSource> List<T>.selectWeighted(random: Random = Random.Default): T? {
+    val highestPriority = maxOfOrNull { it.priority } ?: return null
+    val candidates = filter { it.priority == highestPriority }
+    val totalWeight = candidates.sumOf { it.weight }
+    val anyNegativeWeights = candidates.any { it.weight < 0 }
+    return if (totalWeight <= 0 || anyNegativeWeights) {
+        candidates[random.nextInt(candidates.size)]
+    } else {
+        val r = random.nextInt(totalWeight)
+        var cumulative = 0
+        candidates.first { candidate ->
+            cumulative += candidate.weight
+            r < cumulative
+        }
+    }
+}
--- a/purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelectionTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelectionTest.kt
@@ -1,0 +1,111 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import kotlin.random.Random
+
+class WeightedSourceSelectionTest {
+
+    @Test
+    fun `empty list returns null`() {
+        val sources: List<TestSource> = emptyList()
+        assertThat(sources.selectWeighted(Random.Default)).isNull()
+    }
+
+    @Test
+    fun `single source with positive weight is selected`() {
+        val only = TestSource(priority = 0, weight = 10)
+        assertThat(listOf(only).selectWeighted(FakeRandom(0))).isSameAs(only)
+        assertThat(listOf(only).selectWeighted(FakeRandom(1))).isSameAs(only)
+    }
+
+    @Test
+    fun `single source with zero weight is still selected`() {
+        val only = TestSource(priority = 0, weight = 0)
+        assertThat(listOf(only).selectWeighted(FakeRandom(0))).isSameAs(only)
+    }
+
+    @Test
+    fun `highest priority wins regardless of weight`() {
+        val low = TestSource(priority = 1, weight = 999)
+        val high = TestSource(priority = 5, weight = 1)
+        assertThat(listOf(low, high).selectWeighted(FakeRandom(0))).isSameAs(high)
+    }
+
+    @Test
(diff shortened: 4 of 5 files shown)


tonidero:
* **#3458** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3458?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3458?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
```diff
@@ -0,0 +1,24 @@
...
+    val priority: Int
+    val weight: Int
+}
+
+internal fun <T : WeightedSource> List<T>.selectWeighted(random: Random = Random.Default): T? {
+    val highestPriority = maxOfOrNull { it.priority } ?: return null
+    val candidates = filter { it.priority == highestPriority }
+    val totalWeight = candidates.sumOf { it.weight }
+    return if (totalWeight <= 0) {
+        candidates[random.nextInt(candidates.size)]
```
ajpallares: The PR description claims
> Falls back to uniform random when all top-priority weights are zero (or negative, defensively)

But this implementation means that this fallback only happens when the **sum** of the weights is non-positive. 

Since it is not expected to have any non-positive weights (correct me if I'm wrong), maybe it would be simpler to fall back to uniform random when **any** top-priority is ≤ 0. Otherwise, dealing with positive and zero/negative values could get tricky. 
WDYT?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
```diff
@@ -0,0 +1,24 @@
...
+    } else {
+        val r = random.nextInt(totalWeight)
+        var cumulative = 0
+        candidates.firstOrNull { candidate ->
+            cumulative += candidate.weight
+            r < cumulative
+        } ?: candidates.last()
```
ajpallares: If at this point `totalWeight > 0`, that means candidates is not empty, so we could just do

```suggestion
        candidates.first { candidate ->
            cumulative += candidate.weight
            r < cumulative
        }
```
?

ajpallares (commented):
Some initial comments on the weight-based randomization logic. Will keep looking 👀

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
tonidero: Hmm it's a valid point indeed... Will change it to that. As you said this is unexpected, but good to be defensive 👍

tonidero: Fixed!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
tonidero: Fixed!

purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelectionTest.kt
```diff
@@ -0,0 +1,104 @@
...
+class WeightedSourceSelectionTest {
+
+    @Test
+    fun `empty list returns null`() {
+        val sources: List<TestSource> = emptyList()
+        assertThat(sources.selectWeighted(Random.Default)).isNull()
+    }
+
+    @Test
+    fun `single source with zero weight is still selected`() {
```
ajpallares, line 23: One more test suggestion: a single source with a positive weight. Adding `single source with positive weight is selected` would also pin down `nextInt(1)` / `r=0 < cumulative=1` in the weighted branch.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigResponse.kt
```diff
@@ -20,17 +20,17 @@ internal data class RemoteConfigResponse(
 internal data class ApiSource(
     val id: String,
     val url: String,
-    val priority: Int,
-    val weight: Int,
-)
+    override val priority: Int,
+    override val weight: Int,
+) : WeightedSource
```
ajpallares, line 25: NABD, but I believe in this PR we're not actually using the conformance of `ApiSource` to `WeightedSource`

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -55,8 +58,7 @@ internal class RemoteConfigManager(
             errorLog { "Failed to fetch remote config: ${e.error}" }
             return e.error
         }
-        // WIP: We should have some logic to pick the correct source for this. Right now, hardcoded to the first source.
-        val source = response.blobSources.firstOrNull()
+        val source = response.blobSources.selectWeighted(random)
```
ajpallares, line 61: I'm guessing this will be tackled in follow-up PRs, but, with the current implementation of this PR, we only ever use the highest priority `blobSources`, right? lower-priority sources aren't fallback candidates.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
```diff
@@ -0,0 +1,25 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+import kotlin.random.Random
+
+internal interface WeightedSource {
+    val priority: Int
+    val weight: Int
+}
+
+internal fun <T : WeightedSource> List<T>.selectWeighted(random: Random = Random.Default): T? {
```
ajpallares, line 16: Should we add some documentation on how this works?
1. higher `priority` wins
2. when the uniform fallback fires, including the defensive behavior for `weight` (any negative value disables weighted selection across the candidates)

The current behavior could be non-obvious enough that the next contributor will have to read the tests to figure it out.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
```diff
@@ -0,0 +1,25 @@
...
+    val highestPriority = maxOfOrNull { it.priority } ?: return null
+    val candidates = filter { it.priority == highestPriority }
+    val totalWeight = candidates.sumOf { it.weight }
+    val anyNegativeWeights = candidates.any { it.weight < 0 }
+    return if (totalWeight <= 0 || anyNegativeWeights) {
+        candidates[random.nextInt(candidates.size)]
```
ajpallares, line 22: Cursor highlighted some concerns about thread safety in `random`. Seems like the default `Random.Default` is thread safe, so this would only really affect tests I guess. Just wanted to highlight here as well

ajpallares (commented):
I think it looks good! I have some doubts though, mostly about whether the lower-priority blob sources not being picked at all for now.
Great job btw!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: That's indeed correct! That's tackled in this follow-up PR: https://github.com/RevenueCat/purchases-android/pull/3466

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigResponse.kt
tonidero: Correct! The idea is that we can reuse the same logic to select the ApiSource, but that's not implemented yet (will be in future PRs)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
tonidero: Added some docs in https://github.com/RevenueCat/purchases-android/pull/3458/commits/9d28ec15a4a84fa4efd13783195b3af685fc8977

purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelectionTest.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3458/commits/9d28ec15a4a84fa4efd13783195b3af685fc8977

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
tonidero: Hmm can you elaborate? I think for tests as well, we create new instances on each test. It's true we don't test with the "actual" `Random.Default`, but this is to make the tests deterministic.

ajpallares: You're right, yes. With the uses we have for it now, it's not a problem. Please ignore me

ajpallares (approved):
I think this looks great!

PR #3462, opened 2026-05-11
Add AdMob Reward Verification API facade

polmiro:
### Checklist

### Motivation
This adds the Android API facade for AdMob reward verification so we can align the SDK surface with iOS before stacking implementation details.

### Description
This PR introduces the reward verification API shape in `feature/admob` without implementing runtime behavior yet.

Changes included:
- Added new experimental extensions for rewarded and rewarded interstitial ads:
  - `enableRewardVerification()`
  - verification `show(...)` overloads
- Added experimental reward verification models:
  - `RewardVerificationResult`
  - `VerifiedReward` (with value semantics for virtual currency payloads)
- Updated generated API signatures in `feature/admob/api.txt`
- Updated `feature/admob/README.md` with reward-verification usage examples based on `loadAndTrack...` helpers
- Verification `show(...)` requires `rewardVerificationResult`
- Documented callback thread contract in KDoc

---

> [!NOTE]
> **Medium Risk**
> Adds new experimental public API surface for rewarded ad verification that currently throws `NotImplementedError`, so adopters could hit runtime failures if used. Changes are otherwise isolated to the `feature/admob` module and documentation.
> 
> **Overview**
> Introduces an **experimental AdMob reward verification facade** for rewarded and rewarded-interstitial ads, including `enableRewardVerification()` and new `show(...)` overloads that return a `RewardVerificationResult`.
> 
> Adds new verification models (`RewardVerificationResult`, `VerifiedReward` with a `@Poko` `VirtualCurrency` payload and sentinel types) and wires them into the published surface (`feature/admob/api.txt`) plus module build config (adds the `poko` plugin).
> 
> Updates `feature/admob/README.md` with reward-verification usage examples and refreshes the supported-formats table.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a56b266d5e98ad7916ef3cbe7419d95f255393cb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Reward payload delivered when server-side reward verification succeeds.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+public sealed interface VerifiedReward {
+
+    /**
+     * A virtual-currency reward with a code and amount.
+     */
+    public data class VirtualCurrency(
```
tonidero: We try to avoid using data classes in public APIs. [context](https://jakewharton.com/public-api-challenges-in-kotlin/). Instead, we should use a normal class annotated with `@Poko` for these.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
```diff
@@ -0,0 +1,28 @@
+package com.revenuecat.purchases.admob
+
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+
+/**
+ * Reward payload delivered when server-side reward verification succeeds.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+public sealed interface VerifiedReward {
```
tonidero: Also, we avoid adding "sealed" (either interfaces or classes) to the public APIs. Alternative is to create an interface/abstract class, and have each of the elements implement that. 

This has downsides (can't switch-case on the possible results, can't get compile time safety, and developers could potentially implement their own subclasses/implementations). Still, using sealed means that adding new cases would cause breaking changes, so we try to avoid this.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationResult.kt
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Non-null when verification succeeded.
+     */
+    public val verifiedReward: VerifiedReward?
+        get() = (this.storage as? Storage.Verified)?.reward
+
+    /**
+     * True when verification did not complete successfully.
+     */
+    public val isFailed: Boolean
```
tonidero: `isFailed` sounds weird... Maybe just `failed`?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationResult.kt
```diff
@@ -0,0 +1,46 @@
...
+        get() = this.storage is Storage.Failed
+
+    public companion object {
+
+        /**
+         * Server verification succeeded for this ad transaction.
+         */
+        @JvmStatic
+        public fun verified(reward: VerifiedReward): RewardVerificationResult {
```
tonidero: Hmm I'm unsure of how these will be used, but do these need to be public?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+import kotlin.jvm.JvmSynthetic
+
+/**
+ * Enables RevenueCat reward verification for this ad.
+ *
+ * Call after the ad has loaded and before showing it when using reward-verification APIs.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@JvmSynthetic
+public fun RewardedAd.enableRewardVerification() {
```
tonidero, line 16: I wonder... could these just be a optional "enableRewardVerification" parameter when loading the ad (instead of a separate method) and we can just enable it ourselves in that case?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Shows a rewarded ad with optional reward-verification callbacks.
+ *
+ * When [rewardVerificationResult] is provided, [enableRewardVerification] must be called first.
+ * Callback invocations are delivered on the main thread.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@JvmSynthetic
+public fun RewardedAd.show(
```
tonidero, line 41: I'm wondering, should we hold these until the rest of the implementation is ready? Seems weird to have a public API even if experimental that does not work at all... Maybe we can just change these to `internal` temporarily, until we're ready to go public?

tonidero (commented):
Mostly, I think we should avoid using data classes and sealed classes/interfaces. Other than that, just some comments. Thanks for doing this!

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
polmiro: I remember now, I'll make those updates now. I'll make those updates now.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationResult.kt
polmiro: I think this got derived from the iOS implementation. I'll update :)

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
polmiro: Ah! My plan was to make this PR point to a `admob-reward-verification` branch. I made a mistake when creating the PR. Thanks for the heads up!

tonidero: Ah perfect :)

polmiro: Just updated updated the base branch of the PR.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationResult.kt
polmiro: Good point. The will be used internally before returning the result in the public API so I can adjust visibilty.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
polmiro: One advantage of the `enableRewardVerification` method 3 is that it lives exactly where the custom SSV configuration would live, which may help making mistakes and/or potentially overriding our own settings. I think there’s value too in keeping the API as close as possible to AdMob’s given we are extending AdMob’s types too.  This is the main reason I went for this option.

Notice that iOS already follows this pattern to given that the AdMob iOS SDK is shaped in a very similar way. The PRs for reward verification were recently merged. They are still experimental so we could course-correct. Just bringing it up since there is also value in being consistent across purchases SDKs too.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationResult.kt
```diff
@@ -0,0 +1,46 @@
...
+     * Non-null when verification succeeded.
+     */
+    public val verifiedReward: VerifiedReward?
+        get() = (this.storage as? Storage.Verified)?.reward
+
+    /**
+     * True when verification did not complete successfully.
+     */
+    public val failed: Boolean
+        get() = this.storage is Storage.Failed
```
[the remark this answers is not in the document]
polmiro: @tonidero perhaps we should have stayed the same? Are you okay with me reverting this change?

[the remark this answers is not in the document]
tonidero: Hmm TBH, I think that use case `RewardVerificationResult.failed.failed` seems quite awkward, since it basically means you already know it failed, so why would you check it failed? 😅 

Still, it's a valid point... TBH, I would say no action is needed IMO , or as they suggest, renaming the result type to `failure` could also work... I just think `failed` is preferred than doing a more common `rewardVerificationResult.isFailed` vs `rewardVerificationResult.failed`... Not a strong opinion though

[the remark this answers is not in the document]
polmiro: I'm good with either option so I will leave it as is. I agree it's awkward to want to follow that path, but it's true it's a close overlap 🤷‍♂️

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import kotlin.jvm.JvmSynthetic
+
+/**
+ * Enables RevenueCat reward verification for this ad.
+ *
+ * Call after the ad has loaded and before showing it when using reward-verification APIs.
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
```
tonidero: Hmm my guess is that we don't need to opt in to this annotation if we're propagating it with the annotation itself? Same for the other extensions below

tonidero (approved):
Looks great!

PR #3464, opened 2026-05-12
Add reward verification status polling endpoint plumbing

polmiro:
### Checklist

### Motivation
This adds the core status polling endpoint plumbing in `:purchases`, so adapter-level polling behavior can be stacked in subsequent PRs.

### Description
This change adds end-to-end SDK wiring for fetching reward verification status by `clientTransactionId`:

- Adds `Endpoint.GetRewardVerificationStatus`.
- Adds `Backend.getRewardVerificationStatus(...)` with response parsing and deduped callback handling.
- Adds internal status/wire models:
  - `RewardVerificationStatus`
  - `RewardVerificationStatusResponse`
- Exposes internal entry points through:
  - `PurchasesOrchestrator`
  - `Purchases` (defaults/internal)
  - `awaitGetRewardVerificationStatus(...)` coroutine extension
  - `GetRewardVerificationStatusCallback` (internal)

Test coverage includes:
- Endpoint path generation.
- Response status mapping.
- Backend success/error behavior and deduping.
- Purchases forwarding behavior.

---

> [!NOTE]
> **Medium Risk**
> Adds a new backend API call and response parsing with callback de-duping; issues would primarily surface as incorrect status mapping or request caching/dispatch behavior for this new endpoint.
> 
> **Overview**
> Adds end-to-end internal plumbing to fetch ad reward verification status by `clientTransactionId`.
> 
> Introduces a new `Endpoint.GetRewardVerificationStatus` and `Backend.getRewardVerificationStatus` (with callback de-duping and JSON parsing into the new `RewardVerificationStatus` enum), then wires it through `PurchasesOrchestrator` and `Purchases`, including a coroutine wrapper `awaitGetRewardVerificationStatus`.
> 
> Includes unit tests covering endpoint path/name, response-to-enum mapping (including unknown statuses), backend success/error handling, and concurrent-call de-duping plus forwarding from `Purchases`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1e9852a501d03befa5791e9dac83b827061882e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -711,6 +712,17 @@ public class Purchases internal constructor(
         purchasesOrchestrator.createSupportTicket(email, description, onSuccess, onError)
     }
 
+    @InternalRevenueCatAPI
+    public fun getAdMobRewardVerificationStatus(
```
tonidero: I would say, since these APIs are only meant to be `@InternalRevenueCatAPI`, let's only expose one (either the coroutine version and keep this one `internal`, or this callback version). No need to expose more I think.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/AdMobRewardVerificationStatusResponse.kt
```diff
@@ -0,0 +1,22 @@
+package com.revenuecat.purchases.common.networking
+
+import com.revenuecat.purchases.AdMobRewardVerificationStatus
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal data class AdMobRewardVerificationStatusResponse(
+    @SerialName("status")
```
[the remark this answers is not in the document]
tonidero: This seems true, no need to make it explicit when there are no underscores in the field.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/AdMobRewardVerificationStatusResponse.kt
```diff
@@ -0,0 +1,22 @@
...
+    val status: String,
+) {
+    @OptIn(InternalRevenueCatAPI::class)
+    fun toAdMobRewardVerificationStatus(): AdMobRewardVerificationStatus {
+        return when (status.lowercase()) {
+            "pending" -> AdMobRewardVerificationStatus.PENDING
+            "verified" -> AdMobRewardVerificationStatus.VERIFIED
+            "failed" -> AdMobRewardVerificationStatus.FAILED
+            else -> AdMobRewardVerificationStatus.UNKNOWN
```
tonidero: Could be worth logging something as well... but NABD

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -104,6 +104,16 @@ internal sealed class Endpoint(
     ) {
         override fun getPath(useFallback: Boolean) = pathTemplate.format(Uri.encode(userId))
     }
+    data class GetAdMobRewardVerificationStatus(
+        val userId: String,
+        val clientTransactionId: String,
+    ) : Endpoint(
+        pathTemplate = "/v1/subscribers/%s/ads/reward_verifications/%s",
```
tonidero, line 111: Probably you have thought about this before, but just thinking... Would it be worth adding `AdMob` somewhere in this request? Just thinking whether this is something that we could reuse for other ad providers or not... But I'm totally fine keeping it as is, if we think we can reuse it :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -104,6 +104,16 @@ internal sealed class Endpoint(
...
+        name = "get_admob_reward_verification_status",
```
tonidero: [Not a blocker for this PR]
Just curious, but should this endpoint be supported in fortress and/or a fallback url or something like that? mostly thinking on what happens in the unlikely case that servers are down, since this seems like a somewhat critical path (though probably not the most critical)

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1114,6 +1124,73 @@ internal class Backend(
         }
     }
 
+    fun getAdMobRewardVerificationStatus(
+        appUserID: String,
+        clientTransactionId: String,
+        appInBackground: Boolean,
```
tonidero: Hmm I'm wondering... should this be a background-aware call? Just thinking that this seems like a request that should always happen only in foreground, so we might not care about this, and we should always perform it without delay? (might be missing the entire flow here, in which case, FFTI!)

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1114,6 +1124,73 @@ internal class Backend(
...
+                        }
+                    } else {
+                        onErrorHandler(result.toPurchasesError().also { errorLog(it) })
+                    }
+                }
+            }
+        }
+
+        synchronized(this@Backend) {
+            val delay = if (appInBackground) Delay.DEFAULT else Delay.NONE
```
tonidero: Similar to the above question, not sure if we would want to have some jittering for this request, since it seems it should be performed immediately?

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetAdMobRewardVerificationStatusCallback.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.revenuecat.purchases.AdMobRewardVerificationStatus
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.PurchasesError
+
+/**
+ * Interface to be implemented when making calls that return an [AdMobRewardVerificationStatus].
+ */
+@InternalRevenueCatAPI
+public interface GetAdMobRewardVerificationStatusCallback {
```
tonidero: If we make the callback API `internal`, we wouldn't need to expose this either, but depends on which one we want to use.

tonidero (commented):
Haven't reviewed tests yet. Just some nitpicks, but overall looking great!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
polmiro: Good call, makes sense. And if anything, better to add the parameter if the situation arises later and not prematurely.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
polmiro: It actually started having admob in the path. Then we realized that the piece that is actually coupled to AdMob is the loading piece (setting SSV options) and the webhook piece. But the polling is agnostic. But this brings the question if there are types in this PR that should have their AdMob prefix stripped! I'm going to take a pass.

polmiro: I just renamed and removed all references to AdMob in the PR. Thanks for the callout!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
polmiro: It's a good suggestion. It's something to think through since I could see several paths to follow like compensating actions after the possible outage. I will take a note and bring it up to the team!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
polmiro: yeah, we will for the polling loops but not here. Thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetRewardVerificationStatusCallback.kt
```diff
@@ -0,0 +1,21 @@
+package com.revenuecat.purchases.interfaces
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.RewardVerificationStatus
+
+/**
+ * Interface to be implemented when making calls that return a [RewardVerificationStatus].
+ */
+@InternalRevenueCatAPI
```
tonidero: No need to keep this annotation now that it's `internal` :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -711,6 +712,17 @@ public class Purchases internal constructor(
         purchasesOrchestrator.createSupportTicket(email, description, onSuccess, onError)
     }
 
+    @InternalRevenueCatAPI
```
tonidero: Same here, no need to keep this annotation now that it's internal.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -908,6 +909,19 @@ internal class PurchasesOrchestrator(
         )
     }
 
+    @InternalRevenueCatAPI
```
tonidero: Same here, there is no need for this annotation here since this entire class is internal.

tonidero (approved):
Looks great!

purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetRewardVerificationStatusCallback.kt
polmiro: oh, I was not sure and stayed on the safer side but makes sense.

PR #3466, opened 2026-05-12
Fall back across blob sources when topic downloads fail

tonidero:
### Motivation

`RemoteConfigManager` previously picked a single blob source per refresh and gave up if any topic failed to download from it. With weighted multi-source distribution in place (#previous stack PR), we want to gracefully fall back to other sources when one is degraded so a refresh can still succeed.

### Description

- **Classify topic-fetch outcomes** (`TopicFetcher.kt`): `fetchTopicIfNeeded` now returns a `TopicFetchResult` sealed type — `Success`, `TransientFailure` (timeouts / no network — worth retrying against a different source), or `InvalidatingFailure` (malformed blob ref, checksum mismatch, other IO). Transient/invalidating is decided via `ConnectionErrorReason.fromIOException`.
- **Fallback loop in `RemoteConfigManager.kt`**: introduces `downloadTopicsWithFallback`, which repeatedly picks the next weighted source, attempts the still-failing tasks against it, and stops as soon as everything succeeds or sources are exhausted. The most recent source's error is preserved for the completion callback so callers see the freshest failure reason rather than the first one encountered.
- **Sticky source selection**: tracks the source that produced the last successful refresh in `cachedSourceId`. On the next refresh, that source is tried first (bypassing weighted selection) to keep clients on a known-good source. If the sticky source returns an invalidating failure, the cached id is cleared so a fresh weighted draw can take over next time.
- **Pin reset on response change**: when the backend returns a `RemoteConfigResponse` different from the one currently cached (e.g. an updated source list, new priorities, or new weights), `cachedSourceId` is cleared before the fallback loop runs. This guarantees a new top-priority source can take over instead of the client staying pinned to a now-lower-priority cached source.
- **`selectWeightedExcluding` helper** (`WeightedSource.kt`): weighted selection filtered to sources not already tried in the current refresh. Negative weights are clamped to zero so a single misconfigured entry can't poison the picker.
- **Refactored refresh result** into a `RefreshOutcome` sealed class (`Success` / `VacuousSuccess` / `Failure`) so caching, cleanup, and cached-source clearing are driven by a single decision rather than nested null checks.

### Tests

- `RemoteConfigManagerTest`: new coverage for cross-source fallback on transient failures, exhaustion behavior, invalidating-failure short-circuit per task, sticky-source preference on subsequent refreshes, `cachedSourceId` invalidation when the sticky source fails with an invalidating error (rewritten with a staged `QueuedRandom` so the test actually distinguishes cleared/not-cleared paths), and `cachedSourceId` reset when the backend returns a new response.
- `TopicFetcherTest`: updated for the new `TopicFetchResult` return type, including timeout/no-network → `TransientFailure` and checksum/malformed/other-IO → `InvalidatingFailure` classification.
- `WeightedSourceSelectionTest`: new tests for `selectWeightedExcluding` and for negative-weight clamping behavior.

### Checklist
- [x] Unit tests

---

> [!NOTE]
> **Medium Risk**
> Changes the remote-config refresh path to retry topic downloads across multiple blob sources with sticky source selection, which can affect config update reliability and caching behavior under network failures.
> 
> **Overview**
> Remote config refresh now **falls back across blob sources** when topic downloads fail, retrying only the still-failing topics until one source succeeds or all sources are exhausted; the completion error now reflects the latest failure.
> 
> `TopicFetcher.fetchTopicIfNeeded` now returns a `TopicFetchResult` (`Success`, `TransientFailure`, `InvalidatingFailure`) with error classification driven by `ConnectionErrorReason`, and `RemoteConfigManager` tracks a **sticky** `cachedSourceId` (prefer last-good source, clear it on invalidating failures or when the backend response changes). `WeightedSource` adds `selectWeightedExcluding` to avoid re-trying sources within a refresh, and tests are expanded/updated to cover fallback, stickiness/reset, and new result classifications.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4650c7119398cce612b29f4d18e69fd130fe7b04. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -62,6 +62,10 @@ internal class RemoteConfigManager(
             errorLog { "Failed to fetch remote config: ${e.error}" }
             return e.error
         }
+        val previouslyCachedResponse = cache?.response
+        if (previouslyCachedResponse != null && previouslyCachedResponse != response) {
+            cachedSourceId = null
+        }
         val referenced = buildReferenceSet(response.manifest)
         val tasks = response.manifest.topics.mapNotNull { (topic, entries) ->
             val entry = entries[DEFAULT_ENTRY_ID] ?: return@mapNotNull null
@@ -109,7 +113,7 @@ internal class RemoteConfigManager(
             triedIds += source.id
 
             val attempt = attemptDownloads(source, remaining)
-            lastError = lastError ?: attempt.firstError
+            lastError = attempt.firstError ?: lastError
             if (attempt.invalidating && source.id == previouslyCachedSourceId) {
                 cachedSourceFailedInvalidating = true
             }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -2,6 +2,7 @@ package com.revenuecat.purchases.common.remoteconfig
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.PurchasesErrorCode
 import com.revenuecat.purchases.PurchasesException
 import com.revenuecat.purchases.common.Backend
 import com.revenuecat.purchases.common.DateProvider
@@ -37,6 +38,9 @@ internal class RemoteConfigManager(
     @Volatile
     private var cache: CacheEntry? = null
 
+    @Volatile
+    private var cachedSourceId: String? = null
+
     fun updateRemoteConfigIfNeeded(
         appInBackground: Boolean,
         completion: ((PurchasesError?) -> Unit)? = null,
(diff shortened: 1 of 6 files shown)


tonidero:
* **#3466** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3466?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
```diff
@@ -29,3 +29,9 @@ internal fun <T : WeightedSource> List<T>.selectWeighted(random: Random = Random
         }
     }
 }
+
+internal fun <T : WeightedSource> List<T>.selectWeightedExcluding(
+    excludedIds: Set<String>,
+    idOf: (T) -> String,
```
ajpallares, line 35: Would it be cleaner to add `id: String` to the `WeightedSource` interface instead?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -58,41 +62,122 @@ internal class RemoteConfigManager(
...
-                        topicFetcher.fetchTopicIfNeeded(
-                            topic = task.topic,
-                            entryId = task.entryId,
-                            topicEntry = task.entry,
-                            source = source,
-                        )
-                    }
-                }.awaitAll().firstNotNullOfOrNull { it }
-            }
+            downloadTopicsWithFallback(response.blobSources, tasks)
```
ajpallares, line 77: This is a HUGE nit, but we could defer the calculation of `tasks` to only when `response.blobSources` is not empty (that would avoid some unnecessary work). Total nit as I said, so FFTI.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/TopicFetcher.kt
```diff
@@ -30,11 +37,13 @@ internal class TopicFetcher(
...
         source: BlobSource,
-    ): PurchasesError? {
+    ): TopicFetchResult {
         if (!topicEntry.blobRef.matches(BLOB_REF_PATTERN)) {
             val message = "Topic $topic ($entryId) has a malformed blob_ref; refusing to fetch."
             errorLog { message }
-            return PurchasesError(PurchasesErrorCode.UnexpectedBackendResponseError, message)
+            return TopicFetchResult.InvalidatingFailure(
+                PurchasesError(PurchasesErrorCode.UnexpectedBackendResponseError, message),
+            )
```
ajpallares, line 46: I think this is incorrect. Returning `InvalidatingFailure` here makes it so that the **source** is invalidated, whereas it's the blob in the topic what's invalid. In addition, since this blobRef would be the same for all sources, this would effectively mean that all sources would be invalidated **for all** topics only because one blobRef is invalid.

First, to avoid these confusions, I would rename `InvalidatingFailure` to `SourceInvalidatingFailure`, meaning the source is invalid. Then, for this kind of error (where the blob is invalid regardless of the source) I would introduce a new `TopicFetchResult` meaning "this task is bad regardless of the source", maybe calling it `TaskInvalid`.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -58,41 +62,122 @@ internal class RemoteConfigManager(
...
+        tasks: List<TopicTask>,
+    ): RefreshOutcome {
+        val triedIds = mutableSetOf<String>()
+        var remaining = tasks
+        var lastError: PurchasesError? = null
+        var cachedSourceFailedInvalidating = false
+        val previouslyCachedSourceId = cachedSourceId
+
+        while (remaining.isNotEmpty()) {
+            val source = pickNextSource(blobSources, triedIds, previouslyCachedSourceId) ?: break
```
ajpallares, line 112: Should we add any logs when using a fallback? when we switch sources, when we fall back from the cached source, or when we exhaust all sources.  `TopicFetcher` logs individual fetch failures but the manager's _decisions_ are invisible.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -37,6 +38,9 @@ internal class RemoteConfigManager(
     @Volatile
     private var cache: CacheEntry? = null
 
+    @Volatile
+    private var cachedSourceId: String? = null
+
     fun updateRemoteConfigIfNeeded(
```
ajpallares, line 44: I think I raised this in a previous PR as well, so I'm sorry if I'm repeating myself. But I worry about race conditions here. 

If caller A and caller B both invoke `updateRemoteConfigIfNeeded`:

Thread Main:
```
A: updateRemoteConfigIfNeeded()
   → cache is stale ✓
   → scope.launch { refresh() }   ← returns immediately, coroutine A queued on IO pool
B: updateRemoteConfigIfNeeded()
   → cache is stale ✓ (A hasn't written cache yet)
   → scope.launch { refresh() }   ← returns immediately, coroutine B queued on IO pool
```

**Then in parallel:**

| IO Thread 1 (coroutine A)         | IO Thread 2 (coroutine B)                 |
| --------------------------------- | ----------------------------------------- |
| `refresh()`                       | `refresh()`                               |
| `getRemoteConfig()` …             | `getRemoteConfig()` …                     |
| read `cache?.response`            | read `cache?.response`                    |
| write `cachedSourceId`            | write `cachedSourceId` ← race             |
| `downloadTopicsWithFallback()`    | `downloadTopicsWithFallback()`            |
| write `cache`                     | write `cache` ← race                      |
| `diskCache.write(response)`       | `diskCache.write(response)` ← race        |

ajpallares (commented):
Wow, there's some mindbending stuff in here (or maybe I'm not having my best day 😅). I think I finally understood the logic here and I think it looks mostly ok. But I believe there are a couple of things to fix 🙏 
Great job on this! 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -111,8 +196,20 @@ internal class RemoteConfigManager(
...
+        val stillFailing: List<TopicTask>,
+        val firstError: PurchasesError?,
+        val invalidating: Boolean,
+    )
+
     private data class CacheEntry(val response: RemoteConfigResponse, val cachedAt: Date)
 
+    private sealed class RefreshOutcome {
+        data class Success(val source: BlobSource) : RefreshOutcome()
+        object VacuousSuccess : RefreshOutcome()
```
JayShortway, line 209: I don't think many people know what this means without looking it up (unless it's just me) 😅 Maybe just `EmptySuccess`?

vegaro: first time I've seen Vacuous as well 😄

JayShortway: Haha so it's not just me 😛

tonidero: That’s… true 🤣 will fix. Thanks!

JayShortway: Hello Toni from the plane! 👋 ✈️

ajpallares: It was the first time for me as well. And then, I just saw this
https://github.com/RevenueCat/purchases-ios/blob/fed4baea51cb961f6238f068aa39eb0404b33316/RulesEngine/Operators/LogicOperators.swift#L33
It seems like AI likes this term 😄

JayShortway: Haha it's the new "delve"

tonidero:
I'm going to close this PR since I'm working off a different stack of PRs (see https://github.com/RevenueCat/purchases-android/pull/3534)  due to changes in the remote config and it's easier to do it separately

PR #3468, opened 2026-05-12
Integrate AdMob reward verification flow in adapter

polmiro:
### Checklist

### Motivation
The AdMob adapter reward verification APIs are scaffolded, and core reward verification result plumbing is available in `:purchases`. This PR wires the adapter callback flow end-to-end so AdMob reward callbacks can return deterministic verification outcomes now, while keeping full polling/retry orchestration out of scope for a follow-up PR.

### Description
- Connects the AdMob reward verification service locator to Purchases lifecycle registration.
- Adds the AdMob reward verification poller and maps core reward verification results into adapter `RewardVerificationResult` values.
- Stores per-ad client transaction IDs on `enableRewardVerification()` and consumes that state when AdMob reports a reward.
- Uses a lifecycle-owned `RewardVerificationManager` to own state, verification scope, cleanup, and main-thread callback delivery.
- Fails safely with `RewardVerificationResult.failed` when reward verification show overloads are used without prior `enableRewardVerification()` or when runtime state is unavailable.
- Treats `PENDING`, `UNKNOWN`, and backend errors as failed results for now; retry/backoff orchestration remains out of scope.
- Keeps AdMob adapter callback delivery on the main thread and trusts AdMob to invoke the reward callback once per show.
- Adds focused unit coverage for locator registration and poller result mapping; show behavior coverage remains Robolectric-based.

Out of scope in this PR:
- polling loops
- retries/backoff
- timeout orchestration

---

> [!NOTE]
> **Medium Risk**
> Touches rewarded-ad grant paths and Purchases service registration; incorrect mapping or callback threading could grant or deny rewards incorrectly, though fail-safe defaults and tests mitigate this.
> 
> **Overview**
> Implements **AdMob reward verification** end-to-end: `enableRewardVerification()` and reward-verification `show` overloads no longer throw; they register a per-ad client transaction ID (keyed by `responseId`), poll Purchases once on reward earned, and deliver **`rewardVerificationCompleted`** on the main thread.
> 
> A **`RewardVerificationManager`** / **`RewardVerificationRuntime`** pair registers with **`Purchases`** lifecycle, maps core verification results into adapter **`RewardVerificationResult`**, and **fails safely** when verification was not enabled, Purchases is not configured, or status is `PENDING`/`UNKNOWN`/errors (retry/polling loops deferred). Public API surface updates include extra `show` overloads without placement and renaming **`rewardVerificationResult`** → **`rewardVerificationCompleted`**. AdMob logging is centralized in **`Logger`**.
> 
> Adds unit/Robolectric tests for poller mapping, runtime lifecycle, manager integration, and show fail-safe/placement behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 228357f069c495fe6d83b9372fc4290eb295b1a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
```diff
@@ -0,0 +1,49 @@
...
+        clientTransactionId: String,
+        fetchStatus: suspend (String) -> RewardVerificationStatus = {
+            Purchases.sharedInstance.awaitGetRewardVerificationStatus(clientTransactionId = it)
+        },
+    ): RewardVerificationResult {
+        return pollOutcome(clientTransactionId, fetchStatus).toResult()
+    }
+
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+    private suspend fun pollOutcome(
```
polmiro, line 21: Not really a poller but this PR is focused on all the wiring and a one-shot for easier incremental changes and facilitating reviews too.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Setup.kt
```diff
@@ -0,0 +1,60 @@
...
+    private class StateStore {
+        private val stateByAd: MutableMap<Any, State> = WeakHashMap()
+
+        @Synchronized
+        fun set(ad: Any, state: State) {
+            stateByAd[ad] = state
+        }
+
+        @Synchronized
+        fun get(ad: Any): State? {
```
tonidero: Seems there is no way to clear an ad anymore, which means we might be leaking those ads entirely? We might need to think of a way to clear those after we're done with the ad somehow...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Setup.kt
```diff
@@ -0,0 +1,60 @@
...
+            return stateByAd[ad]
+        }
+    }
+
+    private val stateStore = StateStore()
```
tonidero: So I've been thinking... Keeping references to objects statically is not great IMO. So I've been thinking whether we could do something so the data was "tied" to the purchases instance.

One possibility that we've discussed before in the @RevenueCat/coresdk team was to use [Service locators](https://developer.android.com/training/dependency-injection#di-alternatives). So we could create a service locator interface, that our SDK initializes at runtime and the admob sdk could create an instance and add it for our SDK to initialize it when configuring, and could be cleared when calling `close`.

Mostly trying to think of ways to avoid this unbounded memory holder...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Dispatcher.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.admob.RewardVerificationResult
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import java.util.concurrent.atomic.AtomicBoolean
+
+private val mainHandler by lazy(LazyThreadSafetyMode.NONE) { Handler(Looper.getMainLooper()) }
+private val scope = CoroutineScope(Dispatchers.IO)
```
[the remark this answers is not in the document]
tonidero: Hmm I think this is true...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Dispatcher.kt
```diff
@@ -0,0 +1,66 @@
...
+import android.os.Looper
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.admob.RewardVerificationResult
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import java.util.concurrent.atomic.AtomicBoolean
+
+private val mainHandler by lazy(LazyThreadSafetyMode.NONE) { Handler(Looper.getMainLooper()) }
```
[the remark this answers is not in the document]
tonidero: Agreed... What's the reasoning to do it this way?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Dispatcher.kt
tonidero: If we end up doing something to bound the state holder to the purchases instance, maybe we could also make the dispatcher a class and just keep an instance so it can be bounded the same way...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
```diff
@@ -44,7 +47,22 @@ public fun RewardedAd.show(
     rewardVerificationStarted: (() -> Unit)? = null,
     rewardVerificationResult: (RewardVerificationResult) -> Unit,
 ) {
-    throw NotImplementedError("AdMob reward verification is not implemented yet.")
+    val state = Setup.verificationState(this)
+    Setup.warnAndAssertIfMissingState(state)
+    val completionDelivered = AtomicBoolean(false)
```
tonidero: Should this be part of the state? Right now seems it's always false

tonidero (commented):
Right now, just concerned a bit with the static nature of the state we are holding... Trying to think of ways we can bound it to the purchases instance at least...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Setup.kt
polmiro: It did catch my attention, I just tried to confirm that it would be released if the ad instance is released via weak reference. But I agree it does not seem ideal. I will try to give this a try, thanks for the suggestion!

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Dispatcher.kt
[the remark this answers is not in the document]
polmiro: Makes sense, some of these concepts are new to me. I see now that the follow up poller calls could fail  if we don't use it.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Setup.kt
polmiro: Looking into the service locator approach at the moment.

Btw Is this something we would want to do in iOS too @ajpallares ?

ajpallares: Hmm I _think_ this is not necessary in iOS because the actual `State`s live on the ads as an Obj-C associated object with `OBJC_ASSOCIATION_RETAIN_NONATOMIC`, so each `State`'s lifetime is tied to the ad's lifetime. The static `stateStore` on iOS holds no ad references. It only has an associated-object key.

So when the ad is deallocated, ARC releases the associated `State`.

polmiro: Alright, not static anymore. Let me know if the approach aligns with what you were thinking :)

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Setup.kt
polmiro: Great callout. I addressed this by moving reward-verification state behind a Purchases-lifecycle runtime and adding explicit cleanup paths. We now remove per-ad state as soon as show(...) starts (consume semantics), and we also clear any remaining state when Purchases closes.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Dispatcher.kt
polmiro: Implemented Dispatcher is now an instance owned by the lifecycle-bound Setup runtime (created on configure, closed on purchases close), so it’s scoped with the same lifecycle as verification state.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
polmiro: completionDelivered is intentionally per-show(...) invocation, not persisted setup state.
It starts false for each call and flips to true on first delivery (compareAndSet) so we guarantee rewardVerificationResult is emitted at most once even if callbacks race or fire twice.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/threading/MainThreadDelivery.kt
```diff
@@ -0,0 +1,16 @@
...
+    mainHandler: Handler,
+    block: (() -> Unit)?,
+) {
+    if (block == null) return
+    if (Thread.currentThread() != Looper.getMainLooper().thread) {
+        mainHandler.post { block() }
+    } else {
+        block()
+    }
+}
```
polmiro, line 16: This is pretty the same as `PurchasesOrchestrator.dispatch` but I was not sure if the wiring was worth while. Perhaps this helper can be promoted to core.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/RewardVerificationExtensions.kt
polmiro: I have actually simplified a lot of this. I think I went overboard with defensiveness.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Manager.kt
```diff
@@ -0,0 +1,191 @@
...
+        @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+        fun handleRewardEarned(
+            onAd: Any,
+            rewardVerificationStarted: (() -> Unit)?,
+            rewardVerificationResult: (RewardVerificationResult) -> Unit,
+        ) {
+            val state = removeState(onAd)
+            warnAndAssertIfMissingState(state)
+
+            deliverStarted(rewardVerificationStarted)
```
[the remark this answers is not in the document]
polmiro: I don't think this is a bug, I expect both to be called and since we call failed right away seems to be what I would expect.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,165 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import java.util.UUID
+import java.util.WeakHashMap
+
+private const val TAG = "PurchasesAdMob"
+
+internal object RewardVerificationManager {
```
tonidero, line 15: [not related to this line] 
This file name should be `RewardVerificationManager` I think?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,165 @@
...
+import com.revenuecat.purchases.admob.threading.runOnMainIfPresent
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
+import java.util.UUID
+import java.util.WeakHashMap
+
+private const val TAG = "PurchasesAdMob"
```
tonidero: Hmm nitpicky but maybe this should be globally defined in some constants file in the module?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesLifecycleListeners.kt
```diff
@@ -2,6 +2,19 @@ package com.revenuecat.purchases
 
 private val defaultLifecycleEventBus = PurchasesLifecycleEventBus()
 
+@OptIn(InternalRevenueCatAPI::class)
```
tonidero: Hmm I think this should propagate the annotation instead of opting in? Like
```suggestion
@InternalRevenueCatAPI
```

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/threading/MainThreadDelivery.kt
tonidero: Right... it's not great to duplicate it, but I think I prefer it than trying to expose this from the core library to use it here... The proper approach would be to move it to a separate module for utils, but we're not doing that yet 😓 . So I think this is fine for now.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesLifecycleListeners.kt
```diff
@@ -2,6 +2,19 @@ package com.revenuecat.purchases
...
+public object PurchasesLifecycle {
+    @JvmSynthetic
+    public fun register(listener: PurchasesLifecycleListener) {
+        defaultLifecycleEventBus.register(listener)
+    }
+
+    @JvmSynthetic
+    public fun unregister(listener: PurchasesLifecycleListener) {
```
tonidero: One thing I've noticed is that we're not unregistering anywhere... I'm wondering if we should... Otherwise we could potentially be leaking some objects?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,171 @@
...
+import java.util.WeakHashMap
+import java.util.concurrent.atomic.AtomicBoolean
+
+private const val TAG = "PurchasesAdMob"
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+internal object RewardVerificationManager {
+    private val mainHandler = Handler(Looper.getMainLooper())
+
+    private val runtime = RewardVerificationRuntime(
```
tonidero, line 18: One thing is that this gets created even before the SDK is configured... I see you tried to avoid instantiating most things until needed, but I think the ideal would be to instantiate this only once the SDK gets configured... but then we might need to use something like the existing [ServiceLoader](https://developer.android.com/reference/java/util/ServiceLoader)... I won't block the PR with this, but might be an improvement we want to do later.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,171 @@
...
+                "Purchases is not configured. Call Purchases.configure() before enabling reward verification.",
+            )
+            return
+        }
+
+        val didStoreClientTransactionId = runtime.setClientTransactionId(
+            ad = onAd,
+            clientTransactionId = UUID.randomUUID().toString(),
+        )
+        if (!didStoreClientTransactionId) {
```
tonidero, line 66: If I'm not wrong, this should never be false, since we already do a check before this for whether purchases is configured. In this case, maybe we should log an error instead of a warning since it's an unexpected situation? I would argue that the warning above could also be an error.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,171 @@
...
+        runtime.handleRewardEarned(
+            onAd = onAd,
+            rewardVerificationStarted = rewardVerificationStarted,
+            rewardVerificationCompleted = rewardVerificationCompleted,
+        )
+    }
+}
+
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+internal class RewardVerificationRuntime(
```
tonidero, line 20: Would it be worth to split to a separate file?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,171 @@
...
+        store[ad] = clientTransactionId
+        return true
+    }
+
+    fun handleRewardEarned(
+        onAd: Any,
+        rewardVerificationStarted: (() -> Unit)?,
+        rewardVerificationCompleted: (RewardVerificationResult) -> Unit,
+    ) {
+        val clientTransactionId = removeClientTransactionId(onAd)
```
tonidero: I see we remove this before starting polling... would it make sense to keep it here until the polling is done? Probably we won't pass the same ad multiple times... but feels safer to avoid duplications?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,171 @@
...
+                notifyCompleted(result, rewardVerificationCompleted)
+            }
+        }
+
+        if (clientTransactionId == null) {
+            deliverOnce(RewardVerificationResult.failed)
+            return
+        }
+
+        val verificationTask = synchronized(this) {
```
tonidero: We're synchronizing here over the poll operation, which means that nothing can else can use the same "lock" while a poll operation is happening, which feels dangerous in case this can take a while... Just to check but is there a reason we need to keep this synchronized? Seems this does not modify the state of the class right now?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    private val runtime = RewardVerificationRuntime(
+        mainHandler = mainHandler,
+    )
+
+    init {
+        Purchases.registerService(runtime)
+    }
+
+    fun install(onAd: Any) {
```
tonidero: I will say it's a bit scary to allow to pass `Any` here, since it's used as the key of the map, which could potentially be leaked unless we call `handleRewardEarned` later... I'm not sure if we could use anything more specific as the id of the Ad and make it typed that way? (not familiar with whether this is possible, just checking in case you know)

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.revenuecat.purchases.RewardVerificationResult as CoreRewardVerificationResult
+import com.revenuecat.purchases.VerifiedReward as CoreVerifiedReward
+
+internal object Poller {
+
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+    suspend fun poll(
+        clientTransactionId: String,
+        fetchResult: suspend (String) -> CoreRewardVerificationResult = {
+            Purchases.sharedInstance.awaitGetRewardVerificationResult(clientTransactionId = it)
```
tonidero: Could be interesting to abstract the `Purchases` instance behind some facade/interface outside this and pass it in as a parameter... But that could be a cleanup later so not a blocker :)

tonidero (commented):
Pretty nice! Mostly some nitpicky comments. Only one more concerning I think is the one to handle the synchronizing over a network request being a bit concerning...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/threading/MainThreadDelivery.kt
polmiro: Agreed, keeping it duplicated for now. We can extract a shared utils module later if the duplication grows.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesLifecycleListeners.kt
polmiro: Addressed by the subsequent rename to `PurchasesService` — the public interface now carries `@InternalRevenueCatAPI` directly. The `@OptIn` remains only on the internal `PurchasesServiceRegistry`, which can't propagate it.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesLifecycleListeners.kt
polmiro: The only registration today is `RewardVerificationManager`'s runtime — both the manager and its runtime are singletons with process lifetime, so unregistering would be a no-op. `Purchases.unregisterService(...)` is available for any future caller that registers something shorter-lived. Does that make sense?

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
polmiro: Agreed — leaving as-is for this PR. Worth noting the eager cost today is just the runtime object allocation; its map + scope are only created on `initialize(purchases)`. Happy to revisit with ServiceLoader (or similar) when we have a second auto-registered service.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
polmiro: Swapped `synchronized(this)` for `@Volatile` on `verificationScope`. The `synchronized` block only wrapped `verificationScope?.launch { ... }` — `launch` returns a `Job` immediately and schedules the lambda on `Dispatchers.IO`, so the poll never executed inside the lock anyway. `@Volatile` gives the same happens-before guarantee against concurrent `initialize()` / `close()` without paying for mutual exclusion. Either close-then-launch interleaving (null scope or cancelled scope) still delivers a `failed` result.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
polmiro: Tackled in the latest commit — introduced a `RewardVerificationFetcher` SAM interface with a default backed by `Purchases.sharedInstance.awaitGetRewardVerificationResult`, and `Poller.poll` now takes the fetcher as a parameter. No more direct `Purchases` reference inside `Poller`.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
polmiro: Tackled — the client transaction id is now kept in the map until the verification poll completes (success, failure, or cancellation via `invokeOnCompletion`). Peek at handoff with `getClientTransactionId`, remove from the completion callback. Entry lifecycle now mirrors the in-flight poll.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        notifyStarted(rewardVerificationStarted)
+
+        verificationTask.invokeOnCompletion { cause ->
+            if (cause != null) {
+                deliverOnce(RewardVerificationResult.failed)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
polmiro: Fixed in the latest commit — `notifyStarted` now runs before the IO coroutine is launched, so on a non-main caller the started post is enqueued on the main handler before the coroutine can ever post completed. The skip-started path on a closed scope is preserved by capturing the scope into a local val and checking it before notifying.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationRuntime.kt
```diff
@@ -0,0 +1,122 @@
...
+                val result = poll(clientTransactionId)
+                deliverOnce(result)
+            }
+        }
+        if (verificationTask == null) {
+            deliverOnce(RewardVerificationResult.failed)
+            return
+        }
+
+        notifyStarted(rewardVerificationStarted)
```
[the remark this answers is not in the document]
polmiro: Fixed in the latest commit — `notifyStarted` now runs before the IO coroutine is launched, so on a non-main caller the started post is enqueued on the main handler before the coroutine can ever post completed. The skip-started path on a closed scope is preserved by capturing the scope into a local val and checking it before notifying.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationManager.kt
```diff
@@ -0,0 +1,170 @@
...
+internal object RewardVerificationManager {
+    private val mainHandler = Handler(Looper.getMainLooper())
+
+    private val runtime = RewardVerificationRuntime(
+        mainHandler = mainHandler,
+    )
+
+    init {
+        Purchases.registerService(runtime)
+    }
```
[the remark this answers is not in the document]
polmiro: Same concern @tonidero raised in https://github.com/RevenueCat/purchases-android/pull/3468#discussion_r3267247453. The only registration today is the `RewardVerificationManager` singleton runtime, which has process lifetime — unregistering would be a no-op. `Purchases.unregisterService(...)` is available for any future caller that registers something shorter-lived.

tonidero (approved):
Looks great! Thank you!!

PR #3470, opened 2026-05-13
refactor(admob): split tracking internals and reward verification scaffolding

polmiro:
### Checklist

### Motivation
Tracking APIs in AdMob are already shipped and should remain compatibility-safe. This refactor improves internal structure by separating tracking internals from the root package while keeping reward-verification APIs as simple unimplemented placeholders.

### Description
This PR is **structure-only** in `feature/admob` (no intended behavior change):

- Moved tracking internals into a dedicated internal area:
  - `.../admob/tracking/TrackIfConfigured.kt`
  - `.../admob/tracking/AdMobPrecisionMapping.kt`
  - `.../admob/tracking/TrackingAdListener.kt`
  - `.../admob/tracking/TrackingOnPaidEventListener.kt`
  - `.../admob/tracking/TrackingFullScreenContentCallback.kt`
- Kept public extension API files as thin entrypoints and rewired imports:
  - tracking entrypoints remain in place and unchanged in signature
- Kept reward-verification extension APIs as direct `NotImplementedError` stubs (no extra internal facade layer).
- Updated tests to reference tracking symbols from the new package.
- Removed old duplicated root-level tracking internal files.

**Before**
- Flat root-level package with tracking internals mixed with public extension files.

**After**
- Root-level public API entrypoints remain thin.
- Tracking internals are grouped under `admob.tracking`.
- Reward-verification APIs remain unimplemented placeholders, unchanged in behavior.

**Behavior parity statement**
- Tracking behavior and public tracking API are intended to remain unchanged.
- Reward-verification public APIs remain unimplemented as before (same `NotImplementedError` behavior).

### Test plan
- Attempted:
  - `./gradlew :feature:admob:test :feature:admob:lint`
  - `./gradlew :feature:admob:testDefaultsBc8DebugUnitTest :feature:admob:testDefaultsBc7DebugUnitTest`
- Blocked by unrelated pre-existing compilation failures in `:purchases`:
  - `purchases/src/main/kotlin/com/revenuecat/purchases/interfaces/GetRewardVerificationStatusCallback.kt`
  - error: internal RevenueCat API annotation usage violation during `:purchases:compileDefaultsBc[7|8][Debug|Release]Kotlin`

No `feature/admob`-specific runtime/lint failures were surfaced before hitting that blocker.


feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/internal/RewardVerificationFacade.kt
```diff
@@ -0,0 +1,39 @@
+package com.revenuecat.purchases.admob.rewardverification.internal
+
+import android.app.Activity
+import com.google.android.gms.ads.rewarded.RewardedAd
+import com.google.android.gms.ads.rewardedinterstitial.RewardedInterstitialAd
+import com.revenuecat.purchases.admob.RewardVerificationResult
+
+internal object RewardVerificationFacade {
```
tonidero: Hmm I would consider the "Facade" as the public API in `RewardVerificationExtensions`, and this would be the internals with the implementations for that facade. So maybe this would be `RewardVerificationOrchestrator`? (This is the pattern we follow with the `Purchases`/`PurchasesOrchestrator`, where `Purchases` is the facade, and we keep the top-level logic in the orchestrator...

This is a bit of a nitipick, since I'm still unsure of the shape of the final implementation, so I wouldn't consider it a blocker, just thinking whether this is the actual facade or it could be a bit of a confusing naming.

tonidero (approved):
One comment regarding the naming, but probably not a blocker. Thanks for the refactor! Should indeed keep things tidier :)

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/internal/RewardVerificationFacade.kt
polmiro: I'm actually cleaning that up, it was slob getting ahead of itself for a future PR which makes no sense. Thanks for the call out. I prematurely opened the PR for review 🤦🏻

polmiro:
> One comment regarding the naming, but probably not a blocker. Thanks for the refactor! Should indeed keep things tidier :)

Of course! I figured this split made sense in the iOS side and cleanly fits here. The good thing now I know the path I can do it all incrementally so the reviews are easier 😄

PR #3475, opened 2026-05-13
Introduce PurchasesService and Purchases.registerService

polmiro:
### Checklist

### Motivation
Provide the lifecycle infrastructure needed by the AdMob reward verification locator while keeping reward verification behavior implementation out of scope for this PR.

### Description
- add `PurchasesLifecycleListener` and `PurchasesLifecycleEventBus` in `:purchases` to publish `configure`/`close` lifecycle events
- add `PurchasesLifecycleListeners` to centralize default lifecycle forwarding
- wire both defaults and custom-entitlement `Purchases` variants to notify lifecycle events on configure/close
- add `RewardVerificationServiceLocator` lifecycle hook registration/dispatch plumbing in `feature:admob`
- update locator dispatch to snapshot hooks under synchronization and invoke callbacks outside the lock
- add focused tests for lifecycle bus behavior, lifecycle wiring (defaults + custom entitlement), and service locator registration/dispatch + concurrency regression behavior
- keep reward verification setup/dispatcher/poller implementation out of scope for this PR

---

> [!NOTE]
> **Low Risk**
> Additive internal lifecycle plumbing with tests; callbacks run under registry synchronization, so poorly behaved services could block others but public SDK behavior is unchanged.
> 
> **Overview**
> Introduces an **internal `PurchasesService` lifecycle hook** so modules can react when a `Purchases` instance is configured or closed, without changing public purchase APIs.
> 
> A synchronized **`PurchasesServiceRegistry`** tracks registered services and the current configured instance: **`onConfigured`** broadcasts `initialize`, **`onClosed`** broadcasts `close` (and clears state when the closed instance matches). Late registrants get an immediate `initialize` if a instance is already active; **re-register** does not replay; **register after close** does not deliver a stale configure. **`Purchases.registerService` / `unregisterService`** wire into this registry via a default forwarder used from both **defaults** and **custom-entitlement** `Purchases` variants.
> 
> **`configure`** now notifies `serviceForwarder.initialize` after creating the singleton; **`close`** notifies `serviceForwarder.close` before tearing down the orchestrator (defaults also still clears the backing singleton). Focused unit tests cover registry ordering, wiring on configure/close, and register/unregister behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a5257b7a427f087ea454e0f1e90580d42e8b8ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


polmiro:
@tonidero this is a preliminary PR that implements the service locator in order to do clean ups for reward verifications. Let me know what do you think, hopefully its in the lines of what you were thinking about!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServiceRegistry.kt
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Synchronized
+    internal fun onConfigured(purchases: Purchases) {
+        configuredPurchases = purchases
+        val listenersToNotify = listeners.toList()
+        listenersToNotify.forEach { listener ->
+            listener.onPurchasesConfigured(purchases)
+        }
+    }
```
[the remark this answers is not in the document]
polmiro: If we want both properties (strict ordering + non-locking callbacks), that likely needs a larger redesign rather than local synchronization tweaks. Wondering how big of a concern is this and what are some possible lightweight ways to address it.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/ServiceLocator.kt
```diff
@@ -0,0 +1,60 @@
+package com.revenuecat.purchases.admob.rewardverification
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.PurchasesLifecycleListener
+
+internal interface RewardVerificationLifecycleHook {
+    fun onPurchasesConfigured(purchases: Purchases)
```
tonidero: Hmm the part that I'm failing to see is how `RewardVerificationServiceLocator` is actually used/registered... I think this is indeed pretty close to what I meant, but to make it clearer, this is what I thought:
- The `purchases` SDK would add a `Service` interface and a "Purchase.registerService" static method (both marked as @InternalRevenueCatAPI). The Service interface would have `initialize` and `close` methods of some kind, possibly passing in the context, or whatever dependency is needed. I think what we're missing in this PR is the `registerService`-like method? In this PR, the `Service` is equivalent to the `PurchasesLifecycleListener`.
- The `admob` SDK can implement an instance of this `Service` with an initialize/close implementations that would actually perform the initialization/cleanup. We have this part mostly I think.
- The `purchases` SDK would call `initialize`/`close`  where appropriate.

So basically what you have here is pretty close, but I'm just missing the part where this is actually "registered". Lmk if I'm missing something though! Just want to make sure we're on the same page.

polmiro: Thanks for the suggestions. I made these changes:
- `PurchasesLifecycleListener` is now `PurchasesService` with `initialize(purchases)` and `close(purchases)` (both `@InternalRevenueCatAPI`).
- New `Purchases.registerService(service)` and `Purchases.unregisterService(service)` (`@InternalRevenueCatAPI`, `@JvmStatic`) on the companion.
- `Purchases.configure(...)` and `Purchases.close()` now drive `initialize` / `close` on every registered service through the internal `PurchasesServiceRegistry`.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServiceRegistry.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    @Synchronized
+    internal fun onConfigured(purchases: Purchases) {
+        configuredPurchases = purchases
+        val servicesToNotify = services.toList()
+        // Keep callback dispatch in this critical section so register/configured/closed
+        // notifications are observed in a single total order.
+        servicesToNotify.forEach { service ->
+            service.initialize(purchases)
```
tonidero, line 35: There might be an argument to do this asynchronously... But I think we can start with this, and iterate later if needed since it's an internal API in any case :)

tonidero (approved):
I think this makes sense for now! cc @JayShortway in case you have any other ideas of features we might want to add to this, since I think this is a pattern we probably will want to reuse in the future 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServiceRegistry.kt
polmiro: Sounds good! definitely can iterate on this down the line

JayShortway:
@polmiro @tonidero Apologies for commenting after merging, but I had some questions 😄 I'm curious how this is going to be used, and why we're not using the "standard" ServiceLoader pattern here? ([This SO answer](https://stackoverflow.com/a/52206921) describes the ServiceLoader quite succinctly.)

tonidero:
It's going to be used by the AdMob module to create a new service that may poll for reward statuses for AdMob on demand. 

Regarding using `ServiceLoader`, I think it would also work indeed... I just thought we could make it more customizable for our needs... but yeah, it might be a good improvement to try to avoid reinventing the wheel. Sorry I didn't think about that!

polmiro:
It's all internal so it should be easy to iterate on this. I'll try to add it to the `admob-reward-verification` branch in a separate PR

PR #3478, opened 2026-05-14
Add :rules-engine-internal skeleton module

ajpallares:
### Checklist

### Motivation
Resolves SDK-4336

Adds a `:rules-engine-internal` module so the upcoming JSON Logic engine can ship as a separate library. Sibling iOS PR: RevenueCat/purchases-ios#6787.

### Description
- Skeleton module with an internal `RulesEngine` namespace, smoke test, and Maven publishing deferred until the engine has a consumer ([#3488](https://github.com/RevenueCat/purchases-android/pull/3488)).
- Module and artifact named `rules-engine-internal` / `purchases-rules-engine-internal` to make it clear that the API is unstable and not meant to be reached for directly, mirroring iOS' `RulesEngineInternal`.
- Metalava and Dokka are also skipped for this module since it has no public-facing API for third-party developers.
- New `test-rules-engine-internal` CircleCI job runs `:rules-engine-internal:testDefaultsDebugUnitTest`.

---

> [!NOTE]
> **Low Risk**
> Scaffolding-only module and build/CI wiring with no runtime behavior or public API surface changes to the main SDK.
> 
> **Overview**
> Adds a new **`:rules-engine-internal`** Android library (artifact `purchases-rules-engine-internal`) as a placeholder for the upcoming JSON Logic rules engine, aligned with iOS’ internal rules module naming.
> 
> The module only exposes an **`internal` `RulesEngine` namespace** plus a smoke unit test. **Maven publishing, Metalava, and Dokka are explicitly skipped** for this path until there is a real consumer; Dokka application is refactored behind **`configureDokka()`** on the public-library convention plugin.
> 
> **CircleCI** gains **`test-rules-engine-internal`** (`:rules-engine-internal:testDefaultsDebugUnitTest`), wired into the main workflow and the release approval gate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d7bf9d1676acaeb0cc8e9dedc845dd6c1233042. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Public namespace for the RevenueCat rules engine.
+ *
+ * This module is an implementation detail of the RevenueCat SDK. Every
+ * declaration here is gated by [InternalRulesEngineAPI] so it can be
+ * consumed by `:purchases` / `:ui:revenuecatui` (and hybrid bridges)
+ * across module boundaries without becoming part of the SDK's public
+ * API. Annotate every new public declaration in this module with
+ * [InternalRulesEngineAPI].
+ */
+@InternalRulesEngineAPI
```
tonidero: Hmm do we need this annotation? I guess someone could depend on the rules engine module directly, but then, I think it's safe to say they are shooting themselves in the foot. We could name the artifact something like `rules-engine-internal` or something to make it clearer.

I'm just a bit reticent to have to add this to all API, since it could get annoying/noisy. And as long as we use this module as `implementation` and not `api`, these APIs shouldn't be exposed to our customers.

tonidero (commented):
It's good in general! Just a thought about whether we need the new annotation... Lmk what you think!

rules-engine/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
ajpallares: 100%. We don't have to, really. Definitely, it would be simpler for us not to have to deal with this annotation at all.

> And as long as we use this module as `implementation` and not `api`

This, I think we should enforce somehow, to avoid using some of these as `api` by mistake. WDYT?

tonidero: Hmm indeed... not sure of ways to enforce this, other than by using some danger rule or something like that... But yeah, I would say not super critical. I think we should always be very careful with anything we use `api` with (since it ends up being public API.

ajpallares: True, true. In iOS, I did come up with a very simple lint rule that indirectly prevents us from doing it (RevenueCat/purchases-ios#6788). But I agree, it's not super critical and not a blocker at all 👍

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureConditionalPublishing.kt
```diff
@@ -10,6 +10,10 @@ import org.gradle.api.Project
  * - For other modules: only apply publishing for variants that don't contain "customEntitlementComputation"
  */
 internal fun Project.configureConditionalPublishing() {
+    // `:rules-engine-internal` is a skeleton with no functionality and no consumers yet,
+    // so we deliberately don't publish `purchases-rules-engine-internal` to Maven Central.
+    if (project.path == ":rules-engine-internal") return
```
tonidero, line 15: I think at some point (once we start using it) we will need to publish this module to maven as well... But we can do that later

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureDokka.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.revenuecat.purchases.android.buildlogic.ktx.libs
+import com.revenuecat.purchases.android.buildlogic.ktx.plugins
+import org.gradle.api.Project
+
+/**
+ * Configures Dokka for RevenueCat libraries that expose a public API surface.
+ */
+internal fun Project.configureDokka() {
+    // Skip Dokka for `:rules-engine-internal` since it's SDK-internal by design.
+    if (project.path == ":rules-engine-internal") return
```
tonidero, line 12: Could be interesting to have a list of "internal" modules that can be skipped for this and metalava... But considering we only have one right now, we can keep this and add that later when needed :)

scripts/api-dump.sh
```diff
@@ -2,6 +2,10 @@
 
 exit_code=0
 
+# Note: `:rules-engine-internal` is SDK-internal by design and opts out of
+# Metalava entirely (see `ConfigureMetalava.kt`), so it has no signature
+# generation task to invoke from this script.
```
tonidero: Hmm I don't think this comment is needed IMO... but FFTI

tonidero (approved):
Very nice!! Thanks for adding this!

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureConditionalPublishing.kt
ajpallares: That's right. That I will add in https://github.com/RevenueCat/purchases-android/pull/3488 (draft for now)

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureDokka.kt
ajpallares: Good idea. But I agree 👍

scripts/api-dump.sh
ajpallares: True. Removed in https://github.com/RevenueCat/purchases-android/pull/3478/commits/a196f0243256e69bea5e4853fe7cfb8e4649f458

PR #3480, opened 2026-05-14
Enforce :rules-engine has no public API outside @InternalRulesEngineAPI

ajpallares:
Resolves SDK-4340

### Checklist

### Motivation
A baseline-diff API check (regenerate `api.txt`, `git diff --exit-code`) is silenceable: a contributor can add a non-internal `public` declaration and just regenerate the baseline.

### Description
Adds an intrinsic CI assertion that the metalava signature dump (`rules-engine/build/api-dump.txt`) contains nothing beyond the `@InternalRulesEngineAPI` annotation itself. Runs in the existing `metalava` job. Bypassing it requires editing the script's allow-list, not regenerating a file.

Verified locally: passes on current source; fails (with the offender printed) when a non-annotated `public fun` is added.

Stacked on #3478.

---

> [!NOTE]
> **Low Risk**
> Adds a CI-only API surface check for `:rules-engine`; no runtime or customer-facing behavior changes.
> 
> **Overview**
> Adds a **CI guard** in the existing **`metalava`** job so `:rules-engine` cannot expose a public API beyond what is marked **`@InternalRulesEngineAPI`**.
> 
> Instead of relying only on a baseline **`api.txt`** diff (which someone can “fix” by regenerating), the job inspects the Metalava signature dump at **`rules-engine/build/api-dump.txt`** and **fails** if any public surface appears that is not on an explicit allow-list (the annotation itself). Bypassing the check means changing that allow-list/script, not refreshing a checked-in API file.
> 
> This is aimed at **SDK-4340** and is stacked on **#3478**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86cdfe37290f7537d35a3508a8ff70b1a5303d6b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero (commented):
I think the logic makes sense in general, but from my comment in the previous PR, I do wonder if it makes sense to just expose the API as "public" but just make it clearer devs shouldn't use it with the artifact name...

ajpallares:
Closing. We won't enforce this in Android for now

PR #3482, opened 2026-05-14
RulesEngine: add JSON Logic predicate evaluator

ajpallares:
Resolves SDK-4326

### Checklist
- [x] Unit tests added

### Motivation
First slice of the JSON Logic rules engine.

### Description
- Adds `Logger`, `RuleError`, `Value`, `Evaluator`, and `operators/` with the MVP set: `var`, `missing`, `==`, `!=`, `===`, `!==`, `!`, `!!`, `and`, `or`, `if`.
- `RulesEngine` and `RulesEngineLogger` are **public** so the host SDK can swap the logger at integration time; everything else stays module-internal.
- Test-only `ValueJsonHelper` lets predicates be authored the way they appear in rule artifacts; production callers will build `Value` trees from the host SDK's own JSON parser.

---

> [!NOTE]
> **Medium Risk**
> New rule-evaluation semantics will gate SDK behavior; subtle JS/json-logic-js alignment (equality, missing vars, short-circuit) is heavily tested but still a correctness-sensitive surface for future rule payloads.
> 
> **Overview**
> Introduces the first **JSON Logic predicate evaluator** in `rules-engine-internal`: callers pass a typed `Value` tree plus a variable map; `Evaluator.evaluate` returns whether the predicate is truthy.
> 
> **`RulesEngine`** is now a **public** namespace with a swappable **`RulesEngineLogger`** (`setLogger`); diagnostics (e.g. missing `var`) go through it instead of threading loggers everywhere. **`Value`** models JSON-shaped data with JSON Logic truthiness and JS-aligned **`looseEq` / `strictEq`** (including array/object stringify coercion). **`Operators`** dispatches the MVP set: **`var`**, **`missing`**, **`==` / `!=` / `===` / `!==`**, **`!` / `!!` / `and` / `or` / `if`** (short-circuit where required). Unknown operators throw **`RuleError.UnsupportedOperator`**.
> 
> Tests add **`ValueJsonHelper`** (test-only JSON → `Value`), **`CapturingLoggerRule`**, and broad unit coverage for evaluators, accessors, equality, and logic. Production parsing stays outside this module; the smoke **`RulesEngineTest`** is removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ffa209da1e17a9dd0cd7af835c1a104c389087a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/AccessorOperators.kt
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * Like [lookupVar], but maps a non-resolving path to [Value.Null]
+     * instead of `null` — the shape `missing` needs.
+     */
+    private fun varLookup(vars: Value, path: String): Value =
+        lookupVar(vars, path) ?: Value.Null
+
+    /**
+     * Mirrors `String(path).split(".")` in json-logic-js — preserves empty
+     * segments at the start, middle, and end (unlike Kotlin's default split).
```
tonidero, line 154: Hmm I'm not sure if we could split + filter for this... but not a strong opinion I guess.

rules-engine-internal/api.txt
```diff
@@ -0,0 +1,16 @@
+// Signature format: 4.0
```
tonidero: Hmm I thought we were disabling metalava for this module since it's internal?

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Logger.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Diagnostic warnings are routed through [RulesEngine.logger].
+ */
+public interface RulesEngineLogger {
+    public fun warn(message: String)
+}
+
+/** Default logger for [RulesEngine.logger]. */
+internal object PrintLogger : RulesEngineLogger {
+    override fun warn(message: String) {
+        System.err.println("[RulesEngine] $message")
```
tonidero: Hmm I guess this will print as a normal error and since it's just using `println`, it might not appear everywhere it should... Maybe we should use android's `Log`ger... 

But then again, it would mean not making this a completely platform agnostic module... And I guess we shouldn't be using the default logger in production... So I think it's ok for now :)

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Value.kt
```diff
@@ -0,0 +1,188 @@
+package com.revenuecat.purchases.rules
+
+/**
+ * A JSON-shaped value for JSON Logic predicates and variable data.
+ *
+ * Numbers are split into [IntValue] and [FloatValue] to preserve type intent.
+ * Cross-type numeric comparisons still work — see [looseEq] and [strictEq].
+ */
+internal sealed class Value {
```
tonidero, line 9: Will this need to be part of the "public" API of the module?

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Evaluator.kt
```diff
@@ -0,0 +1,69 @@
...
+    ): Boolean {
+        val scope = Value.ObjectValue(variables)
+        return evaluateValue(predicate, scope).isTruthy
+    }
+
+    /**
+     * Recursive evaluator. Module-internal so operator implementations can
+     * call it for short-circuit / nested evaluation.
+     */
+    fun evaluateValue(
```
tonidero: Should we make this `internal` so we don't forget later?

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Logger.kt
```diff
@@ -0,0 +1,17 @@
+package com.revenuecat.purchases.rules
+
+/**
+ * Logging facade for the rules engine.
+ *
+ * Diagnostic warnings are routed through [RulesEngine.logger].
+ */
+public interface RulesEngineLogger {
+    public fun warn(message: String)
```
tonidero: Should we add a `tag` parameter here? Just thinking about eventually hooking this to the existing `LogHandler`... then again, we can probably just hardcode the tag in the mapping to `LogHandler` later, so NABD.

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Logger.kt
tonidero: Very nitpicky but we might want to extract the `[RulesEngine]` part to a `TAG` constant, so we can reuse for the "proper" logger later on?

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
```diff
@@ -1,9 +1,7 @@
...
- * Namespace for RevenueCat's internal rules engine.
- *
- * SDK implementation detail; not part of the public API and not intended for use by
- * third-party developers.
- */
-internal object RulesEngine
+/** Namespace for the RevenueCat rules engine. */
+public object RulesEngine {
+    @Volatile
+    public var logger: RulesEngineLogger = PrintLogger
```
tonidero: Hmm maybe we could add a setter method instead of making this public variable, like `setLogger`? which can be synchronized. Feels safer to me :)

tonidero (commented):
Some nitpicky comments but nothing really blocking. Looking great!

tonidero (approved):

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Value.kt
ajpallares: That's right. It will need to since the SDK will be injecting a `Map<String, Value>` for the arguments. The idea is to tackle that in a follow-up PR

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Logger.kt
ajpallares: Exactly. `PrintLogger` is intentionally minimal and platform-agnostic. Production is expected to set the Logger with an adapter wired to the SDK’s `LogHandler`, so this default is basically a fallback for tests and pre-integration use

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
ajpallares: Makes sense! Done in https://github.com/RevenueCat/purchases-android/pull/3482/changes/6ee2297a93e898551b0c53c03d3d8ec9c0b94963

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Logger.kt
ajpallares: Sure! Done in https://github.com/RevenueCat/purchases-android/pull/3482/changes/6ee2297a93e898551b0c53c03d3d8ec9c0b94963

ajpallares: We can always iterate on this, but for now it's a public constant in the Rules Engine. I guess that's what you meant?

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Logger.kt
ajpallares: Not sure I understood it right with the other comment, but I added a tag parameter in https://github.com/RevenueCat/purchases-android/pull/3482/changes/6ee2297a93e898551b0c53c03d3d8ec9c0b94963 that defaults to `"[RulesEngine]"`. Lmk if you meant something else!

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/Evaluator.kt
ajpallares: True, good catch! Done in https://github.com/RevenueCat/purchases-android/pull/3482/changes/6ee2297a93e898551b0c53c03d3d8ec9c0b94963

rules-engine-internal/api.txt
ajpallares: Oh, right. For some reason it was a leftover after I removed it from the previous PR. Removed it in https://github.com/RevenueCat/purchases-android/pull/3482/changes/6ee2297a93e898551b0c53c03d3d8ec9c0b94963. Thanks!

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/AccessorOperators.kt
ajpallares: json-logic-js uses String(path).split(".")`, which keeps empty segments. Kotlin's `split(".")` is close, but I think that doing `split` + `filter { it.isNotEmpty() }` would drop segments we must preserve:

```
"a..b".split(".").filter { it.isNotEmpty() }  → ["a", "b"]       // want ["a", "", "b"]
".a".split(".").filter { it.isNotEmpty() }    → ["a"]            // want ["", "a"]
"a.".split(".").filter { it.isNotEmpty() }    → ["a"]            // want ["a", ""]
".".split(".").filter { it.isNotEmpty() }     → []              // want ["", ""]
"".split(".").filter { it.isNotEmpty() }      → []              // want [""] (whole scope)
```
Those empty segments are real lookup keys, so we walk the path manually.

Does this make sense to you? Note that I'm trying to stick with the JSON logic spec as closely as possible, trying to take it as the unique source of truth. That's why there are some weird things like this.

tonidero: Ohhh ok ok... Yeah, I think that makes sense then.

tonidero (approved):
Thank you!

PR #3484, opened 2026-05-14
Add JSON Logic comparison operators (<, <=, >, >=)

ajpallares:
Resolves SDK-4328

### Motivation
Adds the JSON Logic comparison operators to express numeric thresholds and ranges.
iOS counterpart: https://github.com/RevenueCat/purchases-ios/pull/6792

### Summary
- Implements `<`, `<=`, `>`, `>=` per the JSON Logic spec.
- `<` / `<=` accept the 3-arg between form (`{"<=": [1, x, 10]}` reads as `1 <= x <= 10`); `>` / `>=` are binary only, matching the JS reference.
- Two-string operands compare lexicographically; mixed types coerce through `Value.toNumberOrNull` and compare as `Double` (any comparison against `NaN` is `false`, so a malformed operand fails closed).

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> New predicate surface for targeting rules; semantics are subtle (lex vs numeric, NaN) but heavily tested and aligned with json-logic-js.
> 
> **Overview**
> Adds JSON Logic **comparison operators** (`<`, `<=`, `>`, `>=`) to the rules engine so predicates can express thresholds and ranges (e.g. launch count ≥ 3).
> 
> A new `ComparisonOperators` module implements spec-aligned behavior: **two strings** compare lexicographically; mixed types use numeric coercion via `toNumberOrNull`, with **NaN comparisons always false**. `<` and `<=` support the **3-argument “between”** form (`a < b < c`); `>` and `>=` stay **binary only**, matching `json-logic-js`. Arrays/objects are stringified for comparison when needed.
> 
> `Operators.dispatch` routes the four operator names to this module. **Unit tests** cover edge cases (lex vs numeric, missing args, extra args dropped) plus **evaluator integration** tests with `var` and between-form predicates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93c3c79736bce2d4ac9f2de7c931b019b824a37b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/ComparisonOperators.kt
```diff
@@ -0,0 +1,145 @@
...
+            return cmp.apply(lhs.value, rhs.value)
+        }
+        return cmp.apply(lhs.asDouble(), rhs.asDouble())
+    }
+
+    /**
+     * Shared 2-or-3 arg "chain" evaluator used by `<` and `<=`.
+     * `json-logic-js` declares the operator as `function(a, b, c)`:
+     * missing operands resolve to `undefined` (NaN comparisons are
+     * always `false`); the 3-arg form is the between-form
```
tonidero, line 117: I guess we don't need to support > 3 arg forms? Like say, whether a list of x items is sorted 😅

rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/operators/ComparisonOperatorsTest.kt
```diff
@@ -0,0 +1,318 @@
+package com.revenuecat.purchases.rules.operators
+
+import com.revenuecat.purchases.rules.Value
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class ComparisonOperatorsTest {
```
tonidero, line 7: Have you thought about writing all these as parameterized tests? IMO, it might be cleaner, and also easier to share later on with other tests once we extract these accross platforms.

tonidero (approved):
Just a comment about changing the tests to make them easier to read/add and extract to share with other platforms... But not a blocker I would say.

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/ComparisonOperators.kt
ajpallares: Right. Honestly, I had the exact same thought. And also, why do `<` and `<=` support 2 and 3 args while `>` and `>=` only support 2 args?

The thing is, that's exactly how JSON logic works.

https://github.com/jwadhams/json-logic-js/blob/master/logic.js#L53-L64

```
    ">": function(a, b) {
      return a > b;
    },
    ">=": function(a, b) {
      return a >= b;
    },
    "<": function(a, b, c) {
      return (c === undefined) ? a < b : (a < b) && (b < c);
    },
    "<=": function(a, b, c) {
      return (c === undefined) ? a <= b : (a <= b) && (b <= c);
    },
```

For now, we're trying to fully follow the spec, to make sure all pieces (backend, SDKs and rc-workflows) have the same logic.

But I agree, that's a weird limitation 😅

rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/operators/ComparisonOperatorsTest.kt
ajpallares: Yes and no. 

So, we do intend to have a unified set of tests driven by a conformance fixtures file in khepri (WIP https://github.com/RevenueCat/purchases-ios/pull/6822). So, for now, I'm thinking of these tests as documentation (and validation) of how the Rules Engine should behave. 

That said, I agree that they could be rewritten as parameterized tests that are easier to read + understand and easier to share with the iOS counterpart if we want to.

That would still mean having the same redundancy we have now (local tests + khepri's fixtures) but we can iterate on that.

For now, I like the idea of parameterizing these so that we can more easily share the file with iOS, even if initially the file is duplicated in both repos. Thank you for the suggestion! I will tackle that in a follow up PR 👌

PR #3485, opened 2026-05-14
Add JSON Logic string + array operators

ajpallares:
Resolves SDK-4332

## Motivation

Implements `in`, `cat`, `substr`, `merge`, and `missing_some` for v1 targeting predicates. Android behavior and tests match iOS (#6793) and the json-logic-js spec.

## Summary

- `StringArrayOperators` covers `in`, `cat`, `substr`, and `merge` using shared `jsString` and `Operators.clampedInt`
- `AccessorOperators.opMissingSome` reuses dot-path lookup from `missing`
- 34 `StringArrayOperatorsTest` cases mirror iOS; 8 `missing_some` accessor tests plus 2 evaluator integration tests

## Notes

- `substr` slices by Unicode code points (differs from JS UTF-16 only for surrogate pairs)
- Stacked on #3484

---

> [!NOTE]
> **Medium Risk**
> Changes rule-evaluation semantics for new operators (including edge cases like empty `in` haystacks and `substr` indexing); well-tested and scoped to the rules engine, not auth or payments.
> 
> **Overview**
> Adds JSON Logic **string/array** operators and **`missing_some`** so v1 targeting predicates can express membership, concatenation, slicing, array flattening, and “at least *N* of these fields present” guards.
> 
> **`StringArrayOperators`** implements **`in`** (substring on strings with empty-haystack false; strict `===` on arrays), **`cat`**, **`substr`** (Unicode code points via `clampedInt`; documented divergence from JS UTF-16 for surrogates), and **`merge`** (one-level flatten). **`Operators.dispatch`** routes these plus **`missing_some`**, which reuses **`missing`** dot-path semantics and compares present count to a threshold via new **`jsToNumber`** (unparseable thresholds → never satisfied). **`jsArrayElementString`** is exposed for **`cat`**; shared coercion stays on existing **`jsString`** / **`strictEq`** helpers.
> 
> Coverage: large **`StringArrayOperatorsTest`**, **`missing_some`** accessor tests, and evaluator integration for **`in`** and **`missing_some`** inside **`if`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3056262b94f6f73b90ea9d53869c38135d445af4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/Operators.kt
```diff
@@ -83,4 +89,16 @@ internal object Operators {
         val rhs = if (evaluated.size >= 2) evaluated[1] else Value.Null
         return lhs to rhs
     }
+
+    /**
+     * Safely truncate a [Double] to [Int] for index / count math.
+     * `NaN` → `0` (matches JS `ToInteger`); `±Infinity` and
+     * out-of-range finite values clamp to [Int.MAX_VALUE] / [Int.MIN_VALUE].
+     */
+    fun clampedInt(value: Double): Int {
```
tonidero, line 99: Hmm I wonder if this should be separated to some utils class... Also, I wonder what happens if we need higher range than int... But maybe that's not a problem for now.

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/StringArrayOperators.kt
```diff
@@ -0,0 +1,118 @@
...
+     * [Value.ArrayValue] haystack, the test is strict element equality
+     * (mirrors JS `Array.prototype.indexOf`, which uses `===`). Any other
+     * haystack type returns `false`. `json-logic-js` implements `in` as
+     * `function(a, b)` (needle, haystack); missing or extra operands
+     * short-circuit to `false`.
+     */
+    fun opIn(args: Value, vars: Value): Value {
+        val evaluated = Operators.evalArgs(args, vars)
+        val needle = evaluated.firstOrNull() ?: Value.Null
+        val haystack = if (evaluated.size >= 2) evaluated[1] else Value.Null
```
tonidero: I must say, these property names 😅

tonidero (approved):
🚢

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/operators/StringArrayOperators.kt
ajpallares: hahah right. Seems it's the convention for this? 😄

PR #3488, opened 2026-05-14
RulesEngine: Maven Central publishing wiring

ajpallares:
Resolves SDK-4338

**Stacked on top of:** [#3478 — `:rules-engine`: internal skeleton module](https://github.com/RevenueCat/purchases-android/pull/3478).

### Description

This PR flips on every Maven Central publishing piece for the `:rules-engine` module:

- Drops the `:rules-engine` short-circuit in `ConfigureConditionalPublishing` so `com.vanniktech.maven.publish` gets applied again.
- Adds the `mavenPublishing { … }` block (`AndroidSingleVariantLibrary("defaultsRelease", …)`), gated by `plugins.withId("com.vanniktech.maven.publish") { … }` so the custom-entitlement-computation deploy lane keeps working.
- Adds `:rules-engine:metalavaGenerateSignatureDefaultsRelease` to `scripts/api-dump.sh` so CI catches API drift in `rules-engine/api.txt`.
- Adds the `dokkaHtmlPartial` suppression so the published multi-module dokka site doesn't include an empty `rules-engine` page.
- Re-adds the `api(project(":rules-engine"))` constraint in `:bom`.

The `metalava { … }` config + the committed `rules-engine/api.txt` baseline already live in the skeleton PR (#3478) — they're standard hygiene for every `revenuecat.public.library` module — so this PR just wires them into CI enforcement.

### Why a separate PR?

The skeleton itself (#3478) ships a kotlin android library gated by `@InternalRulesEngineAPI` with no functionality and no consumer. We deliberately split publishing wiring out so the skeleton can land independently and we don't start publishing empty `purchases-rules-engine` versions to Maven Central on every SDK release (Maven Central versions are immutable, so once `10.7.0` exists we're committed to keeping it published forever).

This PR is kept open as a draft and will be merged alongside (or just before) the first PR that introduces a real consumer of `:rules-engine`. Until then, the skeleton compiles, runs tests, and gets detekt'd on every PR without producing any external artifact.


ajpallares:
Closing in favor of #3838

PR #3508, opened 2026-05-27
Wire workflows fetching and add cache persistence

vegaro:
## Summary

Wires paywall workflows fetching end-to-end and adds in-memory + disk caching for both the workflows list and individual workflow details.

**Workflows list (`WorkflowManager.getWorkflowsList`)**
- Fetches the paywall workflows list (`type=paywall`), persists the response to `DeviceCache`, and builds an in-memory `offeringId → workflowId` map exposed via `workflowIdForOfferingId()`.
- Prefetches workflow details for every entry marked `prefetch == true`.
- `onComplete` fires only after the list fetch **and** all prefetch CDN fetches finish, so `workflowIdForOfferingId()` is safe to call as soon as `getOfferings` succeeds.
- Concurrent callers are deduplicated (one network request, all callbacks drained together); on network failure the `offeringId → workflowId` map is restored from the disk cache.

**Workflow details + caching (`WorkflowsCache`)**
- New `WorkflowsCache` is the single owner of in-memory workflow state: resolved per-workflow `WorkflowDataResult`s and the workflows list with its `offeringId → workflowId` map.
- `getWorkflow` serves a fresh cached result without a backend round-trip, so a prefetched workflow is reused when its paywall opens instead of being re-fetched. Uses the same 5 min / 25 hr foreground/background TTL as offerings.
- `DeviceCache` gains `cacheWorkflowsListResponse` / `getWorkflowsListResponseCache` / `clearWorkflowsListResponseCache` for list persistence, following the offerings-response pattern.

**Lifecycle (mirrors offerings)**
- `getWorkflowsList()` runs after offerings are cached, and offerings `onSuccess` now waits for it to complete when the endpoint is enabled.
- Identity transitions (login, alias, switch/logout) clear the workflow caches alongside offerings: the in-memory `WorkflowsCache` and the disk list cache (via `OfferingsCache.clearCache`), so a user switch never serves the previous user's workflows.

**Gating**
- Behind the `USE_WORKFLOWS_ENDPOINT` build config flag (driven by the `revenuecat.useWorkflowsEndpoint` property, off on CI). When off, `WorkflowManager` is passed as `null` to `OfferingsManager`, so offerings timing is unchanged.
- Replaces the old per-current-offering `WorkflowPreWarmer`.

<details><summary>See this generated image for helpful context</summary>
<p>

poster-full

</p>
</details> 

---

> [!NOTE]
> **Medium Risk**
> Changes when getOfferings completes and adds concurrent prefetch/network paths; identity cache clearing is aligned with offerings but wrong timing could briefly serve stale workflow maps after user switch (documented same race as offerings).
> 
> **Overview**
> Replaces the per-offering **`WorkflowPreWarmer`** with a full workflows pipeline gated by **`USE_WORKFLOWS_ENDPOINT`**: when enabled, **`WorkflowManager`** + new **`WorkflowsCache`** own list/detail caching (memory + disk list in **`DeviceCache`**), prefetch of `prefetch=true` workflows, deduped in-flight list fetches, and an **`offeringId → workflowId`** map.
> 
> **`OfferingsManager`** now waits on **`getWorkflowsList`** before delivering offerings success (cache hits and network paths), and forces the workflows list stale when offerings are freshly fetched from the network (not disk fallback). **`IdentityManager`** clears **`WorkflowsCache`** on login, alias, switch, and logout like offerings.
> 
> **`getWorkflow`** gains detail caching/TTL, broader resolve error handling, and optional **`callbackDispatcher`** for bounded prefetch (4-thread pool + dedicated **`DefaultFileRepository`** with **`limitedParallelism(4)`** for CDN). **`PurchasesOrchestrator.getWorkflow`** is nullable when workflows are off, always **`dispatch`**es callbacks to the main thread, and reports a configuration error if disabled.
> 
> **`DefaultFileRepository`** accepts an injectable **`ioScope`**; downloads no longer wrap **`withContext(IO)`** so concurrency limits apply. **`InMemoryCachedObject.lastUpdatedAt`** is **`@Volatile`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5ae7df77f9826f3e09f9939474bee914ec4a6ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -156,6 +156,10 @@ internal class PurchasesFactory(
                 createEventsExecutor(),
                 runningIntegrationTests = runningIntegrationTests,
             )
+            val concurrentDispatcher = Dispatcher(
+                createConcurrentExecutor(),
+                runningIntegrationTests = runningIntegrationTests,
+            )
 
             var diagnosticsFileHelper: DiagnosticsFileHelper? = null
             var diagnosticsHelper: DiagnosticsHelper? = null
@@ -204,6 +208,7 @@ internal class PurchasesFactory(
                 eventsDispatcher,
                 httpClient,
                 backendHelper,
+                concurrentDispatcher = concurrentDispatcher,
             )
             val coilImageDownloader = CoilImageDownloader(application)
             val fileRepository = DefaultFileRepository(application)
@@ -536,6 +541,12 @@ internal class PurchasesFactory(
         return Executors.newSingleThreadScheduledExecutor(LowPriorityThreadFactory("revenuecat-events-thread"))
     }
 
+    // Bounded pool for backend calls that are issued many-at-a-time (workflow detail prefetch), so
+    // they run concurrently instead of serializing on the single-threaded default executor.
+    private fun createConcurrentExecutor(): ExecutorService {
+        return Executors.newScheduledThreadPool(CONCURRENT_BACKEND_CALLS)
+    }
+
     private class LowPriorityThreadFactory(private val threadName: String) : ThreadFactory {
         override fun newThread(r: Runnable?): Thread {
             val wrapperRunnable = Runnable {
@@ -549,6 +560,10 @@ internal class PurchasesFactory(
     }
 
     companion object {
+        // Size of the concurrent backend executor. Matches the workflow prefetch concurrency cap so
+        // the bounded prefetch fan-out is never starved for threads.
+        private const val CONCURRENT_BACKEND_CALLS = 4
+
         @VisibleForTesting
         internal fun shouldInitializeDiagnostics(
             diagnosticsEnabled: Boolean,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -260,7 +260,10 @@ internal class PurchasesFactory(
                 localeProvider = localeProvider,
             )
 
-            val workflowsCache = WorkflowsCache(deviceCache = cache)
+            // Gated on the same flag as workflowManager below: when workflows are disabled there is
+            // no writer for this cache, so we skip creating it (and the no-op clearCache work in
+            // IdentityManager) entirely. workflowsCache != null <=> workflowManager != null.
+            val workflowsCache = if (BuildConfig.USE_WORKFLOWS_ENDPOINT) WorkflowsCache(deviceCache = cache) else null
 
             val identityManager = IdentityManager(
                 cache,
@@ -358,19 +361,21 @@ internal class PurchasesFactory(
                 fontLoader = fontLoader,
             )
 
-            val workflowManager = WorkflowManager(
-                backend = backend,
-                workflowDetailResolver = WorkflowDetailResolver(
-                    workflowCdnFetcher = FileCachedWorkflowCdnFetcher(
-                        fileRepository = DefaultFileRepository(contextForStorage, "rc_compiled_workflows"),
+            val workflowManager = workflowsCache?.let {
+                WorkflowManager(
+                    backend = backend,
+                    workflowDetailResolver = WorkflowDetailResolver(
+                        workflowCdnFetcher = FileCachedWorkflowCdnFetcher(
+                            fileRepository = DefaultFileRepository(contextForStorage, "rc_compiled_workflows"),
+                        ),
                     ),
-                ),
-                workflowAssetPreDownloader = WorkflowAssetPreDownloader(
-                    paywallComponentsImagePreDownloader = paywallComponentsImagePreDownloader,
-                    offeringFontPreDownloader = offeringFontPreDownloader,
-                ),
-                workflowsCache = workflowsCache,
-            )
+                    workflowAssetPreDownloader = WorkflowAssetPreDownloader(
+                        paywallComponentsImagePreDownloader = paywallComponentsImagePreDownloader,
+                        offeringFontPreDownloader = offeringFontPreDownloader,
+                    ),
+                    workflowsCache = it,
+                )
+            }
 
             val offeringsManager = OfferingsManager(
                 offeringsCache,
@@ -383,7 +388,7 @@ internal class PurchasesFactory(
                 diagnosticsTracker,
                 offeringFontPreDownloader = offeringFontPreDownloader,
                 uiPreviewMode = appConfig.uiPreviewMode,
-                workflowManager = if (BuildConfig.USE_WORKFLOWS_ENDPOINT) workflowManager else null,
+                workflowManager = workflowManager,
             )
 
             log(LogIntent.DEBUG) { ConfigureStrings.DEBUG_ENABLED }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -380,7 +380,7 @@ internal class PurchasesFactory(
                 diagnosticsTracker,
                 offeringFontPreDownloader = offeringFontPreDownloader,
                 uiPreviewMode = appConfig.uiPreviewMode,
-                workflowManager = workflowManager,
+                workflowManager = if (BuildConfig.USE_WORKFLOWS_ENDPOINT) workflowManager else null,
             )
 
             log(LogIntent.DEBUG) { ConfigureStrings.DEBUG_ENABLED }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -380,7 +380,7 @@ internal class PurchasesFactory(
                 diagnosticsTracker,
                 offeringFontPreDownloader = offeringFontPreDownloader,
                 uiPreviewMode = appConfig.uiPreviewMode,
-                workflowPreWarmer = createWorkflowPreWarmer(workflowManager),
+                workflowManager = workflowManager,
             )
 
             log(LogIntent.DEBUG) { ConfigureStrings.DEBUG_ENABLED }
@@ -549,26 +549,5 @@ internal class PurchasesFactory(
             diagnosticsEnabled: Boolean,
             uiPreviewMode: Boolean,
         ): Boolean = diagnosticsEnabled && !uiPreviewMode
-
-        @OptIn(InternalRevenueCatAPI::class)
-        @VisibleForTesting
-        internal fun createWorkflowPreWarmer(
-            workflowManager: WorkflowManager,
-            useWorkflowsEndpoint: Boolean = BuildConfig.USE_WORKFLOWS_ENDPOINT,
-        ): WorkflowPreWarmer? {
-            return if (useWorkflowsEndpoint) {
-                { appUserID, offeringIdentifier, appInBackground ->
-                    workflowManager.getWorkflow(
-                        appUserID = appUserID,
-                        workflowId = offeringIdentifier,
-                        appInBackground = appInBackground,
-                        onSuccess = {},
-                        onError = {},
-                    )
-                }
-            } else {
-                null
-            }
-        }
     }
 }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -156,10 +156,6 @@ internal class PurchasesFactory(
                 createEventsExecutor(),
                 runningIntegrationTests = runningIntegrationTests,
             )
-            val concurrentDispatcher = Dispatcher(
-                createConcurrentExecutor(),
-                runningIntegrationTests = runningIntegrationTests,
-            )
 
             var diagnosticsFileHelper: DiagnosticsFileHelper? = null
             var diagnosticsHelper: DiagnosticsHelper? = null
@@ -208,7 +204,6 @@ internal class PurchasesFactory(
                 eventsDispatcher,
                 httpClient,
                 backendHelper,
-                concurrentDispatcher = concurrentDispatcher,
             )
             val coilImageDownloader = CoilImageDownloader(application)
             val fileRepository = DefaultFileRepository(application)
@@ -376,6 +371,10 @@ internal class PurchasesFactory(
                         offeringFontPreDownloader = offeringFontPreDownloader,
                     ),
                     workflowsCache = it,
+                    prefetchDispatcher = Dispatcher(
+                        createConcurrentExecutor(),
+                        runningIntegrationTests = runningIntegrationTests,
+                    ),
                 )
             }
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -43,6 +43,7 @@ import com.revenuecat.purchases.common.warnLog
 import com.revenuecat.purchases.common.workflows.FileCachedWorkflowCdnFetcher
 import com.revenuecat.purchases.common.workflows.WorkflowDetailResolver
 import com.revenuecat.purchases.common.workflows.WorkflowManager
+import com.revenuecat.purchases.common.workflows.WorkflowsCache
 import com.revenuecat.purchases.identity.IdentityManager
 import com.revenuecat.purchases.paywalls.FontLoader
 import com.revenuecat.purchases.paywalls.OfferingFontPreDownloader
@@ -259,11 +260,14 @@ internal class PurchasesFactory(
                 localeProvider = localeProvider,
             )
 
+            val workflowsCache = WorkflowsCache()
+
             val identityManager = IdentityManager(
                 cache,
                 subscriberAttributesCache,
                 subscriberAttributesManager,
                 offeringsCache,
+                workflowsCache,
                 backend,
                 offlineEntitlementsManager,
                 dispatcher,
@@ -366,6 +370,7 @@ internal class PurchasesFactory(
                     offeringFontPreDownloader = offeringFontPreDownloader,
                 ),
                 deviceCache = cache,
+                workflowsCache = workflowsCache,
             )
 
             val offeringsManager = OfferingsManager(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -559,8 +559,6 @@ internal class PurchasesFactory(
     }
 
     companion object {
-        // Size of the concurrent backend executor. Matches the workflow prefetch concurrency cap so
-        // the bounded prefetch fan-out is never starved for threads.
         private const val CONCURRENT_BACKEND_CALLS = 4
 
         @VisibleForTesting
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -49,6 +49,7 @@ import com.revenuecat.purchases.paywalls.FontLoader
 import com.revenuecat.purchases.paywalls.OfferingFontPreDownloader
 import com.revenuecat.purchases.paywalls.PaywallPresentedCache
 import com.revenuecat.purchases.paywalls.events.PaywallStoredEvent
+import com.revenuecat.purchases.storage.DefaultFileCache
 import com.revenuecat.purchases.storage.DefaultFileRepository
 import com.revenuecat.purchases.strings.ConfigureStrings
 import com.revenuecat.purchases.strings.Emojis
@@ -64,6 +65,10 @@ import com.revenuecat.purchases.utils.PurchaseParamsValidator
 import com.revenuecat.purchases.utils.WorkflowAssetPreDownloader
 import com.revenuecat.purchases.utils.isAndroidNOrNewer
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencyManager
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.NonCancellable
 import java.net.URL
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
@@ -74,7 +79,11 @@ internal class PurchasesFactory(
     private val apiKeyValidator: APIKeyValidator = APIKeyValidator(),
 ) {
 
-    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+    @OptIn(
+        ExperimentalPreviewRevenueCatPurchasesAPI::class,
+        InternalRevenueCatAPI::class,
+        ExperimentalCoroutinesApi::class,
+    )
     @Suppress("LongMethod", "LongParameterList", "CyclomaticComplexMethod")
     fun createPurchases(
         configuration: PurchasesConfiguration,
@@ -363,7 +372,15 @@ internal class PurchasesFactory(
                     backend = backend,
                     workflowDetailResolver = WorkflowDetailResolver(
                         workflowCdnFetcher = FileCachedWorkflowCdnFetcher(
-                            fileRepository = DefaultFileRepository(contextForStorage, "rc_compiled_workflows"),
+                            // Dedicated FileRepository instance with a concurrency-limited scope, so workflow
+                            // CDN downloads are capped without affecting the instances used for images/video.
+                            fileRepository = DefaultFileRepository(
+                                fileCacheManager = DefaultFileCache(contextForStorage, "rc_compiled_workflows"),
+                                ioScope = CoroutineScope(
+                                    Dispatchers.IO.limitedParallelism(MAX_CONCURRENT_WORKFLOW_CDN_FETCHES) +
+                                        NonCancellable,
+                                ),
+                            ),
                         ),
                     ),
                     workflowAssetPreDownloader = WorkflowAssetPreDownloader(
@@ -560,6 +577,10 @@ internal class PurchasesFactory(
 
     companion object {
         private const val CONCURRENT_BACKEND_CALLS = 4
+
+        // Caps concurrent workflow CDN downloads on the dedicated workflows FileRepository scope, the
+        // same way CONCURRENT_BACKEND_CALLS caps concurrent workflow detail fetches on prefetchDispatcher.
+        private const val MAX_CONCURRENT_WORKFLOW_CDN_FETCHES = 4
 
         @VisibleForTesting
         internal fun shouldInitializeDiagnostics(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -260,9 +260,6 @@ internal class PurchasesFactory(
                 localeProvider = localeProvider,
             )
 
-            // Gated on the same flag as workflowManager below: when workflows are disabled there is
-            // no writer for this cache, so we skip creating it (and the no-op clearCache work in
-            // IdentityManager) entirely. workflowsCache != null <=> workflowManager != null.
             val workflowsCache = if (BuildConfig.USE_WORKFLOWS_ENDPOINT) WorkflowsCache(deviceCache = cache) else null
 
             val identityManager = IdentityManager(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -260,7 +260,7 @@ internal class PurchasesFactory(
                 localeProvider = localeProvider,
             )
 
-            val workflowsCache = WorkflowsCache()
+            val workflowsCache = WorkflowsCache(deviceCache = cache)
 
             val identityManager = IdentityManager(
                 cache,
@@ -369,7 +369,6 @@ internal class PurchasesFactory(
                     paywallComponentsImagePreDownloader = paywallComponentsImagePreDownloader,
                     offeringFontPreDownloader = offeringFontPreDownloader,
                 ),
-                deviceCache = cache,
                 workflowsCache = workflowsCache,
             )
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -366,6 +366,7 @@ internal class PurchasesFactory(
                     paywallComponentsImagePreDownloader = paywallComponentsImagePreDownloader,
                     offeringFontPreDownloader = offeringFontPreDownloader,
                 ),
+                deviceCache = cache,
             )
 
             val offeringsManager = OfferingsManager(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -43,7 +43,6 @@ import com.revenuecat.purchases.common.warnLog
 import com.revenuecat.purchases.common.workflows.FileCachedWorkflowCdnFetcher
 import com.revenuecat.purchases.common.workflows.WorkflowDetailResolver
 import com.revenuecat.purchases.common.workflows.WorkflowManager
-import com.revenuecat.purchases.common.workflows.WorkflowPreWarmer
 import com.revenuecat.purchases.identity.IdentityManager
 import com.revenuecat.purchases.paywalls.FontLoader
 import com.revenuecat.purchases.paywalls.OfferingFontPreDownloader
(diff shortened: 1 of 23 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/InMemoryCachedObject.kt
```diff
@@ -8,7 +8,7 @@ import java.util.Date
 internal class InMemoryCachedObject<T>
 @OptIn(InternalRevenueCatAPI::class)
 constructor(
-    internal var lastUpdatedAt: Date? = null,
+    @Volatile internal var lastUpdatedAt: Date? = null,
```
facumenzella, line 11: How does `@Volatile` affect this work? 🤔

facumenzella:
I opened draft PR #3514 with a tiny repro test for this: https://github.com/RevenueCat/purchases-android/pull/3514

The issue is that we mark the workflows-list cache as fresh as soon as the list response comes back, before the prefetched workflow detail has finished. That lets a second caller skip the in-flight work and get `onComplete` early.

The test does this:

1. `getWorkflowsList` succeeds with one `prefetch = true` workflow.
2. `getWorkflow` for that prefetch is held and does not invoke success/error yet.
3. Call `getWorkflowsList` a second time while the prefetch is still pending.
4. Assert the second callback has not completed yet.
5. Then complete the prefetch and assert both callbacks complete.

It currently fails at step 4

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +46,130 @@ internal class WorkflowManager(
...
+        ) {
+            onComplete()
+            return
+        }
+
+        synchronized(callbackLock) {
+            pendingCompletionCallbacks.add(onComplete)
+            if (isFetchingWorkflowsList) return
+            isFetchingWorkflowsList = true
+        }
```
facumenzella: Long-shot here, but shouldn't this take into account `appUserID`?

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
facumenzella, line 88: Isn't this branch skipping `getWorkflowsList`?

facumenzella:
Let's mention that we're caching following the same logic as offerings in the PR 🙏

vegaro:
@facumenzella I think this is ready for another pass, I addressed the user id stuff and the bug you found

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+                    // canceled so further coroutine work in this scope will not execute regardless.
+                    val result = try {
+                        workflowDetailResolver.resolve(response)
+                    } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
                         onError(e.toPurchasesError())
+                        return@launch
                     }
+                    workflowsCache.cacheWorkflow(workflowId, result)
+                    scope.launch {
+                        runCatching { workflowAssetPreDownloader.preDownloadWorkflowAssets(result.workflow) }
```
tonidero: So just to confirm, we will predownload assets for all workflows, even if not "current", right? Edit: Ah nvm, the backend will decide which ones to prefetch 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+                            .onFailure { errorLog(it) { "Failed to pre-download workflow assets" } }
```
tonidero: Hmm I was wondering whether we should call `onError` in this case... but I guess that as long as we can recover even if assets were not downloaded or downloaded midway, we should be ok :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+            FetchDecision.FETCH -> Unit
+        }
+
+        backend.getWorkflows(
+            appUserID = appUserID,
+            appInBackground = appInBackground,
+            type = "paywall",
+            onSuccess = { response ->
+                workflowsCache.cacheWorkflowsList(response, buildOfferingIdMap(response.workflows))
+                deviceCache.cacheWorkflowsListResponse(
```
tonidero: I'm wondering if this should be done inside the `workflowsCache`? Feels like it could be part of its responsibility...

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+                    JsonTools.json.encodeToString(WorkflowsListResponse.serializer(), response),
+                )
+
+                val prefetchWorkflows = response.workflows.filter { it.prefetch }
+                if (prefetchWorkflows.isEmpty()) {
+                    drainCompletionCallbacks(appUserID)
```
tonidero: I wonder if this should just be a local `onComplete` function that we call here... I guess I understand this but I wouldn't have used `drain` in this context 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+                } else {
+                    val remaining = AtomicInteger(prefetchWorkflows.size)
+                    prefetchWorkflows.forEach { summary ->
+                        getWorkflow(
```
tonidero: TBH, I'm wondering if this is something we want to do in parallel... These are mostly going to be static files from CDN right? For the config endpoint, I ended up using a separate mechanism to store those directly to disk (while checking the hash of the content for verification). See https://github.com/RevenueCat/purchases-android/pull/3437/changes#diff-d98723d7ea20dab112f80eb082b16041298d62a0ee40d1f01ced2be967d2002bR22. Happy to walkthrough those changes if you want! But I guess if we have this working, we can keep this and iterate later as well

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+
+    /**
+     * Decides how a [getWorkflowsList] call should proceed, queueing [onComplete] when it must wait.
+     *
+     * Returns [FetchDecision.JOIN] when a fetch for the same [appUserID] is already in flight: the
+     * call joins it regardless of list-cache freshness, because the list response refreshes the
+     * cache before its prefetch details finish — completing on a fresh cache would fire [onComplete]
+     * before the prefetched workflows land. A fetch in flight for a *different* user is not joined,
+     * so an identity switch starts its own fetch instead of inheriting the previous user's list.
+     * [FetchDecision.COMPLETE_NOW] means there is no in-flight work for this user to wait for.
```
tonidero: Hmm should we document these values in the enum itself? Also, it's a bit confusing TBH 😅. I wonder if we should just rely on the deduping mechanisms on the `Backend` class for now for this and just early exit if `!useWorkflowsEndpoint`?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +47,161 @@ internal class WorkflowManager(
...
+            appInBackground = appInBackground,
+            type = "paywall",
+            onSuccess = { response ->
+                workflowsCache.cacheWorkflowsList(response, buildOfferingIdMap(response.workflows))
+                deviceCache.cacheWorkflowsListResponse(
+                    JsonTools.json.encodeToString(WorkflowsListResponse.serializer(), response),
+                )
+
+                val prefetchWorkflows = response.workflows.filter { it.prefetch }
+                if (prefetchWorkflows.isEmpty()) {
```
tonidero: Would it make sense to also filter out workflows summaries without an offering? At least right now, we won't need to get those right? Though I guess the backend can have this logic using the prefetch flag.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
```diff
@@ -0,0 +1,83 @@
...
+ * failure). This layer exists on top of it for the same reason the offerings cache does — to serve
+ * already-fetched and prefetched data synchronously within a session, so opening a paywall reuses a
+ * resolved workflow instead of paying another backend/CDN round-trip. Time-based staleness
+ * ([isWorkflowsListCacheStale] / [isWorkflowCacheStale]) then decides when a refetch is due.
+ */
+@OptIn(InternalRevenueCatAPI::class)
+internal class WorkflowsCache(
+    private val dateProvider: DateProvider = DefaultDateProvider(),
+) {
+    private val cachedWorkflows = mutableMapOf<String, InMemoryCachedObject<WorkflowDataResult>>()
```
tonidero, line 34: Ohh I misunderstood... We won't be caching these workflows on disk? I would think we should to better support possible scenarios without connectivity... It does mean we need to implement a proper cleanup system later to avoid stale data though... So probably for a separate PR.

tonidero (commented):
Mostly some questions but getting close I think!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
vegaro: yeah, good eye! I actually have a local spec for that. Should have probably mentioned that in the PR description

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/InMemoryCachedObject.kt
vegaro: `@Volatile` just makes writes to `lastUpdatedAt` visible across threads without a lock, so a reader on another thread doesn't see a stale value. I added it because the prefetch runs on `Dispatchers.IO` and the staleness check can run on a different thread.

But honestly it's belt-and-suspenders here. Every reader and writer already goes through `@Synchronized` methods in both `WorkflowsCache` and `OfferingsCache`, and that gives the same visibility guarantee. There's no lock-free reader today.

I can drop it if you prefer, your call.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: exactly! and for now they are all prefetch `true`

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: yeah exactly. The asset predownload is best-effort. The paywall will download whatever is missing when it renders, so failing the whole workflow fetch over it would be worse imho. That’s why it’s a `runCatching` that just logs.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: yeah good point. `OfferingsCache.cacheOfferings` already does its own disk write with `deviceCache.cacheOfferingsResponse`, so to mirror it the `cacheWorkflowsListResponse` write should live in `WorkflowsCache` too.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: renamed it to `completePendingCallbacks` to match the `pendingCompletionCallbacks` field. I kept it as a method rather than a local function because it's called from 3 spots (empty-prefetch, prefetch success, prefetch error) and it has to grab the `callbackLock`, pull out the queued callbacks for that `appUserID`, and fire them. A local `onComplete` wouldn’t really capture that. lmk if the new name reads better.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: documented the enum values. On relying on Backend dedup instead... it only dedups the list HTTP request, but `onComplete` fires after the list fetch and all prefetch CDN fetches finish. That is what the JOIN case does. Lmk if the docs make it read better now.

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
```diff
@@ -260,13 +260,12 @@ internal class OfferingsManager(
...
-                    workflowPreWarmer?.invoke(appUserID, it.identifier, appInBackground)
                 }
                 offeringFontPreDownloader.preDownloadOfferingFontsIfNeeded(offeringsResultData.offerings)
                 offeringsCache.cacheOfferings(offeringsResultData.offerings, offeringsJSON)
-                dispatch {
-                    onSuccess?.invoke(offeringsResultData)
-                }
+                val dispatchSuccess = { dispatch { onSuccess?.invoke(offeringsResultData) } }
+                workflowManager?.getWorkflowsList(appUserID, appInBackground, onComplete = dispatchSuccess)
+                    ?: dispatchSuccess()
```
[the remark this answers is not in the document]
vegaro: @tonidero what do you think? should we create another event?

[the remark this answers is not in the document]
tonidero: Well, the purpose of those events is to track the duration of the operation from the perspective of the developer, so I think the current event is ok. But if we add other APIs in the future to access workflows, it might be worth indeed to add more events to track that.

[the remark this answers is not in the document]
vegaro: ok, that's what I was thinking. That way we can actually see how much time we are adding. Or saving?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: did it in this PR. We were already prefetching in parallel (each prefetch `getWorkflow`), and the storing to disk with hash verify is already there via `FileCachedWorkflowCdnFetcher`.

The only gap vs #3437 was the mechanism I was using. So I swapped the counter with the coroutine scope and launch like in your `RemoteConfigManager` in #3437

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
vegaro: this is stale now, the success path always routes through `getWorkflowsList` via dispatchSuccess. resolving.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: `getWorkflowsList` now keys the in-flight dedup by `appUserID`, so a different user starts its own fetch instead of inheriting the previous one.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +46,191 @@ internal class WorkflowManager(
...
+     * See [FetchDecision] for what each outcome means.
+     *
+     * The decision is keyed by [appUserID]: a call only joins an in-flight fetch for the *same*
+     * user, so an identity switch starts its own fetch instead of inheriting the previous one.
+     */
+    private fun resolveWorkflowsListFetch(
+        appUserID: String,
+        appInBackground: Boolean,
+        onComplete: () -> Unit,
+    ): FetchDecision {
```
tonidero: IMO, this double result style is pretty confusing... Maybe we should just have the `onComplete`, and we can call that immediately when needed. Potentially with a parameter on what happened... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +46,191 @@ internal class WorkflowManager(
...
+        if (!useWorkflowsEndpoint) return FetchDecision.COMPLETE_NOW
+        return synchronized(callbackLock) {
+            val inFlight = pendingCompletionCallbacks[appUserID]
+            when {
+                inFlight != null -> {
+                    inFlight.add(onComplete)
+                    FetchDecision.JOIN
+                }
+                workflowsCache.isWorkflowsListCacheStale(appInBackground) -> {
+                    pendingCompletionCallbacks[appUserID] = mutableListOf(onComplete)
```
tonidero: This isn't atomic I think, so in theory we could start multiple requests simultaneously?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +46,191 @@ internal class WorkflowManager(
...
+     * sequence settles — list fetched, every prefetch finished or failed — so it must always fire
+     * exactly once, otherwise offerings would never be delivered to the caller.
+     *
+     * Concurrent callers for the same [appUserID] while a request is in-flight are deduplicated:
+     * the second call queues its [onComplete] and returns without a new network request, then all
+     * pending callbacks for that user fire together when the in-flight sequence finishes. A call
+     * for a different user starts its own fetch rather than joining the in-flight one.
+     */
+    fun getWorkflowsList(appUserID: String, appInBackground: Boolean, onComplete: () -> Unit = {}) {
+        when (resolveWorkflowsListFetch(appUserID, appInBackground, onComplete)) {
```
tonidero: TBH, I wonder if we should remove the `FetchDecision` and `resolveWorkflowsListFetch` and calculate what to do inside this method instead, in a thread-safe way... I also don't like the signature of `resolveWorkflowsListFetch` that both returns a value and has an `onComplete` parameter. That's always confusing, and we could avoid it by unifying inside here.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -4,27 +4,37 @@ package com.revenuecat.purchases.common.workflows
...
+import kotlinx.coroutines.suspendCancellableCoroutine
 
+@Suppress("LongParameterList")
 internal class WorkflowManager(
     private val backend: Backend,
     private val workflowDetailResolver: WorkflowDetailResolver,
     private val workflowAssetPreDownloader: WorkflowAssetPreDownloader,
+    private val workflowsCache: WorkflowsCache,
     private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
+    private val useWorkflowsEndpoint: Boolean = BuildConfig.USE_WORKFLOWS_ENDPOINT,
```
tonidero: I guess this will always be true, since we only create this manager if this is true? So maybe not needed here?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -260,11 +260,14 @@ internal class PurchasesFactory(
                 localeProvider = localeProvider,
             )
 
+            val workflowsCache = WorkflowsCache(deviceCache = cache)
```
tonidero: I wonder if we should also make this nullable for now and depend on the `BuildConfig.USE_WORKFLOWS_ENDPOINT` property, to avoid caching anything...

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
```diff
@@ -0,0 +1,108 @@
...
+     * cached or the payload can't be parsed (the parse failure is logged). Used to recover the
+     * list after a backend failure, mirroring
+     * [com.revenuecat.purchases.common.offerings.OfferingsCache.cachedOfferingsResponse].
+     */
+    fun cachedWorkflowsListResponseFromDisk(): WorkflowsListResponse? =
+        deviceCache.getWorkflowsListResponseCache()?.let { cached ->
+            runCatching { WorkflowJsonParser.parseWorkflowsListResponse(cached) }
+                .onFailure { errorLog(it) { "Failed to restore workflows list from disk cache" } }
+                .getOrNull()
+        }
```
tonidero, line 116: Should we repopulate the inmemory cache when reading from disk like this?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +46,191 @@ internal class WorkflowManager(
...
+                            launch { prefetchWorkflow(appUserID, summary.id, appInBackground) }
+                        }
+                    }
+                    completePendingCallbacks(appUserID)
+                }
+            },
+            onError = { error ->
+                errorLog { "Failed to fetch workflows list: ${error.underlyingErrorMessage}" }
+                workflowsCache.cachedWorkflowsListResponseFromDisk()?.let { response ->
+                    workflowsCache.cacheWorkflowsList(response, buildOfferingIdMap(response.workflows))
```
tonidero: Hmm interesting... so I guess this is done like this to repopulate the inmemory cache after the network error... though I guess we're overwriting in disk with the same value without need... But then again, this should be an edge case, so probably no need to optimize it.

tonidero: Should we also prefetch the individual workflows in case the list workflow fails just in case those weren't downloaded before for any reason? I think it's probably ok, as long as we can recover from not having them cached later when presenting the workflow, specially considering this is an edge case. So just asking.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -36,29 +46,191 @@ internal class WorkflowManager(
...
+                // prefetching its assets would be wasted work. The backend should already only set
+                // prefetch = true for workflows tied to an offering; this is a defensive guard on top.
+                val prefetchWorkflows = response.workflows.filter { it.prefetch && it.offeringId != null }
+                scope.launch {
+                    // Prefetch each workflow concurrently and wait for all of them before completing,
+                    // so onComplete fires only once the whole sequence settles (empty list completes
+                    // right away). A failed prefetch is logged and does not fail the others.
+                    coroutineScope {
+                        prefetchWorkflows.forEach { summary ->
+                            launch { prefetchWorkflow(appUserID, summary.id, appInBackground) }
```
tonidero: So this will launch all prefetches in parallel if I understand correctly... It could be good to setup a limit, so we don't start downloading a ton of files if there are a ton of workflows? FWIW, in the config endpoint, I set it up to have up to 4 parallel downloads

tonidero: Actually... since we're using the `backend` class, that has a single dispatcher, are we doing these in parallel at all? I believe we're still doing these sequentially?

tonidero (commented):
Mostly some questions and small cleanup but almost there!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: this was added to be able to test with and without the flag. Will go away soon 😄

tonidero: Right... but this class itself is only created in the `PurchasesFactory` if the flag is true right? So the fact that this is created already indicates the flag is true I think

vegaro: Ahhh I see what you mean got it

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: Thanks for the feedback! I dropped `FetchDecision` and `resolveWorkflowsListFetch`, the decision now lives inline in `getWorkflowsList`.

Also, the staleness check and the in-flight registration are in the same synchronized block, so the atomicity concern from [the other thread](https://github.com/RevenueCat/purchases-android/pull/3508#discussion_r3341237276) is now explicit and a concurrent caller always sees the registered key and joins rather than starting a second request.

vegaro: [f6bc53d](https://github.com/RevenueCat/purchases-android/pull/3508/commits/f6bc53d86757497a699befb9b980417136dd4b0a)

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: this is now gone in [f6bc53d](https://github.com/RevenueCat/purchases-android/pull/3508/commits/f6bc53d86757497a699befb9b980417136dd4b0a)

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
tonidero: The inmemory cache is already populated from disk, as I myself mention in my next comment 😅

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
vegaro: Done in [d1f2b54](https://github.com/RevenueCat/purchases-android/pull/3508/commits/d1f2b549739c73e8cf0cc2c680276d76bdc0ad47)

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: I was copying the same mechanism as offerings. OfferingsManager.kt:225 calls offeringsCache.cacheOfferings at line 265, which also re-writes the same JSON to disk.

I think you're correct though and just optimized it in 712b68cbe1f00e5c8dff150137ba19f0c0b3c1c1

vegaro: Right, `getWorkflow` self-heals at presentation time. If a workflow’s detail isn’t cached (or is stale) it fetches from the backend, resolves it (which pulls the CDN assets), caches it, and kicks off asset pre-download.

Also, where would the list to prefetch even come from? The only source on this path is the response restored from disk. When the response doesn’t exist in disk, `cachedWorkflowsListResponseFromDisk()` is null, there’s no list to prefetch from at all. Persisting individual workflow details to disk is the separate follow-up PR, and that’s what would actually help that scenario, becuase we would have in cache the detail responses of the same workflows the list contains.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: oh! good eye on the backend dispatcher! let me see what I can do

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: Good call, added a filter. `getWorkflowsList` now drops summaries with a null offeringId before caching

purchases/src/main/kotlin/com/revenuecat/purchases/common/offerings/OfferingsManager.kt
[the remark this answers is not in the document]
vegaro: This is handled now

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -6,25 +6,32 @@ import com.revenuecat.purchases.InternalRevenueCatAPI
...
+    private val workflowsCache: WorkflowsCache,
     private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
 ) {
 
+    // Guards pendingCompletionCallbacks. A key is present while a workflows-list fetch (plus its
+    // prefetch) is in flight for that appUserID, so concurrent callers join only their own user's
+    // request — a different user starts its own fetch instead of receiving the wrong user's list.
+    private val callbackLock = Any()
+    private val pendingCompletionCallbacks = mutableMapOf<String, MutableList<() -> Unit>>()
+
```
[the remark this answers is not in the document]
vegaro: I don't think this is a real bug... `WorkflowManager.close()` only happens on SDK shutdown and at shutdown the SDK already makes no promise that pending callbacks run. This is how the rest of the SDK works

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -355,18 +367,33 @@ internal class PurchasesFactory(
...
+                WorkflowManager(
+                    backend = backend,
+                    workflowDetailResolver = WorkflowDetailResolver(
+                        workflowCdnFetcher = FileCachedWorkflowCdnFetcher(
+                            // Dedicated FileRepository instance with a concurrency-limited scope, so workflow
+                            // CDN downloads are capped without affecting the instances used for images/video.
+                            fileRepository = DefaultFileRepository(
+                                fileCacheManager = DefaultFileCache(contextForStorage, "rc_compiled_workflows"),
+                                ioScope = CoroutineScope(
+                                    Dispatchers.IO.limitedParallelism(MAX_CONCURRENT_WORKFLOW_CDN_FETCHES) +
```
vegaro, line 380: I don't know if this is too much and we should just not limit the CDN requests. I added paralelization to the getWorfklow backend calls but also here. What do you think @tonidero

tonidero: Hmm my main concern is that downloading too many things simultaneously can increase memory usage and actually slow down everything else... Might be early optimizing but I think having some limit is good 🙏 . 

I guess we weren't limiting parallel downloads of paywall assets though... but I guess there could potentially be more offerings/workflows than assets in a paywall? I guess this could actually be either way 😅. So yeah, I don't feel strongly, but now that we have it, we might as well limit it a bit?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -355,18 +367,33 @@ internal class PurchasesFactory(
...
-                    paywallComponentsImagePreDownloader = paywallComponentsImagePreDownloader,
-                    offeringFontPreDownloader = offeringFontPreDownloader,
-                ),
-            )
+                    workflowAssetPreDownloader = WorkflowAssetPreDownloader(
+                        paywallComponentsImagePreDownloader = paywallComponentsImagePreDownloader,
+                        offeringFontPreDownloader = offeringFontPreDownloader,
+                    ),
+                    workflowsCache = it,
+                    prefetchDispatcher = Dispatcher(
```
tonidero, line 391: [Not related to this line] 
I was thinking that, if it's still downloading the workflows, and the customer opens a specific workflow that hasn't been downloaded yet, we will hook to the existing callback and wait for that to finish right? I guess this means that it could potentially slow down loading if it's on a slow connection and needs to finish downloading other workflows... 

Not sure if there would be a good way to prioritize the "on demand" workflows, but it's probably going to complicate things for an edge case, so I'm ok leaving as is for now.

purchases/src/main/kotlin/com/revenuecat/purchases/storage/DefaultFileRepository.kt
```diff
@@ -128,19 +127,20 @@ internal class DefaultFileRepository(
             if (fileCacheManager.cachedContentExists(it)) it else null
         }
 
-    private suspend fun downloadFile(url: URL): UrlConnection = try {
-        withContext(Dispatchers.IO) {
-            verboseLog { "Downloading remote file from $url" }
+    // Runs on the ioScope's dispatcher (Dispatchers.IO by default). Callers that inject a
+    // concurrency-limited dispatcher rely on the whole download running here, so this must not
+    // switch dispatchers.
```
tonidero: Hmm I wonder if we need this comment... NABD though.

tonidero (approved):
This looks great to me. Thank you so much for all the iterations!! 🙇

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -5,60 +5,238 @@ package com.revenuecat.purchases.common.workflows
...
+                    // Restore the in-memory cache from disk without rewriting it — disk already holds
+                    // this payload.
+                    workflowsCache.cachedWorkflowsListResponseFromDisk()?.let { response ->
+                        val filtered = response.onlyWorkflowsWithOfferingId()
+                        workflowsCache.cacheWorkflowsListInMemory(filtered, buildOfferingIdMap(filtered.workflows))
+                    }
+                    completePendingCallbacks(appUserID)
+                },
+            )
+        }
```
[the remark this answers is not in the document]
vegaro: acknowledged. But won't fix

The whole purpose of forceRefresh is “don’t reuse a stale cached list, go to the network.” If there’s an in-flight `getWorkflowsList`, that batch is going to the network right now. So joining it gives you a real, just-returned backend response, which is fresh by definition. The intent of “force fresh” is fully satisfied; you’re not being served stale cache data, you’re being served a live fetch.

PR #3513, opened 2026-05-28
refactor: extract Offering.presentedOfferingContext() helper and apply across SDK

rickvdl:
Added a small helper function for getting the `PresentedOfferingContext` from an Offering object as a follow up on https://github.com/RevenueCat/purchases-android/pull/3424#discussion_r3318327894

Replacing all occurrences of 
```
offering.availablePackages.firstOrNull()?.presentedOfferingContext
```

with the use of the new helper
```
offering.presentedOfferingContext
```

---

> [!NOTE]
> **Low Risk**
> Mechanical refactor with equivalent semantics for offerings with packages; empty offerings still yield null context where callers pass it through directly.
> 
> **Overview**
> Adds an **`@InternalRevenueCatAPI`** `Offering.presentedOfferingContext` property (placement/targeting from the first available package, or **null** when there are no packages) and updates the public API stub files accordingly.
> 
> Call sites across **purchases** and **RevenueCat UI** that previously inlined `availablePackages.firstOrNull()?.presentedOfferingContext` now use `offering.presentedOfferingContext`, including custom paywall impressions, paywall activity launch options, and embedded paywall views. **`PaywallViewModel`** keeps a local `presentedOfferingContextOrDefault` extension that falls back to `PresentedOfferingContext(identifier)` for paywall analytics when the offering has no packages.
> 
> Unit tests cover the new property and the empty-offering case for `CustomPaywallImpressionParams`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 184c1371af274e0726fa1d78743389b51ca42a59. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/paywalls/events/CustomPaywallEventTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/paywalls/events/CustomPaywallEventTest.kt
@@ -404,6 +404,20 @@ class CustomPaywallEventTest {
         assertThat(params.presentedOfferingContext).isEqualTo(expectedContext)
     }
 
+    @Test
+    fun `CustomPaywallImpressionParams Offering-based init with no packages has null presentedOfferingContext`() {
+        val offering = Offering(
+            identifier = "empty-offering",
+            serverDescription = "",
+            metadata = emptyMap(),
+            availablePackages = emptyList(),
+        )
+        val params = CustomPaywallImpressionParams(paywallId = "pw", offering = offering)
+
+        assertThat(params.offeringId).isEqualTo("empty-offering")
+        assertThat(params.presentedOfferingContext).isNull()
+    }
+
     // endregion
 
     // region Data: placement and targeting fields
--- a/purchases/api-defaults-bc7.txt
+++ b/purchases/api-defaults-bc7.txt
@@ -230,6 +230,7 @@ package com.revenuecat.purchases {
     method @kotlin.jvm.Throws(exceptionClasses=NoSuchElementException::class) public com.revenuecat.purchases.Package getPackage(String identifier) throws java.util.NoSuchElementException;
     method public com.revenuecat.purchases.paywalls.PaywallData? getPaywall();
     method public com.revenuecat.purchases.Offering.PaywallComponents? getPaywallComponents();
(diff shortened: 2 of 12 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/OfferingExtensions.kt
```diff
@@ -0,0 +1,5 @@
+package com.revenuecat.purchases
+
+@InternalRevenueCatAPI
+public fun Offering.presentedOfferingContext(): PresentedOfferingContext? =
```
tonidero: Hmm does this need to be an extension function? Could it just be part of the `Offering` itself?

tonidero: Additionally, maybe it could be a computed property, like
```
@InternalRevenueCatAPI
public val presentedOfferingContext: PresentedOfferingContext?
   get() = availablePackages.firstOrNull()?.presentedOfferingContext
```

Feels nicer as a property :)

tonidero (commented):
Many thanks for doing this!! 🙇  Just some comments regarding the API and location.

purchases/src/main/kotlin/com/revenuecat/purchases/OfferingExtensions.kt
rickvdl: Agree, thanks for the suggestion 🙌  Updated

purchases/src/test/java/com/revenuecat/purchases/paywalls/events/CustomPaywallEventTest.kt
```diff
@@ -404,6 +404,20 @@ class CustomPaywallEventTest {
         assertThat(params.presentedOfferingContext).isEqualTo(expectedContext)
     }
 
+    @Test
+    fun `CustomPaywallImpressionParams Offering-based init with no packages has null presentedOfferingContext`() {
```
tonidero, line 408: Same as in iOS, could be good to add some tests to the new property, but NABD.

tonidero (approved):

PR #3517, opened 2026-05-30
Add reward verification polling for AdMob

polmiro:
## Polling reward verification

Implements the retry/polling orchestration.

`Poller.poll()` now:
- Retries up to **`maxAttempts = 10`** with a **jittered ~1s backoff** (`Random.nextDouble(0.75, 1.25)`), giving the ~10s verification window from the parent ticket ADS-198.
- Treats `Verified` / `FAILED` as **terminal** and `PENDING` / `UNKNOWN` as **non-terminal** (keep polling) — mapped inline in the poll loop.
- Retries transient backend errors (`NetworkError`, `UnknownBackendError`); any other error ends as `Failed`.
- Rethrows `CancellationException` from both fetch and backoff to preserve cooperative coroutine cancellation.
- Emits debug/error logging across the poll lifecycle (start, per-attempt, status, transient/terminal/unexpected error, exhaustion), mirroring the iOS poller.
- Exposes injectable `sleepSeconds` / `jitterSeconds` / `maxAttempts` (defaults unchanged) so tests run without real delays.

The poll loop is structured for a single exit with concrete-exception catches, so no detekt suppressions are needed. `RewardVerificationRuntime` picks up the new behavior automatically via its default `poll` delegate. No public API surface changed (`Poller` is `internal`), so no `api.txt` update.

### Tests
Rewrote `PollerTest` to cover: terminal verified/failed without retry, pending→verified, persistent pending/unknown exhaustion → failed, transient network/backend retry→success, non-transient stop, backoff-scheduler failure, and cancellation from both fetch and backoff.

---

> [!NOTE]
> **Medium Risk**
> Changes reward-grant verification timing and internal error/callback contracts; behavior is well covered by expanded Poller and Backend tests but affects AdMob integration paths.
> 
> **Overview**
> Adds **retry polling** for AdMob reward verification: the internal `Poller` loops up to **10** attempts with **jittered ~1s** backoff, treats `PENDING`/`UNKNOWN` as non-terminal, and only stops on `Verified`/`FAILED` or exhaustion. **Transient** failures (`NetworkError`, HTTP **5xx** via `RewardVerificationException.isServerError`) are retried; other errors fail immediately, with lifecycle logging and injectable `sleep`/`jitter` for tests.
> 
> The **purchases** layer now surfaces **`RewardVerificationError`** / **`RewardVerificationException`** (including **`isServerError`** on HTTP errors from `Backend.getRewardVerificationResult`), and the suspend **`awaitGetRewardVerificationResult`** path throws that exception type instead of a plain `PurchasesException` without server context.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f4b3055f4ced192f190fac5fb4887dffcdf8982. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
+++ b/feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
@@ -54,7 +54,6 @@ internal object Poller {
         ).toResult()
     }
 
-    @Suppress("ReturnCount", "TooGenericExceptionCaught")
     private suspend fun pollOutcome(
         clientTransactionId: String,
         fetcher: RewardVerificationFetcher,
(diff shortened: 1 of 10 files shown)


peterporfy (commented):
LGTM!

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
```diff
@@ -2,49 +2,148 @@ package com.revenuecat.purchases.admob.rewardverification
...
-                Outcome.Verified(result.reward.toAdMobReward())
-            // Polling/retry orchestration lands in a follow-up PR; one-shot non-terminal/unknown statuses fail for now.
-            CoreRewardVerificationResult.PENDING,
-            CoreRewardVerificationResult.UNKNOWN,
-            CoreRewardVerificationResult.FAILED,
-            -> Outcome.Failed
+    private fun PurchasesException.isTransientPollingError(): Boolean {
+        return when (code) {
+            PurchasesErrorCode.NetworkError,
+            PurchasesErrorCode.UnknownBackendError,
```
tonidero: Hmm I don't know if unknown backend error could be considered transient... It means the backend returned an error with a code the SDK doesn't know about I think...

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
```diff
@@ -2,49 +2,148 @@ package com.revenuecat.purchases.admob.rewardverification
...
+            -> true
+            else -> false
+        }
+    }
+
+    private fun CoreRewardVerificationResult.logDescription(): String {
```
tonidero, line 141: Hmm I wonder if we should just log the entire `CoreRewardVerificationResult`... but I guess this is fine as well.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
```diff
@@ -2,49 +2,148 @@ package com.revenuecat.purchases.admob.rewardverification
...
         fetcher: RewardVerificationFetcher,
+        sleepSeconds: suspend (Double) -> Unit,
+        jitterSeconds: () -> Double,
+        maxAttempts: Int,
     ): Outcome {
+        Logger.d("Reward verification poll start transactionId=$clientTransactionId maxAttempts=$maxAttempts")
+        var outcome: Outcome? = null
+        var attempt = 0
+        while (outcome == null && attempt < maxAttempts) {
+            Logger.d(
```
tonidero: FFTI but could be `verbose` logging I guess...

tonidero (approved):
I think this makes sense!

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
polmiro: Sounds good!

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
polmiro: Good idea, I'll switch over since it can get verbsoe with the polling.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/Poller.kt
polmiro: I wrapped it in a `RewardVerificationException` for now. Part of this will be greatly simplified once I moved the polling into core which is where it should be since it's not AdMob specific. But that will be in different PR.

tonidero: Right... I think it's okish for now, but yeah, as you said, could indeed be cleaner to move this logic inside the core sdk.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1221,7 +1222,7 @@ internal class Backend(
                 synchronized(this@Backend) {
                     rewardVerificationResultCallbacks.remove(cacheKey)
                 }?.forEach { (_, onErrorHandler) ->
-                    onErrorHandler(error)
+                    onErrorHandler(RewardVerificationException(error, isServerError = false))
```
tonidero: It is a bit weird to pass an exception instance as a parameter since exceptions are normally thrown (it's like we have a `PurchasesError` type which is different than `PurchasesException`). One alternative could be to make a `RewardVerificationError` instead of the exception... But I'm not going to block this if we're planning to improve it in a follow-up PR.

PR #3521, opened 2026-06-01
Clear virtual currencies cache on AdMob virtual currency reward

polmiro:
### Checklist

### Motivation
When AdMob reward verification grants a virtual-currency reward, the locally cached virtual currency balances become stale, so a subsequent `getVirtualCurrencies()` could return outdated values.

### Description
When a reward-verification result is delivered with a `VerifiedReward.VirtualCurrency`, the adapter now calls `Purchases.sharedInstance.invalidateVirtualCurrenciesCache()` (gated on `Purchases.isConfigured`) before delivering the completion callback — mirroring the iOS adapter's `SideEffects.invalidateVirtualCurrenciesCacheIfConfigured` approach and the existing `trackIfConfigured` idiom in this module. The invalidation is an injectable lambda on `RewardVerificationRuntime` and fires exactly once, only for virtual-currency rewards (not no-reward, unsupported, or failed). Added unit tests covering all four outcomes and a README note documenting the automatic invalidation. Note: this branch is stacked on `admob-reward-verification` (PR #3465), so it targets that integration branch rather than `main`.

---

> [!NOTE]
> **Low Risk**
> Small, targeted change in reward-verification completion path with tests and a configured-Purchases guard; no auth or payment flow changes.
> 
> **Overview**
> After AdMob reward verification completes with a **`VerifiedReward.VirtualCurrency`**, **`RewardVerificationRuntime`** now invalidates the Purchases virtual-currencies cache **once**, **before** the completion callback runs, so later **`getVirtualCurrencies()`** is less likely to return stale balances. Invalidation is skipped for failed verification and for non–virtual-currency verified outcomes; if **`Purchases`** is not configured, it logs a warning and does nothing.
> 
> The invalidation hook is injectable on the runtime (for tests). The AdMob README notes the automatic invalidation and that apps should refetch only when the UI needs fresh values. New unit tests cover virtual-currency vs other/failed paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8217b33f6f087f76d2123123ed3a131b4e530a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationSideEffects.kt
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Invalidates the virtual currencies cache if the SDK is configured.
+ *
+ * Called after reward verification grants a virtual-currency reward so the next
+ * [Purchases.getVirtualCurrencies] fetch returns the updated balance instead of a stale cached value.
+ * If [Purchases] has not been configured yet, logs a warning and skips invalidation.
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal fun invalidateVirtualCurrenciesCacheIfConfigured() {
```
tonidero: Hmm are we planning to use this somewhere else? Just wondering if this should instead be a private function in RewardVerificationRuntime instead... Maybe we did it to be able to inject a custom invalidator for tests, but I think that can still be done with a private function there?

tonidero (approved):
Just a comment but I think it looks good!

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationSideEffects.kt
polmiro: Good catch, there's no reason to do it. I'll move inline.

PR #3522, opened 2026-06-01
Replace custom service registry with ServiceLoader pattern

polmiro:
### Checklist
- [x] ~~create follow-up issues for `purchases-ios` and hybrids~~ — N/A, internal-only refactor with no public API change

### Motivation
Resolves [ADS-277](https://linear.app/revenuecat/issue/ADS-277/implement-serviceloader-pattern). The custom `Purchases.registerService`/`unregisterService` + `PurchasesServiceRegistry` solution for wiring optional modules (AdMob reward verification) into the `Purchases` lifecycle can be replaced by the standard `java.util.ServiceLoader` mechanism, removing internal API surface and the need for imperative registration.

### Description
The default service forwarder now discovers `PurchasesService` implementations via `ServiceLoader.load(PurchasesService::class.java, …classLoader)` (loaded lazily once, in the form R8 can optimize, and guarded so a faulty provider degrades to a no-op instead of crashing `configure()`/`close()`), so `PurchasesServiceRegistry` and the `@InternalRevenueCatAPI` `register/unregisterService` methods are deleted — public `api.txt` is unchanged since they were internal. The AdMob module declares a `RewardVerificationService` provider in `META-INF/services/com.revenuecat.purchases.PurchasesService` with a `consumer-rules.pro` keep rule, delegating lifecycle to its existing runtime singleton instead of calling `registerService`. Verified with a new core `PurchasesServicesTest` (discovery + fan-out), the AdMob `RewardVerificationManagerTest`/`RewardVerificationRuntimeTest`, and the core `PurchasesLifecycleWiring*` tests on both flavors, plus `detektAll` and the metalava API check.

> [!NOTE]
> Stacked on top of #3465 (`admob-reward-verification`); based against that branch so the diff shows only the ServiceLoader refactor. Retarget to `main` once #3465 merges.

---

> [!NOTE]
> **Medium Risk**
> Changes how configure/close fan out to optional modules (including AdMob reward verification); release builds depend on ServiceLoader metadata and ProGuard keeps, though failures are designed to degrade without crashing configure.
> 
> **Overview**
> **Optional modules now hook into `Purchases` configure/close via `java.util.ServiceLoader` instead of `registerService` / `unregisterService` and the old registry.**
> 
> The core module drops `PurchasesServiceRegistry` and the internal register APIs, and wires `Purchases` through a **`PurchasesServiceDispatcher`** that loads `PurchasesService` implementations from `META-INF/services`, wraps each `initialize`/`close` in error isolation, and **closes prior providers before reconfigure** so resources do not leak across configurations.
> 
> **AdMob** registers **`RewardVerificationService`** in `META-INF/services` (with a **consumer ProGuard keep** for reflective instantiation). Reward verification state moves from a manager-registered singleton runtime to a **per-configuration service instance** exposed through `RewardVerificationManager.activeService`; `RewardVerificationRuntime` is a plain lifecycle-owned object again.
> 
> Tests shift to the dispatcher/ServiceLoader model and add coverage for discovery, reinitialize teardown, and fault-tolerant provider behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d22e7e94f10152700218eba379d042ae7fd956e2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationService.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * [PurchasesService] that wires AdMob reward verification into the [Purchases] lifecycle. Discovered
+ * via [java.util.ServiceLoader], so it must keep a no-argument constructor. Forwards to
+ * [RewardVerificationManager]'s shared runtime.
+ */
+@OptIn(InternalRevenueCatAPI::class)
+internal class RewardVerificationService : PurchasesService {
+    override fun initialize(purchases: Purchases) {
+        RewardVerificationManager.runtime.initialize()
```
tonidero: Wdyt about creating a new instance of the `RewardVerificationRuntime` here, and clear it in the `close` function? That way we avoid the need to hold extra data statically. It could mean we need to make sure the service remains instantiated in the core SDK but we already are doing that and feels like it's a better approach?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServices.kt
```diff
@@ -1,31 +1,39 @@
...
+ * Forwards [Purchases] lifecycle events to every [PurchasesService] declared on the classpath.
+ *
+ * Implementations are discovered lazily (once) with [ServiceLoader], passing the interface's own
+ * [ClassLoader] so the call stays in the shape R8 can optimize.
+ */
 @OptIn(InternalRevenueCatAPI::class)
-private class PurchasesServiceRegistryForwarder(
-    private val registry: PurchasesServiceRegistry,
-) : PurchasesService {
+private class ServiceLoaderForwarder : PurchasesService {
```
tonidero: Hmm I wonder, does this need to be a `PurchasesService`? Since this is not a service itself, I think it might be confusing IMO.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServices.kt
```diff
@@ -1,31 +1,39 @@
...
+    }
+
     override fun initialize(purchases: Purchases) {
-        registry.onConfigured(purchases)
+        services.forEach { service -> service.initialize(purchases) }
     }
 
     override fun close(purchases: Purchases) {
-        registry.onClosed(purchases)
+        services.forEach { service -> service.close(purchases) }
```
tonidero: If we go with my other suggestion of instantiating a new `RewardVerificationRuntime`, we might want to clear the list of services here. And we could potentially just instantiate the services in the `initialize` function instead of on a lazy property

tonidero (commented):
Just some comments trying to avoid holding data statically if possible. Let me know what you think! 🙏

polmiro:
@tonidero thanks for the review — addressed all three (resolved the threads):

1. **Fresh runtime per configuration** (248bf0e): `RewardVerificationService` now creates a fresh `RewardVerificationRuntime` in `initialize()` and drops it (cancelling its scope) in `close()`. `RewardVerificationManager` holds it as a nullable that is `null` before configuration / after close, so no verification state is held statically across configurations. I kept the holder on the manager (not the service) because the static `install()`/`handleRewardEarned()` entry points need to reach it.
2. **Dispatcher is not a service** (6004da2): it no longer implements `PurchasesService` — introduced an internal `PurchasesServiceDispatcher` type (field `serviceForwarder` → `serviceDispatcher`).
3. **Discover in initialize / clear on close** (6004da2): services are discovered in `initialize()` and cleared in `close()` instead of via a lazy property, so no service instances are retained after `close()`.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationService.kt
polmiro: Sounds good, I just gave it a shot.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServices.kt
polmiro: I added a new type and renamed as well.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServices.kt
polmiro: Done!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesService.kt
```diff
@@ -0,0 +1,15 @@
+package com.revenuecat.purchases
+
+/**
+ * A service that wants to observe the [Purchases] lifecycle.
+ *
+ * Implementations are discovered automatically through [java.util.ServiceLoader]. To register one,
+ * declare the fully qualified implementation class name in a resource file named
+ * `META-INF/services/com.revenuecat.purchases.PurchasesService` and provide a public no-argument
```
JayShortway, line 8: Mostly for context, there are tools like [AutoService](https://github.com/google/auto/tree/main/service) to avoid this manual work, but I think doing it manually is totally fine for now. It's not like we have 100s of services.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationService.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * [PurchasesService] that wires AdMob reward verification into the [Purchases] lifecycle. Discovered
+ * via [java.util.ServiceLoader], so it must keep a no-argument constructor. Owns the
+ * [RewardVerificationManager.runtime] lifecycle: a fresh runtime per configuration, dropped on close.
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+internal class RewardVerificationService : PurchasesService {
+    override fun initialize(purchases: Purchases) {
+        RewardVerificationManager.runtime = RewardVerificationRuntime()
```
tonidero: Hmm I think we could potentially remove the `RewardVerificaitonManager`, or just create an instance of that here in the service, and keep the manager reference inside this Service itself. That way, the holder of all references is the Service itself, and once the service is cleared, all data can be cleared along with it. We don't store anything statically

FWIW, I don't think we're leaking anything right now, but just mostly a cleanup thing I would say.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesService.kt
tonidero: Oh nice! Wasn't aware of it. I would say it could be nice so, in case we move the service, we don't forget to update the META-INF. Otherwise, we should try to add some test that makes sure we can instantiate the RewardVerificationService...

tonidero (commented):
Just one more 🙏 Thank you for all the cleanup! 🙇

polmiro:
Addressed the latest review (resolved the three threads):

1. **Old services not closed on reconfigure** (@cursor / Bugbot) — fixed in f354247. `ServiceLoaderDispatcher.initialize()` now closes existing services before reloading, so a second `configure()` without an explicit `close()` cancels the previous runtime's coroutine scope. Added a test for the reconfigure path.
2. **ServiceLoader discoverability** (@JayShortway / @tonidero) — added a test in b1e9907 asserting `ServiceLoader` discovers and instantiates `RewardVerificationService`, guarding the `META-INF/services` descriptor against the class being moved/renamed. Kept it manual (no AutoService) for now.
3. **Service owns its state** (@tonidero) — done in 23cc4b5. `RewardVerificationService` now owns the `RewardVerificationRuntime` (created with the service, cancelled on `close`); `RewardVerificationManager` only holds a nullable pointer to the active service and reads the runtime through it. All verification *data* now lives on the lifecycle-scoped service and is cleared when it closes. One static pointer remains on the manager because the public API is static extension functions (`ad.enableRewardVerification()`) that need an entry point — but no verification state is static anymore.

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/rewardverification/RewardVerificationService.kt
polmiro: Sounds good, cleaner too. Done!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesService.kt
polmiro: Does seem a bit of an overkill here no? I will leave it for a future improvement when we have a second or third service hooked in.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesServices.kt
```diff
@@ -1,31 +1,63 @@
...
+
+    @Synchronized
     override fun initialize(purchases: Purchases) {
-        registry.onConfigured(purchases)
+        // Tear down a previous configuration's services before replacing them, so reconfiguring without
+        // an explicit close() doesn't leave the old services holding resources.
+        closeServices()
+        services = loadServices()
+        configuredPurchases = purchases
+        services.forEach { service -> service.initialize(purchases) }
```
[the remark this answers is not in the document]
tonidero: Could be good to safeguard against this.

tonidero (approved):
I think this makes sense, we should test it thoroughly, and also make a build on release to make sure there are no issues, but looking good!

PR #3523, opened 2026-06-02
Enable Gradle build cache for local/worktree builds

vegaro:
`org.gradle.caching` was not enabled. This PR turns it on but only locally. CI forces it off.

The real win is local and across worktrees. On a dev machine the cache lives in `~/.gradle/caches/build-cache-1` and accumulates on every build. Some numbers on a clean rebuild of `:purchases:compileDefaultsBc8DebugUnitTestKotlin`, same JDK, cache the only difference:

| | Time | Tasks |
|---|---|---|
| no cache (`--no-build-cache --rerun-tasks`) | 31s | 27 executed |
| cache hit (build dir wiped, cache kept) | 2s | 18 from cache |

The caches dir is machine-global, not per-checkout, so every git worktree shares it. If you build a task in one worktree, it's a cache hit in another.

CI doesn't get this benefit, so it's disabled there. The CircleCI cache key (created by the orb) is a checksum of `build.gradle*`/`settings.gradle*`, not of the source files. So editing a module's code never changes the key and CI keeps serving the same frozen cache snapshot for older sources, meaning any module you actually touch never gets a real cache hit for its current content.

`GRADLE_OPTS` on both CI executors now forces `-Dorg.gradle.caching=false` unconditionally. `gradle.properties` still sets `org.gradle.caching=true` for local/worktree builds.

Releases build fresh regardless: the publish builds also pass `--no-build-cache` as a second line of defense.

---

> [!NOTE]
> **Low Risk**
> Build-configuration only; CI and release paths explicitly opt out of caching, with no runtime or SDK behavior changes.
> 
> **Overview**
> Turns on **Gradle build cache** in `gradle.properties` for local and multi-worktree builds (shared `~/.gradle/caches/build-cache-1`), while keeping CI and Maven publish paths from relying on cached task outputs.
> 
> CircleCI **android-executor** and **android-machine-emulator** now set `GRADLE_OPTS` with `-Dorg.gradle.caching=false` so CI never uses the build cache despite the repo default. Fastlane **deploy** / **deploy_specific_package** publish tasks add **`--no-build-cache`** so release artifacts are always built fresh.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9677dc2065fa2831be5ef0258b447473cc789fd7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gradle.properties
```diff
@@ -55,3 +55,6 @@ SONATYPE_HOST=CENTRAL_PORTAL
 SONATYPE_AUTOMATIC_RELEASE=true
 
 org.gradle.jvmargs=-Xmx4096M -XX:MaxMetaspaceSize=1g -Dfile.encoding=UTF-8
+
+# Enable the Gradle build cache (local task-output cache) for the main build.
```
rickvdl, line 59: Nice simple change! Just wondering if this also applies to releases? If so, I think that's a case where we want to disable it and always have a fresh run just to be on the safer side.

JayShortway: Good one! I think this depends on the cache key we use in the CircleCI config.

tonidero: Agreed, very cool change! And yeah, it would be good if, for the release branches, we do clean builds just in case.

vegaro: Added. The only way I could think of was specifying the no cache flag when building the release in the Fastfile. I think that works?

tonidero: Hmm I think ideally we would run tests without cache before making the release just in case... Maybe we can do a similar thing but checking the current branch? IMO, it's fine to use the cache in main but probably for release branches, it would be good to run without cache.

.circleci/config.yml
```diff
@@ -66,6 +66,22 @@ commands:
               echo "Error installing Java SDK through SDKMAN, continuing with default Java" >&2
             fi
 
+  # On release branches we build and test without the Gradle build cache, so the
+  # release is verified from scratch. Cache stays on for main and feature branches.
+  # No-op off release branches. Relies on $BASH_ENV being sourced before later steps.
+  disable-build-cache-on-release:
```
AlvaroBrey: I think we can configure this in Gradle settings instead, so we don't have to add it to every job / future jobs

```groovy
buildCache {
    local {
        isEnabled = !System.getenv("CIRCLE_BRANCH").orEmpty().startsWith("release/")
    }
}
```

AlvaroBrey (approved):
🙌

PR #3525, opened 2026-06-02
Fade the multipage paywall header during page transitions

vegaro:
Multipage paywall headers used to pop. When one page has a header and the next doesn't, the header vanished the instant the slide finished (or snapped in at the start). Now the header rides the slide.

The header fades with the page transition:
- present → absent: the outgoing header fades out
- absent → present: the incoming header fades in
- both present, or neither: unchanged (no fade)

It rides the same `animatable.value` progress as the slide, read inside `graphicsLayer { alpha }`, so it stays in lock-step and adds no recompositions per frame. The old `!toHasHeroImage` special-case is dropped: the header now fades even when the incoming page has a hero image, instead of dropping abruptly.

https://github.com/user-attachments/assets/ab0e7516-5feb-4c1a-beee-284b87a5dcbe

---

> [!NOTE]
> **Low Risk**
> Paywall UI and transition animation only; layout workaround for leaving headers is localized but worth a quick visual pass on header↔no-header and header↔hero flows.
> 
> **Overview**
> Multipage workflow paywalls now **fade the header in sync with the slide** when a step gains or loses a header, instead of popping at the end of the transition.
> 
> Header selection returns a **`WorkflowHeaderPresentation`** (step id + **ENTERING / LEAVING / STABLE**). Present→absent uses the outgoing header with **LEAVING**; absent→present uses the incoming with **ENTERING**; when both steps have headers, behavior stays **STABLE** with the existing outgoing/incoming pick rules. The old case that **dropped the header immediately** when navigating to a hero page without a header is replaced by a **LEAVING** fade.
> 
> Alpha follows the same transition progress as the slide via **`graphicsLayer`** (no per-frame recomposition). **LEAVING** headers are drawn as a **Box overlay** in main content—not in the scaffold header slot—so **`headerHeightPx`** on the incoming step is not set from the fading header (hero top inset stays correct).
> 
> Tests cover presentation selection (including the hero behavior change) and **`headerAlpha`** mapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f63293167bb9f5e0551e6613f898d08803a6d452. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
@@ -97,9 +97,8 @@ internal fun LoadedWorkflowPaywall(
                     onClick = onClick,
                     modifier = Modifier
                         .fillMaxWidth()
-                        // animatable.value is a draw-phase state read (no recomposition per frame, same as
-                        // workflowTransition); headerRender.role is a plain closed-over value, refreshed when
-                        // a recomposition rebuilds this modifier (role only changes when the transition changes).
+                        // Read animatable.value inside graphicsLayer (draw phase), like workflowTransition,
+                        // so the fade stays in lock-step with the slide without recomposing every frame.
                         .graphicsLayer { alpha = headerAlpha(headerRender.role, transitionState.animatable.value) },
                 )
             }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
@@ -31,6 +31,14 @@ internal data class WorkflowHeaderStepInfo(
     val hasHeader: Boolean,
 )
 
+internal enum class WorkflowHeaderTransitionRole { ENTERING, LEAVING, STABLE }
+
+internal fun headerAlpha(role: WorkflowHeaderTransitionRole, progress: Float): Float = when (role) {
+    WorkflowHeaderTransitionRole.ENTERING -> progress
+    WorkflowHeaderTransitionRole.LEAVING -> 1f - progress
+    WorkflowHeaderTransitionRole.STABLE -> 1f
+}
+
 @Suppress("LongParameterList")
 @Composable
 internal fun LoadedWorkflowPaywall(
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
@@ -33,6 +33,11 @@ internal data class WorkflowHeaderStepInfo(
 
 internal enum class WorkflowHeaderTransitionRole { ENTERING, LEAVING, STABLE }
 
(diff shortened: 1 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
```diff
@@ -223,21 +249,34 @@ private fun WorkflowStepContent(
...
+    val toInfo = stepInfoByStepId[currentStepId]
+    val fromHasHeader = fromInfo?.hasHeader == true
+    val toHasHeader = toInfo?.hasHeader == true
+
+    return when {
+        fromHasHeader && !toHasHeader ->
+            WorkflowHeaderPresentation(fromStepId, WorkflowHeaderTransitionRole.LEAVING)
+        !fromHasHeader && toHasHeader ->
+            WorkflowHeaderPresentation(currentStepId, WorkflowHeaderTransitionRole.ENTERING)
+        fromInfo != null && toInfo != null && fromHasHeader && toHasHeader -> {
```
facumenzella: `fromInfo != null && toInfo != null` is redundant here. `fromHasHeader`/`toHasHeader` are defined as `fromInfo?.hasHeader == true` / `toInfo?.hasHeader == true`, so both already imply their info is non-null. This branch is equivalent to just `fromHasHeader && toHasHeader`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
```diff
@@ -223,21 +249,34 @@ private fun WorkflowStepContent(
...
+            } else {
+                currentStepId
+            }
+            WorkflowHeaderPresentation(stepId, WorkflowHeaderTransitionRole.STABLE)
+        }
+        else -> WorkflowHeaderPresentation(currentStepId, WorkflowHeaderTransitionRole.STABLE)
+    }
 }
 
 private fun shouldUseOutgoingHeader(
```
facumenzella, line 302: `shouldUseOutgoingHeader`'s leading `fromStepInfo.hasHeader &&` is always true at the only call site. This is now only reached from the both-header branch, where `fromHasHeader` is already established, so I think the check is dead and can be dropped.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/LoadedWorkflowPaywall.kt
```diff
@@ -31,6 +32,24 @@ internal data class WorkflowHeaderStepInfo(
...
 )
 
+internal enum class WorkflowHeaderTransitionRole { ENTERING, LEAVING, STABLE }
+
+internal data class WorkflowHeaderPresentation(
+    val headerStepId: String,
+    val role: WorkflowHeaderTransitionRole,
+)
+
+internal data class WorkflowHeaderRender(
```
facumenzella: niiiiit:

I think we could unify these two to simplify a little bit. `WorkflowHeaderRender` is just `WorkflowHeaderPresentation` with `headerStepId` resolved to a state (`stepStates[id] ?: currentState`). 

We could drop `WorkflowHeaderRender`, have `workflowHeaderState` return the `Presentation` directly, and do the lookup at the call site, the duplicated `role` field goes away. Worth keeping `Presentation` string-keyed though, that's what keeps the selection tests cheap. 

Not blocking though

facumenzella (approved):
Haven't manually tested it but looks legit. A video in the PR description would be amazing 👍

PR #3526, opened 2026-06-02
Fix 1px seam between sliding multipage paywall pages

vegaro:
A 1px white line showed between the two pages during a multipage paywall slide. The two surfaces were positioned by independent `translationX` expressions, so their shared edge landed at a sub-pixel position and the two graphics layers didn't fully cover it, letting the background show through.

The problem was a calculation that output a float number, so the number wouldn't match between in and out transitions

---

> [!NOTE]
> **Low Risk**
> Localized Compose graphics-layer math for workflow paywall animations only; no purchase, auth, or data-path changes.
> 
> **Overview**
> Fixes a **1px seam** visible between multipage workflow paywall pages during horizontal slide transitions.
> 
> Slide `translationX` is no longer computed independently for incoming and outgoing steps. A new **`workflowSlideTranslationX`** helper rounds the **incoming** surface to whole pixels and sets the **outgoing** offset from that value so both layers share the same edge and sub-pixel gaps do not show the background. **`WorkflowSlideTranslationXTest`** locks in the rounding and derived-offset behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 665784f1e038bc4d5997818375f3810e694e0dce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/WorkflowSlideTranslationXTest.kt
+++ b/ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/WorkflowSlideTranslationXTest.kt
@@ -9,58 +9,17 @@ internal class WorkflowSlideTranslationXTest {
     private val from = "from"
 
     @Test
-    fun `incoming and outgoing surfaces stay exactly adjacent so there is no sub-pixel gap`() {
-        val width = 1001f
-        val directionFactor = 1f
-        listOf(0f, 0.2f, 0.333f, 0.5f, 0.6665f, 0.9f, 1f).forEach { progress ->
-            val toTx = workflowSlideTranslationX(to, to, from, progress, width, directionFactor)
-            val fromTx = workflowSlideTranslationX(from, to, from, progress, width, directionFactor)
-            // Outgoing sits exactly one width behind incoming in the slide direction: edges coincide.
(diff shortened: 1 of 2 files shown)


ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/WorkflowSlideTranslationXTest.kt
```diff
@@ -0,0 +1,66 @@
...
+        listOf(0f, 0.2f, 0.333f, 0.5f, 0.6665f, 0.9f, 1f).forEach { progress ->
+            val toTx = workflowSlideTranslationX(to, to, from, progress, width, directionFactor)
+            val fromTx = workflowSlideTranslationX(from, to, from, progress, width, directionFactor)
+            // Outgoing sits exactly one width behind incoming in the slide direction: edges coincide.
+            assertThat(toTx - fromTx).isEqualTo(directionFactor * width)
+        }
+    }
+
+    @Test
+    fun `both surfaces are snapped to whole pixels`() {
```
facumenzella, line 12: I feel like this test is the most valid one, and the rest just add noise. Have you verified all the tests can fail? Not a blocker, just a thought

facumenzella (approved):

PR #3534, opened 2026-06-04
refactor(remote-config): RC Container Format v1 support

tonidero:
This adds support for a new binary format that will be used for remote config support from our SDKs. This is currently unused and untested against the backend, but will be in later PRs of the stack

---

> [!NOTE]
> **Low Risk**
> New internal networking types with no production call sites yet; risk is limited to future integration and binary parsing edge cases, which are heavily tested.
> 
> **Overview**
> Introduces an **internal** parser for **RC Container Format v1** binary payloads, intended for upcoming remote config delivery. `RCContainer.parse` reads the `RC` header, treats element 0 as config, and exposes further blobs as zero-copy `RCElement` views with optional SHA-256 checksum verification and lazy lookup by URL-safe base64 refs (matching backend `blob_ref` encoding).
> 
> Invalid or truncated wire data surfaces as `RCContainerFormatException`. Parsing is careful about **8-byte alignment**, **little-endian** sizes, and **not mutating** the caller’s `ByteBuffer` position.
> 
> Coverage includes in-memory unit tests plus **frozen `.bin` fixtures** so accidental v1 wire-format regressions are caught before integration with the backend.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a53127da2cc8409e1702499527ae3a933b252be3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3601** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3601?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3607** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3534** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3534?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3534?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
```diff
@@ -0,0 +1,152 @@
...
+ * Layout (all multi-byte integers little-endian):
+ * ```
+ * Header (8 bytes): magic byte[2]="RC" | version u8 | flags u8 | reserved byte[4]
+ * Element:          checksum byte[24]  | element_size u32 | reserved u32 | element[element_size] | pad→8
+ * ```
+ * Elements repeat until the backing buffer is exhausted (the format stores no count). Element 0 is
+ * always the [config]; the remaining elements are content-addressed by checksum.
+ *
+ * [config] and each [RCElement] expose read-only views that share the backing buffer, so parsing
+ * copies no field bytes. The content-addressed [elements] map is built lazily: only its hex keys
```
ajpallares: We're using `checksumBase64()` for the keys, so I think this should read
```suggestion
 * copies no field bytes. The content-addressed [elements] map is built lazily: only its base64 keys
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
```diff
@@ -0,0 +1,152 @@
...
+            position(position() + HEADER_RESERVED_SIZE)
+            return version to flags
+        }
+
+        /** Throws [RCContainerFormatException] with [message] if fewer than [needed] bytes remain. */
+        private fun ByteBuffer.require(needed: Int, message: () -> String) {
+            if (remaining() < needed) throw RCContainerFormatException(message())
+        }
+
+        /** Reads a big-endian unsigned 32-bit integer as a [Long]. */
```
ajpallares: Above we're using `LITTLE_ENDIAN` instead
```suggestion
        /** Reads a little-endian unsigned 32-bit integer as a [Long]. */
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
```diff
@@ -0,0 +1,152 @@
...
+
+            val parsed = mutableListOf<RCElement>()
+            while (source.hasRemaining()) {
+                source.require(ELEMENT_HEADER_SIZE) {
+                    "Truncated element header: need $ELEMENT_HEADER_SIZE bytes, " +
+                        "got ${source.remaining()}."
+                }
+                val checksum = source.sliceBytes(CHECKSUM_SIZE.toLong(), "checksum")
+                val size = source.readUnsignedInt()
+                val reserved = source.int
```
ajpallares: Both `element_size` and `reserved` fields are defined as `u32` in the docs above (`element_size u32 | reserved u32`) but we're parsing them differently. Shouldn't it be

```suggestion
                val size = source.readUnsignedInt()
                val reserved = source.readUnsignedInt()
```
?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
```diff
@@ -0,0 +1,55 @@
...
+     * [data] is not copied into an intermediate array.
+     */
+    fun isChecksumValid(): Boolean {
+        val digest = MessageDigest.getInstance(SHA_256_ALGORITHM)
+        digest.update(data.duplicate())
+        val computed = digest.digest()
+
+        val expected = checksum.duplicate()
+        val length = expected.remaining()
+        if (length > computed.size) return false
```
ajpallares: Doesn't this always translate to 
```
        if (24 > 32) return false
```
by construction?

This check could be removed if that's the case

purchases/src/test/java/com/revenuecat/purchases/common/networking/RCContainerTest.kt
ajpallares, line 1: The `parse(ByteBuffer)` overload is documented to parse "from its current position", but all tests start at position 0. We could add a test that wraps a larger array, advances the position (e.g. a few bytes of unrelated prefix), and parses from there. E.g.

```
    @Test
    fun `parses from the buffer's current position`() {
        val container = buildContainer(config = "cfg".toByteArray(), elements = listOf("e".toByteArray()))
        val prefix = ByteArray(8) { 0xAB.toByte() } // unrelated bytes before the container
        val combined = prefix + container

        val buffer = ByteBuffer.wrap(combined)
        buffer.position(prefix.size) // start parsing after the prefix

        val parsed = RCContainer.parse(buffer)

        assertThat(parsed.config.data.readBytes()).isEqualTo("cfg".toByteArray())
        assertThat(parsed.contentElements[0].data.readBytes()).isEqualTo("e".toByteArray())
        // The overload must not consume the caller's position.
        assertThat(buffer.position()).isEqualTo(prefix.size)
    }
```

ajpallares: Related to this, not sure if we need to cover this, but just in case. A similar test where the buffer's current position is a non-8-aligned prefix would fail
```kotlin
    @Test
    fun `parses from a non-8-aligned position`() {
        val container = buildContainer(config = "abc".toByteArray(), elements = listOf("element".toByteArray()))
        val prefix = ByteArray(3) { 0xAB.toByte() } // NOT a multiple of 8
        val combined = prefix + container

        val buffer = ByteBuffer.wrap(combined)
        buffer.position(prefix.size)

        val parsed = RCContainer.parse(buffer)

        assertThat(parsed.config.data.readBytes()).isEqualTo("abc".toByteArray())
        assertThat(parsed.contentElements[0].data.readBytes()).isEqualTo("element".toByteArray())
    }
```

ajpallares (approved):
Took a look, and have some comments. Mostly documentation nits and one suggestion about a missing test. But I think it looks good! Good job!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
```diff
@@ -0,0 +1,152 @@
...
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+
+/**
+ * A parsed RC Container Format v1 payload, providing zero-copy access to its fields.
+ *
+ * Layout (all multi-byte integers little-endian):
+ * ```
+ * Header (8 bytes): magic byte[2]="RC" | version u8 | flags u8 | reserved byte[4]
+ * Element:          checksum byte[24]  | element_size u32 | reserved u32 | element[element_size] | pad→8
```
rickvdl, line 12: I believe the format [definition](https://github.com/RevenueCat/khepri/blob/17bacb7e4f2ade1b57786c20ef62357b8139cb74/khepri/serializers/multipart/rc_format/README.md#L15) mentions the checksum as `byte[32]` instead

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
```diff
@@ -0,0 +1,152 @@
...
+    val elements: Map<String, RCElement> by lazy { contentElements.associateBy { it.checksumBase64() } }
+
+    companion object {
+        private const val MAGIC_R = 'R'.code.toByte()
+        private const val MAGIC_C = 'C'.code.toByte()
+        private const val SUPPORTED_VERSION = 1
+        private const val HEADER_RESERVED_SIZE = 4
+        private const val HEADER_FIXED_SIZE = 8
+        private const val ALIGNMENT = 8
+        private const val CHECKSUM_SIZE = 24
```
rickvdl, line 46: Same here should be 32 IIRC

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
```diff
@@ -0,0 +1,152 @@
...
+
+            if (parsed.isEmpty()) {
+                throw RCContainerFormatException("Missing config element (element 0).")
+            }
+
+            return RCContainer(
+                version = version,
+                flags = flags,
+                config = parsed.first(),
+                contentElements = parsed.drop(1),
```
rickvdl, line 97: Ha the `.drop(1)`  confused me for a second, I thought it was dropping index 1 (the first element after the config) 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
```diff
@@ -0,0 +1,55 @@
...
+     * Computes the SHA-256 of [data], truncates it to the stored [checksum]'s length, and compares.
+     *
+     * The backend stores `sha256(data)` truncated to its leftmost [checksum] bytes (192 bits), so
+     * we compare only that many leading bytes of the computed digest. This reads directly off the
+     * backing buffer (via [ByteBuffer.duplicate]) so neither [data] nor [checksum] is consumed, and
+     * [data] is not copied into an intermediate array.
+     */
+    fun isChecksumValid(): Boolean {
+        val digest = MessageDigest.getInstance(SHA_256_ALGORITHM)
+        digest.update(data.duplicate())
```
rickvdl: FYI it looks like (even on readonly buffers) the `.position()` is still mutable, in other words, if something already read part of the data before duplicating the position may no longer be at the expected location. This may be fine in the current setup / order of reading, but something to keep in mind. 

If we want to guard against this I think we can set the position to 0 or use `rewind()`

rickvdl (commented):
Great work!! 💪 Couple of questions

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
tonidero: I believe we changed to a 24 byte format in a follow-up PR on the backend. Happy to chat more of course!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
tonidero: Good catch!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
rickvdl: You're right, the latest PR in the stack updated that. Closing this one

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3534/commits/9cd9f130d9536d597b1a0412c0b7dab8ccc81903 together with the other comments

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
tonidero: yeah... It indeed can be removed. Thought about keeping it as an extra check (and bounding it to 24, not 32), but yeah I think it's not doing anything right now.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContainer.kt
tonidero: Added a small clarifying comment 🙏 https://github.com/RevenueCat/purchases-android/pull/3534/commits/9cd9f130d9536d597b1a0412c0b7dab8ccc81903

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
tonidero: Good point, better to be safe indeed! https://github.com/RevenueCat/purchases-android/pull/3534/commits/9cd9f130d9536d597b1a0412c0b7dab8ccc81903

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
ajpallares: I'm not against it, FWIW, if it can serve as future-proof. Just wanted to raise in case it was an overlook 🙏

PR #3538, opened 2026-06-04
fix(test-store): don’t mention Play Store in error logs when the Test Store is used

rickvdl:
Android counterpart for https://github.com/RevenueCat/purchases-ios/pull/6906, making offering product loading errors test-store aware. This prevents these errors from pointing developers to the Play Store when Test Store is used.

These errors should rarely happen (if at all) but figured it was good to add for consistency.

---

> [!NOTE]
> **Low Risk**
> String-only branching on configured store for logs and configuration errors; behavior unchanged aside from messaging, with unit test coverage.
> 
> **Overview**
> When the SDK is configured for **Test Store**, offering/product configuration errors no longer tell developers to fix things in the **Play Store**. The change threads `appConfig.store` into offerings building and picks **Test Store–specific** copy from `OfferingStrings` instead of the Play Store templates.
> 
> **`OfferingsFactory`** now uses Test Store strings for missing-product warnings and for the configuration error when no dashboard products resolve. **`OfferingParser.createOfferings`** gains a `configuredStore` parameter (default Play Store) and routes empty-offering warnings through a small helper that selects the right template. New constants cover cannot-find-product, products-not-found, and empty-offering cases for Test Store (RevenueCat dashboard wording). Tests assert string content and that Test Store flows return the Test Store configuration error without mentioning Play Store.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a1b9c31ae72484f65f5b9dd7eb269d200d6f77c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
@@ -8,6 +8,7 @@ import com.revenuecat.purchases.Offerings
 import com.revenuecat.purchases.Package
 import com.revenuecat.purchases.PackageType
 import com.revenuecat.purchases.PresentedOfferingContext
+import com.revenuecat.purchases.Store
 import com.revenuecat.purchases.UiConfig
 import com.revenuecat.purchases.models.StoreProduct
 import com.revenuecat.purchases.paywalls.PaywallData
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Indeed... I would say the only scenarios where this would happen are if there are no test store products configured at all, or for some reason, the requests to fetch those fail... But yeah, I think it doesn't hurt.

PR #3537, opened 2026-06-04
Persist workflow detail envelopes for recovery when backend is down

vegaro:
This PR persists the detail envelopes to disk, mirroring what we do with `OfferingsCache`, so a cold start with the backend down renders the prefetched workflows. We are already saving to disk the list of workflows, and we were missing this part.

---

> [!NOTE]
> **Medium Risk**
> Changes paywall/workflow delivery and disk cache behavior on network failure and cold start; scope is bounded (prefetch + prune) but incorrect recovery could affect paywall rendering.
> 
> **Overview**
> Adds **disk persistence of workflow detail envelopes** (alongside the existing workflows list cache) so prefetched paywalls can be **re-resolved offline** after restart or when the backend is unreachable.
> 
> **Prefetch-only writes:** `getWorkflow` gains `persistEnvelopeOnResolve` (default `false`). Only the workflows-list **prefetch** path sets it `true` after a successful resolve, storing the raw `WorkflowDetailResponse` in `DeviceCache` via `WorkflowsCache` (merge-by-workflow-id map). On-demand fetches do **not** persist, to avoid unbounded disk use.
> 
> **Backend-down recovery:** When `getWorkflowsList` fails, the manager still restores the list from disk; if persisted envelopes exist, it **re-resolves** them in parallel into the in-memory workflow cache (`restoreWorkflowFromEnvelope`), so subsequent `getWorkflow` can be a cache hit without calling the backend. Failures per envelope are logged and isolated; `onComplete` still runs once.
> 
> **Cache hygiene:** `cacheWorkflowsList` **prunes** stored envelopes to IDs in the latest list; `clearCache` clears the envelope store on identity transitions. `WorkflowJsonParser` adds parsing for the envelope map.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bc39f941c43d3b10185966d370714e3b9f865bef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3540, opened 2026-06-05
Stale-while-revalidate for the workflow detail fetch

vegaro:
Linear https://linear.app/revenuecat/issue/WFL-281/stale-while-revalidate-for-the-workflow-detail-fetch

Stacked on #3537.

I noticed that in `OfferingsManager` we do stale-while-revalidate serving (`vendCachedOfferingsAndMaybeRefresh`). Meaning that we serve stale data while fetching in the background more fresh offerings. We were not doing that for workflows, so there was a change in behavior. Before this PR a "stale" cache was a "miss" that blocked the paywall on a full backend round-trip.

To understand it better, I've generated these diagrams:

## How offerings already behaves

When a cached value exists it is always served right away; staleness only decides whether a background refresh is also kicked off. The caller never waits on the network when the cache has something to give.

Captura de pantalla 2026-06-05 a las 8 14 17

## Workflows before this PR

`getWorkflow` only short-circuited on a fresh hit. A cached-but-stale workflow fell through to the same path as a cache miss and blocked the render on a full refetch, even though a perfectly usable workflow was sitting in the cache.

Captura de pantalla 2026-06-05 a las 8 15 25

## Workflows after this PR

Now `getWorkflow` serves the cached value immediately, then fire a background refresh that delivers no callbacks (a success just updates the cache, a failure is logged and swallowed).

Captura de pantalla 2026-06-05 a las 8 16 53

<details>
<summary>📋 Design context (for reviewers and AI agents)</summary>

### Why

This is follow-up #3 from the workflow-detail-envelope-persistence decision. It is a **different layer** from the backend-down recovery work in #3537 underneath it: this is about *normal* in-memory serving, not disk fallback. The goal is parity with `OfferingsManager`, which has always done stale-while-revalidate. The parity question applies to any SDK that caches workflow details (iOS once it has the cache layer — it is a stage behind; no iOS action here).

### The change

`getWorkflow` is rewritten from a two-way (fresh → serve, everything-else → block-and-fetch) into a three-way branch mirroring `vendCachedOfferingsAndMaybeRefresh`:

| Cache state | Behavior | Changed? |
|---|---|---|
| **Fresh hit** | serve cached, no network | unchanged |
| **Stale hit** | serve cached immediately via `onSuccess`, **then** fire a background refresh that updates the cache only — no caller `onSuccess`/`onError`; a failed refresh is logged and swallowed | **this PR** |
| **Miss** | fetch from backend, deliver via `onSuccess`/`onError` | unchanged |

Mechanically: the fetch + resolve + cache + callback block was extracted into a private `fetchAndCacheWorkflow(...)` (pure extraction, no behavior change), and `getWorkflow` now dispatches into it from the three branches.

### Prefetch carve-out — the one deliberate exception

The split is expressed as a single parameter: `getWorkflow(..., staleWhileRevalidate: Boolean = true)`. Fresh and miss are identical across callers, so one boolean captures the difference cleanly (it only affects the stale-present case).

- **Render / on-demand path** (`PurchasesOrchestrator.getWorkflow`): `staleWhileRevalidate = true` (default).
- **Prefetch path** (`prefetchWorkflow`): `staleWhileRevalidate = false`.

Prefetch must NOT adopt SWR. Its job is to force a real refresh and **persist the detail envelope** (the #3537 backend-down-recovery feature). If prefetch served stale + background-refreshed, its stale case would stop blocking-and-persisting and weaken that guarantee. So prefetch keeps today's blocking-fetch-and-persist behavior on a stale workflow.

The background refresh inherits the caller's `persistEnvelopeOnResolve`. On the render path that is `false`, so the SWR background refresh does **not** persist — keeping this work cleanly out of the scope of the on-demand-persistence follow-up (#1).

### No new dedup

Two concurrent stale render calls can each fire a background refresh. This matches offerings — `fetchAndCacheOfferings` on the stale path isn't deduped either — so it is parity-acceptable. Documented in a code comment rather than guarded.

### Test coverage

- stale hit serves the cached value *before* the background refresh resolves
- stale hit fires exactly one background refresh that updates the cache
- stale hit does **not** surface a failing background refresh to the caller
- prefetch path blocks and persists on a stale workflow (no SWR)
- the existing "re-fetches when cache is stale" test stays green (backend still called), with intent/assertions updated to reflect that the stale value is now served first

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes paywall load timing when workflow cache is stale (immediate render vs blocking fetch); prefetch path is explicitly preserved but behavior diverges by caller.
> 
> **Overview**
> **`getWorkflow`** now matches offerings-style **stale-while-revalidate**: a **fresh** hit is unchanged; a **stale** in-memory hit returns the cached workflow immediately and kicks off a **background** refetch that only updates the cache (failures are logged, not delivered to the caller); a **miss** still blocks on the network.
> 
> The fetch/resolve/cache path is factored into **`fetchAndCacheWorkflow`**. A new **`staleWhileRevalidate`** flag (default **`true`** for on-demand paywall loads) controls the stale branch; **`prefetchWorkflow`** passes **`false`** so prefetch still **blocks**, refetches, and **persists detail envelopes** when the cache is stale.
> 
> Tests cover immediate stale serving, background cache updates, swallowed background errors, blocking behavior when SWR is off, and stale prefetch persistence. **`WorkflowsCache`** prune docs are clarified only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8509e19f7bd64f672b6eeaef7caef0099d29d98e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
@@ -20,6 +20,7 @@ import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 
+@Suppress("TooManyFunctions")
 internal class WorkflowManager(
     private val backend: Backend,
     private val workflowDetailResolver: WorkflowDetailResolver,
@@ -68,6 +69,27 @@ internal class WorkflowManager(
             onSuccess(cached)
             return
         }
+        fetchAndCacheWorkflow(
+            appUserID = appUserID,
+            workflowId = workflowId,
+            appInBackground = appInBackground,
+            callbackDispatcher = callbackDispatcher,
+            persistEnvelopeOnResolve = persistEnvelopeOnResolve,
+            onSuccess = onSuccess,
+            onError = onError,
+        )
+    }
+
+    @Suppress("LongParameterList")
+    private fun fetchAndCacheWorkflow(
+        appUserID: String,
+        workflowId: String,
+        appInBackground: Boolean,
+        callbackDispatcher: Dispatcher?,
+        persistEnvelopeOnResolve: Boolean,
+        onSuccess: (WorkflowDataResult) -> Unit,
+        onError: (PurchasesError) -> Unit,
+    ) {
         val onSuccessHandler: (WorkflowDetailResponse) -> Unit = { response ->
             scope.launch {
                 // resolve() can fail in several ways (missing inline data, CDN fetch/parse failures,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
@@ -43,9 +43,16 @@ internal class WorkflowManager(
     }
 
     /**
-     * Fetches and resolves a single workflow, serving a fresh in-memory cache hit when one exists
-     * and otherwise fetching from the backend.
+     * Fetches and resolves a single workflow using stale-while-revalidate, mirroring
+     * [com.revenuecat.purchases.common.offerings.OfferingsManager]'s offerings serving:
+     * a fresh cache hit is served directly; a stale-but-present hit is served immediately and the
+     * cache is refreshed in the background (no callbacks, failures logged); a miss fetches from the
+     * backend and delivers the outcome through [onSuccess]/[onError].
      *
+     * @param staleWhileRevalidate when true (the default, used by the on-demand render path), a
+     * stale-but-present cached workflow is served immediately with a background refresh. When false
+     * (the prefetch path), a stale workflow blocks on a full refetch instead — so prefetch keeps
+     * forcing a fresh fetch and persisting its envelope rather than serving a stale value.
      * @param persistEnvelopeOnResolve when true, also persists the raw detail envelope to disk after
      * a successful resolve, so it survives an app restart and can be re-resolved while the backend is
      * down (see [getWorkflowsList]'s recovery path). Only the prefetch path sets this: prefetched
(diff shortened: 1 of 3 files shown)


vegaro:
* **#3541** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3541?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3540** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3540?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3540?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3537** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3537?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
```diff
@@ -120,11 +129,71 @@ internal class WorkflowsCache(
...
+
+    /**
+     * Drops persisted envelopes whose workflowId is no longer in the latest list. This is the
+     * keyed-map equivalent of how [com.revenuecat.purchases.common.offerings.OfferingsCache] stays
+     * bounded by wholesale-replacing its single response blob: the persisted set always equals what
+     * the latest backend response says exists, so workflows the backend stopped sending don't
+     * linger on disk.
+     */
+    @Synchronized
+    private fun pruneWorkflowDetailEnvelopesToList(workflowIds: Set<String>) {
```
facumenzella, line 175: I think it might be worth a note here saying the store is per-API-key and not per-user. 

I guess its not a big deal, but for any PR about this topic makes me wonder if we should be more defensive.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -49,6 +61,7 @@ internal class WorkflowManager(
         onSuccess: (WorkflowDataResult) -> Unit,
         onError: (PurchasesError) -> Unit,
         callbackDispatcher: Dispatcher? = null,
+        persistEnvelopeOnResolve: Boolean = false,
```
facumenzella, line 72: why default false?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
```diff
@@ -120,11 +129,71 @@ internal class WorkflowsCache(
...
+            runCatching { WorkflowJsonParser.parseWorkflowDetailEnvelopes(cached) }
+                .onFailure { errorLog(it) { "Failed to restore workflow detail envelopes from disk cache" } }
+                .getOrNull()
+        }
+
+    /**
+     * Drops persisted envelopes whose workflowId is no longer in the latest list. This is the
+     * keyed-map equivalent of how [com.revenuecat.purchases.common.offerings.OfferingsCache] stays
+     * bounded by wholesale-replacing its single response blob: the persisted set always equals what
+     * the latest backend response says exists, so workflows the backend stopped sending don't
```
facumenzella: Only `only prefetch = true` workflows that successfully resolved are ever persisted, right?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: The kdoc explains the reasoning: on-demand fetches (not automatically prefetched) skip persistence to avoid unbounded disk growth. A user could open many distinct paywalls that are not prefetched, and there's no LRU cap today (will come in a future PR). The idea is to only store workflows that are prefetched.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
vegaro: Yes, exactly. Only `cacheWorkflowDetailEnvelope` writes to this store, and it is called only from the prefetch path after a successful resolve. Updated the `pruneWorkflowDetailEnvelopesToList` kdoc to make that invariant explicit.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsCache.kt
vegaro: Yes, the store is per-API-key, not per-user, but that’s safe here. `IdentityManager` calls `workflowsCache.clearCache()` on every logIn/logOut transition, which wipes the entire disk envelope store before the new user’s data lands. So the store always reflects the current user.

Looks like there's one existing race (a prefetch completing after the clear can repopulate the store with old data). But it looks like it's the same we have in offerings, so maybe something we can tackle in a future PR?

facumenzella: sure

facumenzella (approved):

PR #3543, opened 2026-06-05
Add Danger check to discourage large PRs

tonidero:
## Why

Large PRs are slow and error-prone to review, and problems are more likely to slip through. This adds a Danger check that nudges us toward smaller, more focused PRs.

## What

Fails the Danger check when a PR changes more than 300 lines of production Kotlin/Java code (insertions + deletions). Test sources and generated/build output are excluded so only "real" code counts.

When a large PR is genuinely necessary, add the `skip-pr-lines-changed-check` label to bypass the failure.

---

> [!NOTE]
> **Low Risk**
> CI-only Danger policy in Dangerfile; no runtime or app code changes, with an explicit bypass label for exceptions.
> 
> **Overview**
> Adds a **Danger CI gate** that **fails** pull requests when combined insertions and deletions in production **`.kt` / `.java`** files exceed **300 lines**.
> 
> Churn is summed only for non-test, non-generated paths (excludes `src/test`, `androidTest`, `testFixtures`, `build`, `generated`, and `*Test(s).kt/java`). PRs that must stay large can bypass with the **`skip-pr-lines-changed-check`** label, which turns the failure into an informational message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 95c2c7f83ed7fe66ccfcae04f5ede2b89928ff51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Dangerfile
```diff
@@ -24,3 +24,35 @@ revenuecatui_kt.each do |file|
...
+# Large PRs are hard to review well, so we cap the churn (insertions + deletions)
+# across real .kt/.java files, excluding test sources and generated/build output.
+# Authors who legitimately need a large PR can add the bypass label.
+PROD_LINES_LIMIT = 300
+SKIP_SIZE_LABEL = "skip-pr-lines-changed-check"
+
+prod_code_files = (git.modified_files + git.added_files).uniq.select do |f|
+  next false unless f.end_with?(".kt") || f.end_with?(".java")
+  next false if f.include?("/src/test/") || f.include?("/src/androidTest/") ||
+                f.include?("/src/testFixtures/")
```
tonidero, line 38: Note how in the end, decided to exclude tests... I do think tests are very important, but I'm afraid that if we include tests in the count, we might write less tests...

Dangerfile
```diff
@@ -24,3 +24,35 @@ revenuecatui_kt.each do |file|
...
+
+# Fail PRs that change too many lines of production Kotlin/Java code.
+# Large PRs are hard to review well, so we cap the churn (insertions + deletions)
+# across real .kt/.java files, excluding test sources and generated/build output.
+# Authors who legitimately need a large PR can add the bypass label.
+PROD_LINES_LIMIT = 300
+SKIP_SIZE_LABEL = "skip-pr-lines-changed-check"
+
+prod_code_files = (git.modified_files + git.added_files).uniq.select do |f|
+  next false unless f.end_with?(".kt") || f.end_with?(".java")
```
tonidero, line 36: Not including gradle files or anything like that... Lmk if you have any thoughts on this!

Dangerfile
```diff
@@ -24,3 +24,35 @@ revenuecatui_kt.each do |file|
     end
   end
 end
+
+# Fail PRs that change too many lines of production Kotlin/Java code.
+# Large PRs are hard to review well, so we cap the churn (insertions + deletions)
+# across real .kt/.java files, excluding test sources and generated/build output.
+# Authors who legitimately need a large PR can add the bypass label.
+PROD_LINES_LIMIT = 300
```
tonidero, line 32: Set the limit to 300 right now, which, without including tests, I think it's already pretty big... Lmk if we should aim for lower amounts.

Dangerfile
```diff
@@ -24,3 +24,35 @@ revenuecatui_kt.each do |file|
...
+SKIP_SIZE_LABEL = "skip-pr-lines-changed-check"
+
+prod_code_files = (git.modified_files + git.added_files).uniq.select do |f|
+  next false unless f.end_with?(".kt") || f.end_with?(".java")
+  next false if f.include?("/src/test/") || f.include?("/src/androidTest/") ||
+                f.include?("/src/testFixtures/")
+  next false if f.include?("/build/") || f.include?("/generated/")
+  next false if f.match?(/Tests?\.(kt|java)$/)
+  true
+end
```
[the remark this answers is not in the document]
tonidero: Hmm this might be ok? I guess it does depend on the file in the end and how much logic it has... But I think adding this might discourage refactors...

Dangerfile
rickvdl: Just checked a few PRs for reference, I think it's a reasonable start and let's tweak if needed :)

Dangerfile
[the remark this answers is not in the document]
rickvdl: That's fine I'd say, deletions are a lot easier to review as well.

Dangerfile
rickvdl: I see `/purchases/src/testDefaults/..` (same for CEC) wouldn't match the directory check, but it would match the files as long as they end with `Test(s).kt/.java` so I think we're good. Helper files etc would't be caught, but probably not the biggest deal.

rickvdl (approved):
Great to have this 💪

ajpallares (approved):
Thanks for adding this! Once included in all SDK repos, we should add the new `skip-pr-lines-changed-check` label to our internal `SDK GitHub labels` notion page, for reference

Dangerfile
AlvaroBrey, line 1: Nice! Single question here: have you considered adding this to the shared Dangerfile instead?

Especially since we already have `check_pr_size_increase` (bytes) there

AlvaroBrey (approved):

Dangerfile
tonidero: Good points! Those folders are the minority, since normally most code is part of the common source set. But it's true it's something we will need to keep an eye out in the future if we add more flavors.

Dangerfile
tonidero: I thought about it... My main reason to keep it split in each repo is that there are different files/folders we need to analyze on each SDK. Additionally, I think we might want to have a somewhat smaller diff max on some hybrids since most logic lives in the natives, ideally not such big PRs are needed in the hybrids. Of course, we can revisit this later if needed :)

PR #3545, opened 2026-06-05
Migrate base RulesEngineInternal operator unit tests to JSON predicate fixtures

ajpallares:
Android counterpart to RevenueCat/purchases-ios#6885

### Motivation
Migrate the hand-written `RulesEngineInternal` operator unit tests into declarative JSON predicate fixtures run by one parameterized runner.

### Description
- Adds per-operator JSON fixtures (exact copies of the RevenueCat/purchases-ios#6885 fixtures) under `src/test/resources/predicate-fixtures`, run by a shared parameterized runner. A guard test pins the loaded count to make sure we're not missing any test case.
- Deletes the migrated arithmetic/comparison/equality/logic suites; trims accessor/evaluator/string-array to the cases not expressible as predicate→boolean.
- Test-only; no production changes.

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Low Risk**
> Test-only refactor with no production code changes; risk is limited to possible coverage gaps if fixtures diverge from deleted Kotlin tests.
> 
> **Overview**
> Moves most **RulesEngineInternal** operator and evaluator coverage from large Kotlin unit test classes into **declarative JSON fixtures** under `src/test/resources/predicate-fixtures/`, aligned with the iOS PR.
> 
> A **parameterized** `PredicateFixtureTest` loads every fixture via `PredicateConformanceFixtureLoader` (kotlinx.serialization) and asserts outcomes through `PredicateConformanceRunner`—expected truthiness, `RuleError` types, and optional warning substrings. `PredicateFixtureLoaderTest` pins **244** loaded cases so a broken resource load cannot shrink the suite silently.
> 
> **Removed** standalone suites for arithmetic, comparison, equality, and logic; **slimmed** `EvaluatorTest`, `AccessorOperatorsTest`, and `StringArrayOperatorsTest` to cases that cannot be expressed as predicate→boolean fixtures (malformed JSON parse, array-scoped `var`, structural `merge` nesting, whole-data `var` results).
> 
> **Test-only** Gradle change: Kotlin serialization plugin plus `kotlinx.serialization.json` on the test classpath. No production rules engine code changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53b9e92843ba7fb9e3a0caf65cda459c52c97b7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/helpers/PredicateConformanceRunner.kt
```diff
@@ -0,0 +1,66 @@
...
+        if (expected.contains.isEmpty()) {
+            assertThat(warnings)
+                .withFailMessage("Fixture %s: expected no warnings, got %s", id, warnings)
+                .isEmpty()
+            return
+        }
+        expected.contains.forEach { substring ->
+            assertThat(warnings)
+                .withFailMessage("Fixture %s: missing warning containing \"%s\", got %s", id, substring, warnings)
+                .anyMatch { it.contains(substring) }
```
rickvdl, line 63: Could it be that we're not failing if we have the expected result + other results? In other words, is it exhaustive?

rickvdl (approved):
Great work! One question

rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/helpers/PredicateConformanceRunner.kt
ajpallares: That's right. My first iteration of this was indeed to make it exhaustive. However, I changed it intentionally. My reasoning was that if a specific test is checking that a specific warning is logged, then other logs that are independent from that test could be added/removed and that shouldn't affect that test passing.

rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/helpers/PredicateConformanceFixture.kt
```diff
@@ -0,0 +1,116 @@
...
+        return File(url.toURI())
+    }
+
+    private fun loadCases(file: File): List<PredicateConformanceFixtureCase> {
+        val root = JSONTokener(file.readText()).nextValue() as JSONObject
+        val fixtures = root.getJSONArray("fixtures")
+        return (0 until fixtures.length()).map { parseCase(fixtures.getJSONObject(it)) }
+    }
+
+    private fun parseCase(json: JSONObject): PredicateConformanceFixtureCase =
```
tonidero: I guess we could make `PredicateConformanceFixtureCase` serializable and use that... But TBH, I don't really mind. Happy to keep as is :)

tonidero (approved):
Looks great!

rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/helpers/PredicateConformanceRunner.kt
rickvdl: Alright, that also makes sense. Thanks!

rules-engine-internal/src/test/kotlin/com/revenuecat/purchases/rules/helpers/PredicateConformanceFixture.kt
ajpallares: Right. I was doubting about this but Cursor told me there wasn't really a lot of benefit. Your comment was the last push I needed, so I migrated it in https://github.com/RevenueCat/purchases-android/pull/3545/commits/53b9e92843ba7fb9e3a0caf65cda459c52c97b7e

I also like it better like this, being more similar to iOS

PR #3547, opened 2026-06-05
other(workflows): reproduce late workflow envelope persistence

facumenzella:
Adds a focused failing regression test for PR #3537.

The test simulates a workflow-list prefetch starting, then `WorkflowsCache.clearCache()` running before the detail callback completes. The late detail callback should not write the old workflow detail envelope back to `DeviceCache`, because that makes a previous user/session envelope durable again after the cache clear.


facumenzella:
closing due to staleness

PR #3548, opened 2026-06-05
[codex] Reproduce stale workflow envelope restore

facumenzella:
Adds a focused failing regression test for PR #3537.

The test simulates a backend-down restore where the cached workflows list still contains `wf_1`, but now with `prefetch=false`, while an older persisted detail envelope for `wf_1` remains on disk. The restore path should not re-resolve that stale envelope into the in-memory workflow cache.

Verification:
- `rtk ./gradlew :purchases:testDefaultsBc8DebugUnitTest --tests "com.revenuecat.purchases.common.workflows.WorkflowManagerTest.list onError does not restore a persisted envelope for a workflow no longer marked prefetch"`
- Fails with `WorkflowManagerTest > list onError does not restore a persisted envelope for a workflow no longer marked prefetch FAILED` at `WorkflowManagerTest.kt:1280`, where the test expects `workflowsCache.cachedWorkflow("wf_1")` to remain null.


facumenzella:
Closing

PR #3550, opened 2026-06-05
Preview paywalls via deep link

davedelong:
### Checklist

Motivation
This changes introduces a new API to allow customers to handle a deep link intent that will result in a paywall being presented.

Description
Adds a new method to the Purchases type in RevenueCatUI that, when given a URL, will attempt to extract offering and paywall identifiers from the value. If successful, it will present the corresponding offering's paywall for the developer to preview.

Included tests exercise the parsing and validation around URL recognition, as well as the logic for presenting the paywall.

---

> [!NOTE]
> **Medium Risk**
> New public entry point that asynchronously fetches offerings and starts PaywallActivity; mitigated by strict URL/paywall validation and activity lifecycle checks, but still affects paywall presentation flow.
> 
> **Overview**
> Adds **Preview Paywall deep link** support on Android so apps can open a published paywall from dashboard preview URLs (aligned with iOS).
> 
> Developers call **`Purchases.presentPaywall(intent, activity)`** from something like `onNewIntent`. The extension delegates to an internal **`PaywallPreviewPresenter`** that recognizes `{scheme}://rc-paywall-preview?offering_id=…&paywall_id=…`, returns `false` immediately when the URL is invalid, and otherwise loads offerings, checks that the paywall id matches the offering, skips launch if the activity is finishing/destroyed, then starts **`PaywallActivity`** with a dismiss button.
> 
> The public surface is recorded in **`api.txt`**. **`PaywallPreviewPresenterTest`** covers URL parsing, fetch failures, id mismatches, and lifecycle guards via injectable **`locateOffering`** / **`launchPaywall`** hooks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80defa85993c692e25b26d93dcc65fae0e543337. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
```diff
@@ -0,0 +1,143 @@
...
+     */
+    private fun parseLink(intent: Intent): PreviewLinkParams? {
+        val uri = intent.data?.takeIf { it.host == PAYWALL_PREVIEW_HOST } ?: return null
+
+        val queryParamCount = uri.queryParameterNames.size
+        val offeringId = uri.getQueryParameter("offering_id")?.takeIf { it.isNotBlank() }
+        val paywallId = uri.getQueryParameter("paywall_id")?.takeIf { it.isNotBlank() }
+
+        return when {
+            queryParamCount != 2 -> {
```
tonidero: I would make this more flexible and support > 2, so, if we want to add new parameters in the future (without removing existing ones), older clients can keep working. I guess it's not a huge deal if this is meant for developers only... but probably good to still future-proof it?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @param intent The [Intent] received by your Activity.
+ * @param activity The [Activity] to present the paywall from. Defaults to the currently resumed
+ * Activity tracked by the SDK.
+ * @return `true` if the intent was a valid Preview Paywall link and handling has begun;
+ * `false` if the intent is not a valid paywall preview link.
+ */
+@MainThread
+@OptIn(InternalRevenueCatAPI::class)
+public fun Purchases.presentPaywall(intent: Intent, activity: Activity? = null): Boolean {
```
tonidero: One random thought... For the WebPurchaseRedemption case, we split the handling of deep links into 2:
- First the [parsing](https://github.com/RevenueCat/purchases-android/blob/4b64a99cefcee8d91bc283c93af8abd23a6ed239/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt#L1213), which returns an object indicating whether there is a valid web purchase redemption link
- The [actual](https://github.com/RevenueCat/purchases-android/blob/4b64a99cefcee8d91bc283c93af8abd23a6ed239/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt#L1195) handling of that link 

I think this pattern can potentially be much more flexible, and I would actually expect that some devs would like to have this flexibility, to customize their app when a paywall is to be presented.

Having said that, I do think if the only use case of this is to preview paywalls by developers and not to allow developers to use this with customers, this is probably ok so FFTI! :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
```diff
@@ -0,0 +1,143 @@
...
+                Logger.w("Invalid rc-paywall-preview link. Expected 2 parameters, but found $queryParamCount")
+                null
+            }
+            offeringId == null -> {
+                Logger.w("Invalid rc-paywall-preview link: Bad offering_id parameter")
+                null
+            }
+            paywallId == null -> {
```
tonidero, line 119: We might want to sync with the workflows team... since we might need to change this parameter name later with workflows, so maybe this should be a soft warning but still continue? In the end, I think this only is used to perform some validations?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
```diff
@@ -0,0 +1,51 @@
...
+ * attempts to present that paywall for previewing.
+ *
+ * The [activity] parameter is optional. If omitted, the SDK uses the currently resumed Activity
+ * it tracks automatically via [ActivityLifecycleCallbacks]. If no Activity can be found
+ * (e.g. the app is backgrounded), a warning is logged and the method returns `false`.
+ *
+ * Expected deep link format:
+ * `{yourScheme}://rc-paywall-preview?offering_id={ID}&paywall_id={ID}`
+ *
+ * Example — minimal (let the SDK discover the Activity):
```
tonidero: Hmm, I'm wondering, should we make the activity a required parameter? I think for most devs, getting the android activity from where they get the intent should be straightforward and it would avoid us having to keep track of the activity... It would also allow us to fetch the intent directly I think.

I can see this more useful for hybrids though... Though most hybrids have ways to fetch the current activity we could use, so just want to make sure I understand the use case for this 🙏

tonidero (commented):
Looking pretty good! The only thing I'm not sure about is the need to keep the current activity ourselves... I think for most cases in Android, it's fine to require the activity as a parameter but just making sure I understand :)

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
davedelong: i was originally planning on support more than two, but i figured that if this ever allows more than two parameters (for example, specifying unpublished and/or draft versions), it could be confusing to scan a qr code for a paywall and see the _wrong_ paywall in your app (because the code in your app wouldn't know how to handle a hypothetical `&draft=true` parameter). so i erred on the side of doing _nothing_ instead of doing the wrong thing

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
davedelong: yep, i completely agree in principle. i chose this current approach for two reasons:

1. the web purchase redemption stuff is in the non-ui module, which makes that sort of thing easier. doing that in this case would require the type definition in the non-ui package and then the handler in the UI package. for current concision, i decided to err on the side of more localized changes, especially since we haven't decided on a deep link handling API yet
2. the code is trivially refactorable later to the approach you describe

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
davedelong: yeah, this is basically just a sanity check value

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
davedelong: This pattern is a translation of the corresponding code in iOS, where (because of SwiftUI) it's not always easy to have a ready handle to your `UIWindow` or `UIViewController` to present from, so I allow `nil` there and attempt to pick a sensible default.

tonidero: Right, in Android Jetpack compose (and also the older view system) is relatively easy to get an instance of the current activity. As I mentioned, I do think this might make it easier to implement in some hybrids, but each hybrid usually has a mechanism to get the activity as well, so I'm just wondering it's worth to keep references to the activities, even as weak in the main SDK vs using each hybrid technology...

I kinda would lean towards making it mandatory in Android for now and we can make it optional later if we find it's hard on the hybrids 🙏. This way, it's also easier to iterate without causing breaking changes.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
tonidero: I can see your point... But I do wonder, if we add, for example, a tracking parameter later, (say so we know from what dashboard/user it came from) it would break this and I think it's possible we will add that... So I'm still wondering if we should be more lenient on this... But I won't fight it.

davedelong: I relaxed this to allow additional query values

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
davedelong: The activity is now required ✅

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -68,6 +68,14 @@ public class Purchases internal constructor(
     public val currentConfiguration: PurchasesConfiguration
         get() = purchasesOrchestrator.currentConfiguration
 
+    /**
+     * The currently resumed Activity, if any. Tracked automatically via ActivityLifecycleCallbacks.
+     * Used internally by RevenueCatUI to present paywalls without requiring an explicit Activity parameter.
+     */
+    @InternalRevenueCatAPI
+    public val currentActivity: Activity?
```
tonidero: I think this new API is now unused and we should probably remove it?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -164,6 +165,23 @@ internal class PurchasesOrchestrator(
     val adTracker: AdTracker = AdTracker(adEventsManager),
 ) : LifecycleDelegate, CustomActivityLifecycleHandler {
 
+    // All activities that have started but not yet stopped, in start order. Using a list rather
+    // than a single ref so that stopping a dialog/translucent overlay (which never stops the
+    // host behind it) correctly leaves the host in the set.
+    private val startedActivityRefs = mutableListOf<WeakReference<Activity>>()
```
tonidero: Same for all this state/logic.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
```diff
@@ -0,0 +1,39 @@
...
+ * }
+ * ```
+ *
+ * @param intent The [Intent] received by your Activity.
+ * @param activity The [Activity] to present the paywall from.
+ * @return `true` if the intent was a valid Preview Paywall link and handling has begun;
+ * `false` if the intent is not a valid paywall preview link.
+ */
+@MainThread
+public fun Purchases.presentPaywall(intent: Intent, activity: Activity): Boolean {
```
tonidero, line 33: Nitpick: I wonder if we shoudl rename to `presentPaywallIfNeeded`... but NABD

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
```diff
@@ -0,0 +1,126 @@
...
+     * been started; `false` if any synchronous URL check fails.
+     */
+    fun handle(
+        locateOffering: suspend (offeringId: String) -> Offering?,
+        intent: Intent,
+        activity: Activity,
+    ): Boolean {
+        val params = parseLink(intent) ?: return false
+
+        MainScope().launch {
```
tonidero, line 63: It could be worth adding a `Purchases.isConfigured` check before this, so we only try to use this with a configured SDK early in the process. And if not, maybe even log some warning/error?

tonidero (commented):
I like that making the activity mandatory makes it much simpler :) Just a couple of suggestions, but I think once we remove the now unused code, we should be ready to go!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
davedelong: This currently only gets called from an instance of `Purchases`, which means it _must_ be configured at that point, right?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
davedelong: I avoided the "IfNeeded" nomenclature because the word "need" implies to me some sort of decision-making process about whether the user might have the corresponding entitlement already. This is an unconditional presentation, as long as the intent can be parsed.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallPreviewPresenter.kt
tonidero: Ah indeed good point... Probably fine then yeah! Please ignore!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PurchasesPaywallPreviewExtensions.kt
tonidero: Right... I think that's fine indeed 👍

tonidero (approved):
I think this looks good!

PR #3552, opened 2026-06-07
Add JSON Logic min and max operators

ajpallares:
### Checklist

### Motivation
Adds `min`/`max` so audience predicates can compare against numeric extrema. Resolves SDK-4347. Android counterpart of RevenueCat/purchases-ios#6825.

### Description
- Variadic flat-list shape per json-logic-js; operands coerce via JS `Number()` (`Value.toNumberOrNull`, like `-`/`/`/`%`, not `parseFloat` as in `+`/`*`); no iteration/projection
- Empty input → ±∞, mirroring `Math.min()` / `Math.max()`
- Non-numeric operands → NaN poisoning so a malformed predicate fails closed against a numeric threshold
- Coverage is the shared declarative JSON fixtures (`min.json` / `max.json`, byte-identical to iOS), run by the existing fixture runner

---

> [!NOTE]
> **Medium Risk**
> Changes predicate evaluation semantics for audience rules; malformed operands intentionally yield NaN/±∞ so thresholds fail closed, but new operators can alter which users match existing predicates once deployed.
> 
> **Overview**
> Adds JSON Logic **`min`** and **`max`** so audience predicates can compute numeric extrema over flat operand lists, aligned with **`Math.min` / `Math.max`** (Android counterpart to iOS).
> 
> Implementation lives in new **`MinMaxOperators`**: operands are evaluated via **`evalArgs`**, coerced with **`Value.toNumberOrNull`** (same path as `-` / `/` / `%`), folded with NaN poisoning, and empty args return **±∞**. **`Operators.dispatch`** routes `"min"` and `"max"` to these handlers.
> 
> Conformance is covered by shared **`min.json`** / **`max.json`** fixtures (coercion, empty args, fail-closed thresholds, vars); **`PredicateFixtureLoaderTest`** expected count is bumped **273 → 307**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6940dadb5a385562a85fb047c2050630926a06ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine-internal/src/test/resources/predicate-fixtures/max.json
```diff
@@ -0,0 +1,326 @@
...
+            ]
+          }
+        ]
+      },
+      "variables": {},
+      "expected": true
+    },
+    {
+      "id": "max_of_empty_args_is_negative_infinity",
+      "description": "{\"max\": []} is -Infinity (matches JS Math.max()); -Infinity is produced via division-by-zero since JSON has no Infinity literal",
```
tonidero, line 213: I do wonder if we should have a way to represent these values (like infinity or NaN) in json, so once parsed into Kotlin, we can perform the test directly, instead of using these "similar behavior" tests, which seem a bit fragile...

tonidero (approved):
I think it looks good! Just an idea to try to improve the tests, but can be applied separately.

rules-engine-internal/src/test/resources/predicate-fixtures/max.json
ajpallares: Very good point. I'll investigate this! There must be something to avoid these workarounds that, as you say, can be fragile. Thanks for the suggestion!

ajpallares: JSON itself has no `Infinity`/`-Infinity`/`NaN` tokens, they're not valid JSON, so they can't appear literally 😓 

So to add them, we'd have to pick some normal JSON value and agree "this one secretly means infinity." But every candidate is already used for something real in a rule:
- If we said the text `"Infinity"` means ∞, we'd lose the ability to use `"Infinity"` as an ordinary piece of text, since the engine could no longer tell whether you meant the word or the number.
- If we said a special object like `{"$float": "Infinity"}` means ∞, that clashes with how objects are normally read: an object in a rule is treated as an instruction/expression, not a plain value.

I don't think we should repurpose any JSON value for this, because any "shortcut" would change the meaning of inputs that are already valid. So I can only just calculate the value instead (e.g. `{"/": [-10, 0]}` gives `-Infinity`).

ajpallares: **Proposal: a self-documenting way to write `±Infinity` in fixtures**

Instead of computing the value with `{"/": [-10, 0]}`, the test harness could seed a few well-known constants into the variables map, in code:

```swift
// Swift (test harness only)
"+Infinity": .float(.infinity),
"-Infinity": .float(-.infinity),
```

```kotlin
// Kotlin (test harness only)
"+Infinity" to Value.FloatValue(Double.POSITIVE_INFINITY),
"-Infinity" to Value.FloatValue(Double.NEGATIVE_INFINITY),
```

Then a fixture reads much more clearly:

```json
{ "===": [ { "max": [] }, { "var": "-Infinity" } ] }
```

**Important:** these constants would be injected only by the test harness (they wouldn't be part of the engine and would never exist in real rules/data).

**Tradeoffs:**
- The "magic" now lives in test code on **both** platforms (Swift + Kotlin) and has to be kept identical.
- It reserves the variable names `+Infinity` / `-Infinity` for every fixture.

Note this only helps `±Infinity`; `NaN` can't be asserted by equality anyway, so it still should still use the `x != x` trick to avoid false positives.

WDYT @tonidero?

tonidero: Hmm right... the alternative would be to run these tests over something that supports outputs other than booleans... But yeah that might be trickier... This could be a good compromise I think, as long as we make sure this won't affect the real rule engine as mentioned 🙏

ajpallares: Sure! Will do this in separate PRs for both platforms
iOS: https://github.com/RevenueCat/purchases-ios/pull/6936
Android: https://github.com/RevenueCat/purchases-android/pull/3556

PR #3557, opened 2026-06-08
Narrow consumer R8 rules

tonidero:
## Why

The SDK's consumer Proguard rules kept the entire SDK:

```
-keep class com.revenuecat.** { *; }
```

This means that all the code of the SDK is kept and included in our customers apps, which is not great. This PR attempts to reduce what we keep to a few strategic locations that might be more dangerous. It's possible that this can be reduced further, but we are taking a conservative approach for now.

There were a couple of changes required on hybrids:
- purchases-unity: https://github.com/RevenueCat/purchases-unity/pull/949
- cordova-plugin-purchases: https://github.com/RevenueCat/cordova-plugin-purchases/pull/1047

Other hybrids seem to work fine with these changes

## What's still kept

- **Enum constant names** — `EnumDeserializerWithDefault` matches JSON against enum constant names read reflectively and *silently falls back to a default* on mismatch, so an obfuscated name would be a silent wrong-value bug, not a crash.
- **Galaxy reflective constructor** (`feature/galaxy`) — `GalaxyBillingWrapper` is instantiated via `Class.forName` + `getDeclaredConstructor`.
- **`PaywallKt` existence probe** (`ui/revenuecatui`) — `canUsePaywallUI` checks for the class via `Class.forName`.

---

> [!NOTE]
> **High Risk**
> Shrinking/obfuscation behavior for all SDK consumers changes materially; mis-keep rules can cause subtle runtime failures (reflection, serialization, Galaxy) that only show up in release/minified apps.
> 
> **Overview**
> Replaces the SDK’s blanket **keep entire `com.revenuecat.**`** consumer ProGuard rule with **targeted keeps** (enum constant names for reflective JSON matching, `Parcelable` `CREATOR`, plus Galaxy’s reflective `GalaxyBillingWrapper` constructors). Drops the broad kotlinx.serialization companion/serializer keep block while retaining the existing ClassValue workarounds.
> 
> **`EnumDeserializerWithDefault`** now takes an explicit stable **`serialName`** for every call site so obfuscated enum class names cannot collide in kotlinx.serialization descriptors or silently map JSON to the wrong default.
> 
> **CI / Maestro** builds a **minified release** e2e APK to exercise consumer rules end-to-end: most flows run on that build; workflows that need **`configureForE2ETests`** are tagged **`debug-only`** and run on the debug APK. The e2e app enables release minification, uses debug signing, and opts into **`DangerousSettings.forTestStoreInReleaseBuild()`** so Test Store keys work in non-debuggable builds for internal testing only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 403e8cdc29c0955206eeb17a55b90039e27b094c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/consumer-rules.pro
```diff
@@ -1,4 +1,19 @@
--keep class com.revenuecat.** { *; }
+# Keep public API names stable for consumers; members may still be shrunk/optimized by R8.
+-keepnames class com.revenuecat.purchases.** { *; }
```
JayShortway: Why is this needed? They're only obfuscated when the app is compiled right?

purchases/consumer-rules.pro
```diff
@@ -1,4 +1,19 @@
...
+
+# Keep all @Serializable model classes intact so kotlinx (de)serialization and the SDK's custom
+# KSerializers (polymorphic/sealed hierarchies, surrogates) keep working under minification.
+-keep @kotlinx.serialization.Serializable class com.revenuecat.** { *; }
```
JayShortway: Doesn't `kotlinx.serialization` ship with the required rules already? Or is this only for our custom serializers?

ui/revenuecatui/consumer-rules.pro
```diff
@@ -1,2 +1,7 @@
 -dontwarn com.emergetools.snapshots.annotations.IgnoreEmergeSnapshot
 -dontwarn com.emergetools.snapshots.annotations.EmergeSnapshotConfig
+
+# Existence is probed via Class.forName from the purchases module (canUsePaywallUI in
+# common/utils.kt) to detect whether the paywalls UI module is on the classpath. Force-keep the
+# file class so the probe resolves even if the app doesn't otherwise reference it directly.
+-keep class com.revenuecat.purchases.ui.revenuecatui.PaywallKt
```
JayShortway: Alternatively we can annotate with `@Keep`, although I don't know if that works for a class generated from a Kotlin file that only contains toplevel functions.

JayShortway (commented):
Almost can't believe it 😭

purchases/consumer-rules.pro
tonidero: I... think it's not needed indeed... Originally added it to keep some places where we use `class.simpleName` in logs... But those cases are usually there just for completeness and log some error, so I don't think it's needed to keep the original names (This would still allow shrinking just keeping the names, but yeah probably not needed 👍 )

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -29,6 +29,7 @@ public enum class WorkflowTriggerType {
 }
 
 internal object WorkflowTriggerTypeDeserializer : EnumDeserializerWithDefault<WorkflowTriggerType>(
+    serialName = "com.revenuecat.purchases.common.workflows.WorkflowTriggerType",
```
tonidero, line 39: So basically we need each deserializer to have a different name, and after minifiying, the previous strategy didn't work so I'm now passing the FQN as a parameter... In theory, any different string could work, but with the FQN we make sure to avoid duplicates.

purchases/src/main/kotlin/com/revenuecat/purchases/DangerousSettings.kt
```diff
@@ -39,5 +41,17 @@ public class DangerousSettings internal constructor(
...
+
+        /**
+         * Creates a [DangerousSettings] that allows configuring the SDK with a Test Store API key
+         * in a release (non-debuggable) build, bypassing the safety check that otherwise blocks it.
+         *
+         * For RevenueCat-internal end-to-end testing only. Do not use in production apps.
+         */
+        @InternalRevenueCatAPI
+        @JvmStatic
+        public fun forTestStoreInReleaseBuild(): DangerousSettings = DangerousSettings(
```
tonidero, line 67: I do wonder if we would want to end up making this public... But for now, kept it internal

ui/revenuecatui/consumer-rules.pro
tonidero: I kinda prefer to keep the rules here... But it might indeed be my preference. Happy to change it if you feel strongly

JayShortway: No not at all!

ui/revenuecatui/consumer-rules.pro
```diff
@@ -6,3 +6,8 @@
 -keep class com.revenuecat.purchases.ui.revenuecatui.paywalls.PaywallAssetWarmerImpl {
     <init>();
 }
+
+# Existence is probed via Class.forName from the purchases module (canUsePaywallUI in
+# common/utils.kt) to detect whether the paywalls UI module is on the classpath. Force-keep the
+# file class so the probe resolves even if the app doesn't otherwise reference it directly.
+-keep class com.revenuecat.purchases.ui.revenuecatui.PaywallKt
```
AlvaroBrey: No longer needed, I moved it to a ServiceLoader in https://github.com/RevenueCat/purchases-android/pull/3913/changes#diff-a003ed020cddef2c784a374cd886a7745171eafc3f8f56049ec7a756e92db2e4L56

(the rule just above this one in the diff)

.circleci/config.yml
```diff
@@ -830,12 +830,12 @@ jobs:
           install: true
       - android/start_emulator:
           avd_name: test-e2e-tester
-          post_emulator_launch_assemble_command: ./gradlew test-apps:e2etests:assembleDebug
+          post_emulator_launch_assemble_command: ./gradlew test-apps:e2etests:assembleRelease
```
AlvaroBrey: should we move the maestro workflow tests job to release builds too?

purchases/consumer-rules.pro
```diff
@@ -1,58 +1,24 @@
--keep class com.revenuecat.** { *; }
+# EnumDeserializerWithDefault matches JSON against enum CONSTANT NAMES read reflectively via
+# Class.enumConstants (value.name.lowercase()) and silently falls back to a default on mismatch.
+# Keep enum constant names so an obfuscated name can't turn into a silent wrong-value bug.
+-keepclassmembers enum com.revenuecat.** {
+    <fields>;
+    public static **[] values();
+    public static ** valueOf(java.lang.String);
+}
```
AlvaroBrey, line 8: Should  we add parcelable too?

```
 -keepclassmembers class com.revenuecat.** implements android.os.Parcelable {
     public static final ** CREATOR;
 }
```

It's in `getDefaultProguardFiles()`  (which is not guaranteed) and right now we're also safe because we publish it it from a transitive dep (`play-services-basement`), but it's better to be explicit IMO

ui/revenuecatui/consumer-rules.pro
tonidero: Ohh good point! I will do that 👍

.circleci/config.yml
tonidero: Oh well... actually they were unified in this other PR https://github.com/RevenueCat/purchases-android/pull/3912, but now I do need to rebase 😅 Will do so

purchases/consumer-rules.pro
tonidero: Right... Good call 👍 Better to be explicit indeed

ui/revenuecatui/consumer-rules.pro
tonidero: Removed in https://github.com/RevenueCat/purchases-android/pull/3557/commits/bdce965d8d829c42b81a34638dc7f8df488cd0b9

purchases/consumer-rules.pro
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3557/commits/bdce965d8d829c42b81a34638dc7f8df488cd0b9

.circleci/config.yml
tonidero: So this was actually a bit tricky 😅 Some tests use a debug-only configuration API, so can't be run on release... So for now, I've added that as separate steps within the same job, to run debug-only tests in https://github.com/RevenueCat/purchases-android/pull/3557/commits/403e8cdc29c0955206eeb17a55b90039e27b094c

AlvaroBrey (approved):
:shipit:

test-apps/e2etests/build.gradle.kts
```diff
@@ -40,8 +40,12 @@ android {
 
     buildTypes {
         release {
-            isMinifyEnabled = false
+            // Minified so the Maestro e2e flow exercises the SDK's consumer R8 rules end-to-end.
+            isMinifyEnabled = true
```
vegaro, line 44: great

vegaro (approved):
looks great to me

PR #3559, opened 2026-06-08
fix(paywalls): clamp negative margins/padding to zero

facumenzella:
## Description

Android does not support negative padding, and the RevenueCatUI SDK applies component **margins** via Compose's `Modifier.padding(...)` as well. The Paywalls dashboard can send negative `top` / `bottom` / `leading` / `trailing` values, which render incorrectly on Android.

This clamps negative values to `0` at the single conversion choke point, `Padding.toPaddingValues()`, so both margins and padding are protected. Because clamping happens before the value is stored in component styles/state, any downstream consumers that compute from the resulting `PaddingValues` (e.g. `ButtonComponentView`'s `margin.calculateTopPadding()`) get clamped values for free.

When clamping occurs, a warning is logged so the dashboard misconfiguration is visible. The warning is opt-in via a `warnIfNegative` flag that only `StyleFactory` passes (once per paywall build); the conversion itself is pure, so hot recompute paths (`derivedStateOf`) clamp silently and never spam the same warning on tab/package/size/variable changes.

### Why this layer
- `toPaddingValues()` is the one place every margin and padding flows through (both `StyleFactory` base styles and per-component partial overrides).
- Not clamping in the `Padding` model itself: that is the serialized source of truth, and clamping there would corrupt round-trips.

### Scope note
The ask was about margins specifically, but this choke point covers both margins and padding. There are zero intentional negative padding/margin values anywhere in source or JSON fixtures (only `0` and positives), and negative padding is unsupported on Android regardless, so clamping both is correct behavior rather than scope creep, and avoids a future footgun where a new margin call site forgets to clamp.

## Tests
- Added `PaddingTest` next to the conversion (the existing `PaddingTests` lives in `:purchases` and tests the model, not the conversion). Covers positive pass-through, all-negative clamping, mixed, and clamping with `warnIfNegative` enabled. Runs under Robolectric so `android.util.Log` (used by the warning) is available.
- Added a `StackComponentView` preview with negative padding and margin so the clamped rendering is visually validated via Emerge snapshots.

<details>

# AI Context

## Metadata

- PR: 3559
- Branch: facundo/clamp-negative-margins
- Agent(s): Claude Code (Opus 4.8, 1M context)
- Generated: 2026-06-10
- Context document version: 3

## Goal

Clamp negative spacing values sent by the Paywalls dashboard to zero on Android, since Android does not support negative padding and margins are applied via `Modifier.padding(...)`. Focus on margins.

## Initial Prompt

"Apparently we don't support negative padding in android, but the dashboard can send it. Can we clamp those? Margins specifically."

- "open a draft and let's take a look" — open a draft PR.
- "let's add the debug log" + "and let's add the preview" — address reviewer suggestions.
- Pointed at Cursor Bugbot comment re: repeated warnings on state refresh — gate the warning so it only fires once at build time, not on recompose.

## Agent Contribution

- Located the conversion choke point `Padding.toPaddingValues()` and mapped all margin/padding call sites (StyleFactory once-per-build, per-component `derivedStateOf` recompute paths, view application sites).
- Verified no intentional negative padding/margin exists in Kotlin source or JSON fixtures.
- Implemented clamping via `coerceAtLeast(0.0)` on all four axes (unconditional).
- Added an opt-in `warnIfNegative` flag; only StyleFactory's 17 conversion sites pass it, so the warning fires once per paywall build and the `derivedStateOf` paths stay silent.
- Wrote unit tests (`PaddingTest`) under Robolectric; added a `StackComponentView` negative-padding preview.
- Ran unit tests + detekt (passing), rebased onto main, replied to all review threads.

## Human Decisions

- Directed scope toward margins; agent confirmed (with evidence) that clamping at the shared conversion covers both safely.
- Reviewer (tonidero) approved and suggested logging + a negative-padding preview. Cursor Bugbot flagged repeated warnings on recompose. All addressed.

- Decision: Clamp negatives in `Padding.toPaddingValues()`; gate the warning behind an opt-in `warnIfNegative` flag used only by StyleFactory.
  - Rationale: Single choke point for clamping; logging only at the once-per-build ingest point avoids re-emitting the warning when partial overrides recompute in `derivedStateOf`.
  - Rejected: Logging unconditionally inside `toPaddingValues()` (spams on state changes, per Cursor); clamping in the `Padding` model (corrupts serialized round-trips); clamping at the ~20 `.padding(...)` application sites (footgun); a process-wide dedupe set (mutable global state).

## Files / Symbols Touched

- `ui/revenuecatui/.../components/ktx/Padding.kt`
  - Why: clamp negatives (always) + opt-in warning
  - Symbols: `Padding.toPaddingValues(warnIfNegative: Boolean = false)`
  - Review relevance: clamping is unconditional; `warnIfNegative` only gates the log
- `ui/revenuecatui/.../components/style/StyleFactory.kt`
  - Why: pass `warnIfNegative = true` at the 17 once-per-build conversion sites
  - Review relevance: confirm only build-time sites opt in (not `derivedStateOf`)
- `ui/revenuecatui/.../components/ktx/PaddingTest.kt`
  - Why: positive, all-negative, mixed, and warnIfNegative coverage; Robolectric for `Log`
  - Symbols: `PaddingTest`
- `ui/revenuecatui/.../components/stack/StackComponentView.kt`
  - Why: preview validating clamped negative padding/margin
  - Symbols: `StackComponentView_Preview_NegativePaddingAndMarginClamped`

- None.

## Validation

- Commands run:
  - `./gradlew :ui:revenuecatui:testDefaultsBc8DebugUnitTest --tests "...ktx.PaddingTest"`: BUILD SUCCESSFUL
  - detekt (pre-commit): passed
- Manual verification:
  - Not run (no on-device rendering performed locally).
- CI:
  - Emerge snapshot: 1 added (the new preview), 0 modified, 0 errored — confirms no existing snapshot changed. New snapshot needs approval in the Emerge dashboard.
  - Other CI green on prior commit; re-running on the latest commit.

## Validation Gaps

- Local Paparazzi verification unavailable: `ui/revenuecatui/src/test/snapshots` is git-ignored/absent in this checkout; snapshots run on CI / the snapshots repo.
- The opt-in warning does not fire for negative values that come only from a partial/conditional override (those convert on the recompute path, which intentionally stays silent). Base component padding/margin still logs once at build.

## Review Focus

- Confirm clamping both margins and padding (vs margins-only) is acceptable.
- Confirm only StyleFactory (build-time) opts into `warnIfNegative`, not `derivedStateOf`.
- Review the Emerge snapshot of the new preview to confirm clamped rendering looks correct, then approve it.

## Risks / Reviewer Notes

- Risk: clamping padding (not just margin) could change rendering if a paywall intentionally used negative padding.
  - Evidence: grep of source and JSON fixtures found only `0`/positive values; negative padding is unsupported on Android.
  - Mitigation: CI Emerge/Paparazzi snapshots surface any visual change (0 modified observed).

## Non-goals / Out of Scope

- No change to the serialized `Padding` model or backend contract.
- No iOS changes.

## Omitted Context

- Raw transcript, unrelated exploration, sensitive details, repetitive attempts, and chain-of-thought-style content were omitted.

</details>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Padding.kt
```diff
@@ -8,12 +8,16 @@ import com.revenuecat.purchases.paywalls.components.properties.Padding
...
 internal fun Padding.toPaddingValues(): PaddingValues =
     PaddingValues(
-        start = leading.dp,
-        top = top.dp,
-        end = trailing.dp,
-        bottom = bottom.dp,
+        start = leading.coerceAtLeast(0.0).dp,
+        top = top.coerceAtLeast(0.0).dp,
+        end = trailing.coerceAtLeast(0.0).dp,
+        bottom = bottom.coerceAtLeast(0.0).dp,
```
tonidero, line 35: I wonder if we should log something when receiving negative values...

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/PaddingTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.ui.unit.dp
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.paywalls.components.properties.Padding
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+@OptIn(InternalRevenueCatAPI::class)
+internal class PaddingTest {
```
tonidero, line 14: It could be interesting to add a preview (for example to StackComponent) with a negative paddings. That way we validate how it looks as well.

tonidero (approved):
A couple of suggestions, but I think this is ok!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Padding.kt
facumenzella: Good call. Added a warning log in `toPaddingValues` when any value is negative, so the dashboard misconfiguration is surfaced (`7b79037`). It logs the offending values and notes they're clamped to 0.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/PaddingTest.kt
facumenzella: Done. Added `StackComponentView_Preview_NegativePaddingAndMarginClamped` (negative padding + margin via `Padding(...).toPaddingValues()`), so Emerge snapshots it and we can see the clamped result renders like zero padding/margin (`7b79037`).

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Padding.kt
```diff
@@ -5,15 +5,29 @@ package com.revenuecat.purchases.ui.revenuecatui.components.ktx
...
-        end = trailing.dp,
-        bottom = bottom.dp,
+internal fun Padding.toPaddingValues(): PaddingValues {
+    if (minOf(top, bottom, leading, trailing) < 0.0) {
+        Logger.w(
+            "Received negative padding/margin value(s) " +
+                "(top=$top, bottom=$bottom, leading=$leading, trailing=$trailing). " +
+                "Negative padding/margin is not supported on Android; clamping to 0.",
+        )
+    }
```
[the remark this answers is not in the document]
facumenzella: Good catch. Fixed in `0279786`: `toPaddingValues` is now pure (clamps unconditionally, no logging), and the warning is gated behind an opt-in `warnIfNegative` flag that only StyleFactory passes (once per paywall build). The `derivedStateOf` paths call it without the flag, so partial-override recomputes (tab/package/size/variable changes) no longer re-emit the warning. Clamping still happens on every call regardless.

PR #3560, opened 2026-06-08
Add `contains` operator (case-insensitive substring)

ajpallares:
### Checklist

### Motivation
Adds the khepri-specific `contains` operator for case-insensitive substring checks in workflow predicates ([SDK-4348](https://linear.app/revenuecat/issue/SDK-4348)). Companion iOS PR: https://github.com/RevenueCat/purchases-ios/pull/6943

### Description
Implemented as `function(haystack, needle)` with locale-independent lowercasing; non-string haystacks return `false`, with no json-logic falsy guard on the haystack. Twenty shared predicate fixtures added (fixture count 347 → 367).


ajpallares:
Closing this PR. We've decided to only implement operators that are part of the JSON Logic specification in the SDKs. `contains` is a khepri-specific operator and is not part of the JSON Logic spec, so we will not be adding it to the SDK.

PR #3562, opened 2026-06-08
Remove workflows fallback on 4xx errors

vegaro:
## Summary

If the workflows list fetch fails, we restore the cache from disk on **any** error, including 4xx. So a 404 (workflow removed) or 403 (misconfigured key) made the SDK resurrect the previous list from disk and keep serving it.

This PR mirrors what we do for offerings policy. It does not require any backend change.

Fallback behavior (identical to offerings):

| Outcome | Behavior |
|---|---|
| Transport failure (no HTTP response) | fall back to cached workflows |
| HTTP 5xx | fall back to cached workflows |
| HTTP 4xx | do not fall back |
| Malformed 2xx body | fall back to cached workflows |

The disk copy is left untouched on 4xx (no purge).

Doing the same for the workflow detail endpoint will be done in a future PR.

---

> [!NOTE]
> **Medium Risk**
> Changes offerings-gated workflows list caching on network errors; behavior is aligned with offerings and heavily tested but affects paywall/offering mapping when the API returns 4xx.
> 
> **Overview**
> Workflows list fetch failures now follow the same fallback rules as offerings, instead of always restoring from disk on any error.
> 
> `Backend.getWorkflows` reports a `GetWorkflowsErrorHandlingBehavior` with each error: **transport failures, 5xx, and malformed success bodies** still signal fallback to cached workflows; **HTTP 4xx** signals `SHOULD_NOT_FALLBACK`. `WorkflowManager` uses that signal on list errors—on 4xx it **does not** read the workflows list or detail envelopes from disk, **invalidates** the in-memory list freshness so the next call retries, and still **completes** pending `onComplete` callbacks so offerings delivery is not blocked. Fallback-eligible errors keep the existing disk restore path, now in `restoreWorkflowsListFromDisk`. Tests cover the new behavior mapping and the 4xx manager path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99321ba02ac533f9c4a2bb8fe9f538a00d3ec1c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -106,7 +106,10 @@ internal typealias RemoteConfigCallback = Pair<(RemoteConfigResponse) -> Unit, (
 @OptIn(InternalRevenueCatAPI::class)
 internal typealias WorkflowDetailCallback = Pair<(WorkflowDetailResponse) -> Unit, (PurchasesError) -> Unit>
 
-internal typealias WorkflowsListCallback = Pair<(WorkflowsListResponse) -> Unit, (PurchasesError) -> Unit>
+internal typealias WorkflowsListCallback = Pair<
+    (WorkflowsListResponse) -> Unit,
+    (PurchasesError, errorHandlingBehavior: GetWorkflowsErrorHandlingBehavior) -> Unit,
+    >
 
 internal enum class PostReceiptErrorHandlingBehavior {
     SHOULD_BE_MARKED_SYNCED,
@@ -116,6 +119,11 @@ internal enum class PostReceiptErrorHandlingBehavior {
 
 internal enum class GetOfferingsErrorHandlingBehavior {
     SHOULD_FALLBACK_TO_CACHED_OFFERINGS,
+    SHOULD_NOT_FALLBACK,
+}
+
+internal enum class GetWorkflowsErrorHandlingBehavior {
+    SHOULD_FALLBACK_TO_CACHED_WORKFLOWS,
     SHOULD_NOT_FALLBACK,
 }
 
@@ -1068,7 +1076,7 @@ internal class Backend(
         appInBackground: Boolean,
         type: String? = null,
         onSuccess: (WorkflowsListResponse) -> Unit,
-        onError: (PurchasesError) -> Unit,
+        onError: (PurchasesError, GetWorkflowsErrorHandlingBehavior) -> Unit,
     ) {
         val endpoint = Endpoint.GetWorkflows(appUserID, type)
         val path = endpoint.getPath()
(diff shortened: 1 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1103,12 +1111,23 @@ internal class Backend(
...
-                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
+                            onErrorHandler(
+                                e.toPurchasesError().also { errorLog(it) },
+                                GetWorkflowsErrorHandlingBehavior.SHOULD_FALLBACK_TO_CACHED_WORKFLOWS,
+                            )
                         } catch (e: IllegalArgumentException) {
-                            onErrorHandler(e.toPurchasesError().also { errorLog(it) })
+                            onErrorHandler(
+                                e.toPurchasesError().also { errorLog(it) },
+                                GetWorkflowsErrorHandlingBehavior.SHOULD_FALLBACK_TO_CACHED_WORKFLOWS,
```
facumenzella, line 1121: On iOS I've generalized instead of duplicating the whole logic. But it's fine. We can always follow-up after releasing.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -246,43 +247,57 @@ internal class WorkflowManager(
...
-                    if (!restoredFromDisk) {
-                        workflowsCache.invalidateWorkflowsListTimestamp()
-                    }
-                    val envelopes = workflowsCache.cachedWorkflowDetailEnvelopesFromDisk().orEmpty()
-                    if (envelopes.isEmpty()) {
+                    if (behavior == GetWorkflowsErrorHandlingBehavior.SHOULD_NOT_FALLBACK) {
+                        // A 4xx means the server intentionally changed/removed these workflows. Don't
+                        // resurrect a stale list from disk; just settle the callbacks so offerings
+                        // delivery isn't stranded. Caches are left as they are (no restore, no purge).
                         completePendingCallbacks(appUserID)
```
facumenzella, line 259: nit: 4xx leaves the timestamp stale, so we re-hit the backend every call while still serving the stale in-memory map. Before this PR the restore path re-stamped it, so we used to back off. 

Want to keep retrying every call, or stamp it so a sticky 4xx backs off? Offerings does the same thing so either way is fine, just flagging.

facumenzella (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: let's mimic what offerings does

vegaro: there was actually a divergence with offerings here. In offerings we call forceCacheStale on 4xx errors

PR #3567, opened 2026-06-09
Remove workflow details fallback on 4xx errors

vegaro:
The workflow detail fetch had no fallback on failure, unlike offerings and the workflow list fetch. On transport errors, 5xx responses, or malformed bodies, the error was surfaced immediately even when a persisted envelope was available on disk.

Mirrors #3562 but in the workflow detail path. The fallback applies to every caller (foreground miss, prefetch, and the SWR background refresh), matching `OfferingsManager`:

```mermaid
flowchart TD
    A([Backend detail fetch fails]) --> B{Error-handling<br/>behavior}

    B -->|"4xx<br/>(SHOULD_NOT_FALLBACK)"| INV
    B -->|"transport · 5xx · malformed<br/>(SHOULD_FALLBACK)"| C{Disk envelope<br/>for workflowId?}

    C -->|absent| INV
    C -->|present| D[Re-resolve envelope]

    D -->|resolve fails| INV
    D -->|resolve ok| E[cacheWorkflow<br/>re-stamps fresh]
    E --> OK

    INV[Invalidate in-memory<br/>timestamp] --> ERR

    OK(["onSuccess<br/>resolved result"]):::ok
    ERR(["onError<br/>original backend error"]):::err

    classDef ok fill:#d4edda,stroke:#28a745,color:#155724
    classDef err fill:#f8d7da,stroke:#dc3545,color:#721c24
```

---

> [!NOTE]
> **Medium Risk**
> Changes paywall/workflow loading when the backend is down or returns errors; behavior is aligned with existing offerings fallbacks but affects what users see offline and after 4xx responses.
> 
> **Overview**
> Workflow detail fetches now carry **`GetWorkflowsErrorHandlingBehavior`** from `Backend.getWorkflow`, matching offerings and the workflows list: **4xx** → `SHOULD_NOT_FALLBACK`, **transport / 5xx / malformed body** → `SHOULD_FALLBACK_TO_CACHED_WORKFLOWS`.
> 
> **`WorkflowManager`** uses that signal on foreground misses and prefetch (not on stale-while-revalidate background refreshes, via `alreadyServedStaleValue`). On fallback-eligible failures it **re-resolves a persisted detail envelope from disk** and succeeds when possible; on **4xx** or when disk recovery fails it **invalidates the in-memory workflow timestamp** so the next call retries instead of serving a stale cached value. SWR background refresh failures stay logged-only and do not touch disk or re-pin from envelope.
> 
> **`WorkflowsCache`** adds `invalidateWorkflowTimestamp` and `cachedWorkflowDetailEnvelopeFromDisk` to support that policy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8d33a376ffcac4a2f0d01d3889389793ef0c19b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -104,7 +104,10 @@ internal typealias WebBillingProductsCallback = Pair<(WebBillingProductsResponse
 internal typealias RemoteConfigCallback = Pair<(RemoteConfigResponse) -> Unit, (PurchasesError) -> Unit>
 
 @OptIn(InternalRevenueCatAPI::class)
-internal typealias WorkflowDetailCallback = Pair<(WorkflowDetailResponse) -> Unit, (PurchasesError) -> Unit>
+internal typealias WorkflowDetailCallback = Pair<
+    (WorkflowDetailResponse) -> Unit,
+    (PurchasesError, errorHandlingBehavior: GetWorkflowsErrorHandlingBehavior) -> Unit,
+    >
 
 internal typealias WorkflowsListCallback = Pair<
     (WorkflowsListResponse) -> Unit,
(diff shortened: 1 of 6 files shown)


facumenzella (approved):
Small thing I noticed: on the 4xx path we only clear the timestamp, so the value sticks around and we keep re-serving a workflow the server already dropped. Already fixed it in #3578 stacked on top, so good to go once that lands.

PR #3568, opened 2026-06-09
Fix paywall crash when a downloaded font file fails to load

tonidero:
## Summary

We received reports of a crash in downloaded paywall fonts:

```
java.lang.IllegalStateException: Unable to load font
Font(file=/data/user/0/<pkg>/cache/rc_paywall_fonts/<hash>.otf, weight=FontWeight(weight=600), style=Normal)
```

### Root cause

When compose is resolving the fonts, this happened outside our currently controlled stack trace. This means that any issue loading the typeface (cache file evicted by the OS after the earlier `exists()` check, corrupt/truncated, or undecodable by the platform) could cause a crash. We should never crash due to fonts, so this changes how we load fonts to avoid any possible crashes by adding a fallback.

### Fix

Route downloaded fonts through a new `LocalFileFont` (custom `AndroidFont` + `TypefaceLoader`) whose `loadBlocking` catches load failures and returns a safe fallback `Typeface` (weight/style aware on API 28+, `Typeface.DEFAULT` below) instead of throwing. Empty families short-circuit to `FontFamily.Default`. Per-weight/style matching is preserved because each `DownloadedFont` remains a distinct `Font`.

---

> [!NOTE]
> **Low Risk**
> Scoped to paywall font resolution and rendering; behavior change is graceful fallback instead of crash, with tests covering failure paths.
> 
> **Overview**
> Fixes paywall crashes when cached downloaded font files fail during Compose text layout by **not** using Compose’s blocking `Font(file = …)` loader, which throws `IllegalStateException("Unable to load font …")` on missing, evicted, or corrupt cache files.
> 
> **Downloaded fonts** in `FontSpec.resolve` now build families via **`localFileFont`**, a custom `AndroidFont` whose `loadBlocking` catches load failures and returns a fallback `Typeface` (weight/style-aware on API 28+, `Typeface.DEFAULT` below). **`FileTypefaceLoader`** abstracts disk loading (`Typeface.Builder` on API 26+). Empty downloaded families resolve to **`FontFamily.Default`**.
> 
> Unit tests cover empty vs multi-face `FontSpec.Downloaded` resolution and fallback behavior on loader errors/null across API levels.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8aa34209f3ce9254455dbe76e6a5883e139a390b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
@@ -6,8 +6,7 @@ import android.graphics.Typeface
 import java.io.File
 
 /**
- * Loads an [android.graphics.Typeface] from a local font [file]. Exposed as a seam so the platform's
- * file-based loaders can be substituted in tests.
+ * Loads an [android.graphics.Typeface] from a local font [file].
  */
 internal fun interface FileTypefaceLoader {
     fun load(file: File): Typeface?
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
@@ -10,5 +10,5 @@ import java.io.File
  * file-based loaders can be substituted in tests.
  */
 internal fun interface FileTypefaceLoader {
-    fun load(file: File): Typeface
+    fun load(file: File): Typeface?
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
@@ -1,0 +1,14 @@
+@file:JvmSynthetic
+
+package com.revenuecat.purchases.ui.revenuecatui.components.properties
+
+import android.graphics.Typeface
+import java.io.File
+
+/**
+ * Loads an [android.graphics.Typeface] from a local font [file]. Exposed as a seam so the platform's
+ * file-based loaders can be substituted in tests.
+ */
+internal fun interface FileTypefaceLoader {
+    fun load(file: File): Typeface
+}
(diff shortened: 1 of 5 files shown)


ajpallares (approved):
I think this makes sense. Thanks for the improvement!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
```diff
@@ -0,0 +1,14 @@
+@file:JvmSynthetic
+
+package com.revenuecat.purchases.ui.revenuecatui.components.properties
+
+import android.graphics.Typeface
+import java.io.File
+
+/**
+ * Loads an [android.graphics.Typeface] from a local font [file]. Exposed as a seam so the platform's
+ * file-based loaders can be substituted in tests.
```
JayShortway: I'm not sure "seam" explains much here.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/LocalFileFont.kt
```diff
@@ -0,0 +1,93 @@
...
+    override suspend fun awaitLoad(context: Context, font: AndroidFont): Typeface =
+        loadBlocking(context, font)
+
+    @Suppress("TooGenericExceptionCaught")
+    override fun loadBlocking(context: Context, font: AndroidFont): Typeface {
+        val localFont = (font as? LocalFileFont) ?: run {
+            Logger.w(
+                "Expected font to be LocalFileFont, but was ${font::class.java.simpleName}. " +
+                    "Falling back to default typeface.",
+            )
```
JayShortway: Maybe this (and other logs in this file) are errors? Not sure actually.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/LocalFileFontTests.kt
JayShortway, line 1: ❤️

JayShortway (approved):
Nice, this looks like the proper solution!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/properties/FileTypefaceLoader.kt
ajpallares: Is this today's "vacuous"? I just saw this in another PR and wondered the same thing 😆

PR #3571, opened 2026-06-09
Look for embedded paywall fonts in public/assets

tonidero:
### Why

Capacitor apps that use our paywalls SDK under the hood bundle their fonts into `src/main/assets/public/assets/` (which maps to the AssetManager path `public/assets/` at runtime). The SDK only ever looked in `assets/fonts/`, so these locally bundled fonts were never found and the SDK fell through to downloading them over the network instead of using the local copy.

### What

Search `public/assets/` in addition to `fonts/` when resolving embedded fonts, in both places that look up bundled fonts:
- `OfferingFontPreDownloader` — the pre-download "is this font bundled?" check that decides whether to skip the download.
- `PaywallResourceProvider.getAssetFontPaths` — the render-time asset font resolution.

---

> [!NOTE]
> **Low Risk**
> Narrow change to asset font path resolution with ordered fallback and tests; no auth, billing, or network contract changes beyond skipping unnecessary font downloads when assets are found.
> 
> **Overview**
> Embedded paywall fonts are now resolved from **`public/assets/`** in addition to **`fonts/`**, so Capacitor apps that bundle fonts under `public/assets` are treated as local copies instead of triggering network downloads.
> 
> **`OfferingFontPreDownloader`** walks an ordered list of asset directories when deciding if a font is bundled (skipping pre-download). **`PaywallResourceProvider.getAssetFontPaths`** uses the same ordered search at render time, with **`fonts/`** winning when the same `.ttf` exists in both places.
> 
> Tests cover Capacitor-style fonts in `public/assets`, precedence, and missing fonts; **`MockResourceProvider`** test data was aligned with the `fonts/` path shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 36173ed766e556149a47fe6511fe17d70ef6ca82. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
@@ -19,7 +19,7 @@ internal class OfferingFontPreDownloader(
 ) {
 
     // Directories within the app's assets where embedded fonts may be bundled, searched in order.
-    // "public/assets" is where Capacitor apps that use our paywalls SDK place their fonts.
+    // "public/assets" is a likely location for Capacitor apps that use our paywalls SDK to place their fonts.
     private val assetsFontsDirs = listOf("fonts", "public/assets")
 
     // GenericFontFamily names as defined by Compose. Restated here because we don't include any Compose dependencies.
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
@@ -18,7 +18,9 @@ internal class OfferingFontPreDownloader(
     private val fontLoader: FontLoader,
 ) {
 
-    private val assetsFontsDir = "fonts"
+    // Directories within the app's assets where embedded fonts may be bundled, searched in order.
+    // "public/assets" is where Capacitor apps that use our paywalls SDK place their fonts.
+    private val assetsFontsDirs = listOf("fonts", "public/assets")
 
     // GenericFontFamily names as defined by Compose. Restated here because we don't include any Compose dependencies.
     private val genericFonts = setOf(
@@ -79,8 +81,10 @@ internal class OfferingFontPreDownloader(
     private fun Context.getAssetFontPath(name: String): String? {
         val nameWithExtension = if (name.endsWith(".ttf")) name else "$name.ttf"
 
-        return resources.assets.list(assetsFontsDir)
-            ?.find { it == nameWithExtension }
-            ?.let { "$assetsFontsDir/$it" }
+        return assetsFontsDirs.firstNotNullOfOrNull { dir ->
+            resources.assets.list(dir)
+                ?.find { it == nameWithExtension }
+                ?.let { "$dir/$it" }
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/testdata/TestData.kt
```diff
@@ -497,7 +497,7 @@ internal class MockResourceProvider(
     override fun getAssetFontPaths(names: List<String>): Map<String, String>? {
         val foundPaths = names.associateWith { name ->
             val nameWithExtension = if (name.endsWith(".ttf")) name else "$name.ttf"
-            "${ResourceProvider.ASSETS_FONTS_DIR}/$nameWithExtension"
+            "${ResourceProvider.ASSETS_FONTS_DIRS.first()}/$nameWithExtension"
```
rickvdl: Nit: This keeps the existing behavior as expected, but I'm wondering if we should just hardcode `fonts` here? This now assumes `fonts` is always first in the list. Alternatively we could also document this somewhere (the `ASSETS_FONTS_DIRS` definition?

rickvdl (approved):
Good find!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/testdata/TestData.kt
tonidero: That's a valid point... I will change it 👍

tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/3571/commits/693bc542c65f9ad47f9f4d445754c370aaf56a46

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
```diff
@@ -18,7 +18,9 @@ internal class OfferingFontPreDownloader(
     private val fontLoader: FontLoader,
 ) {
 
-    private val assetsFontsDir = "fonts"
+    // Directories within the app's assets where embedded fonts may be bundled, searched in order.
+    // "public/assets" is where Capacitor apps that use our paywalls SDK place their fonts.
+    private val assetsFontsDirs = listOf("fonts", "public/assets")
```
JayShortway, line 23: So it's kinda `assets/public/assets`? 😅 Did you happen to check the fonts are really at this path?

JayShortway (approved):
Nice to get this in! 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/OfferingFontPreDownloader.kt
tonidero: Great question! So... basically it depends on where in the original capacitor location the fonts are 😅 . If in the capacitor app, the fonts are inside `public/assets`, then, they are copied to `assets/public/assets`. If they have it in `/public/fonts` then they would be in `assets/public/fonts`. So this depends a bit on the implementation in Capacitor TBH... These comments certainly make it seem like this is the standard in Capacitor, even though there isn't any, so I will update those... And well, for other places, I think the common is to have either an assets or fonts folder, so maybe we could add that `public/fonts` folder too... but I'm afraid of doing too many disk checks... Wdyt?

JayShortway: `public/assets` is the default or most used right? (I seem to remember that.) If so, maybe we can check just that folder, and add more if customers ask about it?

tonidero: Agreed, that seems like a sensible approach.

PR #3573, opened 2026-06-09
Change multipage paywalls compiler flag to `@InternalRevenueCatAPI` `DangerousSetting`

vegaro:
Replaces the compile-time `BuildConfig.USE_WORKFLOWS_ENDPOINT` with a **runtime** opt-in on `DangerousSettings`. So we can release to maven the feature.

---

> [!NOTE]
> **Medium Risk**
> Changes how multipage paywall / workflow fetching is enabled and affects paywall resolution and WorkflowsCache lifecycle, though default behavior stays off and the API is internal opt-in only.
> 
> **Overview**
> **Multipage paywalls (Workflows)** are no longer gated by a compile-time `BuildConfig.USE_WORKFLOWS_ENDPOINT` flag. They are controlled at **configure time** via `@InternalRevenueCatAPI` **`DangerousSettings.useWorkflows`** and **`DangerousSettings.forWorkflows()`**, so the same Maven artifacts can ship with workflows off by default and enabled only when RevenueCat-internal callers opt in.
> 
> That setting flows through **`AppConfig.useWorkflows`** (e.g. **`WorkflowsCache`** creation in **`PurchasesFactory`**) and **`PurchasesType.useWorkflows`** in RevenueCat UI, so **`PaywallViewModel`** resolves ID-based paywalls via the workflows endpoint when configured instead of reading UI module **`BuildConfig`**. The **`USE_WORKFLOWS_ENDPOINT`** **`buildConfigField`** entries are removed from **`purchases`** and **`ui/revenuecatui`** Gradle configs; public API dumps add **`getUseWorkflows()`**, with api-tester and unit tests covering the new surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a595689b31f70d7b55473a449f574d77ededf5ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -163,7 +163,12 @@ internal class PaywallViewModelImpl(
     private val shouldDisplayBlock: ((CustomerInfo) -> Boolean)?,
     preview: Boolean = false,
     private val productChangeCalculator: ProductChangeCalculator = ProductChangeCalculator(purchases),
-    private val useWorkflowsEndpoint: Boolean = BuildConfig.USE_WORKFLOWS_ENDPOINT,
+    private val useWorkflowsEndpoint: Boolean =
+        if (Purchases.isConfigured) {
+            Purchases.sharedInstance.currentConfiguration.dangerousSettings.useWorkflows
```
tonidero: Hmm we should probably obtain this through the `purchases` instance above from the `PurchasesType`.

purchases/src/main/kotlin/com/revenuecat/purchases/DangerousSettings.kt
```diff
@@ -22,8 +22,21 @@ public class DangerousSettings internal constructor(
     internal val uiPreviewMode: Boolean = false,
 
     internal val applyObfuscatedAccountIdToSubscriptionChanges: Boolean = false,
+
+    /**
+     * Enables RevenueCat Workflows (multipage paywalls). Internal RevenueCat use only.
+     */
+    @property:InternalRevenueCatAPI
```
tonidero: Hmm with only applying to the property, I guess this means it's not applied to the constructor? Which is internal already so should be fine I think... Just wondering if we can just use `@InternalRevenueCatAPI` directly... but I think I'm likely missing something 😅

purchases/src/main/kotlin/com/revenuecat/purchases/DangerousSettings.kt
```diff
@@ -39,5 +52,19 @@ public class DangerousSettings internal constructor(
             uiPreviewMode = true,
             applyObfuscatedAccountIdToSubscriptionChanges = false,
         )
+
+        /**
+         * Creates a [DangerousSettings] with RevenueCat Workflows (multipage paywalls) enabled.
+         * Internal RevenueCat use only; behavior may change without warning.
+         */
+        @InternalRevenueCatAPI
+        @JvmStatic
```
tonidero, line 61: I would say we probably don't need `@JvmStatic` if it's `@InternalRevenueCatAPI`... but I guess it doesn't hurt either.

purchases/src/test/java/com/revenuecat/purchases/common/AppConfigTest.kt
```diff
@@ -506,7 +507,8 @@ class AppConfigTest {
                 "autoSyncPurchases=true, " +
                 "customEntitlementComputation=false, " +
                 "uiPreviewMode=false, " +
-                "applyObfuscatedAccountIdToSubscriptionChanges=false), " +
+                "applyObfuscatedAccountIdToSubscriptionChanges=false, " +
+                "useWorkflows=false), " +
```
tonidero, line 511: Had the same thing in my proguard PR where this has a close parentheses and continues 😅 Weird.

tonidero (commented):
Only thing I think is to try to use the `PurchasesType` interface instead of using `Purchases.shared` from the view model. Other than that, looks great!

purchases/src/main/kotlin/com/revenuecat/purchases/DangerousSettings.kt
vegaro: nah I think you're correct

purchases/src/main/kotlin/com/revenuecat/purchases/DangerousSettings.kt
vegaro: Ah right, I will leave it in case I want to try this from the hybrids?

tonidero (approved):
I think this makes sense! 🚢

PR #3575, opened 2026-06-09
Add internal RulesEngine evaluate API surface

ajpallares:
### Checklist
- [x] Unit tests

### Motivation
Expose the minimal entry point for the JSON Logic rules engine so SDK call sites can evaluate predicates. Resolves [SDK-4315](https://linear.app/revenuecat/issue/SDK-4315).

### Description
* Makes the existing `Value` public
* Adds a new `RulesEngine.evaluate(predicate, variables)` public method to the `:rules-engine-internal` module, with failures surfaced as `RulesEngine.EvaluationError`. `evaluate` takes the predicate as a JSON string plus a native `Map<String, Value>` scope and returns `Result<Boolean>`.
* No SDK wiring yet.

## Note on API tests
The public APIs of `:rules-engine-internal` are not pinned by API tests because they are not meant for public usage.

---

> [!NOTE]
> **Medium Risk**
> New public API surface and shared JSON parsing path affect how predicates are parsed and how errors propagate; behavior is covered by unit tests but not yet wired into the SDK.
> 
> **Overview**
> Exposes the rules engine entry point for SDK use: **`Value`** is now public, and **`RulesEngine.evaluate(predicate, variables)`** accepts a JSON predicate string plus a `Map<String, Value>` and returns **`Result<Boolean>`**, with failures as nested **`RulesEngine.EvaluationException`** (`Parse`, `TypeMismatch`, `UnsupportedOperator`, `Unknown`). The old internal **`RuleError`** type is removed in favor of that public hierarchy.
> 
> Predicate parsing moves into production **`ValueJson`** (`org.json` / `JSONTokener` → `Value`); tests route through the same parser via **`ValueJsonHelper`**. Operator and conformance tests are updated to the new exception types, and **`RulesEngineEvaluateTest`** covers success, variables, malformed JSON, and unsupported operators.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8dcff6c4deb829fcb5c1a87ce735b8dedd671aab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
```diff
@@ -12,4 +12,52 @@ public object RulesEngine {
...
     }
+
+    /**
+     * Errors surfaced by the rules engine.
+     *
+     * Note on missing variables: the evaluator does **not** raise an error
+     * for them — per the JSON Logic spec, they resolve to `null` and a warning
+     * is logged instead.
+     */
+    public sealed class EvaluationError(message: String) : RuntimeException(message) {
```
tonidero: To make it clearer this is an exception, should we call this `EvaluationException`?

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
```diff
@@ -12,4 +12,52 @@ public object RulesEngine {
...
+    /**
+     * Evaluates a JSON Logic predicate against a native variable scope.
+     *
+     * @param predicate The rule predicate as a JSON string.
+     * @param variables The resolved variable scope.
+     * @return [Result.success] with `true` when the predicate is truthy,
+     *  `false` otherwise, or [Result.failure] carrying an [EvaluationError]
+     *  when parsing or evaluation fails.
+     */
+    public fun evaluate(
```
tonidero, line 53: I guess we're still not using it so we're safe but should we start publishing this module soon?

tonidero (approved):
Just a rename suggestion but looks good otherwise!

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
ajpallares: That's right. The work for publishing it is basically ready in https://github.com/RevenueCat/purchases-android/pull/3488, but I'm holding off until we really do need it. Indeed, it'll be soon

rules-engine-internal/src/main/kotlin/com/revenuecat/purchases/rules/RulesEngine.kt
ajpallares: Good point. I had it like this to make it equal to iOS, but it's true that `EvaluationException` makes more sense since it extends `EvaluationException`.

I wonder if an `RuntimeException` makes sense for this use case, tbh. What's the Android/kotlin convention for these kind of errors @tonidero?

tonidero: `RuntimeException` has traditionally been used for exceptions that you want to keep "unhandled" so they would crash the app. Whereas `Exception` would usually be used for exceptions you want to handle. So great catch! I would probably use a simple `Exception` for this.

ajpallares: Thank you for the clarification Toni! Renamed + inherit from `Exception` done in https://github.com/RevenueCat/purchases-android/pull/3575/commits/dd25ba0b695990beb11d2db35a7483650d79923e

PR #3579, opened 2026-06-10
Inject workflows for multipage paywall preview in RC App

vegaro:
Adds `@InternalRevenueCatAPI` everywhere required for the RC App to render workflows, same way single-page paywalls are already previewed.

There's one difference with single-page paywalls. Those load via the offering (`offering.paywallComponents`), so they're injected there. Workflows are **not** loaded inside the offerings, they're fetched separately via `/workflows`, even if they have an offering associated right now. So I added an injected workflow on the presentation config (`PaywallOptions`), which also keeps it consistent with iOS #6958 (which has a `injectedWorkflowContext`). The offering is still used for the workflow's packages via `setOffering(...)`.

---

> [!NOTE]
> **Low Risk**
> Changes are gated behind InternalRevenueCatAPI and uiPreviewMode; normal apps still fetch workflows unless they opt into injection.
> 
> **Overview**
> Adds **`@InternalRevenueCatAPI` workflow injection** so the RevenueCat app can preview **multipage (workflow) paywalls** without calling `/workflows`, mirroring how single-page previews use injected offerings.
> 
> **`PaywallOptions`** gains an internal `injectedWorkflow` and **`Builder.injectedWorkflow(workflow, offering)`**, which stores the workflow and sets the offering for package resolution. **`PaywallViewModel`** checks the injected workflow first in `startWorkflowPresentationIfNeeded` and calls `startWorkflowPresentation` locally—**no `awaitGetWorkflow`**.
> 
> **`PurchasesOrchestrator.getWorkflow`** now returns **`ConfigurationError`** when **`uiPreviewMode`** is on, so preview builds cannot hit the network for workflows. Tests cover injected rendering and the preview-mode guard; the API tester documents the new builder API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0401635b4a6cd89b496fe48f059e1fe29ad9a3f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -577,6 +577,17 @@ internal class PurchasesOrchestrator(
         // back synchronously on the caller's thread while a miss resolves on its IO scope, and the
         // prefetch path routes detail callbacks onto a dedicated dispatcher — so normalizing here, at
         // the consumer boundary, is what gives callers (including awaitGetWorkflow) a stable thread.
+        if (appConfig.uiPreviewMode) {
```
vegaro, line 580: we do the same in offerings to prevent making any call, since we already have the data we need

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallOptionsAPI.kt
```diff
@@ -31,4 +33,10 @@ private class PaywallOptionsAPI {
         val purchaseLogic2: PaywallPurchaseLogic? = options.purchaseLogic
         val customVariables2: Map<String, CustomVariableValue> = options.customVariables
     }
+
+    @OptIn(InternalRevenueCatAPI::class)
+    private fun checkInjectedWorkflow(builder: PaywallOptions.Builder, workflow: WorkflowDataResult) {
+        val configured: PaywallOptions.Builder = builder.injectedWorkflow(workflow)
+        val cleared: PaywallOptions.Builder = builder.injectedWorkflow(null)
+    }
```
ajpallares, line 44: Are these needed in API tester? I guess it doesn't hurt

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -752,7 +752,25 @@ internal class PaywallViewModelImpl(
         updatePaywallState(currentOffering)
     }
 
-    private suspend fun startWorkflowPresentationFromEndpointIfNeeded(offeringSelection: OfferingSelection): Boolean {
+    private suspend fun startWorkflowPresentationIfNeeded(offeringSelection: OfferingSelection): Boolean {
+        // Injected workflow (e.g. dashboard preview): render locally, no /workflows fetch.
```
ajpallares: I don't understand this comment about the "dashboard". Does it refer to the mobile app preview instead?

```suggestion
        // Injected workflow (e.g. mobile app preview): render locally, no /workflows fetch.
```

facumenzella (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -752,7 +752,25 @@ internal class PaywallViewModelImpl(
...
+        val injectedWorkflow = options.injectedWorkflow
+        if (injectedWorkflow != null) {
+            val offering = offeringSelection.offering
+            if (offering == null) {
+                Logger.w(
+                    "Paywalls: injectedWorkflow set without a concrete Offering (use setOffering); " +
+                        "workflow screens may fail to resolve their packages.",
+                )
+            }
```
ajpallares, line 764: Is this possible to happen? `injectedWorkflow != null` but `offering == null`. I guess not, so probably this log and trying to show the workflow is ok

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -196,6 +205,19 @@ public class PaywallOptions internal constructor(
             this.customVariables = CustomVariableKeyValidator.validateAndFilter(variables)
         }
 
+        /**
```
ajpallares, line 208: Same comment as before about "dashboard preview". Unless I'm missing something, I don't think it's clear

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
```diff
@@ -196,6 +205,19 @@ public class PaywallOptions internal constructor(
...
+         * it from the backend. Internal RevenueCat use only (e.g. dashboard preview).
+         *
+         * The workflow is rendered against the [Offering] supplied via [setOffering]; its screens
+         * resolve packages from that single offering. Use [setOffering] (not the offering-id
+         * variants) when injecting a workflow, and prefer single-offering workflows in preview.
+         */
+        @InternalRevenueCatAPI
+        public fun injectedWorkflow(workflow: WorkflowDataResult?): Builder = apply {
+            this.injectedWorkflow = workflow
+        }
```
ajpallares: Would it be possible to have this method inject both the workflow and the offering? That way, we don't need the app to call `setOffering` and it'd be less error prone (and would simplify this method's docs)

ajpallares (approved):
I think it makes sense! I only have some suggestions that are not blocking

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
vegaro: right, I think at this point there will be a mocked offering?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallOptions.kt
vegaro: good idea!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
ajpallares: Correct. Coming from the app, right?
But still, we need to deal with the optional object, so I think this is fine

api-tester/src/defaults/kotlin/com/revenuecat/apitester/kotlin/revenuecatui/PaywallOptionsAPI.kt
vegaro: right, not needed but will leave them since we already have them

PR #3589, opened 2026-06-11
Lazy-load paywall components and optimize parsing

alexrepty:
### Checklist
- [x] If applicable, create follow-up issues for `purchases-ios` and hybrids (PWENG-109 for iOS)

### Motivation

Resolves PWENG-106.

Customer reports of large CPU spikes during initialization with deeply nested component trees pointed at some inefficiencies during initialization.

### Description

Three fixes:

- **Serializer (`PaywallComponent.kt`):** decode the already-parsed element directly via `decodeFromJsonElement(json)` instead of `decodeFromString(json.toString())`. Output-identical; removes the per-node re-stringify/re-parse.
- **Lazy paywall deserialization (`Offering.kt`, `OfferingParser.kt`, `OfferingToStateMapper.kt`):** `Offering.PaywallComponents.data` is now decoded lazily on first access. `OfferingParser` captures the raw `paywall_components` JSON plus a cheap top-level shape check instead of eagerly decoding the tree, so offerings-load no longer pays for paywalls that are never shown. `hasPaywall` stays decode-free, and `Offering.PaywallComponents` equality/hash/`toString` compare a cheap content hash of the raw JSON (not the decoded tree), so comparing offerings (e.g. cached vs network in the cache) never triggers a decode. The UI consumer (`validatePaywallComponentsDataOrNull`) guards the first `.data` access and falls back gracefully if a structurally-invalid tree fails to decode, preserving the previous "no crash, fall back" contract.
- **SharedPreferences migration (`SharedPreferencesManager.kt`):** read legacy `.all` once (was 2–3×) instead of per key, so the large blob is no longer re-inflated repeatedly during migration. Migration copies the `com.revenuecat.purchases.*` keys into the dedicated store **without modifying the legacy default store** (copy-only, to avoid altering existing behavior; host-app keys untouched).

> [!WARNING]
> **Behavior change — `hasPaywall`:** `Offering.hasPaywall` now reflects the *presence* of paywall data, not its validity. A `paywall_components` payload that passes the cheap top-level shape check but later fails to fully decode now reports `hasPaywall = true`, whereas the previous eager-parse path decoded it at load and reported `false`. Presentation still falls back to a default paywall when a tree is structurally invalid, so a `hasPaywall`-gated entry point shows a paywall rather than nothing — but host apps that gate UI on `hasPaywall` should be aware of this. Worth calling out in the release notes.

**Measured impact** (synthetic benchmark, desktop JVM; a low-end device is ~10× slower but ratios hold):

| Measurement | Before | After |
|---|---|---|
| `createOfferings` mapping @ 33 paywalls (load) | 297 ms | ~0 ms (no trees decoded at load) |
| `createOfferings` mapping @ 200 paywalls (load) | 1,771 ms | 0.1 ms |
| Single paywall decode (per-display cost) | 8.4 ms | 1.2 ms (~7×) |
| Single-tree decode @ depth 64 | 24.2 ms | 1.6 ms (~15×; µs/node now falls with depth — quadratic gone) |

**Testing:**
- Added regression tests: copy-only migration (legacy store left untouched) in `SharedPreferencesManagerTest`; lazy-decode, decode-free `hasPaywall`, and decode-free equality/hash/`toString` in `OfferingPaywallComponentsLazyTest`; decode-failure handling in the image/video pre-downloaders; and `tab_control_*` deserialization.
- Captured the before/after numbers with a local synthetic benchmark (run manually; not committed).
- Ran unit tests across all `:purchases` flavors (defaults/customEntitlementComputation × bc7/bc8) and both `:ui:revenuecatui` billing variants plus `:ui:debugview` — ~13,400 tests, 0 real failures (the one failing `PaywallComponentsTemplatePreviewRecorder` snapshot test requires the uninitialized `upstream/paywall-preview-resources` submodule and fails identically on `main`).
- Metalava API check: no signature diff (the reshaped `Offering.PaywallComponents` is `@InternalRevenueCatAPI`, hidden from the tracked surface). detekt: clean.

---

> [!NOTE]
> **Medium Risk**
> Meaningful paywall/offering behavior change (`hasPaywall` semantics) plus new lazy-decode failure paths, though presentation and pre-download paths are guarded to fall back safely.
> 
> **Overview**
> Reduces offerings-load and paywall-parse cost by **deferring V2 paywall component-tree deserialization** until something actually reads `paywallComponents.data`, instead of decoding every offering at parse/cache time.
> 
> **`Offering.PaywallComponents`** now wraps lazy decode with a cheap **`componentsHash`** (SHA-256 of raw JSON from the parser; structural hash when built from decoded data) so **equals/hashCode/toString** and **offering comparisons** never force a full tree decode. **`OfferingParser`** keeps raw `paywall_components` JSON, runs a **top-level shape check** (`hasPaywallComponentsShape`), and skips malformed payloads at parse time with a warning.
> 
> **`PaywallComponentSerializer`** decodes via **`decodeFromJsonElement`** instead of re-stringifying each node, removing roughly quadratic work on deep trees.
> 
> **Startup path:** **`SharedPreferencesManager`** reads legacy **`.all` once** per migration instead of multiple passes over a large XML blob.
> 
> **Resilience:** Image/video pre-downloaders and **`validatePaywallComponentsDataOrNull`** catch lazy decode failures so offerings/caching and paywall UI still **fall back** instead of crashing.
> 
> **Behavior note:** **`hasPaywall`** is now **presence-only** (shape-valid wrapper counts even if full decode would fail later); apps gating UI on `hasPaywall` should treat this as a release-note change—display still falls back when validation fails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 118c6e56c58da962583783b1b8a24ced163a88fd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -45,12 +45,13 @@ internal class SharedPreferencesManager(
...
         if (!alreadyHasVersion) {
-            if (legacySharedPreferences.value.all.keys.any {
-                        key ->
-                    key.startsWith(SHARED_PREFERENCES_PREFIX)
-                }
-            ) {
-                performMigration()
+            // Read the legacy preferences only once. `.all` forces a full synchronous load of the backing XML
+            // file, which is expensive when it holds a large value (e.g. the cached offerings response), so we
+            // avoid touching it more than necessary on this (typically main-thread) startup path.
```
tonidero, line 50: This would only really run once... But I guess it's ok to optimize it a bit. I'm also kinda surprised since I thought shared preferences kept references in memory after accessing things once... But I'm not sure about the `all` accessor` so makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/paywalls/components/PaywallComponent.kt
```diff
@@ -36,29 +37,31 @@ internal class PaywallComponentSerializer : KSerializer<PaywallComponent> {
         val jsonDecoder = decoder as? JsonDecoder
             ?: throw SerializationException("Can only deserialize PaywallComponent from JSON, got: ${decoder::class}")
         val json = jsonDecoder.decodeJsonElement().jsonObject
+        // Decode the already-parsed JsonElement directly instead of re-serializing it back to a String and
+        // re-parsing (`decodeFromString(json.toString())`). The latter made deserialization ~quadratic in tree
+        // depth, since every nested PaywallComponent re-stringified its whole subtree at each level.
```
tonidero: I would probably remove this comment. I agree the change is good, but the comment seems superfluous.

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
```diff
@@ -51,11 +51,40 @@ constructor(
...
+         * Constructor for callers that already hold decoded [PaywallComponentsData] (e.g. previews, tests,
+         * hybrid SDKs).
+         */
+        public constructor(uiConfig: UiConfig, data: PaywallComponentsData) : this(uiConfig, lazyOf(data))
+
+        /**
+         * Constructor used by the offerings parser. The (potentially large) component tree referenced by
+         * [dataProvider] is only deserialized the first time [data] is accessed — i.e. when the paywall is
+         * actually displayed — rather than eagerly for every cached offering at load time.
+         */
```
tonidero: Same regarding these comments. I feel they don't provide that much value

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
```diff
@@ -124,16 +124,27 @@ internal fun Offering.validatePaywallComponentsDataOrNull(
 ): RcResult<PaywallValidationResult.Components, NonEmptyList<PaywallValidationError>>? {
     val paywallComponents = paywallComponents ?: return null
 
+    // Force the (lazily-decoded) component tree up front. If decoding fails — e.g. a tree that passed the cheap
+    // shape check at parse time but is structurally invalid — treat it as "no components paywall" so the caller
+    // falls back, mirroring the previous eager-parse behavior where a decode failure yielded null paywallComponents.
+    @Suppress("TooGenericExceptionCaught")
+    val componentsData: PaywallComponentsData = try {
+        paywallComponents.data
```
tonidero, line 132: I guess that delaying the validation means that we indeed won't know if the paywall is valid until presentation, so `hasPaywall` which would return false for invalid paywalls before, won't work... I would say the difference in behavior is not a big deal and maybe ok though. cc @RevenueCat/coresdk for thoughts.

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/SizeAddMarginTests.kt
```diff
@@ -0,0 +1,114 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.components.ktx.addMargin
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Regression tests ensuring that zero margin (whether from an absent JSON key or explicit zeros)
+ * produces the same computed size as when no margin is applied.
+ */
+@RunWith(AndroidJUnit4::class)
+internal class SizeAddMarginTests {
```
tonidero: These tests seem weird in this PR? Just wondering if they should be on a separate PR.

purchases/src/test/java/com/revenuecat/purchases/paywalls/OfferingsParsingBenchmarkTest.kt
```diff
@@ -0,0 +1,219 @@
...
+ * NOTE on stage 4: Robolectric's SharedPreferences are in-memory, so this measures the *code-path* overhead
+ * (the `.all` deep copy + migration copy), NOT the real-device XML inflate that causes the on-device "98% CPU
+ * while reading SharedPreferences". Stages 1–3 are the faithful CPU proxy; stage 4 only demonstrates the
+ * extra main-thread work the migration adds on top of the framework's own XML load.
+ */
+@RunWith(AndroidJUnit4::class)
+@Config(manifest = Config.NONE)
+@Ignore("Manual benchmark for PWENG-106 — run explicitly via --tests, not in CI")
+@OptIn(InternalRevenueCatAPI::class)
+internal class OfferingsParsingBenchmarkTest {
```
tonidero: So these don't seem like they assert anything and just log things... I would say either we add some assertions (which will likely be brittle) or we should remove these from the tests. NABD though.

purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
```diff
@@ -148,20 +148,23 @@ internal abstract class OfferingParser {
...
+            // Defer the (potentially expensive) component-tree deserialization until the paywall is actually
+            // accessed/displayed. Capturing the raw JSON string here is cheap; without this we would eagerly
+            // deserialize every cached offering's component tree at load, even those that are never shown.
+            val rawPaywallComponents = paywallComponentsJson.toString()
+            Offering.PaywallComponents(uiConfig) {
+                json.decodeFromString<PaywallComponentsData>(rawPaywallComponents)
             }
-
-        val paywallComponents = if (paywallComponentsData != null && uiConfig != null) {
-            Offering.PaywallComponents(uiConfig, paywallComponentsData)
```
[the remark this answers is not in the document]
tonidero: I don't think it would crash since we do have a try-catch around this code... But I wonder if we should move that inside here to make it more explicit?

purchases/src/main/kotlin/com/revenuecat/purchases/common/SharedPreferencesManager.kt
```diff
@@ -79,23 +87,12 @@ internal class SharedPreferencesManager(
         ) { "Finished shared preferences migration from legacy to RevenueCat-specific preferences" }
     }
 
-    private fun getRevenueCatKeysToMigrate(): List<String> {
-        val legacyPrefs by legacySharedPreferences
```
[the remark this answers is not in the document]
tonidero: I guess this was the case before (we marked the migration as finished even if some failed) but now it's complaining because we're deleting the older keys... I would say we can probably leave that cleanup out to avoid modifying behavior just in case?

tonidero (commented):
Amazing improvements! Just some comments but good to get this in soon indeed!

purchases/src/test/java/com/revenuecat/purchases/common/SharedPreferencesManagerTest.kt
```diff
@@ -177,6 +177,38 @@ class SharedPreferencesManagerTest {
         verify(exactly = 2) { mockRevenueCatEditor.apply() } // One for the expected version, another for the actual migration
     }
 
+    @Test
+    fun `getSharedPreferences removes migrated RevenueCat keys from legacy preferences but leaves others`() {
```
aboedo: are these about the super old legacy preferences? like from 2019? or is this more recent? 
asking because in practice if it is about the really old ones, I'm not sure I'd even bother, very few people would have them and they'd be very small in size

tonidero: No, this is more recent from ~1y ago, where we migrated to our own shared preferences file, instead of using the default.

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
```diff
@@ -51,14 +51,41 @@ constructor(
...
+        // Used by the parser: the component tree is decoded lazily on first [data] access, not at load time.
+        internal constructor(
+            uiConfig: UiConfig,
+            dataProvider: () -> PaywallComponentsData,
+        ) : this(uiConfig, lazy(LazyThreadSafetyMode.SYNCHRONIZED, dataProvider))
+
+        public val data: PaywallComponentsData by dataProvider
+
+        // Hand-written instead of @Poko because `data` is lazily evaluated; equality/hash force the decode.
+        override fun equals(other: Any?): Boolean =
```
MonikaMateska, line 82: One thing I'm concerned about is that Offering is @Poko, its generated equals/hashCode include paywallComponents, and these hand-written ones force the lazy decode. So any time two Offerings get compared (e.g. in the cache when we check whether offerings actually changed) we end up synchronously decoding every offering's whole component tree, which is exactly the work we're trying to defer here.
Could we base equality on something cheap and stable instead of the decoded tree?
We could maybe compare the raw JSON string we already capture for the lazy decode. Hold onto it as a field and use it (plus uiConfig as we have it now) for equals/hashCode.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
MonikaMateska: This is a public API, and host apps might use `hasPaywall` to gate UI, eg/ decide whether to route to a paywall screen. With this change they'd show that entry point for a paywall that can't actually render. If we agree to go with this direction, we should probably call it out in the release notes as nothing really protect callers who already wrote code against the old meaning.

MonikaMateska (changes requested):

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
tonidero: Hmm the problem I see of keeping the original string is that it would increase memory usage for all offerings. Maybe we could keep some hash of the string... It would be some compute power though but maybe more acceptable. Good point about this possibly causing a decoding before presentation.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/OfferingToStateMapper.kt
tonidero: Well, my main reason to consider this not a huge deal is that, ideally, this shouldn't happen commonly (may happen if using a newer paywall version in some cases though and such... In most of these situations, we would default to the "default" paywall, so this change could indeed increase the number of times we display the default paywall, but again, this would be in uncommon scenarios. That's why I think the change in behavior is acceptable. But yeah, calling it out on the release is probably a good idea.

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
MonikaMateska: Yes, hash actually could be way better regarding memory, and it's still reliable.

alexrepty: @MonikaMateska @tonidero This is good call, I've updated it so that equality is now based on a cached hash.

MonikaMateska (approved):

alexrepty:
@RCGitBot please test

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
```diff
@@ -51,14 +51,51 @@ constructor(
...
+    public class PaywallComponents private constructor(
         public val uiConfig: UiConfig,
-        public val data: PaywallComponentsData,
-    )
+        // Cheap, stable equality key that never touches the lazily-decoded [data]. For parser-built instances
+        // this is a hash of the raw component-tree JSON; for instances built from already-decoded data it is
+        // the data itself (where comparing it is free, since there is nothing to decode). This keeps
+        // Offering.equals/hashCode/toString — which are @Poko-generated over `paywallComponents` — from forcing
+        // a decode of every offering's component tree (e.g. when the cache compares cached vs network offerings).
+        private val identity: Any,
```
tonidero: Since this only refers to the `PaywallComponentsData` but not to the `UiConfig`, maybe we could rename to `componentsDataIdentity`?

tonidero: Also, should this just be renamed to `componetsHash` of type string directly? That way we avoid the `Any` and make it clearer what it is IMO...

tonidero (commented):
Just a small suggestion to simplify the identity property... Other than that, I think this is ready!

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
alexrepty: > Also, should this just be renamed to `componetsHash` of type string directly? That way we avoid the `Any` and make it clearer what it is IMO...

Done! Good call.

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
```diff
@@ -51,14 +51,50 @@ constructor(
...
-    public class PaywallComponents(
+    public class PaywallComponents private constructor(
         public val uiConfig: UiConfig,
-        public val data: PaywallComponentsData,
-    )
+        // Cheap, stable equality key that never touches the lazily-decoded [data]. For parser-built instances
+        // this is a SHA-256 of the raw component-tree JSON; for instances built from already-decoded data it is
+        // that data's structural hash (cheap — there is nothing to decode). This keeps
+        // Offering.equals/hashCode/toString — which are @Poko-generated over `paywallComponents` — from forcing
+        // a decode of every offering's component tree (e.g. when the cache compares cached vs network offerings).
```
tonidero, line 60: IMO, this whole comment is unnecessary and verbose, so I would probably remove it.

tonidero (approved):
I think this looks great! Amazing improvement!!

PR #3596, opened 2026-06-12
Add `paywall_id` and `workflow_id` to paywall events and other fixes

vegaro:
## Summary

- Fix paywall impression/close events for non-paywall workflow steps. Only steps that actually render a paywall should emit paywall events
- Do not fire a close event when the user navigates back within the workflow. Back navigation is not a dismiss
- The backend requires `context` object and was failing to post without one.

---

> [!NOTE]
> **Medium Risk**
> Changes analytics payloads and workflow paywall event gating; backward-compatible for cached events without `workflowId`, but altered close/impression semantics can shift reported metrics.
> 
> **Overview**
> Extends **paywall analytics** so workflow-driven presentations can be correlated on the backend: `PaywallEvent.Data` gains optional `workflowId`, backend paywall payloads add top-level `workflow_id` and enrich `presented_offering_context` with `paywall_id` / `workflow_id` (including when placement/targeting alone would previously omit context). **Workflow backend events** now always serialize `context` with `platform`, `sdk_version`, and optional `locale` so posts match the expected schema.
> 
> **PaywallViewModel** workflow behavior changes: paywall impression/close/interaction events only fire on steps that actually render a paywall (`singleStepFallbackId`); non-paywall steps are suppressed, in-workflow navigation no longer emits `paywall_close`, and `closePaywall` clears workflow state so reused VMs do not leak `workflowId` into later impressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 425ea2611737a4d2eed2f92e03e954e9c23b77f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallComponentsConfigExtensions.kt
```diff
@@ -0,0 +1,79 @@
...
+import com.revenuecat.purchases.paywalls.components.TabControlComponent
+import com.revenuecat.purchases.paywalls.components.TabControlToggleComponent
+import com.revenuecat.purchases.paywalls.components.TabsComponent
+import com.revenuecat.purchases.paywalls.components.TextComponent
+import com.revenuecat.purchases.paywalls.components.TimelineComponent
+import com.revenuecat.purchases.paywalls.components.VideoComponent
+import com.revenuecat.purchases.paywalls.components.common.PaywallComponentsConfig
+
+@JvmSynthetic
+internal fun PaywallComponentsConfig.containsPaywallEventComponent(): Boolean =
```
facumenzella: This new recursive helper has no direct tests, just one indirect ViewModel check (packageless step). A bunch of branches go untested: button -> Sheet destination, carousel pages, tabs control, countdown stacks, header/stickyFooter. A bug in any silently breaks event emission. Since it's a pure function it's cheap to cover directly, e.g.:

```kotlin
@Test
fun `package nested in a button sheet counts as a paywall`() {
    val sheet = ButtonComponent.Destination.Sheet(
        stack = StackComponent(components = listOf(TestData.Components.monthlyPackageComponent)),
        /* ...other sheet fields... */
    )
    val button = ButtonComponent(
        stack = StackComponent(components = listOf(text())),
        action = ButtonComponent.Action.NavigateTo(destination = sheet),
    )
    assertThat(configWith(button).containsPaywallEventComponent()).isTrue()
}

@Test
fun `button to a URL does not count`() {
    val button = ButtonComponent(
        stack = StackComponent(components = listOf(text())),
        action = ButtonComponent.Action.NavigateTo(destination = ButtonComponent.Destination.Url(/* ... */)),
    )
    assertThat(configWith(button).containsPaywallEventComponent()).isFalse()
}
```

Could we add a case per component type (package, purchase button, carousel, tabs, countdown, sticky footer) plus a text-only negative?

facumenzella (approved):
LGTM. The serialization, EncodeDefault context fix, and workflow event-suppression logic all look correct. Only non-blocking note is the test coverage suggestion on `containsPaywallEventComponent` above.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/helpers/PaywallComponentsConfigExtensions.kt
vegaro: I've updated the implementation for now to use `workflow.singleStepFallbackId` instead of walking the tree. We discussed sending it from the backend, but that can come in a future PR

PR #3601, opened 2026-06-15
refactor(remote-config): Trusted-entitlements signing for the binary remote config response

tonidero:
Stacked on #3607 (which adds the binary RC Container request/parsing).

Enables trusted-entitlements signature verification for the binary `GET /v2/config` response. This hasn't been tested against the actual endpoint yet, but this endpoint is currently unused so should be safe to merge.

## What
- `GetRemoteConfig.supportsSignatureVerification = true`. The response signature is verified over the config element's 24-byte truncated SHA-256 checksum (no nonce). Two independent checks are both required, else `FAILED`:
  1. `X-Signature` (Ed25519) covers the checksum -> backend authenticity.
  2. `RCElement.isChecksumValid()` ties the checksum to the actual config bytes.
- `SigningManager`: signed body is now `ByteArray?` (binary-safe).
- `HTTPClient.verifyBinaryResponse` parses the container, checks integrity, and verifies the signature over the config checksum.
- The resulting `VerificationResult` is surfaced through the `getRemoteConfig` callback.

---

> [!NOTE]
> **Medium Risk**
> Touches signature verification and changes the getRemoteConfig callback shape, but scope is limited to the still-unused remote config endpoint with strong test coverage.
> 
> **Overview**
> Enables **trusted-entitlements signature verification** for `GET /v2/config` RC Container responses and threads the outcome to callers.
> 
> **`GetRemoteConfig`** is now in the set of endpoints that require response signing. Verification is **not** over the raw body: the client parses the container, requires **`RCElement.isChecksumValid()`** on the config element, then verifies **`X-Signature`** over that element’s **24-byte truncated SHA-256 checksum** (no nonce / post-params). Empty **204** responses are treated as **`VERIFIED`** without calling the signer.
> 
> **`SigningManager.verifyResponse`** now takes **`ByteArray?`** for the signed payload so JSON bodies and binary checksums use the same API. **`getRemoteConfig`** success callbacks are **`(RCContainer?, VerificationResult)`** instead of container-only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de07e2bc82504b40b817198c0ddfc37a599a9c17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -21,6 +21,8 @@ import com.revenuecat.purchases.common.networking.HTTPResult
 import com.revenuecat.purchases.common.networking.HTTPTimeoutManager
 import com.revenuecat.purchases.common.networking.MapConverter
 import com.revenuecat.purchases.common.networking.NullPointerReadingErrorStreamException
+import com.revenuecat.purchases.common.networking.RCContainer
+import com.revenuecat.purchases.common.networking.RCContainerFormatException
 import com.revenuecat.purchases.common.networking.RCHTTPStatusCodes
 import com.revenuecat.purchases.common.verification.SignatureVerificationException
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
@@ -394,7 +396,11 @@ internal class HTTPClient(
         val verificationResult = if (shouldSignResponse &&
             RCHTTPStatusCodes.isSuccessful(responseCode)
         ) {
-            verifyResponse(path, connection, payloadText, nonce, postFieldsToSignHeader)
+            if (endpoint.expectsRCFormatResponse) {
+                verifyRCFormatResponse(path, connection, bodyBytes)
+            } else {
+                verifyResponse(path, connection, payloadText, nonce, postFieldsToSignHeader)
+            }
         } else {
             VerificationResult.NOT_REQUESTED
         }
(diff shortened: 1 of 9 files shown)


tonidero:
* **#3601** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3601?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3607** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3607?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3534** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetRemoteConfigTest.kt
```diff
@@ -91,6 +98,7 @@ class BackendGetRemoteConfigTest {
         assertThat(parsed.config.data.let { buf -> ByteArray(buf.remaining()).also { buf.duplicate().get(it) } })
             .isEqualTo(config)
         assertThat(parsed.contentElements).hasSize(1)
+        assertThat(verification).isEqualTo(VerificationResult.VERIFIED)
```
tonidero, line 101: I will want to add an integration test for this request once the backend is working and we have tested it... for now, I haven't been able to test this yet.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -575,13 +580,47 @@ internal class HTTPClient(
...
+            errorLog { NetworkStrings.VERIFICATION_ERROR.format(urlPath) }
+            VerificationResult.FAILED
+        } else {
+            signingManager.verifyResponse(
+                urlPath = urlPath,
+                signatureString = connection.getHeaderField(HTTPResult.SIGNATURE_HEADER_NAME),
+                nonce = null,
+                bodyBytes = config.checksumBytes(),
+                requestTime = getRequestTimeHeader(connection),
+                eTag = getETagHeader(connection),
```
rickvdl, line 624: How is this handled, since we don't have etag support for these?

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -393,7 +394,11 @@ internal class HTTPClient(
         val verificationResult = if (shouldSignResponse &&
             RCHTTPStatusCodes.isSuccessful(responseCode)
```
rickvdl, line 397: How do we handle 204's here? Should we early exit and not do any verification since we don't have any content?

rickvdl (approved):
Awesome work! As far as I can see this is in line with what we discussed. Just two questions :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Right, I left this a bit as a failsafe TBH... We are indeed not storing etags for this endpoint. The idea here is that, if the backend does send an etag, even if we don't store it, as long as it uses the same signing mechanism, it should be part of the signature, so we get it from the response headers to calculate the signature. This might not happen if the backend does not provide an etag, but this would also handle it. 

So yeah, basically a guard just in case the backend does send etags.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: That's a valid question!! And yeah, right now I wasn't handling it specially or anything, but if we get a 204 (and no content), I don't think there is a need to verify anything (no chance of MiTM attack I would say, so we can just assume verified. Let me know if you think that's ok!

tonidero: I added a check for this case 🙏 https://github.com/RevenueCat/purchases-android/pull/3601/commits/de07e2bc82504b40b817198c0ddfc37a599a9c17

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
rickvdl: Makes sense, thanks for clarifying!

PR #3604, opened 2026-06-15
Normalize cache keys on `getWorkflow` to workflow id

vegaro:
### Checklist

### Motivation

The render path asks for a workflow by **offering id** (the backend lazily converts it into a workflow), while the prefetch path asks by **workflow id**. `WorkflowManager.getWorkflow`'s `workflowId` parameter was therefore overloaded — sometimes a workflow id, sometimes an offering id — and the cache was keyed inconsistently. A workflow prefetched under its workflow id missed when the render later asked by offering id, and a lazily-converted workflow cached under the offering id missed once the workflows list populated the `offeringId → workflowId` map. Either way: a redundant fetch on every render.

### Description

`getWorkflow` now **normalizes the incoming identifier to the canonical workflow id** (via the `offeringId → workflowId` map) before the cache read, so an ask by either id resolves to the same entry. The parameter is renamed to `workflowOrOfferingId` to make that contract honest. An as-yet-unconverted offering id has no mapping and passes through; its real id is learned from the response.

On a successful resolve the result is cached **once**, under the resolved workflow's own id (via a shared `cacheResolvedWorkflow` helper), and any newly discovered `offeringId → workflowId` mapping is recorded (new `WorkflowsCache.recordWorkflowIdForOfferingId`, copy-on-write under the existing lock) so later asks by offering id resolve to it. A response that resolves with an empty id can't be keyed in the cache, so it's delivered for the current render but not cached under a wrong key — the next render refetches. The disk-fallback path (`resolveDiskFallback`) caches the recovered envelope through the same helper, so it keys off the resolved id too.

No public API change. Unit tests cover: ask-by-offering-id hitting the entry prefetched under the workflow id; the lazy-conversion single-entry caching + mapping record + cache-hit on the next offering-id ask; the empty-id response being delivered but not cached; the disk-fallback re-pin keying off the resolved id; and the `WorkflowsCache` map record/replace behavior.

### Follow-ups (not in this PR)

- Mirror this gating to `purchases-ios` and hybrids.

---

> [!NOTE]
> **Medium Risk**
> Touches paywall workflow fetch/caching logic where wrong keys caused redundant network calls; behavior is well-covered by tests but affects the render hot path.
> 
> **Overview**
> Fixes workflow cache misses when **prefetch uses workflow ids** but **paywall render asks by offering id** (and when lazy conversion previously stored under the wrong key).
> 
> `WorkflowManager.getWorkflow` now takes **`workflowOrOfferingId`**, resolves it through the **`offeringId → workflowId`** map before any cache read, and centralizes writes in **`cacheResolvedWorkflow`**: one in-memory entry keyed by the **resolved workflow id**, optional **`recordWorkflowIdForOfferingId`** after lazy conversion, and **no cache** when the resolved id is empty. Disk fallback and prefetch use the same helper; prefetch turns off mapping recording.
> 
> `PurchasesOrchestrator` only forwards the new parameter name internally—**no public API change**. Tests cover offering-id hits on prefetched workflows, lazy mapping + second-render cache hit, disk fallback keying, and map merge/replace behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 549f67de8f2a1614229e665263ab907a8d33543b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -601,7 +601,7 @@ internal class PurchasesOrchestrator(
         }
```
facumenzella, line 601: Nit: this still calls it `workflowId` while forwarding to `workflowOrOfferingId`. Worth renaming for consistency, though it'd propagate to `Purchases`/`awaitGetWorkflow`, so fine to leave.

facumenzella (approved):
LGTM. Clean fix and the test coverage is thorough (verifies the actual backend-call counts, not just cache mechanics). One naming nit inline, non-blocking.

PR #3607, opened 2026-06-15
refactor(remote-config): Support RC Container Format for remote config request, remove stale implementation temporarily

tonidero:
Stacked on #3534.

## What
- Introduces `HTTPResult.Payload` (`Text` / `RCFormat`) so the HTTP layer can carry raw bytes, with a `payloadText` accessor. Pure refactor; text call sites are unchanged.
- `GetRemoteConfig` now requests the RC Container Format (`Accept: application/x-rc-format`) and parses the response into an `RCContainer`. RCFormat responses bypass the ETag cache.
- Removes unused previous remote config implementation files after modifications in the model format.

This PR doesn't change any behavior, but adds initial support for RC Container format requests.

---

> [!NOTE]
> **Medium Risk**
> Touches shared HTTP/ETag behavior and deletes the remote-config refresh/download path, so config-related behavior is incomplete until replaced; JSON endpoints are largely unchanged via `payloadText`.
> 
> **Overview**
> Refactors the HTTP layer so responses can be **text (JSON)** or **raw RC Container bytes**, via `HTTPResult.Payload` and `payloadText` for existing JSON call sites.
> 
> **`GetRemoteConfig`** now negotiates **`Accept: application/x-rc-format`**, skips ETag caching, treats **204** as “no update” (`null` `RCContainer`), and parses successful bodies with **`RCContainer.parse`**. **`HTTPClient`** reads bodies as bytes, sets RC-format headers only when `endpoint.expectsRCFormatResponse`, and bypasses the ETag manager for those endpoints.
> 
> The prior **JSON-based remote config** implementation is **removed** (`RemoteConfigManager`, `TopicFetcher`, `RemoteConfigResponse`, disk cache, and related tests) until a follow-up wires the new container format end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62f6beadcc70fe7041fec3f7f5c87e6520833694. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3601** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3601?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3607** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3607?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3607?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3534** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3534?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -105,6 +112,7 @@ internal sealed class Endpoint(
         name = "get_remote_config",
     ) {
         override fun getPath(useFallback: Boolean) = pathTemplate
+        override val expectsBinaryResponse: Boolean = true
```
tonidero: This endpoint is currently unused so not a problem to change it.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -52,66 +53,75 @@ internal class RemoteConfigManager(
...
-                val previousResponse = cache?.response
-                topicFetcher.cleanupUnreferencedTopics(referenced)
-                cache = CacheEntry(response, dateProvider.now)
-                if (previousResponse != response) {
-                    diskCache.write(response)
-                }
-            }
-        }
-        return firstError
+        // WIP: Just cache the fetched container for now. The RemoteConfigResponse-based processing
```
tonidero: All this code is not currently used, and with the changes to the binary format, it was being tricky to consolidate them, so I actually removed most of this in a [follow-up PR](https://github.com/RevenueCat/purchases-android/pull/3610) (to readd it in another follow-up PR 😓). I would focus more on the networking layer than this for this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -368,7 +389,7 @@ internal class HTTPClient(
         val verificationResult = if (shouldSignResponse &&
             RCHTTPStatusCodes.isSuccessful(responseCode)
         ) {
-            verifyResponse(path, connection, payload, nonce, postFieldsToSignHeader)
+            verifyResponse(path, connection, payloadText, nonce, postFieldsToSignHeader)
```
ajpallares, line 397: I think right now GetRemoteConfig doesn't support signature verification, but if it ever does, couldn't this (passing `null` / empty string to verify) be a problem?

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1343,12 +1344,19 @@ internal class Backend(
...
+                        val payload = result.payload
+                        if (payload !is HTTPResult.Payload.Binary) {
+                            onErrorHandler(
+                                PurchasesError(
+                                    PurchasesErrorCode.UnknownError,
+                                    "Expected a binary RC Container Format payload for remote config.",
+                                ).also { errorLog(it) },
+                            )
+                            return@forEach
+                        }
```
ajpallares: I think this is effectively unreached, since in `HTTPClient` we're only mapping to `Binary` in the success case

```
        val payload: HTTPResult.Payload = if (
            endpoint.expectsBinaryResponse && RCHTTPStatusCodes.isSuccessful(responseCode)
        ) {
            HTTPResult.Payload.Binary(payloadBytes)
        } else {
            HTTPResult.Payload.Text(String(payloadBytes, Charsets.UTF_8))
        }
```

I guess this doesn't hurt to future proof though.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -325,10 +334,21 @@ internal class HTTPClient(
...
+        // Binary endpoints expose successful responses as raw bytes; everything else (including error
+        // responses, which are still JSON) is decoded as UTF-8 text.
+        val payload: HTTPResult.Payload = if (
+            endpoint.expectsBinaryResponse && RCHTTPStatusCodes.isSuccessful(responseCode)
+        ) {
+            HTTPResult.Payload.Binary(payloadBytes)
+        } else {
+            HTTPResult.Payload.Text(String(payloadBytes, Charsets.UTF_8))
+        }
+        val payloadText = (payload as? HTTPResult.Payload.Text)?.value
```
ajpallares: Hmm this confused me because we are also defining
```kotlin
    val payloadText: String
        get() = when (val payload = payload) {
            is Payload.Text -> payload.value
            is Payload.Binary -> ""
        }
```

But then I realized `payloadText` was defined on `HTTPResult`. So, supernit: should we unify and define 
```kotlin
        val text: String
            get() = when (this) {
                is Text -> value
                is Binary -> ""
            }
```

in `Payload`? This way we can have 
```kotlin
    val payloadText: String get() = payload.text
```

and, most importantly, we remove the nullable nature of it. Having it consistent and only differentiating between empty / nonempty

FFTI

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -466,9 +500,11 @@ internal class HTTPClient(
         nonce: String?,
         shouldSignResponse: Boolean,
         postFieldsToSignHeader: String?,
+        endpoint: Endpoint,
     ): Map<String, String> {
         return mapOf(
             "Content-Type" to "application/json",
+            "Accept" to if (endpoint.expectsBinaryResponse) RC_FORMAT_CONTENT_TYPE else null,
```
ajpallares: The name could be a bit misleading (see the actual `"Content-Type"` above). Maybe...?

```suggestion
            "Accept" to if (endpoint.expectsBinaryResponse) RC_FORMAT_ACCEPT else null,
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -466,9 +500,11 @@ internal class HTTPClient(
         nonce: String?,
         shouldSignResponse: Boolean,
         postFieldsToSignHeader: String?,
+        endpoint: Endpoint,
     ): Map<String, String> {
         return mapOf(
             "Content-Type" to "application/json",
```
ajpallares, line 511: Is it okay to still send `"Content-Type" to "application/json"` for the remote config endpoint?

ajpallares (commented):
Looking good! I have some comments!
Thanks for making the effort to split these into smaller PRs. Really appreciated 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: That's actually done in [this follow-up PR](https://github.com/RevenueCat/purchases-android/pull/3601), and yeah, so it has a different method to handle this case which will use only the main config hash to simplify it

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Nice cleanup indeed!

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: That's a good question... need to check with the backend folks since I haven't been able to test this against that yet (code is unused right now though)

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Indeed... it also allows the compiler to know after this that it will be of type Binary, so I would prefer to leave it 🙏

ajpallares: Makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -8,6 +8,13 @@ internal sealed class Endpoint(
     val fallbackPath: String? = null,
 ) {
     abstract fun getPath(useFallback: Boolean = false): String
+
+    /**
+     * Whether this endpoint returns a binary (RC Container Format) response rather than JSON. When
+     * `true`, the request advertises `Accept: application/x-rc-format` and the response body is exposed
+     * as [HTTPResult.Payload.Binary] instead of being decoded as text.
+     */
+    open val expectsBinaryResponse: Boolean = false
```
rickvdl: Should we explicitly mention the RC container format here? Binary could still be broader (it isn't at the moment, but still)

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3607/commits/42d3db738cf0937a54c1b5ac1cecd827a44beb75

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Addressed in https://github.com/RevenueCat/purchases-android/pull/3607/commits/42d3db738cf0937a54c1b5ac1cecd827a44beb75

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Makes sense, just one question; would it be a lot of effort to just remove it out of this PR instead of commenting out? Would make the diff a bit smaller and since it will be re-added in (probably) a different way anyways.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Actually, I think this is right. `Content-Type` is supposed to be the format of the request, which will indeed still be plain json. `Accept` determined the accepted formats for the response, so I don't think this should be an issue.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: That's indeed a good point. Will move the removal to this PR instead 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
tonidero: I ended up renaming mentiones of "Binary" to "RCFormat". Agree it's better to be concise. Thank you!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: I've moved all removals to this PR which should simplify things. Thank you!

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -325,10 +334,21 @@ internal class HTTPClient(
...
         debugLog { NetworkStrings.API_REQUEST_COMPLETED.format(connection.requestMethod, path, responseCode) }
-        if (payload == null) {
+        if (payloadBytes == null) {
             throw IOException(NetworkStrings.HTTP_RESPONSE_PAYLOAD_NULL)
         }
 
+        // RC Format endpoints expose successful responses as raw bytes; everything else (including error
+        // responses, which are still JSON) is decoded as UTF-8 text.
+        val payload: HTTPResult.Payload = if (
+            endpoint.expectsRCFormatResponse && RCHTTPStatusCodes.isSuccessful(responseCode)
```
rickvdl, line 349: Would be good to understand if a 204 would return any body, if not probably the RCContainer decoding will fail? cc @bisho

rickvdl (approved):
Awesome work, just one remaining question 💪

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: 204 (No content) shouldn't send data, I actually handle that in a [follow-up PR](https://github.com/RevenueCat/purchases-android/pull/3612/changes#diff-8dfb4fa4310c04b8147572cf71f0d487dde8355de674796135e38ca8fa02d4efR1362)... But I will think about moving that part to this PR instead, since it makes more sense here.

rickvdl: Right! Got it. Yes if possible it would centralize things :)

tonidero: Applied in https://github.com/RevenueCat/purchases-android/pull/3607/commits/8d4c69a257433adf752a5f4651be72b0dda23645

PR #3615, opened 2026-06-16
refactor(remote-config): inline blob store + extraction

tonidero:
### Motivation
`/v1/config` responses inline content blobs (content-addressed). We need a place to persist those blobs and reuse them across syncs, so the server can stop re-sending blobs the client already holds.

### Description
Adds a content-addressed, on-disk blob store plus the logic that extracts inlined blobs from the config response, keeping only the blobs the resolved configuration still references (and verifying each blob against its content-address ref).

Also adds `clear()` to both the disk cache and the blob store so the whole cache can be wiped on identity change.

Additionally, this also fixes some older tests that were outdated and adds some integration tests for the whole process to store and retrieve inline blobs. 

---

> [!NOTE]
> **Medium Risk**
> Changes remote-config persistence ordering and on-disk cache correctness (checksum gating, pruning); mistakes could drop blobs or report wrong prefetch state, but scope is internal SDK config caching rather than auth or purchases.
> 
> **Overview**
> Adds a **content-addressed on-disk blob cache** for `/v1/config` inlined payloads and wires it into the sync path so blobs can be reused across refreshes.
> 
> **`RemoteConfigManager`** now reports only **prefetch blob refs that are actually on disk**, and on a `200` it **persists manifest/state first**; only after a successful disk write does it **extract checksum-valid inlined elements** the active config still references and **`retainOnly`** prune everything else. Failed manifest persist skips blob writes/eviction.
> 
> **`RemoteConfigDiskCache.write`** returns success/failure; **`clear()`** is added on both disk cache and blob store for wiping state on identity change (call sites may follow in a later PR).
> 
> Tests add blob-store unit coverage, manager behavior around tampering/unwanted blobs/failed persist, semi-integration tests over real disk I/O, and refresh RC container fixtures toward the **workflows**-shaped config JSON.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ada22dba81aa3dd4a63f909ec6a89db262de7d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3615** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3615?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3621](https://github.com/RevenueCat/purchases-android/pull/3621) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3621?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3672](https://github.com/RevenueCat/purchases-android/pull/3672) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3672?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3648** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -49,11 +52,22 @@ internal class RemoteConfigManager(
...
         val mergedBlobRefs = (previousBlobRefs + response.topics.toTopicBlobRefs())
             .filterKeys { it in response.activeTopics }
+
+        extractInlineBlobs(container)
+
+        // Retain only blobs the current config still references: the prefetch set plus any topic blob ref.
+        val referencedRefs = response.prefetchBlobs.toSet() + mergedBlobRefs.values.flatten()
+        blobStore.retainOnly(referencedRefs)
+
         diskCache.write(
```
rickvdl: FYI: On iOS we check if the remote config was persisted successfully before syncing the blobs to disk, that's probably a bit safer?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -49,11 +52,22 @@ internal class RemoteConfigManager(
...
         val previousBlobRefs = previous?.topicBlobRefs ?: emptyMap()
         // Changed topics overwrite their refs; topics no longer active are pruned.
         val mergedBlobRefs = (previousBlobRefs + response.topics.toTopicBlobRefs())
             .filterKeys { it in response.activeTopics }
+
+        extractInlineBlobs(container)
+
+        // Retain only blobs the current config still references: the prefetch set plus any topic blob ref.
+        val referencedRefs = response.prefetchBlobs.toSet() + mergedBlobRefs.values.flatten()
+        blobStore.retainOnly(referencedRefs)
```
rickvdl: On iOS I added some extra checks before writing the blobs to ensure they're still 'wanted' after applying the new response. Wrote down the logic in the PR description [here](https://github.com/RevenueCat/purchases-ios/pull/7073)

Basically I think the main differences is that iOS does not store all inlined blobs first, but will first check if they are wanted / needed.

Happy to discuss of course!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Bubot had a valid concern about the iOS approach as well though.. https://github.com/RevenueCat/purchases-ios/pull/7073#discussion_r3472054755

Let's discuss :)

tonidero: Both good points thank you!!

tonidero: I've been thinking about this... IMO, the most important part is the configuration itself, while the blobs should be recoverable even if the inlined values couldn't be stored for some reason... So I'm thinking, we might want to keep storing the configuration first, and maybe if there is an issue storing the inlined items in disk, it's "okish" since we can still fetch the latest version from network? Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
```diff
@@ -0,0 +1,122 @@
...
+        parent.listFiles()
+            ?.filter { it.isFile && it.name !in refs }
+            ?.forEach { deleteQuietly(it) }
+    }
+
+    /**
+     * Deletes every cached blob unconditionally. Used on identity change to keep config from bleeding across
+     * users (unlike [retainOnly], which only prunes unreferenced blobs).
+     */
+    fun clear() {
```
tonidero, line 91: @rickvdl note that I added a `clear` method here and to the disk cache. Not used right now but will be later on, to support clearing the cache on identity changes.

purchases/src/test/java/com/revenuecat/purchases/common/networking/RCContainerTestData.kt
```diff
@@ -18,54 +18,56 @@ internal object RCContainerTestData {
...
     /**
-     * A representative `/v1/config` payload, modeled on `get_remote_config_success.json`. Stored as
-     * the config element (element 0) of most fixtures so they resemble real wire data.
+     * Arbitrary placeholder workflow blobs, the kind of static content the config below references by ref.
+     * The exact shape is not asserted anywhere; they only need stable bytes so their content-address is stable.
+     * Declared before [CONFIG_JSON_TEXT] because that payload embeds their content-addresses (via [refOf]).
      */
     // language=json
-    private val CONFIG_JSON_TEXT: String = """
+    private val WORKFLOW_BLOB_TEXT: String = """
```
tonidero, line 26: Note that this used to use old format of the remote config json, so I updated the tests here to use the latest format, and make use of the commited fixtures to add some integration tests that go through the whole flow of processing, storing the blobs and retrieving them, to get more comprehensive coverage cc @rickvdl

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: I believe I addressed the behavior differences, but I mostly kept storing the config first, since I think the blobs can be stored/recovered later if needed. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
```diff
@@ -0,0 +1,122 @@
...
+            ?: emptySet()
+    }
+
+    /** Deletes every cached blob whose ref is not in [refs]. */
+    fun retainOnly(refs: Set<String>) {
+        val parent = blobsDir()
+        if (!parent.exists()) return
+        parent.listFiles()
+            ?.filter { it.isFile && it.name !in refs }
+            ?.forEach { deleteQuietly(it) }
```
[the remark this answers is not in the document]
tonidero: I think this shouldn't be a problem since they should always match. If not, it would be an issue.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -81,6 +97,25 @@ internal class RemoteConfigManager(
                 lastRefreshAt = dateProvider.now.time,
             ),
         )
+
+        if (persisted) {
+            extractInlineBlobs(container, blobRefsToKeep)
```
ajpallares, line 103: Just making sure: at this point we don't need to keep track of those blobRefs in `blobRefsToKeep` that weren't found in `container`, right? I guess we download those later on demand?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -66,12 +70,24 @@ internal class RemoteConfigManager(
...
         val mergedBlobRefs = (previousBlobRefs + response.topics.toTopicBlobRefs())
             .filterKeys { it in response.activeTopics }
-        diskCache.write(
+
+        // Blobs the current config still wants: the prefetch set plus any active-topic blob ref.
+        val blobRefsToKeep = response.prefetchBlobs.toSet() + mergedBlobRefs.values.flatten()
+
+        // Persist the configuration first: it is the source of truth (the manifest the server diffs against),
+        // whereas inline blobs are recoverable over the network. Only touch the blob store
+        // once the state is durably committed, so a failed persist never orphans or evicts blobs.
```
ajpallares, line 90: Shouldn't this be the other way around? First, persist the recoverable/discardable info, and finally persist the source of truth (configuration). If there were a crash in the middle, the system would heal itself. Otherwise, we would be persisting a manifest that references blobs that really weren't saved 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -42,15 +46,15 @@ internal class RemoteConfigManager(
             domain = persisted?.domain ?: DEFAULT_DOMAIN,
             // Opaque manifest replayed verbatim; null on the first run when nothing is persisted yet.
             manifest = persisted?.manifest,
-            // No blob store yet (Phase 3 reports the refs actually cached); the SDK holds nothing here.
-            prefetchedBlobs = emptyList(),
+            // Report only the prefetch blobs we actually hold, so the server stops re-inlining them.
+            prefetchedBlobs = persisted?.prefetchBlobs?.filter { blobStore.contains(it) } ?: emptyList(),
```
ajpallares: The documentation for `prefetched_blobs` is 
> Blob refs the SDK has actually cached locally.

I know the `persisted` returns the `PersistedRemoteConfigurationState` source of truth. But if I'm not mistaken, here we're actually sending the last config response's `"prefetch_blobs"` field, which according to khepri's doc, contains

> Blob refs the SDK should have cached for the resolved configuration.

So, my question is, shouldn't we send here the blobs we actually know we have cached? 

If not, should we add some validation when handling the config endpoint's response to make sure that we have locally all the blobs claimed by the response's `"prefetch_blobs"` field?

ajpallares (commented):
I think this is looking great! I have some comments and questions

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: That's correct! That's planned to be executed later, depending on whether the `prefetch` key in them is set to true, or on demand later on.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Well, I've been a bit of back and forth with this 😅 The current idea is that, even if we miss the inline content (like if there is a crash while saving those), they should still be available to obtain through another request as long as we have the configuration persisted from which we can know what blobs are missing. And consumers will need to wait until those requests are done fetching before actually using them. So at most it should only cause a delay in getting the actual blobs I think. Do let me know if you have any questions since I've been a bit back and forth with this, and @rickvdl also asked a similar question above 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: >  shouldn't we send here the blobs we actually know we have cached?

Yes, so the reasoning here is to get all the blobs the backend expects the SDK to fetch ("prefetch_blobs"), then, filter those that can be found in the `blobStore`, that is, that have already been stored, either from inline, or downloaded, and that's the list of blobs that we have already stored right? So it's possible I'm missing something, if so, please let me know 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: > they should still be available to obtain through another request as long as we have the configuration persisted from which we can know what blobs are missing

This makes sense, but wouldn't this require some extra logic? Seems to me that like the other way around would work automatically (take this with a grain of 🧂, since I haven't thought about this all for so long)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: I understand, and, In theory yes. My approach is more pessimistic, in the sense that it doesn't assume that prefetchBlobs is actually exhaustive, i.e. potentially there could be a stored blob that is not included in `prefetch_blobs`. Probably that means that something failed after receiving the last response from the config endpoint. But my principle was: try not to assume that there's consistency. Like, we send what the SDK actually has cached (using the `cachedRefs()` method of this PR). Again, it's pessimistic, but potentially the same. Not a strong preference and also, I might be missing something as well 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: I think that's a valid argument indeed, let's go with that!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Hm if I'm reading your comment correctly you'd mean that in cases where we have *more* blobs in the blob storage than the config endpoint expects us to have, we'd not send them in the request?

I guess in that case the response would essentially be the same right? The config endpoint would reply with the blobs it want us to have, which wouldn't include those, and we'd clean them up?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Yep I think the new approach makes sense. As to the extra logic, I think the `TopicHandler` (or what this will be called in the end) would have an async way to fetch the blobs anyways, so it could come from disk, but if missing would still fetch it externally.

One caveat here, I think that would assume the blob (that failed to write to disk) is available externally through the CDN. If it was inlined we'd have to re-request the config endpoint? Maybe that's what you meant @ajpallares?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: > in cases where we have more blobs in the blob storage than the config endpoint expects us to have, we'd not send them in the request?

Exactly. 

My main argument for sending the actually-stored blobs was this documentation:
> Blob refs the SDK has actually cached locally.

The only way we're sure that we're sending "the blob refs that the SDK has cached locally" is reading from the files. The last-received `"prefetch_blobs"` should match that, but they could diverge in theory (some crash in the middle, some implementation error).

> I guess in that case the response would essentially be the same right? The config endpoint would reply with the blobs it want us to have, which wouldn't include those, and we'd clean them up?

I think so, yes. The system would heal itself. So not a big deal. I mostly wanted to highlight to have this discussion and make sure we want this approach.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: > This makes sense, but wouldn't this require some extra logic?

Yes, that should be part of the eventual blob fetcher, that will use the `blob` sources url + the blob ref on each topic to download the blob on demand. (Not done in this PR). 

> Seems to me that like the other way around would work automatically

Right... my only small concern is that if there is a lot of inline blobs, we would potentially be slowing down getting other data from the config like `sources` until everything has been written to disk... But yeah, It should work fine as well.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Right... in my mind, we only really need to care about what the backend expects us to have (the `prefetch_blobs`), and if there is anything else in disk, we don't really care about it, and as you said, it should be cleaned later so I think doing it this way, we avoid having to do an extra read from disk and send the backend blobs that are not even needed at all anymore. 

If there was something missing on disk, that does mean the backend needs to resend the blob, so it would behave as expected.

Do you see any downside of continuing using the `prefetch_blobs` with this?

rickvdl (approved):
Amazing work! Thanks for iterating 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: Fair points.

We've discussed this internally on Slack and we've decided to keep it this way (store config first and then blobs). Thanks everyone!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: Makes sense! 

> we avoid having to do an extra read from disk

If I'm not mistaken, here
```
prefetchedBlobs = persisted?.prefetchBlobs?.filter { blobStore.contains(it) } ?: emptyList(),
```
we're doing a disk read per item in `prefetchedBlobs` anyway. But again, probably similar to reading all items in the `blobStore`

ajpallares (approved):
Makes sense! Thanks for all the iteration and all the discussions! 😄

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Hmm true... This might be something we might want to optimize (so only do one read and keep it in memory) but can probably come in follow-up PRs.

ajpallares: Indeed 👍

PR #3616, opened 2026-06-16
Report ad event capture_method (adapter vs manual)

polmiro:
### Checklist

### Motivation
We can't currently tell whether an ad event was auto-captured by a RevenueCat ad-network adapter or reported by a manual/hybrid integration, which hinders debugging customer integration issues.

### Description
Adds a `capture_method` dimension (`adapter` | `manual`) to every ad event, stamped at the SDK's emit paths: the public `trackAd*` API stamps `manual` (covering native custom integrations and all hybrid SDKs that bridge through it), while new `@InternalRevenueCatAPI` overloads let RevenueCat ad-network adapters stamp `adapter` (used by the bundled AdMob adapter).

No public API change. Events from pre-feature SDK versions omit the field, which the backend treats as `unknown`.

---

> [!NOTE]
> **Medium Risk**
> Changes the ad event schema and all emit paths (including revenue), but public APIs are unchanged and `capture_method` is optional for backward compatibility.
> 
> **Overview**
> Adds a **`capture_method`** dimension on every ad analytics event so the backend can tell **adapter auto-capture** from **manual / hybrid** reporting.
> 
> **Core SDK:** New internal `AdCaptureMethod` (`adapter` | `manual`). All `AdEvent` variants carry `captureMethod`. Public `AdTracker.trackAd*` APIs still behave the same and default to **`manual`**; `@InternalRevenueCatAPI` overloads accept an explicit method. Stored/backend ad payloads include optional `capture_method` (older SDKs omit it → backend `unknown`).
> 
> **AdMob module:** New `trackFromAdapter` helpers route adapter instrumentation through the internal overloads with **`ADAPTER`**. Load helpers, listeners, full-screen callbacks, and paid-event wiring now use those instead of calling `trackAd*` directly.
> 
> Tests updated across `AdTracker`, backend conversion, `EventsManager` JSON, and AdMob tracking behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96a6e4a0d867d38ab99cbb4ef7794a1bf88ce402. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
@@ -11,8 +11,12 @@ import com.revenuecat.purchases.common.events.EventsManager
 
 /**
  * Tracks ad-related events such as ad displays, opens, and revenue.
+ *
+ * Events tracked through the public `trackAd*` API are stamped with
+ * [AdCaptureMethod.MANUAL]. RevenueCat ad-network adapters use the
+ * [InternalRevenueCatAPI] overloads to stamp [AdCaptureMethod.ADAPTER] instead.
  */
-@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class, InternalRevenueCatAPI::class)
+@OptIn(InternalRevenueCatAPI::class)
 @ExperimentalPreviewRevenueCatPurchasesAPI
 public class AdTracker internal constructor(
     private val eventsManager: EventsManager,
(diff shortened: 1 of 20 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/ads/events/AdTracker.kt
```diff
@@ -23,7 +27,14 @@ public class AdTracker internal constructor(
...
      */
-    public fun trackAdDisplayed(data: AdDisplayedData) {
+    public fun trackAdDisplayed(data: AdDisplayedData): Unit =
+        trackAdDisplayed(data, AdCaptureMethod.MANUAL)
+
+    /**
+     * Tracks an ad displayed event, stamping the capture method that emitted it.
+     */
+    @InternalRevenueCatAPI
+    public fun trackAdDisplayed(data: AdDisplayedData, captureMethod: AdCaptureMethod) {
```
tonidero, line 37: Hmm I'm wondering if it would make more sense to just add a new `@InternalRevenueCatAPI` property/constructor to `AdDisplayedData` (and others) instead of a new track method... But well, it feels either could work, and since it's annotated as internal, we can potentially change it later if needed so let's keep it as is.

tonidero (approved):
Just curious on the branch name 😅 But it looks great!

tonidero:
Also, I'm just wondering if we should change the label with `pr:other`... It does add some extra data, but feels like it can just be a `pr:other`?

polmiro:
Updated the label, your suggestion to use a different constructor sounds good but I agree we can easily improve this in the future so I'll keep as is to keep moving forward.

PR #3620, opened 2026-06-16
feat(workflows): drive paywall-event gating by step screen_type

facumenzella:
Implements per-step paywall impression gating on Android, driven by the backend's `screen_type` tag. iOS shipped an earlier variant of this feature; it will be updated to match the behavior in this PR (on non-paywall steps, track only workflow events rather than leaving purchases unattributed).

## Motivation

Workflows need the SDK to identify which steps are paywalls so paywall events are tracked per step (WFL-334). The backend (khepri #21429) tags each step's `metadata.screen_type`: `["paywall"]` for the step whose id equals the workflow's `single_step_fallback_id`, and `[]` for every other step.

## What changed

- **Parse** (`WorkflowModels.kt`): adds `WorkflowScreenType` (`screen_type` / `paywall` constants) and `WorkflowStep.stepScreenType: List<String>?`, reading `metadata.screen_type`. `null` = untagged, `[]` = tagged non-paywall. A present-but-non-array value (`null`, scalar, object) is treated as untagged, matching iOS.
- **Gate** (`PaywallViewModel.kt`): `WorkflowStep.tracksPaywallEvents()` now reads `screen_type`. A step reports paywall events when its `screen_type` contains `paywall`, and is suppressed when tagged without `paywall` (including `[]`). When `screen_type` is absent (untagged / pre-rollout / legacy payload) it falls back to the prior structural inference `id == singleStepFallbackId`, so untagged workflows behave exactly as on `main` (only the fallback step reports) until the backend republishes them.

<details>

# AI Context

## Metadata
- PR: #3620, branch `port/7021-workflows-screen-type-impressions`
- Agent: Claude Code (Opus 4.8, 1M context)
- Related: backend khepri #21429. iOS shipped an earlier variant (purchases-ios #7021); it will be updated to match this PR's non-paywall-step behavior (track workflow events only).
- Reviewers: Codex `/codex:review` + `/codex:adversarial-review` (review-fix loop, converged on this final diff)

## Goal
Consume the backend's per-step `screen_type` on Android and use it to gate which workflow steps report paywall events.

## Key decisions (human-owned)
- **Reconcile approach**: re-port onto `main`'s existing `tracksPaywallEvents` scaffold (swap the signal to `screen_type`), rather than keeping a parallel mechanism or abandoning the port. `main` had concurrently shipped the feature via `id == singleStepFallbackId`.
- **Untagged behavior**: untagged steps fall back to `main`'s structural inference (`id == singleStepFallbackId`), so untagged workflows behave exactly as before the rollout (only the fallback step reports). An earlier revision had untagged steps always report (literal iOS semantics); that was reversed after review (vegaro) showed every step of an un-republished workflow would over-report a `paywall_impression`, which is worse than `main` on the very metric this feature exists to fix.

## Agent contribution
- Parsing: `stepScreenType` reads `metadata.screen_type` via cast-to-`JsonArray` then `mapNotNull` of string primitives; non-array shapes return `null` (faithful to iOS `case .array` guard).
- Gate: `tracksPaywallEvents(workflow)` reads `screen_type` (contains `paywall` → report; tagged without `paywall` → suppress; untagged → fall back to `id == workflow.singleStepFallbackId`).
- Tagged the `makeContextPackageWorkflow` test fixture so existing suppression tests hold under the new signal.

## Decision log / course corrections
- Initial implementation built a **parallel** gate (`shouldReportPaywallImpression`, `paywallPresentationReportsImpression`, two-part fingerprint, an `InternalPaywall` `LaunchedEffect` key) before discovering `main` already had `tracksPaywallEvents`. Codex review + adversarial-review on that approach surfaced (and the agent fixed) a same-screen-navigation stale-gate bug. That entire approach was **discarded** after the rebase revealed `main`'s mechanism; the same-screen concern is moot because `main` sets the gate per step at state-build time.
- A rebase onto `origin/main` (22 commits ahead) exposed the conflict; the branch was reset to `origin/main` and re-implemented as the smaller Option A diff.

## Files
- `purchases/.../common/workflows/WorkflowModels.kt` — `WorkflowScreenType`, `WorkflowStep.stepScreenType`.
- `ui/revenuecatui/.../data/PaywallViewModel.kt` — `tracksPaywallEvents()` reads `screen_type`.
- Tests: `WorkflowModelsDeserializationTest`, `PaywallViewModelWorkflowTest`.

## Validation
- `./gradlew detektAll`: green (also runs as the pre-commit hook).
- `:purchases:testDefaultsBc8DebugUnitTest --tests WorkflowModelsDeserializationTest`: 18 pass (8 new).
- `:ui:revenuecatui:testDefaultsBc8DebugUnitTest --tests PaywallViewModelWorkflowTest --tests PaywallViewModelTest`: pass, 0 failures (PaywallViewModelWorkflowTest 57).

## Review focus
- Confirm building on `main`'s `tracksPaywallEvents` flag (vs. iOS's per-view threading) is the right altitude on Android.

## Validation gaps
- Untagged steps fall back to `id == singleStepFallbackId`. When `singleStepFallbackId` itself is `null` this reports on no step (faithful to `main`); a `?: true` "report everything" fallback was considered and not adopted to keep zero divergence from `main`.
- No on-device test of navigating to a `[]`-tagged step and purchasing there (covered at unit level via the per-step flag and `trackComponentInteraction` path on `main`).
</details>

---

> [!NOTE]
> **Medium Risk**
> Changes which workflow steps emit paywall analytics; incorrect null vs empty handling could over- or under-report impressions, though tests and conservative malformed-json behavior mitigate this.
> 
> **Overview**
> Workflow paywall **impression/close** gating now uses the backend **`metadata.screen_type`** tag instead of only inferring from **`single_step_fallback_id`**.
> 
> **Parsing:** `WorkflowScreenType` and **`WorkflowStep.stepScreenType`** read `screen_type` as a string array; **null** means untagged, **[]** means explicitly non-paywall; malformed non-array values are treated as untagged (iOS-aligned).
> 
> **Gating:** `tracksPaywallEvents` reports when the tag includes **`paywall`**, suppresses when tagged without it, and for untagged steps still falls back to **`id == singleStepFallbackId`** so legacy workflows match pre-rollout behavior. Multi-step test fixtures now set **`screen_type`** on context vs paywall steps; new unit tests cover parsing and gating.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f352b933491035f544ade542f717a998db6180b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
@@ -16,8 +16,10 @@ import com.revenuecat.purchases.utils.serializers.URLSerializer
 import kotlinx.serialization.Contextual
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
 import java.net.URL
 
 @InternalRevenueCatAPI
@@ -59,6 +61,22 @@ public data class WorkflowTrigger(
     @SerialName("component_id") val componentId: String,
 )
 
+/**
+ * Step `screen_type` classification values and the metadata key they are returned under.
+ *
+ * The backend tags each step with its screen classification inside `metadata` (`metadata.screen_type`),
+ * e.g. `["paywall"]`. The SDK uses this to decide which workflow steps report paywall events.
+ */
+@InternalRevenueCatAPI
+public object WorkflowScreenType {
+
+    /** The key the classification list is stored under inside a step's `metadata`. */
+    public const val METADATA_KEY: String = "screen_type"
+
+    /** A step whose screen is a paywall; it should report paywall impression/close events. */
+    public const val PAYWALL: String = "paywall"
+}
+
 @InternalRevenueCatAPI
 @Serializable
 public data class WorkflowStep(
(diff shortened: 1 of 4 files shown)


vegaro:
I was testing the latest changes and it looks like we are still sending a paywall impression event on every step

Captura de pantalla 2026-06-17 a las 15 00 55

fwiw, this workflow doesn't contain `screen_type` because it's not been republished

vegaro (approved):
tested it and it works great

PR #3621, opened 2026-06-16
refactor(remote-config): user-scoped invalidation

tonidero:
### Description
This adds support so we clear all the remote config data and stop (or ignore) any in progress operations when the user changes, so we make sure we get fully up to date data for each user.

---

> [!NOTE]
> **Medium Risk**
> Changes concurrency and cache lifecycle for remote config; incorrect epoch/lock behavior could block refreshes or persist wrong-user data, though the PR adds targeted race tests.
> 
> **Overview**
> Adds **identity-scoped remote config invalidation** via `clearCache()`, which wipes disk cache and blob store, cancels in-flight coroutine work, bumps an **epoch**, and releases the refresh guard under a shared lock so config cannot leak across users.
> 
> **`refreshRemoteConfig`** now captures epoch when starting a sync and **drops** late `/v1/config` success/error callbacks when the epoch changed (e.g. after `clearCache()`), since the HTTP call cannot be cancelled. The **200** persist path runs on an injectable **`CoroutineScope`**; persist is re-checked under `cacheLock` so a wipe cannot race with writing the previous user’s config. Guard release uses **`releaseGuardIfOwned`** so stale callbacks do not clear the guard for a newer sync.
> 
> Also adds **`close()`** for scope teardown, handles **`RCContainerFormatException`** on decode like parse failures, and extends unit tests (epoch drops, concurrent persist vs clear, decode failure releases guard) with **`UnconfinedTestDispatcher`** for the async 200 path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d22c847dbc337eaf1c8420cb4d728abf5d6d28e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -8,20 +8,33 @@ import com.revenuecat.purchases.common.debugLog
 import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.networking.RCContainer
 import com.revenuecat.purchases.common.networking.RCContainerFormatException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.cancelChildren
+import kotlinx.coroutines.launch
 import kotlinx.serialization.SerializationException
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
 
 /**
  * Orchestrates a single `/v1/config` sync: replays the persisted opaque manifest, then on `204` keeps the cache
  * untouched and on `200` persists the fresh server manifest plus the full per-topic item index — the
- * configuration (incl. each item's inline content) is the source of truth, and persisting it is the entire
- * commit. Only then, gated on a successful persist, it writes the inlined blobs the resolved config still wants
- * and prunes the rest. It reports which prefetch blobs are now cached locally on the next request.
+ * configuration (incl. each item's inline content) is the source of truth, and persisting it is the **entire**
+ * sync commit, advancing the manifest unconditionally. Only then, gated on a successful persist, it writes the
+ * inlined blobs the resolved config still wants and prunes the rest; a missing or un-parseable blob is
+ * recoverable later (fetched on demand in a future phase) and never blocks the commit. It reports which prefetch
+ * blobs are now cached locally on the next request.
  *
  * The manifest is opaque (stored and replayed verbatim); the active-topic set and removed-topic detection come
- * from the response's [RemoteConfiguration.activeTopics]. Only blobs that arrive inlined in the response are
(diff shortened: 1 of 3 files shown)


tonidero:
* **#3679** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3679?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3682](https://github.com/RevenueCat/purchases-android/pull/3682) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3682?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3683](https://github.com/RevenueCat/purchases-android/pull/3683) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3683?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3621** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3621?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -51,27 +76,70 @@ internal class RemoteConfigManager(
             // Report only the prefetch blobs we actually hold, so the server stops re-inlining them.
             prefetchedBlobs = persisted?.prefetchBlobs?.filter { storedBlobs.contains(it) } ?: emptyList(),
             onSuccess = { container, _ ->
-                // Hold the in-flight guard until persistence completes so a concurrent refresh can't read the
-                // disk cache mid-write and merge against a stale snapshot.
-                try {
-                    if (container != null) {
+                if (epoch.get() != requestEpoch) {
```
ajpallares, line 87: Clever solution!

ajpallares (approved):
Makes sense to me!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -51,27 +83,77 @@ internal class RemoteConfigManager(
...
```
rickvdl, line 87: Hm shouldn't all `epoch.get()` and `isRefreshing.set()` here be guarded with the lock synchronization? Maybe we can introduce a helper?

rickvdl (approved):
Looking good! Just one comment about a possible race condition still.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: I applied a [small change](https://github.com/RevenueCat/purchases-android/pull/3621/commits/15c50509d7766646aa02e15d3f1c80b2c7f4bc87) in other places to mark it as synchronized there to avoid some possible race conditions there. As for this one, I believe it shouldn't be an issue since the idea of this check is to check if it changed from the latest epoch stored for this attempt. So I believe it shouldn't be an issue 🙏

PR #3622, opened 2026-06-17
Introduce offline paywall snapshot testing (purchases-ui-testing + recorder plugin)

skydoves:
SDK consumers currently have no way to snapshot test their server driven paywalls (Paywalls V2 "components"). The only public entry point, `Paywall(options)`, needs a configured `Purchases` instance, a network call for offerings, and Google Play Billing products, so it cannot run in a JVM unit test.

This PR adds a small, supported way to render a consumer's paywall fully offline (no `Purchases.configure()`, no network, no Billing) so it can be captured with Paparazzi or Roborazzi:

1. A new published library, `purchases-ui-testing`, that renders a paywall from a recorded fixture.
2. A new Gradle plugin that records those fixtures (offerings JSON plus the CDN images) from a public SDK key.

The render path reuses the exact same composable the SDK uses in production (`LoadedPaywallComponents`), so the snapshot matches what the app renders, with the only fabricated input being prices (which are not part of the offerings response).

There are new modules:
- `:ui:revenuecatui-testing` (artifact `purchases-ui-testing`): the consumer facing API. `PaywallFixtures.load()`, `PaywallFixtureView`, `PaywallFixtureViewOptions`, and `PaywallFixturesTestRule`.
- `:paywall-fixtures-plugin` (plugin id `com.revenuecat.purchases.paywallfixtures`): registers the `recordPaywallFixtures` task and, when the consumer applies Paparazzi, wires up the test dependency and `testOptions`.

Supporting changes:
- `:purchases`: `FixtureOfferingsFactory` (annotated `@InternalRevenueCatAPI`) builds `Offerings` from an offerings shaped JSON, substituting `TestStoreProduct`s for real store products. The hardcoded preview products were extracted so the previews and the factory share them.
- `:ui:revenuecatui`: a thin internal bridge (`ComponentsPaywallForTesting`, also `@InternalRevenueCatAPI`) that runs validate then state then `LoadedPaywallComponents` with an injected image loader. Promoting these internals as hidden API keeps `api.txt` unchanged.
- `examples/paywall-tester`: a consumer perspective `PaywallSnapshotTest` plus committed snapshot images, exercising only the kit's public API.

### 1. Version catalog

The plugin and the kit are versioned with the SDK, so use the same version as your `purchases` dependency. Pick a Paparazzi version compatible with your AGP.

```toml
# gradle/libs.versions.toml
[versions]
paparazzi = "1.3.5"

[plugins]
paparazzi = { id = "app.cash.paparazzi", version.ref = "paparazzi" }
paywallfixtures = { id = "com.revenuecat.purchases.paywallfixtures", version = "<your purchases SDK version>" }

[libraries]
revenuecat-ui-testing = { module = "com.revenuecat.purchases:purchases-ui-testing", version = "<your purchases SDK version>" }
```

### 2. Module build.gradle.kts

Apply Paparazzi (your chosen version) and the recorder plugin. The kit test dependency and `testOptions.unitTests.isIncludeAndroidResources` are wired automatically when Paparazzi is present.

```kotlin
plugins {
    alias(libs.plugins.paparazzi)
    alias(libs.plugins.paywallfixtures)
}
```

### 3. The test

```kotlin
class PaywallSnapshotTest {
    @get:Rule val paparazzi = Paparazzi(deviceConfig = DeviceConfig.PIXEL_6)
    @get:Rule val paywallFixturesRule = PaywallFixturesTestRule()

    private val fixtures = PaywallFixtures.load()

    @Test fun defaultPaywall() {
        paparazzi.snapshot { PaywallFixtureView(fixtures.offering()) }
    }
}
```

### 4. Recording fixtures (where the API key is used)

`recordPaywallFixtures` calls the RevenueCat offerings API to download the offerings JSON and the paywall images into `src/test/resources/revenuecat-paywall-fixtures`. It needs your app's public SDK key, the same key you pass to `Purchases.configure` (for example `goog_...` or `appl_...`). The key is used only at record time. The snapshot tests themselves render offline from the committed fixtures and need no key.

Provide the key without committing it, either via the `REVENUECAT_API_KEY` environment variable (the default source):

```
REVENUECAT_API_KEY=<public sdk key> ./gradlew recordPaywallFixtures
```

or via the extension, for example reading a local Gradle property:

```kotlin
// build.gradle.kts
paywallFixtures {
    apiKey.set(providers.gradleProperty("revenuecatApiKey"))
    // offerings.set(setOf("default"))  // optional: only record specific offerings
}
```

Then record (or verify) the snapshots:

```
./gradlew recordPaparazziDebug   # writes golden images
./gradlew verifyPaparazziDebug   # CI: compares against committed goldens
```

### 5. Opting out

The plugin is lean: it does not bring its own Paparazzi, so applying it never forces a Paparazzi version onto your classpath. To skip the auto wiring (for example to use Roborazzi, or to manage the test dependency and `testOptions` yourself), set `paywallFixtures { setupSnapshotTesting = false }` or pass `-Prevenuecat.paywallFixtures.snapshotTesting=false`.

## Production code changes to review

Two changes touch shipped `:ui:revenuecatui` code and deserve attention:

1. Image loader injection moved from the debug source set (which had a no op twin in release) into the main source set, and the consumption sites in `RemoteImage` and `BackgroundStyle` now key off "an image loader was injected" rather than `LocalInspectionMode`. Production behavior is unchanged because nothing in production provides the loader. This is the change that lets the published (release) artifact render fixture images.
2. `PaywallResourceProvider.getXmlFontFamily` now returns null instead of crashing when the platform `getXml` returns null. layoutlib (Paparazzi) returns null where a device throws `NotFoundException`, which previously surfaced as a Kotlin intrinsic NPE. The kit also forces fonts to a system fallback during offline rendering so tests do not depend on the consumer bundling the dashboard's fonts.

---

> [!NOTE]
> **Medium Risk**
> Touches shipped revenuecatui image and font resolution paths (guarded by injected loader) and adds new published APIs/plugins; production paywall rendering should be unchanged when consumers do not opt into the testing kit.
> 
> **Overview**
> Adds **offline JVM snapshot testing** for Paywalls V2 (components) so apps can golden-test dashboard paywalls without `Purchases.configure()`, network, or Play Billing.
> 
> **New consumer surface:** published **`purchases-ui-testing`** (`PaywallFixtures`, `PaywallFixtureView`, `PaywallFixturesTestRule`) loads recorded `offerings.json` + mirrored CDN images from test resources and renders through the same **`LoadedPaywallComponents`** path as production, with fabricated **`TestStoreProduct`** prices. **`com.revenuecat.purchases.uifixtures`** registers **`recordPaywallFixtures`** (offerings API + asset download) and optionally wires Paparazzi test deps when Paparazzi is applied.
> 
> **Core SDK support:** **`FixtureOfferingsFactory`** builds `Offerings` from offerings-shaped JSON; preview defaults move to **`TestStoreProductDefaults`**. **`ComponentsPaywallForTesting`** (internal) injects a classpath image loader, fixed date, mock purchases, and font fallback for Paparazzi.
> 
> **Production `revenuecatui` tweaks (behavior unchanged when no loader is injected):** preview image loading keys off **`LocalPreviewImageLoader`** instead of inspection mode (moved to main; release no-ops removed); **`getXmlFontFamily`** tolerates layoutlib returning null.
> 
> **Example:** `paywall-tester` gets parameterized Paparazzi tests; regenerated fixture trees under test resources are gitignored while snapshot images stay committed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2b6f2c41936ef538af67904464cdb637215e19d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


paywall-fixtures-plugin/src/main/kotlin/com/revenuecat/purchases/paywallfixtures/internal/AssetUrlCollector.kt
```diff
@@ -0,0 +1,56 @@
+package com.revenuecat.purchases.paywallfixtures.internal
```
tonidero: [Not related to this line]
I've been thinking whether it could make sense to rename this module to something more generic like `ui-fixtures-plugin`? Mostly thinking about eventually supporting the different screens for workflows...

ui-fixtures-plugin/src/main/kotlin/com/revenuecat/purchases/uifixtures/internal/FixtureRecorder.kt
```diff
@@ -0,0 +1,157 @@
...
+     * Keeps only offerings that have a components paywall and, when [offeringsFilter] is non-empty, are
+     * part of the filter. Everything else in the response (current_offering_id, ui_config, placements,
+     * etc.) is recorded verbatim.
+     */
+    private fun filterOfferings(response: JSONObject): JSONObject {
+        val offerings = response.getJSONArray("offerings")
+        val kept = JSONArray()
+        for (i in 0 until offerings.length()) {
+            val offering = offerings.getJSONObject(i)
+            val hasComponentsPaywall = offering.has("paywall_components")
```
tonidero, line 94: I think we will need to change this soon, due to workflows (since information about paywalls will come from a separate endpoint) cc @vegaro . Should hopefully be not a huge change 🤞

tonidero (commented):
Still reviewing but looking great so far!!

paywall-fixtures-plugin/src/main/kotlin/com/revenuecat/purchases/paywallfixtures/internal/AssetUrlCollector.kt
skydoves: Agree with this!

PR #3627, opened 2026-06-18
fix: attribute presented offering to queue-synced active purchases

ajpallares:
Part of [SDK-4373](https://linear.app/revenuecat/issue/SDK-4373).

### Motivation
When a just-purchased transaction is discovered via `queryPurchases` (synced as an unsynced active purchase) before or instead of the BillingClient purchase callback, its receipt is POSTed without the presented offering, showing up as "Unknown" offerings in charts.

### Description
- `makePurchaseAsync` stamps the in-memory `PurchaseContext` with an `initiationDate`.
- `queryPurchases` results pass through a new `attachCachedPurchaseContext` helper that back-fills the cached presented offering context onto transactions missing one, guarded by `purchaseTime >= initiationDate`.

Limitation: the context cache is in-memory, so this only recovers attribution for the same-session race, not purchases synced in a later session.

---

> [!NOTE]
> **Medium Risk**
> Touches purchase/receipt attribution on the Google billing path; wrong time gating could mis-attribute offerings, but scope is narrow and guarded by initiation date and existing context checks.
> 
> **Overview**
> Fixes **Unknown** offering attribution when a new Google Play purchase is synced via `queryPurchases` (unsynced active purchases) before the BillingClient purchase callback runs.
> 
> `makePurchaseAsync` now records **`initiationDate`** on the in-memory `PurchaseContext`. **`queryPurchases`** passes results through **`attachCachedPurchaseContext`**, which copies the cached context (including **presented offering**) onto transactions that lack it, only when **`purchaseTime >= initiationDate`** so older purchases are not mis-tagged.
> 
> Recovery is **same-session only** because the cache is in-memory.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2501a1c92dc966e7829732866839f4e0dbc57fcd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ajpallares:
I'm thinking that in order not to introduce misattributions with this PR, we need to merge #3165 first. This other PR wasn't really needed by itself, but this PR makes it important to avoid adding the attribution from a cancelled purchase. WDYT @RevenueCat/coresdk?

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
```diff
@@ -1723,6 +1723,104 @@ class BillingWrapperTest {
...
+            onError = { fail("should be a success") },
+        )
+
+        assertThat(purchasesByHashedToken).isNotNull
+        val transaction = purchasesByHashedToken!!.values.first()
+        assertThat(transaction.presentedOfferingContext).isEqualTo(presentedOfferingContext)
+    }
+
+    @Test
+    fun `queryPurchases does not attach cached presented offering context when purchase time is before initiation date`() {
```
tonidero, line 1765: Should we also have a test for when there is a cached context for a different productId? (Shouldn't attach I think)

tonidero (approved):
I think this looks good! Just a possible missing test.

Regarding the other PR. I think it wouldn't be bad to add it indeed... We've confirmed the behavior correct? If so I think it makes sense. I would just see if we need to address the existing comment (I've lost context on that PR TBH)

purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
ajpallares: Ohh true. Good call!

PR #3637, opened 2026-06-22
refactor(ads): move reward verification primitives to core

peterporfy:
### Checklist

Also tested it fully locally with the sample admob app and the refactored code.

### Motivation

The goal of this change is purely technical: move two rewarded ad primitives to the core module so in a future iteration it will be possible to work with it from hybrids - and it makes it possible today for app developers to roll their own hybrid bridges, if they would choose to do so.

### Description

The two primitives are:

1. `generateRewardVerificationToken`
2. `pollRewardVerification`

These can be used to implement a semi-custom reward verification flow without having our admob adapter as dependency.

#### Main changes

- carved out admob-adapter-independent code parts to the core module
- moved and adapted tests to the new code location
- we are now using the core primitives from the admob adapter and it is purely for convenience

---

> [!NOTE]
> **Medium Risk**
> Touches experimental public SDK surface and rewarded-ad/virtual-currency correlation, but behavior is largely relocated with deprecated aliases and broad test updates rather than new product logic.
> 
> **Overview**
> Moves rewarded-ad **server verification** building blocks from the AdMob feature module into **`com.revenuecat.purchases.ads.rewardverification`** on the core `Purchases` SDK, so apps (and future hybrids) can run verification without depending on the AdMob adapter.
> 
> **New public primitives on `Purchases`:** `generateRewardVerificationToken(impressionId)` (SSV custom data + `clientTransactionId`), `pollRewardVerification` / `awaitPollRewardVerification`, plus `RewardVerificationToken`, `RewardVerificationResult`, and `VerifiedReward`. Polling, retry/backoff, and user-facing failure strings live in core; verified **virtual currency** rewards now invalidate the virtual currencies cache from `Purchases` (removed from the AdMob runtime).
> 
> The AdMob module is rewired to call those APIs for SSV setup and polling; former `com.revenuecat.purchases.admob` result types are **deprecated typealiases** to the new package. The internal per-request backend status type is renamed to **`RewardVerificationPollStatus`** to avoid clashing with the public poll result. API dumps, samples, and tests follow the new types.
> 
> **CI:** `Dangerfile` counts production line churn via `git.diff.stats[:files]` with a nil guard so renamed files no longer crash the size check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a8feed858b2aca1b7cef466f890b402f5861975. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
+++ b/feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
@@ -1,0 +1,17 @@
+package com.revenuecat.purchases.admob
+
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+
+/**
+ * Moved to [com.revenuecat.purchases.ads.rewardverification.VerifiedReward].
+ */
+@Deprecated(
+    "Moved to com.revenuecat.purchases.ads.rewardverification.VerifiedReward",
+    ReplaceWith(
+        "VerifiedReward",
+        "com.revenuecat.purchases.ads.rewardverification.VerifiedReward",
+    ),
+)
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+public typealias VerifiedReward =
+    com.revenuecat.purchases.ads.rewardverification.VerifiedReward
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -781,7 +781,7 @@ public class Purchases internal constructor(
         return pollRewardVerification(clientTransactionId) { Poller.poll(it) }
     }
 
-    @ExperimentalPreviewRevenueCatPurchasesAPI
+    @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
     internal suspend fun pollRewardVerification(
         clientTransactionId: String,
         poll: suspend (String) -> RewardVerificationResult,
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -9,6 +9,9 @@ import androidx.annotation.VisibleForTesting
 import com.revenuecat.purchases.Purchases.Companion.configure
 import com.revenuecat.purchases.Purchases.Companion.debugLogsEnabled
 import com.revenuecat.purchases.ads.events.AdTracker
+import com.revenuecat.purchases.ads.rewardverification.Poller
+import com.revenuecat.purchases.ads.rewardverification.RewardVerificationResult
+import com.revenuecat.purchases.ads.rewardverification.VerifiedReward
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.PlatformInfo
 import com.revenuecat.purchases.common.errorLog
@@ -749,6 +752,45 @@ public class Purchases internal constructor(
         onError: (PurchasesError) -> Unit,
     ) {
         purchasesOrchestrator.createSupportTicket(email, description, onSuccess, onError)
+    }
+
+    /**
+     * Generates a reward verification token for a loaded rewarded ad.
+     *
+     * Call after the ad has loaded. Forward [RewardVerificationToken.customData] and
+     * [RewardVerificationToken.appUserID] to your ad network's server-side verification options, then keep
+     * [RewardVerificationToken.clientTransactionId] for use with [pollRewardVerification] when the reward
+     * callback fires.
+     *
+     * @param impressionId The ad network's impression identifier for the loaded ad.
+     */
+    @InternalRevenueCatAPI
+    public fun generateRewardVerificationToken(impressionId: String): RewardVerificationToken {
+        return purchasesOrchestrator.generateRewardVerificationToken(impressionId)
+    }
+
+    /**
+     * Polls the backend until reward verification completes or the attempt budget is exhausted.
+     *
+     * Call when your ad network's reward callback fires, passing the `clientTransactionId` returned by
+     * [generateRewardVerificationToken]. Invalidates the virtual currencies cache automatically on a
+     * verified virtual-currency reward.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    public suspend fun pollRewardVerification(clientTransactionId: String): RewardVerificationResult {
+        return pollRewardVerification(clientTransactionId) { Poller.poll(it) }
+    }
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    internal suspend fun pollRewardVerification(
+        clientTransactionId: String,
+        poll: suspend (String) -> RewardVerificationResult,
+    ): RewardVerificationResult {
+        val result = poll(clientTransactionId)
+        if (result.verifiedReward is VerifiedReward.VirtualCurrency) {
+            purchasesOrchestrator.invalidateVirtualCurrenciesCache()
+        }
+        return result
     }
 
     @OptIn(InternalRevenueCatAPI::class)
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -572,6 +572,7 @@ public class Purchases internal constructor(
     public fun close() {
         notifyLifecycleClosed()
         purchasesOrchestrator.close()
+        rewardVerificationPollLauncher.close()
         backingFieldSharedInstance = null
     }
 
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -9,7 +9,7 @@ import androidx.annotation.VisibleForTesting
 import com.revenuecat.purchases.Purchases.Companion.configure
 import com.revenuecat.purchases.Purchases.Companion.debugLogsEnabled
 import com.revenuecat.purchases.ads.events.AdTracker
-import com.revenuecat.purchases.ads.rewardverification.Poller
+import com.revenuecat.purchases.ads.rewardverification.RewardVerificationPollLauncher
 import com.revenuecat.purchases.ads.rewardverification.RewardVerificationResult
 import com.revenuecat.purchases.ads.rewardverification.RewardVerificationToken
 import com.revenuecat.purchases.ads.rewardverification.VerifiedReward
@@ -31,6 +31,7 @@ import com.revenuecat.purchases.interfaces.GetStorefrontCallback
 import com.revenuecat.purchases.interfaces.GetStorefrontLocaleCallback
 import com.revenuecat.purchases.interfaces.GetVirtualCurrenciesCallback
 import com.revenuecat.purchases.interfaces.LogInCallback
+import com.revenuecat.purchases.interfaces.PollRewardVerificationCallback
 import com.revenuecat.purchases.interfaces.PurchaseCallback
 import com.revenuecat.purchases.interfaces.ReceiveCustomerInfoCallback
 import com.revenuecat.purchases.interfaces.ReceiveOfferingsCallback
@@ -68,6 +69,9 @@ import kotlin.coroutines.suspendCoroutine
 public class Purchases internal constructor(
     @get:JvmSynthetic internal val purchasesOrchestrator: PurchasesOrchestrator,
 ) : LifecycleDelegate {
+
+    private val rewardVerificationPollLauncher = RewardVerificationPollLauncher()
+
     /**
      * The current configuration parameters of the Purchases SDK.
      */
@@ -789,11 +793,19 @@ public class Purchases internal constructor(
      *
      * Call when your ad network's reward callback fires, passing the `clientTransactionId` returned by
      * [generateRewardVerificationToken]. Invalidates the virtual currencies cache automatically on a
-     * verified virtual-currency reward.
+     * verified virtual-currency reward. The [callback] is invoked on the main thread.
+     *
+     * For coroutines, use the `awaitPollRewardVerification` suspend extension instead.
      */
     @ExperimentalPreviewRevenueCatPurchasesAPI
-    public suspend fun pollRewardVerification(clientTransactionId: String): RewardVerificationResult {
-        return pollRewardVerification(clientTransactionId) { Poller.poll(it) }
+    public fun pollRewardVerification(
+        clientTransactionId: String,
+        callback: PollRewardVerificationCallback,
+    ) {
+        rewardVerificationPollLauncher.launch(
+            poll = { awaitPollRewardVerification(clientTransactionId) },
+            onCompleted = { callback.onCompleted(it) },
+        )
     }
 
     @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -11,6 +11,7 @@ import com.revenuecat.purchases.Purchases.Companion.debugLogsEnabled
 import com.revenuecat.purchases.ads.events.AdTracker
 import com.revenuecat.purchases.ads.rewardverification.Poller
 import com.revenuecat.purchases.ads.rewardverification.RewardVerificationResult
+import com.revenuecat.purchases.ads.rewardverification.RewardVerificationToken
 import com.revenuecat.purchases.ads.rewardverification.VerifiedReward
 import com.revenuecat.purchases.common.LogIntent
 import com.revenuecat.purchases.common.PlatformInfo
@@ -48,8 +49,10 @@ import com.revenuecat.purchases.strings.BillingStrings
 import com.revenuecat.purchases.strings.ConfigureStrings
 import com.revenuecat.purchases.utils.DefaultIsDebugBuildProvider
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencies
+import org.json.JSONObject
 import java.net.URL
 import java.util.Locale
+import java.util.UUID
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
@@ -766,7 +769,19 @@ public class Purchases internal constructor(
      */
     @ExperimentalPreviewRevenueCatPurchasesAPI
     public fun generateRewardVerificationToken(impressionId: String): RewardVerificationToken {
-        return purchasesOrchestrator.generateRewardVerificationToken(impressionId)
+        val clientTransactionId = UUID.randomUUID().toString()
+        // Keys inserted in sorted order so the serialized customData is deterministic and matches the
+        // other SDKs byte-for-byte. The backend parses by key, so order is not semantically significant.
+        val customData = JSONObject()
+            .put("api_key", purchasesOrchestrator.currentConfiguration.apiKey)
+            .put("client_transaction_id", clientTransactionId)
+            .put("impression_id", impressionId)
+            .toString()
+        return RewardVerificationToken(
+            customData = customData,
+            clientTransactionId = clientTransactionId,
+            appUserID = purchasesOrchestrator.appUserID,
+        )
     }
 
     /**
--- a/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
+++ b/purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
@@ -764,7 +764,7 @@ public class Purchases internal constructor(
      *
      * @param impressionId The ad network's impression identifier for the loaded ad.
      */
-    @InternalRevenueCatAPI
+    @ExperimentalPreviewRevenueCatPurchasesAPI
     public fun generateRewardVerificationToken(impressionId: String): RewardVerificationToken {
         return purchasesOrchestrator.generateRewardVerificationToken(impressionId)
     }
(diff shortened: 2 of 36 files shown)


purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -751,6 +754,45 @@ public class Purchases internal constructor(
...
+    }
+
+    /**
+     * Polls the backend until reward verification completes or the attempt budget is exhausted.
+     *
+     * Call when your ad network's reward callback fires, passing the `clientTransactionId` returned by
+     * [generateRewardVerificationToken]. Invalidates the virtual currencies cache automatically on a
+     * verified virtual-currency reward.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
```
tonidero, line 801: This one is annotated with `ExperimentalPreviewRevenueCatPurchasesAPI` but `generateRewardVerificationToken` is instead annotated with `InternalRevenueCatAPI`? So I guess the question is, are these APIs meant to be used by developers? if so, they should be annotated with Experimental. If not, as Internal.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -751,6 +754,45 @@ public class Purchases internal constructor(
...
+    public suspend fun pollRewardVerification(clientTransactionId: String): RewardVerificationResult {
+        return pollRewardVerification(clientTransactionId) { Poller.poll(it) }
+    }
+
+    @ExperimentalPreviewRevenueCatPurchasesAPI
```
tonidero: Since this is `internal`, there is no need to annotate it as experimental as well? Unless we were planning to expose it. In which case, same question as above.

purchases/src/main/kotlin/com/revenuecat/purchases/RewardVerificationToken.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Ties a loaded rewarded ad to its server-side reward verification.
+ *
+ * Produced by [Purchases.generateRewardVerificationToken]. Forward [customData] and [appUserID] to your
+ * ad network's server-side verification options, and keep [clientTransactionId] to correlate the reward
+ * callback with [Purchases.pollRewardVerification].
+ */
+@InternalRevenueCatAPI
+public data class RewardVerificationToken(
```
tonidero: Should we move this class inside the `ads/rewardverification` package?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -943,6 +945,23 @@ internal class PurchasesOrchestrator(
         )
     }
 
+    @OptIn(InternalRevenueCatAPI::class)
+    fun generateRewardVerificationToken(impressionId: String): RewardVerificationToken {
```
tonidero: I wonder if we should allow devs to pass custom attributes here... but that can/should be done separately if so :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Also, if we were planning to expose it, I do wonder if we need both TBH... maybe one method with an optional parameter for the poller would be better in that case.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -751,6 +754,45 @@ public class Purchases internal constructor(
...
+    /**
+     * Polls the backend until reward verification completes or the attempt budget is exhausted.
+     *
+     * Call when your ad network's reward callback fires, passing the `clientTransactionId` returned by
+     * [generateRewardVerificationToken]. Invalidates the virtual currencies cache automatically on a
+     * verified virtual-currency reward.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    public suspend fun pollRewardVerification(clientTransactionId: String): RewardVerificationResult {
+        return pollRewardVerification(clientTransactionId) { Poller.poll(it) }
```
tonidero: TBH, now that we have moved the `Poller` to the main SDK, I would love if we could refactor the architecture and avoid using static methods like this, and instead create instances of the poller and similar classes as part of the `PurchasesOrchestrator` (or another helper class if we want to clean it up better)

tonidero (commented):
Some initial comments. Lmk what you think!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: Indeed, good catch. It should be experimental for PHC and potential raw usages. The iOS version has the same. I fixed that (and ran the missing api-dump so it is a bit bigger): https://github.com/RevenueCat/purchases-android/pull/3637/changes/36d82fab734da9c87ee1625e6e953012405dc951

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: This is internal, yes 👍 the only reason we have it is because it makes testing easier (by injecting the poller). Fixed here: https://github.com/RevenueCat/purchases-android/pull/3637/changes/f77353a6fc1e288bb6cae0995292701b0f91b8e3

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
peterporfy: Hm, that might be a great idea for the future to get it back later during polling. However, for that we would need to reconsider how we store the ephemeral polling data on the backend (as it might grow big depending on what the user pushes into it). But it can be definitely a nice convenience feature :) like getting back your custom data in during SSO flows. Can come handy so you have more context when the verification resolves.

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationToken.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+public data class RewardVerificationToken(
+    /** Set as the ad network's server-side verification custom data. */
+    val customData: String,
+    /** Correlates the ad with its verification. */
+    val clientTransactionId: String,
+    /** The app user the reward is attributed to; set as the ad network's SSV user identifier. */
+    val appUserID: String,
+)
```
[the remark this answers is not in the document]
peterporfy: fixed

purchases/src/main/kotlin/com/revenuecat/purchases/RewardVerificationToken.kt
peterporfy: Addressed here in a bigger commit: https://github.com/RevenueCat/purchases-android/pull/3637/changes/36ccbdd1aa02091eacaa74732c9de8f137030889

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: Are you ok if I do this in a followup change - so this can remain scoped to exposing the 2 reward primitives in the core as is today?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Note that PHC usage can be annotated with `@InternalRevenueCatAPI`. We should only use `@ExperimentPreviewRevenueCatPurchasesAPI` if we do mean developers using purchases-android to use this (which seems we want, but just confirming)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Sure that's fine :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: yes, we would like to have it both for PHC and in general as a vanilla API (like how we have vanilla and admob adapter for tracker)

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -751,6 +757,57 @@ public class Purchases internal constructor(
...
+     *
+     * Call after the ad has loaded. Forward [RewardVerificationToken.customData] and
+     * [RewardVerificationToken.appUserID] to your ad network's server-side verification options, then keep
+     * [RewardVerificationToken.clientTransactionId] for use with [pollRewardVerification] when the reward
+     * callback fires.
+     *
+     * @param impressionId The ad network's impression identifier for the loaded ad.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    public fun generateRewardVerificationToken(impressionId: String): RewardVerificationToken {
```
tonidero, line 776: I've actually been thinking, it could be great to move these new methods to the `adTracker` class, so usage would be `Purchases.sharedInstance.adTracker.generateRewardVerificationToken(...)`. Mostly thinking of keeping things within that namespace and avoid increasing the `Purchases` class API surface too much... I'm a bit on the fence since I guess this is technically not actually tracking... Any thoughts on this @RevenueCat/coresdk ?

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -751,6 +757,57 @@ public class Purchases internal constructor(
...
+
+    /**
+     * Polls the backend until reward verification completes or the attempt budget is exhausted.
+     *
+     * Call when your ad network's reward callback fires, passing the `clientTransactionId` returned by
+     * [generateRewardVerificationToken]. Invalidates the virtual currencies cache automatically on a
+     * verified virtual-currency reward.
+     */
+    @ExperimentalPreviewRevenueCatPurchasesAPI
+    public suspend fun pollRewardVerification(clientTransactionId: String): RewardVerificationResult {
```
tonidero: I just noticed, we are only exposing the suspend function version of this. Usually we expose both a callback based API, and then add the convenience suspend function for folks using coroutines in `coroutineExtensions.kt`... I think we might want to do the same here? This will also keep it more compatible with Java.

tonidero (commented):
A couple more points but pretty close I would say!

peterporfy:
This change is part of the following stack:

- #3637 ◀
    - #3657

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: Yes, it is not really a tracker but a more integrated (with other RC features) way to use ads

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: ah, nice catch! I think we would need that for PHC as well. Going to fix this

peterporfy: @tonidero I kept the suspend function as the core internally, but following the conventions of the sdk exposed the callback style on `Purchases` - and exposed the coroutine one on the extensions. Because we are suspend fn first it needed some helpers. Let me know what you think, thanks!

https://github.com/RevenueCat/purchases-android/pull/3637/changes/c557575062f19a469026ac70e7ccc408b80e788c

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -10,6 +12,7 @@ import com.revenuecat.purchases.interfaces.GetRewardVerificationResultCallback
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencies
 import kotlinx.coroutines.suspendCancellableCoroutine
 import java.util.Locale
+import com.revenuecat.purchases.RewardVerificationResult as CoreRewardVerificationResult
```
tonidero: Hmm so we have 2 types named `RewardVerificationResult`? We should try to avoid that or it could be superconfusing... Since one of them is actually `@InternalRevenueCatAPI`, maybe we should make this explicit like `InternalRewardVerificationResult`?

purchases/src/defaults/kotlin/com/revenuecat/purchases/interfaces/PollRewardVerificationCallback.kt
```diff
@@ -0,0 +1,15 @@
+package com.revenuecat.purchases.interfaces
+
+import com.revenuecat.purchases.ExperimentalPreviewRevenueCatPurchasesAPI
+import com.revenuecat.purchases.ads.rewardverification.RewardVerificationResult
+
+/**
+ * Callback for [com.revenuecat.purchases.Purchases.pollRewardVerification].
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+public fun interface PollRewardVerificationCallback {
```
tonidero, line 10: We might want to add an `onError` callback to this so we might not want to have this be a `fun interface` but just a normal interface?

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationPollLauncher.kt
```diff
@@ -0,0 +1,24 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ * Bridges the suspend reward-verification poll to the callback-based public API, so coroutine usage stays
+ * out of [com.revenuecat.purchases.Purchases]. The poll loop only suspends (never blocks), so running it on
+ * [Dispatchers.Main] keeps the callback on the main thread like the SDK's other callback APIs.
+ */
+@OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
+internal class RewardVerificationPollLauncher(
+    private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Main),
```
tonidero: We might want to consider using IO dispatchers instead of main dispatcher for the polling? And then we can call the callback method on the main thread to keep it clean.

tonidero (commented):
Almost there! Sorry I didn't see these before 🙏

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
peterporfy: Indeed that is confusing. They are similar but exist for a different reason. I renamed it to match the ios sdk: https://github.com/RevenueCat/purchases-android/pull/3637/changes/b0f0b75333d54a816f87122af95f27e1644d8fe5

purchases/src/defaults/kotlin/com/revenuecat/purchases/interfaces/PollRewardVerificationCallback.kt
peterporfy: We are using `RewardVerificationResult` to communicate outcome and it folds any result into the binary `failed` or `verifiedReward`. It is on par with ios and I would keep it as is so the handling can happen at the same place.

tonidero: Ohh right 🤦 I completely forgot. But yeah, we should be ok then I think

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/Poller.kt
```diff
@@ -1,17 +1,18 @@
-package com.revenuecat.purchases.admob.rewardverification
+package com.revenuecat.purchases.ads.rewardverification
```
tonidero, line 1: [Not related to this line]
One thing I mention in the follow-up PR, is that we're adding all these to the `defaults` flavor instead of to the `main` source set. I think ideally all of this should be part of the main source set and just make the public API in `Purchases` be present in the `defaults` flavor of it. Won't block the PR for this, but might be a refactor we want to do for cleanup 🙏

tonidero (approved):
I think the remaining things are small so I won't block the PR with this :) Thank you for all the iterations!

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationPollLauncher.kt
peterporfy: I think it was running in the background implicitly already - but I agree it is better to be sure and safe! Fixed it here: https://github.com/RevenueCat/purchases-android/pull/3637/changes/c3462fc3d0307db6305f6695edf4cfaf92dd8290

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -751,6 +761,65 @@ public class Purchases internal constructor(
         purchasesOrchestrator.createSupportTicket(email, description, onSuccess, onError)
     }
 
+    /**
```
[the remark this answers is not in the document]
peterporfy: @tonidero fyi I fixed it here: https://github.com/RevenueCat/purchases-android/pull/3637/changes/c3331f7ea59ea42082b2636e41ae044f9c39d316

I don't think it was a super realistic case (also the worst case was questionable) but it did fit nicely into the current way of doing things and it might remove some future edge case problems. Wdyt?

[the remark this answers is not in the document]
tonidero: Right, very few people would ever close the SDK, so it's indeed quite an edge case. And also, I believe that we would create a new instance with a new scope... but yeah, since we're keeping things statically it could potentially leak a bit, so this makes sense to add 👍

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/Poller.kt
peterporfy: Makes sense, I think I will open another PR with this 👍

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationResult.kt
```diff
@@ -1,6 +1,7 @@
-package com.revenuecat.purchases.admob
```
AlvaroBrey, line 1: Since `RewardVerificationResult` is a `public` class, the package move is a breaking API change. It's acceptable because its an `@Experimental` API but we should at least include it in `CHANGELOG.md` (on release PR, see below)

We could also add a `@Deprecated typealias` on the old location, if appropriate.

The same is true for `VerifiedReward`

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
```diff
@@ -1,4 +1,4 @@
-package com.revenuecat.purchases.admob
```
AlvaroBrey, line 1: Same as above, needs at least `CHANGELOG.md` entry on release

AlvaroBrey (changes requested):

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationResult.kt
tonidero: CHANGELOG.md is generated automatically from the merged PRs so we will need to wait and remember to update it on the generated PR before making the release (can also be modified after if needed). But yeah, definitely great catch with that breaking change!!

peterporfy: @AlvaroBrey it was an internally used class before - I am not sure why it is a breaking change? However, maybe we need to mark the PR as `pr:feat` given there is the new feature available @tonidero?

AlvaroBrey: It was [a public class with a private constructor](https://github.com/RevenueCat/purchases-android/pull/3637/changes#diff-8616d52849aa21285c1aae90b9e857d77a0fe6e89e158ee90ef22e8bcae18803L9) unless the diff is playing tricks on me, so it was publicly exposed

tonidero: As mentioned ^ , the class was exposed in the admob library and now has moved to a different package, so it's indeed an API breaking change I'm afraid...

> maybe we need to mark the PR as pr:feat given there is the new feature available

Hmm well, I'm a bit on the fence since it's marked as experimental, but since we're ready for devs to use it, I'm ok adding the `pr:feat` 👍  In the end, it just affects the version number (minor, instead of patch) and the changelog generation.

peterporfy: I've added typealias for both at the old location: https://github.com/RevenueCat/purchases-android/pull/3637/changes/8059c6d6e95b3f4486362a7d966b1c686f40c521

feature/admob/src/main/kotlin/com/revenuecat/purchases/admob/VerifiedReward.kt
peterporfy: I've added typealias for both at the old location: https://github.com/RevenueCat/purchases-android/pull/3637/changes/8059c6d6e95b3f4486362a7d966b1c686f40c521

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationResult.kt
AlvaroBrey: Thank you!

AlvaroBrey (approved):

PR #3643, opened 2026-06-24
Add WeightedSourceSelector for remote config source selection

ajpallares:
### Checklist

### Motivation
Groundwork for API/blob source failover on the `/v2/config` endpoint: a reusable object that decides which source to try and in what fallback order. Not yet wired to networking.

### Description
Adds a generic `WeightedSourceSelector` that decides the order in which sources should be tried:

- **Priority first:** sources with the highest `priority` come first. A whole priority tier is exhausted before any lower-priority source is considered.
- **Weighted-random within a tier:** among sources sharing the same `priority`, the order is decided by a weighted-random draw, so a higher `weight` is more likely (but not guaranteed) to come first.

The full fallback order is computed once, at init. Callers walk it with a small API:

- `current` — the source to use right now (`null` when none remain).
- `advance()` — move to the next source in the order (e.g. after the current one fails); returns `null` once exhausted.
- `reset()` — rewind back to the first source.

Weight edge cases:

- A negative weight is treated as `0`, so it acts as a last resort within its tier.
- If every weight in a tier is `0`, that tier is ordered uniformly at random.

iOS counterpart: RevenueCat/purchases-ios#7077.

---

> [!NOTE]
> **Low Risk**
> Internal remote-config selection utility with no production wiring yet; behavior change is limited to how negative weights are handled within a tier.
> 
> **Overview**
> Replaces the one-shot `selectWeighted` list helper with **`WeightedSourceSelector`**, which **precomputes a full try order** for remote-config-style sources and exposes **`current`**, **`advance()`**, and **`reset()`** for walking that order when a source fails.
> 
> Ordering is **highest `priority` first**, then a **weight-biased random shuffle within each tier** (instead of picking only the first source). **Negative weights are clamped to 0** (last resort in the tier) rather than forcing uniform random for the whole tier; **zero-sum tiers** still use uniform random. Weight sums use **`sumOrIntMax`** to avoid overflow. **`WeightedSource`** moves into the new file with clearer docs; tests are renamed and expanded for advance/reset and the new weight rules. **Not wired to networking yet** (groundwork for `/v2/config` failover).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 68b33f1acd101c40da84f40dbbdaeeea02683d33. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSource.kt
```diff
@@ -1,31 +0,0 @@
...
-    val weight: Int
-}
-
-/**
- * Helper function to select a source from a list of [WeightedSource]s
- * - Highest priority
- * - Tied priority uses weights to choose randomly.
- * - If weights add up to 0 or any negative weight, uniform random selection.
- */
-internal fun <T : WeightedSource> List<T>.selectWeighted(random: Random = Random.Default): T? {
```
ajpallares, line 16: This helper was unused, and note that now the negative-weight handling is slightly different (only using uniformly random when all weights are ≤ 0)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelector.kt
```diff
@@ -0,0 +1,94 @@
...
+            val ordered = ArrayList<T>(remaining.size)
+            while (remaining.size > 1) {
+                ordered.add(remaining.removeAt(weightedPickIndex(remaining, random)))
+            }
+            ordered.add(remaining.first())
+            return ordered
+        }
+
+        fun <T : WeightedSource> weightedPickIndex(sources: List<T>, random: Random): Int {
+            val weights = sources.map { maxOf(0, it.weight) }
```
rickvdl, line 76: Same question as iOS here regarding capping the max weight to ensure the sum won't go over Int.max. Also we could add a test on both platforms ensuring this is covered.

rickvdl (approved):
Nice work! Solid implementation and consistent with iOS. LGTM! :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelector.kt
ajpallares: Sure! Done in https://github.com/RevenueCat/purchases-android/pull/3643/commits/a061466721f26dde5aa1670677813547c71079b0

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/WeightedSourceSelector.kt
```diff
@@ -0,0 +1,97 @@
...
+            var target = random.nextInt(totalWeight)
+            var selected = weights.lastIndex
+            for (index in weights.indices) {
+                target -= weights[index]
+                if (target < 0) {
+                    selected = index
+                    break
+                }
+            }
+            return selected
```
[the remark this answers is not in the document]
ajpallares: I think this is fine. We shouldn't really surpass Int.MAX_VALUE, so I think we can accept that the weights won't really be fully respected in these scenarios...

tonidero (approved):
Looks great!

PR #3646, opened 2026-06-25
[EXTERNAL] Remove -dontoptimize from Amazon consumer ProGuard rules

### LINKED ISSUE #3496: App not optimized by R8 due to the `-dontoptimize` rule in `feature/amazon/consumer-rules.pro`
- [ ] I have updated Purchases SDK to the latest version

**Describe the bug**

1. Environment
   1. Platform: Android
   2. SDK version:  Used via purchases-kmp version `2.10.2+17.55.1`
   3. OS version: Irrelevant
   4. Android Studio version: Irrelevant
   5. How widespread is the issue: apps using AGP < 9 if I understand correctly

R8 recently shipped an analyzer that shows how good your app is shrinked/optimized/obfuscated. Running it on an app that uses RevenueCat leads to getting a scary-looking score of 0% for optimization.

Image

If I exclude the amazon module

```
// in android/app/build.gradle
configurations.configureEach {
  exclude group: "com.revenuecat.purchases", module: "purchases-store-amazon"
}
```

The score is back to normal levels

Image

Running lint on the app also raises a warning about it

> [../../../../../../../../.gradle/caches/9.5.1/transforms/f235dccd418e2ee34b196b2710a84e68/transformed/purchases-store-amazon-9.29.0/proguard.txt](https://github.com/.gradle/caches/9.5.1/transforms/f235dccd418e2ee34b196b2710a84e68/transformed/purchases-store-amazon-9.29.0/proguard.txt): The consumer keep rules at purchases-store-amazon-9.29.0/proguard.txt (from com.revenuecat.purchases:purchases-store-amazon:9.29.0) contains a global option which should not be specified in library consumer rules: -dontoptimize
Libraries often include consumer keep rules to instruct R8 how to optimize the library, especially if the library uses reflection. These keep rules typically indicate to R8 of which classes, methods and fields shouldn't be fully optimized so e.g. reflection continues to work.
>
> Global keep rules can be used by an application to disable significant optimization features, debug R8 behavior, suppress warnings, etc. However, these global options should not be included in library consumer rules -- those distributed with a library.
>
>Starting in Android Gradle Plugin version 9.0, all such rules are not supported in library builds, and ignored by application builds if included in a library. Libraries should not rely on these global options in consumer rules to function, and should remove them.
>
>If you see a flagged library, first try to update to a newer version that doesn't have the embedded global rule. If an updated version without these global options is not available, contact the library vendor to ask about their plans to remove global options, and thus support better R8 configuration.
>
>As an application developer using Android Gradle Plugin 9.0, verify that Android Gradle Plugin removing these global options isn't causing issues in your application. If they are, you can add them temporarily to a local keep rule file.
More info: https://developer.android.com/topic/performance/app-optimization/choose-libraries-wisely
To suppress this error, use the issue id "GlobalOptionInConsumerRules" as explained in the [Suppressing Warnings and Errors](https://github.com/RevenueCat/purchases-android/issues#SuppressInfo) section.
>
>GlobalOptionInConsumerRules Correctness Warning Priority 2/10

I used r8 `9.3.7-dev`, which includes [global options](https://r8-review.googlesource.com/c/r8/+/123801) in its report. 
According to the lint warning, this shouldn't be a problem for apps built with AGP 9+, as those flags are now ignored.
If I understand correctly, it still means that apps using AGP < 9 aren't getting optimized.

**Additional context**

Image

https://github.com/RevenueCat/purchases-android/blob/0857296bba2c2a869d23e6faffec50b768ef83f4/feature/amazon/consumer-rules.pro#L4

https://developer.android.com/topic/performance/app-optimization/r8-configuration-analyzer

```
// <rootDir>/build.gradle
// to override r8

buildscript {
  repositories {
    maven { url = uri("https://storage.googleapis.com/r8-releases/raw") }
  }

  dependencies {
    // https://r8.googlesource.com/r8/+/refs/heads/main/README.md#replacing-r8-in-agp
    classpath(com.android.tools:r8:9.3.7-dev)
  }
}
```

tsushanth:
## Why

`feature/amazon/consumer-rules.pro` contains a `-dontoptimize` directive that disables R8/ProGuard optimization globally for any app that depends on the Amazon billing feature. This causes the entire consuming app to skip optimization, not just the Amazon SDK classes — silently inflating APK size and degrading runtime performance for all customers using the Amazon appstore variant.

The rule was almost certainly a defensive copy-paste from Amazon's own SDK documentation, which advises it for their internal SDK. It is not needed for the RevenueCat wrapper and has no documented justification in the file.

## What changes

Remove the `-dontoptimize` line from `feature/amazon/consumer-rules.pro`. The remaining rules (`-dontwarn`, `-keep class com.amazon.**`, `-keepattributes *Annotation*`) are unaffected and continue to prevent stripping of Amazon SDK reflection targets.

## Validation

R8 optimization is now re-enabled for the Amazon feature module's consumers. The existing `-keep class com.amazon.** {*;}` rule still prevents any Amazon SDK class from being removed or renamed, so there is no regression risk for Amazon IAP calls. No logic in the RevenueCat codebase relies on optimization being disabled.

Fixes #3496

---

> [!NOTE]
> **Low Risk**
> Single-line ProGuard consumer rule change with Amazon SDK classes still fully kept; no runtime logic changes.
> 
> **Overview**
> Removes **`-dontoptimize`** from `feature/amazon/consumer-rules.pro`, so apps that depend on the Amazon billing feature no longer inherit a rule that turned off R8/ProGuard optimization for the **whole** app.
> 
> **`-dontwarn`**, **`-keep class com.amazon.**`**, and **`-keepattributes *Annotation*`** are unchanged and still protect Amazon SDK classes from stripping/renaming.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7b22faf5fe5532695a01d5cda0123113b5e8978f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/feature/amazon/consumer-rules.pro
+++ b/feature/amazon/consumer-rules.pro
@@ -1,4 +1,3 @@
 -dontwarn com.amazon.**
 -keep class com.amazon.** {*;}
 -keepattributes *Annotation*
--dontoptimize


tonidero:
Hi @tsushanth, thanks for your contribution! While we agree this is far less than ideal, we need to make sure this is valid since Amazon actually recommends adding this to the proguard files: https://developer.amazon.com/docs/in-app-purchasing/iap-obfuscate-the-code.html
image

In any case, we agree this is not ideal specially in hybrid SDKs where using amazon is optional and we should split up the amazon module so it only affects folks using it.

tsushanth:
Thanks for the context, @tonidero! I understand — Amazon does officially recommend `-dontoptimize` in their IAP obfuscation docs for their SDK internals, so removing it outright without validation isn't straightforward.

The split-module approach sounds like the right long-term fix: scoping the rule to only consumers who actually pull in the Amazon IAP integration avoids the blanket performance penalty for everyone else while still respecting Amazon's recommendation for those who need it. Happy to close this PR and wait for that refactor if that's the direction you're taking, or help with any part of it if useful. Let me know how you'd like to proceed.

AlvaroBrey:
Hey @tsushanth , we've verified this internally and it seems like a safe enough change which will benefit consumers.

Per our external contribution guidelines, I will merge this PR into an intermediate branch so that we can run CI on it (while of course still crediting you). Thank you very much!

AlvaroBrey (approved):

PR #3657, opened 2026-06-25
feat(ads): support multi grant rewards

peterporfy:
### Checklist

### Motivation

A rewarded ad can grant multiple entitlements and currencies - this change adapts the SDK to that.

### Description

A new field called `more_rewards` (tail) appear next to `reward` (head). It is a list of rewards (same object shape) as `reward` (but excluding `reward`). It contains elements in case the rewarded ad resulted in more than 1 grants.

We also reload the customer info in case we detect that entitlements have been granted.

### Other notes

**This PR is stacked on ads-293 - need to merge that first**

---

> [!NOTE]
> **Medium Risk**
> Changes preview reward-verification behavior: verified polls can surface as failed when CustomerInfo refresh fails after entitlement grants, affecting apps that gate UX on poll success.
> 
> **Overview**
> Reward verification now models **multiple grants** from a single ad: the API’s `more_rewards` list is parsed and exposed on `RewardVerificationResult` as **`moreRewards`** alongside the primary **`verifiedReward`**. **`VerifiedReward.Entitlement`** (identifier + `expiresAt`) is added end-to-end from the network response through the poller to the public API.
> 
> After a successful poll, **`pollRewardVerification`** applies local side effects across **both** the head and tail rewards: virtual-currency grants still invalidate the virtual currencies cache (including VC only in `moreRewards`), and any entitlement grant triggers a **`CustomerInfo` refresh** with up to three attempts and 300ms backoff on network errors. If an entitlement was granted but refresh cannot complete, the callback gets **`RewardVerificationResult.failed`** even though verification succeeded server-side (the grant remains on the backend).
> 
> The AdMob sample maps entitlement verification to dedicated UI copy. Poll failure log strings are inlined in `Outcome` (replacing `RewardVerificationStrings`); poller tests now assert log severity only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e2e85886b2e62168400da773a29a8fdb54dc2a0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


peterporfy:
This change is part of the following stack:

- #3657 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

examples/admob-sample/src/main/java/com/revenuecat/sample/admob/ui/ads/verification/RewardedVerificationMessage.kt
```diff
@@ -42,9 +42,14 @@ internal data class RewardedVerificationMessage(
...
                     RewardedVerificationMessage(
-                        "✅ Verified\n🎁 Reward granted: ${reward.amount} ${reward.code}",
+                        "✅ Verified\n🎁 Currency granted: ${reward.amount} ${reward.code}",
                         Severity.SUCCESS,
                     )
+                is VerifiedReward.Entitlement ->
+                    RewardedVerificationMessage(
+                        text = "✅ Verified\n🎁 Entitlement granted: ${reward.identifier}",
+                        Severity.SUCCESS
+                    )
```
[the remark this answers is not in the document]
peterporfy: That's fine and deliberate: it is a sample app to show the main loop

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -814,10 +816,35 @@ public class Purchases internal constructor(
...
+
+        if (rewards.any { it is VerifiedReward.VirtualCurrency }) {
             purchasesOrchestrator.invalidateVirtualCurrenciesCache()
         }
-        return result
+        // Don't surface an entitlement reward we couldn't reflect locally; the grant persists
+        // server-side and syncs on a later fetch.
+        val entitlementReflected = rewards.none { it is VerifiedReward.Entitlement } ||
+            refreshCustomerInfoAfterEntitlementGrant(clientTransactionId)
+        return if (entitlementReflected) result else RewardVerificationResult.failed
```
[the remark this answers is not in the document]
peterporfy: That is by design: the verification result's reward collection is informational. If we fail to reflect the entitlements then we won't let the app know about the changes because it can cause wrong UX: the dev might show something was granted but it is not available due to failed fetch. The grant happens anyway, so at one point the user will see it, but presenting failure here is the honest behavior.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -814,10 +816,35 @@ public class Purchases internal constructor(
...
+                awaitCustomerInfo(fetchPolicy = CacheFetchPolicy.FETCH_CURRENT)
+                return true
+            } catch (e: PurchasesException) {
+                // Retry only transient (network) failures; a terminal error won't improve on retry.
+                if (e.code != PurchasesErrorCode.NetworkError) break
+            }
+            attempt++
+        }
+        warnLog { RewardVerificationStrings.entitlementCustomerInfoRefreshFailed(clientTransactionId) }
+        return false
```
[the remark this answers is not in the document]
peterporfy: fixed

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/Outcome.kt
```diff
@@ -7,7 +7,7 @@ import com.revenuecat.purchases.InternalRevenueCatAPI
 // capture *why* polling ended so the poller can log an actionable reason without adding public cases.
 @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
 internal sealed interface Outcome {
-    class Verified(val reward: VerifiedReward) : Outcome
+    class Verified(val reward: VerifiedReward, val moreRewards: List<VerifiedReward>) : Outcome
```
tonidero, line 10: Hmm is this concept something like "a main reward, then other smaller ones"? I'm just wondering if it would be better to just have a list of rewards, instead of 1 then a list. Don't have all the context, so lmk if this is intended :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationStrings.kt
```diff
@@ -26,4 +26,13 @@ internal object RewardVerificationStrings {
     fun terminalError(error: String): String =
         "Reward verification stopped after an unrecoverable error: $error. This is unexpected; if it " +
             "persists, contact RevenueCat support with the error above."
+
+    fun entitlementFetchingCustomerInfo(clientTransactionId: String): String =
```
tonidero: I know this is more like how it works in iOS, but personally, I'm not really a fan of extracting log strings outside like this (unless they are reused, though normally they shouldn't). It just adds an extra step when searching for strings, and we are not planning to localize these dev-oriented strings

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/VerifiedReward.kt
```diff
@@ -18,6 +19,15 @@ public interface VerifiedReward {
         public val amount: Int,
     ) : VerifiedReward
 
+    /**
+     * A temporary entitlement grant with an identifier and expiration.
+     */
+    @Poko
+    public class Entitlement(
+        public val identifier: String,
+        public val expiresAt: Date,
```
tonidero, line 28: Should this be nullable? (Just wondering if we will ever want to grant lifetime entitlements)

tonidero: Unrelated question but I'm wondering, how does this translate to the CustomerInfo? I'm thinking that it will be tricky, since many of the fields in [EntitlementInfo](https://github.com/RevenueCat/purchases-android/blob/9bdbb5b3d07693d18f85bd1ddaa70fdf14c4433c/purchases/src/main/kotlin/com/revenuecat/purchases/EntitlementInfo.kt#L56) do not really match well for entitlements obtained through ads I think, like `store` or `productIdentifier` (You might have a plan for this mapping already, just making sure :) )

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -814,10 +817,37 @@ public class Purchases internal constructor(
         poll: suspend (String) -> RewardVerificationResult,
     ): RewardVerificationResult {
         val result = poll(clientTransactionId)
-        if (result.verifiedReward is VerifiedReward.VirtualCurrency) {
+        val rewards = result.verifiedReward?.let { listOf(it) + result.moreRewards } ?: return result
```
tonidero, line 820: TBH, I wonder if we should move most of the logic here to the `PurchasesOrchestrator` and just keep this as a very thin wrapper, though I think that would mean, we would need to move most classes from the `defaults` source set to `main` so probably not for this PR.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RewardVerificationResponse.kt
```diff
@@ -66,4 +73,23 @@ internal data class RewardVerificationResponse(
...
+
+    @OptIn(InternalRevenueCatAPI::class)
+    private fun JsonObject.toEntitlementRewardOrUnsupported(): VerifiedReward {
+        val identifier = this["identifier"]?.jsonPrimitive?.contentOrNull
+        // `expires_at` is an ISO-8601 string (e.g. "2026-06-16T12:00:00Z"), not epoch seconds.
+        val expiresAtRaw = this["expires_at"]?.jsonPrimitive?.contentOrNull
+
+        if (identifier.isNullOrBlank() || expiresAtRaw.isNullOrBlank()) {
+            warnLog { "Malformed entitlement reward payload: $this" }
+            return VerifiedReward.UnsupportedReward
```
tonidero, line 85: I guess this is related to a previous question I did but I guess that means we won't support lifetime entitlements?

tonidero (commented):
Mostly some questions around the architecture. But pretty close. Thank you!!

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/Outcome.kt
ajpallares: I also had [the exact same doubt](https://github.com/RevenueCat/purchases-ios/pull/7039#discussion_r3498188279) when reviewing the iOS counterpart

peterporfy: @tonidero @ajpallares This is intentional and it makes it 1. backward compatible 2. covers 99% of the cases with a more convenient API.

This was something suggested a few weeks ago by @JayShortway and we adopted the idea as it fits into our use case nicely. It is more like a `head+tail` situation than `main` and `smaller ones`. Most of the users will have a single `head` reward, but in case you configure more than one you can access them in `moreRewards`.

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/VerifiedReward.kt
peterporfy: It shouldn't be nullable. Expiration can be configured up to 50years which is virtually lifetime. The expiration is fundamentally needed (to be not null) for optimal backend processing.

Regarding `store` and `productIdentifier`: those are going to be synthetic.

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/RewardVerificationStrings.kt
peterporfy: Inlined them here: https://github.com/RevenueCat/purchases-android/pull/3657/changes/0e016df65d418c83faa838310837003a7eb027ea

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RewardVerificationResponse.kt
peterporfy: It is about future compatibility - in case we would change our mind on expiresAt or if we would add different grant types.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
peterporfy: I will check this separately - I see this is the same topic as what you raised in the previous PR. This might be a bigger change and I am not fully sure about how to go about it, as we don't want this code to show up in the `customEntitlementComputation` flavor - but I am also not super familiar with the Android project so I suggest to discuss it on slack :)

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/Outcome.kt
tonidero: Right... on the other hand, devs might be incentivized to check for the additional rewards even if they don't have any just in case... But well, It's true that it's backwards compatible (though it's not a huge deal considering it's experimental APIs) but ok let's go with this for now 👍

tonidero (approved):
Looks good!!

purchases/src/defaults/kotlin/com/revenuecat/purchases/ads/rewardverification/Outcome.kt
ajpallares: Right. It being backwards compatible shouldn't be a concern since it's experimental as Toni says. But I do see the point of having this. In the end it's about the `head+tail` situation you mention, and I can't find a simpler way to do that without overcomplicating the API with a new type...

PR #3666, opened 2026-06-26
[WIP] [POC] Use config endpoint for workflows

vegaro:


purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsTopicHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+        val refsToPrefetch = items.values.filter { it.prefetch }.mapNotNull { it.blobRef }
+        if (refsToPrefetch.isEmpty()) return
+        coroutineScope {
+            refsToPrefetch.map { ref -> async { blobFetcher.ensureDownloaded(ref) } }.awaitAll()
+        }
+    }
+
+    override fun clear() {
+        // No handler-owned state: the manager prunes this topic's persisted metadata when it leaves the
+        // active set, and blob retention drops any body it kept alive.
```
tonidero: We will need to support clearing I think, so we clear blobs upon a user change (since some of this will be user-dependent (at least that was my thinking)

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+ * through `topic()` / `body()`. It never sees the blob store, the fetcher, or the disk cache.
+ */
+internal class WorkflowsConfigProvider(
+    private val manager: RemoteConfigManager,
+) {
+
+    fun workflowIdForOfferingId(offeringId: String): String? =
+        manager.topic(TOPIC_NAME)
+            ?.entries
+            ?.firstOrNull { (_, item) -> item.content.stringOrNull(KEY_OFFERING_IDENTIFIER) == offeringId }
```
tonidero: Hmm I wonder if we should optimize this somehow... Maybe we could keep in memory a map of offering id to workflow id, so we could then access by id directly... but maybe I'm eagerly optimizing it... we should test with apps with a ton of offerings and workflows later on to make sure this works fine 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -366,32 +366,27 @@ internal class PurchasesFactory(
...
             )
 
-            val workflowManager = workflowsCache?.let {
-                WorkflowManager(
+            // Workflows are now served from the `/v1/config` layer. WorkflowManager is the unchanged consumer
+            // seam; behind it sit the RemoteConfig stack (sync + blob store + on-demand fetch) and the
+            // WorkflowsConfigProvider. The manager is hoisted so the orchestrator can drive its lifecycle
+            // (foreground refresh + teardown).
+            var remoteConfigManager: RemoteConfigManager? = null
+            val workflowManager = if (appConfig.useWorkflows) {
```
tonidero: We might want to rename the feature flag to something more generic, since it will mean more than just workflows I guess... But not a blocker.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -405,7 +400,13 @@ internal class PurchasesFactory(
...
                 offeringFontPreDownloader = offeringFontPreDownloader,
                 uiPreviewMode = appConfig.uiPreviewMode,
-                workflowManager = workflowManager,
+                // Reinstates "wait for workflows before returning offerings", now gated on the `workflows`
+                // config topic instead of a per-offerings workflows-list fetch.
+                offeringsConfigGate = remoteConfigManager?.let { manager ->
+                    OfferingsConfigGate { appInBackground, appUserID, onReady ->
+                        manager.awaitTopicsReady(listOf("workflows"), appInBackground, appUserID, onReady)
+                    }
+                },
```
tonidero, line 412: Interesting mehanism... I believe it should work 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -127,13 +151,102 @@ internal class RemoteConfigManager(
...
+     * The bytes of the item [itemKey] within [topicName], or `null` when the item is unknown or its body can be
+     * neither read nor downloaded. Awaits an in-flight sync, then resolves the payload the same way for both
+     * delivery modes: a `blob_ref` item reads from the blob store (downloading on demand, deduped, if absent),
+     * while an inline item returns its `content` bytes. Providers go through here instead of touching the blob
+     * store or fetcher directly.
+     */
+    suspend fun body(topicName: String, itemKey: String): ByteArray? {
+        awaitCurrentSync()
+        val item = topic(topicName)?.get(itemKey) ?: return null
+        val ref = item.blobRef ?: return item.content.toString().toByteArray()
```
tonidero: Hmm I wonder if we want to just return back the item string in case blobRef is null.... I would consider this body to only return in case there is an actual blob associated. For the "metadata" part of the config, that probably needs to be fetched from the RemoteConfigTopicStore?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobFetcher.kt
```diff
@@ -0,0 +1,15 @@
...
+ * collapse into one download.
+ *
+ * Declared as an interface, not implemented here: the dedup + checksum verification + transport (and the
+ * `sources` topic that resolves blob URLs) are a separate phase. The wiring binds a no-op (`{ false }`) for now,
+ * so on-demand fetch is inert and only inline/prefetched blobs resolve. This interface is the seam that work
+ * plugs into.
+ */
+internal fun interface RemoteConfigBlobFetcher {
+    /** Ensures [ref] is cached, returning whether it is now on disk. The default wiring is a no-op (`false`). */
+    suspend fun ensureDownloaded(ref: String): Boolean
```
tonidero: We will also need to add some prioritization system, so when we call this, the corresponding blob gets downloaded ASAP, taking priority of any in progress.

tonidero (commented):
Nice!! Only gave it a quick look but this is pretty much what I was thinking indeed! Still a lot of things left to do, but it's nice to have the skeleton with this 🙌 cc @RevenueCat/coresdk would be great for you to take a look as well to see if you find any issues with this architecture!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
vegaro: right! I find the gate function very elegant haha

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
vegaro: we had that map before and it helped, so definityely an improvement. But that can probably be a followup yeah

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowModels.kt
```diff
@@ -141,7 +141,9 @@ public data class PublishedWorkflow(
     @SerialName("initial_step_id") val initialStepId: String,
     val steps: Map<String, WorkflowStep>,
     val screens: Map<String, WorkflowScreen>,
-    @SerialName("ui_config") val uiConfig: UiConfig,
+    // Not sent by the config-topic path (ui_config is its own topic); defaults rather than failing the whole
+    // workflow, and the reader swaps the resolved UiConfig in.
+    @SerialName("ui_config") val uiConfig: UiConfig = UiConfig(),
```
vegaro: This is not correct. We need to decouple ui_config completely

PR #3667, opened 2026-06-26
feat(remote-config): add RemoteConfigSourceProvider for source failover

ajpallares:
### Checklist

### Motivation
Android counterpart of the iOS `RemoteConfigSourceProvider` (RevenueCat/purchases-ios#7093): the SDK needs an address book that tracks the healthy api/blob source and fails over when one is reported unhealthy.

### Description
- Hands out the current healthy api and blob source, failing over independently per purpose.
- Dedupes sources by url (keeping highest priority, tie-broken by weight) and orders once via `WeightedSourceSelector`.
- Identifies a source by its url so a stale unhealthy report can't advance the provider twice; thread-safe.

---

> [!NOTE]
> **Medium Risk**
> New infrastructure for which remote config endpoints the SDK uses; incorrect failover or dedup logic could affect config fetch reliability, though it is isolated and heavily tested and not yet integrated into the fetch path in this diff.
> 
> **Overview**
> Adds a **thread-safe remote config “address book”** that tracks the current healthy **API** and **blob** endpoints separately and advances through a fallback list when a caller reports a source unhealthy.
> 
> `DefaultRemoteConfigSourceProvider` dedupes sources by URL (highest **priority**, then **weight**), orders them once via existing `WeightedSourceSelector`, and exposes `getCurrent`, `reportUnhealthy`, and `restart` per purpose. Handouts use `RemoteConfigSourceHandle` with an opaque **token** so duplicate, concurrent, or post-`restart` unhealthy reports cannot skip ahead more than once.
> 
> Includes broad unit coverage (dedup, independent API/blob failover, stale reports, restart, concurrency).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b3c95d5149872b6c0a8ae6150a8b532368c7433. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
```diff
@@ -0,0 +1,154 @@
...
+            RemoteConfigSourceHandle.Purpose.API -> api.restart()
+            RemoteConfigSourceHandle.Purpose.BLOB -> blob.restart()
+        }
+    }
+
+    private companion object {
+
+        /**
+         * Collapses duplicate urls to the occurrence with the highest priority (tie-broken by
+         * weight), keeping first-seen order. Done once at init so reads never need to re-dedupe.
```
tonidero, line 90: Hmm TBH, I'm not sure if the backend would ever do this... but if it did, maybe we should leave it be duped?

tonidero (approved):
Looking great! TBH, I'm not sure if we need the deduping logic at all, and that's my only question. Other than that 🚢

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
ajpallares: Right. This was not really about the backend. The main reason for doing the dedup was [this proposal](https://revenuecat.slack.com/archives/C0AK23U9DBN/p1782462894950239) of having the SDKs bundle a safety fallback config that **we would always append**. If we do that, we'll certainly have duplicates...

tonidero: Ohh hmm right... that makes sense then 👍 Thank you for explaining!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
```diff
@@ -0,0 +1,154 @@
...
+) {
+
+    /** What the source is used for: calling the config api or downloading a blob. */
+    enum class Purpose { API, BLOB }
+
+    /** A plain URL or a URL format with placeholders, to be resolved by the caller. */
+    val url: String get() = source.url
+}
+
+internal interface RemoteConfigSourceProviderType {
```
tonidero: Actually I was thinking, the `Type` naming is not really common in android... Instead, the interface would be `RemoteConfigSourceProvider`, and the impl would be `DefaultRemoteConfigSourceProvider` or `RemoteConfigSourceProviderImpl`.

tonidero (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
ajpallares: Ohh good catch! Thanks for pointing this out. Done in https://github.com/RevenueCat/purchases-android/pull/3667/commits/a5e06873fab6807414d16b75e2d6eeb5512ba2f0

PR #3672, opened 2026-06-29
refactor(remote-config): Support per-element compression in RC config format

tonidero:
## Why

The backend is moving `/v1/config` rc-format compression from a transport-level `Content-Encoding` on the whole stream to a **per-element codec** in each element header, so a blob can be compressed once, cached pre-compressed, and served verbatim (enabling high-effort compression off the request path).

## What

Adds SDK support for that format change:
- Reads the per-element codec and verifies the checksum against the **decompressed** bytes, so blob refs (content addresses) stay stable across codec/level changes.
- Advertises `Accept-RC-Element-Encoding: gzip` so the server never picks a codec the SDK can't decode.
- To avoid new dependencies, supports **gzip + uncompressed only**; any other codec degrades gracefully (inline blob skipped, config kept).

---

> [!NOTE]
> **Medium Risk**
> Touches remote config sync and RC container parsing on the network path; failures degrade to skipped blobs or cached config rather than crashing, but gzip-only negotiation could mis-handle server responses if backend sends compressed config without matching signatures (signing still uses on-wire config bytes).
> 
> **Overview**
> Adds **per-element content encoding** for RC Container Format `/v1/config` responses, replacing transport-level compression with a codec id in each element header.
> 
> **Negotiation:** RC format requests now send `Accept-RC-Element-Encoding: gzip` (separate from `Accept-Encoding`) so the server only uses codecs Android can decode.
> 
> **Parsing & integrity:** `RCContainer` reads the codec from the reserved u32 low byte; `element_size` is on-wire length while checksums hash **uncompressed** bytes. New `RCContentEncoding` decodes `NONE` and `GZIP`; brotli/zstd/unknown codecs fail gracefully. `RCElement.decode()` / `matchesChecksum()` support verify-after-decode; unknown header flags and reserved upper bits log warnings but still parse.
> 
> **Remote config:** `RemoteConfigManager` parses config via `config.decode()` and stores inline blobs as uncompressed, content-addressed payloads; undecodable blobs are skipped while config persists.
> 
> Tests cover gzip fixtures, unsupported codecs, and updated production integration/golden expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit da1259e634010a822f778e18acb1b0db9e35eaf8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3672** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3672?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3672?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3648** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3648?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -54,7 +55,7 @@ internal class RemoteConfigManager(
                 // disk cache mid-write and merge against a stale snapshot.
                 try {
                     if (container != null) {
-                        val response = RemoteConfiguration.parse(container.config.data)
+                        val response = RemoteConfiguration.parse(container.config.decode())
```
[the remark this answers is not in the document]
tonidero: These errors are catched later on, in the backend.kt class, where we just log an error for now, which I think it's ok as the behavior for now.

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -520,6 +524,8 @@ internal class HTTPClient(
         return mapOf(
             "Content-Type" to "application/json",
             "Accept" to if (endpoint.expectsRCFormatResponse) RC_FORMAT_ACCEPT else null,
+            // Advertise supported per-element codecs for RC Container Format responses (gzip-only on Android).
+            "Accept-Encoding" to if (endpoint.expectsRCFormatResponse) RC_FORMAT_ACCEPT_ENCODING else null,
```
ajpallares: One thing I'm thinking. We're kind of "abusing" the `Accept-Encoding` header, aren't we? If I'm not mistaken, its standard HTTP meaning is transport-level ("gzip the whole response body"), but we're using it to negotiate the per-element codec *inside* the RC container, which is a different layer.

Any CDN/proxy in between could read it the standard way too and transport-gzip the whole body. Or, in the best case, it could be confusing for us in the future.

Would it be cleaner to use a dedicated custom header for this negotiation (something like `X-RC-Accept-Element-Encoding: gzip`) instead? Not a blocker from my side, but I think that it would help us better understand the difference in the behavior wrt the standard `Accept-Encoding` header

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContentEncoding.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.nio.ByteBuffer
+import java.util.zip.GZIPInputStream
+
+/**
+ * The per-element content-encoding codec carried in an [RCElement]'s header.
+ *
+ * The backend compresses each element body independently and stores the codec id in one of the element
+ * header's reserved bytes. `element_size` is the **on-wire (compressed)** length, while the element checksum
+ * always covers the **uncompressed** bytes, so a blob's content address (its ref) is stable across codec/level
+ * changes: callers must [decode] before verifying the checksum or consuming the content.
```
ajpallares, line 14: Nice and important documentation to have here, tbh 😄 
Otherwise, this would be much more confusing

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Hmm I see your point... I think might be worth asking in Slack 👍

tonidero: I've asked, and will change it if we agree on changing it. For now, this should still be ready to go :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
```diff
@@ -8,29 +8,50 @@ import java.security.MessageDigest
...
+     * Decodes the element and checks its SHA-256 (truncated to the stored [checksum]'s length) against
+     * [checksum]. The backend stores `sha256(uncompressed)` truncated to its leftmost [checksum] bytes
+     * (192 bits). An unsupported codec or corrupt stream makes [decode] throw, which is treated as a
+     * failed verification (returns `false`).
+     */
+    @Suppress("SwallowedException") // An undecodable element simply fails verification.
+    fun isChecksumValid(): Boolean = try {
+        matchesChecksum(decode())
+    } catch (e: RCContainerFormatException) {
+        false
```
ajpallares, line 43: Should we add a log here? Not blocking

ajpallares (approved):
Good job on this. It looks good to me. There's the header decision to make, but we can totally change it in a follow-up PR. I have just one small nit, but FFTI

rickvdl (approved):
This looks great! My only question would be what we already discussed over Slack; regarding permissive handling of flags and reserved bytes we don't expect.

As you suggested we could at least add the logging for this :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
tonidero: We do have a log in the call site of this so didn't add one here. Lmk if you think it would be better to have it here 🙏

rickvdl (approved):
Looks great! :shipit: 

One nit: the PR description appears to be a bit stale, mentioning the old header.

PR #3676, opened 2026-06-29
feat(remote-config): add stable content hash to ConfigTopic

ajpallares:
### Motivation
When the `sources` topic changes, the source provider needs to start its list of API/blob sources again from the top. But it reads topics lazily and has no cheap way to tell whether one specific topic changed. The config manifest changes whenever any topic changes, so leaning on it would make the provider start over even when the sources themselves didn't change.

### Description
- Promote `ConfigTopic` from a `Map` typealias to a `data class` that delegates `Map` (so existing access is unchanged) and exposes a `hash`: a canonical SHA-256 of its items, computed once at parse time.
- The hash is computed locally by the SDK and kept out of the JSON, so the data written to the disk cache stays the same.

---

> [!NOTE]
> **Low Risk**
> Internal remote-config modeling with backward-compatible serialization; no auth or payment paths touched in this diff.
> 
> **Overview**
> **`ConfigTopic`** is no longer a bare `Map` typealias: it is a **`Map`-delegating data class** with a lazy **`contentHash`** (canonical SHA-256 of topic items) so callers can detect when a specific topic’s body changed without comparing full payloads or relying on the opaque manifest.
> 
> Hashing **canonicalizes JSON object key order** (nested objects included) while **preserving array order**; **`blob_ref` / `prefetch`** and inline content all affect the hash. **`ConfigTopicSerializer`** still reads/writes the **same bare item map** on the wire and in disk cache—**`contentHash` is never serialized**.
> 
> Adds **`ConfigTopicTest`** and extends **`RemoteConfigurationTest`** for hash behavior and parse → serialize → parse round-trips.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9fa7a573562f276d004609764964449909a25690. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfiguration.kt
```diff
@@ -87,7 +91,63 @@ internal data class RemoteConfiguration(
...
+ * Promoted from a bare `Map` typealias so consumers can cheaply detect whether a topic changed without
+ * re-reading its contents (e.g. `RemoteConfigSourceProvider` restarting its source list when `sources`
+ * changes). [hash] is computed once at construction (i.e. parse) time over a canonical form of [items], so an
+ * unchanged topic keeps the same hash across syncs. It delegates [Map], so existing `topic["key"]` /
+ * `topic.values` access is unchanged, and [hash] is SDK-derived: it never appears in the persisted/wire JSON.
+ */
+@Serializable(with = ConfigTopicSerializer::class)
+internal data class ConfigTopic(
+    val items: Map<String, RemoteConfiguration.ConfigItem>,
+) : Map<String, RemoteConfiguration.ConfigItem> by items {
```
ajpallares, line 104: This is something the AI came up with. Looks cool, but idk if it's necessary or could be problematic

AlvaroBrey: It's a good way to wrap the map 👍 no notes

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfiguration.kt
```diff
@@ -87,7 +91,58 @@ internal data class RemoteConfiguration(
...
+ * A single topic's items, with a stable content [hash] so consumers can tell whether a topic changed
+ * without re-reading its contents. Delegates [Map], so it can be accessed like the item map it wraps.
+ */
+@Serializable(with = ConfigTopicSerializer::class)
+internal data class ConfigTopic(
+    val items: Map<String, RemoteConfiguration.ConfigItem>,
+) : Map<String, RemoteConfiguration.ConfigItem> by items {
+
+    /** Stable content hash of [items], computed once at construction. */
+    val hash: String = computeHash(items)
```
AlvaroBrey: Two nits:

- I'd call it `contentHash` for clarity
- Maybe this could be `lazy`? Do we always need to have it precomputed?

AlvaroBrey (approved):
Couple nits, but looks good!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfiguration.kt
ajpallares: Great suggestions! Especially the `lazy`(ness), it really doesn't need to be precomputed, and, in fact, many times it won't be computed at all. Thanks! Done in https://github.com/RevenueCat/purchases-android/pull/3676/commits/2ca8b9da573fd53d56f8a13c1f6f5729f6805f25

purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/ConfigTopicTest.kt
```diff
@@ -0,0 +1,99 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class ConfigTopicTest {
+
+    @Test
+    fun `two identical topics have the same hash`() {
```
tonidero, line 12: I think we should add some explicit tests for serializing and deserializing. And probably even some at the `RemoteConfiguration` model level with this new serialization?

tonidero (approved):
Looks good! Just a suggestion for some more tests

purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/ConfigTopicTest.kt
ajpallares: Good idea! I added some in https://github.com/RevenueCat/purchases-android/pull/3676/commits/a1cc91a6a94222c6ee66ea0369b844367f180fbd
Lmk if you had some additional ones in mind!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfiguration.kt
```diff
@@ -87,7 +91,58 @@ internal data class RemoteConfiguration(
...
+ * A single topic's items, with a stable [contentHash] so consumers can tell whether a topic changed
+ * without re-reading its contents. Delegates [Map], so it can be accessed like the item map it wraps.
+ */
+@Serializable(with = ConfigTopicSerializer::class)
+internal data class ConfigTopic(
+    val items: Map<String, RemoteConfiguration.ConfigItem>,
+) : Map<String, RemoteConfiguration.ConfigItem> by items {
+
+    /** Stable content hash of [items], computed lazily on first access. */
+    val contentHash: String by lazy { computeHash(items) }
```
JayShortway, line 107: I'm curious why the built-in `hashCode()` wasn't sufficient here?

tonidero: Hmm we could have used it I think TBH... I believe on a previous iteration we were thinking we had to verify this hash against the hash of the topic passed in the manifest... But I don't think that's necessary now, so this could indeed be simplified 🙏 Great catch!

tonidero: Probably not a huge impact though, since the `sources` topic where this is used should be pretty small always

PR #3677, opened 2026-06-29
feat(remote-config): drive RemoteConfigSourceProvider from the topic store

ajpallares:
### Motivation
The source provider must reflect the latest `sources` topic and restart its failover lists when that topic changes, while leaving progress intact when it doesn't. Reading lazily from the config store.

### Description
- `DefaultRemoteConfigSourceProvider` now takes a `RemoteConfigTopicStore` instead of pre-parsed lists; it parses the `sources` topic (api uses `url`, blob uses `url_format`) and rebuilds only when `ConfigTopic.contentHash` changes.
- When the sources change, the provider builds fresh failover lists and discards any late "this source failed" reports left over from the old lists, so they can't skip a source in the new one.

---

> [!NOTE]
> **Medium Risk**
> Changes which config API/blob URLs the SDK uses and when failover state resets; incorrect parsing or hash handling could misroute fetches or drop failover progress.
> 
> **Overview**
> **`DefaultRemoteConfigSourceProvider`** no longer takes static api/blob source lists; it takes a **`RemoteConfigTopicStore`** and lazily reads the **`sources`** topic, parsing api entries from **`url`** and blob from **`url_format`**. Failover lists are rebuilt only when **`ConfigTopic.contentHash`** changes—unchanged topics keep failover position; a changed topic restarts both api and blob from the top.
> 
> Until a usable **`sources`** topic exists, **API** uses embedded defaults (`api.revenuecat.com` then `api.rc-backup.com`); **blob** stays empty with no embedded fallback. On rebuild, handle **tokens** are bumped so late **`reportUnhealthy`** calls from the previous generation are ignored (including skipping **`restart`**/**`reportUnhealthy`** when a rebuild already happened). **`SourceFailover`** locking moves to the provider’s single lock.
> 
> Adds **`RemoteConfigTopicStore`** and tests for defaults, topic hash behavior, and stale reports across topic changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2194d0608a4e560d62ce84baa17fc158e8a6e97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
```diff
@@ -50,44 +54,109 @@ internal interface RemoteConfigSourceProvider {
...
 
-    private val api = SourceFailover(RemoteConfigSourceHandle.Purpose.API, dedupe(apiSources), random)
-    private val blob = SourceFailover(RemoteConfigSourceHandle.Purpose.BLOB, dedupe(blobSources), random)
+    private val lock = Any()
+
+    // Content hash of the `sources` topic the current failovers were built from. Null means there is no
+    // sources topic (absent, or none seen yet).
+    private var builtHash: String? = null
+    private var api = SourceFailover(RemoteConfigSourceHandle.Purpose.API, emptyList(), random)
+    private var blob = SourceFailover(RemoteConfigSourceHandle.Purpose.BLOB, emptyList(), random)
```
tonidero: Should we start it with some hardcoded versions of the sources? I think we can use `https://api.revenuecat.com` and `https://config.revenuecat-static.com/{blob_ref}` (until we have final versions that we can embed into the SDK. Alternatively we could also store them on disk (in an embedded json), but I guess that requires an extra disk access... Probably NABD...

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigTopicStore.kt
```diff
@@ -0,0 +1,7 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+/** Read-only access to a topic's persisted item index (metadata only — no blob bytes, no waiting). */
```
tonidero, line 3: Note that this is still doing a disk access (I imagine) so we should make sure we call this on background threads only (mentioning since the `no waiting` was a bit confusing)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigTopicStore.kt
```diff
@@ -0,0 +1,7 @@
...
+internal fun interface RemoteConfigTopicStore {
```
tonidero, line 4: We will need to hook this up to the RemoteConfigDiskCache, but that can be in a separate PR :)

tonidero (approved):
I think this makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
ajpallares: I remember we discussed the version-on-disk. However, I think it's a great idea to start with some hardcoded values indeed 👍
Great suggestion!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigTopicStore.kt
ajpallares: Right. The `no waiting` here means "not waiting for the network".

Now, is it doing disk access really? Will we do a disk access every time we query for a topic? Or will there be some kind of in-memory caching for this?

tonidero: It could indeed be good to keep some values in memory yeah, so hopefully it would only be the first time, which should happen shortly after configuring the SDK normally.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
```diff
@@ -50,44 +54,140 @@ internal interface RemoteConfigSourceProvider {
...
+         * the topic is absent.
+         */
+        fun sourcesFor(
+            topic: ConfigTopic?,
+            purpose: RemoteConfigSourceHandle.Purpose,
+        ): List<RemoteConfigSource> = when (purpose) {
+            RemoteConfigSourceHandle.Purpose.API ->
+                if (topic == null) DEFAULT_API_SOURCES else parseSources(topic, API_ITEM, URL_KEY)
+            RemoteConfigSourceHandle.Purpose.BLOB ->
+                if (topic == null) DEFAULT_BLOB_SOURCES else parseSources(topic, BLOB_ITEM, URL_FORMAT_KEY)
```
ajpallares: Probably not a big deal, but I'm on the fence on what we should do here:
1. Only use the hardcoded urls when the Sources topic is not available
2. Always merge the hardcoded sources with the Sources from the topic
WDYT @RevenueCat/coresdk?

ajpallares: (note that we already have a dedup mechanism) https://github.com/RevenueCat/purchases-android/blob/9bdbb5b3d07693d18f85bd1ddaa70fdf14c4433c/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt#L92

ajpallares: I'm leaning towards keeping it this way. If we want the hardcoded urls to be used, we should assume those will be included in the Sources topic of the Config endpoint. In addition, the dedup would also emit a log in case of duplication, which would be weird. Therefore, I'm leaving it like this: as a fallback for when we don't have a Sources topic yet

Again, we can always iterate on this, and we probably will, since we'll likely bundle a local Config file. We'll need to rethink the deduplication logic (or even if it's needed at all).

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigTopicStore.kt
ajpallares: Exactly 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigSourceProvider.kt
ajpallares: Done in https://github.com/RevenueCat/purchases-android/pull/3677/commits/1353b4aa490eb604615c5e1ef46d1b17dee89170

tonidero (approved):
I think this makes sense for now indeed, and it's probably a bit simpler, so happy to go with this!

tonidero (approved):
Looks great!

PR #3680, opened 2026-06-29
First pass at local encrypted storage

davedelong:
### Checklist

### Motivation

Local encrypted storage of sensitive information. This uses a derived symmetric key to encrypt blobs before writing them out to disk. It has a functionally similar API to the corresponding changes on iOS.

---

> [!NOTE]
> **Medium Risk**
> New crypto and sensitive-data persistence layer; security depends on password handling and default salt, though it is internal-only and not integrated yet.
> 
> **Overview**
> Introduces **on-device encrypted storage** for sensitive blobs, aligned with the iOS secure-storage API shape.
> 
> Adds an internal **`SecureItemStorage`** contract (read/save/delete by identifier, optional **`modifyItem`**, **`SecureItemAttributes`**, **`SecureStorageException`**) and **`EncryptedItemStorage`**, which derives an AES-256-GCM key via PBKDF2 from a caller-supplied password, encrypts payloads with the item id as GCM associated data, and persists Base64 ciphertexts in two JSON stores. **`includedInBackup`** routes items to **`filesDir`** (Auto Backup) vs **`noBackupFilesDir`** (excluded), including moves when attributes change. Initialization uses **`suspend create()`** with coroutine dispatchers for key derivation and file load; writes use **`AtomicFile`** and disk-before-memory updates under synchronization.
> 
> Robolectric unit tests cover the interface defaults and **`EncryptedItemStorage`** routing, crypto round-trips, and partition behavior via an internal test constructor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b714316891e3e9a9004c06969e234ddec522d80a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,237 @@
...
+ * A [SecureItemStorage] implementation that derives an AES-256-GCM encryption key from a
+ * password using PBKDF2, then stores ciphertexts as JSON files on the device's internal storage.
+ *
+ * ## Key derivation
+ *
+ * The symmetric key is derived once at initialization via `PBKDF2WithHmacSHA256` using the
+ * supplied [password] and an optional salt taken from [AccessGroup.accessGroup]. Because the
+ * key is derived deterministically, it survives backup/restore — as long as the same password
+ * is provided, the data is readable on any device, in contrast to Android Keystore-backed
+ * approaches where keys are hardware-bound and cannot be restored.
```
tonidero: Hmm TBH, I wondering how important is to have the backups support (since otherwise, we could just use `EncryptedSharedPreferences`)... But I guess we want to try to keep people authenticated across installations... So let's go with this!

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,237 @@
...
+ * different identifier, even with the same key.
+ *
+ * ## Cross-app sharing
+ *
+ * The [AccessGroup] class exists for API parity with the iOS SDK. It provides a file-name
+ * namespace that prevents collisions when multiple apps from the same developer use
+ * RevenueCat. It does **not** enable cross-app item sharing on Android; each app's internal
+ * storage directory is OS-sandboxed.
+ */
+internal class EncryptedItemStorage @VisibleForTesting internal constructor(
```
tonidero: I don't think we need the `VisibleForTesting` annotation for internal classes? Tests should be able to instantiate this class I think?

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,237 @@
...
+     *
+     * Mirrors the `Keychain.AccessGroup` concept from the iOS SDK. On Android, the
+     * [accessGroup] value is used as the PBKDF2 salt, so two different access groups produce
+     * two different encryption keys from the same password. The [appIdentifier] is used to
+     * scope the storage file names within the group.
+     *
+     * @param accessGroup The logical group identifier (analogous to a keychain access group
+     *        entitlement on iOS). Used as the PBKDF2 salt on Android.
+     * @param appIdentifier An identifier for this specific app, used to scope storage file
+     *        names so that different apps within the same group do not share items.
```
tonidero: As you said above, each app has its own storage and is not shared in Android. So just wondering, even if it's to keep parity with iOS, is it needed to keep the `AccessGroup` class at all? Or should we just keep a "salt" instead?

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,237 @@
...
+    internal data class AccessGroup(
+        val accessGroup: String,
+        val appIdentifier: String,
+    )
+
+    companion object {
+        private const val PBKDF2_ALGORITHM = "PBKDF2WithHmacSHA256"
+
+        // 100 000 iterations meets current NIST SP 800-132 guidance. This runs once at
+        // initialization, not on every read/write, so the one-time cost is acceptable.
```
tonidero, line 65: Hmm initialization is actually an intensive time where we do many things... I imagine this might run on a background thread, but we might want to run some tests to see how long this may last (ideally in older devices)

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,237 @@
...
+
+            return EncryptedItemStorage(
+                backupFile = File(context.filesDir, "${baseName}_backup.json"),
+                noBackupFile = File(context.noBackupFilesDir, "${baseName}_no_backup.json"),
+                key = key,
+            )
+        }
+    }
+
+    // Both stores are loaded once at construction time and kept in memory.
```
tonidero: Just a note that this is fine for small pieces of data, we should just make sure we don't store anything too big here to not increase memory usage. Not a problem right now, just wanted to mention it :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,237 @@
...
+    // Reads are served entirely from these maps; writes update the map then flush the
+    // affected file to disk. loadStore() is private and only called here.
+    private val backupStore: MutableMap<String, String> = loadStore(backupFile)
+    private val noBackupStore: MutableMap<String, String> = loadStore(noBackupFile)
```
tonidero: Hmm these are reading from disk on the main thread which is not ideal... Should we try to make things async so we can do the loading asynchronously?

tonidero (commented):
Still wrapping my head around all the crypto stuff, so posting a first round of comments today 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
davedelong: yeah, my first implementation used `EncryptedSharedPreferences`, but i wanted to have a path where a user restores their device from a backup and the tokens are still present, so I switched to this file-based approach

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
davedelong: good catch, i'll remove that.

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
davedelong: I'll change this to be async and make the key derivation be async as well

davedelong:
> Still wrapping my head around all the crypto stuff

@tonidero the gist of the crypto stuff is:
- i want to store encrypted blobs 
- i can't use the OS's KeyStore for saving an encryption key, because it's not included in device backups
- i want a symmetric key, so that the same key is used for both encryption and decryption
- i want a key that can be _derived_, so that i can re-make it when the app relaunches after device restoration
- PBKDF2 is a recommend way to do key derivation when you have low-entropy input/salts

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
davedelong: yeah, fair point. simplicity in this case likely beats out interface parity, since this is internal-only

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+import javax.crypto.spec.SecretKeySpec
+
+/**
+ * A [SecureItemStorage] implementation that derives an AES-256-GCM encryption key from a
+ * password using PBKDF2, then stores ciphertexts as JSON files on the device's internal storage.
+ *
+ * ## Key derivation
+ *
+ * The symmetric key is derived once at initialization via `PBKDF2WithHmacSHA256` using the
+ * supplied [password] and an optional salt taken from [AccessGroup.accessGroup]. Because the
```
AlvaroBrey: Outdated ref to `AccessGroup`

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+
+    // endregion
+
+    // region File I/O
+
+    private fun saveStore(file: File, store: Map<String, String>) {
+        val json = JSONObject()
+        store.forEach { (k, v) -> json.put(k, v) }
+        file.parentFile?.mkdirs()
+        // Write to a temp file then rename for atomicity.
```
AlvaroBrey: You can probably just use `AtomicFile` here instead

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+ * approaches where keys are hardware-bound and cannot be restored.
+ *
+ * ## Backup behaviour
+ *
+ * The [SecureItemAttributes.includedInBackup] attribute controls which directory a file is
+ * written to:
+ *
+ * - `true` (the default): the item is stored under [Context.getFilesDir], which participates
+ *   in Android Auto Backup.
+ * - `false`: the item is stored under [Context.getNoBackupFilesDir], which is explicitly
```
AlvaroBrey, line 40: This says we don't need to include the no_backup in the backup exclusions, but the docs in `SecureItemAttributes` contradict it

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    override fun readItem(identifier: String): ByteArray? {
+        // Grab the encoded ciphertext under the lock, then decrypt outside it so we don't
+        // hold the monitor during a potentially slow crypto operation.
+        val encoded = synchronized(this) {
+            backupStore[identifier] ?: noBackupStore[identifier]
+        } ?: return null
+        return try {
+            decrypt(Base64.decode(encoded, Base64.DEFAULT), identifier)
```
AlvaroBrey: Likely should use Base64.NO_WRAP to avoid newline wrapping (in the encoding too)

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+         * @param salt the PBKDF2 salt; defaults to [DEFAULT_SALT]
+         * @throws GeneralSecurityException if key derivation fails
+         */
+        @Throws(GeneralSecurityException::class)
+        suspend fun create(
+            context: Context,
+            password: CharArray,
+            salt: String = DEFAULT_SALT,
+        ): EncryptedItemStorage {
+            val key = withContext(Dispatchers.Default) {
```
AlvaroBrey: It would be best to inject dispatchers through constructor rather than hardcoding them, especially if we end up using them at more spots.

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    private fun decrypt(data: ByteArray, identifier: String): ByteArray {
+        val iv = data.copyOfRange(0, GCM_IV_LENGTH)
+        val ciphertext = data.copyOfRange(GCM_IV_LENGTH, data.size)
+        val cipher = Cipher.getInstance(CIPHER_TRANSFORMATION)
+        cipher.init(Cipher.DECRYPT_MODE, key, GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv))
+        cipher.updateAAD(identifier.toByteArray(Charsets.UTF_8))
+        return cipher.doFinal(ciphertext)
+    }
```
AlvaroBrey, line 221: Consider whether we want to provide an async (`suspend`) version of the public accessors, since encrypt/decrypt may be heavy operations that should be offloaded from the main thread.

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
```diff
@@ -0,0 +1,228 @@
...
+        // reuse it without duplicating the parsing logic.
+        internal fun loadStore(file: File): MutableMap<String, String> {
+            if (!file.exists()) return mutableMapOf()
+            return try {
+                val json = JSONObject(file.readText())
+                val map = mutableMapOf<String, String>()
+                for (k in json.keys()) {
+                    map[k] = json.getString(k)
+                }
+                map
```
AlvaroBrey: We have `JSONObject.toMap()`  extension somewhere

AlvaroBrey (commented):
Generally very nice, left some suggestions

Also take a look at the exception comment by Cursor above (I 👍  it)

purchases/src/main/kotlin/com/revenuecat/purchases/common/security/EncryptedItemStorage.kt
davedelong: As algorithms go, this one isn't heavy-weight. It's O(N) with respect to the length of the thing being processed. The heavier stuff would be the 100,000 iterations that happen during key derivation, which is already async.

AlvaroBrey: Fair enough!

AlvaroBrey (approved):
Looks great!

tonidero (approved):
This makes sense to me. Great job!!

PR #3694, opened 2026-07-02
Fix backend integration tests

rickvdl:
Fixes failing backend integration tests with a couple of different causes:
- The remote config endpoint now inlines items, we previously asserted on an empty list of inlined items
- The remote config (subsequent) requests include the manifest token, which is isn't stable, causing the golden snapshot request files to fail because of difference every time (added in follow up [PR](https://github.com/RevenueCat/purchases-android/pull/3695))
- Updated golden snapshot files (will be merged into this PR from follow up [PR](https://github.com/RevenueCat/purchases-android/pull/3695))

---

> [!NOTE]
> **Low Risk**
> Test-only and golden-file updates; no production SDK behavior changes.
> 
> **Overview**
> Aligns **backend integration tests** with current remote-config API behavior and stabilizes golden HTTP recordings.
> 
> **Remote config assertions** in `ProductionRemoteConfigIntegrationTest` now expect inlined config (drops the empty `contentElements` check) and assert the config blob passes **`isChecksumValid()`**.
> 
> **Golden recording** adds **`manifest`** to `GoldenFileRecorder`’s ignored request-body fields so follow-up `/v1/config` calls with a session manifest compare as `TEST_STATIC_VALUE` instead of failing on every run.
> 
> **Golden fixtures** add request/response snapshots for all `ProductionRemoteConfigIntegrationTest` cases (including manifest replay → **204**), and bump offerings paywall **`revision`** from 5 to 6 in existing offerings golden responses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1dfbb83cd67418228c3e73b4df22801ec6588967. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero (approved):
Ohh right. Thanks for fixing!!

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/GoldenFileRecorder.kt
```diff
@@ -141,6 +144,10 @@ class GoldenFileRecorder(
             "events > [offering_id]",
             "events > [timestamp]",
         )
+
+        val DEFAULT_REQUEST_BODY_FIELDS_TO_IGNORE_BY_PATH = mapOf(
```
tonidero: Would it make sense to use the `DEFAULT_BODY_FIELDS_TO_IGNORE` instead of this? I guess it means it will be ignored in all requests... But maybe not a huge deal?

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/GoldenFileRecorderTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.json.JSONObject
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.io.File
+
+@RunWith(RobolectricTestRunner::class)
+internal class GoldenFileRecorderTest {
```
tonidero: Testing the test utils 😅

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionRemoteConfigIntegrationTest.kt
```diff
@@ -25,12 +25,12 @@ internal class ProductionRemoteConfigIntegrationTest : BaseBackendIntegrationTes
         assertThat(error).isNull()
         assertThat(verification).isEqualTo(VerificationResult.VERIFIED)
         val rcContainer = requireNotNull(container) { "Expected a 200 container, got 204 (no content)." }
+        assertThat(rcContainer.config.isChecksumValid()).isTrue()
```
tonidero, line 28: ❤️

tonidero (approved):
I do wonde if we could use the already existing mechanism instead of adding this... But I think we can always iterate on it if needed.

PR #3682, opened 2026-06-30
refactor(remote-config): Resolve remote config sources lazily from the committed `sources` topic

tonidero:
## Description

Part of Phase 5 of the topic-based remote-config feature (`/v1/config`). Stacked on #3679 (the blob fetcher).

Makes blob source resolution **lazy from the committed `sources` topic** instead of relying on a compile-time constant. The merged `DefaultRemoteConfigSourceProvider` takes fixed source lists at construction; this adds a lazy wrapper that reads the committed config.

### What's here
- **`RemoteConfigTopicStore`** — a thin read facade over the disk cache (`read()` / `topic(name)`). The minimal surface the source provider needs now; later phases extend it with consumer `body()` resolution.
- **`LazyRemoteConfigSourceProvider`** — wraps a memoized `DefaultRemoteConfigSourceProvider` built from the parsed `sources` topic **plus always-present hardcoded fallbacks** (API + blob) at `Int.MIN_VALUE` priority:
  - Initialized from the cached configuration on first use (before any fetch).
  - Configured/cached sources are always tried first; the hardcoded blob template is the last resort, so blob fetching never regresses to empty.
  - **Non-blocking**: the memo is keyed on the persisted manifest, so only the first resolve after a config change touches disk; every call runs on the fetcher's IO pool, never the main thread.
  - Defensive parsing: `url_format` preferred over `url`, api/blob classified by the `{blob_ref}` placeholder, missing/extra keys tolerated.
- `RemoteConfigBlobFetcher.sourceProvider` becomes a required constructor param (the lazy provider is now the source of the fallback); the dead default factory is removed.

Wiring the provider into `RemoteConfigManager` (prefetch loop, re-arming, `clearCache`) is the next stacked PR.

## Tests
- New `LazyRemoteConfigSourceProviderTest` (fallbacks, parsing, memo invalidation, `clear()`, concurrent failover).
- `./gradlew :purchases:testDefaultsBc8DebugUnitTest --tests "*RemoteConfig*"`, `detektAll`, `:purchases:lintDefaultsBc8Debug` all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3679, opened 2026-06-29
refactor(remote-config): Add `RemoteConfigBlobFetcher`

tonidero:
### Description
Adds initial version of the `RemoteConfigBlobFetcher` that will be in charge of downloading blobs that are not inlined in the remote config response. It includes a priority system so requested blobs have higher priority than others, so we minimize waiting.

---

> [!NOTE]
> **Medium Risk**
> New networked remote-config download path with integrity checks and concurrency, but isolated behind new types and not yet implied as wired into production sync from this diff alone.
> 
> **Overview**
> Introduces **`RemoteConfigBlobFetcher`** to download remote-config blobs by content-address ref into **`RemoteConfigBlobStore`**, with **`ensureDownloaded`** (HIGH) for on-demand reads, **`prefetch`** (LOW) for background warming, deduped in-flight work, a 4-worker cap, and priority boosting when a prefetch item is needed immediately.
> 
> Downloads substitute `{blob_ref}` in configured source URLs, verify payloads with **`RemoteConfigUtils.contentAddressRef`**, fail over unhealthy sources via **`RemoteConfigSourceProvider`**, and return **`Boolean`** without throwing; completion treats a concurrent cache write as success.
> 
> Adds **`RemoteConfigUtils`** for shared ref validation and checksum computation; **`RemoteConfigBlobStore.write`** now returns whether persistence succeeded, using the shared ref guard. Broad unit coverage for fetcher scheduling, verification, failover, and store write outcomes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 756e5ae4108305dd2959c4e0f955a6009c8942fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3698, opened 2026-07-02
refactor(remote-config): Re-arm blob sources only when exhausted + in-memory blob-store index

tonidero:
## Changes

- Add `restartIfExhausted(purpose)` to the source provider: re-arms a purpose only if it is currently exhausted, otherwise a no-op that preserves failover progress.
- On-demand `ensureDownloaded` now re-arms exhausted blob sources before enqueuing, so a read after a prior cycle drained the sources retries instead of failing silently.
- `RemoteConfigManager.prefetchBlobs` re-arms blob sources only when exhausted rather than unconditionally every sync.
- `RemoteConfigBlobStore` keeps an in-memory index of on-disk refs so `contains`/`cachedRefs` no longer stat the disk on every call.

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path blob cache checks and network source selection for remote config; behavior shifts from always retrying primary blob sources each sync to preserving failover state, which could delay recovery if exhaustion detection is wrong.
> 
> **Overview**
> Remote config blob **source failover** no longer resets every sync or on every download. **`restartIfExhausted`** re-winds blob (and API) sources to the first entry only when that purpose has no healthy source left; otherwise failover progress is kept so a known-bad primary is not retried on each sync.
> 
> **On-demand** `ensureDownloaded` calls `restartIfExhausted` before enqueueing a missing blob so reads after an exhausted cycle can retry recovered sources. **Sync prefetch** uses the same conditional re-arm instead of unconditional `restart`. Individual download workers still do not restart inside the failover loop (one re-arm per on-demand request or prefetch cycle).
> 
> **`RemoteConfigBlobStore`** maintains a synchronized in-memory ref index (lazy one-time disk scan, updated on write / `retainOnly` / `clear`). `contains` and `cachedRefs` avoid per-call disk stats; **`read`** drops stale index entries when the file is missing. **`retainOnly`** still scans the blobs directory so orphan temp and invalid-named files are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa9c39d09080b685c95a52ed112dd263f2de87e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
@@ -25,14 +25,20 @@ internal class RemoteConfigBlobStore(
 
     // Refs known to be on disk. Loaded once from a disk scan on first access, then kept in sync by write /
     // retainOnly / clear, so contains() and cachedRefs() answer from memory instead of stat-ing the disk each
-    // call. The store is the sole writer of its directory, so this stays authoritative; read() is still forgiving
-    // (returns null if a file is unexpectedly gone), so a rare divergence is tolerated.
+    // call. The store is the sole writer of its directory, so this stays authoritative; if disk and index ever
+    // diverge, read() self-heals by evicting the ref on a miss, so a stale "cached" flag can't strand a re-fetch.
     private var knownRefs: MutableSet<String>? = null
 
     fun contains(ref: String): Boolean = synchronized(lock) { loadedRefs().contains(ref) }
 
     fun read(ref: String): ByteArray? {
-        val target = blobFile(ref)?.takeIf { it.exists() } ?: return null
+        val target = blobFile(ref)?.takeIf { it.exists() }
+        if (target == null) {
+            // Disk and index disagree (ref gone from under us): correct the index so a later
+            // ensureDownloaded/prefetch re-fetches instead of trusting a stale "cached" flag.
+            synchronized(lock) { knownRefs?.remove(ref) }
+            return null
+        }
         return try {
             target.readBytes()
         } catch (e: IOException) {
@@ -78,13 +84,19 @@ internal class RemoteConfigBlobStore(
     /** The refs currently cached on disk (only files whose name is a valid ref). */
     fun cachedRefs(): Set<String> = synchronized(lock) { loadedRefs().toSet() }
 
-    /** Deletes every cached blob whose ref is not in [refs]. */
+    /**
+     * Deletes every cached blob whose ref is not in [refs]. Scans the directory (not just the in-memory index)
+     * so orphans the index never tracks are pruned too: leftover `rc_blob_*.tmp` files from a write interrupted
+     * by a process kill, and any invalid-named file. Runs once per sync, so the scan is not on a hot path.
+     */
     fun retainOnly(refs: Set<String>) {
-        val toDelete = synchronized(lock) { loadedRefs().filterNot { it in refs } }
-        if (toDelete.isEmpty()) return
         val parent = blobsDir()
-        toDelete.forEach { deleteQuietly(File(parent, it)) }
-        synchronized(lock) { loadedRefs().removeAll(toDelete.toSet()) }
+        if (parent.exists()) {
+            parent.listFiles()
+                ?.filter { it.isFile && it.name !in refs }
+                ?.forEach { deleteQuietly(it) }
+        }
+        synchronized(lock) { loadedRefs().retainAll(refs) }
     }
 
     /**
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
@@ -21,7 +21,15 @@ import java.nio.ByteBuffer
 internal class RemoteConfigBlobStore(
     private val applicationContext: Context,
 ) {
-    fun contains(ref: String): Boolean = blobFile(ref)?.exists() == true
+    private val lock = Any()
+
+    // Refs known to be on disk. Loaded once from a disk scan on first access, then kept in sync by write /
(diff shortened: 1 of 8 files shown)


tonidero:
* **#3703** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3703?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3705](https://github.com/RevenueCat/purchases-android/pull/3705) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3705?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3708](https://github.com/RevenueCat/purchases-android/pull/3708) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3708?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3700** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3698** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3698?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3698?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3683** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3679** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3679?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3682](https://github.com/RevenueCat/purchases-android/pull/3682) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3682?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
```diff
@@ -67,34 +76,43 @@ internal class RemoteConfigBlobStore(
...
     fun retainOnly(refs: Set<String>) {
+        val toDelete = synchronized(lock) { loadedRefs().filterNot { it in refs } }
+        if (toDelete.isEmpty()) return
         val parent = blobsDir()
-        if (!parent.exists()) return
-        parent.listFiles()
-            ?.filter { it.isFile && it.name !in refs }
-            ?.forEach { deleteQuietly(it) }
+        toDelete.forEach { deleteQuietly(File(parent, it)) }
+        synchronized(lock) { loadedRefs().removeAll(toDelete.toSet()) }
```
vegaro: Looks like there's a minor behavior change here. The old implementation deleted every file in the blobs dir whose name wasn't in refs. The new one only deletes files present in the in-memory index, and the index only ever contains valid refs. So there are two categories of files are now never pruned, orphaned `rc_blob_*.tmp` files left by a process kill mid-write (the finally cleanup doesn’t run on a kill) and any invalid file that ends up in the directory for other reasons.

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
```diff
@@ -21,7 +21,15 @@ import java.nio.ByteBuffer
...
     private val applicationContext: Context,
 ) {
-    fun contains(ref: String): Boolean = blobFile(ref)?.exists() == true
+    private val lock = Any()
+
+    // Refs known to be on disk. Loaded once from a disk scan on first access, then kept in sync by write /
+    // retainOnly / clear, so contains() and cachedRefs() answer from memory instead of stat-ing the disk each
+    // call. The store is the sole writer of its directory, so this stays authoritative; read() is still forgiving
+    // (returns null if a file is unexpectedly gone), so a rare divergence is tolerated.
+    private var knownRefs: MutableSet<String>? = null
```
rickvdl, line 30: Hm I have one concern with this; if I'm reading the implementation correct the following could happen

- Android purges a file from cache
- `ensureDownloaded(ref)` returns true (because a file is in the `knownRefs`
- We try to load the blob using `read(ref)` which will return `null`

In that case I don't think there is a workaround for forcing the fetcher to download it again? And also that seems like an implementation detail that's then leaking, because as far as the consumers of the blob store go if it exists they should be able to load it?

tonidero: I believe it shouldn't be an issue because we first initialize the in memory cache from the files on disk, so if we purge a file, it should also update this... however, I will add a check that updates the in memory cache if we somehow enter into this state, so it self heals and it works on the next attempt.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
tonidero: Good point! Addressed in https://github.com/RevenueCat/purchases-android/pull/3698/commits/0535058c75f6c862619b6685ad22715fc9ee9d09

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigBlobStore.kt
tonidero: Addressed that self heal in https://github.com/RevenueCat/purchases-android/pull/3698/commits/0535058c75f6c862619b6685ad22715fc9ee9d09

rickvdl: Awesome, thanks for iterating 🙌

rickvdl (approved):
Awesome! Thanks for iterating 🫶

PR #3703, opened 2026-07-02
refactor(remote-config): add RemoteConfigManager read facade (topic/body)

tonidero:
Adds a read API to `RemoteConfigManager` so consumers can read remote-config data:

- `topic(name)` — a topic's committed metadata.
- `blobData(topicName, itemKey)` — a resolved item payload (its inline content, or its blob).

On a cache miss, a read resolves from an in-progress or on-demand sync instead of returning empty.

---

> [!NOTE]
> **Medium Risk**
> New suspend read paths can block on network and trigger on-demand `/v1/config` syncs; behavior is complex but heavily tested and gated on identity/disable state.
> 
> **Overview**
> Adds a **suspend read API** on `RemoteConfigManager`: `topic(RemoteConfigTopic)` for committed item metadata, and `blobData` (transform + JSON reified) for blob-backed payloads. Reads run on IO, respect the session **4xx kill-switch** (`isUnavailable` → `isDisabled`), and on a cache miss **wait for an in-flight sync or trigger a foreground on-demand refresh** when an app user ID is available (wired from `PurchasesFactory` via `appUserIDProvider`).
> 
> Introduces **`RemoteConfigTopic`** and threads it through `RemoteConfigTopicStore` / source parsing. Renames config item inline fields **`content` → `metadata`** and documents that payloads without `blob_ref` are index-only (`blobData` returns null).
> 
> `syncAttributesAndOfferingsIfNeeded` now calls **`refreshRemoteConfig`** before fetching offerings. Refresh completion is signaled with **`CompletableDeferred`** so blocked readers unblock on 204, errors, `clearCache`, and `close`. Unit, integration, and production golden tests cover the facade and waiting behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3f74fbf745f3a2862b9917d76ecd270565a0a6a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -279,7 +279,7 @@ internal class PurchasesFactory(
                     backend = backend,
                     diskCache = RemoteConfigDiskCache(contextForStorage),
                     blobStore = RemoteConfigBlobStore(contextForStorage),
-                    // Lets a cold on-demand read self-trigger a sync for the current user (see body()).
+                    // Lets a cold on-demand read self-trigger a sync for the current user (see blobData()).
                     appUserIDProvider = { cache.getCachedAppUserID() },
                 )
             } else {
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -279,6 +279,8 @@ internal class PurchasesFactory(
                     backend = backend,
                     diskCache = RemoteConfigDiskCache(contextForStorage),
                     blobStore = RemoteConfigBlobStore(contextForStorage),
+                    // Lets a cold on-demand read self-trigger a sync for the current user (see body()).
+                    appUserIDProvider = { cache.getCachedAppUserID() },
                 )
             } else {
                 null
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -67,18 +67,12 @@ internal class RemoteConfigManager(
     private val blobStore: RemoteConfigBlobStore,
     private val dateProvider: DateProvider = DefaultDateProvider(),
     private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
-    // The manager owns the threading for its reads: disk-touching read methods hop to this dispatcher, so
-    // callers never read disk on their own thread. Injectable so tests can run reads deterministically.
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
-    // The single read seam over the persisted item index: shared by the read facade (topic()/body()) and the
-    // source provider. Synchronous + metadata-only (no blob, no wait); the manager adds the IO hop / wait.
     private val topicStore: RemoteConfigTopicStore =
-        RemoteConfigTopicStore { diskCache.read()?.topics?.get(it) },
+        RemoteConfigTopicStore { diskCache.read()?.topics?.get(it.wireName) },
     private val sourceProvider: RemoteConfigSourceProvider =
         DefaultRemoteConfigSourceProvider(topicStore),
     private val blobFetcher: RemoteConfigBlobFetcher = RemoteConfigBlobFetcher(blobStore, sourceProvider),
-    // The current app user id, used only to self-trigger a sync on a cold read (see [body]). A blank/null result
-    // means "no user yet", so no sync is triggered. Defaults to none so tests never self-trigger unless they opt in.
     private val appUserIDProvider: () -> String? = { null },
 ) {
     private val isRefreshing = AtomicBoolean(false)
@@ -217,7 +211,6 @@ internal class RemoteConfigManager(
             epoch.incrementAndGet()
             isRefreshing.set(false)
             lastRefreshedAt = null
-            // Unblock any read waiting on the refresh we just superseded; it re-reads the now-wiped cache.
             completeRefresh()
             // Intentionally NOT resetting `unavailable`: a 4xx is an endpoint/app-level fact that outlives an
             // identity change. It clears only on app restart.
@@ -247,11 +240,6 @@ internal class RemoteConfigManager(
         owned
     }
 
-    /**
-     * Unblocks any read waiting on the in-flight refresh and clears the handle. Completes (never cancels) so a
-     * waiting [body] resumes and re-reads the committed state rather than throwing. Must be called while holding
-     * [cacheLock] — every call site (terminal refresh points and [clearCache]) already does.
-     */
     private fun completeRefresh() {
         refreshCompletion?.complete(Unit)
         refreshCompletion = null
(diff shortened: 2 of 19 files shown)


tonidero:
* **#3703** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3703?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3705](https://github.com/RevenueCat/purchases-android/pull/3705) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3705?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3708](https://github.com/RevenueCat/purchases-android/pull/3708) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3708?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3703?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3700** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3700?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3698** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3698?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3683** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3679** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3679?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3682](https://github.com/RevenueCat/purchases-android/pull/3682) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3682?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -206,10 +230,133 @@ internal class RemoteConfigManager(
...
+     * When the topic isn't committed yet, [awaitConfigForRead] first waits for a refresh in progress — or
+     * triggers one on demand — and then re-reads before giving up, so a read during the initial sync (or before
+     * any sync) returns fresh data instead of `null`, mirroring [blobData]. A committed topic returns
+     * immediately, never delayed by an unrelated in-flight refresh. Use [blobData] for a resolved item payload
+     * that also resolves the referenced blob. Reads disk on [ioDispatcher].
+     */
+    suspend fun topic(topic: RemoteConfigTopic): ConfigTopic? = withContext(ioDispatcher) {
+        if (disabled) return@withContext null
+        topicStore.topic(topic)?.let { return@withContext it }
+        awaitConfigForRead()
```
rickvdl, line 295: Hm would this fetch the config every time we ask for a (non existent) topic?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl, line 221: I think this one could be expanded now with the new responsibilities of the manager;
- Completing the `refreshCompletion`
- clearing `isRefreshing`
- Prevent future calls? Maybe not needed if we assume the instance will be deallocated.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -206,10 +230,133 @@ internal class RemoteConfigManager(
...
+    /**
+     * The resolved blob payload for `itemKey` in [topic], parsed from JSON into [T], or `null` when the item
+     * is unknown, has no `blob_ref`, its blob can't be resolved, or its bytes don't deserialize into [T]. [T]
+     * must be a concrete `@Serializable` type; parsing uses the shared [JsonProvider.defaultJson]. For non-JSON
+     * payloads use the `transform` overload, which also documents the resolution and waiting rules.
+     */
+    suspend inline fun <reified T> blobData(topic: RemoteConfigTopic, itemKey: String): T? =
+        blobData(topic, itemKey) { bytes ->
+            try {
+                JsonProvider.defaultJson.decodeFromString<T>(bytes.decodeToString())
```
rickvdl, line 308: I think this is what you meant over slack; but decoding errors are now also returned as null (which is expected in it's current state), but indeed maybe some typed errors would be nice

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -279,6 +279,8 @@ internal class PurchasesFactory(
                     backend = backend,
                     diskCache = RemoteConfigDiskCache(contextForStorage),
                     blobStore = RemoteConfigBlobStore(contextForStorage),
+                    // Lets a cold on-demand read self-trigger a sync for the current user (see body()).
```
rickvdl: `body()` has been renamed :)

rickvdl (commented):
Awesome work! Couple of notes / questions

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Right, in the end I guess it depends on the consumers and whether and how we want to expose this later on. We already log these internally, so a consumer of this API right now can only know that something went wrong, but not what... So yeah, adding a better response might make more sense, but might leave that for a separate PR and also see if consumers do actually need that info.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Yes... it would indeed. However, since we pass in a `RemoteConfigTopic` which has a set of values the SDK understands, it wouldn't be possible to ask for a non-existent topic, and we should have a value for those existing ones. So I don't think it should be an issue?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Well... we indeed can, but I believe all of this should be fine since basically we're assuming the instance and all of these will be deallocated together with the instance. So feels to me a bit unnecessary... But yeah, I can do it just in case :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Right, that indeed should narrow the surface and that's true as long as we never remove a topic. So I guess it should be fine indeed. We could introduce a list of topics that don't exist and keep an in-memory cache of those, but that depends on how topics will be used in the future, so let's not go that route for now..

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Alright that makes sense, let's leave it like that :)

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/3703/commits/e3f74fbf745f3a2862b9917d76ecd270565a0a6a

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/3703/commits/e3f74fbf745f3a2862b9917d76ecd270565a0a6a

rickvdl (approved):
Awesome work! Almost there 💪

PR #3711, opened 2026-07-03
UI Config via config endpoint

vegaro:
Adds `UiConfigProvider` + `getUiConfig`/`awaitGetUiConfig`. Nothing consumes this yet, the WorkflowManager will use it directly

---

> [!NOTE]
> **Medium Risk**
> New remote-config fetch path affects paywall UI settings with all-or-nothing merge and silent default fallback on failure; guarded by feature flags and preview mode but will be security-sensitive once workflows depend on it.
> 
> **Overview**
> Adds a path to load **UiConfig** from the **`/v1/config`** remote-config **`ui_config`** topic instead of only from legacy offerings JSON.
> 
> A new **`UiConfigProvider`** merges four blob-backed items (`app`, `localizations`, `variable_config`, `custom_variables`) via **`RemoteConfigManager.mergeItemsBlobData`** and decodes them into **`UiConfig`**. If merge or decode fails, callers get a default **`UiConfig()`** rather than a partial object.
> 
> **`PurchasesFactory`** wires **`UiConfigProvider`** when **`RemoteConfigManager`** is enabled. **`PurchasesOrchestrator.getUiConfig()`** and internal **`awaitGetUiConfig()`** expose this; they throw **`ConfigurationError`** in UI preview mode or when remote config / the provider is unavailable. **RevenueCatUI**’s **`PurchasesType`** gains **`awaitGetUiConfig`** for upcoming workflow integration; nothing in the SDK consumes it yet.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9154059143d8aee9e5c15e06fe9e86e9f5c12897. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -48,9 +48,9 @@ import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.common.offerings.OfferingsManager
 import com.revenuecat.purchases.common.offlineentitlements.OfflineEntitlementsManager
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigManager
-import com.revenuecat.purchases.common.remoteconfig.UiConfigProvider
 import com.revenuecat.purchases.common.sha1
 import com.revenuecat.purchases.common.subscriberattributes.SubscriberAttributeKey
+import com.revenuecat.purchases.common.uiconfig.UiConfigProvider
 import com.revenuecat.purchases.common.verboseLog
 import com.revenuecat.purchases.common.warnLog
 import com.revenuecat.purchases.common.workflows.WorkflowDataResult
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -48,6 +48,7 @@ import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.common.offerings.OfferingsManager
 import com.revenuecat.purchases.common.offlineentitlements.OfflineEntitlementsManager
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigManager
+import com.revenuecat.purchases.common.remoteconfig.UiConfigProvider
 import com.revenuecat.purchases.common.sha1
 import com.revenuecat.purchases.common.subscriberattributes.SubscriberAttributeKey
 import com.revenuecat.purchases.common.verboseLog
@@ -106,6 +107,11 @@ import com.revenuecat.purchases.utils.Result
 import com.revenuecat.purchases.utils.isAndroidNOrNewer
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencies
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencyManager
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.launch
 import java.net.URL
 import java.util.Collections
 import java.util.Date
@@ -162,9 +168,12 @@ internal class PurchasesOrchestrator(
     private val backupManager: BackupManager = BackupManager(application),
     val fileRepository: FileRepository = DefaultFileRepository(application),
     private val remoteConfigManager: RemoteConfigManager? = null,
+    private val uiConfigProvider: UiConfigProvider? = null,
     @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
     val adTracker: AdTracker = AdTracker(adEventsManager),
 ) : LifecycleDelegate, CustomActivityLifecycleHandler {
+
+    private val uiConfigScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
 
     internal var state: PurchasesState
         get() = purchasesStateCache.purchasesState
@@ -619,6 +628,44 @@ internal class PurchasesOrchestrator(
     fun workflowIdForOfferingId(offeringId: String): String? =
         workflowManager?.workflowIdForOfferingId(offeringId)
 
+    fun getUiConfig(
+        onSuccess: (UiConfig) -> Unit,
+        onError: (PurchasesError) -> Unit,
+    ) {
+        if (appConfig.uiPreviewMode) {
+            dispatch {
+                onError(
+                    PurchasesError(
+                        PurchasesErrorCode.ConfigurationError,
+                        "UI config cannot be fetched in UI preview mode.",
+                    ),
+                )
+            }
+            return
+        }
+        if (uiConfigProvider == null) {
+            dispatch {
+                onError(
+                    PurchasesError(
+                        PurchasesErrorCode.ConfigurationError,
+                        "UI config is not enabled.",
+                    ),
+                )
+            }
+            return
+        }
+        uiConfigScope.launch {
+            try {
+                val uiConfig = uiConfigProvider.getUiConfig()
+                dispatch { onSuccess(uiConfig) }
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                dispatch {
+                    onError(PurchasesError(PurchasesErrorCode.UnknownError, "Failed to fetch UI config: ${e.message}"))
+                }
+            }
+        }
+    }
+
     fun getProducts(
         productIds: List<String>,
         type: ProductType? = null,
@@ -852,6 +899,7 @@ internal class PurchasesOrchestrator(
         }
         this.backend.close()
         this.remoteConfigManager?.close()
+        this.uiConfigScope.cancel()
         this.workflowManager?.close()
 
         billing.close()
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -631,13 +631,7 @@ internal class PurchasesOrchestrator(
             }
             throw PurchasesException(PurchasesError(PurchasesErrorCode.ConfigurationError, message))
         }
-        return try {
-            provider.getUiConfig()
-        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
-            throw PurchasesException(
-                PurchasesError(PurchasesErrorCode.UnknownError, "Failed to fetch UI config: ${e.message}"),
-            ).apply { initCause(e) }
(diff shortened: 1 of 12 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/UiConfigProvider.kt
```diff
@@ -0,0 +1,32 @@
+package com.revenuecat.purchases.common.remoteconfig
```
tonidero: Not a huge deal but thinking maybe we could move these to a specific package? Like a `providers` package. Alternatively, this could be part of a workflows package outside the `remoteconfig` package, since it would be more specific to that feature.

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+ * the gap.
+ */
+@OptIn(InternalRevenueCatAPI::class)
+internal class UiConfigProvider(
+    private val manager: RemoteConfigManager,
+) {
+
+    suspend fun getUiConfig(): UiConfig {
+        val parts = PART_KEYS.mapNotNull { key ->
+            manager.blobData<JsonObject>(RemoteConfigTopic.UiConfig, key)?.let { key to it }
```
tonidero: Hmm would it work if you pass in the `UiConfig` type directly
```suggestion
            manager.blobData<UiConfig>(RemoteConfigTopic.UiConfig, key)
```
I believe the method should perform the deserialization automatically (or that was the intention 😅 )

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+        }.toMap()
+        return JsonTools.json.decodeFromJsonElement(JsonObject(parts))
+    }
+
+    private companion object {
+        private val PART_KEYS = listOf("app", "localizations", "variable_config", "custom_variables")
```
tonidero: If we do the above, we won't need this

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -162,10 +168,13 @@ internal class PurchasesOrchestrator(
     private val backupManager: BackupManager = BackupManager(application),
     val fileRepository: FileRepository = DefaultFileRepository(application),
     private val remoteConfigManager: RemoteConfigManager? = null,
+    private val uiConfigProvider: UiConfigProvider? = null,
     @OptIn(ExperimentalPreviewRevenueCatPurchasesAPI::class)
     val adTracker: AdTracker = AdTracker(adEventsManager),
 ) : LifecycleDelegate, CustomActivityLifecycleHandler {
 
+    private val uiConfigScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```
tonidero: Hmm I wonder if this is needed... the `blobData` operation should already run in an IO thread if I'm correct.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -619,6 +628,44 @@ internal class PurchasesOrchestrator(
     fun workflowIdForOfferingId(offeringId: String): String? =
         workflowManager?.workflowIdForOfferingId(offeringId)
 
+    fun getUiConfig(
```
tonidero: If we don't need the scope here, we could make this function `suspend` directly and have the UI module handle the result/errors (shouldn't error I think but maybe add a try-catch just in case :) )

tonidero (commented):
Some questions that could potentially cleanup a few things, but nothing big. Great work!

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
vegaro: ah that's pretty obvious hahaha, missed this one when self review. Thanks, will clean it up!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
vegaro: ok. I was a bit hesitant to add this, so was kinda waiting for your review to see what you thought 😆

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
vegaro: hmm thinking about this a bit more... will that work?

the ui config topic has 4 separate blobs

```
 ui_config topic
    "app" → blob_ref "Ewnl…"
    "localizations"  → blob_ref "Jewn…"
    "variable_config" → blob_ref "xXyp…"
    "custom_variables"→ blob_ref "l6yC…"
```

Wouldn't that only work if `ui_config` was a single blob?

vegaro: I could do this though:

```kotlin
UiConfig(
    app = manager.blobData<AppConfig>(topic, "app") ?: AppConfig(),
    // `localizations` needs its property-level serializer, which skips unknown VariableLocalizationKeys.
    localizations = manager.blobData(topic, "localizations") { bytes ->
        JsonTools.json.decodeFromString(
            LocalizedVariableLocalizationKeyMapSerializer,
            bytes.decodeToString(),
        )
    } ?: emptyMap(),
    variableConfig = manager.blobData<VariableConfig>(topic, "variable_config") ?: VariableConfig(),
    customVariables = manager.blobData<Map<String, CustomVariableDefinition>>(topic, "custom_variables")
        ?: emptyMap(),
)
```

tonidero: Ohh yes you're right... if we keep these as separate blobs, indeed we will need to load them separately. Good catch!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -619,6 +621,25 @@ internal class PurchasesOrchestrator(
...
+            }
+            throw PurchasesException(PurchasesError(PurchasesErrorCode.ConfigurationError, message))
+        }
+        return try {
+            provider.getUiConfig()
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+            throw PurchasesException(
+                PurchasesError(PurchasesErrorCode.UnknownError, "Failed to fetch UI config: ${e.message}"),
+            ).apply { initCause(e) }
+        }
```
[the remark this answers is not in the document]
facumenzella: This seems legit, right?

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
vegaro: I updated it but now I am considering if it's better than what we had before. For example `localizations` has a custom deserializer:

```kotlin
  @Serializable(with = LocalizedVariableLocalizationKeyMapSerializer::class)
  public val localizations: Map<LocaleId, Map<VariableLocalizationKey, String>> = emptyMap(),
```

Since we are now constructing `UiConfig` by hand, we need to add defaults and call the `decodeFromString` passing the deserializer `LocalizedVariableLocalizationKeyMapSerializer`

What do you think? The previous version was this, which applies custom deserializers and defaults automatically:

```kotlin
suspend fun getUiConfig(): UiConfig {
    val parts = PART_KEYS.mapNotNull { key ->
        manager.blobData<JsonObject>(RemoteConfigTopic.UiConfig, key)?.let { key to it }
    }.toMap()
    return JsonTools.json.decodeFromJsonElement(JsonObject(parts))
}
```

tonidero: Hmm right... it's certainly not as clean as it is now... TBH, with this I don't really mind either way, happy to leave it as you had it before 👍

tonidero: @rickvdl [suggested](https://github.com/RevenueCat/purchases-ios/pull/7140#discussion_r3529449311) adding a helper to the manager to handle these cases where we need to fetch multiple blobs simultaneously. Wdyt? Happy to add it and could also handle the parallelization of this process.

rickvdl: We could also add an optional `mapInto:` parameter or something like that that allows constructing a container type (`UIConfig` in this case) from the set of fetched blobs if that helps?

vegaro: that would be great yeah, I imagine this will become a common pattern, so better to have a helper for it

rickvdl: Alright, will work on that as well :). 

cc @facumenzella for ios, no need to wait or change anything I'd say, I'll implement the new API in my PR right away.

tonidero: Done in Android in this PR: https://github.com/RevenueCat/purchases-android/pull/3725

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+ * topic (not inline metadata), resolved and decoded into its target type through [RemoteConfigManager.blobData].
+ * A part with no resolvable blob (topic absent, item absent, or blob unresolvable) falls back to [UiConfig]'s
+ * own field default.
+ */
+@OptIn(InternalRevenueCatAPI::class)
+internal class UiConfigProvider(
+    private val manager: RemoteConfigManager,
+) {
+
+    suspend fun getUiConfig(): UiConfig {
```
facumenzella: On iOS, this can fail 🤔 
Both `app` and `localizations` are treated as required, and if one fails the whole thing fails. 

What's the expectation here? We should be doing the same

vegaro: that's changing when I start using this helper https://github.com/RevenueCat/purchases-android/pull/3725/files#r3530180787

facumenzella (approved):
This makes sense to me

PR #3713, opened 2026-07-03
fix(paywalls): pre-download Paywalls V2 icons from the full base URL path

### LINKED ISSUE #3712: PaywallComponentsImagePreDownloader drops base URL path for IconComponent — icon pre-download 403s on every app start (Uri.Builder.path() replaces instead of appends)

**Describe the bug**

`PaywallComponentsImagePreDownloader` builds malformed URLs for Paywalls V2 `IconComponent`s, so the icon pre-download fails with **HTTP 403 on every app start, for every Android user**, for any paywall containing icon components.

The cause is in [`PaywallComponentsImagePreDownloader.kt#L76`](https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt#L76):

```kotlin
is IconComponent -> {
    setOf(Uri.parse(component.baseUrl).buildUpon().path(component.formats.webp).build())
}
```

`Uri.Builder.path()` **replaces** the URI's path instead of appending to it. The offerings response serves `"base_url": "https://icons.pawwalls.com/icons"`, so for `check.webp` this produces `https://icons.pawwalls.com/check.webp` — the `/icons` segment is dropped — which returns 403 (S3 `AccessDenied` via CloudFront). The correct URL returns 200.

The paywall still *renders* correctly because `IconComponentState.url` concatenates correctly (`"$baseUrl/${formats.webp}"`) — so the pre-warm silently never works (icons are fetched cold at render time) and the only visible symptom is the failed requests. iOS is unaffected: `PaywallV2CacheWarming.swift` uses `base.appendingPathComponent(...)`, which appends properly.

1. Environment
   1. Platform: Android (via React Native)
   2. SDK version: react-native-purchases / react-native-purchases-ui 10.4.1 → purchases-android **10.11.0**. The bug is version-independent — the code is still present on `main` (link above).
   3. OS version: all Android versions (observed fleet-wide via APM across our user base)
   4. Android Studio version: N/A
   5. How widespread is the issue. Percentage of devices affected: **~100% of Android devices** — the pre-downloader runs on every app start after offerings are fetched, no paywall interaction needed. We see ~1,150 failed requests/week in New Relic (`MobileRequestError`, 3 icons × every session).

2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue

```
[2026-07-03T12:52:57.321Z] [DEBUG] [RevenueCat] ℹ️ SDK Version - 10.11.0
[2026-07-03T12:52:57.552Z] [DEBUG] [RevenueCat] 😻 Start Offerings update from network.
[2026-07-03T12:52:57.829Z] [DEBUG] [RevenueCat] OfferingImagePreDownloader: starting image download
[2026-07-03T12:52:58.214Z] [DEBUG] [RevenueCat] Pre-downloading Paywall V2 image: https://assets.pawwalls.com/1262234_low_res_1756473183.webp   <-- ImageComponent: full URL from config, works
[2026-07-03T12:52:58.240Z] [DEBUG] [RevenueCat] Pre-downloading Paywall V2 image: https://icons.pawwalls.com/lock-open-2.webp                   <-- IconComponents: /icons/ segment dropped, all 403
[2026-07-03T12:52:58.242Z] [DEBUG] [RevenueCat] Pre-downloading Paywall V2 image: https://icons.pawwalls.com/x.webp
[2026-07-03T12:52:58.244Z] [DEBUG] [RevenueCat] Pre-downloading Paywall V2 image: https://icons.pawwalls.com/check.webp
```

Note the `ImageComponent` download in the same batch uses a full URL from the config and succeeds — only the `IconComponent` branch (base URL + filename) builds broken URLs, which isolates the bug to that `when` branch of `findImageUrisToDownload()`.

   1. Create a Paywalls V2 paywall containing any icon component (ours uses `check`, `x`, and `lock-open-2`) and attach it to the current offering.
   2. Launch an Android app with `purchases-ui` present and RevenueCat debug logging enabled.
   3. **Expected:** icons pre-downloaded from `https://icons.pawwalls.com/icons/<name>.webp` (HTTP 200).
   4. **Actual:** requests go to `https://icons.pawwalls.com/<name>.webp` → HTTP 403 for each icon, on every app start.

   Reproducible without a device:

   ```
   $ curl -so /dev/null -w "%{http_code}\n" https://icons.pawwalls.com/lock-open-2.webp        # 403 (URL the SDK builds)
   $ curl -so /dev/null -w "%{http_code}\n" https://icons.pawwalls.com/icons/lock-open-2.webp  # 200 (correct URL)
   ```

4. **Other information**

   Suggested fix — append instead of replace, matching the render path in `IconComponentState`:

   ```kotlin
   is IconComponent -> {
       setOf(Uri.parse("${component.baseUrl}/${component.formats.webp}"))
   }
   ```

   Impact beyond the wasted pre-warm: the failing requests flood APM tools that auto-instrument OkHttp (New Relic, Datadog, Sentry, …) with 403 network errors on every session, which is how we found this.

**Additional context**

The 403 rather than 404 is S3 returning `AccessDenied` for a nonexistent key via CloudFront (`server: AmazonS3`, `x-cache: Error from cloudfront`), which made this initially look like a CDN permissions problem — it's actually a client-side URL construction bug.

If you want to cross-link your RevenueCat community topic, add the URL at the end of section 4 — I left it out since I don't have the link. And remember the LOG_LEVEL.DEBUG revert in revenueCatService.ts:48 when you're done.

bilqisium:
### Checklist
- [x] If applicable, create follow-up issues for `purchases-ios` and hybrids — none needed: iOS appends correctly (`PaywallV2CacheWarming.swift` uses `appendingPathComponent`), and hybrids inherit this fix through `purchases-android`.

### Motivation
Resolves #3712

`PaywallComponentsImagePreDownloader` builds `IconComponent` pre-download URLs with `Uri.Builder.path()`, which **replaces** the base URL's path instead of appending to it. Icon `base_url`s are served as `https://icons.pawwalls.com/icons`, so the `/icons` segment was silently dropped and every pre-download request hit `https://icons.pawwalls.com/<name>.webp` → S3/CloudFront 403, on every offerings fetch, for every Android user whose paywall contains icon components (reported as fleet-wide APM noise in New Relic, see the linked issue and [community thread](https://community.revenuecat.com/questions/6407)).

The paywall still rendered correctly because `IconComponentState.url` concatenates correctly — which also means the pre-warm has been silently useless since it shipped in 8.12.0 (#2143): icons were always fetched cold at render time.

### Description
- Build the icon URL by concatenation, exactly like `IconComponentState.url` does. Matching the render-path construction is the actual requirement here (not just producing a *valid* URL): the pre-warm only helps if it populates the image cache under the **same URL string** the render path requests.
- Updated the `OfferingImagePreDownloaderTest` icon fixtures to use a base URL with a path segment (`https://pawwalls.com/icons`), matching how icon `base_url`s are served in production. The previous fixtures used a pathless base (`https://pawwalls.com`), for which *replace* and *append* coincidentally produce the same URL — which is why the existing test never caught this.

Intentionally untouched:
- The two `Uri.Builder.path()` usages on the Paywalls **V1** path (`OfferingImagePreDownloader`, `TemplateConfigurationFactory`): V1 `asset_base_url` has no path segment, and the V1 render + pre-download paths use the same construction, so they agree with each other today.
- Icon `overrides` are not pre-warmed (only the base icon), unlike image/stack overrides — pre-existing gap, left out of scope to keep this a pure bug fix.

**Testing**: updated tests fail on `main` (`paywalls V2 - if images, it downloads all of them` captures `…/test_icon_1.webp` without the `/icons` segment) and pass with this change (`:purchases:testDefaultsBc8DebugUnitTest`). `detektAll` clean. Also verified against production CDN: `curl https://icons.pawwalls.com/check.webp` → 403, `curl https://icons.pawwalls.com/icons/check.webp` → 200.

---

> [!NOTE]
> **Low Risk**
> Narrow URL-construction fix in best-effort image pre-warming; no auth, purchase, or data-model changes.
> 
> **Overview**
> Fixes Paywalls V2 icon **pre-download** URLs so they include the full `baseUrl` path (e.g. `/icons`) instead of dropping it.
> 
> `PaywallComponentsImagePreDownloader` now builds icon URIs with `appendEncodedPath` on the parsed base URL, aligning with how `IconComponentState` concatenates at render time so Coil warms the **same cache key** the UI requests. Previously `Uri.Builder.path()` replaced the base path, causing 403s on offerings fetch while paywalls still rendered from cold loads.
> 
> Unit test fixtures now use `https://pawwalls.com/icons` as the icon `baseUrl` and assert `…/icons/test_icon_*.webp` so the bug is covered (pathless bases masked the mismatch).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84ec1d564338a9b75b4ee004a4396ad4dbf6439c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
@@ -73,7 +73,10 @@ internal class PaywallComponentsImagePreDownloader(
                             component.overrides.imageUrisToDownload { it.background.findImageUrisToDownload() }
                     }
                     is IconComponent -> {
-                        setOf(Uri.parse(component.baseUrl).buildUpon().path(component.formats.webp).build())
+                        // Concatenate like IconComponentState.url does. Uri.Builder.path() would replace the
+                        // base URL's path (dropping e.g. /icons), and any mismatch with the render-time URL
+                        // defeats the pre-warm since the image cache is keyed by the exact URL string.
+                        setOf(Uri.parse("${component.baseUrl}/${component.formats.webp}"))
                     }
                     is CarouselComponent -> {
                         // pages are visited by BFS; only extract this component's own background
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
@@ -73,10 +73,16 @@ internal class PaywallComponentsImagePreDownloader(
                             component.overrides.imageUrisToDownload { it.background.findImageUrisToDownload() }
                     }
                     is IconComponent -> {
-                        // Concatenate like IconComponentState.url does. Uri.Builder.path() would replace the
-                        // base URL's path (dropping e.g. /icons), and any mismatch with the render-time URL
-                        // defeats the pre-warm since the image cache is keyed by the exact URL string.
-                        setOf(Uri.parse("${component.baseUrl}/${component.formats.webp}"))
+                        // appendEncodedPath, not path() (which replaces the base URL's path, dropping e.g.
+                        // /icons) nor appendPath (which percent-encodes the segment). The result must match
+                        // the URL IconComponentState builds at render time by raw concatenation, since the
+                        // image cache is keyed by the exact URL string.
+                        setOf(
+                            Uri.parse(component.baseUrl)
+                                .buildUpon()
+                                .appendEncodedPath(component.formats.webp)
+                                .build(),
+                        )
                     }
                     is CarouselComponent -> {
                         // pages are visited by BFS; only extract this component's own background
--- a/purchases/src/test/java/com/revenuecat/purchases/utils/OfferingImagePreDownloaderTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/utils/OfferingImagePreDownloaderTest.kt
@@ -150,7 +150,7 @@ class OfferingImagePreDownloaderTest {
         val expectedImageDownloads = listOf(
             "https://pawwalls.com/test_stack_light_low_res.webp",
             "https://pawwalls.com/test_stack_dark_low_res.webp",
-            "https://pawwalls.com/test_icon_1.webp",
+            "https://pawwalls.com/icons/test_icon_1.webp",
             "https://pawwalls.com/test_image_light_low_res.webp",
             "https://pawwalls.com/test_image_dark_low_res.webp",
             "https://pawwalls.com/test_image_override_light_low_res.webp",
@@ -168,11 +168,11 @@ class OfferingImagePreDownloaderTest {
             "https://pawwalls.com/test_sticky_footer_override_dark_low_res.webp",
             "https://pawwalls.com/test_sticky_footer_override_2_light_low_res.webp",
             "https://pawwalls.com/test_sticky_footer_override_2_dark_low_res.webp",
-            "https://pawwalls.com/test_icon_2.webp",
+            "https://pawwalls.com/icons/test_icon_2.webp",
             "https://pawwalls.com/test_header_low_res.webp",
-            "https://pawwalls.com/test_icon_5.webp",
-            "https://pawwalls.com/test_icon_3.webp",
-            "https://pawwalls.com/test_icon_4.webp",
+            "https://pawwalls.com/icons/test_icon_5.webp",
+            "https://pawwalls.com/icons/test_icon_3.webp",
+            "https://pawwalls.com/icons/test_icon_4.webp",
         )
 
         preDownloader.preDownloadOfferingImages(createOfferingWithV2Paywall(
@@ -193,7 +193,7 @@ class OfferingImagePreDownloaderTest {
                             color = ColorScheme(light = ColorInfo.Alias(ColorAlias(""))),
                         ),
                         IconComponent(
-                            baseUrl = "https://pawwalls.com",
+                            baseUrl = "https://pawwalls.com/icons",
                             iconName = "test_icon",
                             formats = IconComponent.Formats(
                                 webp = "test_icon_1.webp",
@@ -292,7 +292,7 @@ class OfferingImagePreDownloaderTest {
                                         color = ColorScheme(light = ColorInfo.Alias(ColorAlias(""))),
                                     ),
                                     icon = IconComponent(
-                                        baseUrl = "https://pawwalls.com",
+                                        baseUrl = "https://pawwalls.com/icons",
                                         iconName = "test_icon",
                                         formats = IconComponent.Formats(
                                             webp = "test_icon_3.webp",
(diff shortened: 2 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
```diff
@@ -73,7 +73,10 @@ internal class PaywallComponentsImagePreDownloader(
                             component.overrides.imageUrisToDownload { it.background.findImageUrisToDownload() }
                     }
                     is IconComponent -> {
-                        setOf(Uri.parse(component.baseUrl).buildUpon().path(component.formats.webp).build())
+                        // Concatenate like IconComponentState.url does. Uri.Builder.path() would replace the
+                        // base URL's path (dropping e.g. /icons), and any mismatch with the render-time URL
+                        // defeats the pre-warm since the image cache is keyed by the exact URL string.
+                        setOf(Uri.parse("${component.baseUrl}/${component.formats.webp}"))
```
tonidero: Maybe we can just fix the previous format? Like:
```
setOf(
   Uri.parse(component.baseUrl)
    .buildUpon()
    .appendPath(component.formats.webp)
    .build()
)
```
I like it more than just string interpolation.

tonidero (approved):
Ohhh 😮‍💨 Thank you so much for fixing! Just a suggestion that I believe might be a bit cleaner but looks great!

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
bilqisium: Thanks @tonidero!

I checked the pros and cons of each method, and unless I'm mistaken, I think `appendPath()` percent-encodes the segment, while the render path (IconComponentState.url) concatenates raw.

It's identical for the current icon set, but if `formats.webp` ever carried a sub-path or reserved char, the pre-warm URL would diverge from the render URL again, and the cache key has to match exactly for the pre-warm. `appendEncodedPath()` keeps the builder style without the encoding.

Happy either way given today's data, it's your call!

tonidero: Good point, I wouldn't expect the `formats.web` to require encoding... So yeah we can use `appendEncodedPath` to keep the same behavior 👍

bilqisium: Great! Fixed in ecada6929bc21502280aab758d8d113bfc6d485f

tonidero (approved):
Looks great, thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/utils/PaywallComponentsImagePreDownloader.kt
bilqisium: Just so you know, I can't merge even if it's approved, so if it's okay, feel free to merge 🙂 

CleanShot 2026-07-03 at 18 05 30@2x

PR #3715, opened 2026-07-06
feat(remote-config): source API base host from remote-config sources

ajpallares:
### Checklist

### Motivation
Main-API requests hardcode their host to `AppConfig.baseURL`, so they can't fail over across the API hosts the backend advertises via remote config. This is the first, behavior-preserving step toward that failover. Mirrors revenuecat/purchases-ios#7123.

### Description
- Gated behind the `usesRemoteConfigAPISources` dangerous setting (off by default): when off, requests keep targeting the static `AppConfig` base URL, so production behavior is unchanged.
- When on, main-API endpoints (`usesAPISources`) resolve their base host from the shared `RemoteConfigSourceProvider`, wired into `HTTPClient`.
- Preserves existing precedence: a set `proxyURL` pins every request, an overridden base URL pins the host, and per-endpoint fallback URLs are untouched.
- No cross-host failover yet: only the primary source is used here; `reportUnhealthy`-driven failover comes in a follow-up PR.

---

> [!NOTE]
> **Medium Risk**
> Touches core networking and request host selection for many API endpoints, but behavior is gated off by default and preserves proxy/override/fallback precedence.
> 
> **Overview**
> Adds optional routing of **main-API** requests to the host from remote config’s API `sources` topic, behind the **`usesRemoteConfigAPISources`** dangerous setting (off by default).
> 
> When enabled, `HTTPClient` resolves the request base URL from a shared `RemoteConfigSourceProvider` for endpoints that opt in via **`Endpoint.usesAPISources`**; diagnostics/events and fallback-host attempts are excluded. **Proxy URLs**, non-default base URLs, and existing per-endpoint fallback URLs are unchanged. Only the **current primary** API source is used—no `reportUnhealthy` failover yet.
> 
> `PurchasesFactory` builds one `RemoteConfigTopicStore` + `DefaultRemoteConfigSourceProvider` for both `HTTPClient` and `RemoteConfigManager` when workflows remote config is on. Unit tests cover host resolution, ETag retries, and endpoint classification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccbced1017a061aa90f92d53abe4e8588b69d3f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -67,10 +67,8 @@ internal class RemoteConfigManager(
     private val dateProvider: DateProvider = DefaultDateProvider(),
     private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.IO),
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
-    private val topicStore: RemoteConfigTopicStore =
-        RemoteConfigTopicStore { diskCache.read()?.topics?.get(it.wireName) },
-    private val sourceProvider: RemoteConfigSourceProvider =
-        DefaultRemoteConfigSourceProvider(topicStore),
+    private val topicStore: RemoteConfigTopicStore,
+    private val sourceProvider: RemoteConfigSourceProvider,
```
ajpallares, line 84: I removed the default values here since we're always passing them (except for tests)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -209,4 +209,36 @@ internal sealed class Endpoint(
...
+            is GetRemoteConfig,
+            PostCreateSupportTicket,
+            PostRedeemWebPurchase,
+            is GetVirtualCurrencies,
+            is GetRewardVerification,
+            ->
+                true
+            PostDiagnostics,
+            PostEvents,
+            is WebBillingGetProducts,
```
tonidero: Hmm I do wonder if `WebBillingGetProducts` should use the sources as well?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -200,12 +202,21 @@ internal class PurchasesFactory(
             val cache = DeviceCache(prefs, apiKey)
 
             val localeProvider = DefaultLocaleProvider()
+
+            val remoteConfigEnabled = BuildConfig.ENABLE_REMOTE_CONFIG && !appConfig.customEntitlementComputation
+            val remoteConfigDiskCache = if (remoteConfigEnabled) RemoteConfigDiskCache(contextForStorage) else null
+            val remoteConfigTopicStore = RemoteConfigTopicStore {
+                remoteConfigDiskCache?.read()?.topics?.get(it.wireName)
+            }
+            val apiSourceProvider = DefaultRemoteConfigSourceProvider(remoteConfigTopicStore)
```
tonidero, line 213: Hmm it's true that now we're creating a bunch of classes we weren't creating before since they were behind the flag... 

I guess it makes sense since the sources mechanism could work independently of the remote config (with the default values), but we should test this well, since it means we will actually be running some of this code in published versions with the flag off. I'm not aware of any issues but still better to be safer 🙏 .

tonidero (approved):
It's looking great! Just one question and also a warning that we should test this change well since it means we might be instantiating some things for the first time without the flag being on, but I think it should be safe from what I can see.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
ajpallares: Yep. You're right. I'll only merge this once I've tested this thoroughly 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
ajpallares: Great catch! I don't see any reason why not. This all comes from iOS [here](https://github.com/RevenueCat/purchases-ios/blob/dd93a4a34fb8d33466dc4fa66e8b22270c550e29/Sources/Networking/WebBillingHTTPRequestPath.swift#L16-L19), where the `HTTPRequest.WebBillingPath` is declared separately from the main `HTTPRequest.Path`, but they both have the same `"https://api.revenuecat.com"` as the base URL. So I think these endpoints should use the sources as well. Changing that to use sources in both platforms! Thanks

ajpallares: Changed it in https://github.com/RevenueCat/purchases-android/pull/3715/commits/bed632eb4ebe306a616a51e898e0163ec35ce666

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -226,7 +230,7 @@ internal class HTTPClient(
             timeoutManager.recordRequestResult(requestResult)
 
             trackHttpRequestPerformedIfNeeded(
-                baseURL,
+                requestBaseURL,
```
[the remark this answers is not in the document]
ajpallares: The test added in https://github.com/RevenueCat/purchases-android/pull/3715/commits/310406d1e64dedd61c85d5a4374a79b116680c91 contradicts this

PR #3722, opened 2026-07-06
fix(remote-config): don't sync a cold read for the previous user on identity change

tonidero:
## Why

On an identity change, `IdentityManager` wiped remote config before caching the new app user ID. A cold `topic()`/`blobData()` read racing that change could trigger an on-demand sync for the **previous** user and, because that request starts after the cache is wiped, persist the wrong user's config into the new user's freshly-wiped cache and return it. This violates the offerings-parity invariant `clearCache()` exists to protect (config must never bleed across users).

The path is currently latent (the `topic()`/`blobData()` facade has no production callers yet), but it's a landmine for the first consumer.

## Fix

`clearCache()` now binds the incoming app user to the drop-epoch atomically, and cold on-demand reads sync for that bound identity rather than the potentially-stale cached ID.

---

> [!NOTE]
> **Medium Risk**
> Touches identity-sensitive remote config sync and locking around `/v1/config`; wrong user binding would leak config across users, but the change is narrowly scoped with targeted race tests.
> 
> **Overview**
> Fixes a race where **remote config could bleed across users** after login, alias, switch, or logout. Identity flows wipe remote config before the device cache always reflects the new app user ID, so a cold `topic()` / `blobData()` read could trigger `/v1/config` for the **previous** user and persist that response into the wiped cache.
> 
> **`RemoteConfigManager.clearCache`** now takes the new `appUserID` and sets **`currentAppUserID`** in the same critical section as the identity **epoch** bump. **`refreshRemoteConfig`** snapshots **`requestAppUserID`** with the epoch under `cacheLock` (`currentAppUserID` wins; the caller’s ID is only a pre–first-identity-change bootstrap). Cold reads still use `appUserIDProvider` only until the first `clearCache`.
> 
> **`IdentityManager`** passes the target user into `remoteConfigManager?.clearCache(...)` on alias, login, switch, and logout. Tests cover the lagging-provider and mid-read identity-change cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53840787f10eaeb63018baf8d785344cc9ffa671. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -132,16 +132,24 @@ internal class RemoteConfigManager(
             return
         }
         val requestEpoch: Int
+        val requestAppUserID: String
         // Acquire the in-flight guard and capture the epoch together under the lock that also guards
         // clearCache()'s epoch bump + guard release. Otherwise an identity change could slip its bump
         // between getAndSet(true) and epoch.get(), stranding this request with a stale epoch: its
         // callbacks would drop on the mismatch and never release the guard, freezing all future syncs.
+        // The user is snapshotted here too, so the (user, epoch) pair the request carries is consistent:
+        // any clearCache() that would change the user also bumps the epoch under this same lock, so it
+        // either lands fully before this capture (we see its new user) or after (the epoch mismatch drops
+        // our response) — never a stale user paired with the post-clear epoch. currentAppUserID (bound by
+        // clearCache) wins; the passed appUserID is only the pre-first-identity-change bootstrap fallback.
+        // Read only in-memory state under the lock — never appUserIDProvider(), which can re-enter clearCache().
         synchronized(cacheLock) {
             if (isRefreshing.getAndSet(true)) {
                 debugLog { "Remote config refresh already in progress. Skipping." }
                 return
             }
             requestEpoch = epoch.get()
+            requestAppUserID = currentAppUserID ?: appUserID
             refreshCompletion = CompletableDeferred()
         }
         val persisted = diskCache.read()
@@ -149,7 +157,7 @@ internal class RemoteConfigManager(
         logRefreshStart(persisted, appInBackground)
         backend.getRemoteConfig(
             appInBackground = appInBackground,
-            appUserID = appUserID,
+            appUserID = requestAppUserID,
             domain = persisted?.domain ?: DEFAULT_DOMAIN,
             // Opaque manifest replayed verbatim; null on the first run when nothing is persisted yet.
             manifest = persisted?.manifest,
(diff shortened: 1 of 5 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -269,15 +285,18 @@ internal class RemoteConfigManager(
...
      */
     private suspend fun awaitConfigForRead() {
         if (awaitInFlightRefresh()) {
             verboseLog { "Cold remote config read waiting on the refresh already in progress." }
             return
         }
         // Nothing in flight: trigger a sync on demand, unless the endpoint is disabled or no user is known yet.
-        val appUserID = appUserIDProvider()?.takeIf { it.isNotBlank() }
+        // Prefer the identity clearCache() bound (see [currentAppUserID]); the provider is only the bootstrap.
+        val appUserID = (currentAppUserID ?: appUserIDProvider())?.takeIf { it.isNotBlank() }
```
rickvdl, line 309: One concern here; we're saving the `appUserID` here already, but before we're incrementing / storing the epoch. Shouldn't these be tied together in a single locking operation? Right now the user ID can still change before that operation happens.

This (Codex) test should demonstrate it
```kotlin
@Test
fun `on-demand sync uses bound user if identity changes after provider reads stale user`() = runTest {
    every { diskCache.read() } returns null

    lateinit var manager: RemoteConfigManager

    manager = readManager(
        appUserIDProvider = {
            // Simulates the race:
            // Device cache/provider is still returning the previous user,
            // but identity clear happens before the refresh epoch is captured.
            manager.clearCache("new-user")

            "old-user"
        },
    )

    val read = launch(UnconfinedTestDispatcher(testScheduler)) {
        manager.topic(RemoteConfigTopic.Workflows)
    }

    verify(exactly = 1) {
        backend.getRemoteConfig(any(), any(), any(), any(), any(), any(), any())
    }

    assertThat(capturedAppUserID).isEqualTo("new-user")

    onSuccess.invoke(null, VerificationResult.VERIFIED)
    assertThat(read.isCompleted).isTrue()
}
```

rickvdl (commented):
Great catch! Just one more edge-case we may have to consider

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Great catch!! I think [this](https://github.com/RevenueCat/purchases-android/pull/3722/commits/53840787f10eaeb63018baf8d785344cc9ffa671) should solve that 🙏

rickvdl (approved):
Thank you!

PR #3725, opened 2026-07-06
refactor(remote-config): Add `mergeItemsBlobData` to `RemoteConfigManager`

tonidero:
## Why

Sometimes a single logical object in a remote-config topic needs data spread across multiple items. Previously a consumer would have to fetch each item's blob one by one. This adds `mergeItemsBlobData`, which fetches the requested items' blobs in parallel, merges them into one JSON object, and decodes it into a single `T`.

---

> [!NOTE]
> **Low Risk**
> Additive read-side API on remote config that reuses existing blob resolution and sync behavior; no changes to persist/sync commit paths.
> 
> **Overview**
> Adds **`mergeItemsBlobData`** on `RemoteConfigManager` so callers can load several topic items in one shot instead of repeated **`blobData`** calls.
> 
> The API resolves each requested item’s blob **in parallel** (same rules as **`blobData`**: on-demand sync, blob fetch deduping), nests each blob’s JSON under its **item key** (e.g. `{"wf1":{...},"wf2":{...}}`), then decodes into a single `@Serializable` type—or returns **`null`** with logging when anything is missing, non-JSON, or fails deserialization (**all-or-nothing**). Empty key lists and the existing 4xx **disabled** kill-switch also yield **`null`** without I/O.
> 
> Tests cover the happy path, failure modes, disabled endpoint behavior, and a single shared on-demand `/v1/config` sync when multiple uncached items are merged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 94ee12925e519e814f68a30b8bb69144c0f4d928. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -19,8 +19,10 @@ import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.cancelChildren
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import kotlinx.serialization.SerializationException
@@ -370,6 +372,56 @@ internal class RemoteConfigManager(
         transform: (ByteArray) -> T?,
     ): T? = withContext(ioDispatcher) {
         resolveBlobBytes(topic, itemKey)?.let(transform)
+    }
+
+    /**
+     * The resolved blob payloads for [itemKeys] in [topic], each parsed from JSON into [T], keyed by item key.
+     * The batch counterpart of the single-key reified overload: every requested key is present in the result,
+     * mapping to `null` when that item is unknown, has no `blob_ref`, its blob can't be resolved, or its bytes
+     * don't deserialize into [T]. Duplicate keys collapse to one entry; an empty [itemKeys] returns an empty
+     * map. [T] must be a concrete `@Serializable` type; parsing uses the shared [JsonProvider.defaultJson]. For
+     * non-JSON payloads use the `transform` batch overload, which documents the resolution and waiting rules.
+     */
+    suspend inline fun <reified T> blobData(topic: RemoteConfigTopic, itemKeys: Collection<String>): Map<String, T?> =
+        blobData(topic, itemKeys) { bytes ->
+            try {
+                JsonProvider.defaultJson.decodeFromString<T>(bytes.decodeToString())
+            } catch (e: SerializationException) {
+                errorLog(e) { "Failed to parse remote config blob as JSON." }
+                null
+            }
+        }
+
+    /**
+     * Resolves the blob payload bytes for each key in [itemKeys] within [topic] **concurrently** and maps each
+     * through [transform], keyed by item key. The batch counterpart of the single-key `transform` overload:
+     * every requested key is present in the result, mapping to `null` when that item is unknown or its blob
+     * can't be resolved. Duplicate keys collapse to one entry; an empty [itemKeys] returns an empty map.
+     *
+     * Each key resolves through the same path as the single-key overload (see its KDoc for the `blob_ref`,
+     * on-demand fetch, and waiting rules), so the per-item semantics are identical — this only fans them out.
+     * The fan-out is safe: a shared in-flight `/v1/config` refresh is deduped across all keys, and the blob
+     * fetcher dedupes concurrent downloads of the same ref. When the endpoint is [isDisabled] (the 4xx session
+     * kill-switch) every key resolves to `null` without any read. Runs on [ioDispatcher].
+     */
+    suspend fun <T> blobData(
+        topic: RemoteConfigTopic,
+        itemKeys: Collection<String>,
+        transform: (ByteArray) -> T?,
+    ): Map<String, T?> = withContext(ioDispatcher) {
+        coroutineScope {
+            itemKeys.distinct()
+                .associateWith { key -> async { resolveBlobBytes(topic, key)?.let(transform) } }
+                .mapValues { (_, deferred) -> deferred.await() }
+        }.also { resolved ->
+            val unresolved = resolved.filterValues { it == null }.keys
+            if (unresolved.isNotEmpty()) {
+                warnLog {
+                    "Could not resolve remote config blob(s) for ${unresolved.size} of ${resolved.size} " +
+                        "requested item(s) in topic '${topic.wireName}': $unresolved."
+                }
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -372,6 +377,73 @@ internal class RemoteConfigManager(
         resolveBlobBytes(topic, itemKey)?.let(transform)
     }
 
+    /**
+     * Resolves the blobs for every key in [itemKeys] within [topic] **concurrently**, merges each (parsed as a
+     * JSON object) into one combined JSON object, and decodes that into a single [T]. Use this when a single
+     * logical object is assembled from data spread across several items in a topic (e.g. item `a` carries
+     * `{"x":...}` and item `b` carries `{"y":...}`, together decoding into `T(x, y)`).
+     *
+     * This is **all-or-nothing**: if any requested item is unknown, has no `blob_ref`, or its blob can't be
```
tonidero, line 387: This was an assumption on my side. Lmk if you think we should proceed without some of the blobs, or if some of the blobs will be optional. Should probably modify this if so. @vegaro

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -372,6 +377,73 @@ internal class RemoteConfigManager(
...
+     * doesn't deserialize into [T]. On a key collision the later item in [itemKeys] wins. Duplicate keys are
+     * de-duplicated; [T] must be a concrete `@Serializable` type and parsing uses [JsonProvider.defaultJson].
+     *
+     * Each item resolves through the same path as the single-key [blobData] (see its KDoc for the `blob_ref`,
+     * on-demand fetch, and waiting rules) — this only fans them out. The fan-out is safe: a shared in-flight
+     * `/v1/config` refresh is deduped across all keys, and the blob fetcher dedupes concurrent downloads of the
+     * same ref. When the endpoint is [isDisabled] (the 4xx session kill-switch) every item resolves to `null`,
+     * so the call returns `null` without any read. Runs on [ioDispatcher].
+     */
+    suspend inline fun <reified T> mergeItemsBlobData(topic: RemoteConfigTopic, itemKeys: Collection<String>): T? {
```
tonidero: Lmk what you think of this API @vegaro @rickvdl . I think usage now should be something like:
```
mergeItemsBlobData<UiConfig>(RemoteConfigTopic.UiConfig, listOf(....))
```

purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManagerTest.kt
```diff
@@ -1093,6 +1094,194 @@ class RemoteConfigManagerTest {
...
+            every { blobStore.read(REF_TAMPERED) } returns """{"value":"y"}""".toByteArray()
+
+            val result = readManager()
+                .mergeItemsBlobData<MergedBlob>(RemoteConfigTopic.Workflows, listOf("wf1", "wf2"))
+
+            assertThat(result).isEqualTo(MergedBlob(wf1 = Section("x"), wf2 = Section("y")))
+        }
+
+    @Test
+    fun `mergeItemsBlobData returns null when an item's blob cannot be fetched`() = runTest {
```
ajpallares: Isn't this test redundant with the (more complete, I think) test `fun `mergeItemsBlobData warns and returns null when one or more items cannot be resolved`()` below?

ajpallares (approved):
I think it makes sense! 

Just a small question about a test, but nothing blocking

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -372,6 +377,89 @@ internal class RemoteConfigManager(
...
+
+    /**
+     * Resolves every item in [itemKeys] concurrently and builds a JSON object keyed by item key, each mapping
+     * to that item's parsed blob JSON, or `null` if any item can't be resolved or any resolved blob isn't valid
+     * JSON. Assumes it is already running on [ioDispatcher] (its only caller wraps it), so it doesn't switch
+     * context itself.
+     */
+    @Suppress("ReturnCount")
+    private suspend fun mergedBlobObject(topic: RemoteConfigTopic, itemKeys: Collection<String>): JsonObject? {
+        if (disabled) {
```
rickvdl, line 432: Should we also early exit here (and return null instead of an object) when the itemKeys list is empty?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Hmm on one hand I think this makes sense, ~~on the other hand (thinking about iOS), you could also say this is up to the object's decoder to decide? If it can live with one object missing, it should be able to construct the object?~~

Actually I changed my mind here, I think there's a difference between 'the value is null' or the value failed to fetch. So we probably shouldn't just treat it as the same. I think your tradeoff is the best here.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -372,6 +377,89 @@ internal class RemoteConfigManager(
...
+    /**
+     * Resolves the blobs for every key in [itemKeys] within [topic] **concurrently**, builds a single JSON
+     * object mapping **each item key to that item's parsed blob JSON**, and decodes it into a single [T]. Use
+     * this to assemble one object from several items in a topic: given items `wf1 -> {"a":...}` and
+     * `wf2 -> {"b":...}`, the merged object is `{"wf1": {"a":...}, "wf2": {"b":...}}`, so [T] declares a field
+     * per item key.
+     *
+     * This is **all-or-nothing**: if any requested item is unknown, has no `blob_ref`, or its blob can't be
+     * resolved, the call returns `null` (a partial object is never produced) and warn-logs the missing keys.
+     * It also returns `null` if any resolved blob isn't valid JSON, or the merged object doesn't deserialize
```
rickvdl, line 389: I think we're missing tests for these `null` cases

rickvdl (commented):
Few comments but looking very nice overall!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3725/commits/94ee12925e519e814f68a30b8bb69144c0f4d928

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Good one, added in https://github.com/RevenueCat/purchases-android/pull/3725/commits/94ee12925e519e814f68a30b8bb69144c0f4d928

purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManagerTest.kt
tonidero: Good catch! Removed in https://github.com/RevenueCat/purchases-android/pull/3725/commits/58bc18de3e1ced00ae721a1854c38769f4b0d28a

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
vegaro: I agree with you here, I don't see any reason to do something different for ui config for example

rickvdl (approved):
Awesome! Great API addition 💪

vegaro (approved):
tested it in https://github.com/RevenueCat/purchases-android/pull/3711/files#diff-5ce8eaeb3a3161cdb699340fd0e0a87071e57f22006a9ac47a89ba4bf61e97d8R26 and it looks great

PR #3726, opened 2026-07-06
Experimental IAM configuration option

davedelong:
### Checklist

### Motivation
Adds a new experimental configuration option that will be used as part of ongoing authentication work.

---

> [!NOTE]
> **Medium Risk**
> Touches authentication-related configuration, but behavior is gated behind an internal API with default disabled and no SDK wiring in this diff.
> 
> **Overview**
> Adds an experimental **`iamEnabled`** option to **`PurchasesConfiguration`** (default **false**), exposed on the builder as **`iamEnabled(Boolean)`** behind **`@InternalRevenueCatAPI`**. The flag is stored as an internal property, included in **`copy`**, **`equals`**, and **`hashCode`**, and documented as preferring token-based user sessions for server communication.
> 
> Unit tests assert the default is false and that the builder can set it to true.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4afa25b686bbcf9cec40da7159f9740f2f1471d0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -321,6 +330,18 @@ public open class PurchasesConfiguration(builder: Builder) {
...
         }
 
+        /**
+         * Enabling this instructs the SDK to prefer using token-based user sessions
+         * for communicating with the server. Default is disabled.
+         *
+         * @warning This function is marked as [ExperimentalPreviewRevenueCatPurchasesAPI] and may change in the
+         * future.
+         */
+        @ExperimentalPreviewRevenueCatPurchasesAPI
```
tonidero: I would probably keep this at least as `@InternalRevenueCatAPI` until it's functional. FWIW:
- `InternalRevenueCatAPI` means it's not meant to use by outside developers
- `ExperimentalPreviewRevenueCatPurchasesAPI` means developers can use it but the API and/or behaviors are subject to change. 

So if something is not functional, I don't think we should expose it as experimental either.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesConfiguration.kt
```diff
@@ -321,6 +330,18 @@ public open class PurchasesConfiguration(builder: Builder) {
             this.galaxyBillingMode = galaxyBillingMode
         }
 
+        /**
+         * Enabling this instructs the SDK to prefer using token-based user sessions
+         * for communicating with the server. Default is disabled.
+         *
+         * @warning This function is marked as [ExperimentalPreviewRevenueCatPurchasesAPI] and may change in the
```
tonidero: I don't think this comment is needed, the annotation itself should be self explanatory

tonidero (commented):
Mostly, I just think this shouldn't be made available to consumers until it's functional so I would suggest a change in annotation. Other than that, looks good!

JZDesign (approved):

PR #3728, opened 2026-07-07
Drop `ENABLE_REMOTE_CONFIG`, and gate the config layer on useWorkflows

vegaro:
Workflows are the only consumer of `/v1/config` right now, and turning them on already implies needing the config layer, so there's no reason to keep a separate `ENABLE_REMOTE_CONFIG` build flag in sync by hand.

Drops the flag (`buildConfigField` + `local.properties.example` entry) and gates `RemoteConfigManager` on `useWorkflows` alone. Still off in the `customEntitlementComputation` flavor, same as before.

---

> [!NOTE]
> **Low Risk**
> Small wiring change with no auth or payment logic; the only behavioral shift is dropping the unused standalone remote-config flag when workflows are off.
> 
> **Overview**
> Removes the standalone **`ENABLE_REMOTE_CONFIG`** debug/build path so the `/v1/config` stack is no longer toggled separately from workflows.
> 
> **`RemoteConfigManager`** is now created only when **`appConfig.useWorkflows`** is on (still skipped for **`customEntitlementComputation`**). The **`BuildConfig.ENABLE_REMOTE_CONFIG`** field, **`local.properties.example`** entry, and related **`PurchasesFactory`** import are deleted.
> 
> Apps that previously turned on remote config via the build flag without **`useWorkflows`** will no longer instantiate the config layer; that path was test-only and workflows are the sole consumer today.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 245ca3908223258a1d7bb096a04cbd691040c3aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/local.properties.example
+++ b/local.properties.example
@@ -74,11 +74,3 @@
 ## Debug flags
 ## ──────────────────────────────────────────────
 #ENABLE_EXTRA_REQUEST_LOGGING=true
-
-## ──────────────────────────────────────────────
-## Enable Remote Config (test-only)
-## When true, the SDK instantiates the /v1/config manager and refreshes it on app
-## foreground (staleness-gated). The response is not consumed yet; this only exercises
-## the request/parse/persist path from test apps ahead of the production wiring.
-## ──────────────────────────────────────────────
-#ENABLE_REMOTE_CONFIG=true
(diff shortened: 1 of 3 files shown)


tonidero (approved):
Thank you!!

PR #3732, opened 2026-07-08
refactor(remote-config): Add `awaitTopicAndPrefetchBlobsReady` to `RemoteConfigManager`

tonidero:
### Description

Adds `RemoteConfigManager.awaitTopicAndPrefetchBlobsReady(topic)`.

`topic()` guarantees only that the config request has committed and any **inlined** blobs are cached. It does **not** wait for **non-inlined** blobs the config marks `prefetch: true` — those are enqueued fire-and-forget on the fetcher's LOW-priority queue and may still be downloading when `topic()` returns.

The new method reads the committed topic, then waits for that topic's `prefetch`-marked blobs to finish caching (joining/boosting the in-flight prefetch; already-cached inline blobs return immediately; failed downloads don't block), then returns the `ConfigTopic`.

`WorkflowsConfigProvider.awaitReady()` now uses it, so "ready" includes prefetch-marked workflow bodies being downloaded, not just metadata + inline.

Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/7157

---

> [!NOTE]
> **Medium Risk**
> Changes when offerings/workflows are considered "ready" and adds identity-change handling during long blob waits; behavior is well-tested but affects paywall/offering timing.
> 
> **Overview**
> Adds **`RemoteConfigManager.awaitTopicAndPrefetchBlobsReady`**, which returns a committed topic only after **`prefetch`-marked** item blobs are cached (via `blobFetcher.ensureDownloaded`), not just after config metadata commits. **`topic()`** is refactored to share a private **`committedTopic`** helper for the read/wait-on-miss path.
> 
> After the blob wait, the method **re-reads** the topic and loops if identity/cache changed mid-wait, so callers don't return a stale snapshot.
> 
> **`WorkflowsConfigProvider.awaitReady()`** now uses the new API instead of **`topic()`**, restoring parity with the old workflows list gate: offerings success can wait for prefetch workflow bodies, not only topic metadata and inlined blobs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 64de20bdae95453cb27367f626b44c8d30efe8bb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -319,16 +319,51 @@ internal class RemoteConfigManager(
...
+            // Only the prefetch-marked items matter here; on-demand items are resolved lazily by blobData reads.
+            val prefetchRefs = configTopic.values
+                .filter { it.prefetch }
+                .mapNotNull { it.blobRef }
+            if (prefetchRefs.isNotEmpty()) {
+                verboseLog { "Awaiting ${prefetchRefs.size} prefetch blob(s) for topic '${topic.wireName}'." }
+                // Joins/boosts any in-flight LOW-priority prefetch; already-cached (inlined) refs return at once.
+                blobFetcher.ensureDownloaded(prefetchRefs)
+            }
+            configTopic
```
facumenzella: Quick question: since this grabs the topic once and waits on its prefetch blobs, what happens if an identity change (`clearCache()`) fires while we're mid-wait? Looks like we'd still return the original snapshot even though it no longer matches the current user.

iOS hit the same thing and handled it in #7157 by re-reading the topic after the wait and looping if it changed:

```kotlin
suspend fun awaitTopicAndPrefetchBlobsReady(topic: RemoteConfigTopic): ConfigTopic? =
    withContext(ioDispatcher) {
        var committed = committedTopic(topic) ?: return@withContext null
        while (true) {
            val prefetchRefs = committed.values.filter { it.prefetch }.mapNotNull { it.blobRef }
            if (prefetchRefs.isNotEmpty()) blobFetcher.ensureDownloaded(prefetchRefs)
            val latest = committedTopic(topic) ?: return@withContext null
            if (latest == committed) return@withContext committed
            committed = latest
        }
    }
```

`awaitReady()` throws away the return value today so it's not urgent, but is this worth matching for parity, or is there a reason Android doesnt need the retry?

facumenzella:
Separate thought: 

the `awaitReady()` → `onPaywallConfigReady()` chain has no timeout wrapping the blob wait, so a slow/failing source can stretch `getOfferings()` latency by however long the source-failover loop takes. Checked iOS for comparison since this mirrors #7157: same gap there too, `URLSessionRemoteConfigBlobDownloader` uses `URLSession.shared` with no timeout override, so it falls back to the default 60s `timeoutIntervalForRequest` per attempt (looser than Android's 5s connect + 5s read), and there's no `withTimeout` around `awaitTopicAndPrefetchBlobsReady`/`deliverWhenConfigReady` either.

So not something this PR introduces, and not blocking, just flagging since both platforms are now waiting on network blob downloads (not just a config read) on the `getOfferings` critical path. Might be worth a shared follow-up to cap the wait on both.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Hmm you're right, we probably need that, good catch!

tonidero:
> So not something this PR introduces, and not blocking, just flagging since both platforms are now waiting on network blob downloads (not just a config read) on the getOfferings critical path. Might be worth a shared follow-up to cap the wait on both.

Good call, let's chat about it over Slack...

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Added fix for this in https://github.com/RevenueCat/purchases-android/pull/3732/commits/d32f1ccdba7326af1ca63acf8041b2a8fd683463

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -348,16 +348,67 @@ internal class RemoteConfigManager(
...
+                if (prefetchRefs.isNotEmpty()) {
+                    verboseLog { "Awaiting ${prefetchRefs.size} prefetch blob(s) for topic '${topic.wireName}'." }
+                    // Joins/boosts any in-flight LOW-priority prefetch; already-cached (inlined) refs return at once.
+                    blobFetcher.ensureDownloaded(prefetchRefs)
+                }
+                // Re-read: if the committed topic is unchanged the snapshot we waited on is still current, so return
+                // it. If it changed under us (identity change wiped it → null, or a newer sync committed a different
+                // topic) loop: a null exits with null, a different topic re-awaits its own prefetch blobs.
+                val latest = committedTopic(topic)
+                if (latest == committed) break
```
AlvaroBrey, line 390: Should we be using `contentHash` for this comparison, rather than the topic instance?

ajpallares (approved):
Looks good! No comments other than Alvaro's question

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: It's indeed a good question. So since these are data classes it should mostly work. We mostly added `contentHash` for the sources topic to keep a reference of whether it changed... But I believe we could also compare the contents, just would need to keep a reference to the original sources in `DefaultRemoteConfigSourceProvider` I think... But yeah, I think it should work either way, so might leave it as is for now. Lmk if you think otherwise and can change it in follow-up PRs!

AlvaroBrey: If it works either way I'd leave the equals, since it's less expensive 🚢

AlvaroBrey (approved):

PR #3733, opened 2026-07-08
Handle button sheet destination without an inline sheet

MonikaMateska:
### Checklist

## Motivation

While working on bringing Astra to our iOS app, I hit a draft paywall that refused to parse on iOS but rendered fine on the web dashboard, the same applies on Android as well.

A `navigate_to` action with `destination: "sheet"` can arrive without an inline `sheet` object (the backend types it as optional and drops null fields). The `ActionSerializer` required it via `checkNotNull(sheet)`, so a missing sheet
threw `IllegalStateException`. Because that runs during deserialization, one bad button took down the whole paywall parse. 
The web SDK types it as `sheet?: SheetProps | null` and renders a missing sheet as a button with a no-op tap.

## Description

This aligns Android with the web contract:

- `ButtonComponent.Destination.Sheet` fields are now optional (`stack` nullable,  `id` defaulted), so a content-less sheet is representable, and the deserializer  maps a missing sheet to an empty `Sheet` instead of `checkNotNull`-throwing.
- A `navigate_to`/`sheet` destination with no `stack` maps to a new `ButtonComponentStyle.Action.NoOp` in `StyleFactory`, the button renders but its tap does nothing, rather than being routed to `Unknown` (which hides the
  button). Sheets with content are unchanged.
- `NoOp` is handled across the action `when`s (`toPaywallAction` -> null tap, interaction mapping, purchase-related check), and `OfferingToStateMapper`  handles the now-nullable sheet stack.

Added a deserialization test covering a `sheet` destination with no inline `sheet` (decodes to an empty `Sheet` instead of throwing).

---

> [!NOTE]
> **Low Risk**
> Localized paywall JSON/UI behavior change for an edge-case button action; no purchase, auth, or billing flow changes.
> 
> **Overview**
> Fixes **whole-paywall parse failures** when a button uses `navigate_to` with `destination: "sheet"` but the optional inline `sheet` object is absent (payloads that already work on web).
> 
> **Deserialization:** `Destination.Sheet` now allows a nullable `stack` and defaults on other fields; `ActionSerializer` maps a missing inline sheet to an empty `Sheet()` instead of throwing during decode.
> 
> **Rendering:** `StyleFactory` turns sheet destinations with no `stack` into a new **`NoOp`** button action so the control still shows but taps do nothing (web parity). `NoOp` is wired through paywall action mapping, interaction analytics, and unsupported-condition walks; sheet destinations with content are unchanged.
> 
> Adds a unit test that `destination: "sheet"` without an inline `sheet` decodes successfully.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 915f625f86b6c125d931a1884173e20d3be32858. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
@@ -830,10 +830,14 @@ internal class StyleFactory(
                 destination.method,
                 componentInteractionValue = "navigate_to_url",
             )
-            is ButtonComponent.Destination.Sheet ->
-                // A content-less sheet is handled as a no-op before we get here, so `stack` is non-null;
-                // the elvis is only to satisfy the type system.
-                destination.stack?.let { rawStack ->
+            is ButtonComponent.Destination.Sheet -> {
+                val rawStack = destination.stack
+                if (rawStack == null) {
+                    // Unreachable: a content-less sheet is mapped to NoOp before reaching here. Log in
+                    // case that invariant ever breaks, and hide the button rather than crash.
+                    Logger.e("Sheet destination reached convertDestination without a stack.")
+                    Result.Success(null)
+                } else {
                     createStackComponentStyle(rawStack)
                         .map { it.applyBottomWindowInsetsIfNecessary(shouldApply = true) }
                         .map { it.applyHorizontalWindowInsetsIfNecessary(shouldApply = true) }
@@ -846,7 +850,8 @@ internal class StyleFactory(
                                 size = destination.size,
                             )
                         }
-                } ?: Result.Success(null)
+                }
+            }
             // Returning null here, which will result in this button being hidden.
             is ButtonComponent.Destination.Unknown,
             -> Result.Success(null)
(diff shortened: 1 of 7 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -823,18 +831,22 @@ internal class StyleFactory(
...
+                        .map { stackComponentStyle ->
+                            ButtonComponentStyle.Action.NavigateTo.Destination.Sheet(
+                                id = destination.id,
+                                name = destination.name,
+                                stack = stackComponentStyle,
+                                backgroundBlur = destination.backgroundBlur,
+                                size = destination.size,
+                            )
+                        }
+                } ?: Result.Success(null)
```
tonidero: Nitpick but could be good to log an error here where it should never actually arrive.

tonidero (approved):
Makes sense!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
MonikaMateska: Good point, adding it now

PR #3739, opened 2026-07-09
fix(remote-config): require UI config for workflows

rickvdl:
## Description

iOS requires the `UIConfig` to be successfully loaded in order for workflows to be presented. Android was more flexible here in the sense that it used to fall back to an empty `UIConfig`. However this could potentially greatly alter the look / feel / workings of workflows that we've decided that it indeed should be required.

This PR ensures the `UIConfig` is available, or throws an `PurchasesErrorCode.UnknownError` with message ` "UI config is unavailable."`

---

> [!NOTE]
> **Medium Risk**
> Changes paywall/workflow presentation behavior: missing or broken remote UI config now fails workflow fetch instead of rendering with defaults, which may surface new errors in production but avoids incorrect paywall styling.
> 
> **Overview**
> Aligns Android with iOS by treating remote **UI config** as mandatory for workflows instead of silently using an empty default when merge/decode fails.
> 
> **`UiConfigProvider`** now returns `null` when the four-part `ui_config` merge is missing or invalid; it no longer substitutes `UiConfig()`. **`PurchasesOrchestrator.getUiConfig`** throws `PurchasesErrorCode.UnknownError` with `"UI config is unavailable."` when the provider returns null. **`WorkflowManager.getWorkflow`** loads UI config after resolving the workflow and calls **`onError`** (without asset prewarm) if config is null or load throws; **`CancellationException`** still propagates with no callback. The **`getOfferings`** readiness gate (`onPaywallConfigReady`) stays best-effort so offerings are not blocked, but docs note the render path must error if config is still missing at presentation time.
> 
> Tests cover null provider results, workflow errors, cancellation, and unchanged gate completion when UI config is unavailable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52cd759249e0c2b24d160c8bc59c32fbb9b0e20d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -629,7 +629,12 @@ internal class PurchasesOrchestrator(
             }
             throw PurchasesException(PurchasesError(PurchasesErrorCode.ConfigurationError, message))
         }
-        return provider.getUiConfig()
+        return provider.getUiConfig() ?: throw PurchasesException(
+            PurchasesError(
+                PurchasesErrorCode.UnknownError,
+                "UI config is unavailable.",
+            ),
+        )
     }
 
     fun getProducts(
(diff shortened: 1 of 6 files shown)


tonidero (approved):
Thank you for fixing this!! 🙇

PR #3742, opened 2026-07-09
Fix ui_config font/color decoding bug

vegaro:
`RemoteConfigManager` decoded merged `ui_config` blobs with `JsonProvider.defaultJson`, whose `classDiscriminator` is overridden to `"discriminator"` for `BackendEvent`. Fonts and colors send their discriminator as `"type"`, so any paywall using custom fonts or colors silently fell back to a default (fontless/colorless) `UiConfig` instead of decoding correctly. 

This PR switches both `blobData`/`mergeItemsBlobData` to `JsonTools.json`, the instance the legacy offerings-embedded `ui_config` path already uses correctly, and adds a regression test that decodes a real font payload through the actual merge path.

---

> [!NOTE]
> **Medium Risk**
> Changes JSON decoding for all remote-config blob merges (including ui_config); wrong Json instance would affect every paywall using custom fonts/colors, but the fix is narrowly scoped and covered by a regression test.
> 
> **Overview**
> Fixes **silent loss of custom paywall fonts and colors** when `ui_config` is loaded from remote config merges: `RemoteConfigManager` now decodes single-blob and merged-blob JSON with **`JsonTools.json`** instead of **`JsonProvider.defaultJson`**.
> 
> `defaultJson` sets `classDiscriminator` to `"discriminator"` for `BackendEvent`, but backend payloads (e.g. `FontInfo`) use **`"type"`** as the polymorphic key—so fonts/colors failed to deserialize and callers often saw a default, empty-looking `UiConfig`.
> 
> KDoc on `blobData` / `mergeItemsBlobData` documents why the shared default instance must not be used here (aligned with the legacy offerings `ui_config` path). A **regression test** in `UiConfigProviderTest` asserts a real-shaped font payload decodes through the merge path into `UiConfig.AppConfig.FontsConfig.FontInfo.Name`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce939376586b824defb4d6a1ba0d3bc74f053f99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -1,6 +1,7 @@
 package com.revenuecat.purchases.common.remoteconfig
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.JsonTools
 import com.revenuecat.purchases.PurchasesError
 import com.revenuecat.purchases.common.Backend
 import com.revenuecat.purchases.common.DateProvider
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -419,13 +420,15 @@ internal class RemoteConfigManager(
...
+     * must be a concrete `@Serializable` type; parsing uses [JsonTools.json] (not [JsonProvider.defaultJson],
+     * whose `classDiscriminator` is overridden for [com.revenuecat.purchases.common.events.BackendEvent] and
+     * would break any topic payload relying on the default `type` discriminator, e.g. paywall components). For
+     * non-JSON payloads use the `transform` overload, which also documents the resolution and waiting rules.
      */
     suspend inline fun <reified T> blobData(topic: RemoteConfigTopic, itemKey: String): T? =
         blobData(topic, itemKey) { bytes ->
             try {
-                JsonProvider.defaultJson.decodeFromString<T>(bytes.decodeToString())
+                JsonTools.json.decodeFromString<T>(bytes.decodeToString())
```
tonidero, line 431: Note this means we won't be able to deserialize the classes the `JsonProvider` is able to... But that's probably ok for now 👍

purchases/src/test/java/com/revenuecat/purchases/common/uiconfig/UiConfigProviderTest.kt
```diff
@@ -89,6 +91,51 @@ internal class UiConfigProviderTest {
...
 
+    @Test
+    fun `getUiConfig decodes a font whose discriminator uses the real backend key`() = runTest {
+        // Regression test: the backend sends FontInfo's sealed-type discriminator under "type" (as in this
+        // real payload shape). The merge decode must use a Json instance that resolves that key correctly
+        // (JsonTools.json), not JsonProvider.defaultJson, whose classDiscriminator is overridden to
+        // "discriminator" for BackendEvent and would silently drop the font instead of decoding it.
+        stubMergedRead(
+            buildJsonObject {
+                putJsonObject("app") {
```
tonidero, line 102: Hmm could argue that we could just hardcode a json string... But it's ok not a problem to keep it like this :)

tonidero (approved):

purchases/src/test/java/com/revenuecat/purchases/common/uiconfig/UiConfigProviderTest.kt
vegaro: looks like it's a stablished pattern in this file. I think like you do haha

PR #3743, opened 2026-07-09
test: add offerings startup perf tester app

vegaro:
New `test-apps:perftester` app to measure how long `configure()` + `getOfferings()` takes at startup, baseline vs workflows. It's the realistic counterpart to the instrumentation harness: a real app, a button press, timestamps around the two calls, one JSON line appended per press to `results.jsonl` on device.

Why an app and not just the instrumentation test: the harness runs in-process with mocked billing and a simulated lifecycle, so it can't show real process cold-start ordering. This can. The `had_*` cache flags in each row say whether a sample was cold or warm, so run order doesn't have to be tracked separately.

How it works:
- Toggle picks the arm: baseline or `DangerousSettings.forWorkflows()`.
- "Configure + getOfferings" runs one measured cycle: `close()` if configured (so `configure()` doesn't short-circuit on an equal config), snapshot the cache flags, `configure()`, `getOfferings()`, then on the workflows arm `ui_config` + first workflow body readiness.
- "Clear SDK caches" wipes the SDK prefs/dirs for a cold sample without reinstalling.
- `-e autorun true -e workflows true` drives the same cycle for scripted `am start` loops.

A Test Store key is recommended (`PERF_TESTER_API_KEY` in `local.properties`): products resolve through the simulated store, no Play setup, and billing is the same in both arms. No changes to `:purchases`.

Known limitation: the render-readiness `withTimeout` can't actually bound `awaitGetWorkflow`/`awaitGetUiConfig`, because those SDK APIs use non-cancellable `suspendCoroutine`. A hung fetch on the workflows arm keeps the cycle waiting past the guard. The `getOfferings` measurement itself is properly bounded. Left as-is for now, a proper fix needs a detached-async workaround.

## Verification
- [x] :test-apps:perftester:testDebugUnitTest (3/3)
- [x] :test-apps:perftester:assembleDebug
- [x] detektAll
- [ ] on-device measurement runs: needs a Test Store key. The app builds, installs, launches and renders clean on a device, but I couldn't run real measurements here.


vegaro:
Added post-readiness memory measurements to this perftester PR. Each JSONL row now records before/after/delta RSS, RSS anon/file breakdown, and Java/native heap. The after sample is taken only after getOfferings plus the existing ui_config/workflow readiness sequence completes, so this measures retained memory after workflow loading. It is intentionally not a peak-memory metric; RSS values are read from /proc/self/status and are reported in KiB. Unit tests, assembleDebug, and detekt passed.

PR #3745, opened 2026-07-09
test: add on-demand getOfferings benchmark (remote config off vs on)

tonidero:
## Description

Adds an on-demand instrumentation benchmark that measures the median wall-clock time of the full `Purchases.getOfferings()` API on a **cold install**, comparing **remote config disabled vs enabled**.

### How it works
- Configures the real SDK through the public `Purchases.configure(PurchasesConfiguration)` path with a `test_`-prefixed key, so `OfferingsFactory` resolves products via the simulated store (`SimulatedStoreBillingWrapper`) with no Google Play Billing connection required.
- Enabling remote config via `DangerousSettings.forWorkflows()` makes `getOfferings()` additionally sync `/v1/config` and block on the `WorkflowManager.onPaywallConfigReady` gate. Remote config does **not** reroute the offerings data itself today, so the "enabled" scenario measures the **config-layer overhead added to getOfferings**.
- Each cold iteration: `resetSingleton()` → wipe the four RevenueCat `SharedPreferences` files + `noBackupFilesDir/RevenueCat` (remote-config state + blobs) → `configure()` fresh (anonymous user) → time `getOfferings` to its callback.
- Logs one line per run plus a `median/p90/min/max/mean` summary to logcat under tag `GetOfferingsBenchmark`. No pass/fail threshold.

### Running it
`@Ignore`d by default (matches the existing `BlurBenchmarkTest` convention) so it never runs in normal/CI suites. To measure, provide a test-store key and run explicitly:

\`\`\`bash
./gradlew :purchases:connectedDefaultsDebugAndroidTest \
  -PTEST_STORE_REVENUECAT_API_KEY=test_XXXX \
  -PTEST_PACKAGE_FILTER=com.revenuecat.purchases.benchmark
\`\`\`

(remove the class-level `@Ignore`, or run the method directly from the IDE). Without the key it self-skips.

> **Note:** for the "enabled" median to be meaningful, the `test_` key must have offerings with test-store products and `workflows`/`ui_config` topics provisioned; otherwise the readiness gate mostly no-ops and the overhead is understated.

### DIFF
--- a/purchases/build.gradle.kts
+++ b/purchases/build.gradle.kts
@@ -96,6 +96,11 @@ android {
         // Shared
         testInstrumentationRunnerArguments["TEST_PROXY_URL"] = resolveProperty("TEST_PROXY_URL")
 
+        // Test store key (test_-prefixed) used by the getOfferings benchmark. No backend environment
+        // of its own; it hits production and resolves products via the simulated store.
+        testInstrumentationRunnerArguments["TEST_STORE_REVENUECAT_API_KEY"] =
+            resolveProperty("TEST_STORE_REVENUECAT_API_KEY")
+
         // Optional package filter for running a subset of tests (used by CI).
         // e.g. -PTEST_PACKAGE_FILTER=com.revenuecat.purchases.integration.production
         val testPackageFilter = resolveProperty("TEST_PACKAGE_FILTER")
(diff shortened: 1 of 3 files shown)


tonidero:
Closing this for now, since I think it doesn't provide much value without actual assertions.

PR #3746, opened 2026-07-09
Re-tier HTTP request timeouts around per-host memory

ajpallares:
### Checklist
- [x] Unit tests

### Motivation
Now that a single request can fail over across several hosts, the old timeout handling doesn't hold up. We only remember the most recent timeout in one global slot, so one slow host shortens timeouts for unrelated hosts (and a success on any host resets them all). The timeouts are also the same regardless of the request, so we wait too long on a host that's down before failing over to the next one.

### Description
New per-attempt timeout tiers:

| Request kind | Base | Reduced (source timed out in last 10 min) |
|---|---|---|
| Main-source, endpoint has NO fallback-URL support | 15s | 5s |
| Main-source, endpoint HAS fallback-URL support | 5s | 2s |
| Fallback-host request | 30s | n/a (flat) |
| Proxied request | 30s | n/a (flat) |

- Replace the single `AtomicLong` last-timeout slot in `HTTPTimeoutManager` with a per-host `ConcurrentHashMap` (`host -> lastTimeoutDate`, 10-minute per-entry expiry, pruned on access).
- Re-tier the request timeouts as above; fallback-host and proxied requests never consult the memory.
- Record a timeout for any main-source request that timed out (keyed by attempt host), dropping the previous fallback-support-only condition; a successful main-source response clears only that host's entry.
- Rename the `TIMEOUT_ON_MAIN_BACKEND_FOR_FALLBACK_SUPPORTED_ENDPOINT` result to `MAIN_SOURCE_TIMED_OUT`; add `AppConfig.hasProxyURL`.
- **Gated behind `usesRemoteConfigAPISources`:** the no-fallback re-tiering (15s/5s) and the broadened timeout recording apply only when the setting is enabled. With it disabled (the default), no-fallback endpoints keep the legacy 30s flat timeout and only fallback-supported endpoints arm the per-host memory; the fallback-supported and proxied tiers are unaffected.

iOS counterpart: RevenueCat/purchases-ios#7179

---

> [!NOTE]
> **Medium Risk**
> Changes default network timing and failover behavior for API requests across hosts; legacy 30s flat timeouts remain when remote-config API sources are disabled, but enabled configs get shorter waits and broader timeout recording.
> 
> **Overview**
> Replaces **global** timeout memory with **per-host** tracking so one slow API host does not shorten timeouts for unrelated hosts, and success on one host no longer resets fail-fast state for all.
> 
> **`HTTPTimeoutManager`** now keys recent timeouts by host (10-minute expiry) and applies distinct tiers: main-source with fallback URLs **5s / 2s** (reduced), main-source without fallback **15s / 5s** when re-tiering is enabled, fallback-host and proxied attempts **30s** flat (no memory). **`HTTPClient`** arms that memory on main-source `SocketTimeoutException` when the endpoint supports fallback URLs or `usesRemoteConfigAPISources` is on; a successful HTTP response clears only that host’s entry even if signature verification fails later.
> 
> **`Endpoint.targetsFallbackHost`** (`GetRemoteConfigFallback`) and **`AppConfig.hasProxyURL`** align timeout selection and “main backend” bookkeeping with domain-layer fallback-host and proxy routing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db959d037a89eb7ce037fe3f576c12ef4a23349a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -9,56 +9,79 @@ import com.revenuecat.purchases.InternalRevenueCatAPI
...
+     * @return The timeout in milliseconds.
      */
-    fun getTimeoutForRequest(isFallback: Boolean, fallbackAvailable: Boolean): Long {
-        // Check if reset is needed (10 minutes elapsed)
-        if (shouldResetTimeout()) {
-            resetTimeout()
-        }
-
+    fun getTimeoutForRequest(
+        host: String?,
```
AlvaroBrey, line 74: q: can `host` actually be null? how could that happen?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
```diff
@@ -69,16 +92,19 @@ internal class HTTPTimeoutManager(
...
 
     /**
-     * Records the result of a request and updates internal state accordingly.
-     * @param result The result of the request
+     * Records the result of a request attempt and updates the per-host memory accordingly.
+     * @param host The resolved host string of the attempt.
+     * @param result The result of the request.
      */
-    fun recordRequestResult(result: RequestResult) {
+    fun recordRequestResult(host: String?, result: RequestResult) {
```
AlvaroBrey, line 110: same here

AlvaroBrey (approved):
Looks good and makes sense I think! Just one question

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPTimeoutManager.kt
ajpallares: I believe Java's `URL.getHost()` returns a `String?`, so that'd be why. We call it from [here](https://github.com/RevenueCat/purchases-android/pull/3746/changes#diff-1eb7acca809981914bd4da4d178442b7a65747a3c3f9aa0f0c4478c7251f78e0R340)

AlvaroBrey: I see, java shenanigans ☕ . Forget about it then!

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -195,7 +195,7 @@ internal class HTTPClient(
             )
         }
 
-        val isMainBackend = fallbackURLIndex == 0
+        val isMainBackend = fallbackURLIndex == 0 && !endpoint.targetsFallbackHost
```
tonidero, line 198: ❤️

tonidero (approved):
I think this makes a lot of sense. Thank you!!

PR #3747, opened 2026-07-09
Coalesce `onPaywallConfigReady` so `getOfferings` readiness runs once

vegaro:
`getOfferings` does not dispatch success until `WorkflowManager.onPaywallConfigReady` completes. That gate resolves the workflows topic and `ui_config` before returning. If it is invoked more than once like by `getOfferings` and `configure()` it will redundantly call the config manager more than once.

---

> [!NOTE]
> **Low Risk**
> Localized concurrency change to the paywall readiness gate with matching unit tests; behavior for single callers and best-effort failures stays the same.
> 
> **Overview**
> **Overlapping `onPaywallConfigReady` calls now share one in-flight readiness pass** instead of each starting its own workflows-topic sync and `ui_config` fetch. That matters because `OfferingsManager` gates `getOfferings` success on this hook, and concurrent callers (e.g. configure plus get offerings) used to duplicate config work.
> 
> `WorkflowManager` keeps a single `Deferred` under a lock, reuses it while readiness is running, and clears it when that work finishes. Each caller still gets its own `onComplete` after awaiting the shared deferred. Docs note that sequential calls after completion may start a fresh check, which should be cheap once caches are warm.
> 
> Tests cover coalescing (one `awaitReady` / one `getUiConfig`, both callbacks fire) and **no `onComplete` when `close()` cancels the scope** before readiness completes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4f2a041b8d6fd4413f04e518277b4fb5a997abe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
@@ -10,8 +10,10 @@ import com.revenuecat.purchases.common.uiconfig.UiConfigProvider
 import com.revenuecat.purchases.utils.WorkflowAssetPreDownloader
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -118,15 +129,34 @@ internal class WorkflowManager(
...
+                }
+            }.also { deferred ->
+                inFlightReadiness = deferred
+                deferred.invokeOnCompletion {
+                    synchronized(readinessLock) {
+                        if (inFlightReadiness === deferred) {
+                            inFlightReadiness = null
+                        }
+                    }
+                }
```
tonidero, line 155: Hmm I'm unsure of how big of an issue this is.. Ideally, the RemoteConfigManager should resist calling the same thing multiple times/from multiple places and at least deduplicate requests there, so it wouldn't be a big deal if multiple requests come simultaneously. If it doesn't, maybe it's something we need to fix at that layer...

But in any case, I guess this does allow for some deduplication, so happy to ship this for now.

tonidero (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: Yeah, I did benchmark before and after and did see an improvement. I thought it was double fetching but it's not because there's only one call performed. I believe the delay is pure execution time, since double calling this function re-runs the code even when the blobs are already cached. It was firing on my tests more than once for a cold `getOfferings`, one from `configure` and one from the explicit `getOfferings`.

So there is no need to do a dedupe fix at the `RemoteConfigManager` layer, that looks likeit's covered.

PR #3748, opened 2026-07-09
fix(remote-config): require all UI config fields except custom_variables

ajpallares:
### Checklist

### Motivation
Stacked on #3739. Remote `ui_config` decoding silently fell back to empty defaults when parts were missing; iOS treats them as required.

### Description
- Make `app`, `localizations`, and `variable_config` (plus their nested maps) required in `UiConfig`; only `custom_variables` keeps its empty-map fallback.
- A missing or malformed part now fails to decode instead of producing a degraded config.
- Added decode tests asserting each required field throws when absent.

---

> [!NOTE]
> **Medium Risk**
> Incomplete or legacy remote ui_config payloads will fail decode and surface as unavailable UI config rather than silently degraded paywalls; behavior change is intentional and aligned with iOS.
> 
> **Overview**
> Remote **`ui_config` JSON decoding** no longer fills in empty defaults for missing top-level or nested fields. **`app`**, **`localizations`**, and **`variable_config`** (including **`app.colors`**, **`app.fonts`**, and both compatibility maps) are now **required**; absent or malformed JSON throws **`SerializationException`** instead of producing a partial config. **`custom_variables`** remains optional and still defaults to an empty map when omitted.
> 
> Call sites that built **`UiConfig()`** in memory—tests, previews, injected workflows, paywall-tester samples—now use **`emptyUiConfig()`** / **`uiConfigWithFonts()`** helpers that supply explicit empty structures. Decode tests were extended to assert each required field fails when missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e4efc647f6a4b50f54f480485f2ed4fb0b804f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/SamplePaywalls.kt
+++ b/examples/paywall-tester/src/main/java/com/revenuecat/paywallstester/SamplePaywalls.kt
@@ -71,6 +71,7 @@ class SamplePaywallsLoader {
                 Offering.PaywallComponents(
                     uiConfig = UiConfig(
                         app = UiConfig.AppConfig(
+                            colors = emptyMap(),
                             fonts = mapOf(
                                 primaryLocalFont to FontsConfig(android = FontInfo.Name("lobster_two")),
                                 secondaryGoogleFont to FontsConfig(android = FontInfo.GoogleFonts("Barrio")),
@@ -78,6 +79,10 @@ class SamplePaywallsLoader {
                             ),
                         ),
                         localizations = mapOf(localeId to variableLocalizationKeysForEnUs()),
+                        variableConfig = UiConfig.VariableConfig(
+                            variableCompatibilityMap = emptyMap(),
+                            functionCompatibilityMap = emptyMap(),
+                        ),
                     ),
                     data = data,
                 )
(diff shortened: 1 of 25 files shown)


tonidero (approved):
I think that makes sense... I guess these shouldn't happen, and I think we should already handle the case where one of the ui_config blobs being completely missing, so I was just wondering how you were able to reproduce these? In any case, I think these should be fine.

ajpallares:
> I think that makes sense... I guess these shouldn't happen, and I think we should already handle the case where one of the ui_config blobs being completely missing, so I was just wondering how you were able to reproduce these? In any case, I think these should be fine.

Right @tonidero. This was not something I could reproduce myself. It was more of a theoretical fix to cover the hypothetical cases where some of these properties would be missing from the backend response. In such situation, falling back to empty dictionaries would degrade the paywall/workflow. More context on [this Slack thread](https://revenuecat.slack.com/archives/C0AK23U9DBN/p1783609813377249). cc @rickvdl in case you want to add something to the explanation

tonidero:
Right I missed that thread. I think it's fine to add this, just nothing that concerns me too much I guess :) Thank you!!

PR #3749, opened 2026-07-10
Bound blob downloads with per-source timeouts

ajpallares:
### Motivation
Blob downloads used one flat 5s connect/read timeout for every attempt: tight enough to give up on a first fetch of a large blob, yet slow to abandon a source that just timed out.

### Description
- Bound each blob-source attempt via #3746's shared timeout manager and its no-fallback tier (15s base, 5s if that host timed out in the last 10 min), keyed by the blob source host.
- Record per-host results so a flaky source fails fast on re-hit; a timeout still advances to the next source as before.

Stacks on #3746.

---

> [!NOTE]
> **Medium Risk**
> Changes remote config blob download timing and failover behavior on slow or flaky hosts; shared timeout state with the main HTTP client could affect concurrent downloads to the same host.
> 
> **Overview**
> Remote config **blob downloads** no longer use a fixed 5s connect/read timeout on every attempt. They now share the same **`HTTPTimeoutManager`** as the main API client: **15s** base for blob-source hosts with no alternate URL, **5s** after that host recently timed out (10-minute memory), keyed per host.
> 
> **`RemoteConfigBlobFetcher`** takes `HTTPTimeoutManager`, resolves timeouts before each download, handles **`SocketTimeoutException`** as source-unhealthy (failover), and records results **after** the attempt completes so a 200 with a failing body read does not prematurely clear fail-fast state. **`PurchasesFactory`** wires one shared timeout manager into **`HTTPClient`** and the blob fetcher injected into **`RemoteConfigManager`**.
> 
> **`UrlConnectionFactory.createConnection`** now requires explicit connect/read timeouts; callers that still want 5s use **`DEFAULT_CONNECTION_TIMEOUT_MS`** (health checks, file repo, generic download helpers).
> 
> Broad test coverage was added for timeout tiers, per-host isolation, and interaction with checksum failures and mid-body errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1cd84d7f57fa8cf3add8bc0313b46f32cf604450. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rickvdl (approved):
I think this looks good! Thank you

tonidero (approved):
Looks great!

PR #3750, opened 2026-07-10
fix(remote-config): use a dedicated fallback endpoint

tonidero:
The config endpoint fallback wasn't really working in the previous format, since it has some differences from the main endpoint. For example:
- It's a GET, not a POST
- Uses application/json, not x-rc-format
- No inline items
- No `nonce` for signing

Since it was quite different, we split the fallback to a completely separate endpoint. This will only be hit if there is nothing in cache, preferring cache over the fallback values.

Also added some integration tests for this

Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/7182

---

> [!NOTE]
> **Medium Risk**
> Changes cold-start remote config networking and commit behavior (no inlined blobs on fallback), but scope is limited to empty-cache retry paths and is covered by integration and unit tests.
> 
> **Overview**
> Remote config no longer relies on the generic HTTP client URL fallback for `/v1/config`. The main **`GetRemoteConfig`** call always hits the primary base URL only (no `fallbackBaseURLs` on that request), and a separate **`GetRemoteConfigFallback`** endpoint handles recovery.
> 
> **Dedicated fallback path:** **`Backend.getRemoteConfigFallback`** issues a **GET** to the first configured fallback host, parses plain **`application/json`** into **`RemoteConfiguration`** (not RC container format), and verifies signatures **without a nonce** (like offerings). **`RemoteConfigManager`** only uses this on **cold start**: when the main POST fails with a retryable error **and** there is **no cached config**. If anything is already cached, that data is kept and the fallback is skipped; **4xx** still disables the endpoint without trying fallback.
> 
> **Persistence:** Fallback commits pass **`container = null`**, so inlined blobs are not extracted from the response; wanted blobs are fetched via **`prefetchBlobs`** / the blob fetcher instead.
> 
> Unit, manager, endpoint, and production integration tests (including ETag caching on the fallback host) cover the new flow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cad78ce91e4218a56169d1650a48c5a8845262b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -28,6 +28,7 @@ import com.revenuecat.purchases.common.networking.RewardVerificationResponse
 import com.revenuecat.purchases.common.networking.WebBillingProductsResponse
 import com.revenuecat.purchases.common.networking.buildPostReceiptResponse
 import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMapping
+import com.revenuecat.purchases.common.remoteconfig.RemoteConfiguration
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
 import com.revenuecat.purchases.customercenter.CustomerCenterRoot
@@ -109,6 +110,11 @@ internal typealias RewardVerificationResultCallback =
 
 internal typealias RemoteConfigCallback = Pair<
     (RCContainer?, VerificationResult) -> Unit,
+    (PurchasesError, errorHandlingBehavior: GetRemoteConfigErrorHandlingBehavior) -> Unit,
+    >
+
+internal typealias RemoteConfigFallbackCallback = Pair<
+    (RemoteConfiguration) -> Unit,
     (PurchasesError, errorHandlingBehavior: GetRemoteConfigErrorHandlingBehavior) -> Unit,
     >
 
@@ -209,6 +215,10 @@ internal class Backend(
     @get:Synchronized @set:Synchronized
     @Volatile var remoteConfigCallbacks =
         mutableMapOf<BackgroundAwareCallbackCacheKey, MutableList<RemoteConfigCallback>>()
+
+    @get:Synchronized @set:Synchronized
+    @Volatile var remoteConfigFallbackCallbacks =
+        mutableMapOf<BackgroundAwareCallbackCacheKey, MutableList<RemoteConfigFallbackCallback>>()
 
     fun close() {
         this.dispatcher.close()
@@ -1162,7 +1172,6 @@ internal class Backend(
         val cacheKey = BackgroundAwareCallbackCacheKey(listOf(path, appUserID), appInBackground)
 
         val baseURL = appConfig.baseURL
-        val fallbackBaseURLs = appConfig.fallbackBaseURLs
         // The manifest is an opaque token replayed verbatim; omitted on the first run when there is none.
         val body = buildMap<String, Any?> {
             put(APP_USER_ID, appUserID)
@@ -1178,7 +1187,6 @@ internal class Backend(
                     body = body,
                     postFieldsToSign = null,
                     backendHelper.authenticationHeaders,
-                    fallbackBaseURLs = fallbackBaseURLs,
                 )
             }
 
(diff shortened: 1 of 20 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1250,6 +1258,89 @@ internal class Backend(
...
     }
 
+    fun getRemoteConfigFallback(
+        appInBackground: Boolean,
+        domain: String,
+        onSuccess: (RemoteConfiguration) -> Unit,
+        onError: (PurchasesError, GetRemoteConfigErrorHandlingBehavior) -> Unit,
+    ) {
+        val fallbackURL = appConfig.fallbackBaseURLs.firstOrNull()
+        if (fallbackURL == null) {
```
tonidero, line 1268: I disable this if using a proxy... TBH, not sure if we should, but trying to keep same behavior we had before this for offerings/pem

purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionRemoteConfigFallbackIntegrationTest.kt
```diff
@@ -0,0 +1,154 @@
...
+ * call the fallback host (which is never faked) so they hit the real endpoint, and the manager-flow test relies
+ * on the main `api.revenuecat.com` request being forced to a `5xx` so the domain layer routes to the fallback.
+ *
+ * TODO: Ignored until the backend deploys the fallback config endpoint. As of writing, `GET /v1/config/{domain}`
+ * on the fallback host returns `404 Object Not Found` (the sibling `/v1/offerings` and
+ * `/v1/product_entitlement_mapping` fallbacks return `200`, so the host and auth are fine — the config snapshot
+ * just isn't served yet). Re-enable once it is live. The SDK logic is fully covered by the unit tests in
+ * `BackendGetRemoteConfigTest` and `RemoteConfigManagerTest`.
+ */
+@Ignore("Backend fallback config endpoint (GET /v1/config/{domain}) not deployed yet; returns 404.")
```
tonidero: I'm disabling it since the app we're using for testing doesn't have data yet... Will retry this later

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -199,6 +199,66 @@ internal class RemoteConfigManager(
...
+     * without a fallback.
+     */
+    private fun handleMainRefreshError(
+        requestEpoch: Int,
+        appInBackground: Boolean,
+        persisted: PersistedRemoteConfigurationState?,
+        error: PurchasesError,
+        behavior: GetRemoteConfigErrorHandlingBehavior,
+    ) {
+        if (behavior == GetRemoteConfigErrorHandlingBehavior.SHOULD_RETRY && persisted == null) {
```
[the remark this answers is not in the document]
tonidero: I think this is fine

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1250,6 +1258,89 @@ internal class Backend(
...
+                }
+            }
+
+            override fun onCompletion(result: HTTPResult) {
+                synchronized(this@Backend) {
+                    remoteConfigFallbackCallbacks.remove(cacheKey)
+                }?.forEach { (onSuccessHandler, onErrorHandler) ->
+                    if (result.isSuccessful()) {
+                        try {
+                            onSuccessHandler(RemoteConfiguration.parse(result.payloadText.encodeToByteArray()))
```
rickvdl: We're not reporting the signature verification result here, which we do for the regular endpoint. I think we should be able to report that back as well.

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1250,6 +1258,89 @@ internal class Backend(
...
+                        } catch (e: SerializationException) {
+                            // A 2xx body that doesn't parse is not a client error; keep retrying.
+                            onErrorHandler(
```
rickvdl, line 1314: Hm are we handling 204's here? That body would fail to parse since it's empty I think?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/Endpoint.kt
```diff
@@ -81,14 +81,23 @@ internal sealed class Endpoint(
...
+
+    /**
+     * Fallback for [GetRemoteConfig] served from a fallback base URL. Unlike the main endpoint, it is a plain
+     * `GET` returning `application/json` (no `x-rc-format`, no inlined blobs, no request body), and its response
+     * is signature-verified **without** a nonce (mirroring [GetOfferings]). The domain layer chooses this
+     * endpoint explicitly; it does not participate in the generic HTTPClient URL-fallback mechanism.
+     */
+    data class GetRemoteConfigFallback(val domain: String) : Endpoint(
+        pathTemplate = "/v1/config/%s",
+        name = "remote_config_fallback",
```
rickvdl, line 97: iOS uses `remote_config_static_fallback`, I think it's a bit more descriptive in signaling that it's handeled differently, but no strong opinion. If this is used for diagnostics I think we should consolidate?

tonidero: Ohh you're totally right we should consolidate. Thank you!! And yeah... I think `fallback` would already signal it, but happy to add `static` as well to be more precise 👍

rickvdl: True as well, ok I'll change it :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Right... I forgot about it since it's not exposed anywhere yet (in enforced, it would fail inside the network layer)... But yeah, we should consolidate. Thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: So I think the fallback endpoint wouldn't return 204's, but it would use the preexisting etag system instead, which would return the cached body I think. I'm adding some tests to make sure we handle that, but I think it should work as is

tonidero: Added test in https://github.com/RevenueCat/purchases-android/pull/3750/changes/731670e00c511e5bbec9f3a7d276afee80b1642e#diff-bc80cb0c30485f43f54cef1ff111e6ef04dac47e48ee11ff48bee96a606d1b1c

rickvdl: 🤦 Yep you're right indeed, let me change this on my side as well

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: Added in https://github.com/RevenueCat/purchases-android/pull/3750/commits/e04ed8358ab91b53f77decb710cfb2f482ff7ac8. Thanks again!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -199,6 +199,66 @@ internal class RemoteConfigManager(
...
+    private fun fetchFromFallback(
+        requestEpoch: Int,
+        appInBackground: Boolean,
+        previous: PersistedRemoteConfigurationState?,
+    ) {
+        if (epoch.get() != requestEpoch) return
+        verboseLog { "Main remote config request failed with no cached config; trying the fallback endpoint." }
+        backend.getRemoteConfigFallback(
+            appInBackground = appInBackground,
+            domain = previous?.domain ?: DEFAULT_DOMAIN,
```
rickvdl: Tiny discrepancy here; iOS uses the previous requests domain, whereas android will always use the `DEFAULT_DOMAIN` (`app`), because it will only reach this line when we have no `previous` AFAIK?

rickvdl (approved):
Awesome work, one more question 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3750/commits/cad78ce91e4218a56169d1650a48c5a8845262b1

PR #3755, opened 2026-07-13
refactor(remote-config): Clarify prefetch remote-config blobs in the server-provided order

tonidero:
Prefetch non-inlined prefetch blobs in the order the backend returns them in `prefetch_blobs`. This is already the case right now but it's more incidental (LinkedHashSet preserves order but List is a better type to represent this). Also added tests to preserve this behavior.

---

> [!NOTE]
> **Low Risk**
> Small type/refactor in prefetch assembly plus unit tests; no auth, persistence, or API surface changes beyond prefetch scheduling order being explicit and covered.
> 
> **Overview**
> Remote config blob prefetch now builds an **ordered** `List` (server `prefetch_blobs` first, then topic items with `prefetch: true`) and **dedupes** while keeping the first occurrence—replacing a `Set` where order was only incidental.
> 
> New tests lock in that **`RemoteConfigBlobFetcher`** runs LOW-priority prefetch in list order and that **`RemoteConfigManager`** passes refs to `blobFetcher.prefetch` in server order, then item refs, with duplicates collapsed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0419c982d336a9ab231a0402047965cd7d442e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -733,12 +733,12 @@ internal class RemoteConfigManager(
      */
     private fun prefetchBlobs(response: RemoteConfiguration, mergedTopics: Map<String, ConfigTopic>) {
         sourceProvider.restartIfExhausted(RemoteConfigSourceHandle.Purpose.BLOB)
-        val refs = buildSet {
+        val refs = buildList {
             addAll(response.prefetchBlobs)
             mergedTopics.values.forEach { topic ->
                 topic.values.forEach { item -> if (item.prefetch) item.blobRef?.let(::add) }
             }
-        }
+        }.distinct()
         val toPrefetch = refs.filterNot { blobStore.contains(it) }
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -733,12 +733,12 @@ internal class RemoteConfigManager(
      */
     private fun prefetchBlobs(response: RemoteConfiguration, mergedTopics: Map<String, ConfigTopic>) {
         sourceProvider.restartIfExhausted(RemoteConfigSourceHandle.Purpose.BLOB)
-        val refs = buildSet {
+        val refs = buildList {
```
tonidero, line 736: As I mentioned in the description, it worked before this change, but using a list is more explicit that order is important.

rickvdl (approved):
Great to make this explicit! AFAIK iOS and Android are now aligned as well. Thanks!

PR #3756, opened 2026-07-13
fix(remote-config): throttle failed refresh attempts

rickvdl:
Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/7191

## Motivation

When locally cached remote-config data is available, a failed request to the remote config endpoint does not fall back. However it also does not update the local cache (which makes sense). However this results in it also bypassing the staleness check, causing every subsequent request to be sent to the backend right away.

Now that we're refreshing the remote config state when fetching offerings, any call to `getOfferings()` (or any app-foreground) will make another call to the remote config API as long as it's degraded.

## Changes
This PR adds the `lastRefreshAttemptAt` and will only allow a new request to be sent after the `rereshAttemptCooldownInSeconds` has passed, which I set to `30s` for now (up for discussion)

---

> [!NOTE]
> **Medium Risk**
> Changes refresh scheduling for a core sync path used by offerings and topic reads; mis-tuned cooldown or gating could delay recovery after outages, though explicit refresh and cache clear remain escape hatches.
> 
> **Overview**
> Stops **retry storms** when `/v1/config` fails but local config is still usable: failed syncs never update `lastRefreshedAt`, so every stale check (e.g. offerings / foreground) could hit the API again immediately.
> 
> **`RemoteConfigManager`** now tracks **`lastRefreshAttemptAt`** and routes refreshes through a **`staleGated`** flag. **`refreshRemoteConfigIfStale`**, cold reads (`awaitConfigForRead`), and topic-triggered on-demand syncs use stale-gated mode and are **skipped** until **`REFRESH_ATTEMPT_COOLDOWN` (1 minute)** elapses after the last started attempt. **`refreshRemoteConfig()`** (explicit) **ignores** that cooldown so callers can retry right away. The cooldown is recorded only when a stale-gated refresh **actually starts** (not when one is skipped because another is in flight). **`clearCache`** clears the attempt timestamp so a new user can sync immediately.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40de1c66ceaab972eeaadaa9883b0e9d589d05af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -33,7 +33,7 @@ import kotlinx.serialization.json.decodeFromJsonElement
 import java.util.Date
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicInteger
-import kotlin.time.Duration.Companion.seconds
+import kotlin.time.Duration.Companion.minutes
 
 /**
  * Orchestrates a single `/v1/config` sync: replays the persisted opaque manifest, then on `204` keeps the cache
@@ -803,7 +803,7 @@ internal class RemoteConfigManager(
 
     private companion object {
         private const val DEFAULT_DOMAIN = "app"
-        private val REFRESH_ATTEMPT_COOLDOWN = 30.seconds
+        private val REFRESH_ATTEMPT_COOLDOWN = 1.minutes
     }
 }
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -33,6 +33,7 @@ import kotlinx.serialization.json.decodeFromJsonElement
 import java.util.Date
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicInteger
+import kotlin.time.Duration.Companion.seconds
 
 /**
  * Orchestrates a single `/v1/config` sync: replays the persisted opaque manifest, then on `204` keeps the cache
@@ -98,6 +99,9 @@ internal class RemoteConfigManager(
     @Volatile
     private var lastRefreshedAt: Date? = null
 
+    @Volatile
+    private var lastRefreshAttemptAt: Date? = null
+
     // The app user a cold on-demand read should sync for, rebound by clearCache() under [cacheLock] atomically
     // with the epoch bump, so it can never lag the identity change the way [appUserIDProvider] (backed by the
     // device cache) can. Null until the first identity change; awaitConfigForRead() falls back to the provider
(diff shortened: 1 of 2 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -184,6 +188,17 @@ internal class RemoteConfigManager(
         )
     }
 
+    private fun refreshRemoteConfigIfAttemptCooldownElapsed(appInBackground: Boolean, appUserID: String) {
+        val now = dateProvider.now
+        val lastAttempt = lastRefreshAttemptAt
+        if (lastAttempt == null || now.time - lastAttempt.time >= REFRESH_ATTEMPT_COOLDOWN.inWholeMilliseconds) {
+            lastRefreshAttemptAt = now
```
tonidero: Hmm I was wondering if we should be setting this in the `cacheLock`... Probably not a huge deal, but in case multiple requests happen simultaneously, maybe we could?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -291,6 +306,7 @@ internal class RemoteConfigManager(
         synchronized(cacheLock) {
             if (epoch.get() == requestEpoch) {
                 lastRefreshedAt = dateProvider.now
+                lastRefreshAttemptAt = null
```
tonidero: Hmm I believe we shouldn't reset this to null on a 204?

tonidero (commented):
Just a couple comments but I think we're almost there! Thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: I'm not so sure tbh, I think a 204 is a successful refresh, not resetting it here would mean we wouldn't make new requests for the throttle window because we consider the endpoint unhealthy?

tonidero: Hmm well, basically I think it shouldn't really take any effect (shouldn't ever be reached), because it would be blocked by the staleness check before this new check right? So I actually think it doesn't matter if we reset it or not, but conceptually, I would leave it so that has the time of the last attempt, whether successful or failed (though it only matters later in the checks, if it passes the staleness check which would only happen if we had errored requests).

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Good point! Will add that :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -184,6 +209,11 @@ internal class RemoteConfigManager(
         )
     }
 
+    private fun isRefreshAttemptCooldownElapsed(now: Date): Boolean {
+        val lastAttempt = lastRefreshAttemptAt
+        return lastAttempt == null || now.time - lastAttempt.time >= REFRESH_ATTEMPT_COOLDOWN.inWholeMilliseconds
```
tonidero: We could use `Duration.between` (a custom extension) to calculate the elapsed time, but NABD :)

tonidero (approved):
Looks great! Thank you!!

PR #3759, opened 2026-07-13
fix(remote-config): cache remote-config paywall data in memory to avoid loading flash

tonidero:
## Why

Workflow paywalls load their content (workflow + `ui_config` blobs) from the remote-config layer, which read from disk on every presentation, which is done asynchronously. That made the `LoadingPaywall` flash for a couple of frames.

## How

Cache the parsed `PublishedWorkflow` / `UiConfig` in memory and make the reads memory-first: on a warm cache the reads return synchronously (no dispatch), so the paywall renders on the first frame. On a cold cache it fetches and waits as before. The cache is warmed on init and on each config commit, invalidated on new versions / identity change, and `ui_config` plus the current-offering and `prefetch` workflows are the ones kept in memory.

---

> [!NOTE]
> **Medium Risk**
> Touches identity-change cache ordering and cross-user invalidation for in-memory paywall config; mistakes could leak stale workflow/ui_config across users or regress offerings gating.
> 
> **Overview**
> Fixes a brief **LoadingPaywall** flash when presenting workflow paywalls by keeping parsed **`UiConfig`** and **`PublishedWorkflow`** data in memory instead of re-reading remote config from disk on every presentation.
> 
> **`RemoteConfigManager`** now exposes a monotonic **`configGeneration`** and **`RemoteConfigCommitListener`** hooks (commit / invalidate on persist, identity wipe, and 4xx disable). **`UiConfigProvider`** and **`WorkflowsConfigProvider`** share **`GenerationGuardedCache`** (store-if-newer, stale in-flight reads dropped) and warm from disk on SDK init and after each commit without triggering a network sync via **`committedTopicOrNull`**. **`PurchasesFactory`** wires one shared pair of providers, registers listeners, and preloads caches from committed disk state on cold start.
> 
> **`WorkflowManager.getWorkflow`** is **`suspend`** and throws **`PurchasesException`** instead of callback + main-thread dispatch, so a warm read on **`Dispatchers.Main.immediate`** can finish without suspending. **`onPaywallConfigReady`** completes **synchronously** when both caches are warm; otherwise it warms **`ui_config`** then workflows (including sync when needed for the offerings gate).
> 
> **`IdentityManager`** clears remote config **before** offerings on login/alias/switch so stale per-user config cannot be served while the current offering is briefly null.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6daeafcab87a420a1d1289e350147b7833e1805. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -288,7 +288,26 @@ internal class PurchasesFactory(
...
             val uiConfigProvider = remoteConfigManager?.let { UiConfigProvider(it) }
+            val workflowsConfigProvider = remoteConfigManager?.let {
+                WorkflowsConfigProvider(
+                    it,
+                    currentOfferingIdProvider = { offeringsCache.cachedOfferings?.current?.identifier },
+                )
+            }
+            if (remoteConfigManager != null && uiConfigProvider != null && workflowsConfigProvider != null) {
+                remoteConfigManager.registerListener(uiConfigProvider)
+                remoteConfigManager.registerListener(workflowsConfigProvider)
```
tonidero, line 303: These kinda result in a circular reference at runtime, since `RemoteConfigManager` has the providers, but the providers also depend on the `RemoteConfigManager`. This is relatively acceptable in Android, since it shouldn't produce a memory leak, but this won't work as is in iOS or we could be leaking these classes... Though I guess we can hold the listeners weakly

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
```diff
@@ -17,15 +23,92 @@ import com.revenuecat.purchases.common.remoteconfig.RemoteConfigTopic
...
+    /**
+     * Returns the in-memory [UiConfig] if present, else resolves it from the config layer (which may wait for or
+     * trigger a `/v1/config` sync on a cold cache), caches it, and returns it. `null` when it can't be resolved.
+     */
+    suspend fun getUiConfig(): UiConfig? {
+        cache?.let { return it }
+        val generation = manager.configGeneration
+        val uiConfig = resolve()
+        if (uiConfig != null) store(generation, uiConfig)
+        return uiConfig
```
rickvdl: Theoretically here the generation may no longer match the current `lastGeneration`, should we change this to only return when still current?

rickvdl: There are also a couple of relevant cases in the `WorkflowsConfigerProvider`, maybe we could create a helper for it if we go this route

rickvdl (commented):
I think this makes sense so far! Will have another look after I created the iOS PR to cross-check

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
tonidero: Good point, will address! Thank you!

tonidero: This should address it https://github.com/RevenueCat/purchases-android/pull/3759/commits/8d985f9999371f7dc3c04fc6bfd9c793cf87faad. Thanks!!

rickvdl: That makes sense, I like the helper, thank you!

rickvdl (commented):
Changes make sense to me! Will leave approval to someone from the workflows team with the remaining context on workflow specifics if that's okay.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
```diff
@@ -76,6 +131,80 @@ internal class WorkflowsConfigProvider(
...
+    suspend fun warm() = warm(manager.configGeneration)
+
+    /** Fire-and-forget [warm] on this provider's own scope; used for the cold-start init warm. */
+    fun warmAsync(generation: Int) {
+        scope.launch { warm(generation) }
+    }
+
+    override fun onConfigCommitted(generation: Int) {
+        scope.launch { warm(generation) }
+    }
```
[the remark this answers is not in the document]
tonidero: I think this may indeed happen... but is probably okish and would avoid taking too long to return... It would mean someone would need to be presenting a paywall while logging out though 😅

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
```diff
@@ -18,12 +27,51 @@ import kotlinx.serialization.json.JsonPrimitive
...
+     * offering, the current offering has no workflow mapping (legacy/embedded paywall), or its workflow is parsed.
+     */
+    fun isWarmForCurrentOffering(): Boolean {
+        val cached = cache ?: return false
+        val workflowId = currentOfferingIdProvider()?.let { cached.offeringToWorkflowId[it] }
+        return workflowId == null || cached.workflows.containsKey(workflowId)
+    }
 
     suspend fun workflowIdForOfferingId(offeringId: String): String? {
+        cache?.let { return it.offeringToWorkflowId[offeringId] }
```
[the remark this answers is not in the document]
tonidero: I believe it should be ok, the warm up should happen immediately after a new commit.

vegaro:
@tonidero I ran some tests on this branch and these were my results:

Captura de pantalla 2026-07-14 a las 19 19 28

Kinda what we were expecting but wanted to shared. Although I still want to do a heap dump and take a closer look at the memory

tonidero:
Right… my plan is to remove the paywall paywallComponents from being kept in memory in a follow-up PR so we actually stay similar in memory usage hopefully (for single paged workflows)

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
```diff
@@ -103,8 +103,8 @@ internal class IdentityManager(
                         log(LogIntent.USER) {
                             IdentityStrings.ALIAS_OLD_USER_ID_TO_CURRENT_SUCCESSFUL.format(oldAppUserID, newAppUserID)
                         }
-                        offeringsCache.clearCache()
                         remoteConfigManager?.clearCache(newAppUserID)
+                        offeringsCache.clearCache()
```
vegaro: I asume this change is not really necessary, not asking to revert it but it made me wonder if it's needed and why

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -129,21 +121,24 @@ internal class WorkflowManager(
...
+        }
         val readiness: Deferred<Unit> = synchronized(readinessLock) {
-            inFlightReadiness ?: scope.async<Unit> {
-                coroutineScope {
-                    launch { awaitBestEffort("workflows topic") { workflowsConfigProvider.awaitReady() } }
-                    launch { awaitBestEffort("ui_config") { checkNotNull(uiConfigProvider.getUiConfig()) } }
-                }
+            inFlightReadiness ?: scope.async {
+                awaitBestEffort("ui_config") { checkNotNull(uiConfigProvider.getUiConfig()) }
+                awaitBestEffort("workflows") { workflowsConfigProvider.warm() }
```
vegaro, line 141: This `warm()` made me think. It calls `committedTopicOrNull(...)` so technically it's possible workflows returns null. Before we were doing `awaitReady` which actually made sure it trggered a network sync if not synced yet.

`onPaywallConfigReady` can be called from `vendCachedOfferingsAndMaybeRefresh`

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
tonidero: This was the agent explanation of the bug:
```
The in-memory workflow/ui_config caches are invalidated synchronously inside
remoteConfigManager.clearCache(), which ran after offeringsCache.clearCache()
on identity change. In the gap a concurrent getOfferings/paywall present saw
offerings==null (so isWarmForCurrentOffering() returned true) and the
memory-first reads returned the previous user's per-user (app_user_id /
enrolled_variants) config.
```
TBH, it's quite an edge case but seemed theoretically possible, so I switched the order. Lmk what you think!

vegaro: if it's order dependent, maybe we should have another function that "guarantees" the order. So they are always done in the same order?

tonidero: Yeah we probably should clean that up. Will take a look!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
tonidero: This is indeed valid... it shouldn't be a huge issue since it would self heal but could cause the loading screen to appear. So yeah working on a fix 🙏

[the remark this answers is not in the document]
tonidero: Same as comment above

tonidero: Fixed in https://github.com/RevenueCat/purchases-android/pull/3759/commits/d1a1772921d4baced38426c0d52875cbf65a29e7

purchases/src/main/kotlin/com/revenuecat/purchases/identity/IdentityManager.kt
tonidero: Extracted [here](https://github.com/RevenueCat/purchases-android/pull/3759/commits/f0c90043a3c1bb1b7ca254e46f5d5fdd7f015668)

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
```diff
@@ -17,15 +24,81 @@ import com.revenuecat.purchases.common.remoteconfig.RemoteConfigTopic
...
+     * No-op (no network) when `ui_config` isn't committed yet, so a cold-disk init warm never triggers a sync.
+     *
+     * If the topic **is** committed but its body can't be resolved right now (e.g. a transient blob read),
+     * the previous snapshot is dropped (store-if-newer [GenerationGuardedCache.invalidate]) rather than left
+     * behind tagged with an older generation. Otherwise, since [getUiConfig] is memory-first and never falls
+     * through on a hit, it would keep serving the outdated config after the generation advanced; dropping it
+     * makes the next [getUiConfig] re-resolve the committed config instead.
+     */
+    suspend fun warm(generation: Int) {
+        if (manager.committedTopicOrNull(RemoteConfigTopic.UiConfig) == null) return
```
vegaro, line 73: WorkflowConfigProvider.warm has an early return if already warm, maybe we can add it here as well

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
tonidero: Good point! Added.

PR #3761, opened 2026-07-14
chore(ci): api-check: show diff when api-check fails, only check api files

AlvaroBrey:
When the `metalava` check fails we have no info in the log as to _why_, so debugging it is a bit of a pain.

This PR:
- scopes the check to the metalava signature files (`*/api.txt`, `*/api-*.txt`), so unrelated tracked-file churn (e.g. a `mise.lock` rewrite) no longer fails it
- prints that `git diff` to stdout on failure so the changes show up directly in the log; the guidance message goes to stderr

---

> [!NOTE]
> **Low Risk**
> Changes only affect the CI api-check shell script; no runtime or security-sensitive code paths.
> 
> **Overview**
> **CI `api-check` now only fails on Metalava signature drift**, not on any dirty tracked file (e.g. `mise.lock` rewrites). The post-dump `git diff` is limited to `**/api.txt` and `**/api-*.txt` via pathspec globs.
> 
> **On failure, the log shows what changed:** the script prints a labeled `git diff` of those files to stdout and sends the “run api-dump” guidance to stderr.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 261d99785ea1f7ff9a298c22045a3e718a81fffd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/api-check.sh
+++ b/scripts/api-check.sh
@@ -2,11 +2,18 @@
 
 exit_code=0
 
+# Metalava signature files; only these should gate the check so unrelated
+# tracked-file churn (e.g. mise.lock) doesn't fail it.
+api_paths=(':(glob)**/api.txt' ':(glob)**/api-*.txt')
+
 # Run the api-dump.sh script to generate the API dump
 ./scripts/api-dump.sh || exit_code=$?
-# Check if there are any dirty changes in git
-if ! git diff --quiet; then
-  echo "API dump has changes, run the scripts/api-dump.sh script to generate the API dump, review and commit them."
+# Check if there are any dirty changes in the API signature files
+if ! git diff --quiet -- "${api_paths[@]}"; then
+  echo "Diff:"
+  git --no-pager diff -- "${api_paths[@]}"
+  echo
+  echo "API dump has changes, run the scripts/api-dump.sh script to generate the API dump, review and commit them." >&2
   exit_code=1
 fi
 


scripts/api-check.sh
```diff
@@ -6,7 +6,10 @@ exit_code=0
 ./scripts/api-dump.sh || exit_code=$?
 # Check if there are any dirty changes in git
 if ! git diff --quiet; then
```
tonidero: I do wonder if we should also change this to only diff `.api` files... But I guess this is not what this PR is trying to achieve 😅 . Happy to get this merged!

tonidero (approved):
:shipit: Thank you!!

scripts/api-check.sh
AlvaroBrey: Yeah, makes sense. Done in https://github.com/RevenueCat/purchases-android/pull/3761/changes/6903c6caeba491c5ecd950402d7619b0de28da5a to avoid another tiny PR, keeping the diff too as it's still useful

scripts/api-check.sh
```diff
@@ -2,11 +2,18 @@
...
+
 # Run the api-dump.sh script to generate the API dump
 ./scripts/api-dump.sh || exit_code=$?
-# Check if there are any dirty changes in git
-if ! git diff --quiet; then
-  echo "API dump has changes, run the scripts/api-dump.sh script to generate the API dump, review and commit them."
+# Check if there are any dirty changes in the API signature files
+if ! git diff --quiet -- "${api_paths[@]}"; then
+  echo "Diff:"
+  git --no-pager diff -- "${api_paths[@]}"
```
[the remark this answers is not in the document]
AlvaroBrey: False positive, the paths are single-quoted, so git does the matching, not the shell. But I'll make them globs for clarity

tonidero (approved):
Nice! Thanks for adding that as well 🙏

PR #3763, opened 2026-07-14
fix(paywalls): apply top window insets to the hero ZLayer, not the first nested stack

joshdholtz:
Fixes a paywall with a hero image + close button overlay rendering the close button roughly a status-bar height lower on Android than on iOS.

**Root cause:** the "apply top window insets" mark for a hero ZLayer was a self-resetting flag consumed by the *next stack style constructed*. Children are built before their parent, so a close button's inner stack consumed it instead of the ZLayer. Now the target stack is tracked by reference.

<details>
<summary>Details: mechanism, testing, and known gaps</summary>

### Mechanism

When the first visual component of a paywall is a full-width image inside a ZLayer (the "hero" case), `StyleFactory` marks that ZLayer to receive the top window insets, so its overlay children get pushed below the status bar while the hero image extends behind it.

That mark was a self-resetting boolean read by the next `StackComponentStyle` constructor. Since children styles are built before their parent's style, an overlay child containing a stack of its own — like a close button, whose content is a stack — consumed the flag first:

- the close button's inner stack applied the status-bar inset to its content, pushing the icon down, and
- the ZLayer itself ended up with `applyTopWindowInsets = false`, losing its hero inset handling entirely.

The fix tracks the stack that should apply the top insets by reference (`stackAwaitingTopWindowInsets`), so only the intended ZLayer picks it up regardless of construction order.

### Testing

- New `StyleFactoryTests` case mirroring the affected layout (hero ZLayer with a close-button overlay). It fails on `main` (`zStack.applyTopWindowInsets` is `false`, the button's inner stack got `true`) and passes with this change.
- Existing window-insets tests in `StyleFactoryTests` pass unchanged.
- `detektAll` passes. The two failing tests in the full `:ui:revenuecatui` suite (`TextComponentViewVariablesTests` timezone-dependent date, `PaywallComponentsTemplatePreviewRecorder` missing local preview-resources checkout) also fail on unmodified `main`.

### Known gaps

- Paparazzi snapshots render with zero window insets, so this fix produces no snapshot diff and cannot be screenshot-tested with the current harness.
- On-device before/after verification with the affected paywall layout is still pending.

</details>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
@@ -160,15 +160,18 @@ internal class StyleFactory(
 
         private class WindowInsetsState {
             /**
-             * Whether the current component should apply the top window insets. This field is reset when it is read,
-             * as it should only be set on a single component.
+             * The stack that should apply the top window insets, if any. Tracked by reference (rather than a
+             * flag consumed by the next stack style built) because children styles are built before their
+             * parent's: a flag intended for a hero ZLayer would otherwise be consumed by the first stack
+             * nested inside one of its overlay children (e.g. a close button's stack).
              */
-            var applyTopWindowInsets = false
-                get() {
-                    val value = field
-                    field = false
-                    return value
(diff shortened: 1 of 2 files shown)


joshdholtz:
@cursor please review this and please put extra effort into making sure this is correct and wont have any negative side effects

PR #3764, opened 2026-07-14
Improve memory usage on ETags by making `HTTPResult` body lazy

vegaro:
We got OOM reports pointing at JSONObject construction while ETagManager returns a cached HTTPResult through copy(). Keeping the JSON body lazy avoids parsing that cached payload until a caller actually reads result.body.

---

> [!NOTE]
> **Low Risk**
> Small behavioral refactor in networking with equivalent parse semantics on first `body` access; main effect is deferred work on cache hits.
> 
> **Overview**
> **Defers JSON parsing** on `HTTPResult` by making `body` a `lazy` property backed by a new `parseBody()` helper, instead of building a `JSONObject` in the property initializer on every instance.
> 
> That matters for the ETag path where cached responses are wrapped via `copy()`: those objects no longer pay parse cost unless something actually reads `body` (or derived fields like `backendErrorCode` / `backendErrorMessage` on failed responses). Parsing behavior and error logging on invalid JSON are unchanged once `body` is accessed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6d107522937eee04d7f9c96dd7209d3f122e8d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
@@ -124,17 +124,21 @@ public data class HTTPResult(
         BACKEND, CACHE
     }
 
-    val body: JSONObject = payloadText
-        .takeIf { it.isNotBlank() }
-        ?.let {
-            try {
-                JSONObject(it)
-            } catch (e: JSONException) {
-                errorLog(throwable = e) { "Failed to parse payload as JSON: $it" }
-                null
+    val body: JSONObject by lazy { parseBody() }
+
+    private fun parseBody(): JSONObject {
+        return payloadText
+            .takeIf { it.isNotBlank() }
+            ?.let {
+                try {
+                    JSONObject(it)
+                } catch (e: JSONException) {
+                    errorLog(throwable = e) { "Failed to parse payload as JSON: $it" }
+                    null
+                }
             }
-        }
-        ?: JSONObject()
+            ?: JSONObject()
+    }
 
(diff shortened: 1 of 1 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -124,17 +124,26 @@ public data class HTTPResult(
...
-            try {
-                JSONObject(it)
-            } catch (e: JSONException) {
-                errorLog(throwable = e) { "Failed to parse payload as JSON: $it" }
-                null
+    private val bodyDelegate = lazy { parseBody() }
+
+    val body: JSONObject by bodyDelegate
+
+    internal val isBodyInitialized: Boolean
```
tonidero: Hmm is this for tests only? I would probably remove it IMO. It could actually help simplify the code by just doing:
```
val body: JSONObject by lazy { parseBody() } 
```
We probably can't test this then... But I think that's ok TBH.

tonidero (approved):
Just a possible cleanup opportunity IMO.

tonidero:
Not sure if we should keep the label `pr:fix`... It's quite internal, but I guess it's still somewhat of a fix... A bit on the fence.

vegaro:
Yeah, I thought fix becuase it could fix OOMs? Maybe if I change the title?

btw, I did check and `body` is accessed while still in `onCompletion` before invoking callbacks. I was concerned we were suddenly moving parsing the body into another thread. So we should be good.

PR #3765, opened 2026-07-14
fix: avoid HTTPResult copy re-parsing payload in ETag cache

tonidero:
## Description

Copying an `HTTPResult` re-invokes its constructor, which re-parses the entire payload into a new `JSONObject`. For very large offerings responses this keeps two copies of the parsed JSON in memory at once and could crash with `OutOfMemoryError` when reusing the ETag-cached version (`ETagManager.getHTTPResultFromCacheOrBackend`).

---

> [!NOTE]
> **Medium Risk**
> Touches core HTTP/ETag caching and mutates shared `HTTPResult` instances during store/304 paths; behavior should match prior logic but shared references could surprise callers if they reuse the same object.
> 
> **Overview**
> **ETag cache paths no longer call `HTTPResult.copy()`**, which could re-run the string constructor and duplicate parsed JSON in memory (risk of `OutOfMemoryError` on large offerings when serving 304 cached responses).
> 
> On **304 Not Modified**, `ETagManager` now updates `requestDate` and `verificationResult` on the deserialized cached instance and returns it unchanged otherwise (payload and `CACHE` origin). When **persisting** a backend result, it temporarily sets `origin` to `CACHE` on that same object for serialization, then restores it in a `finally` block.
> 
> `HTTPResult` makes **`origin`**, **`requestDate`**, and **`verificationResult`** mutable (`var`) to support in-place updates. Tests cover 304 behavior and assert stored entries serialize with `Origin.CACHE`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9677b23ed763faeadea1833394fb533ef184eb25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
@@ -93,13 +93,9 @@ internal class ETagManager(
         )
         eTagHeader?.let { eTagInResponse ->
             if (shouldUseCachedVersion(responseCode)) {
-                val storedResult = getStoredResult(urlString)?.let { storedResult ->
-                    storedResult.copy(
-                        // This assumes we won't store verification failures in the cache and we will clear the cache
-                        // when enabling verification.
-                        verificationResult = verificationResult,
-                        requestDate = requestDate ?: storedResult.requestDate,
-                    )
+                val storedResult = getStoredResult(urlString)?.also { storedResult ->
+                    storedResult.requestDate = requestDate ?: storedResult.requestDate
+                    storedResult.verificationResult = verificationResult
                 }
                 return storedResult
                     ?: if (refreshETag) {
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -93,13 +93,9 @@ internal class ETagManager(
         )
         eTagHeader?.let { eTagInResponse ->
             if (shouldUseCachedVersion(responseCode)) {
-                val storedResult = getStoredResult(urlString)?.let { storedResult ->
-                    storedResult.copy(
-                        // This assumes we won't store verification failures in the cache and we will clear the cache
-                        // when enabling verification.
```
vegaro, line 99: this comment still applies I believe

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
```diff
@@ -21,9 +21,9 @@ private const val SERIALIZATION_NAME_IS_FALLBACK_URL = "isFallbackURL"
 public data class HTTPResult(
     val responseCode: Int,
     val payload: Payload,
-    val origin: Origin,
-    val requestDate: Date?,
-    val verificationResult: VerificationResult,
+    var origin: Origin,
+    var requestDate: Date?,
+    var verificationResult: VerificationResult,
```
vegaro, line 26: I think these can be `internal set`?

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
tonidero: Re-added in https://github.com/RevenueCat/purchases-android/pull/3765/commits/ac7ecd89ee6c266ede5de0602afec8d40ffe0d56

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/HTTPResult.kt
tonidero: It's actually tricky because these are constructor parameters... I'm actually thinking of making this whole class internal again (it was exposed for the galaxy changes, but are not needed anymore afaik cc @fire-at-will) but will do so in a separate PR.

tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3770

PR #3766, opened 2026-07-14
Add memory profiling harness

vegaro:
## Summary

Adds a repeatable macrobenchmark for comparing paywall-tester memory with workflows disabled and enabled.

- Reports MemoryUsageMetric Last and Max values across 10 cold-start iterations.
- Adds a BuildConfig switch for DangerousSettings.forWorkflows().
- Supports distinct application ID suffixes so the two runs cannot share app data.
- Updates benchmark-macro-junit4 to 1.4.1 for Android 16 compatibility.

## Measurement scope

This is appropriate for comparing cold-start peak memory and the immediate post-startup sample. It is not by itself a retained-memory or leak measurement; a settled phase or heap-dump analysis should be added before making that claim.

## Validation

Focused Kotlin compilation and BuildConfig generation passed. The harness was also run on a physical Pixel 6a with Android 16 against the latest fix branch.


vegaro:
Superseded by the memory measurements added to PR #3743, which reuses the real-app readiness and cache-control harness.

PR #3768, opened 2026-07-14
feat(remote-config): send fetch_context on config endpoint requests

ajpallares:
### Checklist
- [x] Unit tests

### Motivation
The backend caches resolved config for ~30 minutes, so config changes don't reach clients until the cache expires. Telling the backend why the SDK is fetching lets it decide when to force a fresh resolution. iOS equivalent: RevenueCat/purchases-ios#7214.

### Description
Adds a `fetch_context` field to every `POST /v1/config/{domain}` request, backed by a new `RemoteConfigFetchContext` enum (`app_start`, `foreground`, `identity_change`, `read`) threaded through `RemoteConfigManager` and the orchestrator call sites. It's also part of the backend request cache key, so fetches made for different reasons no longer coalesce (a fresh-wanting fetch can't piggyback on an in-flight `read`).

---

> [!NOTE]
> **Medium Risk**
> Touches all remote config fetch paths and changes request coalescing behavior; mis-assigned contexts could affect backend cache-bypass logic but does not alter auth or purchase flows.
> 
> **Overview**
> Adds **`fetch_context`** to every `POST /v1/config/{domain}` body so the backend knows why the SDK is fetching (e.g. to bypass stale server-side cache). A new internal **`RemoteConfigFetchContext`** enum (`app_start`, `foreground`, `identity_change`, `read`) is threaded through **`RemoteConfigManager`** and **`PurchasesOrchestrator`** call sites: first foreground uses **AppStart**, later foregrounds **Foreground**, login/logout/switch user **IdentityChange**, sync-attributes/offerings and cold topic reads **Read**.
> 
> **`Backend.getRemoteConfig`** now requires `fetchContext`, sends it in the JSON body, and includes **`fetchContext.wireName`** in the in-flight request dedupe key so concurrent fetches for the same user but different reasons do not share one HTTP call.
> 
> Tests and backend integration golden files were updated for the new parameter and body field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96a826779500c7be9e090e577ab8942012bb8160. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -28,6 +28,7 @@ import com.revenuecat.purchases.common.networking.RewardVerificationResponse
 import com.revenuecat.purchases.common.networking.WebBillingProductsResponse
 import com.revenuecat.purchases.common.networking.buildPostReceiptResponse
 import com.revenuecat.purchases.common.offlineentitlements.ProductEntitlementMapping
+import com.revenuecat.purchases.common.remoteconfig.RemoteConfigFetchContext
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfiguration
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.customercenter.CustomerCenterConfigData
@@ -1161,6 +1162,7 @@ internal class Backend(
     fun getRemoteConfig(
         appInBackground: Boolean,
         appUserID: String,
+        fetchContext: RemoteConfigFetchContext,
         domain: String,
         manifest: String?,
         prefetchedBlobs: List<String>,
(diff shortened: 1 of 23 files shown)


rickvdl (approved):
Changes look good! Thanks for adding this.

Indeed just the issue of sending another request with a different `fetchContext` while a request in-flight is still TBD...

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -1171,12 +1173,13 @@ internal class Backend(
         val path = endpoint.getPath()
         // Include the app user in the key: the path carries the domain but not the app_user_id (sent in the
         // request body), so concurrent calls for different users must not be deduped onto a single shared request.
-        val cacheKey = BackgroundAwareCallbackCacheKey(listOf(path, appUserID), appInBackground)
+        val cacheKey = BackgroundAwareCallbackCacheKey(listOf(path, appUserID, fetchContext.wireName), appInBackground)
```
tonidero, line 1176: TBH, I don't think we need to add it to the dedupe key and have to potentially do extra requests. As we talked over Slack, I think it could be ok just by making sure the `app_start` context has preference over orders if needed and we can revisit this in future versions.

tonidero (approved):
Just a comment, but I think it looks great! Thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
ajpallares: Right. I'd say it's not that important because the `RemoteConfigManager` is taking care of deduping the requests anyway. So, I don't think it hurts?

ajpallares: I'm merging this one as it is now. And I'll have to make a PR for forcing the `fetch_context: app_start` the first time anyways, so we can undo this in that PR if needed (also, because the iOS PR is already merged...)

PR #3772, opened 2026-07-15
fix(remote-config): skip decoding paywall components under workflows to save memory

tonidero:
## Description

Stacked on top of #3759.

With this PR, we won't keep the `Offering.paywallComponents` in memory if remote config is enabled. In case we get a killswitch from the config endpoint, we would refetch offerings but then keep it in memory on that attempt

We also add an instrumentation test to confirm the kill switch behaves as expected.

---

> [!NOTE]
> **Medium Risk**
> Changes offerings parsing, caching, and paywall fallback when remote config is disabled—important for paywall display but scoped to workflows and covered by new unit/integration tests.
> 
> **Overview**
> When **workflows** are on and `/v1/config` is still enabled, offerings parsing **no longer retains** `paywallComponents` JSON in memory (components come from remote config instead). **`hasPaywall`** still reflects a components paywall via a new internal **`hasPaywallComponents`** presence flag.
> 
> If remote config hits a **session 4xx kill switch**, the SDK fires **`onRemoteConfigDisabled`**, **clears** the in-memory offerings cache (with a generation guard so stale in-flight parses cannot overwrite the cache), **refetches** offerings, and **captures** paywall components again for the fallback render path.
> 
> Integration tests can pass **`DangerousSettings`** (e.g. workflows) through **`configureSdk`**. A new **production integration test** fakes `/v1/config` to assert the skip-then-fallback behavior end-to-end.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c7516ec2c69cafbe2f56ef9d8ea080510312745. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/api-defauts.txt
+++ b/purchases/api-defauts.txt
@@ -221,10 +221,11 @@ package com.revenuecat.purchases {
 
   @dev.drewhamilton.poko.Poko public final class Offering {
     ctor public Offering(String identifier, String serverDescription, java.util.Map<java.lang.String,?> metadata, java.util.List<com.revenuecat.purchases.Package> availablePackages);
-    ctor public Offering(String identifier, String serverDescription, java.util.Map<java.lang.String,?> metadata, java.util.List<com.revenuecat.purchases.Package> availablePackages, optional @com.revenuecat.purchases.InternalRevenueCatAPI com.revenuecat.purchases.paywalls.PaywallData? paywall, optional @com.revenuecat.purchases.InternalRevenueCatAPI com.revenuecat.purchases.Offering.PaywallComponents? paywallComponents, optional java.net.URL? webCheckoutURL);
+    ctor public Offering(String identifier, String serverDescription, java.util.Map<java.lang.String,?> metadata, java.util.List<com.revenuecat.purchases.Package> availablePackages, optional @com.revenuecat.purchases.InternalRevenueCatAPI com.revenuecat.purchases.paywalls.PaywallData? paywall, optional @com.revenuecat.purchases.InternalRevenueCatAPI com.revenuecat.purchases.Offering.PaywallComponents? paywallComponents, optional java.net.URL? webCheckoutURL, optional @com.revenuecat.purchases.InternalRevenueCatAPI boolean hasPaywallComponents);
     method public operator com.revenuecat.purchases.Package get(String s);
     method public com.revenuecat.purchases.Package? getAnnual();
     method public java.util.List<com.revenuecat.purchases.Package> getAvailablePackages();
+    method public boolean getHasPaywallComponents();
     method public String getIdentifier();
     method public com.revenuecat.purchases.Package? getLifetime();
     method public java.util.Map<java.lang.String,java.lang.Object> getMetadata();
@@ -244,6 +245,7 @@ package com.revenuecat.purchases {
     property public final com.revenuecat.purchases.Package? annual;
     property public final java.util.List<com.revenuecat.purchases.Package> availablePackages;
     property public final boolean hasPaywall;
+    property public final boolean hasPaywallComponents;
     property public final String identifier;
     property public final com.revenuecat.purchases.Package? lifetime;
     property public final java.util.Map<java.lang.String,java.lang.Object> metadata;
--- a/purchases/api-defauts.txt
+++ b/purchases/api-defauts.txt
@@ -221,11 +221,10 @@ package com.revenuecat.purchases {
 
   @dev.drewhamilton.poko.Poko public final class Offering {
     ctor public Offering(String identifier, String serverDescription, java.util.Map<java.lang.String,?> metadata, java.util.List<com.revenuecat.purchases.Package> availablePackages);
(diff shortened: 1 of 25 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -291,6 +292,17 @@ internal class PurchasesOrchestrator(
...
+
+        // When the `/v1/config` 4xx kill-switch trips, workflow-served paywalls are no longer available, so
+        // refetch offerings from the network: the parser skips capturing paywall components while the endpoint
+        // is live, and this refetch (with the endpoint now disabled) decodes them for the fallback render path.
+        remoteConfigManager?.registerListener(object : RemoteConfigCommitListener {
+            // Only the disable transition matters here; commits/invalidations are handled by the config providers.
+            override fun onConfigCommitted(generation: Int) = Unit
+            override fun onRemoteConfigDisabled(generation: Int) {
+                offeringsManager.fetchAndCacheOfferings(appUserID, state.appInBackground)
+            }
```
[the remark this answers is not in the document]
tonidero: Hmm true... we should probably first invalidate the cache so we force a refresh. On it.

purchases/api-defauts.txt
```diff
@@ -221,10 +221,11 @@ package com.revenuecat.purchases {
...
+    ctor public Offering(String identifier, String serverDescription, java.util.Map<java.lang.String,?> metadata, java.util.List<com.revenuecat.purchases.Package> availablePackages, optional @com.revenuecat.purchases.InternalRevenueCatAPI com.revenuecat.purchases.paywalls.PaywallData? paywall, optional @com.revenuecat.purchases.InternalRevenueCatAPI com.revenuecat.purchases.Offering.PaywallComponents? paywallComponents, optional java.net.URL? webCheckoutURL, optional @com.revenuecat.purchases.InternalRevenueCatAPI boolean hasPaywallComponents);
     method public operator com.revenuecat.purchases.Package get(String s);
     method public com.revenuecat.purchases.Package? getAnnual();
     method public java.util.List<com.revenuecat.purchases.Package> getAvailablePackages();
+    method public boolean getHasPaywallComponents();
```
rickvdl: Could we make this a private property in order to not expose it in the public API?

purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
```diff
@@ -181,12 +172,47 @@ internal abstract class OfferingParser {
...
+        }
+        if (!shouldParsePaywallComponents()) return null
+
+        // Defer the (potentially expensive) component-tree deserialization until the paywall is actually
+        // accessed/displayed. Capturing the raw JSON string here is cheap; without this we would eagerly
+        // deserialize every cached offering's component tree at load, even those that are never shown.
+        val rawPaywallComponents = paywallComponentsJson.toString()
+        // A content hash of the raw JSON serves as the equality key, so comparing offerings (e.g. cached vs
+        // network) never forces the lazy decode.
+        return Offering.PaywallComponents(uiConfig, componentsHash = rawPaywallComponents.sha256()) {
```
rickvdl, line 210: If I understand correctly this is not specific to the remote config enabled case, and this is a behavior change for regular offerings when remote config is disabled as well right? 

Although this probably improves performance offerings parse time, I'm a bit on the fence whether we should also add this now if we try to keep changes to a minimum..

rickvdl (approved):
I think this makes sense! Two questions mostly

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -291,6 +292,21 @@ internal class PurchasesOrchestrator(
...
+        // window before the refetch lands takes the cache-miss -> network path and gets freshly decoded
+        // components, instead of being served the stale null-component objects. The refetch (with the endpoint
+        // now disabled) then repopulates the cache proactively.
+        remoteConfigManager?.registerListener(object : RemoteConfigCommitListener {
+            // Only the disable transition matters here; commits/invalidations are handled by the config providers.
+            override fun onConfigCommitted(generation: Int) = Unit
+            override fun onRemoteConfigDisabled(generation: Int) {
+                offeringsManager.clearInMemoryOfferingsCache()
+                offeringsManager.fetchAndCacheOfferings(appUserID, state.appInBackground)
+            }
```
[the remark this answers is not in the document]
rickvdl: This one is probably valid

purchases/api-defauts.txt
tonidero: Right, we probably can... In any case, even though it appears here, it's annotated with `@InternalRevenueCatAPI`, but it's true we don't need it right now, so will make it private. Thank you!

purchases/src/main/kotlin/com/revenuecat/purchases/common/OfferingParser.kt
tonidero: Yes indeed, this is from before any changes for remote config. And yeah, it's a possible performance issue that we can also try to look at TBH. cc @vegaro But I think it should be worked on separately yeah

PR #3773, opened 2026-07-15
feat(remote-config): force app_start on the first config request

ajpallares:
### Motivation
On a fresh app open the backend should bypass its 30-min config cache. If another trigger (e.g. a `Read` from `getOfferings`) wins the race to send the first `/v1/config` request, the backend would only see that context and could serve stale config. Follow-up to the `fetch_context` work (#3768), per the team decision.

### Description
- `RemoteConfigManager` overrides the first committed request of the session to `AppStart`, regardless of the caller's context; subsequent requests report their real context.
- Removed `fetch_context` from the request dedup cache key, so concurrent requests differing only by context now coalesce.
- Added unit tests for the forced first request and that only the first one is overridden.

### Checklist
- [x] Unit tests

---

> [!NOTE]
> **Medium Risk**
> Changes which `fetch_context` the backend receives on cold start and how concurrent config requests coalesce, which can affect server-side cache bypass and telemetry; behavior is well covered by new unit tests.
> 
> **Overview**
> Ensures the backend sees **`app_start`** on a fresh app open even when another trigger (e.g. a **`Read`** from offerings) wins the race to call `/v1/config`.
> 
> **`RemoteConfigManager`** tracks **`hasCommittedInitialConfig`** and, under **`cacheLock`**, overrides the outgoing **`fetch_context`** to **`AppStart`** until config is actually committed—a successful persist (200 or fallback), or a **204** that confirms the cache is current. Failed requests, unparseable 200s, or failed persists keep forcing **`AppStart`** on later attempts; after commit, callers’ real contexts are sent again.
> 
> **`Backend.getRemoteConfig`** drops **`fetch_context`** from the in-flight dedup key (still keyed by path + **`app_user_id`** + background flag), so concurrent config calls that differ only by context share one HTTP request. Tests and integration golden files were updated for **`app_start`** on cold first requests and for dedup behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2869d89cd212a745bb576ee54331f2e65beeded7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -104,10 +104,10 @@ internal class RemoteConfigManager(
     @Volatile
     private var lastRefreshAttemptAt: Date? = null
 
-    // Forces the first committed request of the session to report AppStart, regardless of the caller's context, so
-    // the backend always sees app_start on a fresh app open. Guarded by [cacheLock]; set once when that request
-    // is committed.
-    private var hasCommittedInitialRequest = false
+    // Forces requests to report AppStart until the first one succeeds, regardless of the caller's context, so the
+    // backend always sees app_start on a fresh app open. Guarded by [cacheLock]; set when a request succeeds, so a
+    // failed attempt keeps forcing AppStart until a later attempt delivers it.
+    private var hasSucceededInitialRequest = false
 
     // The app user a cold on-demand read should sync for, rebound by clearCache() under [cacheLock] atomically
     // with the epoch bump, so it can never lag the identity change the way [appUserIDProvider] (backed by the
@@ -196,7 +196,7 @@ internal class RemoteConfigManager(
                     isRefreshing.set(true)
                     requestEpoch = epoch.get()
                     requestAppUserID = currentAppUserID ?: appUserID
-                    requestFetchContext = resolveCommittedFetchContext(fetchContext)
+                    requestFetchContext = fetchContextForRequest(fetchContext)
                     refreshCompletion = CompletableDeferred()
                     true
                 }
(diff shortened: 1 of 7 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -230,6 +237,13 @@ internal class RemoteConfigManager(
         return lastAttempt == null || Duration.between(lastAttempt, now) >= REFRESH_ATTEMPT_COOLDOWN
     }
 
+    // Overrides the first committed request's context to AppStart. Must be called within [cacheLock].
+    private fun resolveCommittedFetchContext(requested: RemoteConfigFetchContext): RemoteConfigFetchContext {
+        if (hasCommittedInitialRequest) return requested
+        hasCommittedInitialRequest = true
```
tonidero: Hmm I guess if that initial request fails for any reason, posterior ones won't actually add this... But it feels like a big edge case TBH, so I'm ok keeping as is.

tonidero (approved):
Makes sense to me! A small note, but don't consider it a blocker.

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
ajpallares: Hmmm great point. We could set the boolean at response time then?

ajpallares: Done in https://github.com/RevenueCat/purchases-android/pull/3773/commits/0f7c266ac910eeefe914e35a3180e53f47232f34 to set it at success time

tonidero (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -245,6 +267,7 @@ internal class RemoteConfigManager(
             // response and leave isRefreshing alone: clearCache() reset it, or a newer refresh owns it.
             return
         }
+        markInitialRequestSucceeded(requestEpoch)
```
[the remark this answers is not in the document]
ajpallares: Great point. Improved

PR #3774, opened 2026-07-15
fix: store ETag cache payload verbatim to avoid OOM on large responses

### LINKED ISSUE #3628: `OutOfMemoryError` in `ETagManager.serialize` when caching large /offerings response

vegaro:
Fixes RevenueCat/purchases-android#3628.

`HTTPResultWithETag.serialize()` embedded the payload inside two nested JSON string layers, so every quote in the payload was escaped twice (`"` → `\"` → `\\\"`). For a large `/offerings` response this built a single \~37MB string through `JSONObject.toString()`, which is the allocation OOMing on low-heap devices in the issue.

This splits the cache into two SharedPreferences entries written atomically in one editor commit:

* `<url>` → small flat metadata JSON (eTag, refresh time, response code, request date, verification result, flags)
* `<url>#rc_payload` → the payload string exactly as received — stored by reference, never JSON-escaped

Effects:

* Storing a response performs no payload-sized allocations at all (a test asserts the stored string is the same instance we received).
* `getETagHeaders` no longer deserializes the payload on every request just to read the eTag — it reads only the metadata entry.
* 304 reads return the payload string without the double-unescape.

Measured allocations per operation (5MB offerings-like payload, `ThreadMXBean.getThreadAllocatedBytes`, Robolectric with real SharedPreferences — `ETagManagerMemoryTest`, now a permanent regression gate):

| Operation | Before | After |
| -- | -- | -- |
| Store response in ETag cache (the RevenueCat/purchases-android#3628 crash site) | 49.3 MB | 150 KB |
| `getETagHeaders` on warm cache (every request) | 46.8 MB | 2 KB |
| 304 cache-hit read | 46.8 MB | 3 KB |

Both columns are measured with a warm-up pass that excludes one-time class-loading/first-use allocations, so they reflect steady-state per-operation cost.

* Legacy combined entries are migrated to the split format on first read; unparseable entries read as a cache miss and self-heal through the existing refresh-ETag retry.
* With the store no longer flipping `origin` in place, `HTTPResult`'s fields go back to `val` (reverting the mutability from RevenueCat/purchases-android#3765, which RevenueCat/purchases-android#3764's lazy `body` made unnecessary).

## Known limitation: SDK downgrade

Once this version writes split-format entries, downgrading to an older SDK leaves reads of those entries failing: the old `HTTPResultWithETag.deserialize` throws while building request headers — before any response could overwrite the entry — so it never self-heals until something clears the cache (identity change, app data clear). The failures surface as request errors through the existing error paths, not crashes, and device-level caches still serve their own copies. Accepted tradeoff for keeping the URL as the metadata key; the alternative (a separate metadata key, deleting the legacy entry on migration) trades this for a messier staleness story on re-upgrade.

<details><summary>Full Explanation</summary>

Let me walk through what actually happens to the bytes, with a tiny payload standing in for the real 15MB one.

## What we get from the server

Say `/offerings` returns this (in reality it's < />15MB of this kind of thing):

```json
{"offerings":[{"id":"premium"}]}
```

That string is sitting in memory. **We already have it. Done, right?** No — here's what the cache does to it.

## Step 1: `HTTPResult.serialize()` wraps it in JSON

The payload gets stuffed into a JSON object *as a string value*. To be a valid string value, every `"` inside it must be escaped to `\"`:

```json
{"responseCode":200,"payload":"{\"offerings\":[{\"id\":\"premium\"}]}","origin":"CACHE"}
```

Notice the payload part grew — every one of its quotes is now 2 characters instead of 1. Real offerings JSON is *full* of quotes (every key, every string value), so this pass alone grows the payload noticeably. And this is a brand-new full copy of the whole thing.

## Step 2: `HTTPResultWithETag.serialize()` wraps *that* in JSON again

Now the entire string from step 1 gets embedded into *another* JSON object, again as a string value. So every character that's special gets escaped **again** — each `"` becomes `\"`, and each `\` from step 1 becomes `\\`:

```json
{"eTag":"abc123","httpResult":"{\"responseCode\":200,\"payload\":\"{\\\"offerings\\\":[{\\\"id\\\":\\\"premium\\\"}]}\",\"origin\":\"CACHE\"}"}
```

Look at what happened to a single quote from the original payload:

| Stage | The character | Size |
| -- | -- | -- |
| Original payload | `"` | 1 char |
| After step 1 | `\"` | 2 chars |
| After step 2 | `\\\"` | 4 chars |

Every quote in the original payload is now **4 characters**. A 15MB payload dense with quotes balloons to roughly 2x+, and building that string goes through `StringBuilder`, which grows by doubling its internal array — that `Arrays.copyOf` of a < />36MB array is exactly the line that OOMs in issue RevenueCat/purchases-android#3628.

So to "cache a string we already had in memory," we built two more full copies of it, each bigger than the last.

## The fix: don't wrap the payload at all

Store it under **two SharedPreferences keys** instead of one:

```
Key "https://api.revenuecat.com/v1/offerings"
  → {"eTag":"abc123","responseCode":200,"origin":"CACHE",...}    ← tiny, ~200 bytes

Key "https://api.revenuecat.com/v1/offerings#rc_payload"
  → {"offerings":[{"id":"premium"}]}                             ← the ORIGINAL string, untouched
```

The metadata is small so escaping it costs nothing. The payload never goes through JSON escaping because it's not *inside* JSON anymore — it's its own entry. `putString` just stores the reference to the string we already have. **Zero new copies. The 36MB allocation never happens.**

And reading gets the same win in reverse: today, just to send the `X-RevenueCat-ETag` header on a request, we deserialize the whole blob — un-escaping twice, allocating a full payload copy — to pluck out `"abc123"`. With split keys, we read the tiny metadata entry and never touch the payload.

That's the whole idea: the payload was already in its final form the moment it arrived; the fix is to stop passing it through two rounds of packaging it never needed.

</details>

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

---

> \[!NOTE\]<br>**Medium Risk**<br>Touches every networked request’s ETag cache and changes on-disk format (SDK downgrade leaves stale split entries until cache clear), but behavior is heavily tested and failures degrade to cache misses/request errors rather than crashes.
>
> **Overview**<br>Fixes OOM when caching large `/offerings` responses by **splitting the ETag SharedPreferences cache**: URL key holds small `ETagCacheMetadata` JSON; `url#rc_payload` stores the response body **without nested JSON escaping** (the old `HTTPResultWithETag` double-wrap path could allocate tens of MB per store/read).
>
> `getETagHeaders` now reads metadata only (no payload). **304 cache hits** merge verification/request date via `HTTPResult.copy()` instead of mutating cached results. **Legacy combined entries** migrate to the split format on first read; corrupt or partial entries behave as cache misses.
>
> `HTTPResult` fields `origin`, `requestDate`, and `verificationResult` are `val` **again** because caching no longer flips `origin` in place.
>
> Adds `ETagManagerMemoryTest` as a regression gate (allocations stay under 1MB on a \~5MB payload). `ETagManagerTest` covers split storage, legacy migration, and edge cases.
>
> **Downgrade note:** older SDKs cannot read split entries until cache clear (documented tradeoff).
>
> Reviewed by [Cursor Bugbot](<https://cursor.com/bugbot>) for commit 526b341574b9ecb3ab4982a214105c6289323cfa. Bugbot is set up for automated code reviews on this repo. Configure [here](<https://www.cursor.com/dashboard/bugbot>).

---

> [!NOTE]
> **Medium Risk**
> Changes the on-disk ETag format and code path for every conditional request; failures degrade to cache misses rather than crashes, but SDK downgrades or legacy prefs on API &lt; 24 can leave stale or unreadable cache until clear.
> 
> **Overview**
> Fixes OOM when caching large API bodies (e.g. `/offerings`) by **replacing the monolithic `HTTPResultWithETag` SharedPreferences blob** with a **v2 prefs file** that stores only small **`ETagCacheMetadata` JSON** per URL, while response bodies go to a new **`ETagPayloadStore`** under `cacheDir` (one UTF-8 file per URL, streamed write + atomic rename, size-checked reads).
> 
> **`getETagHeaders`** now loads metadata only—no payload deserialize on every request. **304 handling** rebuilds `HTTPResult` from file + metadata via **`copy()`** instead of mutating cached results. **`HTTPResult`** drops prefs serialization and makes **`origin` / `requestDate` / `verificationResult`** immutable again.
> 
> On first use, **`_preferences_etags`** is **deleted without parsing** (API 24+); corrupt or partial entries behave as **cache misses** that self-heal via existing ETag refresh. Adds **`ETagManagerMemoryTest`** allocation gates and broad unit coverage for the payload store and manager edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50ff0d1f57047f26e9e979ad87337d88fd4d7c65. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
This makes sense, _but_ we're still storing all of the responses in a single `SharedPreferences`, which is then loaded into memory for the whole process life. We are still loading _much less_  data into memory but the next step here would be to push the responses to disk (a file each) instead. I'm gonna toy around with a separate PR for this

vegaro:
Does that mean that we hold in memory the SharedPreferences and the Offerings themselves for the whole process life?!? @AlvaroBrey

AlvaroBrey:
> Does that mean that we hold in memory the SharedPreferences and the Offerings themselves for the whole process life?!? @AlvaroBrey

As far as I know,  yes. as long as `ETagManager.prefs` lives (which it will for the entire process, just like `HTTPClient`), it's entire content will be in memory. That's how `SharedPreferences` works.

AlvaroBrey:
* **#3777** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3777?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3774** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -115,9 +231,19 @@ internal class ETagManager(
...
+        val metadata = ETagCacheMetadata.deserialize(serialized)
+        if (metadata != null) {
+            // Read without the monitor: a concurrent storeResult() landing between the two getString calls can
+            // pair the old metadata with the new payload for this one read. The window is nanoseconds, requires
+            // concurrent same-URL traffic, and the worst case is one response served with the previous entry's
+            // metadata fields — accepted to keep reads lock-free.
+            val payload = prefs.value.getString(payloadKey(urlString), null) ?: return null
+            return metadata.toHTTPResult(payload)
+        }
+        return migrateLegacyEntry(urlString, serialized)?.httpResult
```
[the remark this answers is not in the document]
AlvaroBrey: Accepted, this is a minor tradeoff (it will just be re-requested) to not be blocking.

AlvaroBrey (approved):
Absolutely insane improvement 🤯  this is amazing

#3777  moves the store to disk to prevent permanent allocation, but that's just a secondary improvement. Kudos!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -50,7 +56,115 @@ internal data class HTTPResultWithETag(
...
+ * under the URL key, while the payload is stored verbatim under a separate key ([ETagManager.payloadKey])
+ * so caching a large response never re-encodes a string we already hold in memory
+ * (https://github.com/RevenueCat/purchases-android/issues/3628).
+ *
+ * `origin` is intentionally not persisted: stored results always read back as [HTTPResult.Origin.CACHE].
+ *
+ * Downgrade note: older SDK versions cannot parse this format — after a downgrade, reads of split
+ * entries fail (as request errors, not crashes) and do not self-heal until the cache is cleared
+ * (e.g. identity change) since the failure happens before any response could overwrite the entry.
+ * Accepted tradeoff for keeping the URL as the metadata key; see PR #3774.
```
tonidero: Hmm I wonder if we should move the content to another file to avoid this downgrade issue... But maybe it would be addressed in the follow-up PR? Still reviewing 😅 

Also, we can probably clean this comment to make it less agentic :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -93,11 +207,13 @@ internal class ETagManager(
         )
         eTagHeader?.let { eTagInResponse ->
             if (shouldUseCachedVersion(responseCode)) {
-                val storedResult = getStoredResult(urlString)?.also { storedResult ->
-                    storedResult.requestDate = requestDate ?: storedResult.requestDate
-                    // This assumes we won't store verification failures in the cache and we will clear the cache
-                    // when enabling verification.
-                    storedResult.verificationResult = verificationResult
+                val storedResult = getStoredResult(urlString)?.let { storedResult ->
+                    storedResult.copy(
```
tonidero, line 177: I guess this reverts back the previous change we did... It's true it shouldn't be a problem anymore so I guess I'm ok with this 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -141,22 +267,48 @@ internal class ETagManager(
...
-        return serializedHTTPResultWithETag?.let {
-            HTTPResultWithETag.deserialize(it)
+    /**
+     * Pre-split entries embedded the payload (double-escaped) inside one combined JSON string. Rewrite
+     * them in the split format on first read so later reads never pay the unescape cost again, and
+     * return the parsed legacy entry. Unparseable data reads as a cache miss.
+     */
+    @Suppress("SwallowedException", "ReturnCount")
+    @Synchronized
+    private fun migrateLegacyEntry(urlString: String, serialized: String): HTTPResultWithETag? {
```
tonidero: I do wonder if we should just clear the cached etag once when migrating... TBH, feels like extra complexity... But well, since we have it, I think it should be ok.

tonidero (commented):
Looking good! Going to review the follow up PR before approving

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
vegaro: oh right, I think it might need rebase?

tonidero: I think that change was merged so possibly... but seems it was modified in this PR, not sure if there is an exact reason.

tonidero (approved):
I think this makes sense! Left some comments in the follow-up PR in case we want to wait until that is ready to merge this.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
vegaro: nah, I think it must have been some merge/rebase problem tbh

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: Using a new file makes sense I think, I'll do it in the follow-up

And the comments are also cleaned up in the follow up already 🥲

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: I believe this was intentional, see https://github.com/RevenueCat/purchases-android/pull/3774/changes/9671f05bcc76128bc7135f0100faa8b40a341515

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: So, rather than using a new file I've prefixed the new keys in the same file (https://github.com/RevenueCat/purchases-android/pull/3777/changes/937e0da28dd6d9bd1dd3e79b89a73997442f5004). This prevents downgrade issues but also doesn't have us with two SharedPreferences in memory and having to juggle priority between them.

It's essentially the same approach as before but rather than removing a key from a pref, we save it with a new prefix and delete the old one.

tonidero: Makes sense! I mentioned it before but it should be ok to clear the etag cache if needed, but this is better :) Thank you!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
tonidero: Ok I guess we're still modifying it, but as mentioned, the reason behind the original change is no more (we were decoding the original string back to a json on each copy 😬 So I'm not opposed to this.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: I've been giving this a think

- If we drop the old entries on migration, the user "pays" with another network request
- But if we _don't_ drop them, we have to read and then reserialize the cached entries (including the whole payload as before), risking OOM, especially with several sequential/parallel requests. Although I assume the risk of OOM here is low because it would have OOM on original serialization and never stored.

Currently I'm leaning more towards actually just removing the migration and dropping old entries altogether 🤷  opinions @vegaro ?

vegaro: I think I agree on droppign the cache. The extra complexity adds more breaking points. I would drop everything and start clean

tonidero: Yeah... feels cleaner + this is just an optimization and shouldn't be a huge deal to have to fetch the whole request again.

AlvaroBrey: Done (https://github.com/RevenueCat/purchases-android/pull/3774/changes/80da962f28f93e15853c31d9e9a3bfe8b3601613 and https://github.com/RevenueCat/purchases-android/pull/3777/changes/bdc4f9ce1cbf67b188a7f667a215deb70aacb83c) and merged #3777 into this one to ensure they're merged as a unit

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -176,6 +268,27 @@ internal class ETagManager(
...
+
+        /**
+         * Removes entries from older SDK formats (bare URL keys), which are dropped rather than
+         * migrated: their values would otherwise stay parsed in the prefs in-memory map for the
+         * process lifetime. Runs on the already-loaded key set and never parses a value; the edit
+         * only happens when something is stale.
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal fun removeLegacyEntries(prefs: SharedPreferences) {
+            val legacyKeys = prefs.all.keys.filterNot { it.startsWith(METADATA_KEY_PREFIX) }
```
tonidero: It could be worth moving this to a background thread so we don't need to read the whole shared preferences file in the main thread for this migration... But then again, I think this is okish considering metadata is much smaller... so I won't block this PR for this and we can optimize later :)

AlvaroBrey: The whole file read would be done here anyway, since we need to read the file for the metadata.

_but_ we can likely just use a new file tbh, and ignore the old one. That way we don't need key namespacing, don't care about deleting the old ones, or anything else. The price is just a dead SharedPrefs on disk but I think that should be okay. Now that we are dropping old values using a new file is basically free

tonidero: Yeah... agreed that should be ok IMO

AlvaroBrey: Done at https://github.com/RevenueCat/purchases-android/pull/3774/changes/eb4a9aebe8d280b21528e7573bd271085de222bf. Thanks for the suggestions!

In the end I did delete the old file since there's a useful method for it in SDK 24+, just left it laying around in sdk 23.

tonidero (approved):
Nice! Changes make sense!

PR #3775, opened 2026-07-15
Eagerly prewarm offering's workflow assets

vegaro:
Workflow images/fonts were only prewarmed at render time, right before a paywall shows. Legacy (non-workflow) offerings have always prewarmed the current offering's images/fonts eagerly, right after `/offerings` resolves. This closes that gap for workflows.

- `WorkflowManager.prewarmAssetsForOffering(offeringId)`: fire-and-forget, resolves the offering to its workflow and reuses `getWorkflow`'s existing resolve → fetch → prewarm path. Exits silently if the offering has no mapped workflow, so it doesn't spam an error log on every `getOfferings` for apps not using workflows yet.
- Wired into both `OfferingsManager` `dispatchSuccess` paths (fresh fetch and cache-vend), inside the `onPaywallConfigReady` completion so the offering→workflow lookup has synced topic data.

No public API change.

---

> [!NOTE]
> **Low Risk**
> Paywall asset prefetch timing and internal wiring only; failures are logged and retried without blocking offerings or workflow reads.
> 
> **Overview**
> Workflow paywall images and `ui_config` fonts are now pre-downloaded when remote config warms the **current offering’s** workflow, not only when `WorkflowManager.getWorkflow` runs at render time. That aligns workflow behavior with legacy offerings, which already prefetch the current offering’s assets after `/offerings`.
> 
> **`WorkflowAssetPreDownloader`** is replaced by **`WorkflowAssetPrewarmer`**, shared by the render path and a new load path. One `warmedWorkflowIds` set dedupes both paths; failed decode or missing `ui_config` does not mark a workflow warmed so the next commit can retry. The load path decodes via **`resolveWorkflowBody`** so the workflows cache stays byte-only (no eager `Lazy` decode).
> 
> **`WorkflowsConfigProvider`** takes optional **`onCurrentWorkflowLoaded`**, invoked fire-and-forget after `warm()` for the current offering’s workflow id—even if byte preload missed that body—using **`::resolveWorkflowBody`** as the transient decoder. **`PurchasesFactory`** constructs the prewarmer earlier, passes it into **`WorkflowManager`**, and wires **`workflowAssetPrewarmer::onCurrentWorkflowLoaded`** into the provider.
> 
> Tests move from **`WorkflowAssetPreDownloaderTest`** to **`WorkflowAssetPrewarmerTest`** and add **`WorkflowsConfigProvider`** coverage for the callback behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6ca7d9868ad87cdbebaae4e614030dbe2dbb101. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowAssetPrewarmer.kt
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /** Load-path callback for [WorkflowsConfigProvider]; see the class KDoc. */
+    suspend fun onCurrentWorkflowLoaded(
+        workflowId: String,
+        transientDecode: suspend (String) -> PublishedWorkflow?,
+    ) {
+        // Dedup before decoding, so a workflow the render path already warmed is never transiently decoded here.
+        val alreadyWarmed = synchronized(warmedWorkflowIds) { warmedWorkflowIds.contains(workflowId) }
+        if (alreadyWarmed) return
```
tonidero, line 57: Hmm I do wonder if we should just call prewarm again, and let it to the warmers to check if the assets are already downloaded (thinking of the case that something failed to load the first time), but NABD to be honest

tonidero (approved):
Looks great to me!

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
```diff
@@ -232,6 +239,21 @@ internal class WorkflowsConfigProvider(
...
+        // reads bytes + parses without touching the retained Lazy above, so prewarming keeps the cache
+        // raw-bytes-only. Prewarming is sequential, so announce the current offering's workflow first: it is the
+        // one most likely to be presented, so its assets should be cached before the prefetch-only workflows'.
+        onWorkflowsLoaded?.let { notify ->
+            val currentWorkflowId = currentOfferingId?.let { offeringToWorkflowId[it] }
+            val orderedIds = if (currentWorkflowId != null && workflows.containsKey(currentWorkflowId)) {
+                linkedSetOf(currentWorkflowId).apply { addAll(workflows.keys) }
+            } else {
+                workflows.keys
+            }
```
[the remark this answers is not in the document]
vegaro: I would expect this to be fixed already?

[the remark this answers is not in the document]
tonidero: Same... I was confused with that comment

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowAssetPrewarmer.kt
vegaro: so like not keep track of them? and just let Coil handle it? with the current code we save some CPU no?

tonidero: Hmm right... I thought we had some checks on the predownloaders but I think we don't... So yeah, this makes sense then, thank you!

PR #3777, opened 2026-07-15
fix: store ETag cache payloads in files instead of SharedPreferences

AlvaroBrey:
Stacked on #3774. **Must ship in the same release**: the intermediate payload format from that PR never ships, so devices migrate legacy entries directly to the file store.

#3774 stopped the ETag cache from re-encoding payloads, but they still lived in SharedPreferences, which keeps its whole file parsed in the app heap for the process lifetime.

This moves the payload to one file per URL under `cacheDir` (sha256-named), keeping the small metadata JSON in prefs:

- Writes stream through fixed-size buffers (no payload-sized allocations) to a temp file committed by an atomic checked rename.
- Writes are not fsynced, keeping disk latency off the request path. Instead, metadata records the payload byte size and reads verify it, so a file truncated by power loss reads as a miss and heals through the existing refresh retry. 
- The payload is read from disk only on 304 hits, decoded strictly so corrupt bytes read as a miss. Metadata reads (every request) stay in SharedPreferences.

## Measurements

5MB offerings-like payload. Allocations from `ETagManagerMemoryTest`; latency from an instrumented release build on a Xiaomi Mi 9T Pro (Snapdragon 855), medians of 10.

Memory, allocated per operation:

<details>
<summary> Click to expand </summary>

| Operation | main | #3774 | this PR |
|---|---|---|---|
| Store response | 49.3 MB | 151 KB | 190 KB |
| `getETagHeaders` (every request) | 46.8 MB | 2 KB | 3 KB |
| 304 cache hit | 46.8 MB | 3 KB | ~20 MB transient |
| **Retained in heap, process lifetime** | **~2x payload** | **~1x payload** | **~200 bytes** |

</details>

Latency, on device:

<details>

Measured on a release build, Xiaomi Mi 9T Pro, 5MB payload, median [min–p90] of 100 iterations:

| Operation | main | #3774 | this PR |
|---|---|---|---|
| Store | 735 ms [670–786] | 151 µs [142–196] | 16.6 ms [15.5–18.5] |
| Headers (every request) | 565 ms [550–631] | 130 µs [126–145] | 53 µs [42–142] |
| 304 hit | 564 ms [551–598] | 151 µs [139–514] | 29.5 ms [28.6–30.1] |
| Miss | 5 / 12 µs | 5 / 12 µs | 5 / 12 µs |

</details>

The 304 read rebuilding the payload from disk is the deliberate trade for zero permanent
retention. 

## Known limitation: SDK downgrade

Same as #3774: older SDKs cannot read the new format; entries fail as
request errors until the cache is cleared. Additionally, orphaned
payload files linger in `cacheDir` after a downgrade until the OS
purges them.

---

> [!NOTE]
> **Medium Risk**
> Changes core HTTP caching for all SDK network responses; mistakes could cause cache misses, stale data, or failed stores, but failures degrade to refetch rather than serving corrupt JSON.
> 
> **Overview**
> Moves **ETag response bodies** out of SharedPreferences into **per-URL files** under `cacheDir` (`ETagPayloadStore`), while **small metadata JSON** (eTag, status, verification, etc.) stays in prefs under **`v2:`-prefixed keys**. This targets [#3628](https://github.com/RevenueCat/purchases-android/issues/3628): large offerings payloads no longer sit in the prefs in-memory map for the process lifetime.
> 
> **`ETagManager`** writes the payload file first (streaming UTF-8 with fixed buffers, temp file + rename), records **`payloadSizeBytes`** in metadata, and only then persists metadata. **304 / cache hits** load the body from disk with a size check and strict decoding so truncated or corrupt files become cache misses that can heal via the existing refresh path. **`clearCaches`** clears prefs and payload files. On upgrade, **`removeLegacyEntries`** drops bare-URL and old `#rc_payload` prefs keys without parsing values.
> 
> New **`ETagPayloadStore`** tests and expanded **`ETagManager`** / memory regression tests cover the split store, legacy sweep, and allocation bounds (304 reads are allowed to allocate proportionally to payload size).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82768df2d1291f869bef9501160a4c4a785976d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
@@ -24,7 +24,8 @@ internal data class ETagData(
 /**
  * Everything the ETag cache persists about a response except its payload, stored as a small flat JSON
  * under [ETagManager.metadataKey]; the payload lives in an [ETagPayloadStore] file. The versioned key
- * keeps downgrades safe: older SDKs only read the bare URL key and simply refetch.
+ * keeps downgrades safe: older SDKs only read the bare URL key and simply refetch, and anything they
+ * wrote there is swept at the next upgraded launch ([ETagManager.removeLegacyEntries]).
  *
  * `origin` is intentionally not persisted: stored results always read back as [HTTPResult.Origin.CACHE].
  */
@@ -124,7 +125,9 @@ internal data class ETagCacheMetadata(
 @Suppress("TooManyFunctions")
 internal class ETagManager(
     context: Context,
-    private val prefs: Lazy<SharedPreferences> = lazy { initializeSharedPreferences(context) },
+    private val prefs: Lazy<SharedPreferences> = lazy {
+        initializeSharedPreferences(context).also(::removeLegacyEntries)
+    },
     private val dateProvider: DateProvider = DefaultDateProvider(),
     private val payloadStore: ETagPayloadStore = ETagPayloadStore(context),
 ) {
@@ -247,7 +250,6 @@ internal class ETagManager(
         return prefs.value.getString(metadataKey(urlString), null)?.let { ETagCacheMetadata.deserialize(it) }
     }
 
-
     private fun shouldStoreBackendResult(resultFromBackend: HTTPResult): Boolean {
         val responseCode = resultFromBackend.responseCode
         return responseCode != RCHTTPStatusCodes.NOT_MODIFIED &&
@@ -271,6 +273,22 @@ internal class ETagManager(
 
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
         internal fun metadataKey(urlString: String): String = METADATA_KEY_PREFIX + urlString
+
+        /**
+         * Removes entries from older SDK formats (bare URL keys), which are dropped rather than
+         * migrated: their values would otherwise stay parsed in the prefs in-memory map for the
+         * process lifetime. Runs on the already-loaded key set and never parses a value; the edit
+         * only happens when something is stale.
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal fun removeLegacyEntries(prefs: SharedPreferences) {
+            val legacyKeys = prefs.all.keys.filterNot { it.startsWith(METADATA_KEY_PREFIX) }
+            if (legacyKeys.isNotEmpty()) {
+                val editor = prefs.edit()
+                legacyKeys.forEach(editor::remove)
+                editor.apply()
+            }
+        }
 
         fun initializeSharedPreferences(context: Context): SharedPreferences =
             context.getSharedPreferences(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
@@ -22,9 +22,10 @@ internal data class ETagData(
 
 /**
  * Everything the ETag cache persists about a response except its payload. Stored as a small flat JSON
- * under the URL key, while the payload is stored verbatim under a separate key ([ETagManager.payloadKey])
- * so caching a large response never re-encodes a string we already hold in memory
- * (https://github.com/RevenueCat/purchases-android/issues/3628).
+ * under the URL key, while the payload is stored verbatim in a file ([ETagPayloadStore]) so caching a
+ * large response never re-encodes a string we already hold in memory
+ * (https://github.com/RevenueCat/purchases-android/issues/3628) nor retains it in the SharedPreferences
+ * in-memory map for the process lifetime.
  *
  * `origin` is intentionally not persisted: stored results always read back as [HTTPResult.Origin.CACHE].
  *
@@ -126,6 +127,7 @@ internal class ETagManager(
     context: Context,
     private val prefs: Lazy<SharedPreferences> = lazy { initializeSharedPreferences(context) },
     private val dateProvider: DateProvider = DefaultDateProvider(),
+    private val payloadStore: ETagPayloadStore = ETagPayloadStore(context),
 ) {
 
     internal fun getETagHeaders(
@@ -193,14 +195,32 @@ internal class ETagManager(
         val serialized = prefs.value.getString(urlString, null) ?: return null
         val metadata = ETagCacheMetadata.deserialize(serialized)
         if (metadata != null) {
-            // Read without the monitor: a concurrent storeResult() landing between the two getString calls can
-            // pair the old metadata with the new payload for this one read. The window is nanoseconds, requires
+            // Read without the monitor: a concurrent storeResult() landing between the metadata read and the
+            // payload read can pair the old metadata with the new payload for this one read. It requires
             // concurrent same-URL traffic, and the worst case is one response served with the previous entry's
             // metadata fields — accepted to keep reads lock-free.
-            val payload = prefs.value.getString(payloadKey(urlString), null) ?: return null
+            val payload = readPayload(urlString) ?: return null
             return metadata.toHTTPResult(payload)
         }
         return null
+    }
+
+    private fun readPayload(urlString: String): String? {
+        return payloadStore.read(urlString) ?: migratePrefsPayload(urlString)
+    }
+
+    /**
+     * An earlier release stored the payload under a second SharedPreferences key instead of a file.
+     * Move such payloads to the file store on first read; the prefs key is only removed once the file
+     * write is durable, so a failed write retries on the next read.
+     */
+    @Synchronized
+    private fun migratePrefsPayload(urlString: String): String? {
+        val prefsPayload = prefs.value.getString(payloadKey(urlString), null) ?: return null
+        if (payloadStore.write(urlString, prefsPayload)) {
+            prefs.value.edit().remove(payloadKey(urlString)).apply()
+        }
+        return prefsPayload
     }
 
     internal fun storeBackendResultIfNoError(
@@ -215,7 +235,10 @@ internal class ETagManager(
 
     @Synchronized
     internal fun clearCaches() {
+        // Prefs (metadata) first: a crash in between leaves orphan payload files, which are harmless and
+        // overwritten later, rather than metadata pointing at purged payloads.
         prefs.value.edit().clear().apply()
+        payloadStore.clear()
     }
 
     @Synchronized
@@ -225,19 +248,23 @@ internal class ETagManager(
         eTag: String,
     ) {
         val metadata = ETagCacheMetadata.fromResult(result, ETagData(eTag, dateProvider.now))
-        prefs.value.edit()
-            .putString(urlString, metadata.serialize())
-            // The payload is stored verbatim under its own key: embedding it inside the metadata JSON
-            // would re-escape a multi-MB string we already hold, which OOMed on large /offerings
-            // responses (https://github.com/RevenueCat/purchases-android/issues/3628).
-            .putString(payloadKey(urlString), result.payloadText)
-            .apply()
+        // Payload file first, and metadata only once that write is durable: metadata present must imply
+        // the payload was written. The payload lives in a file rather than in the prefs JSON because
+        // re-escaping a multi-MB string OOMed on large /offerings responses
+        // (https://github.com/RevenueCat/purchases-android/issues/3628), and SharedPreferences would keep
+        // it parsed in the app heap for the process lifetime.
+        if (payloadStore.write(urlString, result.payloadText)) {
+            prefs.value.edit()
+                .putString(urlString, metadata.serialize())
+                .apply()
+        }
     }
 
     private fun getStoredMetadata(urlString: String): ETagCacheMetadata? {
         val serialized = prefs.value.getString(urlString, null) ?: return null
         return ETagCacheMetadata.deserialize(serialized)
     }
+
 
     private fun shouldStoreBackendResult(resultFromBackend: HTTPResult): Boolean {
         val responseCode = resultFromBackend.responseCode
@@ -256,8 +283,10 @@ internal class ETagManager(
     }
 
     companion object {
-        // Cache keys are endpoint URLs from URL(baseURL, path).toString(), which never carry a '#'
-        // fragment, so this suffix cannot collide with another URL's entry.
+        // Where an earlier release kept the payload in SharedPreferences. Only read (and removed) by
+        // [migratePrefsPayload]; new payloads go to [ETagPayloadStore]. Cache keys are endpoint URLs
+        // from URL(baseURL, path).toString(), which never carry a '#' fragment, so this suffix cannot
+        // collide with another URL's entry.
         private const val PAYLOAD_KEY_SUFFIX = "#rc_payload"
 
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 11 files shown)


AlvaroBrey:
~~Discussion point for the team: this PR keeps #3774's legacy migration (now rewriting the payload into a file, preserving the eTag so the first request after upgrade gets a 304 instead of re-downloading the full body — 15MB for the app in #3628) The alternative is dropping legacy entries on first read: it deletes the migration machinery (the synchronized re-read guard and its races) at the cost of one full re-download per URL per device after the upgrade.~~


Edit: I've given this some thought: the migration only runs once per request (and runs per-request, not all at once), so I think it's an okay price to pay to avoid extra network requests/breakage in low-network scenarios

AlvaroBrey:
@cursor review

AlvaroBrey:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3777?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3777** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3777?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3777?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3774** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3774?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

AlvaroBrey:
@cursor review

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -258,7 +273,11 @@ internal class ETagManager(
 
     @Synchronized
     internal fun clearCaches() {
-        prefs.value.edit().clear().apply()
```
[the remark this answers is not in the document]
AlvaroBrey: Intentional: the split-prefs layout never shipped in any release, it only exists on this unreleased stack. (If the release plan changes and #3774 ships alone first, this needs revisiting.)

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -258,7 +258,10 @@ internal class ETagManager(
 
     @Synchronized
     internal fun clearCaches() {
```
vegaro, line 219: Looks like this does blocking disk I/O on the main thread when called from `configure`

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -297,17 +307,22 @@ internal class ETagManager(
             HTTPResultWithETag.deserialize(serialized)
         } catch (e: JSONException) {
             return null
+        } catch (e: IllegalArgumentException) {
+            // Corrupt enum values: a cache miss like any other unparseable entry.
+            return null
         }
         // Re-read under the monitor: a concurrent clearCaches() or storeResult() may have removed or
-        // replaced this entry after the (unsynchronized) caller captured it — writing the stale capture
+        // replaced this entry after the (unsynchronized) caller captured it; writing the stale capture
```
vegaro: I literally have this rule as well haha

purchases/src/test/java/com/revenuecat/purchases/common/networking/ETagManagerMemoryTest.kt
```diff
@@ -95,12 +95,15 @@ class ETagManagerMemoryTest {
...
+        // Regression gates for #3628. Store and header reads must not allocate anywhere near payload size
+        // (the legacy combined format allocated tens of MB per operation on this 5MB payload; the store's
+        // encoder writes through a fixed buffer). The 304 read rebuilds the payload string from its file, so
+        // its cost is payload-proportional by design (the deliberate tradeoff for not retaining the payload
+        // in the SharedPreferences in-memory map for the process lifetime) but bounded to a small multiple.
         val maxAllowedBytes = 1024L * 1024L
         assertThat(storeBytes).isLessThan(maxAllowedBytes)
         assertThat(headerBytes).isLessThan(maxAllowedBytes)
-        assertThat(notModifiedBytes).isLessThan(maxAllowedBytes)
+        assertThat(notModifiedBytes).isLessThan(3L * payload.length * Char.SIZE_BYTES)
```
vegaro, line 109: don't compoletely understand the `3L` but could this flake?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: Good catch, this was a regression from a review round (commit() for a crash-ordering guarantee the code doesn't need: the in-memory clear already makes readers miss before touching files). Back to apply() in 891923699. payloadStore.clear() still runs on the caller thread; it's metadata-only unlinks, but happy to move it off-thread if you'd rather.

purchases/src/test/java/com/revenuecat/purchases/common/networking/ETagManagerMemoryTest.kt
AlvaroBrey: Documented in 891923699.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -258,7 +258,10 @@ internal class ETagManager(
 
     @Synchronized
     internal fun clearCaches() {
+        // Metadata first: its in-memory clear is immediate, so readers miss before ever touching files.
+        // A crash before the async disk flush leaves metadata without payloads, a self-healing miss.
         prefs.value.edit().clear().apply()
+        payloadStore.clear()
```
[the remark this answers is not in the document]
AlvaroBrey: Fixed in the follow-up commit: clear() now renames the directory (constant-time) and the next write, always on a background thread, deletes the trash.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: Follow-up: the file deletion is off the caller thread too now (clear() renames the directory, the next background write deletes the trash).

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -0,0 +1,157 @@
...
+        val file = fileFor(urlString)
+        return try {
+            if (expectedSizeBytes != null && file.length() != expectedSizeBytes) {
+                return null
+            }
+            // Strict decoding (unlike String(bytes), which silently substitutes U+FFFD) so a corrupt file
+            // reads back as a cache miss instead of serving garbage as a valid cached response.
+            Charsets.UTF_8.newDecoder()
+                .decode(ByteBuffer.wrap(file.readBytes()))
+                .toString()
```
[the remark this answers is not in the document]
AlvaroBrey: Behavior was already REPORT: CharsetDecoder.decode() uses the decoder's defaults, which are REPORT (the REPLACE default belongs to Charset.decode/String(bytes)). The corrupt-file test proves it. Made the config explicit anyway so it stops reading as REPLACE.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
```diff
@@ -78,6 +76,8 @@ internal data class ETagCacheMetadata(
     val verificationResult: VerificationResult,
     val isLoadShedderResponse: Boolean,
     val isFallbackURL: Boolean,
+    /** Verified on read to detect truncated payload files; `null` skips the check. */
+    val payloadSizeBytes: Long? = null,
```
tonidero, line 43: Instead of the bytes size, should we keep a hash to verify the contents?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -0,0 +1,159 @@
...
+            val result = step()
+            if (result.isError) result.throwException()
+            out.write(byteBuffer.array(), 0, byteBuffer.position())
+            byteBuffer.clear()
+            if (result.isUnderflow) return
+        }
+    }
+
+    private companion object {
+        const val DIRECTORY_NAME = "rc_etag_payloads"
```
tonidero: Maybe we can put this within a `RevenueCat` folder? NABD

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -0,0 +1,159 @@
...
+                return null
+            }
+        }
+        val file = fileFor(urlString)
+        // Never opened by readers; a crash mid-write leaves one orphan that the next write overwrites.
+        // Same-URL writes are serialized by [ETagManager].
+        val tempFile = File(directory, file.name + TEMP_SUFFIX)
+        return try {
+            FileOutputStream(tempFile).use { out -> encodeTo(out, payload) }
+            if (tempFile.renameTo(file)) file.length() else null
```
tonidero, line 56: Should we use an `AtomicFile` for this write?

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -0,0 +1,159 @@
...
+        if (!directory.exists()) return
+        val trash = File(directory.parentFile, directory.name + TRASH_SUFFIX + System.nanoTime())
+        if (!directory.renameTo(trash)) {
+            directory.deleteRecursively()
+        }
+    }
+
+    private fun deleteTrash() {
+        directory.parentFile
+            ?.listFiles { file -> file.name.startsWith(directory.name + TRASH_SUFFIX) }
```
tonidero, line 109: If we don't put this inside a `RevenueCat` folder or something like that, we could potentially delete files from the developer that end with `.trash` I think

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -0,0 +1,159 @@
...
+internal class ETagPayloadStore(
+    private val directory: File,
+) {
+    constructor(context: Context) : this(File(context.cacheDir, DIRECTORY_NAME))
+
+    /**
+     * Returns the payload's size in bytes once the file is in place, or `null` when the write failed;
+     * callers must not persist metadata for a failed write.
+     */
+    fun write(urlString: String, payload: String): Long? {
```
tonidero, line 42: If we do end up returning a hash, I think we can potentially calculate as we stream the output if I remember correctly.

tonidero (commented):
Looking very nice!! I think this is pretty close just a couple suggestions of using a checksum + putting these files within a `RevenueCat` folder. Lmk what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
AlvaroBrey: Done in be38ecc2b: cacheDir/RevenueCat/etag_payloads.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
AlvaroBrey: Good point, fixed by the RevenueCat folder in be38ecc2b: the trash sweep now only lists a RevenueCat-owned directory.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
AlvaroBrey: Deliberately not (class KDoc has the summary): our reads are lock-free, and AtomicFile's openRead deletes a concurrent writer's in-flight .new file while finishWrite reports the resulting rename failure only to logcat, which pairs new metadata with an old payload durably. The temp+checked-rename here has neither issue.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
tonidero: Hmm thinking about it again, maybe verifying a hash for big responses could add some latency... I'm mostly trying to avoid possible corruption though so I'm a bit on a back and forth here 😅. I guess I won't fight too strongly for it :P cc @vegaro in case you have any thoughts

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
tonidero: Ohhh makes sense. Thank you for explaining!!

tonidero (approved):
I will approve this since I won't block the PR regarding using a hash... I think it should be ok. Great job!!

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: Nobody likes emdashes 😆

purchases/src/test/java/com/revenuecat/purchases/common/networking/ETagManagerMemoryTest.kt
AlvaroBrey: To clarify (a rogue Claude agent left the previous comment, sorry): this needed to change because the etag read  _does_ allocate memory now (since it's not always in memory in SharedPrefs). 3x the payload size is an estimate of the acceptable maximum since we don't stream so we have at least 1x string + 1x binary in memory.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
AlvaroBrey: I started this with a CRC32 hash but decided against it because the chance we have a failed write that is exactly the same size in bytes as a real one (which is the only case a hash would matter) is tiny I think, not enough to justify the hash on every read

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
AlvaroBrey: 🤦 that was a Claude agent that escaped containment, sorry. but the explanation is mostly right, AtomicFile is not intended for concurrent use and we have no explicit locks for latency

vegaro (approved):
Amazing job. I am super excited about these changes

PR #3757, opened 2026-07-13
feat(paywalls): web_view component (PWENG-98)

JZDesign:
## Summary

Consolidates the Paywalls V2 `web_view` feature into a single PR. This supersedes the stacked draft PRs [#3650](https://github.com/RevenueCat/purchases-android/pull/3650)–[#3729](https://github.com/RevenueCat/purchases-android/pull/3729).
For https://linear.app/revenuecat/issue/PWENG-98
A `web_view` component embeds hosted web content inside a paywall:

1. **Schema** (`:purchases`) — `WebViewComponent` with URL, optional `id` / `protocol_version`, and size (no native fallback stack)
2. **Compose rendering** — `WebViewComponentView` with fit placeholders, HTTPS-only URL resolution; terminal load/renderer failures render nothing
3. **Envelope / value types** — message kinds (`connect` / `init` / `reject` / `message` / `request` / `response` / `error`), flat `payload` for `rc:variables`, non-finite numbers → JSON null
4. **JavaScript bridge** — handshake at hard-coded host protocol version `1`, origin validation, resize / variables messaging
5. **Paywall wiring** — `PaywallOptions.setWebViewMessageHandler` and `PaywallDialogOptions.setWebViewMessageHandler`

### No native fallback (by design)

`web_view` does **not** support a native `fallback` stack. Older/dashboard payloads that still include `fallback` are ignored at decode time (`ignoreUnknownKeys`). On terminal failure the WebView is destroyed and nothing is rendered — apps must use an SDK that supports `web_view`.

### Lifecycle & security hardening (follow-up on this PR)

| Fix | Behavior |
|---|---|
| **Document handshake reset** | `onMainFrameNavigationStarted` (from `onPageStarted`) closes the channel, clears fit/resize thresholds, resets Compose dimensions, and drops outbound work queued for prior document generations so same-origin navigation / reload can `connect` again |
| **Immutable WebView identity** | Compose `key(WebViewIdentity(url, componentId, widthFit, heightFit))` recreates WebView + bridge when those fields change; locale / message handler still update in place; bridge holders are identity-scoped so old `onRelease` cannot release a replacement bridge |
| **Renderer termination** | `onRenderProcessGone` returns `true`, marks an idempotent terminal failure, destroys the dead WebView, and renders nothing |
[...]

PR #3780, opened 2026-07-15
feat(paywalls): internal types for web_view messaging (1/9)

JZDesign:
### Checklist

### Motivation
Part **1 of 8** splitting #3757 (`cursor/paywalls-web-view-consolidated-a396`) into smaller landable PRs. Lands the internal messaging types for Paywalls V2 `web_view` as **inert** code — nothing registers or activates the component until Part 8.

### Description
Adds SDK-internal models used to exchange messages with hosted web content inside a paywall:

- `PaywallWebViewValue` — typed JSON-compatible values
- `PaywallWebViewMessage` / `PaywallWebViewController` / `PaywallWebViewMessageHandler`
- Unit tests for `PaywallWebViewValue`
- Keeps `ui/revenuecatui/api.txt` unchanged; these types are not public API

**Inert until Part 8.** Until activation, a paywall JSON containing `web_view` continues to hit the existing unknown-type fallback on `main`.

Verified: `:ui:revenuecatui:testDefaultsDebugUnitTest` (PaywallWebViewValueTest), `detektAll`, `./scripts/api-check.sh`.

**Labels:** `pr:feat`, `pr:RevenueCatUI`, `feat:Paywalls_V2`

### Series (splitting #3757)

PWENG-98

| Part | PR | Notes |
|------|-----|--------|
| 1/9 | #3780 | main |
| 2/9 | #3784 | ~~dropped~~ — app-facing message handler removed from scope |
| 3/9 | #3781 | independent |
| 4/9 | #3782 | independent |
| 5/9 | #3783 | independent |
| 6/9 | #3796 | after PRs 1+4; replaces half of #3785 |
| 7/9 | #3798 | bridge; stacks on 6 |
| 8/9 | #3786 | after PRs 3+7 |
| 9/9 | #3787 | activation |

> Series is being converted to a linear Graphite stack: main → 1 → 3 → 4 → 5 → 6 → 7 → 8 → 9 (part 2 dropped).

<div><a href="https://cursor.com/agents/bc-8f8d68f5-c4d5-4bd2-93a1-41b9f6e72c86"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8f8d68f5-c4d5-4bd2-93a1-41b9f6e72c86"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
@@ -18,11 +18,11 @@ import dev.drewhamilton.poko.Poko
  * @property error The error description for a `rc:error` message, if any.
  */
 @Poko
-public class PaywallWebViewMessage(
-    public val componentId: String,
-    public val type: String,
-    public val responses: Map<String, PaywallWebViewValue>? = null,
-    public val error: String? = null,
+internal class PaywallWebViewMessage(
+    val componentId: String,
+    val type: String,
+    val responses: Map<String, PaywallWebViewValue>? = null,
+    val error: String? = null,
 )
 
 /**
@@ -32,7 +32,7 @@ public class PaywallWebViewMessage(
  * Outgoing messages preserve the RevenueCat web view postMessage envelope and are validated before
  * being delivered to the web view.
  */
-public interface PaywallWebViewController {
+internal interface PaywallWebViewController {
 
     /**
      * Sends a `rc:variables` message into the web view targeting [componentId]. The SDK already replies
@@ -41,7 +41,7 @@ public interface PaywallWebViewController {
      * `variables` key). Reserved SDK-managed top-level keys cannot be overwritten; provide
      * app-specific values under `custom`.
      */
-    public fun postVariables(
+    fun postVariables(
         componentId: String,
         variables: Map<String, PaywallWebViewValue>,
     )
@@ -51,7 +51,7 @@ public interface PaywallWebViewController {
      * [variables] map is sent as the transport envelope's `payload` field (not nested under a
      * `variables` key). Use [postVariables] for the common `rc:variables` case.
      */
-    public fun postMessage(
+    fun postMessage(
         componentId: String,
         type: String,
         variables: Map<String, PaywallWebViewValue>,
@@ -85,8 +85,8 @@ public interface PaywallWebViewController {
  *     .build()
  * ```
  */
-public fun interface PaywallWebViewMessageHandler {
-    public fun onMessage(
+internal fun interface PaywallWebViewMessageHandler {
+    fun onMessage(
         message: PaywallWebViewMessage,
         controller: PaywallWebViewController,
     )
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
@@ -1,0 +1,93 @@
+package com.revenuecat.purchases.ui.revenuecatui
+
+import dev.drewhamilton.poko.Poko
+
+/**
+ * A validated message received from a Paywalls V2 `web_view` component.
+ *
+ * Messages follow the RevenueCat web view postMessage envelope. Known [type]s include:
+ * - `rc:step-loaded`
+ * - `rc:step-complete` (carries [responses])
+ * - `rc:request-variables`
+ * - `rc:error` (carries [error])
+ *
+ * @property componentId The canonical component id (the `web_view.id` from the paywall schema). This
+ * matches the id API consumers see in the paywall configuration.
+ * @property type The message type, e.g. `rc:step-complete`.
+ * @property responses The structured responses for a `rc:step-complete` message, if any.
+ * @property error The error description for a `rc:error` message, if any.
+ */
+@Poko
+public class PaywallWebViewMessage(
+    public val componentId: String,
+    public val type: String,
+    public val responses: Map<String, PaywallWebViewValue>? = null,
+    public val error: String? = null,
+)
+
+/**
+ * Lets app code send messages back into a Paywalls V2 `web_view` component, for example in response to
+ * a `rc:request-variables` message.
+ *
+ * Outgoing messages preserve the RevenueCat web view postMessage envelope and are validated before
+ * being delivered to the web view.
+ */
+public interface PaywallWebViewController {
+
+    /**
+     * Sends a `rc:variables` message into the web view targeting [componentId]. The SDK already replies
+     * with SDK-managed variables (such as `locale`); use this to provide additional values. The flat
+     * [variables] map is sent as the transport envelope's `payload` field (not nested under a
+     * `variables` key). Reserved SDK-managed top-level keys cannot be overwritten; provide
+     * app-specific values under `custom`.
+     */
+    public fun postVariables(
+        componentId: String,
+        variables: Map<String, PaywallWebViewValue>,
+    )
+
+    /**
+     * Sends a message of the given [type] into the web view targeting [componentId]. The flat
+     * [variables] map is sent as the transport envelope's `payload` field (not nested under a
+     * `variables` key). Use [postVariables] for the common `rc:variables` case.
+     */
+    public fun postMessage(
+        componentId: String,
+        type: String,
+        variables: Map<String, PaywallWebViewValue>,
+    )
+}
+
+/**
+ * Receives validated messages from Paywalls V2 `web_view` components.
+ *
+ * Set this on [PaywallOptions.Builder.setWebViewMessageHandler]. The handler is always invoked on the
+ * main thread. The app decides what to do with each message: the SDK does not automatically dismiss the
+ * paywall or trigger a purchase in response to `rc:step-complete`.
+ *
+ * ### Usage
+ * ```kotlin
+ * PaywallOptions.Builder { /* dismiss */ }
+ *     .setWebViewMessageHandler { message, controller ->
+ *         when (message.type) {
+ *             "rc:request-variables" -> controller.postVariables(
+ *                 componentId = message.componentId,
+ *                 variables = mapOf(
+ *                     "custom" to PaywallWebViewValue.Object(
+ *                         mapOf("app_segment" to PaywallWebViewValue.String("high_intent")),
+ *                     ),
+ *                 ),
+ *             )
+ *             "rc:step-complete" -> { /* read message.responses, navigate, log analytics, etc. */ }
+ *             "rc:error" -> { /* log message.error */ }
+ *         }
+ *     }
+ *     .build()
+ * ```
+ */
+public fun interface PaywallWebViewMessageHandler {
+    public fun onMessage(
+        message: PaywallWebViewMessage,
+        controller: PaywallWebViewController,
+    )
+}
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewValue.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewValue.kt
@@ -12,14 +12,13 @@ import org.json.JSONObject
  * @see PaywallWebViewMessage
  * @see PaywallWebViewController
  */
-// Modeled as an abstract class with a closed set of subtypes (mirroring [CustomVariableValue]) rather
-// than a sealed class, to stay binary-compatible as public API.
-public abstract class PaywallWebViewValue internal constructor() {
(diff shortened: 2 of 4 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewValue.kt
```diff
@@ -0,0 +1,144 @@
...
+     * Non-finite values (`NaN`, infinities) cannot be represented in JSON and serialize as JSON
+     * `null` when sent into the web view. Equality follows [Double.equals] semantics (consistent
+     * with [hashCode]): `NaN` equals `NaN`, and `-0.0` does not equal `0.0`.
+     */
+    public class Number(public val value: kotlin.Double) : PaywallWebViewValue() {
+        public constructor(value: kotlin.Int) : this(value.toDouble())
+        public constructor(value: kotlin.Long) : this(value.toDouble())
+        public constructor(value: kotlin.Float) : this(value.toDouble())
+
+        override fun equals(other: Any?): kotlin.Boolean = other is Number && value.equals(other.value)
```
AlvaroBrey, line 39: IINM all of these clases could use `Poko` instead of handrolling equals/hashcode/tostring

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+import dev.drewhamilton.poko.Poko
+
+/**
+ * A validated message received from a Paywalls V2 `web_view` component.
+ *
+ * Messages follow the RevenueCat web view postMessage envelope. Known [type]s include:
+ * - `rc:step-loaded`
+ * - `rc:step-complete` (carries [responses])
+ * - `rc:request-variables`
+ * - `rc:error` (carries [error])
```
AlvaroBrey, line 12: If we're exposing these strings and letting the apps match on them, we should place them as public `const val`s for easier use

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @property componentId The canonical component id (the `web_view.id` from the paywall schema). This
+ * matches the id API consumers see in the paywall configuration.
+ * @property type The message type, e.g. `rc:step-complete`.
+ * @property responses The structured responses for a `rc:step-complete` message, if any.
+ * @property error The error description for a `rc:error` message, if any.
+ */
+@Poko
+public class PaywallWebViewMessage(
```
AlvaroBrey: I wonder whether we should spec this as different types (or just primitives) and have more than one callback in the listener instead. Right now this class is sparse, we'll never have error at the same type as responses, and `loaded` and `request-variables` have neither

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Sends a message of the given [type] into the web view targeting [componentId]. The flat
+     * [variables] map is sent as the transport envelope's `payload` field (not nested under a
+     * `variables` key). Use [postVariables] for the common `rc:variables` case.
+     */
+    public fun postMessage(
+        componentId: String,
+        type: String,
+        variables: Map<String, PaywallWebViewValue>,
+    )
```
AlvaroBrey: Under which circumstances is `postMessage` expected to be used from the app? Right now it looks like it's granting access to internals the app shouldn't have: if our only supported usecase for now is posting variables, expose `postVariables` only.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewValue.kt
```diff
@@ -0,0 +1,144 @@
...
+                is JSONArray -> {
+                    if (remainingDepth <= 0) return null
+                    val list = ArrayList<PaywallWebViewValue>(value.length())
+                    for (index in 0 until value.length()) {
+                        list.add(fromJson(value.get(index), remainingDepth - 1) ?: return null)
+                    }
+                    Array(list)
+                }
+                // Numbers stored by org.json as BigDecimal/BigInteger, or any other type, are not supported.
+                else -> (value as? kotlin.Number)?.let { Number(it.toDouble()) }
```
AlvaroBrey, line 112: If a BigDecimal/BigInteger reaches here it will be converted lossily and produce potentially different values than it meant to. I'd argue it's better to just fail early, or at the very least return `null` and log an error

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewValue.kt
```diff
@@ -0,0 +1,144 @@
...
+    public class Boolean(public val value: kotlin.Boolean) : PaywallWebViewValue() {
+        override fun equals(other: Any?): kotlin.Boolean = other is Boolean && value == other.value
+        override fun hashCode(): Int = value.hashCode()
+        override fun toString(): kotlin.String = "PaywallWebViewValue.Boolean(value=$value)"
+    }
+
+    /**
+     * A JSON object: a map of string keys to [PaywallWebViewValue]s.
+     */
+    public class Object(public val value: Map<kotlin.String, PaywallWebViewValue>) : PaywallWebViewValue() {
```
AlvaroBrey: `Object` and `Array` should probably copy their constructor parameters instead of keeping the original ref to prevent aliasing

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
AlvaroBrey: Actually see [this other comment](https://github.com/RevenueCat/purchases-android/pull/3780#discussion_r3597195608) as well

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
+package com.revenuecat.purchases.ui.revenuecatui
+
+import dev.drewhamilton.poko.Poko
+
+/**
+ * A validated message received from a Paywalls V2 `web_view` component.
+ *
+ * Messages follow the RevenueCat web view postMessage envelope. Known [type]s include:
+ * - `rc:step-loaded`
+ * - `rc:step-complete` (carries [responses])
```
AlvaroBrey, line 10: Why is `step-complete` here, isn't that a Workflows thing?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
AlvaroBrey: to be clear I mean something like:

```kt
public abstract class PaywallWebViewMessageHandler {
    public fun onLoaded(
        componentId: String,
        controller: PaywallWebViewController,
    ) {}
   public fun onError(
        componentId: String,
       error: String?()
        controller: PaywallWebViewController,
    ) {}
   public fun onRequestVariables(
        componentId: String,
        controller: PaywallWebViewController,
    ) {}
}
```

(abstract instead of interface to prevent breaking if we add more types)

AlvaroBrey (commented):
My first doubt is is: do we actually need this to be public? I thought for V1 we weren't going to have apps pass variables into the web components, and for internal SDK usage (sizing or whatever else) internal visibility should be enough.

If we don't have an immediate use case for public visibility, it doesn't really pay off to publish the API, since then we would be tied to it.

Either way I've reviewed below assuming we actually need the public visibility

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+ * - `rc:request-variables`
+ * - `rc:error` (carries [error])
+ *
+ * @property componentId The canonical component id (the `web_view.id` from the paywall schema). This
+ * matches the id API consumers see in the paywall configuration.
+ * @property type The message type, e.g. `rc:step-complete`.
+ * @property responses The structured responses for a `rc:step-complete` message, if any.
+ * @property error The error description for a `rc:error` message, if any.
+ */
+@Poko
```
AlvaroBrey, line 20: Does not need `@Poko` since it's internal.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+internal class PaywallWebViewMessage(
```
AlvaroBrey, line 21: nit: this class is unnecesssarily sparse 
- `error` will never carry `responses`
- `step-complete` will never carry `error`
-  the rest will never carry either

IMO this should be a sealed class for internal use. If we ever want to expose this in the API I'd argue that different callbacks would be warranted to avoid leaking the internal model

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+     * Sends a message of the given [type] into the web view targeting [componentId]. The flat
+     * [variables] map is sent as the transport envelope's `payload` field (not nested under a
+     * `variables` key). Use [postVariables] for the common `rc:variables` case.
+     */
+    fun postMessage(
+        componentId: String,
+        type: String,
+        variables: Map<String, PaywallWebViewValue>,
+    )
+}
```
AlvaroBrey, line 59: IINM, this entire interface is unused save for tests? `WebViewJavaScriptBridge`  implements it but only tests call it.

If it's useful for test mocking purposes, please rephrase the documentation to remove the refs to "let app code send messages" etc

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewMessage.kt
```diff
@@ -0,0 +1,93 @@
...
+ *                 ),
+ *             )
+ *             "rc:step-complete" -> { /* read message.responses, navigate, log analytics, etc. */ }
+ *             "rc:error" -> { /* log message.error */ }
+ *         }
+ *     }
+ *     .build()
+ * ```
+ */
+internal fun interface PaywallWebViewMessageHandler {
```
AlvaroBrey, line 88: Same, no implementers outside tests, dead code

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/PaywallWebViewValue.kt
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * A JSON-compatible value used in messages exchanged with a Paywalls V2 `web_view` component.
+ *
+ * Only JSON-like values are supported: [String], [Number], [Boolean], [Object], [Array] and [Null].
+ * Functions, binary blobs, dates, and platform-specific objects are intentionally not representable.
+ *
+ * @see PaywallWebViewMessage
+ * @see PaywallWebViewController
+ */
+// Modeled as an abstract class with a closed set of subtypes (mirroring [CustomVariableValue]).
```
AlvaroBrey, line 15: Similar comment as https://github.com/RevenueCat/purchases-android/pull/3780/changes#r3602221302: since this is now for internal use, it can just be a sealed class of data classes and greatly simplified. Or omitted altogether

AlvaroBrey:
Superseded and dropped from the series. The `web_view` messaging value types (`PaywallWebViewValue`) this PR introduced were removed when the app-facing message handler was cut for v1; the only remaining SDK->content payload (locale) is now a plain `Map<String, String>` handled in the 6a/6b PRs. Part 1 no longer has any content.

PR #3790, opened 2026-07-16
Open the default paywall for offerings without a workflow

vegaro:
With workflows enabled, offerings are parsed without their paywall components, and every non-legacy paywall is served through the workflows path. An offering with no workflow, or one that can't be served, would fall through and fail rendering.

Resolving an offering to its workflow now distinguishes four outcomes, each recovering differently:

- has a workflow → serve it
- has no workflow → render its regular (via offering) or default paywall
- config endpoint disabled by a 4xx → reload `/offerings` (which re-parse with the previously skipped components) and render that
- topic transiently unavailable → surface an error rather than silently degrading

https://github.com/user-attachments/assets/dba14d61-6e27-4f17-9d9f-b107b66a003e

---

> [!NOTE]
> **Medium Risk**
> Changes core paywall render and fallback behavior when workflows and remote config are enabled, including offerings reload on config disable; scope is focused and covered by new unit tests.
> 
> **Overview**
> With workflows enabled, offerings without a mapped workflow or with broken config could fail to render instead of showing a regular or default paywall. **Offering→workflow resolution now returns a `WorkflowResolution` sealed type** (`Found`, `NoWorkflow`, `Disabled`, `Unavailable`) instead of a nullable workflow id, so callers can tell workflowless offerings apart from config kill-switch and transient failures.
> 
> **`WorkflowsConfigProvider.resolveWorkflow`** checks `isDisabled` before the warm-cache fast path (so stale cache cannot skip an offerings reload after a 4xx disable) and maps a missing workflows topic to `Disabled` vs `Unavailable`.
> 
> **`PaywallViewModel`** branches on each outcome: present the workflow on `Found`; fall back to embedded/default paywall on `NoWorkflow`; reload `/offerings` on `Disabled` to recover skipped paywall components; show error on `Unavailable` or when a resolved workflow body fails to load (no silent degrade). Public/internal API surface switches from `workflowIdForOfferingId` to `resolveWorkflow` through `Purchases` and `PurchasesType`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit befb92fcefc713809b3a93bbec97a9349b58edb3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowResolution.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowResolution.kt
@@ -1,5 +1,3 @@
-@file:OptIn(InternalRevenueCatAPI::class)
-
 package com.revenuecat.purchases.common.workflows
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowResolution.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowResolution.kt
@@ -1,0 +1,28 @@
+@file:OptIn(InternalRevenueCatAPI::class)
+
+package com.revenuecat.purchases.common.workflows
+
+import com.revenuecat.purchases.InternalRevenueCatAPI
+
+/**
+ * Outcome of resolving an offering to its workflow through the `/v1/config` workflows topic. It lets the paywall
+ * render path tell apart three situations that a bare `workflowId?` conflates:
+ *
+ * - [Found]: the offering maps to a workflow id, which should be served through the workflows path.
+ * - [NoWorkflow]: the workflows topic was readable and the offering genuinely has no workflow. This is a
+ *   workflowless offering that should render its regular or default paywall.
+ * - [Unresolved]: the workflows topic could not be read at all — the `/v1/config` endpoint is disabled by a 4xx
+ *   kill switch, or a sync failed transiently — so whether the offering has a workflow is unknown. The caller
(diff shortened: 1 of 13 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowResolution.kt
```diff
@@ -0,0 +1,34 @@
+@file:OptIn(InternalRevenueCatAPI::class)
```
tonidero: Hmm do we need this?

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowResolution.kt
```diff
@@ -0,0 +1,34 @@
...
+ *   workflowless offering that should render its regular or default paywall.
+ * - [Disabled]: the topic could not be read because the `/v1/config` endpoint is disabled for the session by a
+ *   4xx kill switch. The offering was parsed with its paywall components skipped while workflows were enabled,
+ *   so the caller should reload offerings — which now re-parse with those components — to recover its paywall.
+ * - [Unavailable]: the topic could not be read for some other (transient) reason, so whether the offering has a
+ *   workflow is unknown and reloading would not recover anything. The caller should surface an error.
+ */
+@InternalRevenueCatAPI
+public sealed class WorkflowResolution {
+    @InternalRevenueCatAPI
```
tonidero: Also not sure if we need the annotation if the overall sealed class has the annotation... But well, nitpicking :P

tonidero (approved):
Looks great to me!! Thank you!!

PR #3791, opened 2026-07-16
refactor(remote-config): RCContainer improvements

bisho:
### Motivation

- Move checksum check to `decode()`, no alternative ways of getting the content, guarantees checksums are done.
- The config is converted to bytes right away, as it is always needed and non-optional. The config validation can also fail for trusted entitlement signatures, so if the hash doesn't match it should also behave in a similar way.
- When decoding elements, we allocate byte array, copying from the buffer if we must (when data is not compressed) to simplify usages that currently need to duplicate and handle this themselves.
- The design ensures all elements are only decoded once: "consumed": config is decoded right away, and getting blobs removes them from RCContainer, no double decoding by mistake.
- RCElement leaks a bit into the blob processing, we want to check if references are in cache before attempting to decode.

The `reserved` field is removed from `RCElement` since it was stored but never read by any caller. We can re-add it later if needed, no point in having it in code when unused.

Tests are updated throughout: mocks no longer stub `decode()` returning a `ByteBuffer` or `matchesChecksum`, and the `elements` map is no longer asserted. A new test covers the case where a content element's checksum doesn't match (parse succeeds, but `decode()` throws), and the config-checksum-mismatch case now asserts that `parse` itself throws.

---

> [!NOTE]
> **Medium Risk**
> Changes remote-config container parsing, response signature inputs, and inline blob extraction—security-adjacent paths—but behavior is covered by updated unit and integration tests; config checksum mismatches now fail earlier than before.
> 
> **Overview**
> Refactors how **RC Container Format v1** is consumed so integrity checks are harder to skip and large payloads stay compressed until needed.
> 
> **`RCContainer` / `RCElement`:** Element 0 is **decoded and checksum-verified during `parse`**, exposed as `config: ByteArray` (not an `RCElement`). Content blobs stay as zero-copy `contentElements`; the lazy **`elements` checksum map is removed**—callers use `checksumBase64()` on each element. **`RCElement.decode()`** now always returns a detached **`ByteArray`** and **throws on checksum mismatch** (replacing separate `isChecksumValid` / `matchesChecksum` / `decodedBytes` paths). **`RCContentEncoding.decode`** always materializes bytes; unused **`reserved`** on `RCElement` is dropped.
> 
> **Call sites:** `HTTPClient` signs/verifies over `RCContainer.parse(...).config` and **fails RC-format verification when config parse fails** (including bad config checksum). **`RemoteConfigManager`** parses config directly, extracts inline blobs by iterating `contentElements` (ref filter + skip if already cached) before decode. **`RemoteConfigBlobStore` / `RemoteConfigBlobFetcher`** take **`ByteArray`** instead of `ByteBuffer`.
> 
> Tests and integration fixtures are updated for the new API and stricter config checksum behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fefdce31a9a66678830840c86d0a0bd89a65827. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContentEncoding.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContentEncoding.kt
@@ -30,15 +30,14 @@ internal enum class RCContentEncoding(val id: Int) {
         fun fromId(id: Int): RCContentEncoding? = values().firstOrNull { it.id == id }
 
         /**
-         * Returns the uncompressed bytes of [source] for the given [codecId].
-         *
-         * [NONE] returns [source] unchanged (zero-copy). [GZIP] inflates into a fresh read-only buffer. Any
-         * other codec (including the known-but-unsupported [BROTLI]/[ZSTD]) or a corrupt gzip stream throws
+         * The uncompressed bytes of [source] for the given [codecId], as a fresh [ByteArray] detached from
+         * [source]'s backing buffer. [NONE] copies the bytes out; [GZIP] inflates into a new array. Any other
+         * codec (including the known-but-unsupported [BROTLI]/[ZSTD]) or a corrupt gzip stream throws
          * [RCContainerFormatException]; callers treat that as a verification/decode failure.
          */
-        fun decode(source: ByteBuffer, codecId: Int): ByteBuffer = when (fromId(codecId)) {
-            NONE -> source
-            GZIP -> ByteBuffer.wrap(gunzip(source)).asReadOnlyBuffer()
+        fun decode(source: ByteBuffer, codecId: Int): ByteArray = when (fromId(codecId)) {
+            NONE -> source.toByteArray()
+            GZIP -> gunzip(source)
             BROTLI, ZSTD, null ->
                 throw RCContainerFormatException("Unsupported content encoding id $codecId.")
         }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
@@ -7,89 +7,48 @@ import java.security.MessageDigest
 /**
  * A single element within an [RCContainer].
  *
- * Both [checksum] and [data] are zero-copy, read-only views over the container's backing buffer:
- * no element bytes are copied or hashed during parsing. [data] holds the **on-wire** bytes, which
- * may be compressed per [codec]; [decode] yields the uncompressed content. The [checksum] always
- * covers the uncompressed bytes, so callers that need integrity verification opt in via
- * [isChecksumValid] (which decodes first).
+ * [data] is a zero-copy, read-only view over the container's backing buffer — the potentially large payload
+ * is never copied during parsing. It holds the **on-wire** bytes, which may be compressed per [codec]; [decode]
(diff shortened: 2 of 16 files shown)


bisho:
* **#3791** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3791?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3791?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCContentEncoding.kt
```diff
@@ -30,15 +30,14 @@ internal enum class RCContentEncoding(val id: Int) {
...
+         * The uncompressed bytes of [source] for the given [codecId], as a fresh [ByteArray] detached from
+         * [source]'s backing buffer. [NONE] copies the bytes out; [GZIP] inflates into a new array. Any other
+         * codec (including the known-but-unsupported [BROTLI]/[ZSTD]) or a corrupt gzip stream throws
          * [RCContainerFormatException]; callers treat that as a verification/decode failure.
          */
-        fun decode(source: ByteBuffer, codecId: Int): ByteBuffer = when (fromId(codecId)) {
-            NONE -> source
-            GZIP -> ByteBuffer.wrap(gunzip(source)).asReadOnlyBuffer()
+        fun decode(source: ByteBuffer, codecId: Int): ByteArray = when (fromId(codecId)) {
+            NONE -> source.toByteArray()
```
bisho, line 39: This ensures the decode is a separate copy, doesn't hold refs to the buffer, etc, and simplifies how to think about these, and how freely the values can be used...  
  
For example:

`blobStore.write(ref, ByteBuffer.wrap(bytes))``  
`  
Becomes

`blobStore.write(ref, bytes)`

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/RCElement.kt
```diff
@@ -8,52 +8,36 @@ import java.security.MessageDigest
...
-     * [checksum]. The backend stores `sha256(uncompressed)` truncated to its leftmost [checksum] bytes
-     * (192 bits). An unsupported codec or corrupt stream makes [decode] throw, which is treated as a
-     * failed verification (returns `false`).
-     */
-    @Suppress("SwallowedException") // An undecodable element simply fails verification.
-    fun isChecksumValid(): Boolean = try {
-        matchesChecksum(decode())
-    } catch (e: RCContainerFormatException) {
-        false
+    fun decode(): ByteArray {
```
bisho, line 31: Single `decode()`, always verifies checksum, so it can't be missused, and have some codepath forgetting to do the validation.

tonidero (commented):
Love the changes. They do make a lot of sense and leave it cleaner IMO. cc @rickvdl since it might inspire similar changes in iOS. I don't think it's a requirement for V0, but just an improvement though so I wouldn't consider this a blocker for our initial testing.

Some tests seem to be failing (some should be fixed in main if you rebase)

tonidero (approved):
Looks great! Much cleaner indeed, thank you!

PR #3797, opened 2026-07-16
Add Android workflow paywall Maestro flows (local only)

vegaro:
Adds Android Maestro E2E flows for workflow (multipage) paywalls, mirroring the iOS flows added in purchases-ios #7221 and #7224. Both flows run against the existing `test-apps/e2etests` app.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the E2E test app, Maestro assets, and CI test path; production SDK behavior is unchanged and CI still uses the default API key path.
> 
> **Overview**
> Adds **local-only** Maestro coverage for multipage workflow paywalls under `maestro/workflows/`, aligned with the iOS workflow E2E flows. CI Maestro now runs only `maestro/e2e_tests/` so these workflow flows are not executed on CircleCI.
> 
> The `e2etests` app gains an `open_workflow` launch path: Maestro passes `e2e_test_flow` as an intent extra, `MainActivity` shows a new `WorkflowScreen` that presents the `default_workflows` offering via RevenueCatUI, and entitlement status is surfaced for assertions. When `E2E_WORKFLOWS_API_KEY` is set at build time (`local.properties` / Gradle), `E2ETestsApplication` configures that project with `DangerousSettings.forWorkflows()`; without it, the existing CI `test_store_annual_purchase` setup is unchanged.
> 
> `local.properties.example` documents the optional workflows API key for running the flows locally against the test store.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bd741c3c203de60b07753818e5f6ce24f36b664. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


facumenzella (approved):
Looks solid. Haven't run the tests locally, but we're already running this in the CI on iOS, so let's plug that in here too 🚀

maestro/workflows/navigate_workflow_back.yaml
```diff
@@ -0,0 +1,43 @@
+# Opens a workflow paywall (offering `default_workflows`), advances to screen 2, taps the back
```
facumenzella, line 1: on my mind: we might be able to share this one day

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/E2ETestsApplication.kt
```diff
@@ -1,22 +1,35 @@
...
-                context = this,
-                apiKey = Constants.API_KEY,
-            ).build(),
-        )
+        // The workflow E2E flows are built with E2E_WORKFLOWS_API_KEY set (surfaced as
+        // BuildConfig.WORKFLOWS_API_KEY). When it's present we point at that project and enable
+        // workflows; otherwise we keep the default configuration used by the CI
+        // test_store_annual_purchase flow untouched. The Maestro launch argument only selects the
+        // screen (in MainActivity), so configuration stays here in the Application.
+        val builder = if (BuildConfig.WORKFLOWS_API_KEY != WORKFLOWS_API_KEY_PLACEHOLDER) {
```
tonidero, line 22: Hmm I do wonder if we should try to reuse the same E2E project TBH... Don't think there is a necessary reason to split these in the long run and splitting to a different project means we will need to migrate it later... But well, won't block the PR for this :)

tonidero (approved):
So nice to get these in... I guess we're not running them on CI yet though, but well, we can do that in a separate PR.

I would argue whether it might be ok to unify into the existing job instead of a separate one. I think it will add time without need since the current maestro job is not too slow

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/E2ETestsApplication.kt
vegaro: that's a good point, let's think about it @facumenzella 

but yeah, let's merge this, we can iterate on it

PR #3798, opened 2026-07-16
feat(paywalls): web_view JavaScript bridge (6b)

JZDesign:
### Checklist
- [x] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids

### Motivation
Part **6b** of the web_view split (replaces the other half of #3785). **Inert until Part 8** (#3787). Keeps the per-document `CoroutineScope` threading model from #3785.

**Based on `main`.** 6a (#3796) folded into this branch (empty net diff).

### Description
JavaScript bridge for the `rc-web-components` channel:

- `WebViewJavaScriptBridge` (origin helpers from Part 5)
- `androidx.webkit` dependency (`webkit = 1.12.1`)

**App-facing message handler removed** for v1: the bridge keeps only the SDK-internal path (handshake, resize/fit, origin checks). `postMessage` / `postVariables` / `onMessage` and the `PaywallWebViewController` / `PaywallWebViewMessageHandler` types are gone.

Bridge production file: **376 lines**. No public API change.

**Verified locally:**
- `./gradlew :ui:revenuecatui:testDefaultsDebugUnitTest --tests "com.revenuecat.purchases.ui.revenuecatui.components.webview.*"` — green (33 bridge tests + envelope/nav)
- `./gradlew detektAll` — green
- `./scripts/api-check.sh` — no-op / green

**Coordination:**
- Part 7 (#3786 / `webview/7-view`) updated for the new bridge signature (`TestWebViewSlot` + webkit static mocks via RuleChain)
- Part 8 (#3787 / `webview/8-activate`) rebased onto updated 7
- #3785 should be closed in favor of this PR (branch kept as audit reference)

**Danger:** still over the 300-production-line limit (single cohesive security-sensitive class + suite). Apply `skip-pr-lines-changed-check` with that justification (label already applied).

### Series (splitting #3757)

PWENG-98
Parts 1 and 2 were dropped. Not a linear stack: parts 3, 4, 5 are independent
(each off `main`); `webview/deps-1-5` merges all three; the tail stacks linearly on
that merge (6b -> 7 -> 8; 6a folded into 6b, empty net diff).

| Part | PR | Base | Notes |
|------|-----|------|-------|
| 1 | ~~#3780~~ | - | dropped; value types collapsed to `Map<String, String>` |
| 2 | ~~#3784~~ | - | dropped; app-facing message handler cut from v1 |
| 3 | #3781 | `main` | schema component, not yet registered |
[...]

PR #3782, opened 2026-07-15
feat(paywalls): web_view transport envelope + JSON hardening (4)

JZDesign:
### Checklist
- [x] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids

### Motivation
Part **4** splitting #3757; **inert until Part 8** (#3787).

Lands the wire envelope + JSON input hardening used by the web_view message bridge. No runtime wiring yet.

### Description
- `WebViewEnvelope` / `WebViewMessageType` / `WebViewMessageField`
- Envelope unit tests: per-field parse/reject, kind whitelist, `build(...)` emission

Verified: `WebViewEnvelopeTest`, `detektAll`.

**Labels:** `pr:feat`, `pr:RevenueCatUI`, `feat:Paywalls_V2`

### Series (splitting #3757)

PWENG-98

Parts 1 and 2 were dropped. Not a linear stack: parts 3, 4, 5 are independent
(each off `main`); `webview/deps-1-5` merges all three; the tail stacks linearly on
that merge (6a -> 6b -> 7 -> 8).

| Part | PR | Base | Notes |
|------|-----|------|-------|
| 1 | ~~#3780~~ | - | dropped; value types collapsed to `Map<String, String>` |
| 2 | ~~#3784~~ | - | dropped; app-facing message handler cut from v1 |
| 3 | #3781 | `main` | schema component, not yet registered |
| 4 | #3782 | `main` | transport envelope + JSON hardening |
| 5 | #3783 | `main` | navigation policy + origin helpers |
| 6a | #3796 | `webview/deps-1-5` | SDK variables provider |
| 6b | #3798 | #3796 | JavaScript bridge |
| 7 | #3786 | #3798 | Compose view + style |
| 8 | #3787 | #3786 | activation + unsupported protocol_version fallback |

[...]

PR #3813, opened 2026-07-21
refactor(paywalls): use kotlinx.serialization for web_view transport envelope (9)

AlvaroBrey:
### Motivation

The web_view transport envelope (#3782, #3798) does serialization by hand with org.json. Replace it with kotlinx.serialization, which the module already uses.

### Description

Stacked on #3787.

- `WebViewEnvelope` is now a `@Serializable` data class with a `Kind` enum; validation happens at decode time. The `WebViewMessageField` constants are replaced by `@SerialName` annotations.
- The bridge builds and reads payloads as kotlinx `JsonObject` instead of `JSONObject`.
- `WebViewEnvelopeTest` no longer needs Robolectric (kotlinx is pure JVM).

Behavior changes vs the hand-rolled parser, all affecting only out-of-contract frames (the content side always emits `JSON.stringify` output):

- Non-integer `protocol_version` is rejected instead of truncated. Protocol versions are exact-match integers per workflow-web-components-sdk.
- Explicit `null` in optional fields parses as absent instead of rejecting the frame.
- Non-RFC JSON (lenient syntax, duplicate keys) is rejected.

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/webview/WebViewEnvelope.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/webview/WebViewEnvelope.kt
@@ -2,11 +2,15 @@
 
 package com.revenuecat.purchases.ui.revenuecatui.components.webview
 
-import org.json.JSONException
-import org.json.JSONObject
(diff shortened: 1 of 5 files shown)


AlvaroBrey:
@cursor review

AlvaroBrey:
☝️  Not a real merge, just reordering PRs

PR #3817, opened 2026-07-22
Fetch offline entitlement mappings through Remote Config

rickvdl:
Android equivalent of iOS PR https://github.com/RevenueCat/purchases-ios/pull/7265

## Description
Implements fetching the `ProductEntitlementMapping` through the remote config infrastructure. This will only be done when remote config is enabled. And will still fall back to the existing implementation when remote config is disabled through the kill-switch.

There's still an open question around handling the fallback if the kill-switch is activated later [here](https://github.com/RevenueCat/purchases-android/pull/3817#discussion_r3631886849)

---

> [!NOTE]
> **Medium Risk**
> Changes how offline entitlement mappings are sourced for apps using remote config, with legacy fallback but new async/concurrency behavior on the refresh path.
> 
> **Overview**
> When **workflows / remote config** is enabled, stale **product entitlement mapping** updates now read the `product_entitlement_mapping.default` blob through remote config instead of always calling the legacy backend endpoint.
> 
> `ProductEntitlementMappingTopicProvider` decodes that blob; `OfflineEntitlementsManager` runs the refresh on a coroutine, **dedupes** concurrent updates, **falls back** to `backend.getProductEntitlementMapping` when remote config is off or the blob is missing/unparseable, and **`close()`** cancels in-flight work (wired from `PurchasesOrchestrator.close()`). `PurchasesFactory` only creates the topic provider when remote config is enabled.
> 
> `RemoteConfigTopic` adds **`ProductEntitlementMapping`**, and `RemoteConfigManager` tightens the 4xx disable path under `cacheLock` plus small blob-resolution refactors. Tests cover the new path, fallback, concurrency, and production integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 52512be3cacb41f29adebed11ab70bf5139392f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -644,9 +652,59 @@ internal class RemoteConfigManager(
         itemKey: String,
         transform: (ByteArray) -> T?,
     ): T? = withContext(ioDispatcher) {
-        resolveBlobBytes(topic, itemKey)?.let(transform)
+        blobDataSnapshot(topic, itemKey, transform)?.value
     }
 
+    /** Resolves a blob and retains the committed-config identity that produced it. */
+    suspend fun <T> blobDataSnapshot(
```
rickvdl: Equivalent of the iOS method [`blobDataSnapshot`](https://github.com/RevenueCat/purchases-ios/blob/6bb91ce62166d2475f0ca6b5390dfbc7cdc660d0/Sources/Networking/RemoteConfigManager.swift#L423-L451), which is used for preventing the theoretical case of a topic / blob being invalidated before it's returned to the caller. 

This ensures that the PEM in this case is only persisted to disk if it's the current, up to date value. Which should prevent potential race conditions (mostly when switching users).

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/OfflineEntitlementsManager.kt
```diff
@@ -101,23 +109,51 @@ internal class OfflineEntitlementsManager(
     fun updateProductEntitlementMappingCacheIfStale(completion: ((PurchasesError?) -> Unit)? = null) {
         if (isOfflineEntitlementsEnabled() && deviceCache.isProductEntitlementMappingCacheStale()) {
```
rickvdl: This method will fetch the PEM from the legacy implementation only when the remote config one fails / is not available. However it could happen that it is fetched successfully, but remote config is disabled in a later stage. 

If we want to tackle this as well I guess we'd have to always fetch it from the legacy system still?

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMappingTopicProvider.kt
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    private companion object {
+        private const val ITEM_KEY = "default"
+    }
+}
+
+@OptIn(InternalRevenueCatAPI::class)
+internal interface EntitlementMappingTopicProvider {
```
tonidero: Hmm in Android we don't really use these interfaces normally unless we need to... I guess it could be good practices, but not something we do right now, so I would probably remove the interface for now?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
```diff
@@ -435,6 +441,8 @@ internal class RemoteConfigManager(
     fun close() {
         scope.cancel()
         synchronized(cacheLock) {
+            isClosed = true
```
tonidero: Just a note that, after an instance of `Purchases` is "closed" I would assume that all the attached classes would be freed, so normally, I would assume nothing more should be done after this calls close. In general vast majority of folks will never use this, and I consider it mostly useful for tests. So just wondering if we need this... NABD to have it though, since it could potentially avoid some disk writes if the developer keeps the reference in memory for any reason...

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMappingTopicProvider.kt
```diff
@@ -0,0 +1,45 @@
...
+    suspend fun getProductEntitlementMapping(): ProductEntitlementMappingResult?
+}
+
+@OptIn(InternalRevenueCatAPI::class)
+internal class ProductEntitlementMappingResult(
+    val mapping: ProductEntitlementMapping,
+    private val useIfCurrent: ((ProductEntitlementMapping) -> Unit) -> Boolean,
```
tonidero: Hmm TBH, I'm wondering if we need to be so careful specially for PEM... Note that PEM does not change per user + we would trigger a new fetch immediately after changing the user that would override it, so the window of error is quite narrow. I guess this makes it a bit safer, but just wondering if it's worth the extra complexity TBH...

tonidero (commented):
Still reviewing but looking great! Just left some initial comments

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMappingTopicProvider.kt
rickvdl: Ah good one! Indeed they're not needed for mocking on Android, so I've removed it :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
rickvdl: Right, that's indeed a reasonable assumption. So I'll just remove it for simplicity

purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMappingTopicProvider.kt
rickvdl: Right, that's fair. I've removed that complexity for now 👍

PR #3826, opened 2026-07-24
Fix crash when a paywall component tree fails to decode

AlvaroBrey:
### Motivation

`Offering.PaywallComponents.data` decodes the component tree lazily, and Kotlin's `lazy` does not cache a *thrown* initializer, so it re-throws (and re-decodes) on every access after a failure. Since the lazy-parse change (#3589, already released), a decode failure yields a non-null holder that throws on access instead of a null `paywallComponents`. 

Unguarded readers (notably the fallback paywall render, impression fingerprint) therefore crash on the main thread for a components-only offering whose tree can't decode, e.g. an unknown component type with no `fallback` reaching an SDK that doesn't know it.

### Description

- Memoize the decode outcome in `PaywallComponents` (`Lazy<Result<…>>`): the tree decodes once, and success *or* failure is cached.
- Add a non-throwing `dataOrNull`; keep `data` throwing for the one intentional catcher (`validatePaywallComponentsDataOrNull`), which still degrades to the default paywall.
- Route best-effort readers (fallback render, fingerprint, event data, exit offers, tester list) through `dataOrNull`.
- Test: decode a real tree with a synthetic unknown type and no fallback; assert `data` throws, `dataOrNull` is null, and the tree decodes only once.

---

> [!NOTE]
> **Medium Risk**
> Public API change (`data` is now `Result` instead of throwing `PaywallComponentsData`) and broad paywall UI/SDK touchpoints; behavior is defensive but integrators must adopt the new accessors.
> 
> **Overview**
> **Fixes main-thread crashes** when a V2 offering has `paywallComponents` present but the component tree fails to decode (e.g. unknown component type with no fallback). Lazy decode no longer re-throws on every access; success and failure are cached once.
> 
> `Offering.PaywallComponents` now backs lazy decode with `Lazy<Result<PaywallComponentsData>>`: **`data`** is a **`Result`** (callers use `getOrThrow()` / `getOrElse`), and **`dataOrNull`** is added for best-effort reads that treat failure as absence.
> 
> Call sites are updated so optional paths (image/video pre-download, paywall tester listing, impression fingerprints, exit offers, analytics) use **`dataOrNull`** or **`data.getOrElse`**, while validation still handles failure and falls back to the default paywall. Tests cover memoized failure and decode-fallback behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 504eedcc2b448fa2a99472971f44c20d4366ecf3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
@@ -83,9 +83,11 @@ constructor(
             dataProvider: () -> PaywallComponentsData,
         ) : this(uiConfig, componentsHash, lazy(LazyThreadSafetyMode.SYNCHRONIZED) { runCatching(dataProvider) })
 
-        // Decoded once on first access; the outcome (success or failure) is memoized. [data] re-throws the
-        // cached failure; [dataOrNull] returns null instead, for best-effort reads that must not crash.
-        public val data: PaywallComponentsData get() = dataResult.value.getOrThrow()
+        // Decoded once on first access; the outcome (success or failure) is memoized. [data] exposes that
+        // outcome as a [Result] so callers must handle a decode failure explicitly (there is deliberately no
+        // throwing accessor to reach for by mistake). [dataOrNull] is the convenience for best-effort reads
+        // that treat a failure as absence.
+        public val data: Result<PaywallComponentsData> get() = dataResult.value
 
         public val dataOrNull: PaywallComponentsData? get() = dataResult.value.getOrNull()
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
@@ -66,14 +66,14 @@ constructor(
         // Offering.equals/hashCode/toString — which are @Poko-generated over `paywallComponents` — from forcing
         // a decode of every offering's component tree (e.g. when the cache compares cached vs network offerings).
         private val componentsHash: String,
-        dataProvider: Lazy<PaywallComponentsData>,
+        private val dataResult: Lazy<Result<PaywallComponentsData>>,
     ) {
         /**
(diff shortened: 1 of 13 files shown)


AlvaroBrey:
@cursor review

AlvaroBrey:
@cursor review

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
```diff
@@ -66,24 +66,28 @@ constructor(
...
             uiConfig: UiConfig,
             componentsHash: String,
             dataProvider: () -> PaywallComponentsData,
-        ) : this(uiConfig, componentsHash, lazy(LazyThreadSafetyMode.SYNCHRONIZED, dataProvider))
+        ) : this(uiConfig, componentsHash, lazy(LazyThreadSafetyMode.SYNCHRONIZED) { runCatching(dataProvider) })
 
-        public val data: PaywallComponentsData by dataProvider
+        // Decoded once on first access; the outcome (success or failure) is memoized. [data] re-throws the
+        // cached failure; [dataOrNull] returns null instead, for best-effort reads that must not crash.
+        public val data: PaywallComponentsData get() = dataResult.value.getOrThrow()
```
tonidero: Hmm I wonder if we should remove/hide this so we don't forget about this in the future... But I think it should be ok...

tonidero (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/Offering.kt
AlvaroBrey: https://github.com/RevenueCat/purchases-android/pull/3826/changes/504eedcc2b448fa2a99472971f44c20d4366ecf3

Refactored to expose the `Result` instead so if anyone wants the exception  (eg `OfferingToStateMapper` and tests) they can get it but it will be obvious

tonidero (approved):
:shipit:

PR #3812, opened 2026-07-21
refactor(remote-config): add API sources failover with health check endpoint support

tonidero:
### Description

Completes the health-check-gated API source failover feature. Everything remains behind the `usesRemoteConfigAPISources` dangerous setting, so released behavior is unchanged.

**Failure classification per request attempt on the current source:**

| Outcome | Action |
|---|---|
| 2xx/3xx | Success. |
| 4xx | Return result; no failover, no health check (request error, not a source problem). |
| 5xx or connection failure | `GET <source>/v1/health/connectivity` (lazy, never proactive). Health 2xx → source healthy, no failover, surface the original error/result. Health non-2xx or unreachable → mark source unhealthy, transparently retry the in-flight request on the next source. |

Follow-up: Add client side connectivity checks before hitting health checks.

---

> [!NOTE]
> **Medium Risk**
> Changes core HTTP request routing and retry behavior for API traffic when the dangerous setting is on; released apps stay unchanged, but misclassified health/failover could affect which host serves purchases and config.
> 
> **Overview**
> Adds **health-check-gated failover** across remote-config API sources when `usesRemoteConfigAPISources` is enabled. `HTTPClient` no longer only swaps hosts via `RemoteConfigSourceProvider`; it wires **`APISourceFailover`** and walks sources in a loop (capped at **`MAX_API_SOURCE_ATTEMPTS`**) before existing fallback-URL logic runs.
> 
> On **5xx or connection failure**, the SDK probes **`/v1/health/connectivity`** via **`SourceHealthChecker`** (cached ~10s, concurrent checks coalesced). A **2xx health** response means the source is fine—**no host switch** and the original error/result is returned. Unhealthy sources are reported to the provider and the **same in-flight request** retries on the next source. **4xx** never triggers health checks or failover.
> 
> **`DefaultRemoteConfigSourceProvider`** now **re-arms the API list every 10 minutes** after the first failover advance (blob lists unchanged), and **`reportUnhealthy`** returns whether the list actually advanced.
> 
> Factory wiring: `PurchasesFactory` builds `APISourceFailover` with `SourceHealthChecker` and passes it into `HTTPClient` instead of the raw provider.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7f48b94b7a746234d0315680c6d82415b4cd935. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3840, opened 2026-07-27
feat(remote-config): device connectivity pre-check for API source failover

tonidero:
### Why

When a request to an API source fails while the device itself has no connectivity (airplane mode, dead wifi), the health check is guaranteed to fail too — so the failover would walk through and burn every source for a purely client-side problem.

### What

A new `DeviceConnectivityChecker` is consulted before running the health check, and only when the request got **no response at all**: it asks the OS whether there is an active network (`ConnectivityManager.getActiveNetwork()`, a cheap local call, no network I/O). If the device is positively offline, the original error surfaces without health-checking or consuming a source. Any uncertainty (missing permission, `ConnectivityManager` unavailable or throwing) fails open to the existing behavior.

A 5xx response is deliberately *not* gated on connectivity: the source demonstrably answered, so the device is online and the check could only do harm on a false offline verdict.

### Note for reviewers

This adds `ACCESS_NETWORK_STATE` (a normal install-time permission) to the SDK manifest, which merges into host apps regardless of the `usesRemoteConfigAPISources` dangerous setting. Apps can opt out with `tools:node="remove"`.

### Review round (`fad10f3b2`)

Both of @AlvaroBrey's points are addressed:

- **`NET_CAPABILITY_VALIDATED` dropped.** Validation comes from the OS's own captive-portal probe against Google-owned hosts, so a VPN or a network filtering those hosts reports "not validated" while our API hosts are perfectly reachable — a false offline verdict whose cost is suppressing a legitimate failover. `isDeviceOffline()` is now just `activeNetwork == null`, and the reason not to reintroduce validation lives in the class KDoc. **Captive portals are no longer covered**, by design.
- **Connectivity is only consulted when no response arrived.** `onRequestFailure` takes the attempt's `IOException?` (null meaning "the source responded"), and the offline branch is gated on it being non-null.

On granularity: the check fires for any no-response `IOException` rather than an allowlist of subtypes (`ConnectException`/`UnknownHostException`/`SocketTimeoutException`). The gate is a conjunction with "the OS reports no active network", and when that holds no source is reachable whatever the subtype, so an allowlist would only change unlisted-subtype cases — where its failure mode is burning every source again, i.e. the bug this fixes. Sibling `SocketException` subtypes such as `NoRouteToHostException` fall outside `ConnectionErrorReason`'s `NO_NETWORK` bucket, and Android's mapping varies by HTTP stack and OEM.

Test churn worth a look: the 5xx-while-offline tests in `HTTPClientTest` and `BackendAPISourceFailoverIntegrationTest` are **inverted** (they now assert failover still happens), and connection-failure-while-offline coverage was added in both, which was missing before.

---

> [!NOTE]
> **Medium Risk**
> Changes failover behavior for connection failures under remote-config API sources and adds a merged manifest permission; scope is limited and connectivity checks fail open when uncertain.
> 
> **Overview**
> Adds a **device connectivity pre-check** so remote-config **API source failover** does not health-check or burn alternate hosts when the request never got a response and the OS reports no active network.
> 
> **`DeviceConnectivityChecker`** uses `ConnectivityManager.activeNetwork == null` (no `NET_CAPABILITY_VALIDATED`, to avoid false offline on VPN/filtered captive-portal probes). Uncertainty (missing service, permission issues, exceptions) **fails open** to existing failover behavior. **`APISourceFailover.onRequestFailure`** now takes the attempt’s `IOException?`; the offline short-circuit runs only when that exception is non-null. **5xx responses** still run health checks and can fail over, since the source clearly responded.
> 
> **`HTTPClient`** forwards `connectionException` into failover decisions; the separate **fallback-host** path on connection errors is **unchanged** (comment notes connectivity is not consulted there yet). **`PurchasesFactory`** wires a real checker; the SDK manifest adds **`ACCESS_NETWORK_STATE`** (mergeable; apps can remove with `tools:node="remove"`).
> 
> Tests cover unit, HTTP client, and backend integration paths for offline connection failures vs offline 5xx.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b65981ab5cdff9b82678476908dc27f1ac4303e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3840** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3812** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3812?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
tonidero, line 4: This is my main question for this PR... I do think this is a pretty common permission, but still would love Android folks to look at it 🙏 cc @vegaro @AlvaroBrey

vegaro: I think it's fine, it's a normal permission that doesn't trigger dialog, so I think we can add it

AlvaroBrey: Agreed, no big deal if it's autogranted

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/DeviceConnectivityChecker.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * True only when the OS positively reports no validated network. Any uncertainty (missing
+     * ACCESS_NETWORK_STATE permission, ConnectivityManager unavailable, framework exception) is
+     * false, so callers fail open to their normal behavior.
+     */
+    @Suppress("TooGenericExceptionCaught", "ReturnCount")
+    fun isDeviceOffline(): Boolean {
+        return try {
+            val connectivityManager = applicationContext.getSystemService(Context.CONNECTIVITY_SERVICE)
```
AlvaroBrey, line 27: Hmm using `NET_CAPABILITY_VALIDATED` may not be the best for reliability. IIRC this usually just checks connectivity to `gstatic.com`, so we get a false negative (and thus block failovers) wherever that domain is not available (VPNs, networks which filter Google domains, etc), even if connectivity to our domains is unaffected.
Maybe drop it and just take `activeNetwork == null`?

AlvaroBrey: Thinking about this some more, perhaps we can refine it? After all when we reach `AttemptOutcome.Failed` is always after a network call, so we know the reason of the failure. Does it make sense to skip connectivity check if the error could only happen with an actual backend response (`4xx`, `5xx`, but not `SocketTimeoutException`,`UnknownHostException`,`ConnectException`)?

tonidero: > Maybe drop it and just take activeNetwork == null?

Hmm you're right... We shouldn't use that for this optimization, I agree 👍 

> Does it make sense to skip connectivity check if the error could only happen with an actual backend response (4xx, 5xx, but not SocketTimeoutException,UnknownHostException,ConnectException)?

Yeah I think that can be a bit more optimization we can do... We do want to do the health check on a 5xx, but I do think we don't need to do this connectivity check.

Thank you! Will do those changes!

tonidero: I've done those 2 changes in https://github.com/RevenueCat/purchases-android/pull/3840/commits/fad10f3b2e5a32b60db01c89665f270c86b4f8d5 🙏

AlvaroBrey (approved):

PR #3844, opened 2026-07-28
Assume product with no billing period to be a "lifetime" offering

RosieWatson:
### Motivation

Fixed `price_per_period_abbreviated` (and similar period variables) returning an empty string for lifetime products with custom package identifiers

### Description

This updates so we are now detecting "lifetime" via the product having no billing period (`period == null`) instead of the `$rc_lifetime` package identifier

Comparative check in purchases-ios that this change was based on: https://github.com/RevenueCat/purchases-ios/blob/48237e34209b86221aa86629befc3698df2a2640/RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift#L594

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 5f40e1e2b43a6f5ace8d75a2b4a5f1a3d9732586. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackagePeriodExtensions.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackagePeriodExtensions.kt
@@ -7,7 +7,7 @@ import com.revenuecat.purchases.paywalls.components.common.VariableLocalizationK
 import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
 
 internal val Package.isLifetime: Boolean
-    get() = packageType == PackageType.LIFETIME
+    get() = packageType == PackageType.LIFETIME || product.period == null
 
 internal val Package.periodUnitLocalizationKey: VariableLocalizationKey?
     get() = if (isLifetime) VariableLocalizationKey.LIFETIME else product.period?.periodUnitLocalizationKey
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
@@ -133,20 +133,19 @@ internal class VariableDataProvider(
     }
 
     fun periodName(rcPackage: Package): String? {
-        if (rcPackage.packageType == PackageType.CUSTOM ||
-            rcPackage.packageType == PackageType.UNKNOWN
-        ) {
-            return rcPackage.identifier
-        }
-        val stringId = when (rcPackage.packageType) {
-            PackageType.LIFETIME -> R.string.lifetime
-            PackageType.ANNUAL -> R.string.annual
-            PackageType.SIX_MONTH -> R.string.semester
-            PackageType.THREE_MONTH -> R.string.quarter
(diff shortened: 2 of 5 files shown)


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackagePeriodExtensions.kt
```diff
@@ -7,7 +7,7 @@ import com.revenuecat.purchases.paywalls.components.common.VariableLocalizationK
 import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
 
 internal val Package.isLifetime: Boolean
-    get() = packageType == PackageType.LIFETIME
+    get() = packageType == PackageType.LIFETIME || product.period == null
```
vegaro, line 10: I think this is going to break for consumables. I believe that's what making the tests fail actually

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/VariableDataProvider.kt
```diff
@@ -133,20 +133,19 @@ internal class VariableDataProvider(
...
-            PackageType.LIFETIME -> R.string.lifetime
-            PackageType.ANNUAL -> R.string.annual
-            PackageType.SIX_MONTH -> R.string.semester
-            PackageType.THREE_MONTH -> R.string.quarter
-            PackageType.TWO_MONTH -> R.string.bimonthly
-            PackageType.MONTHLY -> R.string.monthly
-            PackageType.WEEKLY -> R.string.weekly
-            PackageType.UNKNOWN, PackageType.CUSTOM -> null
+        val stringId = when {
+            rcPackage.isLifetime -> R.string.lifetime
```
vegaro, line 137: putting this above `PackageType.CUSTOM` will show "Lifetime" for custom packages

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/processed/PackagePeriodExtensions.kt
RosieWatson: Let me dig more into how iOS is currently handling the split more 🤔

RosieWatson: Okay, so (correct me if I'm wrong), iOS has more ways of telling the difference between a consumable and non-consumable product that we are able to do via the play store... so this does break things (as you said) and undoing it will leave it broken unless using `rc_lifetime` which doesn't really work when you have multiple of them in one offering...

I think this would require a bigger change by adding a new marker to the backend to denote products as `lifetime` offerings

Does that line up with your understanding @vegaro?

vegaro: that's exactly right

vegaro:
@RosieWatson gave this a try and opened https://github.com/RevenueCat/purchases-android/pull/3850, check it out!

RosieWatson:
Closed - replaced by https://github.com/RevenueCat/purchases-android/pull/3850

PR #3847, opened 2026-07-28
fix: exclude pending purchases from offline entitlements calculation

tonidero:
### Description

When computing CustomerInfo offline, purchases in `PENDING` state returned by `queryPurchases` were treated like any other purchase. This aligns the offline calculation with the post-receipt path, which already skips pending purchases. `PurchasedProductsFetcher` now filters out `PENDING` transactions before mapping them to purchased products. Transactions with `UNSPECIFIED_STATE` are still included, since Amazon purchases always have that state.

---

> [!NOTE]
> **Medium Risk**
> Changes offline entitlement computation, which can affect what users see when the network is unavailable; scope is narrow (pending-only filter) and aligned with existing receipt handling.
> 
> **Overview**
> Offline **CustomerInfo** could grant entitlements from store purchases still in **`PENDING`** state because `PurchasedProductsFetcher` treated every `queryPurchases` result equally. The fetcher now **drops `PENDING` transactions** before building purchased products, matching the post-receipt path (`PostReceiptHelper` already refuses pending receipts).
> 
> **`UNSPECIFIED_STATE`** purchases are still included so Amazon transactions (which always use that state) are not affected. Tests cover pending exclusion and unspecified-state inclusion; test stubs accept a configurable `purchaseState`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e173c9eb1cc12112f466dc651b93fc2b5b6f395b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/PurchasedProductsFetcher.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/PurchasedProductsFetcher.kt
@@ -8,6 +8,7 @@ import com.revenuecat.purchases.common.BillingAbstract
 import com.revenuecat.purchases.common.DateProvider
 import com.revenuecat.purchases.common.DefaultDateProvider
 import com.revenuecat.purchases.common.caching.DeviceCache
+import com.revenuecat.purchases.models.PurchaseState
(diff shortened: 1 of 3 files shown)


rickvdl (approved):
Thanks for the fix!

PR #3849, opened 2026-07-28
Fix: Fill expanding to Custom Component size after layout measure pass

JZDesign:
### Checklist


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/stack/StackComponentView.kt
JZDesign, line 1: This is actually an insufficient change. It solves the problem of the PR where if a fill stack is overflowed by a webview, it allows you to scroll. But what it doesn't do is force a webview to fill the space any more…

And that's where most of the complexity is. Fill when smaller than the parent space. Fit when larger…

PR #3853, opened 2026-07-29
feat(remote-config): send X-RC-Last-Refresh-Time on config requests

tonidero:
## Description
In order to be able to leverage the load shedder functionality, the backend must be able to determine when the request was last refreshed. Other API requests that rely on ETAG caching send the `X-RC-Last-Refresh-Time` header as a unix epoch in milliseconds. With this PR we're doing the same for remote config requests.

Android equivalent of https://github.com/RevenueCat/purchases-ios/pull/7308

## Changes
- After receiving the initial remote config response, we'll now store a `lastRefreshTime` value along with the received config
  - This is only done for successful responses, determined by the `2xx` status code
- In subsequent requests we'll be including this value as a Unix epoch timestamp in milliseconds under the `X-RC-Last-Refresh-Time` header
- After the next successful response we'll update the local value again

---

> [!NOTE]
> **Medium Risk**
> Changes remote config request headers and disk persistence for sync bookkeeping, but behavior mirrors the existing ETag last-refresh pattern and old cache files degrade safely to a missing header.
> 
> **Overview**
> Remote config `/v1/config` requests now send **`X-RC-Last-Refresh-Time`** (epoch millis), reusing the same header name as ETag-cached APIs (`HTTPRequest.LAST_REFRESH_TIME_HEADER_NAME`, renamed from the ETag-only constant). The header is omitted until a prior successful refresh exists; **`getRemoteConfigFallback` does not send it**.
> 
> **Persistence and replay:** `PersistedRemoteConfigurationState` adds optional `lastRefreshTime`. On a **200**, `RemoteConfigManager` stores it with the config commit using the same clock instant as the in-memory staleness gate. On a **204**, only that timestamp is advanced (config unchanged); the write runs on the manager scope and re-reads disk so a wiped cache is not resurrected. `issueConfigRequest` passes the persisted value into `Backend.getRemoteConfig`.
> 
> **Tests** cover header presence/absence, 200/204 persistence, identity `clearCache`, and backward-compatible disk reads without the new field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ed8f58ac04b852c7e40b717905daaeb57a237403. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -19,6 +19,7 @@ import com.revenuecat.purchases.backendName
 import com.revenuecat.purchases.common.caching.WorkflowMetadata
 import com.revenuecat.purchases.common.events.EventsRequest
 import com.revenuecat.purchases.common.networking.Endpoint
+import com.revenuecat.purchases.common.networking.HTTPRequest
 import com.revenuecat.purchases.common.networking.HTTPResult
 import com.revenuecat.purchases.common.networking.PostReceiptResponse
 import com.revenuecat.purchases.common.networking.RCContainer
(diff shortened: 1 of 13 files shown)


rickvdl (approved):
Nice! 🚢

PR #3854, opened 2026-07-29
test(remote-config): de-flake the inline workflows blob integration test

tonidero:
`ProductionRemoteConfigIntegrationTest.decodes an inline workflows content blob` intermittently failed on CI with `Expected a workflows item with an inlined blob_ref`.

The request is identical every run and this path has no cache, retry or URL fallback, so the variance is in the response: blob inlining is a backend optimization, not a contract, and a blob the config advertises may be served from the CDN instead.

The test now:
- asserts unconditionally that every inlined element's ref is advertised by the config (`prefetch_blobs` ∪ topic `blob_ref`s) and decodes to non-empty bytes, which still exercises the real backend bytes through codec + checksum verification;
- skips (with diagnostics naming the advertised refs and the inlined element refs/sizes) instead of failing when the backend inlined no workflows blob.

Note this is a valid case on the backend since in some cases the blob would need to be warmed before being inlined, so we need to handle that case as well

---

> [!NOTE]
> **Low Risk**
> Test-only change with no production code; reduces false CI failures without weakening coverage of inlined blob decoding.
> 
> **Overview**
> Stops CI flakes in **`decodes an inline workflows content blob`** by aligning the test with backend behavior: blob inlining is optional, so responses may advertise workflows refs without inline bytes.
> 
> The test now **always** checks every inlined element is advertised in config (`prefetch_blobs` plus topic `blob_ref`s) and that `decode()` yields non-empty bytes (codec + checksum path against real backend data). The workflows JSON assertion runs only when a workflows blob was inlined; otherwise **`Assume.assumeTrue`** skips with a **`describe()`** diagnostic (active topics, prefetch refs, workflows items, inlined element refs/sizes).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1fad6197815d14e2c5e4b84f0745f0cd7a1efa7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionRemoteConfigIntegrationTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/backend_integration_tests/ProductionRemoteConfigIntegrationTest.kt
@@ -26,6 +26,7 @@ import kotlinx.serialization.builtins.serializer
 import kotlinx.serialization.json.jsonObject
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.After
+import org.junit.Assume
 import org.junit.Test
 import java.io.File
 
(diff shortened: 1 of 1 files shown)


ajpallares (approved):
Makes sense!

PR #3855, opened 2026-07-29
feat(remote-config): use the server clock for X-RC-Last-Refresh-Time

tonidero:
Follow-up to #3853 (merged), which sent `dateProvider.now` as the refresh time. The backend compares the replayed value against its own clock, so a skewed device clock silently corrupts the optimization. This round-trips the server's own `X-RevenueCat-Request-Time` instead.

The RC-format response path already parses that header into `HTTPResult.requestDate` (including on a 204) — `getRemoteConfig` was just discarding it before `RemoteConfigManager` could see it, so this mostly widens the success callback.

- A response with no such header carries the last server-supplied value forward, rather than substituting local time. Nothing device-derived is ever sent.
- `lastRefreshedAt` stays on the device clock: `isCacheStale` measures elapsed *local* time, so server time there would produce a bogus interval. The two timestamps now use different clocks deliberately.
- The fallback endpoint contributes no refresh time: different host, different clock.

---

> [!NOTE]
> **Medium Risk**
> Changes remote config sync bookkeeping and request headers the server uses for freshness; scope is limited to metadata timing, not config parsing or auth.
> 
> **Overview**
> Remote config no longer persists the device clock as `lastRefreshTime` for `X-RC-Last-Refresh-Time`. It now round-trips the server’s **`X-RevenueCat-Request-Time`** from each successful `/v1/config` response (200 or 204).
> 
> **`getRemoteConfig`** passes `HTTPResult.requestDate` through its success callback (new middle parameter). **`HTTPClient`** parses that header with `toLongOrNull()` so bad values become null instead of crashing the request path.
> 
> **`RemoteConfigManager`** writes `requestDate` into persisted `lastRefreshTime` on 200 and 204; if the header is missing, the previous server value is kept and a warning is logged. **`lastRefreshedAt`** (staleness) still uses the device clock. The **fallback** config path never sets refresh time from the fallback host.
> 
> Tests and a production integration check cover header presence on 200/204 and the new persistence rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddf1b82c1a37ba60cce0787978e10638bfc253da. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetRemoteConfigTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetRemoteConfigTest.kt
@@ -106,7 +106,7 @@ class BackendGetRemoteConfigTest {
             manifest = testManifest,
             lastRefreshTime = null,
             prefetchedBlobs = testPrefetchedBlobs,
-            onSuccess = { result, verificationResult ->
+            onSuccess = { result, _, verificationResult ->
                 container = result
                 verification = verificationResult
             },
@@ -156,7 +156,7 @@ class BackendGetRemoteConfigTest {
             manifest = testManifest,
             lastRefreshTime = null,
             prefetchedBlobs = testPrefetchedBlobs,
-            onSuccess = { result, _ -> container = result },
+            onSuccess = { result, _, _ -> container = result },
             onError = { error, _ -> fail("Expected success. Got error: $error") },
         )
 
@@ -180,7 +180,7 @@ class BackendGetRemoteConfigTest {
             manifest = testManifest,
             lastRefreshTime = null,
             prefetchedBlobs = testPrefetchedBlobs,
-            onSuccess = { _, _ -> fail("Expected error. Got success") },
+            onSuccess = { _, _, _ -> fail("Expected error. Got success") },
             onError = { error, behavior ->
                 obtainedError = error
                 obtainedBehavior = behavior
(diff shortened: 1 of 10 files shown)


purchases/src/test/java/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManagerTest.kt
rickvdl, line 1265: Should we also assert on the last refresh time being nil here for the fallback endpoint?

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetRemoteConfigTest.kt
```diff
@@ -242,6 +242,84 @@ class BackendGetRemoteConfigTest {
         assertThat(verification).isEqualTo(VerificationResult.VERIFIED)
     }
 
+    @Test
+    fun `getRemoteConfig hands the server request date to the success callback on a 200`() {
```
rickvdl, line 246: Should we also add a test / assertion that the backend does return this server time in normal operations? That should give us some insights if this mechanism breaks. Since it's a bit more brittle now that we're no longer constructing the timestamp locally.

rickvdl (approved):
Looking good! Just two tests we should add I think

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendGetRemoteConfigTest.kt
tonidero: Added these tests in https://github.com/RevenueCat/purchases-android/pull/3855/commits/ddf1b82c1a37ba60cce0787978e10638bfc253da! And also added a log in case the header is missing in the response for visibility

PR #3856, opened 2026-07-29
FIX: Web View Scroll Arbitration 

JZDesign:
### Checklist

### Motivation

Web views required 2 fingers to scroll in certain situations. We need it to be 1.

[Ticket](https://linear.app/revenuecat/issue/FUN-2262/fix-fill-height-discrepancies) linear issue FUN-2262

### Description

Adjust the scroll arbitration


AlvaroBrey (approved):
Nice! One nit only: maybe the PR title needs to be more specific, since it will show up in the changelog and just "2 fingers required to scroll" makes it sound like a bigger problem

PR #3861, opened 2026-07-30
fix(remote-config): clear last attempt timestamp for cooldown after successful response

rickvdl:
The `lastRefreshAttemptAt` cooldown functionality was designed to ensure that the `RemoteConfigManager` would not immediately allow another request to be sent if a request failed, because the cache wouldn't be updated and thus still be stale.

However `lastRefreshAttemptAt` was never reset on a successful 200 or 204 response, which caused the logic to also kick in after successful responses. This was fine in production code, since the cache was only considered stale after 5 minutes, and the cooldown was just 1 minute.

However during manual testing (in which I lowered the cache staleness expiration time) this means that we can never really lower the value below the cooldown time (1 minute).

Also added test coverage for this, because this should've been caught really.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to clearing attempt timestamps after successful remote-config responses; production staleness (5 minutes) already dwarfed the 1-minute cooldown, so risk is mainly correctness for shorter windows and testability.
> 
> **Overview**
> Fixes **remote config refresh throttling** so a successful sync does not keep the failed-attempt cooldown active.
> 
> `RemoteConfigManager` sets `lastRefreshAttemptAt` on stale-gated refreshes to avoid hammering the API after failures, but it was **never cleared** on successful **200** or **204** responses. After a good response, the 1-minute cooldown could still block `refreshRemoteConfigIfStale` when the cache window is shorter (e.g. in tests or tuned staleness), effectively capping how often refreshes can run below the real cache TTL.
> 
> The fix **resets `lastRefreshAttemptAt` to null** when config is committed on **200** persist and when **204** bookkeeping completes. **`cacheDuration`** is extracted in `DateExtensions` with an injectable **`cacheDurationProvider`** on the manager so tests can use a short staleness window and assert a second refresh fires after success.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b212d443f47be11009b22db56ef727eda1c0b2f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DateExtensions.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DateExtensions.kt
@@ -15,16 +15,22 @@ import kotlin.time.Duration.Companion.minutes
 private val CACHE_REFRESH_PERIOD_IN_FOREGROUND = 5.minutes
 private val CACHE_REFRESH_PERIOD_IN_BACKGROUND = 25.hours
 
+internal fun cacheDuration(appInBackground: Boolean): Duration {
+    return when {
+        appInBackground -> CACHE_REFRESH_PERIOD_IN_BACKGROUND
+        else -> CACHE_REFRESH_PERIOD_IN_FOREGROUND
+    }
+}
+
 @OptIn(InternalRevenueCatAPI::class)
-internal fun Date?.isCacheStale(appInBackground: Boolean, dateProvider: DateProvider = DefaultDateProvider()): Boolean {
+internal fun Date?.isCacheStale(
+    appInBackground: Boolean,
+    dateProvider: DateProvider = DefaultDateProvider(),
+    cacheDurationProvider: (Boolean) -> Duration = ::cacheDuration,
+): Boolean {
     return this?.let {
         log(LogIntent.DEBUG) { ReceiptStrings.CHECKING_IF_CACHE_STALE.format(appInBackground) }
-        val cacheDuration = when {
-            appInBackground -> CACHE_REFRESH_PERIOD_IN_BACKGROUND
-            else -> CACHE_REFRESH_PERIOD_IN_FOREGROUND
-        }
-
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DateExtensions.kt
```diff
@@ -15,16 +15,18 @@ import kotlin.time.Duration.Companion.minutes
 private val CACHE_REFRESH_PERIOD_IN_FOREGROUND = 5.minutes
 private val CACHE_REFRESH_PERIOD_IN_BACKGROUND = 25.hours
 
+internal fun cacheDuration(appInBackground: Boolean): Duration {
```
tonidero, line 18: Hmm I don't grasp the reason for this refactor... seems it keeps the same behavior?

tonidero (approved):
🚢

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DateExtensions.kt
rickvdl: It is now also reused as the default value for the RemoteConfigManager [here](https://github.com/RevenueCat/purchases-android/pull/3861/changes#diff-cd3ecbe5f1846a428e3125909c02ff3830b7adce82cabdba02c0a3ff90779e9eR89)

rickvdl: But that did point me to the RemoteConfigManager case bypassing the `log()`, so I fixed that and they now both go through the codepath that still logs

tonidero: ohh right right 🤦 
 Thank you!

Issue #3867, opened 2026-07-30
`OutOfMemoryError` caching large offerings responses (`DeviceCache.cacheOfferingsResponse`, `ETagPayloadStore.read`)


pmarino:
- [ ] I have updated Purchases SDK to the latest version (we're running 10.14.1 in production)
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**

We're seeing `OutOfMemoryError` crashes on low-memory Android devices in two SDK code paths that materialize the full offerings response in memory. This is a follow-up to #3628: the `HTTPResultWithETag.serialize` OOM fixed in 10.14.x no longer occurs for us (thank you!), but the same underlying condition (a very large offerings payload) now crashes in:

1. `DeviceCache.cacheOfferingsResponse` (DeviceCache L:601) — `JSONObject.toString()` before the SharedPreferences write fails on a single 37,748,744-byte `char[]` allocation.
2. `ETagPayloadStore.read` (ETagPayloadStore L:85) — `CharsetDecoder.decode()` over the stored payload file fails on a 39,002,984-byte `CharBuffer` allocation.

Both paths hold several full copies of the payload in memory at once (e.g. for the cache write: the parsed `JSONObject` tree, the serializer's doubling buffer, and the resulting `String`), so peak memory usage is a multiple of payload size — and payload size grows with offerings × paywalls × locales. The findings below describe what we believe pushed our project across the allocation threshold now, but the thresholds themselves would apply to any sufficiently large offerings configuration, from either host.

Two findings that may help with triage:

**a) The crashes seem to correlate with the fallback host, whose offerings payload appears to be roughly 1.8× the main API's.** All crash events we sampled log a connection-level failure to `api.revenuecat.com` (DNS resolution, SSL handshake, TCP connect, or connection closed) shortly before the crash, and one stack trace passes through `HTTPClient.performRequest$performRequestToFallbackURL`, so these devices appear to be failing over to `api-production.8-lives-cat.io`. We compared the two endpoints with `curl` (representative headers, not the SDK itself, so treat sizes as approximate):

[...]

PR #3872, opened 2026-07-31
fix: avoid OutOfMemoryError when caching large offerings responses

AlvaroBrey:
- Fixes an `OutOfMemoryError` when caching a large offerings response: `cacheOfferingsResponse`
  serialized a `JSONObject`, a contiguous allocation of several times the payload size. The raw
  response body is now stored as received.
- Stops re-writing the response when it was just read from the disk cache, which made the crash
  repeat on every failed fetch with no network involved.
- `Offerings.originalSource` is no longer persisted, since injecting it into the cached body is what
  forced the re-serialization. It is now nullable, and offerings from the disk cache report null.
- Fixes a stale `baseline-prof.txt` descriptor for `getOfferings` that the callback change
  invalidated.
- The response still lives in SharedPreferences, so ~1x payload stays retained for the process
  lifetime. Only the serialization is gone.

### Checklist

<details>

Part of #3867 (the `DeviceCache.cacheOfferingsResponse` crash; `ETagPayloadStore.read` is #3873).

**Two causes**

- `cacheOfferingsResponse` took a `JSONObject` and called `toString()`. `StringBuilder` grows by
  `(cap << 1) + 2`, so the reported 37,748,744-byte failure is the doubling step to
  `char[18,874,364]`, reached once the serialized response passes ~9.44M chars.
- `OfferingsManager` re-cached the response even when it had just been read from the disk cache, so
  an offline device re-serialized and re-wrote the same JSON on every `getOfferings`.

**Changes**

- The raw response body is threaded through `Backend.getOfferings`' callback and stored as received.
  No payload-sized allocation on the write path.
- `Offerings.originalSource` is no longer persisted. The field has no runtime consumer. SDK-4408
  tracks the two sibling caches that persist the same field.
- `originalSource` is now nullable; offerings rebuilt from the disk cache report null rather than
  claiming `MAIN`.
- The disk-cache path skips only the disk write, so the in-memory cache is still populated.
- Fixed the `baseline-prof.txt` descriptor for `getOfferings` (`Function2` to `Function3`).

**Measured**

Caller-thread allocation (where the crash was) on a 5.24MB offerings-like payload, via
`ThreadMXBean#getThreadAllocatedBytes`:

| | Allocated |
|---|---|
| `cacheOfferings`, after | 147 KB |
| serialize-then-store, before | 28,595 KB (~5.5x the payload) |

The response is parsed either way to build the offerings, so the tree is built outside the measured
window on both sides: this compares only what the caching step itself adds.

**Notes**

- No new prefs keys and nothing to migrate. Old cached bodies still parse and are still served.
- Regression gates: `DeviceCacheTest` pins the stored string; `OfferingsManagerTest` asserts the
  network path caches verbatim and the disk path caches nothing.

</details>

---

> [!NOTE]
> **Medium Risk**
> Touches core offerings fetch/cache/fallback behavior and changes semantics of internal `originalSource` on disk-cache paths; mitigated by tests and no migration of existing cached bodies.
> 
> **Overview**
> Fixes **`OutOfMemoryError`** on low-heap devices when caching large offerings payloads by **storing the raw HTTP body** instead of re-serializing a `JSONObject`, and by **not writing disk again** when offerings were parsed from the existing disk cache (e.g. repeated failed fetches offline).
> 
> `Backend.getOfferings` now passes **`result.payloadText`** through the success callback so `DeviceCache.cacheOfferingsResponse` can persist the response verbatim. `OfferingsCache` no longer injects `rc_original_source` into the JSON (that mutation forced the expensive `toString()` path).
> 
> **`Offerings.originalSource`** is now **nullable** and defaults to **null** when unknown; disk-cache fallbacks use **null** instead of inferring `MAIN`. In-memory cached instances still keep the source from the build that populated them. **`baseline-prof.txt`** is updated for the `getOfferings` callback signature change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2d2b5008869fcc0e0148ec3470aa7db10d61266. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -65,8 +65,12 @@ internal typealias CallbackCacheKey = List<String>
 
 /** @suppress */
 @OptIn(InternalRevenueCatAPI::class)
+internal typealias OfferingsSuccessCallback = (JSONObject, HTTPResponseOriginalSource, responsePayload: String) -> Unit
+
+/** @suppress */
+@OptIn(InternalRevenueCatAPI::class)
 internal typealias OfferingsCallback = Pair<
-    (JSONObject, HTTPResponseOriginalSource) -> Unit,
+    OfferingsSuccessCallback,
     (PurchasesError, errorHandlingBehavior: GetOfferingsErrorHandlingBehavior) -> Unit,
     >
 
@@ -422,7 +426,7 @@ internal class Backend(
     fun getOfferings(
         appUserID: String,
         appInBackground: Boolean,
-        onSuccess: (JSONObject, HTTPResponseOriginalSource) -> Unit,
+        onSuccess: OfferingsSuccessCallback,
         onError: (PurchasesError, GetOfferingsErrorHandlingBehavior) -> Unit,
     ) {
         val endpoint = Endpoint.GetOfferings(appUserID)
@@ -454,7 +458,7 @@ internal class Backend(
                 }?.forEach { (onSuccess, onError) ->
                     if (result.isSuccessful()) {
                         try {
-                            onSuccess(result.body, result.originalDataSource)
+                            onSuccess(result.body, result.originalDataSource, result.payloadText)
                         } catch (e: JSONException) {
                             val errorBehavior = GetOfferingsErrorHandlingBehavior.SHOULD_FALLBACK_TO_CACHED_OFFERINGS
                             onError(e.toPurchasesError().also { errorLog(it) }, errorBehavior)
--- a/purchases/src/main/baseline-prof.txt
+++ b/purchases/src/main/baseline-prof.txt
@@ -628,7 +628,7 @@ HSPLcom/revenuecat/purchases/common/Backend;->addCallback(Ljava/util/Map;Lcom/re
 HSPLcom/revenuecat/purchases/common/Backend;->getCallbacks()Ljava/util/Map;
 HSPLcom/revenuecat/purchases/common/Backend;->getCustomerInfo(Ljava/lang/String;ZLkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
 PLcom/revenuecat/purchases/common/Backend;->getDiagnosticsCallbacks()Ljava/util/Map;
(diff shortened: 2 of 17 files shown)


AlvaroBrey:
@cursor review

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
```diff
@@ -454,7 +458,7 @@ internal class Backend(
                 }?.forEach { (onSuccess, onError) ->
                     if (result.isSuccessful()) {
                         try {
-                            onSuccess(result.body, result.originalDataSource)
+                            onSuccess(result.body, result.originalDataSource, result.payloadText)
```
tonidero, line 461: I expect this means we now keep both the JSONObject + string versions at the same time longer right?

But probably this should indeed save a ton in allocations + the string version should not use nearly as much memory, so I believe it's acceptable.

AlvaroBrey:
I just realized that `originalSource` is only used for tests. Given that, instead of storing it in a separate place I'm just gonna completely remove that behaviour for prod, it doesn't belong there.

AlvaroBrey:
@cursor review

AlvaroBrey:
@cursor review

AlvaroBrey:
@cursor review

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
tonidero: The alternative in my mind is to split the caching of the offerings. So we cache the string version on disk first before even confirming it's a valid offerings json, then decode it and just use the parsed version henceforth... But not sure I like that + a lot of refactoring... So yeah, I think we might be ok with this

tonidero (approved):
Let's hope this helps 🙏 Were you able to measure memory impact?

AlvaroBrey:
> Let's hope this helps 🙏 Were you able to measure memory impact?

I hadn't because it was a pretty clear improvement, but I have now. This is for caching a single 5MB payload on Robolectric (excluding the memory footprint of the payload itself ofc):

> | | Allocated |
> |---|---|
> |  before | 28,595 KB |
> |  after | 147 KB |

😁

purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
AlvaroBrey: Yeah I don't think it's worth it for now, considering this is mitigation.

PR #3873, opened 2026-07-31
fix: reduce memory reading cached responses and recover from corrupted cache entries

### LINKED ISSUE #3628: `OutOfMemoryError` in `ETagManager.serialize` when caching large /offerings response

AlvaroBrey:
- Reduces the memory needed to read a cached HTTP response: the 304 path decoded the payload file
  with a strict `CharsetDecoder`, which allocates a `char[]` the size of the whole file. Measured 3.9x
  down to 2.0x payload bytes.
- Corrupted cache entries are now detected and recover on their own. Previously a same-length
  corruption was served as a cache hit and wedged permanently, because the server kept answering 304.
- Replaces the stored file size with a CRC32, folded into the existing write loop so it costs no
  extra pass over the payload.
- This reduces the OOM but does not eliminate it: ~2x is the floor while the payload has to exist as
  a `String`. #3867 stays open.
- ETag entries written by earlier versions miss once and refetch, one extra request per cached URL on
  the first launch after updating.

### Checklist

<details>

### Motivation

Part of #3867 (the `ETagPayloadStore.read` crash). Regression introduced in **10.14.1** by #3774,
which fixed #3628 by moving the allocation from the write path to the read path.

`CharsetDecoder.decode(ByteBuffer)` allocates `char[fileSizeBytes]` (UTF-8's
`averageCharsPerByte` is 1.0), i.e. 2x the file size contiguous. The reported 39,002,984 bytes
implies a ~19.5MB payload file.

The file-size check it replaces only caught truncation. Same-length corruption was served as a
hit, and that never heals: the payload fails JSON parsing, `HTTPResult.body` swallows it, and the
server keeps answering 304.

### Description

- `write` returns a CRC32 instead of the length, folded into the existing write loop via
  `CheckedOutputStream` (no extra pass, no extra buffer).
- `read` verifies before decoding, then uses `String(bytes, UTF_8)`. Verifying first means a
  mismatch skips the payload-sized `String`.
- `expectedChecksum` is required, so an unverified read is unreachable.
- Missing checksum is a miss, no migration branch.
- `java.util.zip.CRC32` because `CRC32C` is API 34 and `MessageDigest` is too slow here.

Measured by `ETagManagerMemoryTest` on 5MB: 304 read drops from ~3.9x to **2.0x** payload bytes.
Gate tightened 6x to 2.5x.

The regression gate is the new same-length-different-content case in `ETagPayloadStoreTest`, the
one thing the old size check could not catch.

### Known limitations

1. **Reduces the OOM, does not eliminate it.** ~2x is the floor while `Payload.Text` holds a
   `String`. The reported crash had 9.7MB free with the `byte[]` already resident, so a ~19.5MB
   `String` would also have failed. A size ceiling would fix it and was declined for this pass.
   #3867 stays open.
2. Entries from 10.14.1 to 10.16.0 store a size, not a checksum, so they miss and refetch once.
3. The CRC32 CPU cost (est. 5-40ms per 20MB on low-end ARM, background thread) is an estimate,
   not measured.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes core ETag cache read/write behavior and metadata format for all cached API responses; risk is mitigated by one-time miss/refetch for legacy entries and extensive unit/memory tests, but large-payload cache hits remain ~2× heap by design (#3867 not fully closed).
> 
> **Overview**
> Fixes a regression where **`ETagPayloadStore.read`** used strict **`CharsetDecoder`**, allocating a payload-sized `char[]` (~2× file size) and contributing to OOM on large offerings-style responses (#3867).
> 
> **Integrity:** Cache metadata now stores **`payloadChecksum`** (CRC32) instead of optional **`payloadSizeBytes`**. **`write`** returns the checksum via **`CheckedOutputStream`** on the existing encode path; **`read`** requires **`expectedChecksum`**, verifies CRC32 before decoding, then uses lenient **`String(bytes, UTF_8)`** instead of **`CharsetDecoder.decode`**. That catches truncation, stale interrupted renames, and **same-length corruption** that the size check could not—entries that previously wedged behind perpetual 304s now miss once and refetch.
> 
> **Migration:** Entries from 10.14.1–10.16.0 without **`payloadChecksum`** deserialize as misses and stop sending the stored eTag so the server returns a full body.
> 
> **Tests:** Memory gate for 304 reads tightened (~2× payload vs ~3.9×); new same-length corruption and legacy-metadata cases added.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5025b614a938ce9eb68647512f5eb063d8b2c63. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagManager.kt
@@ -12,7 +12,6 @@ import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.strings.NetworkStrings
 import com.revenuecat.purchases.utils.isAndroidNOrNewer
-import com.revenuecat.purchases.utils.optNullableLong
 import org.json.JSONException
 import org.json.JSONObject
 import java.util.Date
@@ -40,8 +39,13 @@ internal data class ETagCacheMetadata(
     val verificationResult: VerificationResult,
     val isLoadShedderResponse: Boolean,
     val isFallbackURL: Boolean,
-    /** Verified on read to detect truncated payload files; `null` skips the check. */
-    val payloadSizeBytes: Long? = null,
+    /**
+     * CRC32 of the stored payload, verified on read. Kept here rather than inside the payload file
+     * because [ETagManager.storeResult] writes the file before committing these prefs: a
+     * self-describing file would still validate after a crash between the two, serving fresh bytes
+     * under the previous entry's eTag.
+     */
+    val payloadChecksum: Long,
 ) {
     fun serialize(): String {
         return JSONObject().apply {
@@ -52,7 +56,7 @@ internal data class ETagCacheMetadata(
             put(SERIALIZATION_NAME_VERIFICATION_RESULT, verificationResult.name)
             put(SERIALIZATION_NAME_IS_LOAD_SHEDDER_RESPONSE, isLoadShedderResponse)
             put(SERIALIZATION_NAME_IS_FALLBACK_URL, isFallbackURL)
-            payloadSizeBytes?.let { put(SERIALIZATION_NAME_PAYLOAD_SIZE_BYTES, it) }
+            put(SERIALIZATION_NAME_PAYLOAD_CHECKSUM, payloadChecksum)
         }.toString()
     }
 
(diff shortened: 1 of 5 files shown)


AlvaroBrey:
@cursor review

tonidero (approved):
Makes sense! Thank you!!

PR #3878, opened 2026-08-03
Add a Maestro flow for the config kill switch tripping mid-session

rickvdl:
## Description
E2E test with coverage for the scenario of triggering the remote config kill-switch mid session.

With the remote config enabled it will first go through a multi page workflow, after which it will trigger the kill-switch and re-open the workflow, asserting that now only the last screen (the paywall) is presented as a single page fallback, and ensuring the product can be purchased on the paywall.

iOS counterpart: [RevenueCat/purchases-ios#7324](https://github.com/RevenueCat/purchases-ios/pull/7324)

---

> [!NOTE]
> **Low Risk**
> Changes are mostly E2E and debug-only hooks; production impact is a small HTTPClient path that honors optional URL modification on all requests when a strategy is injected.
> 
> **Overview**
> Adds a **Maestro E2E workflow** that walks a three-screen workflow paywall, arms a remote-config kill switch mid-session, re-presents the same offering, and asserts degradation to the **classic last-page paywall** (not the generic default) with a successful purchase.
> 
> To drive that scenario in the test app, the PR extends **`ForceServerErrorStrategy`** with **`modifyRequestURL`** and wires **`HTTPClient`** to use it when requests are not redirected to the forced-error URL. Debug-only **`configureForE2ETests`** can append `force_killswitch=true` on **`GetRemoteConfig`** when armed.
> 
> The workflow E2E screen gains **Force Config Killswitch** / **Sync Attributes And Offerings** controls, sync + polling until **`resolveWorkflow`** reports **`WorkflowResolution.Disabled`**, while keeping the pre-switch **`Offering`** instance for paywall presentation. Release builds keep normal **`Purchases.configure`** via a stub configurator.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e64fdf1a7e52fe4b199d0d2e1791fc655e5aac02. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -399,13 +399,14 @@ internal class HTTPClient(
         }
 
         try {
+            val requestURL = URL(baseURL, path)
             fullURL = if (appConfig.runningTests &&
                 forceServerErrorStrategy?.shouldForceServerError(baseURL, endpoint) == true
             ) {
                 warnLog { "Forcing server error for request to ${URL(baseURL, path)}" }
                 URL(forceServerErrorStrategy.serverErrorURL)
             } else {
-                URL(baseURL, path)
+                forceServerErrorStrategy?.modifyRequestURL(requestURL, endpoint) ?: requestURL
             }
 
             nonce = if (shouldAddNonce) signingManager.createRandomNonce() else null
--- a/test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
+++ b/test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
@@ -14,6 +14,7 @@ private val forceServerErrorStrategy = AtomicReference(E2EForceServerErrorStrate
 internal fun configurePurchases(
     configuration: PurchasesConfiguration,
 ) {
+    forceServerErrorStrategy.set(E2EForceServerErrorStrategy.Never)
     Purchases.configureForE2ETests(configuration, forceServerErrorStrategy::get)
 }
 
--- a/test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
+++ b/test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
@@ -14,7 +14,6 @@ private val forceServerErrorStrategy = AtomicReference(E2EForceServerErrorStrate
 internal fun configurePurchases(
     configuration: PurchasesConfiguration,
 ) {
-    forceServerErrorStrategy.set(E2EForceServerErrorStrategy.Never)
     Purchases.configureForE2ETests(configuration, forceServerErrorStrategy::get)
 }
 
--- a/test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
+++ b/test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
@@ -1,0 +1,23 @@
+package com.revenuecat.e2etests
+
+import com.revenuecat.purchases.E2EForceServerErrorStrategy
+import com.revenuecat.purchases.InternalRevenueCatAPI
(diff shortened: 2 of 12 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -399,13 +399,14 @@ internal class HTTPClient(
...
         try {
+            val requestURL = URL(baseURL, path)
             fullURL = if (appConfig.runningTests &&
                 forceServerErrorStrategy?.shouldForceServerError(baseURL, endpoint) == true
             ) {
                 warnLog { "Forcing server error for request to ${URL(baseURL, path)}" }
                 URL(forceServerErrorStrategy.serverErrorURL)
             } else {
-                URL(baseURL, path)
+                forceServerErrorStrategy?.modifyRequestURL(requestURL, endpoint) ?: requestURL
```
tonidero, line 409: Hmm I wonder if we should somehow have the maestro tests set the `appConfig.runningTests` property to true, so we can go with the same path as above and keep the strategy outside with that flag... But well, we can keep this for now I think!

test-apps/e2etests/src/debug/java/com/revenuecat/e2etests/PurchasesConfigurator.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.revenuecat.purchases.E2EForceServerErrorStrategy
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.Purchases
+import com.revenuecat.purchases.PurchasesConfiguration
+import com.revenuecat.purchases.configureForE2ETests
+import java.util.concurrent.atomic.AtomicReference
+
+@OptIn(InternalRevenueCatAPI::class)
+private val forceServerErrorStrategy = AtomicReference(E2EForceServerErrorStrategy.Never)
```
tonidero, line 11: I guess for our current usage (maestro tests) it's not a problem, but not sure if it could be good to clean this up after each test for cleanliness... I would say it's probably not a blocker for now though.

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/workflow/WorkflowScreen.kt
```diff
@@ -98,21 +115,82 @@ fun WorkflowScreen(modifier: Modifier = Modifier) {
...
+    ) {
+        Button(onClick = E2ETestsApplication::forceConfigKillSwitch) {
+            Text("Force Config Killswitch")
+        }
+
+        Button(
+            onClick = {
+                coroutineScope.launch {
+                    if (syncAndWaitForConfigKillSwitch()) {
+                        configEndpointKillSwitchOn = true
```
tonidero: Hmm it's weird that we set this to true on the `Sync attributes and offerings` button... I wonder if it should be split to the `Force config killswitch` button instead... We might need to have a UI way to wait until the sync is done I guess...

tonidero (approved):
Looks good!! Nice to have this test 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
rickvdl: Right, I think setting `appConfig.runningTests` will also have more behavioral changes like reducing network request timeouts. I guess without it we'd have a bit of a more production-like test and have less flakiness.

test-apps/e2etests/src/main/java/com/revenuecat/e2etests/workflow/WorkflowScreen.kt
rickvdl: Good point, changed it to make it two-staged. First the kill-switch is `armed` and when it is hit and the SDK entered kill-switch enabled mode the label is updated to show `on`

purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
tonidero: Yeah indeed... let's keep as is then 👍

PR #3881, opened 2026-08-03
feat: allow getCustomerInfo to not wait for unsynced purchases

AlvaroBrey:
### Checklist

### Motivation

When there are pending purchases, `getCustomerInfo` reports only after they've been posted, on a single-threaded backend dispatcher. A customer on iOS gating their startup paywall on `getCustomerInfo` saw that post hang and blow their 6s budget; the same design exists here, and there was no public way to opt out.

Linear: SDK-4401. iOS counterpart: RevenueCat/purchases-ios#7353.

### Description

Adds an opt-in `UnsyncedTransactionsWaitPolicy` on `PurchasesConfiguration.Builder`. It defaults to `WAIT`, which is the current behavior; `DO_NOT_WAIT` stops `getCustomerInfo` from waiting on pending purchase posts.

Under `DO_NOT_WAIT`, the sync moves to a dedicated dispatcher and the caller is answered by one of:

| Situation | Answer |
|---|---|
| Something cached | Fetch from the backend, concurrently with the sync |
| Nothing cached (first launch) | `CustomerInfo` computed from the device's purchases |
| Nothing cached *and* can't compute | Wait for the sync (unchanged behavior, as a floor) |

Server truth then arrives through `UpdatedCustomerInfoListener` once posting finishes. Fetching is preferred over device computation whenever a cache exists, because device computation can't see purchases made outside the store — otherwise we'd trade a hang for lost entitlements.

Notes for reviewers:

- **Device computation deliberately doesn't enter offline-entitlements mode.** It calls the calculator directly instead of `calculateAndCacheOfflineCustomerInfo`, so it doesn't set `_offlineCustomerInfo`, clear the cache, or report `trackEnteredOfflineEntitlementsMode` — the backend is reachable here, this is the app's choice. Matches what iOS does.
- **`receiptPostDispatcher`** follows the existing `remoteConfigDispatcher: Dispatcher = dispatcher` pattern, defaulting to the shared dispatcher so `WAIT` is unchanged. `Backend.close()` now closes it, via a set that also stops double-closing when the defaults collapse.
- **Because posts and reads can now interleave**, a guard drops a cache write whose `CustomerInfo` is staler than what this session already wrote. Gated on the policy so `WAIT` apps are untouched.
- Small extras in the neighbourhood: `DeviceCache.hasCachedCustomerInfo` avoids decoding a blob just to null-check it (also used by the pre-existing `shouldCalculateOfflineCustomerInfoInGetCustomerInfoRequest`), and `computeOfflineCustomerInfo` now guards on the full `isOfflineEntitlementsEnabled()` rather than the flag alone.

### Testing

Beyond the added unit tests: full `:purchases:testDefaultsDebugUnitTest` run is green (4202 tests), `detektAll` clean, and `scripts/api-check.sh` confirms the Metalava signatures change is additive only.

---

> [!NOTE]
> **Medium Risk**
> Changes core `getCustomerInfo` timing and cache behavior when opted in; default `WAIT` is unchanged but `DO_NOT_WAIT` apps must handle `UpdatedCustomerInfoListener` for correct entitlements.
> 
> **Overview**
> Adds **`UnsyncedTransactionsWaitPolicy`** on `PurchasesConfiguration` (default **`WAIT`**, unchanged behavior). **`DO_NOT_WAIT`** lets `getCustomerInfo` return without blocking on pending purchase posts.
> 
> With **`DO_NOT_WAIT`**, receipt posts use a dedicated dispatcher so they don’t head-of-line block subscriber reads. The immediate answer depends on cache: **backend fetch** when `CustomerInfo` is cached (sync runs in parallel), **on-device computation** on first launch with no cache, or **wait for sync** when computation isn’t possible. Fresher state after posts is expected via **`UpdatedCustomerInfoListener`**.
> 
> **`CustomerInfoUpdateHandler`** skips caching **`CustomerInfo`** that is older than what the session already wrote when **`DO_NOT_WAIT`** is set, avoiding stale GET responses overwriting fresher POST results.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d1354cba50d985f1d3cb2eeb28af1384ed7b53f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/api-defauts.txt
+++ b/purchases/api-defauts.txt
@@ -546,6 +546,7 @@ package com.revenuecat.purchases {
     method @InaccessibleFromKotlin public final java.util.concurrent.ExecutorService? getService();
     method @InaccessibleFromKotlin public final boolean getShowInAppMessagesAutomatically();
     method @InaccessibleFromKotlin public final com.revenuecat.purchases.Store getStore();
+    method @InaccessibleFromKotlin public final com.revenuecat.purchases.UnsyncedTransactionsWaitPolicy getUnsyncedTransactionsWaitPolicy();
     method @InaccessibleFromKotlin public final com.revenuecat.purchases.EntitlementVerificationMode getVerificationMode();
     property public final String apiKey;
     property public final String? appUserID;
@@ -561,6 +562,7 @@ package com.revenuecat.purchases {
     property public final java.util.concurrent.ExecutorService? service;
     property public final boolean showInAppMessagesAutomatically;
     property public final com.revenuecat.purchases.Store store;
+    property public final com.revenuecat.purchases.UnsyncedTransactionsWaitPolicy unsyncedTransactionsWaitPolicy;
     property public final com.revenuecat.purchases.EntitlementVerificationMode verificationMode;
   }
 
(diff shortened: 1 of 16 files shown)


AlvaroBrey:
@cursor review

AlvaroBrey:
@cursor review

PR #3863, opened 2026-07-30
refactor(checkpoints): Checkpoints wiring and sample usage in paywall tester - SDK-4423

tonidero:
Stacked on top of #3889 

This wires checkpoints into the SDK entry points and adds example usage to the paywall-tester app.

https://github.com/user-attachments/assets/4556311f-ee06-4479-8e56-4b1df53dcddc

---

> [!NOTE]
> **Medium Risk**
> Checkpoints can auto-present paywalls/workflows from app code paths and depend on correct current-activity tracking; changes touch core orchestration but stay behind InternalRevenueCatAPI.
> 
> **Overview**
> **Wires checkpoints through the public SDK surface** so `awaitCheckpoint` actually runs workflows instead of throwing unsupported, and **`checkpointListener` is delegated** through `PurchasesOrchestrator` / `CheckpointsManager` like other global listeners.
> 
> In **`PurchasesOrchestrator`**, a **`CheckpointsManager`** is constructed with **`RandomWorkflowCheckpointResolver`** and **`UiCheckpointWorkflowExecutor`**, backed by **`CurrentActivityTracker`** on activity lifecycle so UI checkpoints can use the foreground activity.
> 
> The **paywall-tester** app adds a **Checkpoints** screen (manual hit, result display, persisted recents), navigation from App Info, and a global **`CheckpointListener`** in **`MainApplication`** (log + toast on hit).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6d84040d49b7d0e39159c47efbc80840c8afba0f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3887, opened 2026-08-04
Checkpoints public API v0 (Internally annotated APIs only) - SDK-4418

tonidero:
This PR includes a possible approach for the public API for checkpoints. Not final and marked as internal until we're comfortable, but we can use this to start iterating

---

> [!NOTE]
> **Low Risk**
> Adds new internal types and a stub extension with no runtime checkpoint behavior; existing purchase flows are unchanged unless future code wires the listener.
> 
> **Overview**
> Introduces an **internal** (`@InternalRevenueCatAPI`) checkpoint API so apps can eventually register dashboard-defined checkpoints, observe lifecycle, and auto-present workflows without a core→UI compile dependency.
> 
> **`Purchases`** gains `checkpointListener` for global hit/completion callbacks. A new `checkpoints` package defines `CheckpointInfo`, `CheckpointParams` (typed custom properties for targeting), `CheckpointResult` (`PaywallPresented` vs `NoAction` with reasons like holdout/frequency cap), `CheckpointPaywallOutcome`, and `CheckpointPresenter` / `CheckpointWorkflowPresentation` for **ServiceLoader**-based UI presentation.
> 
> **`awaitCheckpoint`** is added as the coroutine counterpart to the planned `Purchases.checkpoint` API but **always throws** `UnsupportedError` (“not implemented yet”). No orchestration or real checkpoint handling ships in this PR—types and wiring hooks only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit faaeba36f7442f3216408a79facce683ecb7573b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -331,3 +333,24 @@ public suspend fun Purchases.awaitCreateSupportTicket(email: String, description
...
+ * @return The [CheckpointResult] for this checkpoint.
+ */
+@Suppress("UnusedParameter")
+@JvmSynthetic
+@Throws(PurchasesException::class)
+@InternalRevenueCatAPI
+public suspend fun Purchases.awaitCheckpoint(
+    checkpointIdentifier: String,
+    params: CheckpointParams? = null,
+): CheckpointResult = throw PurchasesException(
```
tonidero, line 354: I decided to keep the CheckpointResult type (With some renames). Even if we change it later, I think it's good to start having a return type on the API from the beginning, so we make sure we support it and work towards that. This is not what we talked yesterday but lmk if you have any concerns! @AlvaroBrey @rickvdl

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointParams.kt
```diff
@@ -0,0 +1,34 @@
...
+    customProperties: Map<String, Any?> = emptyMap(),
+) {
+
+    public constructor(vararg customProperties: Pair<String, Any?>) : this(customProperties.toMap())
+
+    public val customProperties: Map<String, Any> = customProperties.mapNotNull { (key, value) ->
+        when (value) {
+            is String, is Number, is Boolean -> key to value
+            else -> {
+                warnLog { "Dropping invalid checkpoint custom property '$key': ${value?.javaClass?.name ?: "null"}" }
```
tonidero, line 22: We could make this part of the API, at the cost of usability... I think I kinda prefer to keep the API simple and add this verification... but we can iterate on it later.

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointResult.kt
```diff
@@ -0,0 +1,50 @@
...
+     * Nothing is served for this checkpoint; the user continues uninterrupted.
+     */
+    @Poko
+    public class NoAction internal constructor(
+        override val checkpoint: CheckpointInfo,
+        public val reason: Reason,
+    ) : CheckpointResult() {
+
+        @Poko
+        public class Reason internal constructor(public val value: String) {
```
tonidero, line 30: I think these are not final... but we can start with these and prune them/change them later if needed. I think it's a decent start though.

tonidero:
* **#3900** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3900?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3898** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3863** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3863?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3889** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3889?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3888** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3887** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3887?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowPresentation.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.revenuecat.purchases.UiConfig
+import com.revenuecat.purchases.common.workflows.PublishedWorkflow
+import dev.drewhamilton.poko.Poko
+
+/**
+ * Everything a [CheckpointPresenter] needs to present the workflow resolved for a checkpoint.
+ */
+@InternalRevenueCatAPI
+@Poko
+public class CheckpointWorkflowPresentation internal constructor(
```
tonidero, line 14: This is actually meant to be kept as internal, but it's public to be used by the UI module. But won't be publicly available in the end.

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -427,6 +438,39 @@ public class Purchases internal constructor(
...
+    @InternalRevenueCatAPI
+    public fun checkpoint(
+        checkpointIdentifier: String,
+        params: CheckpointParams?,
+        callback: CheckpointCallback,
+    ) {
+        callback.onError(
+            PurchasesError(PurchasesErrorCode.UnsupportedError, "Checkpoints are not implemented yet."),
+        )
+    }
```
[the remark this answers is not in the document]
tonidero: I mostly want to keep the final API in mind, for once it's public, so I think this is ok.

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointPaywallOutcome.kt
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Terminal result of a checkpoint-presented paywall. Delivered in [CheckpointResult.PaywallPresented] and
+ * through [CheckpointListener.onCheckpointPaywallFinished].
+ */
+@InternalRevenueCatAPI
+public abstract class CheckpointPaywallResult internal constructor() {
+
+    public class Dismissed : CheckpointPaywallResult() {
+        override fun toString(): String = "Dismissed"
+    }
```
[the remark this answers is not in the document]
tonidero: This is right... fixing

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -427,6 +438,39 @@ public class Purchases internal constructor(
...
+     * Registers that [checkpointIdentifier] was reached. Depending on the configured targeting rules, this may
+     * auto-present an experience (the call resolves when it finishes) or do nothing.
+     * @param [checkpointIdentifier] The checkpoint identifier, as configured in the RevenueCat dashboard.
+     * @param [callback] Response callback
+     */
+    @InternalRevenueCatAPI
+    public fun checkpoint(
+        checkpointIdentifier: String,
+        callback: CheckpointCallback,
+    ) {
```
vegaro: thoughts on just having a coroutine version? I would love if we could start just making coroutine APIs 😄 

we could also think of offering a Flow or similar instead of a listener

Just some thoughts

tonidero: Yeah... I wanted to keep the public API for callbacks as well... But I'm ok removing it for now. Will do so

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointPaywallResult.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.PurchasesError
+import dev.drewhamilton.poko.Poko
+
+/**
+ * Terminal result of a checkpoint-presented paywall. Delivered in [CheckpointResult.PaywallPresented] and
+ * through [CheckpointListener.onCheckpointPaywallFinished].
+ */
+@InternalRevenueCatAPI
+public abstract class CheckpointPaywallResult internal constructor() {
```
vegaro: hmm initially reading the name of the class I thought this was implementing `CheckpointResult` but I see it's not.

tonidero: Right... I see the confusion... The idea is that this will contain the information related for a specific CheckpointResult in a more accessible manner... We could collapse this data inside the CheckpointResult.PaywallPresented... Not sure what's cleaner TBH. Open to suggestions.

vegaro: I think just a rename would do. What do you think of `Outcome`?

tonidero: That could work 👍

tonidero: Renamed done in https://github.com/RevenueCat/purchases-android/pull/3887/commits/21b91609a333c613bf4abbab8b8a92b20dd24a24

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: Removed in https://github.com/RevenueCat/purchases-android/pull/3887/commits/39bb7b867cc34b9988108fbebc3935b19406d8f1

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -331,3 +333,24 @@ public suspend fun Purchases.awaitCreateSupportTicket(email: String, description
         )
     }
 }
+
+/**
+ * Registers that [checkpointIdentifier] was reached. Depending on the configured targeting rules, this may
```
rickvdl: Should we change 'reached' to 'hit' here as well?

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointListener.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+@InternalRevenueCatAPI
+public interface CheckpointListener {
+
+    /** A checkpoint was hit, before evaluation. */
+    public fun onCheckpointHit(checkpoint: CheckpointInfo) {
+        // Default empty implementation
+    }
+
+    /** The outcome was decided. Same result the call site receives. */
```
rickvdl: Maybe we should tweak the comment to say something like "the checkpoint was resolved and the outcome was return" or similar. It now reads like this method is called once the outcome was decided, but before anything was done, but the method name is clear enough in the sense that it mentions that the checkpoint "finished"

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointParams.kt
rickvdl: With part of the API, do you mean accepting an enum or like specific types? In that case I'd say this is probably fine. Otherwise I think every developer would just do exactly this themselves 😅

rickvdl (approved):
Nice! I think this is a great start

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointParams.kt
tonidero: Yes, that's what I meant 😅 (accepting specific types as values)... But yeah, I think this makes for a more ergonomic API indeed.

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/3887/commits/e8d2494d99f89d22f79fbc53dfb34153daef4bfe

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointListener.kt
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/3887/commits/e8d2494d99f89d22f79fbc53dfb34153daef4bfe

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowPresentation.kt
```diff
@@ -0,0 +1,19 @@
...
+    public val checkpoint: CheckpointInfo,
+    public val workflow: PublishedWorkflow,
+    public val uiConfig: UiConfig,
+    public val offering: Offering?,
```
facumenzella: Is there a reason to have offering as optional, but uiConfig as non optional?

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointPaywallOutcome.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.PurchasesError
+import dev.drewhamilton.poko.Poko
+
+/**
+ * Terminal result of a checkpoint-presented paywall. Delivered in [CheckpointResult.PaywallPresented] and
+ * through [CheckpointListener.onCheckpointPaywallFinished].
+ */
+@InternalRevenueCatAPI
+public abstract class CheckpointPaywallOutcome internal constructor() {
```
facumenzella, line 13: One thing I don't really get, but I want to see the final state is why we need this new type, if we already have a way of listening to a paywall output in place. 

What would this give us that we currently don't have?

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowPresentation.kt
tonidero: Right... we can probably make offering non optional right now. The idea was to eventually support non-ui workflows but in that case, the `uiConfig` should also be nullable, so we would need to do changes in any case. So let's make it non nullable for now and support the current cases, and worry about the future later (This API won't actually be part of the public API and remain as internal)

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointPaywallOutcome.kt
facumenzella: Thinking out loud... there might be multiple checkpoints and we need to know which one triggered the presentation?

tonidero: Not really... the idea here is to return the result of the final step of the workflow. In case that's a paywall UI, this would be it. If for example, it was an ad, we would have a different `CheckpointResult` for that... But now that you mention it, it's true we could definitely use the existing `PaywallResult` type. My main reason to make a new one was because that one was a sealed class, and in case we need to add new types, it would be a breaking change. Whereas for this new API, we made it as an abstract class from the beginning, so it's easier to add new cases...

Not completely opposed to reuse `PaywallResult` but I would prefer to keep it like this so we can eventually move away from that API.

facumenzella: Oh true, this makes sense

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointListener.kt
```diff
@@ -0,0 +1,26 @@
...
+@InternalRevenueCatAPI
+public interface CheckpointListener {
+
+    /** A checkpoint was hit, before evaluation. */
+    public fun onCheckpointHit(checkpoint: CheckpointInfo) {
+        // Default empty implementation
+    }
+
+    /** The checkpoint was resolved and the outcome was returned. */
+    public fun onCheckpointResolved(checkpoint: CheckpointInfo, result: CheckpointResult) {
```
vegaro: initially I thought Resolved meant returned by the backend, or that "evaluation finished". But it actually means "the entire interaction completed". What about naming it `onCheckpointCompleted`?

Also thinking about the order of things. Is this the right order?

- For `NoAction`: `onCheckpointHit` → `onCheckpointResolved(NoAction)`, fires immediately.
- For `Paywall`: `onCheckpointHit` → (evaluates, presents paywall, user buys, restores, errors or dismisses) → `onCheckpointPaywallFinished(outcome)` → `onCheckpointResolved(PaywallPresented(outcome))`.

The outcome in `onCheckpointPaywallFinished` is the same one inside PaywallPresented?

And would it make sense to have a callback when the rules are evaluated, before presenting? Something like `onCheckpointEvaluated`. Right now there's no way to know a checkpoint matched until the user is done with the paywall. (I think we've discussed this already but I don't remember correctly)

vegaro (commented):
I think it looks very good. Just one question

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointListener.kt
tonidero: Yes to the order you mentioned!

> The outcome in onCheckpointPaywallFinished is the same one inside PaywallPresented?

And you're right... `onCheckpointPaywallFinished` is probably superfluous, and just does the "casting" to the subclass automatically, so devs can listen to that specific type only if they want to and avoid the subclass handling. But I think that may be early optimizing. I will remove `onCheckpointPaywallFinished` for now.

> Something like onCheckpointEvaluated

And yes, I think adding a `onCheckpointEvaluated` will make sense indeed. But we would also need to add some API to return the result of that evaluation (which right now is internal). If you don't mind I will do that in follow-up PRs!

vegaro (approved):

tonidero:
FYI @rickvdl I did some changes to the API [here](https://github.com/RevenueCat/purchases-android/pull/3887/commits/faaeba36f7442f3216408a79facce683ecb7573b) 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowPresentation.kt
tonidero: Made it non-optional

tonidero:
### Merge activity

* **Aug 5, 11:32 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/RevenueCat/purchases-android/3887).
* **Aug 5, 11:33 AM UTC**: [Graphite](https://app.graphite.com/github/pr/RevenueCat/purchases-android/3887) couldn't merge this PR because it failed for an unknown reason (This repository has GitHub's merge queue enabled. Please configure the GitHub merge queue integration in Graphite settings.).

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
```diff
@@ -186,6 +187,14 @@ public class Purchases internal constructor(
     @InternalRevenueCatAPI
     public var debugEventListener: DebugEventListener? by purchasesOrchestrator::debugEventListener
 
+    /**
+     * Global listener for checkpoint activity, including the disposition of checkpoint-presented experiences.
+     */
+    @get:JvmSynthetic
+    @set:JvmSynthetic
+    @InternalRevenueCatAPI
+    public var checkpointListener: CheckpointListener? = null
```
JayShortway, line 196: Is this gonna be public? If so, what's the use case?

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
```diff
@@ -331,3 +333,24 @@ public suspend fun Purchases.awaitCreateSupportTicket(email: String, description
...
+ * @return The [CheckpointResult] for this checkpoint.
+ */
+@Suppress("UnusedParameter")
+@JvmSynthetic
+@Throws(PurchasesException::class)
+@InternalRevenueCatAPI
+public suspend fun Purchases.awaitCheckpoint(
+    checkpointIdentifier: String,
+    params: CheckpointParams? = null,
+): CheckpointResult = throw PurchasesException(
```
JayShortway, line 354: I suppose this is suspend because we might need to do disk reads to evaluate a rule? Would be really nice if it would be sync, but I definitely agree we shouldn't hide disk reads behind sync methods.

JayShortway (commented):
Had some questions and decided to ask them here, close to the code. Nothing against the mergeability of this PR!

purchases/src/defaults/kotlin/com/revenuecat/purchases/Purchases.kt
tonidero: The idea is yes! The use case would be mostly for tracking purposes, so the developer can track what checkpoints are happening and their outcomes from a single place and even receive real time events on what's happening on the checkpoints if, for example, there is a UI workflow and we take over the screen.

Probably not required for V0... but it was an easy addition I think.

purchases/src/defaults/kotlin/com/revenuecat/purchases/coroutinesExtensions.kt
tonidero: Yes! I originally also added a callback based solution (with optional result handler), but we decided to leave it for later. In any case, that can be built easily on top of this one.

PR #3888, opened 2026-08-04
refactor(checkpoints): Checkpoints core engine - SDK-4423

tonidero:
Stacked on top of RevenueCat/purchases-android#3887

This implements the internal engine that processes a checkpoint: deciding which workflow (if any) should run for it and running it. For this PoC, checkpoint matching is mocked and a random workflow is picked from those available. Note that none of this code is yet used, and will be in the follow-up PRs.

```mermaid
flowchart TD
    Entry["checkpoint() entry points<br/>(wired in #3863)"] --> Manager[CheckpointsManager]
    Manager -- "1. resolve(checkpoint)" --> Resolver[CheckpointWorkflowResolver]
    Resolver -. PoC impl .-> Random[RandomWorkflowCheckpointResolver]
    Random --> WM["WorkflowManager<br/>(workflows config topic)"]
    Random --> UCP[UiConfigProvider]
    Random --> Off[cached Offerings]
    Manager -- "2. execute(presentation)" --> Executor[CheckpointWorkflowExecutor]
    Executor -. PoC impl .-> UiExec[UiCheckpointWorkflowExecutor]
    UiExec --> Tracker[CurrentActivityTracker]
    UiExec -- ServiceLoader --> Presenter["CheckpointPresenter<br/>(implemented by the UI module in #3889)"]
```

---

> [!NOTE]
> **Medium Risk**<br>New paywall presentation path with concurrency and Activity lifecycle assumptions, but internal-only and not yet connected to the public API.
>
> **Overview**<br>Adds the **internal checkpoints engine** that resolves a checkpoint to a workflow and runs it. Nothing in this PR wires it to public `checkpoint()` entry points yet.
>
> `CheckpointsManager` orchestrates hits on the main dispatcher: listener `onCheckpointHit` / `onCheckpointCompleted`, `CheckpointWorkflowResolver`, then `CheckpointWorkflowExecutor`, mapping outcomes to `CheckpointResult` (`NoAction`, `PaywallPresented`, or thrown `PurchasesException` on failed resolution).
>
> **PoC resolution and execution:** `RandomWorkflowCheckpointResolver` picks a random offering-linked workflow from remote config (with hardcoded `unknown_checkpoint` / `error_checkpoint` for no-match and error paths). `UiCheckpointWorkflowExecutor` presents via `ServiceLoader`-loaded `CheckpointPresenter`, enforces one presentation at a time, and suspends until paywall completion.
>
[...]

PR #3889, opened 2026-08-04
refactor(checkpoints): Checkpoints workflow presenter UI - SDK-4423

tonidero:
Stacked on top of #3888 

This adds the UI side of checkpoints: presenting the workflow resolved for a checkpoint full-screen and reporting the outcome back to the SDK.

---

> [!NOTE]
> **Medium Risk**
> New purchase/paywall presentation path tied to checkpoint execution; outcome delivery and process-death edge cases need correct behavior, but scope is isolated to the new presenter/activity flow.
> 
> **Overview**
> Adds the **RevenueCat UI** side of checkpoint workflows: core can discover and invoke a `CheckpointPresenter` through **ServiceLoader** without a compile-time UI dependency.
> 
> **`CheckpointPresenterImpl`** stores delegate, presentation, and a mutable outcome in **`CheckpointCallStore`** (keyed by `callId`), then starts **`CheckpointWorkflowActivity`** with only the call id in the Intent. The activity shows **`Paywall`** with an **injected workflow** from the checkpoint resolution and mirrors **`PaywallActivity`** outcome handling: purchase/restore/errors update the store entry; on finish it reports **`onCheckpointPaywallFinished`** once and removes the entry. Missing or stale call ids (e.g. after process death) close the activity without reporting.
> 
> **ProGuard** keeps `CheckpointPresenterImpl` for reflective loading. Unit tests cover the call store, presenter wiring, ServiceLoader discovery, and graceful finish when the call is unknown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 265cf0be932d4083feb8ae5587142bb515d1dfec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/ui/revenuecatui/src/main/AndroidManifest.xml
+++ b/ui/revenuecatui/src/main/AndroidManifest.xml
@@ -9,6 +9,9 @@
                 android:name=".customercenter.CustomerCenterActivity"
                 android:theme="@android:style/Theme.NoTitleBar"
                 android:exported="false" />
+        <activity
+                android:name=".checkpoints.CheckpointWorkflowActivity"
+                android:exported="false" />
     </application>
 
 </manifest>
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
@@ -1,19 +1,24 @@
 package com.revenuecat.purchases.ui.revenuecatui.checkpoints
 
+import com.revenuecat.purchases.checkpoints.CheckpointPaywallOutcome
 import com.revenuecat.purchases.checkpoints.CheckpointPresenterDelegate
 import com.revenuecat.purchases.checkpoints.CheckpointWorkflowPresentation
 
 /**
  * Carries a checkpoint call's non-Parcelable payload across the Intent boundary into
  * [CheckpointWorkflowActivity], keyed by callId. Entries are removed when the terminal result is reported, so
- * a recreated activity (e.g. rotation) can still read its payload.
+ * a recreated activity (e.g. rotation) can still read its payload — including the outcome staged so far, which
+ * must outlive any single activity instance.
  */
 internal object CheckpointCallStore {
 
     class Entry(
         val delegate: CheckpointPresenterDelegate,
         val presentation: CheckpointWorkflowPresentation,
-    )
+    ) {
+        // Written/read on the main thread by the presenting activity.
+        var stagedOutcome: CheckpointPaywallOutcome = CheckpointPaywallOutcome.Dismissed
+    }
 
     private val entriesByCallId = mutableMapOf<String, Entry>()
 
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
@@ -27,4 +27,9 @@ internal object CheckpointCallStore {
 
     @Synchronized
     fun remove(callId: String): Entry? = entriesByCallId.remove(callId)
+
+    @Synchronized
+    fun clear() {
+        entriesByCallId.clear()
+    }
 }
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
@@ -1,0 +1,30 @@
+package com.revenuecat.purchases.ui.revenuecatui.checkpoints
+
+import com.revenuecat.purchases.checkpoints.CheckpointPresenterDelegate
+import com.revenuecat.purchases.checkpoints.CheckpointWorkflowPresentation
+
+/**
+ * Carries a checkpoint call's non-Parcelable payload across the Intent boundary into
+ * [CheckpointWorkflowActivity], keyed by callId. Entries are removed when the terminal result is reported, so
+ * a recreated activity (e.g. rotation) can still read its payload.
+ */
+internal object CheckpointCallStore {
+
+    class Entry(
+        val delegate: CheckpointPresenterDelegate,
+        val presentation: CheckpointWorkflowPresentation,
+    )
+
+    private val entriesByCallId = mutableMapOf<String, Entry>()
+
+    @Synchronized
+    fun store(callId: String, entry: Entry) {
+        entriesByCallId[callId] = entry
+    }
+
+    @Synchronized
+    fun get(callId: String): Entry? = entriesByCallId[callId]
+
+    @Synchronized
+    fun remove(callId: String): Entry? = entriesByCallId.remove(callId)
+}
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
@@ -7,7 +7,7 @@ import com.revenuecat.purchases.checkpoints.CheckpointWorkflowPresentation
 /**
  * Carries a checkpoint call's non-Parcelable payload across the Intent boundary into
  * [CheckpointWorkflowActivity], keyed by callId. Entries are removed when the terminal result is reported, so
- * a recreated activity (e.g. rotation) can still read its payload — including the outcome staged so far, which
+ * a recreated activity (e.g. rotation) can still read its payload — including the outcome cached so far, which
  * must outlive any single activity instance.
  */
 internal object CheckpointCallStore {
@@ -17,7 +17,7 @@ internal object CheckpointCallStore {
         val presentation: CheckpointWorkflowPresentation,
     ) {
         // Written/read on the main thread by the presenting activity.
-        var stagedOutcome: CheckpointPaywallOutcome = CheckpointPaywallOutcome.Dismissed
+        var outcome: CheckpointPaywallOutcome = CheckpointPaywallOutcome.Dismissed
     }
 
     private val entriesByCallId = mutableMapOf<String, Entry>()
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
@@ -1,0 +1,29 @@
+package com.revenuecat.purchases.ui.revenuecatui.checkpoints
+
+import android.app.Activity
+import android.content.Intent
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.checkpoints.CheckpointPresenter
+import com.revenuecat.purchases.checkpoints.CheckpointPresenterDelegate
+import com.revenuecat.purchases.checkpoints.CheckpointWorkflowPresentation
+
+/**
+ * [CheckpointPresenter] implementation instantiated by the core module through [java.util.ServiceLoader]
+ * (registered in `META-INF/services`).
+ */
+@InternalRevenueCatAPI
+public class CheckpointPresenterImpl : CheckpointPresenter {
+
+    override fun present(
+        activity: Activity,
+        callId: String,
+        presentation: CheckpointWorkflowPresentation,
+        delegate: CheckpointPresenterDelegate,
+    ) {
+        CheckpointCallStore.store(callId, CheckpointCallStore.Entry(delegate, presentation))
+        activity.startActivity(
+            Intent(activity, CheckpointWorkflowActivity::class.java)
+                .putExtra(CheckpointWorkflowActivity.EXTRA_CALL_ID, callId),
+        )
+    }
+}
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
@@ -16,7 +16,7 @@ import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
 
 /**
  * Presents the workflow resolved for a checkpoint and reports the terminal [CheckpointPaywallOutcome] back to
- * the core module exactly once. Terminal purchase/restore events are staged as they happen and delivered when
+ * the core module exactly once. Terminal purchase/restore events are cached as they happen and delivered when
  * the paywall dismisses (or the activity is otherwise finished), mirroring
  * [com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivity]'s result handling.
  */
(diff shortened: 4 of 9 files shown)


tonidero:
* **#3900** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3900?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3898** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3898?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3863** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3863?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3889** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3889?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3889?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3888** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3888?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3887** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3887?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
```diff
@@ -0,0 +1,86 @@
...
+        super.onDestroy()
+    }
+
+    private val resultStagingListener = object : PaywallListener {
+        override fun onPurchaseCompleted(customerInfo: CustomerInfo, storeTransaction: StoreTransaction) {
+            stagedResult = CheckpointPaywallOutcome.Purchased(customerInfo)
+        }
+
+        override fun onRestoreCompleted(customerInfo: CustomerInfo) {
+            stagedResult = CheckpointPaywallOutcome.Restored(customerInfo)
```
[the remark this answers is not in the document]
tonidero: Hmm right... will fix this

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
```diff
@@ -0,0 +1,86 @@
...
+        callId = intent.getStringExtra(EXTRA_CALL_ID)
+        entry = callId?.let { CheckpointCallStore.get(it) }
+        val presentation = entry?.presentation ?: run {
+            // Process death or stray relaunch: the core-side pending call died with the process, so there is
+            // nothing to report to.
+            finish()
+            return
+        }
+        val options = PaywallOptions.Builder(dismissRequest = ::finish)
+            .injectedWorkflow(presentation.workflow, presentation.offering, presentation.uiConfig)
```
tonidero, line 46: Would love the input from the workflows team to make sure this is the preferred approach to display a workflow UI 🙏. We might need to iterate on this later 🤷 cc @vegaro @facumenzella

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
```diff
@@ -0,0 +1,29 @@
...
+        activity: Activity,
+        callId: String,
+        presentation: CheckpointWorkflowPresentation,
+        delegate: CheckpointPresenterDelegate,
+    ) {
+        CheckpointCallStore.store(callId, CheckpointCallStore.Entry(delegate, presentation))
+        activity.startActivity(
+            Intent(activity, CheckpointWorkflowActivity::class.java)
+                .putExtra(CheckpointWorkflowActivity.EXTRA_CALL_ID, callId),
+        )
```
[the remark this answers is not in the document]
tonidero: Hmm this should never happen I would say... In theory it would only happen if the activity is not found, but that shouldn't happen.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
```diff
@@ -0,0 +1,40 @@
...
+ * must outlive any single activity instance.
+ */
+internal object CheckpointCallStore {
+
+    class Entry(
+        val delegate: CheckpointPresenterDelegate,
+        val presentation: CheckpointWorkflowPresentation,
+    ) {
+        // Written/read on the main thread by the presenting activity.
+        var stagedOutcome: CheckpointPaywallOutcome = CheckpointPaywallOutcome.Dismissed
```
rickvdl: `stagedOutcome` sounded a bit confusing to me, would just `outcome` work or was the reasoning that since it's not final yet we'd like to indicate that?

rickvdl (approved):
This looks clean! 🙌

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
tonidero: Yeah pretty much. The actual outcome may still change later after this is set (for some outcomes only).... But having said that, I think `outcome` is probably enough for this and probably easier to understand yeah, so I will change it.

ui/revenuecatui/src/main/AndroidManifest.xml
```diff
@@ -9,6 +9,9 @@
                 android:name=".customercenter.CustomerCenterActivity"
                 android:theme="@android:style/Theme.NoTitleBar"
                 android:exported="false" />
+        <activity
+                android:name=".checkpoints.CheckpointWorkflowActivity"
+                android:exported="false" />
```
vegaro, line 14: do we need the `android:theme="@android:style/Theme.NoTitleBar"`? Like in `CustomerCenterActivity`, but it's true that `PaywallActivity` doesn't have it. I see we request it in the activity itself anyways

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
vegaro: This is what we use in the RC App and I think it's the right thing to use since it prevents calling `/workflows`.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    override fun onDestroy() {
+        if (isFinishing) {
+            report()
+        }
+        super.onDestroy()
+    }
+
+    // Outcomes are staged on the store entry (not this instance) so a configuration change doesn't reset them.
+    private val resultStagingListener = object : PaywallListener {
```
vegaro: I am a bit confused about what `staging` means here

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.revenuecat.purchases.ui.revenuecatui.PaywallOptions
+import com.revenuecat.purchases.ui.revenuecatui.helpers.Logger
+
+/**
+ * Presents the workflow resolved for a checkpoint and reports the terminal [CheckpointPaywallOutcome] back to
+ * the core module exactly once. Terminal purchase/restore events are staged as they happen and delivered when
+ * the paywall dismisses (or the activity is otherwise finished), mirroring
+ * [com.revenuecat.purchases.ui.revenuecatui.activity.PaywallActivity]'s result handling.
+ */
+internal class CheckpointWorkflowActivity : ComponentActivity() {
```
vegaro, line 23: Exit offers won't work in this `CheckpointWorkflowActivity`. I think it's fine since exit offers will be replaced with checkpoints.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
```diff
@@ -0,0 +1,29 @@
...
+@InternalRevenueCatAPI
+public class CheckpointPresenterImpl : CheckpointPresenter {
+
+    override fun present(
+        activity: Activity,
+        callId: String,
+        presentation: CheckpointWorkflowPresentation,
+        delegate: CheckpointPresenterDelegate,
+    ) {
+        CheckpointCallStore.store(callId, CheckpointCallStore.Entry(delegate, presentation))
```
vegaro, line 23: Do we need to clean up this entry and close `CheckpointWorkflowActivity` if the coroutine awaiting the checkpoint is cancelled?

ui/revenuecatui/src/main/AndroidManifest.xml
tonidero: Yeah indeed, I think requesting it in the activity itself has the benefit that we keep the rest of the developer's theme, and we just request not to use the title bar. If we do it in the manifest, it means we're overriding the developer's theme for our activity if they have any... Honestly I could see it either way. I went with the `PaywallActivity` approach for now.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
tonidero: Yeah I will rename this... But basically it meant that we store an outcome temporarily until the activity is actually finished, but the outcome may actually change. Still, I think just making this an `outcomeListener` is probably simpler and avoid confusion.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
tonidero: Ohhh good catch I haven't tested this 🥴... I do think people should ideally migrate to multipage now... but probably something we can discuss over Slack.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
tonidero: We do cleanup the entry if the coroutine is cancelled so we shouldn't have a memory leak there. It's true the activity will remain displayed though, but I don't have a good thought on how to handle that... We would need to be passing the coroutine context across activities which I'm unsure we should do... Wdyt? Maybe it's an acceptable case? as in, the UI will remain displaying even if you cancel the coroutine. In any case, once the checkpoint activity is displayed, any work in the previous activity should have been "stopped" normally.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
```diff
@@ -0,0 +1,29 @@
...
+        activity.startActivity(
+            Intent(activity, CheckpointWorkflowActivity::class.java)
+                .putExtra(CheckpointWorkflowActivity.EXTRA_CALL_ID, callId),
+        )
+    }
```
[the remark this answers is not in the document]
tonidero: Duplicate of https://github.com/RevenueCat/purchases-android/pull/3889#discussion_r3720263934

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointPresenterImpl.kt
vegaro: OK. I think it's acceptable yes

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.revenuecat.purchases.checkpoints.CheckpointPresenterDelegate
+import com.revenuecat.purchases.checkpoints.CheckpointWorkflowPresentation
+
+/**
+ * Carries a checkpoint call's non-Parcelable payload across the Intent boundary into
+ * [CheckpointWorkflowActivity], keyed by callId. Entries are removed when the terminal result is reported, so
+ * a recreated activity (e.g. rotation) can still read its payload — including the outcome staged so far, which
+ * must outlive any single activity instance.
+ */
+internal object CheckpointCallStore {
```
facumenzella, line 13: Totally random question, but this is just a synchronized map right? Feels too specific. Don't we have already something like this that we can reuse?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
facumenzella, line 23: Maybe this is obvious, but what's the advantage of having a new activity? Why can't we reuse what's already in place like PaywallActivity?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
tonidero: Well, this is a global state holder, that holds this state at the application level, instead of at the activity level. And well, the whole `Purchases` instance is the same I guess 😅. We could try to tie it to the purchases instance lifetime, but it's trickier since this lives in the UI module... We could potentially expose it as an internally annotated API though... Could be a slightly cleaner way I guess, and we should actually fix the orphaned calls here which I got confused for a bit... So I will see about doing that change (moving it to `Purchases`)

Regarding specificity, I think I prefer that, since we need to make sure to handle this correctly... Not sure if you have some thoughts though?

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointWorkflowActivity.kt
tonidero: Not a bad question! Had a few thoughts to make it different:
- `PaywallActivity` has a lot of extra stuff that we don't need for this and it's mostly legacy from Paywalls V1 so it felt like a good opportunity to get a clean slate (like the `FontProvider` or the `shouldDisplayDismissButton`)
- For the workflows activity, we handle the whole presentation and result ourselves and needs to also notify the relevant checkpoint listener this is a bit different than the PaywallActivity, and trying to put it in felt more work than it was worth

Now, it's probably not impossible to unify them into one... But honestly, I think the way we use it is different and felt cleaner to keep them separate. Lmk if you have other thoughts. Happy to chat!

facumenzella: That makes sense. thank you 👯 🍬

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointCallStore.kt
facumenzella: > Well, this is a global state holder, that holds this state at the application level, instead of at the activity level. 

Ok I didn't understand this, and I guess it changes my opinion about it haha. Definitely not a blocker. I am not challenging if it's useful, but the code looked simple enough to have a generalization of it. Let's finish the whole thing, and revisit this once everything is in place

PR #3894, opened 2026-08-05
Only error-log an unresolvable ui_config when one is published

tonidero:
### Motivation

A project with no paywalls configured legitimately publishes no `ui_config` at all. The offerings readiness gate treated any missing `ui_config` as a failure, so developers not using paywalls/workflows got a spurious ERROR log on every cold `getOfferings`.

---

> [!NOTE]
> **Low Risk**
> Logging and internal resolution classification only; getOfferings still completes best-effort and existing getUiConfig behavior for successful resolves is preserved.
> 
> **Overview**
> Stops spurious **ERROR** logs on cold `getOfferings` when a project has no paywalls and never publishes `ui_config`.
> 
> **`UiConfigProvider`** now exposes `resolveUiConfig()` with a `UiConfigResolution` outcome (`Found`, `NotConfigured`, `Disabled`, `Superseded`, `Unavailable`). Missing or empty topics are treated as valid “nothing to resolve,” not failures; only published parts that still won’t merge/decode map to `Unavailable`. `getUiConfig()` stays a `UiConfig?` view of `Found` only.
> 
> **`WorkflowManager.onPaywallConfigReady`** uses that classification in `readyUiConfig()`: it error-logs only `Unavailable` (and thrown errors); `NotConfigured`, disabled remote config, and superseded resolves log at verbose/debug and still complete the gate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ded9aaf1e0ba1cb0e07a8fc69dc50510f3c29003. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3896, opened 2026-08-05
Stop background re-syncing /v1/config on every getOfferings for projects without paywalls

tonidero:
Stacked on #3894.

### Problem

A project with no paywalls configured still receives the `ui_config` and `workflows` topics — committed, but with no items in them. Every read of a ui_config part therefore misses, and a miss self-primes a `/v1/config` sync. Since a successful sync clears the only throttle (the failed-attempt cooldown), that meant a fresh config request on **every** `getOfferings`, forever.

With this PR, we would use the existing staleness check before performing this operation so we keep the config requests in this scenario to a minimum.

---

> [!NOTE]
> **Medium Risk**
> Changes remote config read/sync behavior on a hot path (`getOfferings`); logic is well-tested but mis-staleness could delay discovering new paywalls until the next ordinary refresh.
> 
> **Overview**
> Stops **every** `getOfferings` from firing `/v1/config` when paywall topics are committed but empty (no paywalls). Cold reads used to self-prime a sync on any miss; a successful sync cleared throttles, so repeated misses kept re-fetching forever.
> 
> **`awaitConfigForRead`** now calls **`refreshRemoteConfigIfStale`** instead of always starting a stale-gated refresh. Against **fresh** committed config, missing items or inactive topics are treated as the server’s answer—no extra network call. After the staleness window, **`clearCache`**, or when nothing was ever committed (failed persist), priming still runs as before.
> 
> **`refreshRemoteConfigIfStale`** logs when it skips because config is still fresh. Integration tests cover repeated `onPaywallConfigReady` with no paywalls (single config request) and picking up a paywall on the next stale-driven sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 017a797ecb7e4eeb659ed632f21c9c8c7039ca84. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigManager.kt
@@ -72,8 +72,10 @@ import kotlin.time.Duration.Companion.minutes
  * Consumers read through the facade: [topic] for a topic's committed item index (metadata only) and [blobData]
  * for a resolved item's blob payload (fetched on demand). Both run on [ioDispatcher] so callers never touch disk
  * on their own thread. When either read finds no committed data it calls [awaitConfigForRead] rather than
- * failing — waiting for a refresh in progress, or triggering one on demand when none is (a cold read fetches its
- * own data) — unless the endpoint is [isDisabled] or no app user is known yet.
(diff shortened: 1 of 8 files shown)


rickvdl (approved):
Looks good!

tonidero:
Failing test is unrelated to the changes in this branch and we're working on a fix, so merging this for now.

PR #3903, opened 2026-08-06
Default missing text localizations to empty instead of fallback paywall

vegaro:
Android displays the fallback paywall when a base text localization is missing. iOS keeps rendering the configured paywall and uses an empty string for the missing value. This is a recurring point of confusion and folks normally refer to iOS as the expected behavior. I think we should match it for now.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to paywall text resolution for missing base localizations; override paths and image/video localization rules are unchanged.
> 
> **Overview**
> **Missing base text localizations no longer force the fallback paywall on Android.** When a `TextComponent`’s `text_lid` is absent or not a string for a locale, the paywall still validates as components and that locale shows **empty text**, aligned with iOS.
> 
> A new `stringForAllLocalesOrEmpty` helper logs a warning per missing locale and supplies `""`. `createTextComponentStyle` uses it instead of `stringForAllLocales` plus the previous logic that distinguished “orphan” lids (missing in every locale) from partial gaps across locales.
> 
> **Override text** in `LocalizedTextPartial` still uses strict `stringForAllLocales`, so missing override localizations can still fail validation. Tests now expect successful component validation with empty strings where base text was missing per locale.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2372dd825f983bc3790a463e99c8a20153eaa848. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
```diff
@@ -950,19 +950,8 @@ internal class StyleFactory(
...
-                // from a frontend bug (e.g. a badge added only in an override state).
-                if (lidExistsInAnyLocale) {
-                    Result.Error(errors)
-                } else {
-                    Logger.w("Missing text for text_lid '${component.text.value}', using empty string.")
-                    Result.Success(localizations.mapValues { "" })
-                }
-            },
+        // Match iOS behavior by rendering missing base localizations as empty text instead of invalidating the paywall.
+        first = Result.Success(localizations.stringForAllLocalesOrEmpty(component.text)),
```
facumenzella, line 954: `Components.errors` is hardcoded null so `MissingStringLocalization` never reaches the UI anymore. we lose the `MissingLocalization` banner debug builds used to show and only keep the `Logger.w`. is the log enough? 🤔

facumenzella: Not sure if this is something people relied on tbh, so I guess its fine

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Localization.kt
```diff
@@ -49,6 +51,17 @@ internal fun NonEmptyMap<LocaleId, LocalizationDictionary>.stringForAllLocales(
...
+internal fun NonEmptyMap<LocaleId, LocalizationDictionary>.stringForAllLocalesOrEmpty(
+    key: LocalizationKey,
+): NonEmptyMap<LocaleId, String> =
+    mapValues { (locale, localizationDictionary) ->
+        localizationDictionary.string(key).getOrElse {
+            Logger.w(MissingStringLocalization(key, locale).message)
+            ""
+        }
+    }
+
```
facumenzella, line 71: Let's add some doc. Everything has some in this file 👀

facumenzella (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/style/StyleFactory.kt
vegaro: to be honest. I was trying to publish a broken paywall on the dashboard and it's prevented from there, so it looks like the only way to hit what I am fixing is with the RC App

PR #3907, opened 2026-08-06
Ingest checkpoint rules remote config

vegaro:
Adds the real backend contract for the checkpoints topic

iOS conterpart https://github.com/RevenueCat/purchases-ios/pull/7370

---

> [!NOTE]
> **Medium Risk**
> Renames the remote-config topic wire key and defines how checkpoint→workflow rules are parsed; misaligned backend rollout or bad payloads could drop rules silently until consumers act on them.
> 
> **Overview**
> Implements the backend **checkpoint rules** remote-config contract: `CheckpointResponse` now carries a `rules` list of `CheckpointRule` entries (`audience_id`, `workflow_id`, optional `id`), replacing the empty placeholder model.
> 
> Remote config ingestion switches from the `checkpoints` topic to **`checkpoint_rules`** (`RemoteConfigTopic.CheckpointRules`), and `CheckpointsConfigProvider` loads typed `CheckpointResponse` blobs via `blobData<CheckpointResponse>`.
> 
> **`CheckpointRulesSerializer`** decodes each rule independently—invalid array elements, missing/empty required fields, or wrong types are logged and skipped so valid rules keep their order. Unknown JSON fields and malformed top-level `rules` shapes are handled without failing the whole checkpoint when the blob is otherwise valid.
> 
> Tests cover happy path, malformed-rule isolation, empty rules, and null responses for bad JSON or non-object payloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01f286cd22916e523d05334ae2f3003f41f16d7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,112 @@
...
+@Serializable
+private data class CheckpointPayload(
+    val id: String? = null,
+    val rules: JsonElement? = null,
+)
+
+@Serializable
+internal data class CheckpointRule(
+    val id: String? = null,
+    @SerialName("audience")
```
tonidero: Hmm this seems a bit strange? As in have `audience` and `workflow_id`. Shouldn't this be `audience_id`? Might be more of a backend question

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,112 @@
...
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.decodeFromJsonElement
+import java.util.Date
+import kotlinx.serialization.SerializationException as JsonSerializationException
+
+internal data class CheckpointResponse(
+    val identifier: String,
+    val id: String?,
```
tonidero: I'm wondering, is the `id` meant only for analytics? Just wondering if we could just send the `identifier` instead... But well, not a blocker :)

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,112 @@
...
+    val rules: List<CheckpointRule>,
+) {
+    companion object {
+        internal fun parse(identifier: String, bytes: ByteArray): CheckpointResponse? {
```
tonidero: I wonder if we could create custom serializer/deserializers instead of all this... Then, we can use the existing `blobData` mechanism to transform from the bytes to the parsed object?

tonidero (commented):
Mostly some questions 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
vegaro: in any case, `id` and `identifier` are not great. We can probably remove `id` for now in the SDK if we won't use it

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
vegaro: done in [a1d8fa0](https://github.com/RevenueCat/purchases-android/pull/3907/commits/a1d8fa0c201e23c731ee507da90980edb7ae7fab)

vegaro: actually I think I can clean it up a bit more and remove the need for CheckpointResponse and CheckpointPayload just for the identifier

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,104 @@
...
- */
 @Serializable
-internal class CheckpointResponse
+internal data class CheckpointRule(
+    val id: String? = null,
+    @SerialName("audience")
+    val audienceId: String,
+    @SerialName("workflow_id")
+    val workflowId: String,
+    val schedule: CheckpointRuleSchedule? = null,
```
facumenzella: we ended up dropping schedule on iOS since it's not V1 (same reasoning as frequency_cap).

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
vegaro: d42f19240

facumenzella (approved):
lgtm 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,83 @@
...
+    val identifier: String = "",
+    val id: String? = null,
+    @Serializable(with = CheckpointRulesSerializer::class)
+    val rules: List<CheckpointRule> = emptyList(),
+)
+
+@Serializable
+internal data class CheckpointRule(
+    val id: String? = null,
+    @SerialName("audience")
```
facumenzella: `audience_id` here too? https://github.com/RevenueCat/purchases-ios/pull/7370#discussion_r3734538566

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,83 @@
...
 
-/**
- * Empty until the backend checkpoint schema is defined.
- */
 @Serializable
-internal class CheckpointResponse
+internal data class CheckpointResponse(
+    @Transient
+    val identifier: String = "",
+    val id: String? = null,
```
facumenzella: still here, dropped it on iOS https://github.com/RevenueCat/purchases-ios/pull/7370#discussion_r3734544258

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
vegaro: lol yes lost track of life

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
vegaro: updated to `audience_id`

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
```diff
@@ -1,9 +1,82 @@
...
+    val rules: List<CheckpointRule> = emptyList(),
+)
+
+@Serializable
+internal data class CheckpointRule(
+    val id: String? = null,
+    @SerialName("audience_id")
+    val audienceId: String,
+    @SerialName("workflow_id")
+    val workflowId: String,
```
tonidero, line 30: So I believe it's true but just to confirm, this will be the id of the topic item, that we will use in the `WorkflowsConfigProvider.getWorkflow` right? If so it should be easy to link together 🙏

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointsConfigProvider.kt
```diff
@@ -7,5 +7,6 @@ internal class CheckpointsConfigProvider(
     private val manager: RemoteConfigManager,
 ) {
     suspend fun getCheckpoint(identifier: String): CheckpointResponse? =
-        manager.blobData(RemoteConfigTopic.Checkpoints, identifier)
+        manager.blobData<CheckpointResponse>(RemoteConfigTopic.Checkpoints, identifier)
+            ?.copy(identifier = identifier)
```
tonidero: Hmm I'm not sure I'm liking it... Is there a need to embed the `identifier` in the `CheckpointResponse` like this? I guess the caller already knows the `identifier` it's using so there is no need to add it?

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigTopic.kt
```diff
@@ -12,5 +12,5 @@ internal enum class RemoteConfigTopic(val wireName: String) {
     Workflows("workflows"),
     UiConfig("ui_config"),
     Sources("sources"),
-    Checkpoints("checkpoints"),
+    Checkpoints("checkpoint_rules"),
```
tonidero: Nitpick... but should we also rename the enum to `CheckpointRules` for consistency? I don't think we need to add other checkpoint-related topics at the moment, and this is internal anyway so not a blocker.

tonidero (approved):
Looks good!

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointsConfigProvider.kt
vegaro: right. Actually what you say makes sense, no need at all to have the identifier in the `CheckpointResponse`. I guess the only reason would be to diffeenciate it from another Checkpoint, so the `equals` returns false when comparing wiht another one that has the same rules? Am I overthinking? lol

tonidero: Well… This model is internal, so I would say it's not an issue to try to improve its "equals" 😅. I guess callers will need to understant this case, but in any case, I imagine this needs to be keyed by the checkpoint identifier anyway… I would say to remove it until we need it 🙂

purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigTopic.kt
vegaro: [c4e7367](https://github.com/RevenueCat/purchases-android/pull/3907/commits/c4e7367319736ced786a632d0a32d29f76f74be1)

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointResponse.kt
vegaro: yes, it's the public id, the one that starts with `wf...`

purchases/src/main/kotlin/com/revenuecat/purchases/common/checkpoints/CheckpointsConfigProvider.kt
vegaro: [01f286c](https://github.com/RevenueCat/purchases-android/pull/3907/commits/01f286cd22916e523d05334ae2f3003f41f16d7a)

PR #3909, opened 2026-08-06
refactor(checkpoints): Move the checkpoints API to the RevenueCatUI module

tonidero:
Checkpoints currently live entirely in the core `purchases` module, even though everything they do today is present a paywall. While we can make it work from the core module using the ServiceLocator pattern and load the paywall presenter (the way we have it now), in iOS this is more complicated.

In order to keep parity, we will move the checkpoints API to `:ui:revenuecatui`. The core module keeps only what it alone can do, resolving a checkpoint identifier to a workflow, and exposes it through a single `@InternalRevenueCatAPI` seam. The ServiceLocator pattern previously used is no longer needed and is removed.

Additionally, this also adds API tests for the APIs, even if they are annotated as Internal, so it helps also visualizing usage.

---

> [!NOTE]
> **Medium Risk**
> Large internal architectural move across modules with presentation lifecycle and one-at-a-time constraints; public stable APIs are unchanged but internal checkpoint call sites must use RevenueCatUI packages and depend on the UI module.
> 
> **Overview**
> Moves the **checkpoints developer API** (`awaitCheckpoint`, `checkpointListener`, result/outcome types) from `purchases` into `revenuecatui`, while the core module keeps only **resolution** (`resolveCheckpoint` → `CheckpointResolution`) plus shared infrastructure (`currentActivity`, an opaque `checkpointManagerSlot` on `Purchases`).
> 
> **Core (`purchases`)** drops orchestration and UI wiring: `CheckpointsManager`, `UiCheckpointWorkflowExecutor`, `CheckpointPresenter` / `ServiceLoader`, and `Purchases.awaitCheckpoint` / `checkpointListener` are removed. Resolution now returns `CheckpointResolution` (workflow vs no-action reasons, including **unknown checkpoint** for the PoC identifier) and throws on hard failures instead of returning a failed resolution type.
> 
> **UI (`revenuecatui`)** owns end-to-end checkpoint hits: a per-`Purchases` `CheckpointsManager` (stored in `checkpointManagerSlot`) calls `resolveCheckpoint`, presents via `CheckpointWorkflowActivity`, and exposes `Purchases.awaitCheckpoint` and `Purchases.checkpointListener` extensions. Pending presentation state lives on the manager instead of `CheckpointCallStore` + core delegate callbacks.
> 
> Adds **api-tester** coverage for the internal checkpoint surface and updates **paywall-tester** imports to the new package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85bcb0828627dca0ac2e4a0d82c100585964ced9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointResult.kt
```diff
@@ -44,6 +44,9 @@ public abstract class CheckpointResult internal constructor() {
 
                 @JvmField
                 public val DISABLED: Reason = Reason("DISABLED")
+
+                @JvmField
+                public val UNKNOWN_CHECKPOINT: Reason = Reason("UNKNOWN_CHECKPOINT")
```
tonidero, line 49: I added this, for checkpoints that do not have any rules/workflow or anything attached cc @rickvdl

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointsManager.kt
```diff
@@ -1,54 +0,0 @@
...
-import com.revenuecat.purchases.common.errorLog
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
-
-/**
- * Orchestrates a checkpoint hit: fires listener events, resolves the checkpoint to a workflow through
- * [CheckpointWorkflowResolver], and runs the resolved workflow through [CheckpointWorkflowExecutor].
- */
-internal class CheckpointsManager(
```
tonidero, line 16: The manager goes away. Instead, the logic mostly will live in the `CheckpointsCoordinator` in the ui module.

purchases/api-defauts.txt
```diff
@@ -385,7 +385,7 @@ package com.revenuecat.purchases {
     method public void getAmazonLWAConsentStatus(com.revenuecat.purchases.interfaces.GetAmazonLWAConsentStatusCallback callback);
     method @kotlin.jvm.Synchronized public String getAppUserID();
     method public com.revenuecat.purchases.virtualcurrencies.VirtualCurrencies? getCachedVirtualCurrencies();
-    method @kotlin.jvm.JvmSynthetic public com.revenuecat.purchases.checkpoints.CheckpointListener? getCheckpointListener();
+    method @kotlin.jvm.JvmSynthetic public Object? getCheckpointListenerSlot();
```
tonidero: Again, this still remains as internally annotated API so people still need to opt in to use it.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointsManager.kt
```diff
@@ -0,0 +1,160 @@
...
+                "Failed to present checkpoint workflow: $e",
+            )
+        }
+    }
+
+    // Counterpart of onActivityDestroyed for calls that fail or get cancelled before the presented workflow
+    // reports: releases the slot and takes the orphaned paywall down with the caller that asked for it.
+    // Always runs on the main thread, inside the checkpoint() dispatch.
+    private fun abandon(callId: String) {
+        take(callId)?.activity?.get()?.finish()
```
tonidero, line 135: I ended up keeping a weak reference to the activity in the cache so we can also cancel the activity if the coroutine gets cancelled for any reason.

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointsManager.kt
rickvdl: It's back to `CheckpointsManager` right?

rickvdl (approved):
Makes sense! 🚢

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointsManager.kt
tonidero: Yes! Sorry outdated comment. I iterated over this and brought it back to match iOS :)

rickvdl: I figured that, great! 🙌

PR #3902, opened 2026-08-06
refactor(paywalls): collect web_view assets in the existing predownload walk

AlvaroBrey:
- Widens the existing paywall component walk so a single BFS collects image URIs **and** `web_view` descriptors
- **Nothing consumes the `web_view` data yet, and image pre-download behavior is unchanged**, so this is a no-op at runtime. It exists to give the next two PRs a single traversal to read from.

### Checklist
- [x] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids

<details><summary>Agent description</summary>

### Motivation

Both warming paths already funnelled through one BFS over the component tree, with a `when` over component type as the single source of truth for extraction. `web_view` was the one type that returned nothing.

Adding WebView warming as its own traversal would have meant a second walk over the same tree on the offerings-load path, and a second place to keep in sync with the component model. Widening the return type of the walk that already exists costs one pass and keeps the `when` exhaustive, so a new component type still fails to compile until someone decides what it contributes.

This mirrors step 1 of the iOS plan, which consolidates image and video traversals into one. Android was already consolidated (video fallbacks are handled inside the same `when`, fonts never come from the tree), so the Android version of that step is just this widening.

### Description

- `PaywallComponentsImagePreDownloader.findImageUrisToDownload()` moves out to a top-level `PaywallComponentsConfig.collectAssets()` in a new `PaywallComponentAssets.kt`. The extraction `when` is carried over unchanged.
- Returns `PaywallComponentAssets(imageUris, webViews)`. `preDownloadImages` now takes the collected assets rather than the config, so one traversal feeds every warming path.
- `WebViewAsset` derives the Fit axes from the base component's `size`. Overrides can only change `visible`, so the base component is the right source; this matches what `WebViewComponentView` derives at render time.
- Adds `PaywallComponent.flatten()` next to the existing `filter`, so the walk reads as a flatten instead of `filter { true }`.
- Covers the walk with `PaywallComponentAssetsTest`: nesting under header and sticky footer, per-axis Fit derivation, two components sharing a URL, and images plus `web_view` collected in the same pass.

### Known limitations

- `OfferingVideoPredownloader` still does its own `filter { it is VideoComponent }` walk, over `base.stack` only rather than header and sticky footer. It has no production caller today, so it is not a second walk on any live path, but it should fold into `collectAssets()` whenever it gets wired.

</details>

---

> [!NOTE]
> **Low Risk**
[...]

PR #3910, opened 2026-08-06
fix(paywalls): prewarm the current offering's workflow assets on load

AlvaroBrey:
TL;DR: for workflows paywalls, we were prefetching the data but not prewarming images/fonts/etc until we actually launched the paywall. This is a bug, since it differs from iOS and from the offerings paywalls.

- The prewarm is announced at the end of `WorkflowsConfigProvider.warm`, but `warm` runs on config commit, which on a cold start finishes before the offerings response lands. `currentOfferingIdProvider` reads the in-memory offerings cache, so it is null and nothing is announced.
- Then `isWarmForCurrentOffering` is true because the _data_ is there, so the `warm` guard and the `onPaywallConfigReady` fast path short-circuit every later attempt: the announce is never reached again, on cold or warm launch.
- Split the announce out as `prewarmCurrentOfferingAssets` and drive it from `onPaywallConfigReady`, which `OfferingsManager` calls right after caching the offerings: the first point where the current offering is known.
- **This shifts asset downloads earlier for every workflows app.** Images and fonts that used to be fetched at paywall-open are now fetched shortly after offerings resolve. That is the intended design and matches both the non-workflows path and iOS, but it is a runtime behavior change.

### Checklist

### Motivation

Asset prewarming exists so a paywall does not fetch its own images and fonts while rendering. On projects using workflows that never happened: nothing was warmed until a paywall was opened, at which point the render path warmed everything at once. The first screen the user sees was always uncached.

The cause is an ordering race with no recovery. `WorkflowsConfigProvider.warm` byte-caches eligible workflows and then announces the current offering's workflow for asset prewarming. It is driven by `onConfigCommitted`, and on a cold start config commit wins: measured on device, config persisted at `:52.262` and warmed at `:52.291`, while the offerings response only landed at `:52.453`. `currentOfferingIdProvider` reads `offeringsCache.cachedOfferings`, which is in-memory only and therefore empty at process start, so the announce is skipped.

Nothing retries, because the warmth guards measure the wrong thing. `isWarmForCurrentOffering` reports whether the workflow's *bytes* are cached; it says nothing about whether its assets were prewarmed. Those bytes are warm after that first pass, so `warm`'s own guard and `WorkflowManager.onPaywallConfigReady`'s fast path both return before reaching the announce. `warm` ran exactly once in a 75 second session. A warm relaunch is no better: `/v1/config` returns 304, so there is no commit and no warm at all.

iOS does not have this problem because it pushes the offering id in rather than reading it from a cache: `OfferingsManager` passes `includingOfferingId: offerings.current?.identifier` into `scheduleAssetPrewarmingForPrefetchedWorkflows`, so there is no race to lose. It also tracks asset-prewarm state separately from byte state, which is the distinction the Android guard collapses.

### Description

- Extract the announce from `WorkflowsConfigProvider.warm` into `prewarmCurrentOfferingAssets`, which resolves the offering to workflow mapping from the cache rather than requiring a fresh `warm` pass. `warm` calls the same private helper, so its behavior is unchanged.
- Call it from `WorkflowManager.onPaywallConfigReady`, before the warm fast path. `OfferingsManager` already calls that right after `cacheOfferings`, so no change to `OfferingsManager` was needed.
- `WorkflowAssetPrewarmer` already dedups by workflow id, so the already-warmed case costs nothing and the render path cannot double-download.

Measured on a Pixel 7a, cold launch with no paywall opened, on a workflows project whose current offering has a V2 paywall:

| | before | after |
| --- | --- | --- |
| workflow decodes | 0 | 1 |
| image pre-downloads | 0 | 7 |

Ordering after the fix: config warm `:24.790` (no current offering yet), offerings cached `:25.208`, prewarm decode `:25.333`, images from `:25.334`.

The regression gate is `WorkflowManagerTest > onPaywallConfigReady prewarms the current offering's assets even on the warm fast path`: it fails with the `prewarmCurrentOfferingAssets` call removed and passes with it. `WorkflowsConfigProviderTest > prewarmCurrentOfferingAssets announces once the offering is known after a warm that had none` covers the provider half, reproducing the cold-start ordering directly.

Known limitation: only the current offering's assets are warmed, matching the existing offerings path. iOS additionally warms all prefetch-flagged workflows; Android still only byte-caches those. Not changed here.

</details>

---

> [!NOTE]
> **Medium Risk**
> Shifts network work earlier for all workflows apps after offerings load; behavior is intentional but increases background downloads and touches the offerings readiness path.
> 
> **Overview**
> Fixes a cold-start race where workflow **byte** caches looked warm but **images/fonts** for the current offering were never pre-downloaded until paywall open.
> 
> On cold launch, `WorkflowsConfigProvider.warm()` often runs from config commit **before** offerings are cached, so `currentOfferingId` is null and asset prewarm is skipped. Later, `isWarmForCurrentOffering()` and `onPaywallConfigReady`'s fast path never retry the announce.
> 
> **Changes:** Asset announce logic is centralized in `announceCurrentWorkflow`, exposed as `prewarmCurrentOfferingAssets()` for callers that learn the current offering later. `WorkflowManager.onPaywallConfigReady` invokes it **before** the warm fast path (after `OfferingsManager` caches offerings). `warm()` still calls the same helper; `WorkflowAssetPrewarmer` dedupes by workflow id.
> 
> **Runtime effect:** Workflows apps now fetch current-offering paywall assets shortly after offerings resolve (aligned with offerings paywalls and iOS), not only at paywall launch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 026d3cbf1b16731c26499913965b45f7f6082d31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
@@ -138,6 +138,9 @@ internal class WorkflowManager(
      * cache next reads committed data.
      */
     fun onPaywallConfigReady(onComplete: () -> Unit) {
+        // warm() may already have run with no current offering (config commits before offerings land), and the
+        // fast path below would then never re-run it. Deduped by workflow id, so a repeat call is free.
+        workflowsConfigProvider.prewarmCurrentOfferingAssets()
         if (uiConfigProvider.isWarm() && workflowsConfigProvider.isWarmForCurrentOffering()) {
             onComplete()
             return
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowsConfigProvider.kt
@@ -250,21 +250,30 @@ internal class WorkflowsConfigProvider(
         }
         cache.store(generation, Cached(workflows, offeringToWorkflowId))
 
-        // Warm the current offering's workflow assets (images + ui_config fonts) at load time — mirrors the
-        // offerings path, which pre-downloads only the current offering's assets, never other offerings'. So a
-        // prefetch-flagged workflow that isn't the current offering's has its bytes cached (above) but not its
-        // assets: it isn't the paywall about to be shown. Fire-and-forget so it never blocks warm() or the
-        // getOfferings readiness gate. resolveWorkflowBody decodes transiently: it reads bytes + parses without
-        // touching the retained Lazy above, so prewarming keeps the cache raw-bytes-only.
-        onCurrentWorkflowLoaded?.let { notify ->
-            // Notify whenever the current offering maps to a workflow — do NOT gate on `workflows` (the byte-warm
-            // map), whose parallel preload can miss a body that hasn't finished its LOW-priority prefetch yet.
-            // resolveWorkflowBody fetches the body on demand, so gating here would drop the current offering's
-            // asset prewarm purely on preload timing.
-            currentOfferingId?.let { offeringToWorkflowId[it] }?.let { currentWorkflowId ->
-                scope.launch { notify(currentWorkflowId, ::resolveWorkflowBody) }
-            }
-        }
+        announceCurrentWorkflow(offeringToWorkflowId)
+    }
+
+    /** Announces against the already-warmed cache, for callers that learn the current offering after [warm]. */
+    fun prewarmCurrentOfferingAssets() {
+        announceCurrentWorkflow(cache.cached?.offeringToWorkflowId)
+    }
+
+    // Warm the current offering's workflow assets (images + ui_config fonts) at load time — mirrors the
+    // offerings path, which pre-downloads only the current offering's assets, never other offerings'. So a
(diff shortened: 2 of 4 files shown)


AlvaroBrey:
* **#3897** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3897?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3913** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3913?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3910** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

AlvaroBrey:
@cursor review

JZDesign (approved):
I think that this is an appropriate change, but I missed a lot of the context that occurred when these new cache mechanisms were introduced. So I would still wait on another reviewer to be certain.

tonidero (approved):
Ohhh great find!! 🫶 I think this makes sense!

vegaro (approved):

PR #3912, opened 2026-08-07
Unify projects used by Maestro tests

vegaro:
### Checklist

### Motivation

The annual purchase and workflow Maestro tests use two different RevenueCat projects. Keeping the `automated_sdk_tests` project gives us one place for the shared test fixtures, one API key, and one CI job.

### Description

- Use `MAESTRO_TEST_STORE_API_KEY` in the unified Maestro job.
- Fetch `MAESTRO_TEST_STORE_API_KEY` from 1Password during local setup.
- Pass the key through `BuildConfig.API_KEY` instead of editing `Constants.kt` before the build.
- Configure Purchases from the first Activity so Maestro launch arguments are available before configuration.
- Run the annual purchase and workflow suites against the same APK and emulator.
- Select the dedicated `maestro_annual_purchase` offering and assert its copied paywall, annual product, and `pro` entitlement.
- Document the shared local E2E properties.

### Testing

- `circleci config validate .circleci/config.yml`
- `./gradlew :test-apps:e2etests:lintDebug :test-apps:e2etests:assembleDebug --no-daemon`
- `maestro test maestro/e2e_tests/ maestro/workflows/` (13/13 flows passed)

---

> [!NOTE]
> **Low Risk**
> Changes are limited to E2E test app, Maestro flows, and CI/local dev wiring; no production SDK behavior.
> 
> **Overview**
> Maestro **annual purchase** and **workflow** suites now run in one CircleCI job (`run-maestro-tests`) against a single E2E APK and emulator, using the `automated_sdk_tests` test-store project instead of two separate RevenueCat projects and CI jobs.
> 
> The API key is injected at build time via `MAESTRO_TEST_STORE_API_KEY` (`BuildConfig.API_KEY`), replacing CI `sed` on `Constants.kt` and the separate `E2E_WORKFLOWS_API_KEY` / workflow-only build. Local setup pulls the key from 1Password (`setup_local_properties`) and documents it in `local.properties.example`.
> 
> `E2ETestsApplication` always configures Purchases on the first Activity so Maestro launch arguments (locale, debug failure strategy) apply before the SDK starts for all flows. The annual purchase Maestro flow targets offering `maestro_annual_purchase` with updated product/entitlement assertions (`pro_annual_subscription`, `pro`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42a5252d74042aca658e7a62c8a701389b670459. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.circleci/config.yml
```diff
@@ -803,26 +803,21 @@ jobs:
...
+    description: "Run Maestro E2E tests for E2E Tester app"
     <<: *android-machine-emulator
     steps:
       - checkout
       - attach_workspace:
           at: .
       - revenuecat/install-mise-tools:
           tools: java
-      - run:
-          name: Replace API_KEY (Using Test store API key from Production Integration Tests Android project)
```
tonidero, line 823: Hmm I think it's good to add the name of the project somewhere since it can help people looking at this... Maybe we can add this somehow below? Or maybe a comment there is enough...

.circleci/config.yml
```diff
@@ -1073,6 +1066,8 @@ workflows:
           requires:
             - prepare-tests
       - run-maestro-e2e-tests:
+          context:
+            - maestro
           requires:
             - prepare-tests
       - run-maestro-workflow-tests:
```
tonidero: Hmm do we need to remove this (and its job)?

tonidero: Hmm I think the difference is the `E2E_CONFIGURE_PURCHASES_ON_FIRST_ACTIVITY` property? Can this be passed as a parameter from the maestro test itself when launching the app instead so we can unify the jobs?

tonidero (commented):
Mostly questions to see if we can unify both jobs into one.

.circleci/config.yml
vegaro: Yes, we can unify the jobs. `Application.onCreate` runs before the Activity intent extras containing the Maestro parameters are available, so it can’t use those parameters directly. What I can do is that I can always register the existing Activity lifecycle callback and configure Purchases in `onActivityPreCreated`, where the extras are available:

```
Application.onCreate
  → register activity lifecycle callback

onActivityPreCreated
  → read Maestro extras
  → configure Purchases
  → unregister callback
```

The regular purchase flow provides no extras and gets the default configuration, while workflow flows can customize the initial configuration as they already do. I am going to try that approach

.circleci/config.yml
vegaro: Added `automated_sdk_tests` next to the CircleCI build command and in `local.properties.example`.

fastlane/Fastfile
```diff
@@ -61,6 +61,7 @@ platform :android do
     fields = %w[
       PAYWALL_TESTER_API_KEY_A
       PURCHASE_TESTER_API_KEY
+      MAESTRO_TEST_STORE_API_KEY
```
tonidero, line 64: I guess the 1password item has already been added :)

tonidero (approved):
Nice! Thanks for the cleanup! 💪

fastlane/Fastfile
vegaro: Yes, already added. Thanks for pointing it out!

PR #3913, opened 2026-08-07
refactor(paywalls): move image warming to `purchases-ui` and use a ServiceLoader to reach it

AlvaroBrey:
- Replaces by-name reflection with `ServiceLoader` pattern already used for `PurchasesService` (and briefly for checkpoints). 
- Image pre-downloading and the shared `ImageLoader` move to RevenueCatUI, where Coil is. Core stops compiling against a library it does not ship.

### Checklist

### Motivation

`:purchases` cannot depend on `:ui:revenuecatui`, but core is where warming is triggered. Four mechanisms had accumulated for one question, "is RevenueCatUI here, and can it do this for me?":

| Mechanism | Used by | Problem |
|---|---|---|
| `ServiceLoader` + typed interface | `PurchasesService`, registered by `feature/admob` | None. This is the good one. |
| `compileOnly` + `Class.forName` probe | images (Coil) | Core compiles against a library it does not ship, forcing `getImageLoader` to return `Any`. |
| `Class.forName` on a string FQN | WebView preboot, as first written in #3904 | Nothing binds the string at build time. |
| n/a | fonts, videos | None needed: both work with core's own HTTP stack. |

Only images and preboot need something core cannot reach. This PR moves images; #3904 moves preboot.

#3909 removed a core → UI seam entirely by moving checkpoints into RevenueCatUI. That is the better move where it works, and it is why `CheckpointPresenter` is no longer an example of this pattern. It does not work here: warming is triggered by the offerings load, which lives in core, and `PurchasesService` only gets `initialize`/`close`.

### Description

- `PaywallAssetWarmer` in core (`warmImages(context, uris)`), plus an internal `PaywallAssetWarming` holder that loads it lazily, so the `ServiceLoader` scan happens on first warm rather than during `configure()`.
- Every call through the seam is wrapped in `runCatching`. Warming runs inline on the SDK's single-threaded dispatcher inside the offerings success callback, before offerings are cached and handed to the app, so a misbehaving implementation must not take that path down.
- `CoilImageDownloader` and `PaywallComponentsImagePreDownloader` are deleted; the `ImageLoader` factory moves to `PaywallImageLoader` in RevenueCatUI, which `RemoteImage` and `BackgroundStyle` call directly.
- The tree walk is gated on warmer presence, since `collectAssets()` has no consumer other than warming.
- `OfferingVideoPredownloader` now takes the shared `PaywallAssetWarming` as a required parameter rather than defaulting to a new one, so it cannot start a second `ServiceLoader` scan. It has had no production caller since #2717 unwired it, so only its own test constructs it.

### Things the diff does not show

- **R8 verified end to end.** In a minified consumer APK the descriptor survives unrenamed and still names the impl in full: `META-INF/services/...PaywallAssetWarmer -> ...PaywallAssetWarmerImpl`. The AAR carries both the keep rule and the descriptor. `purchases`' own consumer rules already keep `com.revenuecat.**`, so the RevenueCatUI rule is belt-and-braces.
- **Regression gate: `PaywallAssetWarmerImplTest`.** It does a real `ServiceLoader.load` and asserts the impl comes back, so deleting the descriptor or renaming the class fails the build. Verified by pointing the descriptor at a renamed class and watching it fail. The reflection approach it replaces could not have an equivalent.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes the cross-module seam for paywall image pre-download and removes a public internal API; warming runs on the offerings path but is guarded with runCatching and lazy SPI discovery limits configure-time risk.
> 
> **Overview**
> **Core no longer compiles against Coil or RevenueCatUI for paywall images.** Offerings and workflow pre-download now go through a **`PaywallAssetWarmer`** SPI discovered with **`ServiceLoader`** (same pattern as **`PurchasesService`**). Internal **`PaywallAssetWarming`** loads the warmer lazily, gates tree walks on **`isAvailable`**, and wraps every warm in **`runCatching`** so failures cannot break the offerings success path.
> 
> **Removed from core:** **`CoilImageDownloader`**, **`PaywallComponentsImagePreDownloader`**, **`Purchases.getImageLoader`**, **`canUsePaywallUI`**, and the **`compileOnly`** Coil dependency. **`OfferingImagePreDownloader`**, **`WorkflowAssetPrewarmer`**, and **`OfferingVideoPredownloader`** now share one **`PaywallAssetWarming`** instance from **`PurchasesFactory`**.
> 
> **In RevenueCatUI:** **`PaywallImageLoader`** holds the shared Coil **`ImageLoader`** (25 MB disk cache); **`PaywallAssetWarmerImpl`** enqueues **`ImageRequest`s** and is registered via **`META-INF/services`**. **`RemoteImage`** and **`BackgroundStyle`** use **`PaywallImageLoader`** instead of **`Purchases.getImageLoader`**. ProGuard keeps the impl’s no-arg constructor for **`ServiceLoader`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aecc5ea7dd897b5ca0c9619ba3886c7e610c5285. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey:
* **#3897** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3897?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#3918](https://github.com/RevenueCat/purchases-android/pull/3918) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3918?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#3938](https://github.com/RevenueCat/purchases-android/pull/3938) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3938?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3913** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3916** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

AlvaroBrey:
@cursor review

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/paywalls/PaywallImageLoader.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * The single [ImageLoader] every paywall image goes through, so what warming pre-downloads is what
+ * rendering reads back.
+ */
+internal object PaywallImageLoader {
+
+    private const val MAX_CACHE_SIZE_BYTES = 25L * 1024 * 1024
+    private const val CACHE_FOLDER = "revenuecatui_cache"
+
+    private var cached: ImageLoader? = null
```
tonidero, line 12: I guess this won't be reset even if the purchases instance is destroyed... I think this is fine but something to be aware of.

tonidero (approved):
Looks great!! Indeed much cleaner to do it this way I think

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/paywalls/PaywallImageLoader.kt
AlvaroBrey: It was already like this in `PurchasesOrchestrator` tbh.

But also, I've looked at Coil docs and this is not a big deal. `ImageLoader` keeps app context, not activity, so there's no leaks. And clearing _could_ be worse, because we could end up with more than one instance writing to the same disk cache

PR #3917, opened 2026-08-07
chore: stop persisting ProductEntitlementMapping.originalSource to disk

AlvaroBrey:
- Stops persisting `ProductEntitlementMapping.originalSource` to disk. Only used by tests, who still read it from memory.
- Fresh network fetches still set it in-memory via `fromNetwork()`. disk-cache reads now report `null` as source
- Simplified `fromJson()`/`fromNetwork()` to drop the JSON round-trip the field used to need, now that nothing persists or reads it back from disk.
- Partial fix for [SDK-4408](https://linear.app/revenuecat/issue/SDK-4408/stop-persisting-originalsource-three-android-caches-write-a-test-only), issue stays open.

### Checklist

### Motivation

`ProductEntitlementMapping.originalSource` was added to support backend integration tests asserting failover behavior. It has no runtime consumer. Unlike `Offerings`/`CustomerInfo`, this class has no custom `equals()`, so the field was participating in ordinary data class equality.

Investigated together with the analogous fields on `Offerings` (already fixed in #3872, was forcing a costly JSON reserialize on every cache write) and `CustomerInfo` (left as-is: three existing androidTest assertions genuinely depend on it surviving a cache read, and there's no OOM-shaped driver to justify building an in-memory shim just for this).

### Description

- Removed the `rc_original_source` write from `ProductEntitlementMapping.toJson()`.
- `fromNetwork()` now attaches `originalSource` directly via `.copy()` instead of injecting it into a JSON copy just to re-parse it back out through `fromJson()`.
- `fromJson()` no longer reads `rc_original_source` at all. A legacy cache file from an older SDK version that still has the key is now ignored.
- Made `originalSource` nullable, defaulting to `null` instead of `MAIN`: a disk-loaded mapping no longer silently claims a source it doesn't actually know.
- Added a regression test proving the field isn't persisted (`DeviceCacheTest`), and updated the existing tests that encoded the old default-to-`MAIN` behavior.

</details>

---

> [!NOTE]
> **Low Risk**
> Test-only metadata on an internal API with no runtime consumers; entitlement mapping data itself is unchanged.
> 
> **Overview**
> Stops writing **`ProductEntitlementMapping.originalSource`** to SharedPreferences by removing `rc_original_source` from **`toJson()`** and from **`fromJson()`** parsing. Network responses still set the field in memory via **`fromNetwork()`** → **`copy(originalSource = httpResult.originalDataSource)`**, without the previous JSON inject-and-reparse path.
> 
> **`originalSource`** is now **nullable** and defaults to **`null`** instead of **`MAIN`**, so cache reads no longer imply a source the mapping never stored. Legacy cache JSON that still contains **`rc_original_source`** is ignored on load.
> 
> Tests and **`createProductEntitlementMapping`** were updated; **`DeviceCacheTest`** adds a regression that cached JSON omits the key and round-trip **`originalSource`** is null.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57de5806b10b3b989cae1e2d53c35f314a94172f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMapping.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMapping.kt
@@ -12,7 +12,7 @@ import org.json.JSONObject
 @InternalRevenueCatAPI
 public data class ProductEntitlementMapping(
     val mappings: Map<String, Mapping>,
-    val originalSource: HTTPResponseOriginalSource = HTTPResponseOriginalSource.MAIN,
+    val originalSource: HTTPResponseOriginalSource? = null,
     val loadedFromCache: Boolean = false,
 ) {
     public companion object {
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMapping.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/offlineentitlements/ProductEntitlementMapping.kt
@@ -2,10 +2,8 @@ package com.revenuecat.purchases.common.offlineentitlements
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
 import com.revenuecat.purchases.common.HTTPResponseOriginalSource
-import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.networking.HTTPResult
 import com.revenuecat.purchases.common.originalDataSource
-import com.revenuecat.purchases.utils.copy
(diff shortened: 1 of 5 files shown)


tonidero (approved):
:shipit:

tonidero:
It's true that PEM is much smaller than offerings and so, it shouldn't be a big issue to have the round trip... But as you said, only used in tests... I guess we could have a test that verifies reconfiguration, and fetching the PEM from disk and the source... but don't think it provides much value, so this makes sense!

PR #3916, opened 2026-08-07
fix: Declare kotlinx-coroutines-android as a dependency

AlvaroBrey:
- `:purchases` uses `Dispatchers.Main` in two places but only declared `kotlinx-coroutines-core`, which does not provide it. `Dispatchers.Main` throws `IllegalStateException("Module with the Main dispatcher is missing")` without the `kotlinx-coroutines-android` artifact.
- Affected call sites: `BlockstoreHelper` (a constructor-default `CoroutineScope`) and `RewardVerificationPollLauncher` (`withContext(Dispatchers.Main)` to deliver the poll result).
- **Almost certainly unreachable in practice.** Nearly every app already brings `kotlinx-coroutines-android` transitively through `lifecycle-runtime-ktx`, `activity-compose` or similar, which is why this has gone unnoticed. An app using the SDK with no other coroutines-consuming AndroidX library would hit it.
- Adds one line to `libs.versions.toml` and one dependency. Expect a small size delta on an app that did not already have the artifact.

### Checklist
- [x] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids

<details><summary>Agent description</summary>

### Motivation

Found while removing Coil from `:purchases` in a stacked PR. `coil-base` was declared `testImplementation`, and it pulls `kotlinx-coroutines-android` transitively, so the *test* classpath had `Dispatchers.Main` by accident. Dropping Coil broke two reward-verification tests with the missing-Main-dispatcher error, which is what surfaced the gap.

Checking the production side showed the same hole and no accident covering it: `kotlinx-coroutines-android` was on no RevenueCat module's runtime classpath, verified against `:purchases`, `:feature:admob` and `:ui:revenuecatui`. The SDK was relying on the consuming app to supply a dispatcher it uses itself.

### Description

- `implementation(libs.coroutines.android)` in `purchases/build.gradle.kts`, plus the version-catalog alias. Same version reference as `coroutines-core`, so the two cannot drift.

### On the missing test

There is no unit test here, deliberately. The failure mode is a missing artifact on the *published* runtime classpath, not a code path: on this branch the test classpath still gets `kotlinx-coroutines-android` transitively through `coil-base`, so any `Dispatchers.Main` test passes with or without this change and would gate nothing.

The evidence is the dependency graph. Before, `:purchases:dependencies --configuration defaultsReleaseRuntimeClasspath` had no `kotlinx-coroutines-android`; now it does.

</details>

---

> [!NOTE]
[...]

PR #3918, opened 2026-08-07
feat(paywalls): Pool prerendered web_views by url

AlvaroBrey:
- Fixes a real bug in the single-slot design: a paywall with two `web_view` components had whichever composed first destroy the other's prerender. Both Cursor and an internal review flagged it.
- Two changes. The match key narrows from `WebViewIdentity` (url + component id + Fit axes) to the **resolved url alone**, matching iOS; and the single slot becomes a **bounded pool of 3**, keyed on that url.
- Component id and the Fit axes stop being part of identity and become late-bound bridge state, rebound just before activation triggers the handshake that transmits them. So a view warmed for one component now serves any component pointing at the same bundle.
- **The pool is not exercised yet.** `prewarm` is still reached only from the paywall tester; nothing in core calls it, because prerender is not wired to the `PaywallAssetWarmer` seam. This lands the mechanism so the wiring PR does not have to change policy at the same time.
- The bound of 3 comes from on-device memory measurement, not a guess. Details below.
- The pool bound now only governs the *activation* tier. Urls beyond it are not dropped and no longer evict a held entry: they fall through to the sequential http-cache warm added in #3897, so coverage no longer depends on `MAX_PREWARMED`.
- Tripling the ceiling makes the release path matter more than it did, so `onTrimMemory` release comes in with #3897 and drops both tiers at any trim level.
- Stacked on #3897.

### Checklist

### Motivation

`PaywallWebViewPrewarmer` held one slot keyed on the full `WebViewIdentity`. Two independent problems followed.

The key was too narrow. Two components sharing a bundle url but differing in component id, or in a Fit axis, could not reuse each other's prerender even though the prerendered content is identical. iOS keys on component url, so this was also a gratuitous divergence.

The slot was too small. A paywall can carry 0..N `web_view` components, and real offerings have been observed with 4 to 6 on a single paywall. With more than one, `take()` on the first component destroyed the entry the second would have used, so the second rendered cold *and* paid for a wasted prerender.

### Description

**Rekeying.** `PrewarmedWebView` holds `resolvedUrl` instead of a `WebViewIdentity`. `componentId`, `sizeToContentWidth` and `sizeToContentHeight` become `var` on `WebViewJavaScriptBridge`, with a new `rebindComponent(...)` that `activateIn` calls before `loadUrl`. This is safe because those three values only ever cross to the page during the `rc-web-components` handshake (`init` and `fit`), which activation re-triggers, so rebinding before the reload is equivalent to having constructed the bridge with them.

**Pooling.** The slot becomes a `LinkedHashMap<String, PrewarmedWebView>` bounded at `MAX_PREWARMED`:

- Nothing is evicted at capacity. A held entry is worth more than a new one, and the overflow url still gets its bundle into the http cache via the second tier, so evicting would trade a certain win for a speculative one.
- A repeat `prewarm` for a url already held, in flight, or queued is a no-op.
- Each entry gets its own expiry `Runnable` rather than one shared timeout, so an entry added later is not cut short by an earlier one's clock.
- An async `PrerenderException` discards only its own entry.
- A `take()` miss leaves the pool untouched. Under the old design a miss destroyed the held view; with siblings that is exactly wrong, since they belong to other components on the same paywall and are very likely about to be taken.

### Why the bound is 3

Measured on device, holding prerenders costs almost nothing after the first one or two, because the cost is dominated by renderer processes and WebView caps those itself:

| Held prerenders | Marginal cost |
|---|---|
| 1st | ~104 MB |
| 2nd | ~104 MB |
| 3rd to 5th | ~1-6 MB each |

The cap was 2 renderer processes on a 5.6 GB device and 1 on a 4 GB device, regardless of how many prerenders were requested. Renderer processes run as `com.google.android.webview:sandboxed_processN`, outside the app's own process, which is worth knowing if anyone re-measures this with `dumpsys meminfo`.

So 3 covers the multi-`web_view` paywalls we have actually seen while staying inside the platform's own ceiling. It is no longer a coverage ceiling either: with the cache tier, a paywall with 6 `web_view` urls gets 3 activations and 3 cache hits rather than 3 activations and 3 cold loads.

### Known limitation: a replayed size can come from another component

Under the old identity key, warm and adopter agreed on the Fit axes by construction, so replaying the cached content size was always sound. Keying on url alone drops that guarantee: resize reporting is gated on the *warmed* component's Fit axes, and `ignoreDocumentResetFromActivation` deliberately keeps that size across activation. Reachable damage is limited to an adopter whose `Fill` axis turns out unbounded at measure time, which then consumes a size measured under another component's contract. Masking the replay to the adopter's own Fit axes is the fix; not done here.

### Still undecided

The pool *size* is settled by the measurement above; the pool *policy* is not. What to prerender and when is still an open cross-platform question with iOS, which is building the same url-keyed pool starting at one view. Nothing here commits us: `MAX_PREWARMED` is one constant, and the wiring PR chooses which urls get offered.

Also unsettled: what the cache tier is worth today is capped by CDN config, not by this code. The bundle HTML ships no `Cache-Control` and the SDK script ships `max-age=60, must-revalidate`, so a warm outside the freshness window saves the body but not the round trip. Measured headers and the CDN-side fix are in #3897.

### Tests

`PaywallWebViewPrewarmerTest` gates this change with `holds one entry per distinct url`, `warms urls beyond the pool for the cache instead of evicting`, `never holds more than the pool bound plus one cache warm`, `each entry expires on its own schedule`, `an async error only drops the entry it belongs to`, and `a miss leaves the other entries held`.

Two pre-existing tests asserted the behaviour this deliberately inverts (a view not being handed to a different component, or to different Fit axes). They were rewritten to assert the new behaviour, and `does not hand the view to a different url` was added so the negative case is still covered.

</details>

### DIFF
--- a/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/webview/PaywallWebViewPrewarmer.kt
+++ b/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/webview/PaywallWebViewPrewarmer.kt
@@ -50,17 +50,14 @@ internal class PaywallWebViewPrewarmer {
 
     private val mainHandler = Handler(Looper.getMainLooper())
 
-    private var slot: PrewarmedWebView? = null
+    private val pool = LinkedHashMap<String, PrewarmedWebView>()
 
     // A prerender that is never displayed holds a renderer process at foreground priority, which
-    // Android will not reclaim on its own. Bound it.
-    private val releaseOnTimeout = Runnable {
-        Logger.d("Paywalls V2 web_view prewarm expired before display; releasing.")
-        releaseSlot()
-    }
+    // Android will not reclaim on its own. Each entry expires on its own schedule.
+    private val expiries = mutableMapOf<String, Runnable>()
 
-    // Urls the hold tier had no room for, warmed for the http cache alone. One at a time: WebView caps
-    // its own renderer processes low, so concurrent loads contend instead of overlapping.
+    // Urls the pool had no room for, warmed for the http cache alone. One at a time: WebView caps its
+    // own renderer processes low, so concurrent loads contend instead of overlapping.
     private val cacheWarmQueue = ArrayDeque<WebViewIdentity>()
     private var cacheWarmInFlight: PrewarmedWebView? = null
     private var abandonCacheWarm: Runnable? = null
@@ -78,7 +75,7 @@ internal class PaywallWebViewPrewarmer {
     @VisibleForTesting
     internal val trimCallbacks = object : ComponentCallbacks2 {
         override fun onTrimMemory(level: Int) {
-            if (slot != null || cacheWarmInFlight != null) {
+            if (pool.isNotEmpty() || cacheWarmInFlight != null) {
                 Logger.d("Paywalls V2 web_view warming abandoned on memory trim (level $level).")
             }
             releaseAll()
@@ -90,12 +87,12 @@ internal class PaywallWebViewPrewarmer {
     }
 
     /**
-     * Warms [url] for the component that will render it. [url] is resolved through [WebViewUrlResolver]
-     * so the resulting identity matches the display path's byte for byte; an unresolvable URL is a no-op.
+     * Prerenders [url] for the component that will render it. [url] is resolved through
+     * [WebViewUrlResolver] so the resulting identity matches the display path's byte for byte;
+     * an unresolvable URL is a no-op.
      *
-     * The first url is held for adoption. Later urls are queued and warmed for the http cache one at a
-     * time, since the single slot can only serve one of them. A url already held, in flight or queued is
-     * a no-op.
+     * Entries are pooled by resolved URL up to [MAX_PREWARMED]. Beyond that the url is queued and warmed
+     * for the http cache alone, one at a time. A url already held, in flight or queued is a no-op.
      */
     @MainThread
     @Suppress("ReturnCount")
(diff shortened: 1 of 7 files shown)


AlvaroBrey:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3918?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3918** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3918?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3918?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3897** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3897?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3938](https://github.com/RevenueCat/purchases-android/pull/3938) <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3938?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#3913** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3913?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3916** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3916?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

AlvaroBrey:
Superseded. #3897 is rescoped to cache warming only,

PR #3920, opened 2026-08-07
Fix flaky cached offerings integration test and cover the fully-offline case

AlvaroBrey:
tldr: Increased timeout of tests that _do_ call CDN, and added a true "full offline" one

- `cachedOfferingsAreUsedWhenCachedOfferingsAndServerCannotBeReached` has been failing intermittently with `UncompletedCoroutinesError: After waiting for 10s`, on roughly two thirds of production integration runs.
- It is not a code regression: an unchanged sha fails and then passes on a rerun, and only the production project is affected because the load-shedder ones publish no `ui_config`/`workflows` topics.
- The forced-server-error harness only rewrites requests that go through `HTTPClient`. Config blobs are fetched straight from the config CDN, a different host, so "the server is unreachable" left a real download inside a 10s budget.
- Fix: a 60s budget on the three tests that restart the SDK against the real backend. This does not make the test unfailable, it stops it failing for a legitimate wait.
- New: an opt-in hook for forcing direct connections to fail, and a `cachedOfferingsAreUsedWhenFullyOffline` test that uses it with the blob cache cleared.

### Checklist

### Motivation

The failure is a timeout, not a bad assertion: `runTest` on coroutines-test 1.6.4 gives the body a hard 10s of wall clock, and the second `awaitOfferings()` does not come back in time. Reruns of an identical sha pass, so nothing in the diff of any PR caused it. The earliest occurrence found was 2026-08-06T16:17Z, on four unrelated branches at once.

What the test believes it sets up is "nothing is reachable". What it actually sets up is "the RevenueCat API is unreachable, the config CDN is fine". `ForceServerErrorStrategy` is consulted by `HTTPClient` only, and `RemoteConfigBlobFetcher.tryDownloadVerifyStore` opens its connection through `UrlConnectionFactory` directly, so blob traffic to `config.revenuecat-static.com` was never redirected.

That distinction is not a bug in the SDK. Paywall data moved to a separately hosted config layer precisely so it survives an API outage, so continuing to read it from the CDN while the API is down is the design working. The stale assumption was the test's: since `OfferingsManager` gates its `onSuccess` on `WorkflowManager.onPaywallConfigReady`, which can wait on `awaitTopicAndPrefetchBlobsReady`, the test body has always included a real CDN download whose worst case is the 30s blob timeout.

The production integration project publishes `ui_config` and `workflows` (one workflow, a 23KB blob, plus a 54KB localizations blob), which is why only that job is exposed.

### Description

- Raise the budget to 60s on the three tests that restart the SDK against the real backend. 10s is a deadlock guard, not a latency budget.
- Hoist the duplicated unreachable-URL strategy into one `apiUnreachable` field.
- Add `ForceServerErrorStrategy.shouldForceConnectionFailure(url)`, defaulting to `false` so existing tests keep today's semantics: an API outage does not imply the CDN is down.
- Add `ForcedFailureUrlConnectionFactory`, wired into `RemoteConfigBlobFetcher` from `PurchasesFactory` only when a strategy is present. No public API change.
- Add `cachedOfferingsAreUsedWhenFullyOffline`, which clears the blob store before the restart so the readiness gate has to face a dead network, and keeps the tight 10s default on purpose: with no host reachable there is nothing legitimate to wait for, so the budget doubles as the guard.

`cachedOfferingsAreUsedWhenFullyOffline` is the regression gate. It is not passing vacuously: on device it logs 24 forced connection failures, blob downloads failing across both `revenuecat-static.com` and `.net`, and `Error fetching offerings. Using disk cache.`, all inside 10s. Reverting the `ForcedFailureUrlConnectionFactory` wiring makes it exercise nothing, since the blobs simply re-download.

Measured on a Pixel emulator against the production project, the second `awaitOfferings()` takes 103, 103, 180, 105 and 125ms across five cold runs. The margin over 10s is large locally and thin on a CI emulator, which is what the 60s budget absorbs.

Known limitation: the hook covers remote config blobs only. Fonts, the file repository and source health checks take the same direct path and are unaffected by it, so a test asking for "no network" still leaves those three connecting.

</details>

### DIFF
--- a/purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
+++ b/purchases/src/androidTest/kotlin/com/revenuecat/purchases/BasePurchasesIntegrationTest.kt
@@ -88,6 +88,9 @@ abstract class BasePurchasesIntegrationTest {
         override fun modifyRequestURL(url: URL, endpoint: Endpoint): URL {
             return forceServerErrorsStrategy?.modifyRequestURL(url, endpoint) ?: url
         }
+        override fun shouldForceConnectionFailure(url: String): Boolean {
+            return forceServerErrorsStrategy?.shouldForceConnectionFailure(url) ?: false
+        }
         override fun fakeResponseWithoutPerformingRequest(baseURL: URL, endpoint: Endpoint): HTTPResult? {
             return forceServerErrorsStrategy?.fakeResponseWithoutPerformingRequest(baseURL, endpoint)
         }
--- a/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/integration/cachedofferings/BaseCachedOfferingsUsageIntegrationTest.kt
+++ b/purchases/src/androidTestDefaults/kotlin/com/revenuecat/purchases/integration/cachedofferings/BaseCachedOfferingsUsageIntegrationTest.kt
@@ -21,6 +21,25 @@ import java.net.URL
 
 abstract class BaseCachedOfferingsUsageIntegrationTest : BasePurchasesIntegrationTest() {
 
+    private companion object {
+        const val UNREACHABLE_URL = "http://localhost:100/unreachable-address"
+
+        /**
+         * These tests restart the SDK twice against the real backend, and `getOfferings` only returns once the
+         * `/v1/config` paywall data is ready, which can include downloading config blobs from the config CDN.
+         * `runTest`'s 10s default is a deadlock guard, not a latency budget, and it is too tight for that.
+         */
+        const val BACKEND_TEST_TIMEOUT_MS = 60_000L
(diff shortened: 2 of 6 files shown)


AlvaroBrey:
@cursor review

AlvaroBrey:
@tonidero I'm slightly OOTL on the config endpoint, so let me know if this makes sense. I was suprised that blobs didn't go through HTTPClient

tonidero (approved):
Ohhh thank you so much!! This makes a lot of sense!

Regarding not using httpclient, it felt that for these CDN requests that don't need any of the boilerplate we have created on the httpclient made sense... but yeah we could potentially also join things so everything goes through the httpclient. Just would like to refactor the httpclient into smaller pieces TBH 😅

PR #3924, opened 2026-08-08
fix: re-arm billing reconnection ladder when the service disconnects with pending requests

### LINKED ISSUE #3922: BillingClient reconnection permanently stops when the service disconnects during an in-flight connection attempt; queued requests and getOfferings pend forever

**Describe the bug**

If the Play billing service disconnects while a `BillingClient.startConnection()` attempt is already in flight, the SDK's reconnection ladder is never re-armed and never runs again. From that state:

- The in-flight connection attempt is killed with the service, so `onBillingSetupFinished(OK)` never arrives for it.
- A concurrent `startConnection` call during that window gets response code 5 ("Client is already in the process of connecting to billing service."). `BillingWrapper.onBillingSetupFinished` handles code 5 (`DEVELOPER_ERROR`) with a deliberate no-op ("Billing service is already trying to connect. Don't do anything."), relying on the in-flight attempt to complete. It never does.
- `BillingWrapper.onBillingServiceDisconnected()` only logs and tracks a diagnostics event. It schedules nothing.

Result: no code path ever calls `startConnection` again. The exponential backoff machinery (`RECONNECT_TIMER_START_MILLISECONDS` = 1s with the 15 minute cap in `BillingWrapper`, `RETRY_TIMER_START` = 878ms in `BillingClientUseCase`) never engages because nothing schedules a retry from this state. Queued service requests (`pending_request_count=2` in the SDK's own diagnostics) wait indefinitely, so `Purchases.getOfferings()` pends silently: no error, no timeout, no retry, against a Play Store that is healthy again seconds later. The only release we have observed is an app background/foreground cycle, whose lifecycle hook calls `startConnection` and drains the queue within seconds.

We hit this in production behind a paywall: `getOfferings()` pends indefinitely with no callback, the user sees a spinner forever. The likely production trigger is Play Store churn (for example a Play self-update) while a billing operation is in flight.

1. Environment
   1. Platform: Android. We consume the SDK from a Flutter app in production via purchases_flutter 10.8.0 / purchases-hybrid-common 18.29.0, but the defect is entirely in the native Android SDK and reproduces at the native layer.
   2. SDK version: reproduced identically on purchases-android 10.16.0 and 10.16.1 (BillingClient 8.3.0).
   3. OS version: Android 17 (physical Google Pixel 10a, build CP2A.260705.006) and Android 16 (stock emulator with a Google Play image, no Google account signed in; the wedge is store-account-independent).
   4. Android Studio version: n/a (Gradle CLI builds).
   5. How widespread: low frequency, high severity. Our fleet telemetry shows roughly one silent paywall stall per week in this failure class. Once wedged, the device does not recover until the app is backgrounded and foregrounded.

2. Debug logs

Verbose SDK logs are forwarded through our Flutter log handler; the payloads are the SDK's own lines. Wedge anatomy from the physical Pixel 10a on 10.16.0 (device local time):

```
08-08 08:44:06.980  RC debug: Starting connection for com.android.billingclient.api.BillingClientImpl@3182b3c
08-08 08:44:07.008  ActivityManager: Start proc 10200:com.android.vending ... for bound-service {com.android.vending/com.google.android.finsky.billing.iab.InAppBillingService}
08-08 08:44:07.073  W/BillingClient: Client is already in the process of connecting to billing service.
08-08 08:44:07.074  RC debug: Starting connection for com.android.billingclient.api.BillingClientImpl@3182b3c
08-08 08:44:07.132  ActivityManager: Force stopping service ServiceRecord{... com.android.vending/com.google.android.finsky.billing.iab.InAppBillingService}
08-08 08:44:07.137  W/BillingClient: Billing service died.
08-08 08:44:08.092  RC diagnostics: google_billing_setup_finished {"billing_response_code":5,"billing_debug_message":"Client is already in the process of connecting to billing service.","pending_request_count":2}
08-08 08:44:08.092  RC warn: Billing Service disconnected for com.android.billingclient.api.BillingClientImpl@3182b3c
```

Then nothing. Zero "Starting connection" lines from 08:44:09 through 08:57:08 (grep-proven across the full logcat), even though the force-stop loop ended at ~08:44:56 and the Play Store was healthy from ~08:45:40. The pending `getOfferings` sat silent for 13m03s. At 08:57:09 we backgrounded and foregrounded the app:

```
08-08 08:57:09.450  RC debug: Starting connection for com.android.billingclient.api.BillingClientImpl@3182b3c
08-08 08:57:09.665  RC debug: Billing Service Setup finished for com.android.billingclient.api.BillingClientImpl@3182b3c
```

and the queued work completed about 3 seconds later.

The same anatomy reproduced on the emulator on both 10.16.0 and 10.16.1. Example from 10.16.1:

```
08-08 11:51:04.654  RC debug: Starting connection for com.android.billingclient.api.BillingClientImpl@a01c04a
08-08 11:51:04.873  ActivityManager: Force stopping service ... InAppBillingService
08-08 11:51:04.905  W/BillingClient: Billing service died.
08-08 11:51:04.928  RC warn: Billing Service disconnected for com.android.billingclient.api.BillingClientImpl@a01c04a
[zero further "Starting connection" lines until a background/foreground at 11:54:15]
```

Reproduces on a stock Android emulator with a Google Play system image and no Google account, so it should be easy to run anywhere:

1. Configure the SDK with verbose logs and open a screen that calls `Purchases.getOfferings()`.
2. Simulate sustained Play Store churn from the host: run `adb shell am force-stop com.android.vending` in a loop every ~300ms for ~110 seconds.
3. While the loop is running, trigger a `getOfferings()` call so the SDK starts a billing connection.
4. The wedge forms when the loop kills the billing service during an in-flight connection attempt. Signature: "Starting connection", then code 5 "already in the process of connecting" with `pending_request_count` > 0 in the `google_billing_setup_finished` diagnostics entry, then "Billing service died", then permanent silence.
5. Stop the loop. Wait several minutes with a healthy Play Store.
6. Background and foreground the app.

Expected: after the service churn ends, the exponential backoff reconnect ladder re-establishes the connection within seconds to minutes, and the pending `getOfferings` completes or fails with an error.

Actual: no reconnection attempt is ever scheduled. `getOfferings()` pends silently and indefinitely (observed 13m03s on the Pixel, and 201.8s / 192.7s / 203.3s on three emulator runs where the pend length was simply how long we waited before foregrounding). The lifecycle foreground handler is the only thing that ever calls `startConnection` again.

Timing note: a single force-stop does not reproduce this (the SDK rebinds in ~400ms). The disconnect has to land inside the in-flight connection window, which the sustained loop makes reliable (3 out of 3 attempts across emulator runs, plus the Pixel run).

4. **Other information**

Our reading of the current code, consistent with all observed logs:

- `BillingWrapper.onBillingServiceDisconnected()` is a no-op apart from logging and diagnostics, so a disconnect never re-arms the retry ladder by itself.
- The `DEVELOPER_ERROR` (code 5) branch of `onBillingSetupFinished` intentionally does nothing, trusting the in-flight attempt to deliver a terminal callback. When the service dies mid-connect, BillingClient never delivers that callback, so the trust is misplaced and both paths drop the ball.
- `retryBillingServiceConnectionWithExponentialBackoff()` is only invoked from terminal `onBillingSetupFinished` error codes, which never arrive in this scenario.

We have a fix in progress and intend to submit a PR (re-arming reconnection when a disconnect lands while requests are pending / a connection attempt is in flight).

Related: we found a second, independent way `getOfferings()` pends forever (HTTP layer, hung socket not bounded by any timeout) while investigating the same production stall. Filed separately as #3923.

**Additional context**

Production impact for us: a paywall that silently never loads offerings, with no error surfaced to the app, so the app cannot even show a retry affordance based on SDK signals. The `pending_request_count` growth in consecutive diagnostics entries (2, then 4, then 6 in our logs as more calls queued) makes the wedge visible in the SDK's own diagnostics stream.

aelaguiz:
### Checklist
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids (not applicable: Play Billing reconnection is Android specific)

### Motivation

Fixes #3922 (full repro and log evidence in the issue).

When the Play billing service dies while a `startConnection` attempt is in flight, that attempt never delivers `onBillingSetupFinished`. A concurrent `startConnection` call in that state yields `DEVELOPER_ERROR` ("Client is already in the process of connecting to billing service."), which `onBillingSetupFinished` deliberately ignores:

```kotlin
BillingClient.BillingResponseCode.DEVELOPER_ERROR -> {
    // Billing service is already trying to connect. Don't do anything.
}
```

and `onBillingServiceDisconnected` only logged and tracked diagnostics:

```kotlin
override fun onBillingServiceDisconnected() {
    log(LogIntent.WARNING) {
        BillingStrings.BILLING_SERVICE_DISCONNECTED_INSTANCE.format(billingClient?.toString())
    }
    diagnosticsTrackerIfEnabled?.trackGoogleBillingServiceDisconnected()
}
```

With no callback left to fire, nothing in the system ever calls `startConnection` again. Requests queued in `serviceRequests` wait forever, and `getOfferings` pends silently until an app lifecycle event happens to arm a new connection.

### Description

`onBillingServiceDisconnected` now calls the existing `retryBillingServiceConnectionWithExponentialBackoff()` when `serviceRequests` is non-empty. That is the whole change: one call into the retry ladder the wrapper already has (1s start, 15min cap, `reconnectionAlreadyScheduled` dedupe guard). Empty-queue behavior is unchanged, since the next API request still arms its own connection through `executeRequestOnUIThread`. A retry that fires while a live connect attempt is in flight is absorbed by the existing `DEVELOPER_ERROR` no-op or the `isReady` early return in `performStartConnection`, so there is no double-connect risk.

New tests in `BillingWrapperTest` (mocked handlers, no wall-clock waits):

- `onBillingServiceDisconnected with pending service requests re-arms the reconnection ladder`
- `onBillingServiceDisconnected without pending service requests does not schedule a reconnection`
- `disconnect during an in-flight connection attempt eventually executes queued requests` (full wedge choreography: request queued, DEVELOPER_ERROR ignored, disconnect schedules the retry, queued request completes)

A/B: with the fix, `BillingWrapperTest` passes 71/71. With `BillingWrapper.kt` reverted to main, exactly the two wedge tests fail. All four CI unit test lanes (`testDefaultsDebugUnitTest`, `testDefaultsReleaseUnitTest`, and both `CustomEntitlementComputation` variants) and `detektAll` pass locally.

Real-device validation, measured from a production Flutter app's test rig (repeated `force-stop` of the Play Store while a connection attempt is in flight and a `getOfferings` composite is queued):

| Scenario | Stock 10.16.1 | With this fix |
|---|---|---|
| Service death mid-connect with queued requests | Zero `Starting connection` attempts after the death; queued requests pended 13m03s on a physical Pixel (192-203s on an emulator) and released only on an app foreground event | Ladder re-armed after every mid-connect death: `Starting connection` at +1.0s, +2.1s, +4.2s, +8.4s, +16.2s, +32.1s, +64.3s; queued request completed about 18s after the service recovered, with zero user interaction and zero lifecycle events |

No behavior change for apps that never hit this state; the added call is gated on a non-empty request queue and reuses the existing scheduling guard.

---

> [!NOTE]
> **Medium Risk**
> Touches core Google Play Billing connection and retry timing; change is gated on a non-empty queue and reuses existing backoff/dedupe, with strong unit tests.
> 
> **Overview**
> Fixes a wedge where Play Billing dies mid-`startConnection`: the in-flight attempt never calls `onBillingSetupFinished`, and further `startConnection` calls only get `DEVELOPER_ERROR` with no retry, so **queued `serviceRequests` (e.g. offerings) could hang until a lifecycle event**.
> 
> **`onBillingServiceDisconnected`** now calls the existing **`retryBillingServiceConnectionWithExponentialBackoff()`** when the request queue is non-empty; empty-queue disconnects behave as before.
> 
> Adds **`BillingWrapperTest`** coverage for disconnect with/without pending requests and the full mid-connect death scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 833c1e028e130bf6d0e891f8d5694f7c3bdb111b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
@@ -752,6 +752,15 @@ internal class BillingWrapper(
             BillingStrings.BILLING_SERVICE_DISCONNECTED_INSTANCE.format(billingClient?.toString())
         }
         diagnosticsTrackerIfEnabled?.trackGoogleBillingServiceDisconnected()
+        // If the service died while a connection attempt was in flight (e.g. the Play Store was
+        // updating or was killed mid-connection), that attempt will never deliver
+        // onBillingSetupFinished, and a later startConnection() call while BillingClient is still
+        // CONNECTING only produces a DEVELOPER_ERROR ("client is already in the process of
+        // connecting") which does not retry either. Without re-arming the reconnection ladder
+        // here, queued service requests would wait forever.
+        if (serviceRequests.isNotEmpty()) {
+            retryBillingServiceConnectionWithExponentialBackoff()
+        }
     }
 
     /**
--- a/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/google/BillingWrapperTest.kt
@@ -1576,6 +1576,106 @@ class BillingWrapperTest {
     }
 
     @Test
+    fun `onBillingServiceDisconnected with pending service requests re-arms the reconnection ladder`() {
+        every { mockClient.isReady } returns false
+        every { mockClient.queryProductDetailsAsync(any(), any()) } just runs
+
+        val capturedDelays = mutableListOf<Long>()
+        val runnableSlot = slot<Runnable>()
+        every {
+            handler.postDelayed(capture(runnableSlot), capture(capturedDelays))
+        } returns true
+
+        // Queue a request while the client is not ready, mirroring work queued during a
+        // connection attempt.
+        wrapper.queryProductDetailsAsync(
+            productType = ProductType.SUBS,
+            productIds = setOf("product_a"),
+            onReceive = { },
+            onError = { fail("shouldn't be an error") }
+        )
+        val scheduledCallsBeforeDisconnect = capturedDelays.size
+
+        wrapper.onBillingServiceDisconnected()
+
+        // The disconnect must schedule a reconnection using the existing backoff ladder.
+        assertThat(capturedDelays.size).isEqualTo(scheduledCallsBeforeDisconnect + 1)
+        assertThat(capturedDelays.last()).isEqualTo(1000L) // RECONNECT_TIMER_START_MILLISECONDS
+
+        // Running the scheduled reconnection starts a new connection attempt.
+        runnableSlot.captured.run()
+        verify(exactly = 2) { mockClient.startConnection(any()) }
+
+        // Once the new connection succeeds, the queued request executes instead of hanging.
+        every { mockClient.isReady } returns true
+        every {
+            handler.postDelayed(capture(runnableSlot), any())
+        } answers {
+            runnableSlot.captured.run()
+            true
+        }
+        billingClientStateListener!!.onBillingSetupFinished(billingClientOKResult)
+        verify(exactly = 1) { mockClient.queryProductDetailsAsync(any(), any()) }
+    }
+
+    @Test
+    fun `onBillingServiceDisconnected without pending service requests does not schedule a reconnection`() {
+        val capturedDelays = mutableListOf<Long>()
+        every {
+            handler.postDelayed(any(), capture(capturedDelays))
+        } returns true
+
+        wrapper.onBillingServiceDisconnected()
+
+        assertThat(capturedDelays).isEmpty()
+    }
+
+    @Test
+    fun `disconnect during an in-flight connection attempt eventually executes queued requests`() {
+        // Reproduces the wedge where the billing service dies while a connection attempt is in
+        // flight: the in-flight attempt never delivers onBillingSetupFinished, and a concurrent
+        // startConnection() only produces DEVELOPER_ERROR ("client is already in the process of
+        // connecting"), which must not be the end of the line for queued requests.
+        every { mockClient.isReady } returns false
+        every { mockClient.queryProductDetailsAsync(any(), any()) } just runs
+
+        val scheduledRunnables = mutableListOf<Runnable>()
+        every {
+            handler.postDelayed(capture(scheduledRunnables), any())
+        } returns true
+
+        wrapper.queryProductDetailsAsync(
+            productType = ProductType.SUBS,
+            productIds = setOf("product_a"),
+            onReceive = { },
+            onError = { fail("shouldn't be an error") }
+        )
+
+        // A second connection attempt while the first is in flight finishes with DEVELOPER_ERROR,
+        // which intentionally schedules nothing.
+        wrapper.onBillingSetupFinished(BillingClient.BillingResponseCode.DEVELOPER_ERROR.buildResult())
+        val scheduledBeforeDisconnect = scheduledRunnables.size
+
+        // The service then dies mid-connection: the in-flight attempt will never call back, so
+        // the disconnect is the only signal left and it must schedule a reconnection.
+        wrapper.onBillingServiceDisconnected()
+        assertThat(scheduledRunnables.size).isEqualTo(scheduledBeforeDisconnect + 1)
+
+        // The scheduled reconnection runs and the connection succeeds.
+        scheduledRunnables.last().run()
+        every { mockClient.isReady } returns true
+        every {
+            handler.postDelayed(capture(scheduledRunnables), any())
+        } answers {
+            scheduledRunnables.last().run()
+            true
+        }
+        billingClientStateListener!!.onBillingSetupFinished(billingClientOKResult)
+
+        verify(exactly = 1) { mockClient.queryProductDetailsAsync(any(), any()) }
+    }
+
+    @Test
     fun `normalizing Google purchase returns correct product ID and null store user ID`() {
         val expectedProductID = "expectedProductID"
 


purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
```diff
@@ -752,6 +752,15 @@ internal class BillingWrapper(
...
         }
         diagnosticsTrackerIfEnabled?.trackGoogleBillingServiceDisconnected()
+        // If the service died while a connection attempt was in flight (e.g. the Play Store was
+        // updating or was killed mid-connection), that attempt will never deliver
+        // onBillingSetupFinished, and a later startConnection() call while BillingClient is still
+        // CONNECTING only produces a DEVELOPER_ERROR ("client is already in the process of
+        // connecting") which does not retry either. Without re-arming the reconnection ladder
+        // here, queued service requests would wait forever.
+        if (serviceRequests.isNotEmpty()) {
+            retryBillingServiceConnectionWithExponentialBackoff()
```
tonidero, line 762: We think we probably shouldn't attempt a reconnection in this scenario since it might end up in an infinite loop. Billing service disconnections shouldn't trigger reconnections by themselves. One potential change we could do is to clean the pending requests on specific errors so we don't leave them hanging. 

Have you been able to reproduce this in production? We wonder if this may actually happen in real scenarios.

tonidero (commented):
Thank you for your contribution @aelaguiz! Just left a couple of questions/comments. Let me know what you think!

purchases/src/main/kotlin/com/revenuecat/purchases/google/BillingWrapper.kt
gurgen-k-y: @tonidero we could reproduce this issue in production by using third-party SDKs that also use the Billing SDK. We suspect this is due to version issues.

PR #3925, opened 2026-08-08
fix: bound silent reads and complete native login recovery

### LINKED ISSUE #3923: getOfferings hangs indefinitely on an established but silent HTTP connection: no timeout fires, and concurrent calls coalesce onto the hung request

**Describe the bug**

When a connection to `api.revenuecat.com` is established but the far end then goes silent (socket accepted, no bytes ever returned), the SDK's HTTP layer does not bound the request with any timeout. `Purchases.getOfferings()` pends indefinitely with no error, no retry, and no fallback-host attempt. We measured:

- 10.16.0: 402.4s and 155.5s pends, each ended only when we externally killed the hung socket.
- 10.16.1: 192.0s pend, ended only by external socket kill. We held a window of more than 120 seconds after request start with zero SDK timeout, retry, or failover activity in verbose logs. The re-tiered request timeouts that shipped in 10.16.1 (#3746, #3749) do not bound this case.

The SDK's own diagnostics stream proves it saw the hang without surfacing anything to the caller. After we killed the socket, it recorded:

```
http_request_performed {"host":"api.revenuecat.com","endpoint_name":"get_offerings","response_time_millis":403809,"successful":false,"response_code":-1,"connection_error_reason":"NO_NETWORK"}
http_request_performed {"host":"api.revenuecat.com","endpoint_name":"remote_config","response_time_millis":403830,"successful":false,"response_code":-1,"connection_error_reason":"NO_NETWORK"}
```

A 403 second response time against timeouts configured in the tens of seconds, and `connection_error_reason` is reported as `NO_NETWORK` even though the device's network was healthy the entire time.

Two dedupe mechanisms then amplify one hung socket into a full wedge of the offerings surface:

- Every subsequent `getOfferings()` call coalesces onto the hung request: `Same call already in progress, adding to callbacks map with key: BackgroundAwareCallbackCacheKey(cacheKey=[/v1/subscribers/<app_user_id>/offerings], appInBackground=false)`
- Config refresh is skipped: `Remote config refresh already in progress. Skipping.`

So no later call can ever succeed while the first one hangs, and nothing ever times the first one out.

Finally, in runs where we killed the hung socket while leaving the network path healthy, the SDK retried immediately and completed a successful offerings fetch within 29 to 400 ms of the kill. That confirms the network and the API were fine throughout; the only missing piece is a deadline on the hung request.

1. Environment
   1. Platform: Android. We consume the SDK from a Flutter app in production via purchases_flutter 10.8.0 / purchases-hybrid-common 18.29.0, but the behavior is entirely in the native Android SDK's HTTP layer.
   2. SDK version: reproduced on purchases-android 10.16.0 and 10.16.1 (BillingClient 8.3.0, not involved in this issue).
   3. OS version: Android 16 (stock emulator with a Google Play image).
   4. Android Studio version: n/a (Gradle CLI builds).
   5. How widespread: any user whose TCP connection to the API goes silent mid-request (NAT timeout, captive portal, flaky mobile network, middlebox) holds the entire offerings surface hostage for as long as the socket survives. In our tests the pend lasted 2.5 to 7 minutes and ended only because we severed the socket externally; nothing in the SDK's behavior suggests it would ever end it. Fleet telemetry shows roughly one silent paywall stall per week in this failure class, and the "late success the moment the socket dies" signature matches what we see in the wild.

2. Debug logs

Verbose SDK logs forwarded through our Flutter log handler; payloads are the SDK's own lines. During the pend, per additional `getOfferings()` call:

```
08-08 08:21:36.247  RC debug: Same call already in progress, adding to callbacks map with key: BackgroundAwareCallbackCacheKey(cacheKey=[/v1/subscribers/<app_user_id>/offerings], appInBackground=false)
08-08 08:27:42.451  RC debug: Remote config refresh already in progress. Skipping.
```

Zero `API request started` / error / retry lines for the hung endpoint for the entire pend. Then, immediately after we killed the hung socket (network healthy):

```
08-08 08:28:17.495  RC diagnostics: http_request_performed {"host":"api.revenuecat.com","endpoint_name":"get_offerings","response_time_millis":403809,"successful":false,"response_code":-1,"connection_error_reason":"NO_NETWORK"}
[new request attempts follow within ~30-80ms of the socket dying]
```

In the variant where the network path was healthy at socket death, those immediate retries completed a successful offerings fetch (10.16.0: success 0.4s after the kill following a 155.5s pend; 10.16.1: getOfferings completion 29ms and offerings delivered 38ms after the kill following a 192.0s pend).

1. Launch an Android emulator with `-http-proxy http://127.0.0.1:<port>` pointing at a small proxy that accepts the CONNECT for `api.revenuecat.com` (and the fallback host) and then forwards nothing, keeping the socket open and silent. This models an established connection whose far end stops responding.
2. Configure the SDK (verbose logs) and call `Purchases.getOfferings()`.
3. Observe the pend: no callback, no timeout, no retry, no fallback host, indefinitely.
4. Call `getOfferings()` again several times; each logs "Same call already in progress" and joins the hung request. Trigger a foreground config refresh; it logs "Remote config refresh already in progress. Skipping."
5. Kill the hung socket at the proxy while switching the proxy to pass traffic through normally.
6. The SDK instantly retries and succeeds; the diagnostics entry for the hung request reports the full multi-hundred-second `response_time_millis`.

Expected: the request fails with a network error after the HTTP layer's timeout budget (tens of seconds at most), callers get an error callback, and a later `getOfferings()` can start a fresh request.

Actual: the request hangs for as long as the socket survives (402s / 155s / 192s measured), every concurrent caller is silently parked on it, and nothing is ever surfaced to the app.

4. **Other information**

The connect/read timeout configuration appears to only bound the connection handshake and inter-byte gaps in some paths, not a fully silent established stream on this path; whatever the mechanism, the observed behavior on both 10.16.0 and 10.16.1 is that no deadline fires. The request-level coalescing and the config-refresh skip are reasonable dedupe designs on their own, but combined with an unbounded request they convert one bad socket into a permanent wedge of every future offerings call.

We have a fix in progress and intend to submit a PR (an overall deadline on dispatched requests so hung sockets fail loudly and coalesced callers get an error callback).

Related: #3922 covers an independent wedge we reproduced in the same investigation (billing connection ladder permanently stops after a disconnect lands during an in-flight reconnect). Either defect alone produces the same production symptom: `getOfferings()` pending silently and indefinitely at a paywall.

**Additional context**

Production impact for us: a paywall spinner that never resolves, with no signal the app can act on. The immediate-success-after-socket-death behavior means affected users sometimes see offerings load minutes later for no visible reason, which matches the diagnostics we quoted (`response_time_millis=403809`, `successful=false`) followed by an instant successful retry.

aelaguiz:
An established RevenueCat connection can accept a request and then stop sending bytes. The SDK currently leaves reads unbounded, so native login and any joined callers can remain pending indefinitely. This applies the requested flat 30-second socket read timeout and verifies that the public login API completes and can recover on the same Purchases instance.

Fixes #3923. Related application incident: funcountry/psmobile#6001.

## User-Facing Changes

A silent established connection returns the SDK's existing network error instead of leaving callers waiting indefinitely. After connectivity recovers, the same SDK instance can complete login and accept the signed-in user.

No client UI changes.

## Implementation

- `HTTPTimeoutManager` supplies a flat 30-second read timeout, independent of the existing connection timeout tiers. The existing unit-test divider remains applicable.
- `HTTPClient.getConnection` applies that timeout and preserves current main's `useCaches = false` behavior.
- Existing exception mapping, request coalescing, endpoint-specific retry and fallback policies own completion. Login does not gain an offerings fallback or a new retry owner. Thirty seconds bounds an individual silent read, not the total duration of a composite login.

## Reproduction: Before And After

The controlled server accepts the real SDK HTTP request and sends no response. It keeps the connection open through the decisive assertion; teardown does not release the operation being tested.

**Before:** RevenueCat Android 10.20.0 (`7ab95766efd10601a1c790f6129100a2eacf729e`), using its original `readTimeout = 0`. Both public `Purchases.logIn` reproductions fail the external observation deadline: a silent login never calls back, and a silent subscriber-attribute upload prevents the original login from continuing.

**After:** the same 10.20.0 source with the two production changes in this PR. Both reproductions pass. The silent login completes both joined callers exactly once with `NetworkError`; a subsequent login on the same instance receives valid CustomerInfo, caches it, and accepts the requested app user ID. When attribute synchronization is silent, its timeout releases the original login, which completes successfully without replacing Purchases.

The updated upstream branch also passes these public API tests. Its fixtures retain the real Purchases, PurchasesOrchestrator, IdentityManager, subscriber-attribute chain, Backend, Dispatcher, HTTP client, signing manager, and identity cache. Ancillary Android billing/lifecycle integrations are test doubles.

## Validation

On the branch updated against current upstream main: **220 tests passed** across HTTPClientTest, HTTPTimeoutManagerTest, NativeLoginRecoveryTest, and BackendTest; `detektAll` passed.

The HTTP tests use independently bounded executor futures and exact socket/executor cleanup. Removing only the production read-timeout assignment makes both silent-read controls fail at their ten-second observation deadline and exit normally; restoring it makes both pass. Backend coverage explicitly coordinates overlap, releases both joined offerings callers on error, and verifies a later healthy same-user request executes and succeeds. The public login tests cover joined login failure, same-instance retry, and the pre-login attribute dependency. Existing connection-timeout tier tests remain unchanged.

Application adoption and physical-device onboarding verification are owned by funcountry/psmobile#5988 and funcountry/psmobile#6001. The historical sessions prove unfinished native login; they do not identify which historical socket or endpoint stopped responding. This PR fixes the reproduced unbounded-read mechanism and does not claim to eliminate network outages.

## QA Test Plan

Use the repository tests above with a server that accepts a connection without responding. For an integrating Android app, an engineer supplies a test build and controlled network setup: allow the initial request to remain silent, observe the normal error, permit subsequent requests, and retry from the existing screen. Require successful identity acceptance and the original operation's usable result without restarting the app.

---

> [!NOTE]
> **Medium Risk**
> Changes core HTTP behavior for all SDK requests by replacing unbounded reads with a 30s cap; low logic surface area but broad network impact if timeouts are too aggressive on slow links.
> 
> **Overview**
> **Bounds HTTP socket reads** so established connections that stop sending data no longer hang callers indefinitely. `HTTPClient` now sets `readTimeout` from a flat **30-second** value in `HTTPTimeoutManager`, separate from existing connect-timeout tiers and fail-fast reduction.
> 
> User-visible effect: silent reads surface the SDK’s normal **network error** within that window instead of leaving `Purchases.logIn` (and coalesced/joined work) pending forever; the same `Purchases` instance can complete login after connectivity recovers. Existing fallback, coalescing, and exception mapping are unchanged—tests cover silent-read timeout, offerings fallback after a silent main host, coalesced offerings release on timeout, and end-to-end native login recovery (including a silent attribute sync blocking the login chain).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3efa5e82bed4c5921330022fe9fdeb4ed790a61. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
@@ -701,7 +701,11 @@ internal class HTTPClient(
     private fun getConnection(request: HTTPRequest, timeoutMs: Long): HttpURLConnection {
         return (request.fullURL.openConnection() as HttpURLConnection).apply {
             connectTimeout = timeoutMs.toInt()
-            // We leave the read timeout to the default (readTimeout = 0), which means infinite.
+            // Bound reads with the same budget: with the default (readTimeout = 0, infinite), an
+            // established connection that goes silent hangs the request forever, so the request
+            // never errors, coalesced callers waiting on it never release, and the retry/fallback
+            // machinery never runs.
+            readTimeout = timeoutMs.toInt()
             request.headers.forEach { (key, value) ->
                 addRequestProperty(key, value)
             }
--- a/purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
@@ -33,6 +33,7 @@ import io.mockk.spyk
 import io.mockk.verify
 import okhttp3.mockwebserver.MockResponse
 import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.SocketPolicy
 import okio.Buffer
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatThrownBy
@@ -2230,6 +2231,86 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
         ).isEqualTo(HTTPTimeoutManager.REDUCED_TIMEOUT_MS / HTTPTimeoutManager.TEST_DIVIDER)
         verify(exactly = 1) {
             timeoutManager.recordRequestResult(host, HTTPTimeoutManager.RequestResult.OTHER_RESULT)
+        }
+    }
+
+    @Test
+    fun `a connection that goes silent after connecting fails within the configured timeout instead of hanging`() {
+        val endpoint = Endpoint.GetOfferings("test_user_id")
+
+        val appConfig = createAppConfig(proxyURL = null)
+        val timeoutManager = spyk(HTTPTimeoutManager(appConfig))
+        client = createClient(appConfig = appConfig, timeoutManager = timeoutManager)
+
+        // The server accepts the connection but never sends a single response byte, so only the
+        // read timeout can bound this request: the connection itself succeeds.
+        server.enqueue(MockResponse().setSocketPolicy(SocketPolicy.NO_RESPONSE))
+
+        assertThatThrownBy {
+            client.performRequest(
+                baseURL,
+                endpoint,
+                body = null,
+                postFieldsToSign = null,
+                mapOf("" to ""),
+                fallbackBaseURLs = emptyList(),
+            )
+        }.isInstanceOf(SocketTimeoutException::class.java)
+    }
+
+    @Test
+    fun `a connection that goes silent after connecting fails over to the fallback URL`() {
+        val endpoint = Endpoint.GetOfferings("test_user_id")
+        assert(endpoint.supportsFallbackBaseURLs)
+
+        val appConfig = createAppConfig(proxyURL = null)
+        val timeoutManager = spyk(HTTPTimeoutManager(appConfig))
+        val host = baseURL.host
+        client = createClient(appConfig = appConfig, timeoutManager = timeoutManager)
+
+        // The main server accepts the connection but never responds; the fallback server works.
+        server.enqueue(MockResponse().setSocketPolicy(SocketPolicy.NO_RESPONSE))
+
+        val fallbackServer = MockWebServer()
+        val fallbackBaseURL = fallbackServer.url("/v1").toUrl()
+        val validJsonPayload = """{"offerings": [], "current_offering_id": null}"""
+        fallbackServer.enqueue(
+            MockResponse()
+                .setBody(validJsonPayload)
+                .setResponseCode(RCHTTPStatusCodes.SUCCESS)
+        )
+
+        every {
+            mockETagManager.getHTTPResultFromCacheOrBackend(
+                RCHTTPStatusCodes.SUCCESS,
+                validJsonPayload,
+                eTagHeader = any(),
+                urlString = any(),
+                refreshETag = false,
+                requestDate = any(),
+                verificationResult = VerificationResult.NOT_REQUESTED,
+                isLoadShedderResponse = false,
+                isFallbackURL = true,
+            )
+        } returns HTTPResult.createResult(RCHTTPStatusCodes.SUCCESS, validJsonPayload)
+
+        try {
+            val result = client.performRequest(
+                baseURL,
+                endpoint,
+                body = null,
+                postFieldsToSign = null,
+                mapOf("" to ""),
+                fallbackBaseURLs = listOf(fallbackBaseURL),
+            )
+
+            assertThat(result.responseCode).isEqualTo(RCHTTPStatusCodes.SUCCESS)
+            // The silent socket counts as a main-source timeout for the fail-fast memory.
+            verify(exactly = 1) {
+                timeoutManager.recordRequestResult(host, HTTPTimeoutManager.RequestResult.MAIN_SOURCE_TIMED_OUT)
+            }
+        } finally {
+            fallbackServer.shutdown()
         }
     }
 
--- a/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendTest.kt
+++ b/purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendTest.kt
@@ -75,6 +75,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.annotation.Config
 import java.io.IOException
+import java.net.SocketTimeoutException
 import java.net.URL
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.LinkedBlockingQueue
@@ -1687,6 +1688,44 @@ class BackendTest {
         asyncBackend.getOfferings(appUserID, appInBackground = false, onSuccess = { _, _, _ ->
             lock.countDown()
         }, onError = onReceiveOfferingsErrorHandler)
+        lock.await(defaultTimeout, TimeUnit.MILLISECONDS)
+        assertThat(lock.count).isEqualTo(0)
+        verify(exactly = 1) {
+            mockClient.performRequest(
+                mockBaseURL,
+                Endpoint.GetOfferings(appUserID),
+                body = null,
+                postFieldsToSign = null,
+                any(),
+            )
+        }
+    }
+
+    @Test
+    fun `given multiple get offerings calls for same user, all coalesced callers receive the error when the request times out`() {
+        every {
+            mockClient.performRequest(
+                mockBaseURL,
+                Endpoint.GetOfferings(appUserID),
+                body = null,
+                postFieldsToSign = null,
+                any(),
+            )
+        } answers {
+            Thread.sleep(200)
+            throw SocketTimeoutException("Read timed out")
+        }
+        val lock = CountDownLatch(2)
+        asyncBackend.getOfferings(appUserID, appInBackground = false, onSuccess = { _, _, _ ->
+            fail<Unit>("Should be error")
+        }, onError = { _, _ ->
+            lock.countDown()
+        })
+        asyncBackend.getOfferings(appUserID, appInBackground = false, onSuccess = { _, _, _ ->
+            fail<Unit>("Should be error")
+        }, onError = { _, _ ->
+            lock.countDown()
+        })
         lock.await(defaultTimeout, TimeUnit.MILLISECONDS)
         assertThat(lock.count).isEqualTo(0)
         verify(exactly = 1) {


purchases/src/main/kotlin/com/revenuecat/purchases/common/HTTPClient.kt
```diff
@@ -701,7 +701,11 @@ internal class HTTPClient(
     private fun getConnection(request: HTTPRequest, timeoutMs: Long): HttpURLConnection {
         return (request.fullURL.openConnection() as HttpURLConnection).apply {
             connectTimeout = timeoutMs.toInt()
-            // We leave the read timeout to the default (readTimeout = 0), which means infinite.
+            // Bound reads with the same budget: with the default (readTimeout = 0, infinite), an
+            // established connection that goes silent hangs the request forever, so the request
+            // never errors, coalesced callers waiting on it never release, and the retry/fallback
+            // machinery never runs.
+            readTimeout = timeoutMs.toInt()
```
tonidero, line 708: We shouldn't reuse the same connection timeouts for read, since in situations of low connectivity and bigger responses those may be lower. I would probably add a flat `30s` timeout for read, which can also be obtained from the timeout manager, from a separate method, to keep the logic there.

tonidero (commented):
Thank you so much for your contribution @aelaguiz! This makes sense to us! I just left a suggestion to have a single flat and bigger timeout for reads. Let me know if you have any questions though! Once that change is in, we can get this merged 👍

purchases/src/test/java/com/revenuecat/purchases/common/HTTPClientTest.kt
```diff
@@ -2347,6 +2348,86 @@ internal class HTTPClientTest: BaseHTTPClientTest() {
         }
     }
 
+    @Test
+    fun `a connection that goes silent after connecting fails within the configured timeout instead of hanging`() {
```
aelaguiz, line 2358: P2: This new NO_RESPONSE test calls performRequest on the test thread with no independent observation bound. Reverting the production readTimeout assignment hangs the test before @After can close the server. Restore the already-prepared task-owned executor + bounded Future.get + exact server/executor finally cleanup for both new silent-HTTP tests. The unfixed source should fail and exit ordinarily; socket cleanup must not be what makes the fixed assertion pass.

purchases/src/test/java/com/revenuecat/purchases/common/backend/BackendTest.kt
```diff
@@ -1700,6 +1701,44 @@ class BackendTest {
         }
     }
 
+    @Test
+    fun `given multiple get offerings calls for same user, all coalesced callers receive the error when the request times out`() {
```
aelaguiz: P2: The prepared recovery test has regressed to sleep-based overlap and only two error callbacks. Use explicit start/release coordination, then require a third same-user getOfferings call to execute and succeed. Public login exercises identifyCallbacks, so it does not replace this separate offeringsCallbacks release/reuse assertion.

aelaguiz (commented):
## Pro review: changes requested at 5cdb84d4f7716965baa32cf9919386bdb040df43

The production correction is the right small repair: HTTPClient obtains one flat read timeout from HTTPTimeoutManager, retains the existing connection tiers and callback/error ownership, and preserves upstream's useCaches=false. This is a socket-read interval, not a total-login deadline.

I accept the new public Purchases.logIn controls as meaningful native composition evidence. They exercise real orchestration, identity/attribute synchronization, Backend, Dispatcher, HTTP and cache/CustomerInfo handling, including main-looper callback dispatch. The supplied before run fails with the unbounded source; the fixed run releases both joined callers, retains the old identity on failure, then accepts a healthy same-instance login. The attribute-upload case progresses to the original login without a reset. Platform doubles and the still-missing native-to-Flutter device composition are stated honestly. I did not rerun the native suite in this review.

Two previously prepared test repairs have not reached this pushed head:

1. **P2: Restore bounded ownership/cleanup in the two silent-HTTP tests.** `HTTPClientTest.kt`, `a connection that goes silent after connecting fails within the configured timeout instead of hanging` and `a connection that goes silent after connecting fails over to the fallback URL`, call performRequest synchronously against NO_RESPONSE. If the production read-timeout assignment regresses, neither test reaches its assertions or teardown; BaseHTTPClientTest only shuts down the server in @After. The already-prepared local version used a task-owned executor, an independent bounded Future.get, and exact server/executor cleanup in finally. Port that small correction instead of adding another framework. Removing the read-timeout assignment must produce an ordinary failing test and exit, not a stuck test process. Cleanup must occur after the decisive observation, not manufacture the successful timeout.

2. **P2: Restore deterministic offerings callback-release/reuse coverage.** `BackendTest.kt`, `given multiple get offerings calls for same user, all coalesced callers receive the error when the request times out`, currently uses Thread.sleep(200) to arrange overlap and stops after two errors. Use explicit start/release coordination and assert that a following same-user offerings request actually executes and completes. The public-login test proves the identifyCallbacks path, not the separate offeringsCallbacks path. This is the already-prepared coalescing/recovery test, not additional production scope.

No further native production abstraction, retry mechanism or timeout policy is requested. Preserve the recorded earlier intermittent attribute-followup result as an unexplained observation; the later passing public/broader runs do not retroactively explain it.

The app's original-onboarding/real-prices/native-sheet proof belongs to [the application integration (PR #5807)](https://github.com/funcountry/psmobile/pull/5807) and [native-to-onboarding completion (#6001)](https://github.com/funcountry/psmobile/issues/6001), not an expansion of this upstream implementation. This native PR can be judged independently once its bounded regressions are repaired. No merge or release authorization is given.

aelaguiz (commented):
Pro signs off the native silent-read repair at c3efa5e82bed4c5921330022fe9fdeb4ed790a61. The implementation is right, this PR is ready from source/test review, and its native scope is complete. This supersedes my prior requested native test corrections. GitHub rejected a formal APPROVE because this connection acts as the PR author; this comment records the review verdict, not a GitHub approval.

Reviewed the complete six-file change and supplied current raw controls. The flat 30-second read interval is independent of connection tiers; useCaches=false and existing endpoint, exception and callback ownership remain. This is not a total-login deadline.

Both prior test findings are closed: the silent-HTTP tests now have independent observation bounds and exact socket/executor cleanup, with retained ordinary-failure negative controls; the offerings test explicitly coordinates overlap and proves a later healthy same-user request succeeds. Public Purchases.logIn covers joined failure, same-instance successful retry with accepted identity/CustomerInfo, and release of a silent attribute-upload prerequisite. Current supplied XML totals 220 passing tests. I did not independently execute these tests.

No further native production mechanism or application device evidence is required for this native scoped signoff. Application adoption and original-offer completion remain with https://github.com/funcountry/psmobile/pull/5807, https://github.com/funcountry/psmobile/issues/6001 and https://github.com/funcountry/psmobile/issues/5988. This verdict does not claim those outcomes, identify the historical socket, or authorize merge/release/production mutation. The ordinary maintainer and final CI workflow remains separate.

aelaguiz:
@tonidero The updated patch now uses the separate flat 30-second read timeout you requested, through `HTTPTimeoutManager`, while preserving the existing connection tiers, error handling and callback ownership.

The current head is `c3efa5e82bed4c5921330022fe9fdeb4ed790a61`. The focused native suite has 220 passing tests plus `detektAll`, including public `Purchases.logIn` recovery after a silent response, release of joined callbacks, a healthy subsequent login on the same instance, and a silent pre-login attribute upload. Removing the read-timeout assignment makes the bounded negative controls fail normally. Android application qualification also reaches real prices and the Play purchase sheet after the native error and a deliberate retry.

We want to consume this through RevenueCat's maintained releases rather than distribute our own patched SDK. Could you review the updated implementation and let us know what remains for inclusion in an official Android release, and the supported adoption path for Flutter customers if the native release arrives before the wrapper update?

The current Android 10.22.1 source still leaves this read timeout unbounded, so moving to the latest Flutter release alone does not provide this recovery.

PR #3926, opened 2026-08-10
Fix misleading ui_config warnings for projects without paywalls

rickvdl:
## Description
As reported on [a community post](https://community.revenuecat.com/sdks-51/kmp-sdk-warning-could-not-resolve-remote-config-blob-s-for-4-of-4-requested-item-s-in-topic-ui-config-7785), for apps without a published `ui_config` blobs (any app without any paywalls). The `UiConfigProvider` attempts to merge the (4) independently fetched blobs in order to construct a `UIConfig` object.

Merging these fails if any of the 4 blobs are absent (which makes sense). However when a `UIConfig` genuinely isn't available for an app, we should not be logging this as a failure of individual blobs failing to load when merging. But rather just treat this as the `UIConfig` not being available. 

In this case the following was logged every time
```
WARN: Could not resolve remote config blob(s) for 4 of 4 requested item(s) in topic 'ui_config': [app, localizations, variable_config, custom_variables]. Returning null
```

## Changes
We'll now check whether we have any of the (currently 4) available parts (blobs) for the `UIConfig`. If not, we'll consider the `UIConfig` as unavailable, rather than attempting to merge and construct a `UIConfig` object, which would fail and trigger the log. This should be in line with the reasoning of https://github.com/RevenueCat/purchases-android/pull/3894

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to ui_config resolution and logging for projects without paywalls; remote config merge and auth paths are untouched.
> 
> **Overview**
> Fixes noisy **WARN** logs for apps with **no paywalls**, where `UiConfigProvider` used to call the all-or-nothing `mergeItemsBlobData` path even when none of the four `ui_config` parts were published—surfacing *Could not resolve remote config blob(s) for 4 of 4…* on every read.
> 
> **`UiConfigProvider.resolve()`** now primes config via `topic()`, checks whether any of `app`, `localizations`, `variable_config`, or `custom_variables` exist, and returns `null` (treated as **not configured**) without merging when none are present. If the committed topic is empty but a refresh is already in flight, it re-reads via **`committedTopicAfterInFlightRefresh`** so paywall readiness does not treat stale empty cache as final while AppStart is still fetching.
> 
> **`RemoteConfigManager`** adds that helper: await the in-flight refresh without starting a new one, then return the latest committed topic.
> 
> Tests cover no-merge on absent/empty topics, in-flight refresh picking up newly committed parts, and cold `onPaywallConfigReady` completing without blob warnings or extra downloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 20ae109830867438c13af9307670267a3c909e11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
@@ -117,8 +117,15 @@ internal class UiConfigProvider(
         scope.cancel()
     }
 
-    private suspend fun resolve(): UiConfig? =
-        manager.mergeItemsBlobData<UiConfig>(RemoteConfigTopic.UiConfig, ITEM_KEYS)
+    private suspend fun resolve(): UiConfig? {
+        if (manager.isDisabled) return null
+        // topic() waits for or primes the initial config sync on a cold cache. Once that authoritative topic is
+        // available, an absent topic or one with none of the ui_config parts means the project has no ui_config;
+        // avoid asking the generic all-or-nothing merger to resolve four known-missing blobs and warning about it.
+        val topic = manager.topic(RemoteConfigTopic.UiConfig)
+        if (topic == null || ITEM_KEYS.none(topic::containsKey)) return null
+        return manager.mergeItemsBlobData<UiConfig>(RemoteConfigTopic.UiConfig, ITEM_KEYS)
+    }
 
     /**
      * Tells "there is nothing to resolve" apart from "resolving what is published failed". Only ever called
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
@@ -122,7 +122,10 @@ internal class UiConfigProvider(
         // topic() waits for or primes the initial config sync on a cold cache. Once that authoritative topic is
         // available, an absent topic or one with none of the ui_config parts means the project has no ui_config;
         // avoid asking the generic all-or-nothing merger to resolve four known-missing blobs and warning about it.
-        val topic = manager.topic(RemoteConfigTopic.UiConfig)
+        var topic = manager.topic(RemoteConfigTopic.UiConfig)
+        if (topic != null && ITEM_KEYS.none(topic::containsKey)) {
+            topic = manager.committedTopicAfterInFlightRefresh(RemoteConfigTopic.UiConfig)
+        }
         return if (topic == null || ITEM_KEYS.none(topic::containsKey)) {
             null
         } else {
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
@@ -123,8 +123,11 @@ internal class UiConfigProvider(
         // available, an absent topic or one with none of the ui_config parts means the project has no ui_config;
         // avoid asking the generic all-or-nothing merger to resolve four known-missing blobs and warning about it.
         val topic = manager.topic(RemoteConfigTopic.UiConfig)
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Nice!! Makes sense, thank you!

purchases/src/main/kotlin/com/revenuecat/purchases/common/uiconfig/UiConfigProvider.kt
```diff
@@ -117,8 +117,21 @@ internal class UiConfigProvider(
...
-    private suspend fun resolve(): UiConfig? =
-        manager.mergeItemsBlobData<UiConfig>(RemoteConfigTopic.UiConfig, ITEM_KEYS)
+    private suspend fun resolve(): UiConfig? {
+        if (manager.isDisabled) return null
+        // topic() waits for or primes the initial config sync on a cold cache. Once that authoritative topic is
+        // available, an absent topic or one with none of the ui_config parts means the project has no ui_config;
+        // avoid asking the generic all-or-nothing merger to resolve four known-missing blobs and warning about it.
+        var topic = manager.topic(RemoteConfigTopic.UiConfig)
+        if (topic != null && ITEM_KEYS.none(topic::containsKey)) {
+            topic = manager.committedTopicAfterInFlightRefresh(RemoteConfigTopic.UiConfig)
```
tonidero, line 127: Hmm is this part actually needed? I think `manager.topic(RemoteConfigTopic.UiConfig)` would use the cached topic if it exists, or if not yet, wait for the inflight request if any and then return that. I think returning the cached value is probably ok for this check? It's true that if the inflight request would return an actual value, there might be a bit of a race condition, but would self heal on the next attempt, and shouldn't be an issue I think...

rickvdl: Right, yes indeed as you said this was specifically for the second part, when the inflight request would return an actual value. It's true that it would self heal on the next request, so happy to leave this out of scope for simplicity as well if you prefer :)

tonidero: Well... this means we need to await for inflight requests now every time for apps without the ui config topic... So I would leave it as it was to minimize behavior change IMO. Honestly, feels that this "race condition" could just be considered a delay rather than a race condition 😅

rickvdl: Hmm so I think this was already the case if I'm not mistaken. If the empty topic was cached, and a refresh was inflight, resolving every ui config blob (part) would join in and wait for the same refresh to complete. 

The following psuedo code helped me wrap my head around it a bit more 😅 
```
UiConfigProvider.resolve()
    ↓
RemoteConfigManager.mergeItemsBlobData<UiConfig>(
    topic = UiConfig,
    itemKeys = ["app", "localizations", "variable_config", "custom_variables"]
)
    ↓
RemoteConfigManager.mergedBlobObject()
    ↓
async { resolveBlobBytes(UiConfig, "app") }
    ↓
RemoteConfigManager.resolveBlobBytes(UiConfig, "app")
    ↓
RemoteConfigManager.committedItem(UiConfig, "app")
    ↓
RemoteConfigManager.awaitConfigForRead()
    ↓
RemoteConfigManager.awaitInFlightRefresh()
    ↓
refreshCompletion.await()
```

Since this was done for each of the (4) `UIConfig` parts I think the change may actually be a bit of an improvement given the fact that we're now just awaiting once for the existing inflight request through `committedTopicAfterInFlightRefresh`.

tonidero: Ah right 🤦, I was actually comparing against the version before the last commit you pushed 😅 In that one, we only get the topic like `manager.topic(RemoteConfigTopic.UiConfig)`, which would get the topic from cache first and just do a null check/item keys check to skip it. But then again, it's a bit of a behavior change that your last commit does fix. So yeah, we can keep it! Thanks for explaining!

tonidero (approved):

PR #3914, opened 2026-08-07
refactor(checkpoints): Resolve a checkpoint's workflow from its rules

tonidero:
Connects the checkpoints resolver to the `checkpoint_rules` remote config ingested in #3907.

Until now the resolver picked a workflow at random from the workflows topic. Now the identifier selects its workflow through the rules the dashboard configured for it.

Audience evaluation is still out of scope, so rules are walked in the order the backend sent them. That ordering is a placeholder for the rules engine picking the first rule whose audience matches.

Also lets the SDK tell apart a checkpoint that isn't configured, one whose rules couldn't be read, and remote config being disabled entirely - the three used to be indistinguishable, and all three were reported to apps as an unknown checkpoint.

`error_checkpoint` stays simulated: nothing in the config-driven path throws, and it's the only way for the tester apps to exercise that branch.

Note for anyone running CheckpointTester: its identifiers now need real dashboard checkpoints, otherwise nothing is presented.

---

> [!NOTE]
> **Medium Risk**
> Changes how checkpoints pick paywalls in production (config-driven vs random) and alters NoAction reason semantics for unconfigured vs unreadable rules; well-covered by new unit tests but affects customer-facing checkpoint presentation.
> 
> **Overview**
> Replaces the PoC **random workflow** checkpoint resolver with **`CheckpointWorkflowResolverImpl`**, which maps each checkpoint identifier to a workflow via the **`checkpoint_rules`** remote-config topic.
> 
> **`CheckpointsConfigProvider`** now returns **`CheckpointRulesResolution`** (`Found`, `NotConfigured`, `Disabled`, `Unavailable`) instead of a nullable response, with classification after a blob read and a single retry if **`configGeneration`** changes mid-read. The resolver walks rules in dashboard order (audience matching still TODO), skips unservable rules, and maps outcomes to **`UNKNOWN_CHECKPOINT`**, **`NO_MATCH`**, **`DISABLED`**, or **`CONFIGURATION_UNAVAILABLE`**. **`error_checkpoint`** remains a simulated throw for tester apps.
> 
> Wiring adds **`CheckpointsConfigProvider`** in **`PurchasesFactory`** / **`PurchasesOrchestrator`** and updates CheckpointTester docs and inline use cases for the new behavior.
> 
[...]

PR #3381, opened 2026-04-23
Wire multipage workflow navigation into PaywallViewModel

vegaro:
## Summary
- `PaywallViewModel` fetches the workflow result and creates a `WorkflowNavigator` to manage step traversal
- Adds an `isWorkflowActive` flag to `PaywallState.Loaded.Components` (defaults to `false`). No caller sets it to `true` yet, a follow-up PR will.
- `handleWorkflowAction(componentId)` advances the navigator to the next step and re-builds paywall state with `isWorkflowActive = true`
- `handleBackNavigation()` pops the back stack via the navigator; `BackHandler` and `NavigateBack` calls in `InternalPaywall` use this before closing the paywall
- `LoadingPaywall` gains no-op stubs for both new interface methods

---

> [!NOTE]
> **Medium Risk**
> Introduces new navigation/state transitions for workflow-driven paywalls and changes back-button behavior, which could affect dismissal and screen rendering flow. Purchase/restore logic is untouched, but workflow step validation and action nullability may surface new configuration errors.
> 
> **Overview**
> Adds workflow step navigation to paywalls by having `PaywallViewModelImpl` keep a `WorkflowNavigator`, rebuild paywall state from the active workflow step on `handleWorkflowAction`, and consume back presses via `handleBackNavigation` before dismissing.
> 
> Tightens workflow button handling by making `WorkflowTrigger` actions nullable when `componentId` is missing (guarding clicks), and by validating/erroing during style creation with a new `PaywallValidationError.WorkflowButtonMissingComponentId` to prevent misconfigured workflow buttons.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa0f8ad7b395eebdf4f194b1654b66214e863fe8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3928, opened 2026-08-10
fix: Restart a retained workflow at its initial step when re-presented

facumenzella:
### Motivation

A multi-page (workflow) paywall reopens on whatever step the user left it on instead of starting over.

Alternative to #3921

### Description

`PaywallViewModel` is scoped to the host's `ViewModelStoreOwner`, so an embedded or dialog paywall can be
dismissed and shown again on the same instance. 

**Trade vs #3921:** that one re-resolves the workflow on every presentation, this one replays the
dismiss-time snapshot. With checkpoint rules feeding resolution (#3907, #3914), that could matter.

<details>

# AI Context

## Metadata

- PR: this PR (alternative to #3921)
- Branch: `facu/reset-workflow-to-initial-step-on-re-presentation`
- Author / human owner: @facumenzella
- Agent(s): Claude Code, Claude Opus 5 (1M context); review by Codex via `codex-review-fix-loop`
- Generated: 2026-08-10
- Context document version: 2

## Goal

Review #3921 ("multi-page paywall position does not reset after dismissal"), then, on request, build an
alternative fix for the same bug so the two approaches could be compared.

## Initial Prompt

Review https://github.com/RevenueCat/purchases-android/pull/3921.

- Why can the ViewModel outlive an embedded or dialog presentation? Led to tracing `ViewModelStoreOwner`
  scoping and `PaywallOptions.hashCode` key stability.
- Supplied the original bug report; asked whether #3921 makes sense. Confirmed the root cause.
- "can you draft your fix differently? so we compare them?" Led to this branch.
- "how can we prove this fixes it? did you write a test?" Led to the composition-level RED→GREEN test,
  after the first two tests turned out to be unable to fail on the unfixed tree.
- "do you know when the regession was introduced? open a draft straight to main."
- "run /codex-review-fix-loop" Led to six review passes.

## Agent Contribution

- Reviewed #3921, reproduced its two `PaywallViewModelTest` failures, and baselined against `4980371c5`
  to confirm they were PR-introduced.
- Traced the root cause to the `_state` / `_workflowState` desync in `updateWorkflowStepState`.
- Traced the origin to #3381/#3416, with #3596 changing the symptom's shape.
- Implemented this alternative and 12 tests.
- Ran six adversarial review passes and applied the findings below.

## Human Decisions

- Directed the work from review toward an alternative implementation.
- Required RED→GREEN proof, which changed the test strategy from ViewModel-level to composition-level.
- Chose to open as a draft against `main`, and to hold the push until the review loop converged.
- Review of the final diff: `Not captured`.

- Decision: replay via `startWorkflowPresentation` from a dismiss-time snapshot.
  - Rationale: it already resets the navigator, state store, step cache and trace id, with no network.
  - Rejected: `updateState()` (#3921's approach), which costs three fetches per re-present.
- Decision: leave `_state` untouched at the dismiss boundary.
  - Rationale: the host owns the exit animation.
  - Rejected: `_state = Loading`, which caused #3921's two test failures via a null presentation
    fingerprint.
- Decision: keep the hook in `LaunchedEffect`, not `remember`.
  - Rationale: it emits `workflow_step_started`; an abandoned composition would emit phantom events.
  - Rejected: `remember`, which was tried in an earlier commit and reverted for this reason.
- Decision: scope the replay snapshot by the `OfferingSelection` that produced the workflow.
  - Rationale: the options hash covers only the offering identifier, so retargeting can go unnoticed.

## Findings applied during review (6 passes, a defect on every pass)

1. `cancelStateUpdate()` at dismiss stranded a paywall dismissed during its initial load on the skeleton
   forever. Found independently by two lenses. Self-inflicted earlier in the same session.
2. Purchase-dismiss and restore-dismiss re-presentation were untested.
3. The composition test pinned only eventual state, not the absence of a `Loading` frame.
4. The no-refetch test missed `awaitGetUiConfig` and `awaitOfferings`.
5. A double dismiss erased the replay snapshot.
6. `paywallPresentationData` was cleared inline in `closePaywall`, so the two auto-dismiss paths never
   cleared it, letting a replay skip its impression.
7. The single-impression test passed with the hook removed; strengthened with `workflowId`/`traceId`.
8. The replay snapshot was not scoped to the current target.
9. The first version of the retarget test passed with its guard disabled.
10. The guard captured the selection at dismiss instead of at presentation start.
11. The retarget branch left the stale step rendered during the reload window.
12. The selection was captured before an early return, so a malformed workflow mismatched the pair.

## Files / Symbols Touched

- `ui/revenuecatui/.../data/PaywallViewModel.kt` — `onPaywallPresented`, `endPresentationSession`,
  `DismissedWorkflowPresentation`, `currentWorkflowOfferingSelection`, `closePaywall`, `updateState`,
  `startWorkflowPresentation`, `clearWorkflowState`.
- `ui/revenuecatui/.../InternalPaywall.kt` — fires the hook on composition entry.
- `ui/revenuecatui/.../LoadingPaywall.kt` — no-op override.
- `ui/revenuecatui/.../testdata/TestData.kt` — no-op override on `MockViewModel`.
- `ui/revenuecatui/src/test/.../InternalPaywallRePresentationTest.kt` — new, composition-level coverage.
- `ui/revenuecatui/src/test/.../data/PaywallViewModelWorkflowTest.kt` — ViewModel-level coverage.
- `ui/revenuecatui/src/test/.../testfixtures/TwoStepWorkflowFixture.kt` — new, shared fixture.

- None. No new dependencies, build config, feature flags, or public API. `onPaywallPresented` is on the
  `internal` `PaywallViewModel` interface; `scripts/api-check.sh` passes.

## Validation

- `./gradlew :ui:revenuecatui:testDefaultsDebugUnitTest`
  - `4980371c5` (base): `1679 tests completed, 1 failed`
  - `e37f770fc` (#3921): `1680 tests completed, 3 failed`
  - this branch: `1692 tests completed, 1 failed`
  - The constant failure is `PaywallComponentsTemplatePreviewRecorder > initializationError`, a local
    Paparazzi environment issue present on the base commit.
- `InternalPaywallRePresentationTest` on `4980371c5`: FAILED. On this branch and on #3921: PASSED. So
  #3921 also fixes the reported bug.
- Six behaviors verified RED by reverting their production change: the stranded load, the double
  dismiss, the impression reset (which also took down two pre-existing tests), the single-impression
  attribution, the retarget guard, and the presentation-start capture.
- `./gradlew detektAll`: BUILD SUCCESSFUL. `./scripts/api-check.sh`: BUILD SUCCESSFUL.
- Manual verification: **Not run.** No device or emulator run against a real dashboard workflow.
- CI: `Not captured` at time of writing.

## Validation Gaps

- No manual/device verification. The one host shape where this matters is an embedded or dialog paywall,
  not `PaywallActivity`, which clears its ViewModelStore and cannot reproduce the bug at all.
- No Paparazzi snapshot run; the pre-existing local failure masks any new snapshot issue.
- Open finding 1 (suspended-load retarget) is unfixed and untested.
- Open finding 4: the `LaunchedEffect` placement is unpinned in either direction.
- The regression-origin claim comes from reading history at those commits, not from a bisect.

## Review Focus

- Is pinning to the dismiss-time workflow snapshot acceptable given checkpoint rules (#3907, #3914)?
- Open finding 1: worth fixing here, or a follow-up?
- Six review passes each found a real defect, three of them in the previous pass's own fix. All were in
  snapshot lifetime management, a class of bug #3921's re-derive-from-source design does not have. Is the
  instant, analytics-preserving re-present worth that surface?
- `_purchaseCompleted` is still not reset at the session boundary. Pre-existing.

## Non-goals / Out of Scope

- Changing how the ViewModel is retained or keyed.
- Any change to #3921.
- iOS parity work beyond the reported symptom.

## Omitted Context

- Raw transcript, unrelated exploration, sensitive details, repetitive attempts, and
  chain-of-thought-style content were omitted.

</details>


facumenzella:
closing in favor of #3921.

PR #3929, opened 2026-08-10
fix: only log unfetched products once

fire-at-will:
### Motivation
When fetching products, the SDK queries the stores twice: once for subscription products, and again for one-time purchases. The SDK logs warning messages when stores don't return all expected products. This is a good log to keep, but when querying for a one-time product, the SDK will log the "unfetched product" log when the subscription product query returns nothing, even when the one-time purchase query will be successful.

This creates confusion for humans/agents reading through the SDK logs, as the product isn't truly unfetched yet, it just hasn't been fetched yet.

### Description
This PR updates how the SDK generates "product not fetched" logs so that it only logs it once per product request. Since the logs contain store-specific information, they must be generated in each stores' `BillingWrapper` implementation. So, to accomodate this, we've added a `logUnfetchedProducts` param to `BillingAbstract.queryProductDetailsAsync()`, and `queryProductDetailsAsync()` consumers specify whether or not they'd like the unfetched logs to be generated.

### Testing
Added unit tests to confirm whether the `logUnfetchedProducts` param works as intended, and to check that the `queryProductDetailsAsync()` consumers are passing in the proper values.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to logging timing and slightly more efficient product-type chaining; product fetch semantics and billing APIs stay the same aside from an optional parameter defaulting to true.
> 
> **Overview**
> Adds a **`logUnfetchedProducts`** flag through **`BillingAbstract.queryProductDetailsAsync()`** and each store billing wrapper (Google, Amazon, Galaxy, simulated). Store handlers only emit “missing/unfetched product” warnings when that flag is **true**, so multi-step fetches do not warn prematurely.
> 
> **`PurchasesOrchestrator.getProductsOfTypes`** now passes **`logUnfetchedProducts = false`** on intermediate product-type queries and **`true`** only on the last query in the chain. It also stops querying the next product type when there are no remaining IDs, and only queries INAPP for IDs not already returned from SUBS. Diagnostics **`notFound`** tracking uses the remaining requested ID set at completion instead of recomputing from collected products.
> 
> **`OfferingsFactory`** uses the same pattern: SUBS query with logging disabled, INAPP follow-up with logging enabled when needed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 01df168afe204cd86a69ba9273d8f7e6360cb700. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -1437,9 +1437,11 @@ internal class PurchasesOrchestrator(
         val type = typesRemaining.firstOrNull()?.also { typesRemaining.remove(it) }
 
         type?.let {
+            val isFinalQuery = typesRemaining.isEmpty()
             billing.queryProductDetailsAsync(
                 productType = it,
                 productIds = productIds,
+                logUnfetchedProducts = isFinalQuery,
                 onReceive = { storeProducts ->
                     dispatch {
                         getProductsOfTypes(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -1438,7 +1438,7 @@ internal class PurchasesOrchestrator(
         val typesRemaining = types.toMutableSet()
         val type = typesRemaining.firstOrNull()?.also { typesRemaining.remove(it) }
 
-        type?.let { productType ->
+        type?.takeIf { productIds.isNotEmpty() }?.let { productType ->
             val isFinalQuery = typesRemaining.isEmpty()
             billing.queryProductDetailsAsync(
                 productType = productType,
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
@@ -1428,24 +1428,26 @@ internal class PurchasesOrchestrator(
         startTime: Date? = null,
         callback: GetStoreProductsCallback,
     ) {
+        val collectedProductIds = collectedStoreProducts.map { it.purchasingData.productId }
+        val requestedProductIds = productIds + collectedProductIds
         val nonNullStartTime = startTime ?: run {
-            trackGetProductsStarted(productIds)
+            trackGetProductsStarted(requestedProductIds)
             dateProvider.now
         }
 
         val typesRemaining = types.toMutableSet()
         val type = typesRemaining.firstOrNull()?.also { typesRemaining.remove(it) }
 
-        type?.let {
+        type?.let { productType ->
             val isFinalQuery = typesRemaining.isEmpty()
             billing.queryProductDetailsAsync(
-                productType = it,
+                productType = productType,
                 productIds = productIds,
                 logUnfetchedProducts = isFinalQuery,
                 onReceive = { storeProducts ->
                     dispatch {
                         getProductsOfTypes(
-                            productIds,
+                            productIds - storeProducts.map { it.purchasingData.productId }.toSet(),
                             typesRemaining,
                             collectedStoreProducts + storeProducts,
                             nonNullStartTime,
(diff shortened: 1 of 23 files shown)


fire-at-will:
@cursor review

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1437,9 +1437,11 @@ internal class PurchasesOrchestrator(
         val type = typesRemaining.firstOrNull()?.also { typesRemaining.remove(it) }
 
         type?.let {
+            val isFinalQuery = typesRemaining.isEmpty()
             billing.queryProductDetailsAsync(
                 productType = it,
                 productIds = productIds,
+                logUnfetchedProducts = isFinalQuery,
```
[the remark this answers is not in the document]
fire-at-will: Addressed in 8dedd01b75c6ed7afc718d006a295e97d4e4bbb7

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1428,22 +1428,26 @@ internal class PurchasesOrchestrator(
...
             billing.queryProductDetailsAsync(
-                productType = it,
+                productType = productType,
                 productIds = productIds,
+                logUnfetchedProducts = isFinalQuery,
                 onReceive = { storeProducts ->
                     dispatch {
                         getProductsOfTypes(
-                            productIds,
+                            productIds - storeProducts.map { it.purchasingData.productId }.toSet(),
```
fire-at-will, line 1450: This is a behavioral change: follow-up queries only query for un-fetched products. This is probably more efficient on the store side, since they'll be fetching fewer products, but it also prevents us from logging "unfetched" log messages for logs that were fetched in a previous query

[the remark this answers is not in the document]
fire-at-will: Fixed in fe81a58eb958cef27c85f1366f80553e22dc658e

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -1428,22 +1428,26 @@ internal class PurchasesOrchestrator(
...
         val type = typesRemaining.firstOrNull()?.also { typesRemaining.remove(it) }
 
-        type?.let {
+        type?.takeIf { productIds.isNotEmpty() }?.let { productType ->
+            val isFinalQuery = typesRemaining.isEmpty()
             billing.queryProductDetailsAsync(
-                productType = it,
+                productType = productType,
                 productIds = productIds,
+                logUnfetchedProducts = isFinalQuery,
```
tonidero, line 1446: Hmm so I'm wondering if it would be cleaner to just log missing products after all the calls are done, where we then know if there are any missing products after fetching all types... Then, we don't need to be passing the logging logic to each call... Wdyt?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Indeed! I think this makes sense even if we change what I mentioned in my previous comment. I guess it could be a problem if the ids could be duplicated for different types... But I don't *think* that's possible?

tonidero (commented):
I'm just wondering if there is a cleaner way... Wdyt? Happy to chat about it of course! Also, great improvement 🙌

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
fire-at-will: @tonidero  Yeah, I like the idea! It doesn't feel great to have params in our functions just for logging logic. It would also allow us to standardize the logging logic across stores, which is a big plus. The reason I didn't go with that approach is that could lose store-specific logging info, like in this log from the Play Store:

`Product not found: [redacted product ID]- Product Type: subs, Reason: PRODUCT_NOT_FOUND, Serialized doc ID:`

Looking at it though, that reason may not actually be that helpful, and the Galaxy/Amazon stores don't actually include any store-specific info. If you're in favor, I'll refactor this to remove the bool params and centralize the logging outside of the stores' billing wrappers

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
fire-at-will: Yeah, I double checked and it's not possible for the Galaxy, Amazon, or Play Stores, so it shouldn't be a problem today, but could theoretically be possible in the future for other stores? I doubt it though, since by definition, a product identifier should uniquely identify a product

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: Yeah... I think that's fine TBH... maybe if there is a reason other than `PRODUCT_NOT_FOUND` it could be useful? In case there is some useful reason why it doesn't show I guess... But yeah, I feel it's cleaner to move most of the logic outside...

fire-at-will: Cool, thanks! I will refactor this to remove the bool params and place the logging outside of the BillingWrappers :)

PR #3930, opened 2026-08-10
fix: don't share a MessageDigest across coroutines in FontLoader and DefaultFileCache

bilqisium:
### Issue

Crash affecting [this customer](https://revenuecat.zendesk.com/agent/tickets/82655).

`FontLoader` and `DefaultFileCache` each held a single shared `MessageDigest`.

MessageDigest is not thread-safe, and both classes call it from concurrent coroutines, so the digest's internal state gets corrupted and throws ultimately killing the app.

### The fix

Replaced both private helpers with one internal fun `ByteArray.md5Hex()` in `common/utils.kt`, next to the existing sha1()/sha256(), which already build the digest per call. It reuses the existing `ByteArray.toHexString()`.

Added tests as well.

---

> [!NOTE]
> **Medium Risk**
> Touches concurrent download/cache paths for paywall fonts and file prefetching; the fix is localized and digest output should match, but these are customer-visible failure paths that previously crashed the app.
> 
> **Overview**
> Fixes a **production crash** when paywall fonts or cached remote files are hashed from **multiple coroutines at once**. `FontLoader` and `DefaultFileCache` each kept one lazy `MessageDigest` and called `digest()` concurrently; `MessageDigest` is not thread-safe, which corrupted state and could throw (e.g. `ArrayIndexOutOfBoundsException`).
> 
> The PR **centralizes MD5-as-hex** in a new internal `ByteArray.md5Hex()` in `common/utils.kt`, matching the existing per-call pattern used by `sha1()` / `sha256()`, and reuses `toHexString()`. Both classes drop their shared digest and private helpers and call the extension for URL hashing and font integrity checks.
> 
> **Tests** in `HashUtilsTest` cover expected digests (including leading-zero padding) and a multi-threaded regression that would fail with a shared instance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03bb3c5d2b4e7b35429018468247845060ffa46e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/utils.kt
@@ -10,6 +10,7 @@ import android.content.pm.PackageManager
 import android.os.Build
 import android.util.Base64
 import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.models.toHexString
 import java.security.MessageDigest
 import java.util.Locale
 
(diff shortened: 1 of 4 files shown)


MonikaMateska (approved):
LGTM 👌

PR #3933, opened 2026-08-11
fix: avoid OutOfMemoryError parsing large offerings responses

AlvaroBrey:
Fetching offerings parsed the entire response into an `org.json` tree. Paywall components are the bulk of that response, and `OfferingParser` only ever needed them as text to hand to kotlinx.serialization, so most of that tree was built just to be serialized straight back to a string and thrown away. Building it is what ran out of memory.

- **Keep components out of the tree.** The response text is scanned once, each `paywall_components` object is replaced by `null`, and only the much smaller remainder is parsed. Components are kept as offsets into the payload and sliced out as text on demand, so the round-trip through `JSONObject.toString()` is gone.
- **Stop re-parsing the disk copy.** When a fetch fails and parsed offerings are already in memory, the fallback reuses them instead of rebuilding them from the cached response. Its most common trigger is a background refresh whose callbacks are both null, so that work was being discarded anyway.
- **This lowers the ceiling, it does not remove it.** The payload text and the extracted component text are still held, and SharedPreferences still retains a copy of the response. Stripping components server-side (khepri#21713) is still the real fix.
- **Behavior change**, offline-with-warm-memory path only: store products are no longer re-queried, so prices are not refreshed and offerings whose products all disappeared are no longer dropped. If the cache went stale because the device locale changed, offerings stay on the old locale until the cache window expires.
- **Behavior change:** malformed JSON inside a components object no longer fails the whole response; it surfaces when those components are decoded, leaving the other offerings usable.
- Part of #3867, which should stay open until the server-side strip lands.

### Checklist

### How the scan works

It walks the response text tracking strings and escapes, so braces inside paywall copy (mustache variables like `{{ product.price_per_month }}`) are not mistaken for object delimiters. For each offering it records the start and end offset of that offering's `paywall_components` object, then builds the text to parse by copying everything except those spans, writing `null` in their place. `RawPaywallComponents` holds the offsets and produces the text lazily, so `OfferingParser` decodes from the original payload rather than from a re-serialized tree.

Anything the scanner does not recognize throws internally and falls back to a plain full parse, so unrecognized payloads behave exactly as before. Duplicate `offerings` or `paywall_components` keys take that same path deliberately: `org.json` keeps only the last value, so continuing would desynchronize the index-parallel component list and could attach one offering's paywall to another.

### Notes for reviewers

- `Backend` now hands the offerings success callback the payload text rather than `result.body`, so nothing reads `HTTPResult.body` on this path. Coalesced `getOfferings` callbacks therefore each run the scan where the `by lazy` previously parsed once; that is still far cheaper than the old full tree, and each caller now gets its own tree instead of sharing a mutable one.
- `componentsHash` is now derived from the raw payload text rather than `org.json`'s reserialization, so its values change. It is `private`, used only for in-process `equals`/`hashCode`, and never persisted or transmitted; the disk cache stores the payload verbatim, so cached-vs-network comparisons still hash identically.
- The in-memory fallback only takes effect while the cache generation still matches. A mismatch means that copy was parsed with components skipped, so it falls through to the disk copy rather than re-caching a stale parse.
- `TemplatePreviews.kt` binds `createOfferings(JSONObject, Map)` reflectively, so the `@JvmOverloads` two-argument bridge has to keep working; breaking it fails Paparazzi at runtime with no compile error.
- `baseline-prof.txt` needed its `getOfferings` descriptor updated for the narrowed callback arity.

### Regression gates

- `OfferingsResponseParserMemoryTest` gates allocations of the parse, so losing the allocation-free skip path fails the build.
- `OfferingsResponseParserTest` covers braces and escapes inside strings, nested structures, index alignment, duplicate and escaped keys, and the malformed-escape deferral.
- `OfferingsManagerTest` covers the generation guard, the workflow gate, and that the in-memory fallback neither reads the disk copy nor re-enters the factory.
- `OfferingsTest` covers elided-vs-un-elided equivalence; every pre-existing `OfferingParser` test used the un-elided overload, so the production path had no coverage.

### Known, not addressed here

The generation check and the cache write in the fallback are not atomic, so an invalidating clear landing between them can re-cache offerings the kill switch meant to drop. The same check-then-write exists on `main` in `createAndCacheOfferings` with a wider window, so this is pre-existing in kind; fixing it properly means making validation and write indivisible across both sites.

`ui_config` and v1 `PaywallData` still round-trip through `toString()`. Both are eagerly decoded, so slicing them would save only the intermediate string and not the decode. Recorded on SDK-4433.

</details>

### DIFF
--- a/purchases/src/main/baseline-prof.txt
+++ b/purchases/src/main/baseline-prof.txt
@@ -650,7 +650,7 @@ HSPLcom/revenuecat/purchases/common/Backend;->addCallback(Ljava/util/Map;Lcom/re
 HSPLcom/revenuecat/purchases/common/Backend;->getCallbacks()Ljava/util/Map;
 HSPLcom/revenuecat/purchases/common/Backend;->getCustomerInfo(Ljava/lang/String;ZLkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
 PLcom/revenuecat/purchases/common/Backend;->getDiagnosticsCallbacks()Ljava/util/Map;
-HSPLcom/revenuecat/purchases/common/Backend;->getOfferings(Ljava/lang/String;ZLkotlin/jvm/functions/Function3;Lkotlin/jvm/functions/Function2;)V
+HSPLcom/revenuecat/purchases/common/Backend;->getOfferings(Ljava/lang/String;ZLkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;)V
 HSPLcom/revenuecat/purchases/common/Backend;->getOfferingsCallbacks()Ljava/util/Map;
 HSPLcom/revenuecat/purchases/common/Backend;->getPaywallEventsCallbacks()Ljava/util/Map;
 PLcom/revenuecat/purchases/common/Backend;->getProductEntitlementCallbacks()Ljava/util/Map;
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt
@@ -65,7 +65,7 @@ internal typealias CallbackCacheKey = List<String>
 
 /** @suppress */
 @OptIn(InternalRevenueCatAPI::class)
-internal typealias OfferingsSuccessCallback = (JSONObject, HTTPResponseOriginalSource, responsePayload: String) -> Unit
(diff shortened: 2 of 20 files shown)


AlvaroBrey:
@cursor review

AlvaroBrey:
dropping, too complex to be worth it right now just as offering paywalls are being deprecated in favor of workflows/config endpoint

PR #3934, opened 2026-08-11
build: update Kotlin to 2.2.21 and Poko to 0.20.2

AlvaroBrey:
Part of the AGP 9 upgrade pre-work. Bumps Kotlin 2.0.21 to 2.2.21 and Poko 0.17.2 to 0.20.2.

- Kotlin lands on **2.2.21**, not 2.2.10: Poko 0.20.0 dropped support for Kotlin 2.2.0 and 2.2.10 "due to incompatibility", and 0.20.2 is compiled against 2.2.21. Poko's old IR plugin in turn crashes under any Kotlin 2.2 compiler, so both versions have to move together and 2.2.21 is the pairing upstream supports.
- **Consumer-facing caveat, and the reason this wants a major release**: the published POM carries `kotlin-stdlib` at the compiler version in `compile` scope, so consumers on an older Kotlin fail with `Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 2.2.0, expected version is 2.0.0`. This effectively raises the minimum Kotlin a consumer needs.
- `kotlinLanguage` stays at `1.8`, but note that only sets our own `languageVersion`/`apiVersion`. It does not govern the stdlib we publish, so it does not keep the above caveat away.
- `test-apps/sdksizetesting` moves to the same Kotlin version. It consumes the locally published artifacts, so it is a worked example of the consumer break described above rather than an unrelated tweak.
- Drops a stdlib pin in `dokka-hide-internal` that existed only because the repo's Kotlin was older than Dokka's analysis Kotlin; they now match.
- Stacked on #3961. The `api-tester` workaround this PR used to carry is gone, because #3961 exports `compose-foundation-layout` from `:ui:revenuecatui` properly.

### Checklist

<details>

### Motivation

Part of the AGP 9 upgrade, split into small independently-shippable PRs. Poko 0.17.2's IR compiler
plugin crashes under the Kotlin 2.2 compiler (`NoSuchMethodError: irEqeqeq`), so the two version
bumps have to land together.

The exact pairing matters. Poko's changelog for 0.20.0 says "Compile with Kotlin 2.2.20. Drop
support for Kotlin 2.2.0 and 2.2.10 due to incompatibility", and 0.20.2 is compiled with 2.2.21.
An earlier revision of this PR paired Kotlin 2.2.10 with Poko 0.20.2, which built and tested green
but is an unsupported combination upstream. Poko is an IR plugin, so the failure modes are a loud
`NoSuchMethodError` or silently wrong generated `equals`/`hashCode`/`toString`, across 111 `@Poko`
types in published modules. Moving Kotlin to 2.2.21 puts us on the combination Poko actually
targets.

The Kotlin **compiler** version and the Kotlin **language/API level** (`kotlinLanguage`) are
separate knobs in this build (`ConfigureAndroidLibrary.kt`): every library module sets
`languageVersion`/`apiVersion` from `kotlinLanguage`, not from the compiler version. This PR bumps
only the compiler and leaves `kotlinLanguage = "1.8"`.

That does **not** make the change invisible to consumers, and an earlier revision of this
description wrongly claimed it did. The Kotlin Gradle Plugin publishes `kotlin-stdlib` at the
compiler version in `compile` scope, Gradle resolves the highest version, and an older consumer
compiler then cannot read that stdlib's metadata. `sdksizetesting` hit exactly this and needed its
own Kotlin bumped to keep consuming the SDK. The practical minimum Kotlin for consumers therefore
moves with this PR, which is being tracked for the release that carries it.

### Description

- Kotlin 2.0.21 to 2.2.21, Poko 0.17.2 to 0.20.2.
- `test-apps/sdksizetesting` Kotlin bumped to match, so it can still read the SDK's metadata.
- `dokka-hide-internal` stdlib pin removed.

### Ordering note

This PR was originally stacked on #3935 (Metalava 0.4.0-alpha03 to 0.5.0), which has since merged.
Metalava was split out because it stands on its own, but the order mattered and both ways were
tested directly. With Metalava still at 0.4.0-alpha03, bumping Kotlin alone produces a small but
real `api*.txt` diff: Metalava cannot parse Kotlin 2.2's metadata format and silently drops the
`throws` clause it synthesizes from `@Throws` on at least one method. The compiled bytecode was
confirmed unaffected via `javap`, but the tracked signature file loses it. With Metalava bumped
first, this combination produces zero `api*.txt` diff.

### Testing

- `scripts/api-check.sh` clean with zero signature diff.
- `:purchases` unit tests green, `detektAll` clean.
- Compiled every module applying the Poko plugin (`:purchases`, `:ui:revenuecatui`,
  `:feature:admob`, `:feature:galaxy`) to confirm the IR-plugin crash does not reproduce.
- `:api-tester` compiles with no Compose layout dependency of its own, which is the end to end
  check that #3961 underneath this branch does its job.

</details>


api-tester/build.gradle.kts
```diff
@@ -48,6 +48,7 @@ dependencies {
 
     implementation(platform(libs.compose.bom))
     implementation(libs.compose.ui)
+    implementation(libs.compose.foundation.layout)
```
tonidero: Hmm I wonder if the need of doing this should be considered a breaking change... I guess it would be mostly fine, as long as folks were mostly able to make it work without it... But I wonder if we should include/expose this from our ui module...

tonidero (approved):
Nice!! 🙇 Thank you thank you!

AlvaroBrey:
Waiting for https://github.com/RevenueCat/purchases-android/pull/3947 merge (which fixes maestro broken on this branch)

api-tester/build.gradle.kts
AlvaroBrey: This is not a breaking change (not anymore than it already was, I mean), it hasn't hit us for the same reasons as: https://github.com/RevenueCat/purchases-android/pull/3916, basically everyone who ships an android nowadays transitively includes `compose.foundation.layout` anyway.

But we _should_ expose it in the UI module to make it explicit, separate PR incoming.

AlvaroBrey: https://github.com/RevenueCat/purchases-android/pull/3961

AlvaroBrey:
Further testing reveals we actually need kotlin 2.3+ for proper AGP + our supporting plugins compatibility, so this intermediate step is uesless

PR #3936, opened 2026-08-11
Ingest audiences config topic

vegaro:
[WFL-451](https://linear.app/revenuecat/issue/WFL-451/android-ingest-the-audiences-config-topic) adds the app-level `audiences` topic so upcoming SDK rule evaluation can read audience data; rule evaluation itself stays out of scope.

iOS counterpart https://github.com/RevenueCat/purchases-ios/pull/7390

---

> [!NOTE]
> **Low Risk**
> Read-only ingestion on the existing remote-config stack; no auth, purchase, or evaluation logic changes.
> 
> **Overview**
> Adds support for the backend **`audiences`** remote-config topic so future SDK rule evaluation can load audience definitions from synced config.
> 
> **`RemoteConfigTopic.Audiences`** is registered with wire name `audiences`, matching other topic providers. **`AudiencesConfigProvider`** exposes `getAudience(identifier)`, which returns the audience blob as an opaque **`JsonObject`** via **`RemoteConfigManager.blobData`** (or `null` for missing, malformed, or non-object payloads).
> 
> Tests cover wire naming, full JSON round-trip, null handling, and an integration path where prefetched audience items with and without blobs are persisted and read independently off disk.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c857cbb934148514f088ab7982d7539231506098. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/AudiencesConfigProvider.kt
```diff
@@ -0,0 +1,12 @@
+package com.revenuecat.purchases.common.audiences
+
+import com.revenuecat.purchases.common.remoteconfig.RemoteConfigManager
+import com.revenuecat.purchases.common.remoteconfig.RemoteConfigTopic
+import kotlinx.serialization.json.JsonObject
+
+internal class AudiencesConfigProvider(
+    private val manager: RemoteConfigManager,
+) {
+    suspend fun getAudience(identifier: String): JsonObject? =
```
tonidero, line 10: I've been thinking whether we actually need to parse this at all as a json, or if we could just return the raw string... Are we sending anything more other than the json logic string? If so, it could be a performance improvement... But I guess that depends on the backend format and we can iterate on it later.

tonidero (approved):
Looks good! Mostly just a question but nothing blocking.

purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/AudiencesConfigProvider.kt
vegaro: a json string would be interesting yeah. Let's see what the backend sends and we decide

PR #3937, opened 2026-08-11
build: update Paparazzi to 2.0.0-alpha05

AlvaroBrey:
Part of the AGP 9 upgrade pre-work. Paparazzi alpha02 cannot run on Gradle 9 in two separate ways, so it has to move before the AGP/Gradle bump.

<details>

### Motivation

Part of the AGP 9 upgrade, split into small independently-shippable PRs. Paparazzi is a hard
prerequisite for Gradle 9, for the two reasons in the summary above.

**The AGP-9 support claim needed checking, because the sources conflict.** The alpha05 changelog
says *"This release supports pre-AGP 9.0 consumers"*, which reads like AGP 9 is unsupported, and
the PR carrying the AGP-9 API migration (cashapp/paparazzi#2106, "[DNM] AGP 9.0 API Changes") was
closed **unmerged**. Resolved it against the `2.0.0-alpha05` tag's actual source instead:

- `PaparazziPlugin.kt` imports `com.android.build.api.dsl.CommonExtension` and contains **no
  `BaseExtension` reference at all**, so the migration did land, just via a different change.
- It contains `isAgpAtLeast(major = 9)`, used to *enable* ASM instrumentation on AGP 9+: an
  AGP-9-aware feature branch, not a version cap.
- Issue #2095 is `closed / completed`, milestoned `2.0.0-alpha05`.

So alpha05 supports AGP 9, and the changelog line means it *also* retains pre-AGP-9 support.

**On the golden images.** alpha05 bumps LayoutLib 16.1.1 → 16.2.1 and Compose to 1.11.2, which can
legitimately shift rendering, so the two committed `ui/debugview` PNGs were the main risk. What the
measurements showed:

| | `notConfigured` | `configured` |
|---|---|---|
| alpha02 (baseline, local macOS) | 0.047221% | 0.455492% |
| alpha05 (local macOS) | 0.055340% | 0.467417% |

Both already fail locally on **unmodified `main`** by the first row's margins, while
`ci/circleci: test_defaults_debug` is green on `main` (that job runs `testDefaultsDebugUnitTest`
across all modules, which includes `DebugViewSnapshotTest`). So the goldens match CI's Linux
rendering and simply do not match macOS. alpha05 adds roughly +0.008% / +0.012% on top of that
pre-existing gap, which is the same order as the environment noise itself.

Re-recording on macOS would therefore replace CI-valid images with locally-valid ones and break the
job that currently passes. Leaving them untouched; CI is the authority here, and if alpha05's
rendering change does cross the threshold on Linux, `test_defaults_debug` on this PR will say so.

### Description

- `gradle/libs.versions.toml`: `paparrazzi` (sic) `2.0.0-alpha02` → `2.0.0-alpha05`, and removed the
  now-false `# Can't use alpha04 because it requires higher Kotlin version than we're using`
  comment, which this PR's Kotlin dependency resolves.

Also checked while here, no changes needed:

- Paparazzi task names are unchanged between alpha02 and alpha05, and
  `fastlane/Fastfile`'s `<module>:recordPaparazziDebug` (which has no flavor, while both modules
  carry the `apis` dimension) resolves correctly to `recordPaparazziDefaultsDebug` via Gradle's
  camel-case abbreviation. Confirmed with `--dry-run`.
- `fastlane/Fastfile` carries a commented-out `paparazzi.snapshot.dir` property noting it is
  unsupported in the current version. Left alone; not required here and worth its own look.

Tested: build configures cleanly on Kotlin 2.2.10 with no stdlib conflict;
`:ui:revenuecatui:testDefaultsDebugUnitTest` 1678/1679 pass, the single failure being the
pre-existing `PaywallComponentsTemplatePreviewRecorder` error from an uninitialized
`upstream/paywall-preview-resources` submodule in this environment; `detektAll` clean;
`:ui:debugview:recordPaparazziDefaultsDebug` still succeeds, so the path CI actually uses works
(re-recorded output was reverted, goldens are byte-identical to `main`).

</details>


AlvaroBrey:
See https://github.com/RevenueCat/purchases-android/pull/3934#issuecomment-5280425956

PR #3940, opened 2026-08-12
`OutOfMemoryError` caching large offerings responses (`DeviceCache.cacheOfferingsResponse`, `ETagPayloadStore.read`)

tsushanth:
Switch offerings response cache from SharedPreferences to file-based storage (avoiding permanent in-memory map pressure on low-heap devices), and add a 10 MB size guard in ETagPayloadStore.read to prevent OOM when allocating ByteArray + String for oversized cached payloads.

Found via automated repo scanning, fix written and reviewed before opening.

---

> [!NOTE]
> **Medium Risk**
> Changes persistence of offerings responses and ETag cache read behavior, which can affect cold-start/offline offerings availability if file I/O fails. Mitigated by SharedPreferences fallback and treating oversized ETag payloads as cache misses.
> 
> **Overview**
> Fixes **OutOfMemoryError** on low-heap devices when caching multi-MB offerings/ETag responses.
> 
> **Offerings cache** now writes to a file under `cacheDir/RevenueCat/offerings_response` instead of SharedPreferences, so the large payload is no longer kept in the in-memory prefs map. Reads fall back to the legacy prefs entry on upgrade, and successful file writes remove that entry.
> 
> **ETag payload reads** reject files larger than **10 MB** (instead of `Int.MAX_VALUE`), treating them as cache misses so oversized payloads trigger a fresh network request rather than allocating a huge `ByteArray` + `String`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ff183ac79b6bee3ac2628bd0a7620d602ff3213. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -65,6 +65,7 @@ import com.revenuecat.purchases.strings.Emojis
 import com.revenuecat.purchases.subscriberattributes.SubscriberAttributesManager
 import com.revenuecat.purchases.subscriberattributes.SubscriberAttributesPoster
 import com.revenuecat.purchases.subscriberattributes.caching.SubscriberAttributesCache
+import java.io.File
 import com.revenuecat.purchases.utils.CoilImageDownloader
 import com.revenuecat.purchases.utils.DefaultUrlConnectionFactory
 import com.revenuecat.purchases.utils.EventsFileHelper
@@ -206,7 +207,11 @@ internal class PurchasesFactory(
             }
             val signingManager = SigningManager(signatureVerificationMode, appConfig, apiKey)
 
-            val cache = DeviceCache(prefs, apiKey)
+            val offeringsResponseFile = File(
+                File(contextForStorage.cacheDir, "RevenueCat"),
+                "offerings_response",
+            )
+            val cache = DeviceCache(prefs, apiKey, offeringsResponseFile = offeringsResponseFile)
 
             val localeProvider = DefaultLocaleProvider()
 
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
@@ -37,6 +37,9 @@ import kotlinx.serialization.builtins.serializer
 import kotlinx.serialization.json.Json
 import org.json.JSONException
 import org.json.JSONObject
+import java.io.File
+import java.io.FileOutputStream
+import java.io.IOException
 import java.util.Date
 import kotlin.time.Duration.Companion.hours
 
@@ -60,6 +63,7 @@ public open class DeviceCache(
     private val preferences: SharedPreferences,
     private val apiKey: String,
     private val dateProvider: DateProvider = DefaultDateProvider(),
+    private val offeringsResponseFile: File? = null,
 ) : StorefrontProvider {
     private companion object {
         private const val CUSTOMER_INFO_SCHEMA_VERSION_KEY = "schema_version"
@@ -590,15 +594,38 @@ public open class DeviceCache(
 
     @Synchronized
     internal fun getOfferingsResponseCache(): JSONObject? {
+        val file = offeringsResponseFile
+        if (file != null && file.exists()) {
+            return try {
+                JSONObject(file.readText(Charsets.UTF_8))
+            } catch (e: Exception) {
+                null
+            }
+        }
+        // Fallback: read from SharedPreferences for users upgrading from a previous version.
         return getJSONObjectOrNull(offeringsResponseCacheKey)
     }
 
     /**
-     * Serializing a [JSONObject] here instead of storing [offeringsResponse] as received would cost a
-     * contiguous allocation of several times the response size, which OOMs on low-heap devices.
+     * Persists [offeringsResponse] to a file rather than SharedPreferences so the payload is not
+     * loaded into the in-memory preferences map on every app start, which OOMs on low-heap devices
+     * when the response is several megabytes.
      */
     @Synchronized
     internal fun cacheOfferingsResponse(offeringsResponse: String) {
+        val file = offeringsResponseFile
+        if (file != null) {
+            try {
+                file.parentFile?.mkdirs()
+                // Stream through OutputStreamWriter so no full-payload ByteArray is allocated.
+                FileOutputStream(file).writer(Charsets.UTF_8).use { it.write(offeringsResponse) }
+                // Clean up any legacy SharedPreferences entry left by a previous version.
+                preferences.edit().remove(offeringsResponseCacheKey).apply()
+            } catch (e: IOException) {
+                errorLog(e) { "Failed to write offerings response to disk." }
+            }
+            return
+        }
         preferences.edit()
             .putString(offeringsResponseCacheKey, offeringsResponse)
             .apply()
@@ -606,6 +633,7 @@ public open class DeviceCache(
 
     @Synchronized
     internal fun clearOfferingsResponseCache() {
+        offeringsResponseFile?.delete()
         preferences.edit().remove(offeringsResponseCacheKey).apply()
     }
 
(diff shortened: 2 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -206,7 +207,11 @@ internal class PurchasesFactory(
             }
             val signingManager = SigningManager(signatureVerificationMode, appConfig, apiKey)
 
-            val cache = DeviceCache(prefs, apiKey)
+            val offeringsResponseFile = File(
+                File(contextForStorage.cacheDir, "RevenueCat"),
```
tonidero, line 211: I wonder if we should use the `noBackupFilesDir` instead, so we avoid it being cleared by the system.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -590,22 +594,46 @@ public open class DeviceCache(
...
-     * Serializing a [JSONObject] here instead of storing [offeringsResponse] as received would cost a
-     * contiguous allocation of several times the response size, which OOMs on low-heap devices.
+     * Persists [offeringsResponse] to a file rather than SharedPreferences so the payload is not
+     * loaded into the in-memory preferences map on every app start, which OOMs on low-heap devices
+     * when the response is several megabytes.
      */
     @Synchronized
     internal fun cacheOfferingsResponse(offeringsResponse: String) {
+        val file = offeringsResponseFile
+        if (file != null) {
```
tonidero, line 617: Should this be `file == null`? Since we need to create the file?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Also, maybe it can be created within the `DeviceCache` itself with a default parameter, but keep the parameter there for testing.

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -590,22 +594,46 @@ public open class DeviceCache(
 
     @Synchronized
     internal fun getOfferingsResponseCache(): JSONObject? {
+        val file = offeringsResponseFile
+        if (file != null && file.exists()) {
```
tonidero, line 598: We should add tests both for the caching + fetching.

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -169,5 +171,8 @@ internal class ETagPayloadStore(
         const val TRASH_SUFFIX = ".trash"
         const val CHUNK_CHARS = 64 * 1024
         const val WRITE_BUFFER_BYTES = 256 * 1024
+        // A read allocates ~2x this in heap (ByteArray + String). Cap at 10 MB so the spike is at
+        // most ~20 MB; larger payloads fall back to a fresh network request instead of OOMing.
```
tonidero, line 175: I think this might not help much, since, even if the response comes from network, it would be similar size right?

tonidero (commented):
Thanks for the contribution!! 🙇 We're looking into some backend fixes that should also help with these repeating OOM we've been having that should improve the situation a lot. Still, I think the change to use a file different than shared preferences make sense! Thanks!

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -590,22 +594,46 @@ public open class DeviceCache(
...
+            try {
+                file.parentFile?.mkdirs()
+                // Stream through OutputStreamWriter so no full-payload ByteArray is allocated.
+                FileOutputStream(file).writer(Charsets.UTF_8).use { it.write(offeringsResponse) }
+                // Clean up any legacy SharedPreferences entry left by a previous version.
+                preferences.edit().remove(offeringsResponseCacheKey).apply()
+            } catch (e: IOException) {
+                errorLog(e) { "Failed to write offerings response to disk." }
+            }
+            return
```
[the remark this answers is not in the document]
tonidero: This is also valid. We probably should use something like `AtomicFile` to make sure we don't leave the file in an inconsistent state

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
AlvaroBrey: > I wonder if we should use the noBackupFilesDir instead, so we avoid it being cleared by the system.

It's a cache, it _should_ be cleared by the system if the user requests it or under storage pressure. Bypassing it is an antipattern IMO

purchases/src/main/kotlin/com/revenuecat/purchases/common/networking/ETagPayloadStore.kt
```diff
@@ -74,8 +74,10 @@ internal class ETagPayloadStore(
         return try {
             FileInputStream(fileFor(urlString)).use { input ->
                 val sizeBytes = input.channel.size()
-                // Not an integrity check: ByteArray would throw past the IOException catch below.
-                if (sizeBytes > Int.MAX_VALUE) return null
+                // Guards against OOM on low-heap devices: a ByteArray this large, plus the String
+                // built from it, would exhaust the heap. Oversized files fall back to a fresh
+                // network request. MAX_PAYLOAD_BYTES < Int.MAX_VALUE so the cast below is safe.
+                if (sizeBytes > MAX_PAYLOAD_BYTES) return null
```
AlvaroBrey, line 80: If we're to adopt this, we need to  guard the write path too, otherwise we're always reading from network but still writing to disk.

Also this is a significant enough behaviour that I'd want logs for it

However, as @tonidero said [here](https://github.com/RevenueCat/purchases-android/pull/3940#pullrequestreview-4913257140), we're actually on the process of serving these differently so that we don't get the entire dataset in a single response, so I wonder if this cap is even needed in the mid-term.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Right… good point… FWIW, most things we have are caches that could be refetched. I guess the idea being that we should keep anything that helps the app work offline if needed. In this case, this could be helpful in that scenario. But not super important. I'm ok keeping this as is 👍

AlvaroBrey: Actually you make a good point too, this is not actually just a cache in the typical sense, it's an offline datasource... so maybe we do indeed want to store it in a persistent dir.  for `EtagPayloadStore` I used `cacheDir` but that's different, that requires network to begin with

AlvaroBrey (changes requested):

purchases/src/main/kotlin/com/revenuecat/purchases/common/caching/DeviceCache.kt
```diff
@@ -590,22 +594,46 @@ public open class DeviceCache(
...
+        val file = offeringsResponseFile
+        if (file != null && file.exists()) {
+            return try {
+                JSONObject(file.readText(Charsets.UTF_8))
+            } catch (e: Exception) {
+                null
+            }
+        }
+        // Fallback: read from SharedPreferences for users upgrading from a previous version.
         return getJSONObjectOrNull(offeringsResponseCacheKey)
```
AlvaroBrey, line 606: The main benefit of this change is to _not_ have the data loaded in memory for the entire process lifetime, so IMO it isn't worth it without also getting rid of the prefs allocation if pre-existing

That is to say, this change needs to include some migration to either delete or move to disk existing cached sharedpref data, _and_ to stop reading from sharedprefs in the normal paths

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -206,7 +207,11 @@ internal class PurchasesFactory(
...
+                "offerings_response",
+            )
```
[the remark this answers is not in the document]
AlvaroBrey: This is accurate too

PR #3942, opened 2026-08-12
refactor(checkpoints): Local rule evaluation foundation

tonidero:
### Description

Starts introducing the infrastructure for rule evaluation in the SDK and providing dimensions for the evaluation of rules.

One behavior change to mention is that we will now pick the first rule with an audience that matches, instead of the first rule with a servable workflow, which I think is desired and matches iOS.

No audience providers are added yet. This will be done in follow-up PRs.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint workflow selection semantics (no fallthrough when the matched rule is unservable), which can alter which paywall/workflow is shown. Dimension providers are not live yet, so runtime impact is limited until follow-ups.
> 
> **Overview**
> Adds a **local rules evaluation** layer and wires checkpoints through it, so rule selection is based on audience match rather than “first servable workflow.”
> 
> Introduces `common/localrules` with `LocalRulesEvaluator`, dimension providers/resolver, and a `RulesEngine` logger bridge. Checkpoints now pick the **first audience-matching rule**; if that winner’s workflow is unservable, resolution returns `CONFIGURATION_UNAVAILABLE` instead of falling through. No dimension providers are registered yet, and predicates are still the placeholder `"true"`, so published order remains the effective signal until follow-ups land.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4336fae5ccc5d8c26176d6c1f50319dec908440b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3943, opened 2026-08-12
refactor(checkpoints): Add DeviceDimensionProvider

tonidero:
### Description

Adds the first dimension provider, so rule predicates have something to read: `device.appVersion`, `device.locale`, `device.platform`, `device.platformVersion` and `device.sdkVersion`.

---

> [!NOTE]
> **Medium Risk**
> Enables real checkpoint audience matching on device/locale/platform fields; camelCase dimension keys are a new predicate contract, though behavior is additive with tests covering provider and platform mapping.
> 
> **Overview**
> **Local rules can now read device context.** `PurchasesFactory` wires `DeviceDimensionProvider` into `LocalRulesEvaluator` (replacing an empty provider list), so checkpoint predicates can target `device.appVersion`, `device.locale`, `device.platform`, `device.platformVersion`, and `device.sdkVersion`.
> 
> `DeviceDimensionProvider` snapshots most values at configure time but **re-reads locale** on each evaluation (normalized to lowercase `en_us`), omits empty string dimensions, and maps Amazon vs other stores via shared `Store.platformName`. That same extension replaces `HTTPClient`'s private `X-Platform` mapping so headers and rules stay aligned.
> 
> Dimension key documentation and resolver tests move from **snake_case to camelCase** (`appVersion` vs `app_version`) to match the nested `device.*` paths the engine resolves.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d52917764ec5bf1b8d1fc0eb864fb189777fff62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
@@ -1,0 +1,74 @@
+@file:OptIn(InternalRevenueCatAPI::class)
+
+package com.revenuecat.purchases.common.localrules
+
+import android.os.Build
+import com.revenuecat.purchases.InternalRevenueCatAPI
+import com.revenuecat.purchases.common.AppConfig
+import com.revenuecat.purchases.common.Config
+import com.revenuecat.purchases.common.LocaleProvider
+import com.revenuecat.purchases.common.platformName
+import java.util.Date
+
+/**
+ * Device and environment dimensions.
+ *
+ * Everything but the locale is fixed for the life of the process and is read once. The locale is read per
+ * evaluation because the system locale can change mid-session and `preferredUILocaleOverride` can be set at any
+ * time, and an audience keyed on locale has to agree with what the customer will actually be shown.
+ *
+ * A dimension the device has no value for is omitted; see [hasValue].
+ */
+internal class DeviceDimensionProvider(
+    appConfig: AppConfig,
+    private val localeProvider: LocaleProvider,
+) : RulesDimensionProvider {
+
+    override val identifier: String = "device"
+
+    override val namespace: RulesDimensionNamespace = RulesDimensionNamespace.Device
+
+    private val fixedDimensions: Map<String, RulesDimensionValue> = mapOf(
+        KEY_APP_VERSION to RulesDimensionValue.StringValue(appConfig.versionName),
+        KEY_PLATFORM to RulesDimensionValue.StringValue(appConfig.store.platformName),
+        KEY_PLATFORM_VERSION to RulesDimensionValue.IntValue(Build.VERSION.SDK_INT.toLong()),
+        KEY_SDK_VERSION to RulesDimensionValue.StringValue(Config.frameworkVersion),
+    )
+
+    // Snapshotted at configure time, so it is the fallback for the rare case where the live preference list is
+    // empty.
+    private val fallbackLanguageTag: String = appConfig.languageTag
+
+    override suspend fun dimensions(date: Date): Map<String, RulesDimensionValue> =
+        (fixedDimensions + (KEY_LOCALE to RulesDimensionValue.StringValue(currentLocale())))
+            .filterValues { value -> value.hasValue }
+
+    /**
+     * [LocaleProvider.currentLocalesLanguageTags] is the comma-joined preference list sent as a header. A
+     * dimension has to be a scalar to be usable with `==` or `in`, so only the preferred locale is exposed.
+     *
+     * Normalized to lowercase snake case (`en_us`) so one predicate can target a locale on every platform,
+     * whatever casing and separator that platform's locale API happens to use.
+     */
+    private fun currentLocale(): String =
+        localeProvider.currentLocalesLanguageTags
+            .substringBefore(',')
+            .ifEmpty { fallbackLanguageTag }
+            .lowercase()
+            .replace(oldChar = '-', newChar = '_')
+
+    internal companion object {
+        const val KEY_APP_VERSION = "appVersion"
+        const val KEY_LOCALE = "locale"
+        const val KEY_PLATFORM = "platform"
+        const val KEY_PLATFORM_VERSION = "platformVersion"
+        const val KEY_SDK_VERSION = "sdkVersion"
+
+        /**
+         * A dimension the device could not tell us about is omitted rather than reported as an empty string: an
+         * absent key resolves to null in the engine, while `""` would compare equal to another absent value.
+         */
+        private val RulesDimensionValue.hasValue: Boolean
+            get() = this !is RulesDimensionValue.StringValue || value.isNotEmpty()
+    }
+}
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -29,6 +29,7 @@ import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.events.BackendStoredEvent
 import com.revenuecat.purchases.common.events.EventsManager
 import com.revenuecat.purchases.common.isDeviceProtectedStorageCompat
+import com.revenuecat.purchases.common.localrules.DeviceDimensionProvider
 import com.revenuecat.purchases.common.localrules.LocalRulesEvaluator
 import com.revenuecat.purchases.common.localrules.RulesEngineLoggerBridge
 import com.revenuecat.purchases.common.log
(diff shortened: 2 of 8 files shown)


tonidero:
* **#3960** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3960?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3949** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3949?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3943** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3943?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3943?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    override suspend fun dimensions(date: Date): Map<String, RulesDimensionValue> =
+        fixedDimensions + (KEY_LOCALE to RulesDimensionValue.StringValue(currentLanguageTag()))
+
+    /**
+     * [LocaleProvider.currentLocalesLanguageTags] is the comma-joined preference list sent as a header. A
+     * dimension has to be a scalar to be usable with `==` or `in`, so only the preferred locale is exposed.
+     */
+    private fun currentLanguageTag(): String =
+        localeProvider.currentLocalesLanguageTags.substringBefore(',').ifEmpty { fallbackLanguageTag }
```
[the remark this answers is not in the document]
rickvdl: Yes we should standardize this, I opted for `en_us` (lowercased snake_case) for easier matching in rules and less error-prone. But happy to discuss

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+internal class DeviceDimensionProvider(
+    appConfig: AppConfig,
+    private val localeProvider: LocaleProvider,
+) : RulesDimensionProvider {
+
+    override val identifier: String = "device"
+
+    override val namespace: RulesDimensionNamespace = RulesDimensionNamespace.Device
+
+    private val fixedDimensions: Map<String, RulesDimensionValue> = mapOf(
```
rickvdl, line 31: Could any of these values return empty strings? If so we should probably map them to null (omit them)

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+        KEY_APP_VERSION to RulesDimensionValue.StringValue(appConfig.versionName),
+        KEY_PLATFORM to RulesDimensionValue.StringValue(appConfig.store.platformName),
+        KEY_PLATFORM_VERSION to RulesDimensionValue.IntValue(Build.VERSION.SDK_INT.toLong()),
```
rickvdl, line 34: On iOS this is a semver string, let's discuss how we should handle this..

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+        KEY_SDK_VERSION to RulesDimensionValue.StringValue(Config.frameworkVersion),
```
rickvdl, line 35: On iOS I explicitly did not include the `-SNAPSHOT` prefix here for example, should we?

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+    private val localeProvider: LocaleProvider,
+) : RulesDimensionProvider {
+
+    override val identifier: String = "device"
+
+    override val namespace: RulesDimensionNamespace = RulesDimensionNamespace.Device
+
+    private val fixedDimensions: Map<String, RulesDimensionValue> = mapOf(
+        KEY_APP_VERSION to RulesDimensionValue.StringValue(appConfig.versionName),
+        KEY_PLATFORM to RulesDimensionValue.StringValue(appConfig.store.platformName),
```
rickvdl, line 33: Hm would `amazon` as a platform name derived from the store make sense? I think platform == OS in our reasoning

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: ~~Good question! None of these are nullable so it shouldn't happen... Some of them do default to empty string though in some cases, like the locale or the app version...  Not sure if we would like to clear those out in those scenarios... But yeah, probably something to sync about with the audiences folks.~~ Sorry I read nullable not sure why 🤦. So yeah, a couple of these could be empty strings... I'm ok changing them to remove them when empty strings for now I guess yeah... probably a better outcome 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: Right... TBH, I was wondering... I think having it be a number is easier to perform comparison rules, for example "android versions newer than X" is easier with a number than a semver string... Still I agree we should discuss this with audiences team! 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: Yeah... well, in Android there are many versions of OS's. Amazon might be one of the more distinct ones (and it might change soon with Vega). I think however it makes sense we send the same we send to the backend, which is the store platform name. Note that for non-amazon stores, we always default to `android`, which includes galaxy, test store,....

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: Hmmm IMO, we should just use the version as it is... If we remove the snapshot, it might not be a version that's actually published or even that will be published necessarily I guess (if we end up doing a major). But still could be another question we should ask audiences team!

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
rickvdl: Ah right, I figured we'd always report android and separately report the store. In this case I think this makes sense, but I think indeed we may need another property for the store then?

tonidero: yeah true... we will indeed need a separate property for that with a different logic (with values that should match the backend I guess)

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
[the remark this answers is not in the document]
tonidero: I think that makes sense! Will do the same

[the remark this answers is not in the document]
tonidero: Changed!

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: Done!

rickvdl (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: As discussed, will merge this as is, and we can consolidate in follow-up PRs

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/DeviceDimensionProvider.kt
tonidero: As discussed, will merge this as is, and we can consolidate in follow-up PRs

PR #3944, opened 2026-08-12
Fix multi-page paywall state after dismissal

vegaro:
Supersedes #3921.

Multi-page paywalls could reopen on the last rendered workflow step because the retained `PaywallViewModel` outlived the dismissed presentation. This PR resets presentation state on dismissal and reload from the initial step when the paywall enters composition again.

Ideally we use `rememberViewModelStoreProvider`, but it requires updating `lifecycle-viewmodel-compose` to `2.11.0` which requires AGP 9+ and we are blocked on that.

---

> [!NOTE]
> **Medium Risk**
> Touches paywall dismiss/re-present lifecycle, workflow navigation state, and analytics session boundaries; behavior is heavily tested but affects core paywall UX paths.
> 
> **Overview**
> Fixes **embedded/multi-step workflow paywalls reopening on a later step** when the same retained `PaywallViewModel` is shown again after dismiss.
> 
> `InternalPaywall` now wires **`onPaywallPresented` / `onPaywallDismissed`** via `DisposableEffect` so presentation boundaries follow Compose composition (after exit animations), not only `closePaywall()`. On dismiss, **`endPresentationSession()`** clears workflow/session state and sets a flag to **reload on the next presentation**, while **keeping the last loaded UI** until the paywall leaves composition. Re-entry triggers **`updateState()`** from the initial workflow step; impression tracking and async refreshes are **blocked during teardown** so stale steps and events are not applied. Purchase/restore auto-dismiss paths use the same session end logic.
> 
> Adds a **Maestro workflow regression** and **unit/instrumentation tests** for Compose, `PaywallActivity`, `PaywallView`, and workflow ViewModel re-presentation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 926652b89282dc321900edfe17d7b47646216c7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


maestro/workflows/workflow_reopens_at_initial_step.yaml
```diff
@@ -0,0 +1,54 @@
+# Reopening an embedded workflow paywall after dismissing it from a later page must start a new
+# presentation on the initial page instead of retaining the previous presentation's page state.
+# Run on CI by the `run-maestro-workflow-tests` job.
+#
+# Run locally against the test store:
+#   maestro test -e MAESTRO_STORE=test_store maestro/workflows/workflow_reopens_at_initial_step.yaml
+
+appId: com.revenuecat.e2etests
+name: Reopening a workflow paywall starts at the initial page
```
tonidero, line 9: ❤️

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallViewModel.kt
```diff
@@ -396,6 +403,18 @@ internal class PaywallViewModelImpl(
...
+        cancelStateUpdate()
+        paywallPresentationData = null
+        standaloneStateStore = null
+        clearWorkflowState()
+        // This ViewModel can outlive an embedded or dialog presentation. Do not leave the last rendered
+        // workflow step in the generic state, and reload from the initial step when the paywall enters
+        // composition again.
+        _state.value = PaywallState.Loading
+        shouldReloadStateOnNextPresentation = true
+    }
```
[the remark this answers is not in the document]
tonidero: I was going to say this same thing 😅 We should make sure we only set it to loading after the paywall is completely dismissed. And we should test this both with the Compose/View APIs and the activity API.

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
```diff
@@ -120,6 +120,10 @@ internal fun InternalPaywall(
     options: PaywallOptions,
     viewModel: PaywallViewModel = getPaywallViewModel(options),
 ) {
+    LaunchedEffect(viewModel) {
```
tonidero: Hmm we should make sure this isn't called again on configuration changes (like device rotation... I guess it could just cause a state reset so maybe it works, but it would be updated without need?

tonidero (commented):
Just a couple comments. I think the only blocker would be the one about possibly seeing a flash of the loading screen when dismissing... Looks good otherwise!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/InternalPaywall.kt
vegaro: I added a test for this. It's safe because `shouldReloadStateOnNextPresentation` would be false and `onPaywallDismissed` and `onPaywallPresented` won't do anything

JZDesign (approved):
🔥 Looks good to me, thanks for running with it!

tonidero (approved):
I think this looks great!

PR #3949, opened 2026-08-12
feat(checkpoints): Evaluate checkpoint custom variables as custom.* dimensions

tonidero:
### Description

Makes a checkpoint's custom variables readable from a rule predicate as `custom.<key>`, so the same checkpoint can behave differently depending on how it was reached. 

Similar to iOS (RevenueCat/purchases-ios#7393)

---

> [!NOTE]
> **Medium Risk**
> Changes which checkpoint workflow wins when rules use `custom.*` predicates, and renames an internal API (`customProperties` → `customVariables`) that callers must adopt.
> 
> **Overview**
> Renames checkpoint **`customProperties`** to **`customVariables`** on `CheckpointParams`, `Purchases.resolveCheckpoint`, and the resolver stack so naming matches paywalls and iOS.
> 
> Checkpoint resolution now feeds those per-call values into **local rule evaluation** as **`custom.<key>`** dimensions (alongside `device.*`). Invalid keys and unsupported value types are dropped with warnings instead of failing the whole evaluation. Shared **`CustomVariableKeyValidator`** lives in the purchases module and replaces the UI-only copy; paywall and checkpoint paths both use it.
> 
> **`LocalRulesEvaluator.match`** and **`RulesDimensionResolver.snapshot`** accept optional custom variables; an empty map leaves the `custom` namespace absent (not an empty object). **`CheckpointWorkflowResolverImpl`** passes converted variables when matching checkpoint audience rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c5c00a46e9ac424b42fb83825922ce0f6b7d9af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3960** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3960?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3949** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3943** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
```diff
@@ -41,9 +43,11 @@ internal class CheckpointWorkflowResolverImpl(
     private val checkpointsConfigProvider: CheckpointsConfigProvider?,
     private val localRulesEvaluator: LocalRulesEvaluator,
     private val getOfferings: suspend () -> Offerings,
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    private val audiencePredicate: String = PLACEHOLDER_AUDIENCE_PREDICATE,
```
tonidero, line 47: This won't actually be here in the final version, just moved for easier testing

rickvdl (commented):
I think this looks great! I do however see one discrepancy with iOS.

As of https://github.com/RevenueCat/purchases-ios/pull/7400 I've updated the implementation to reuse the `CustomVariable` type from the Paywalls implementation. I think we should do the same on Android, and that also showed that we currently accept both Int and Float values for checkpoints, but in the current custom variable implementation for Paywalls we seem to only accept a number type which is internally preserved as a double. 

I think we should consolidate this. For now this means that we'll only support the 'number' initializer and integers would internally be preserved as Doubles on iOS. But if we plan on adding more types we should do the same for Paywalls I'd say. Wdyt?

tonidero:
> I think this looks great! I do however see one discrepancy with iOS.
> 
> As of [RevenueCat/purchases-ios#7400](https://github.com/RevenueCat/purchases-ios/pull/7400) I've updated the implementation to reuse the `CustomVariable` type from the Paywalls implementation. I think we should do the same on Android, and that also showed that we currently accept both Int and Float values for checkpoints, but in the current custom variable implementation for Paywalls we seem to only accept a number type which is internally preserved as a double.
> 
> I think we should consolidate this. For now this means that we'll only support the 'number' initializer and integers would internally be preserved as Doubles on iOS. But if we plan on adding more types we should do the same for Paywalls I'd say. Wdyt?

Makes sense, and nice catch! I will do the same in Android yeah. For now we can consolidate to Doubles, and we can iterate if needed later

tonidero:
@rickvdl I moved the changes for ^ comment to a follow-up PR: https://github.com/RevenueCat/purchases-android/pull/3960. Lmk what you think! 🙇

rickvdl (approved):
Nice!

vegaro (approved):

PR #3951, opened 2026-08-12
Support headless offering checkpoints

vegaro:
Some checkpoint workflows only select an offering and leave presentation to the app. This inspects the matched workflow and returns `CheckpointResult.ReceivedOffering` when the workflow consists of a single valid offering step.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint resolution and purchase/paywall entry paths (new branch, renamed resolution types) with broad test coverage; behavior for invalid configs is preserved but resolution order and dependency loading differ.
> 
> **Overview**
> Checkpoints can now resolve to **app-owned offerings** instead of always launching RevenueCat UI. When the winning workflow is a single `offering` step with a valid `offering_identifier`, resolution returns **`CheckpointResolution.MatchedOffering`** and the UI layer surfaces **`CheckpointResult.ReceivedOffering`**—no Activity or one-at-a-time presentation slot.
> 
> **UI workflows** are unchanged in behavior but renamed to **`MatchedWorkflow`** (from `Workflow`). Resolution loads the workflow body first via **`WorkflowManager.getWorkflowBody`**, branches on step type, and only fetches workflow→offering mapping, offerings, and asset prewarm for full UI paths. Invalid shapes (mixed steps, bad identifiers, etc.) still fail the matched rule with **`CONFIGURATION_UNAVAILABLE`** without falling through to later rules.
> 
> The checkpoint tester, API tester, and tests are updated for the new result type and offering use case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6cc7ec7f9e6e5fa364115d876294f10dbe201459. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
```diff
@@ -77,39 +82,98 @@ internal class CheckpointWorkflowResolverImpl(
...
+        }
+        if (step.screenId != null) {
+            return unsupportedOfferingRule(rule, "the offering step contains a screen")
+        }
+        if (step.triggers.isNotEmpty()) {
+            return unsupportedOfferingRule(rule, "the offering step contains triggers")
+        }
+        if (step.triggerActions.isNotEmpty()) {
+            return unsupportedOfferingRule(rule, "the offering step contains trigger actions")
+        }
```
vegaro: all this validation could be removed, open for feedback

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointResolution.kt
```diff
@@ -12,11 +12,14 @@ import com.revenuecat.purchases.common.workflows.PublishedWorkflow
 @InternalRevenueCatAPI
 public sealed class CheckpointResolution {
 
+    /** The checkpoint selected [offering] without requiring RevenueCat-managed UI. */
+    public data class Offering(val offering: RevenueCatOffering) : CheckpointResolution()
```
tonidero: Hmm I understand this is meant to remain internal and not a public API, but don't fully like naming it `Offering` with the name collisions it could bring... Maybe keep the `ReceivedOffering` naming of the public API... or maybe `MatchedOffering`, and rename `Workflow` with `MatchedWorkflow`? 

Not sure TBH... Maybe it's ok since usually, when calling from outside, you would access this with `CheckpointResolution.Offering` so there shouldn't be conflicts outside...

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
```diff
@@ -77,39 +82,98 @@ internal class CheckpointWorkflowResolverImpl(
...
-        val offering: Offering = offerings.all[offeringId]
-            ?: return unservableRule(rule, "offering '$offeringId' was not found in offerings")
         val workflow = try {
-            workflowManager.getWorkflow(rule.workflowId)
+            workflowManager.getPublishedWorkflow(rule.workflowId)
         } catch (e: PurchasesException) {
             return unservableRule(rule, "it could not be loaded: ${e.error}")
         }
+        val offeringSteps = workflow.steps.values.filter { it.type == OFFERING_STEP_TYPE }
+        return if (offeringSteps.isEmpty()) {
```
tonidero: Hmm I'm not sure if whether there are offering steps should determine how we process this... Maybe we should just look at the type of the first step for now, and add some validation for now that if the first step is UI, all steps should be UI, or if the first step is an offering step, it should be a single step. Then we can discuss how to support multistep workflows in follow-ups?

tonidero (approved):
Amazing! I think this looks good. Left a couple comments, but I think we can always iterate over those in follow-up PRs if needed.

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointResolution.kt
vegaro: I feel you I went back and forth here as well as soon as I had to do that alias. Left it because of the Workflow being there, but I might just go with MatchedOffering and MatchedWorkflow

tonidero: Yeah... not a blocker in any case since this will remain internal.

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
```diff
@@ -18,14 +18,19 @@ import com.revenuecat.purchases.common.localrules.LocalRule
...
 
 /**
  * Resolves a checkpoint through the `checkpoint_rules` topic: the checkpoint's rules are read from remote config
  * and evaluated in order against locally collected dimensions, and the first rule whose audience matches wins.
  *
- * The winner is final. If its workflow turns out to be unservable — no offering configured for it, that offering
- * absent from the fetched offerings, or its body unavailable — the checkpoint resolves to
+ * UI config remains a prerequisite for every matched checkpoint workflow. After it is loaded, a terminal offering
+ * step returns its offering to the app without presentation, while a UI workflow keeps the existing presentation
+ * path. The winner is final. If its workflow turns out to be unservable, the checkpoint resolves to
```
facumenzella: I find this confusing. It says "prerequisite for every matched checkpoint workflow", but the offering case is the one thing that isn't a workflow presentation.

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
```diff
@@ -84,14 +79,33 @@ internal class WorkflowManager(
...
         }
-        return workflow
     }
 
+    /**
+     * Loads a published workflow body by its workflow id without resolving offering aliases, loading UI config,
+     * or prewarming presentation assets. Checkpoint resolution uses this to inspect the workflow kind before it
+     * decides which additional dependencies, if any, are needed to serve it.
+     */
+    suspend fun getPublishedWorkflow(workflowId: String): PublishedWorkflow =
```
facumenzella: `getWorkflow` and `getPublishedWorkflow` both return a `PublishedWorkflow`, so the names don't say which one loads `ui_config` and prewarms. worth renaming this one to smth like `getWorkflowBody`?

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
vegaro: let me remove this comment. Iniitally I wasn't requiring UI config for the nonUI (checkppoint returning an offering) case. But I think I was overcomplicating things so removed it and this comment is from that discussion.

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
vegaro: yeah makes sense. Let's keep it simple

vegaro: done in a48adad1c9c39a16297ea31623bcfb3070041177

purchases/src/main/kotlin/com/revenuecat/purchases/common/workflows/WorkflowManager.kt
vegaro: done in [1c579c7](https://github.com/RevenueCat/purchases-android/pull/3951/commits/1c579c7faf1002f427007dbe8ffbd48c9429f08d)

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointResolution.kt
```diff
@@ -1,9 +1,9 @@
 package com.revenuecat.purchases.checkpoints
 
 import com.revenuecat.purchases.InternalRevenueCatAPI
-import com.revenuecat.purchases.Offering
 import com.revenuecat.purchases.UiConfig
 import com.revenuecat.purchases.common.workflows.PublishedWorkflow
+import com.revenuecat.purchases.Offering as RevenueCatOffering
```
facumenzella: Does this alias still make sense?

vegaro: I think we can remove it now that we renamed to `MatchedOffering`

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
```diff
@@ -87,40 +86,89 @@ internal class CheckpointWorkflowResolverImpl(
...
+            return unservableRule(rule, "offering '$offeringIdentifier' was not found in offerings")
+        }
+        debugLog {
+            "Checkpoint resolved to offering '${offering.identifier}' from workflow '${rule.workflowId}'"
+        }
+        return CheckpointResolution.MatchedOffering(offering)
+    }
+
+    private fun unsupportedOfferingRule(rule: CheckpointRule, reason: String): CheckpointResolution.NoAction =
+        unservableRule(rule, reason)
```
facumenzella: this `unsupportedOfferingRule` is kind of uselsss right 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointResolution.kt
vegaro: done in [bf5a551](https://github.com/RevenueCat/purchases-android/pull/3951/commits/bf5a5519f71a04a0d0509a2fcb81452e7515332c)

purchases/src/main/kotlin/com/revenuecat/purchases/checkpoints/CheckpointWorkflowResolverImpl.kt
```diff
@@ -155,6 +203,73 @@ internal class CheckpointWorkflowResolverImpl(
...
     }
 }
+
+private class CheckpointResolutionDependencies(
+    private val identifier: String,
+    private val workflowManager: WorkflowManager,
+    private val uiConfigProvider: UiConfigProvider?,
+    private val getOfferings: suspend () -> Offerings,
+) {
+    private var offeringsLoaded = false
```
facumenzella: I don't see much value in these `xLoaded` flags tbh. What am I missing?

vegaro: I actually don't think we need `CheckpointResolutionDependencies` at all anymore. This is leftovers from a previous iteration. Good catch!

facumenzella (approved):

PR #3960, opened 2026-08-13
refactor(checkpoints): Reuse CustomVariableValue for checkpoint custom variables

tonidero:
### Description

`CheckpointParams` now takes `Map<String, CustomVariableValue>` instead of `Map<String, Any?>`, so the compiler enforces the value types that were previously checked (and dropped) at runtime.

Android equivalent of RevenueCat/purchases-ios#7400. 

---

> [!NOTE]
> **Medium Risk**
> Breaking change to `@InternalRevenueCatAPI` checkpoint APIs affects how custom variables reach audience matching and paywalls; behavior for invalid keys is unchanged but callers must migrate from raw primitives.
> 
> **Overview**
> **Checkpoint custom variables are now typed as `CustomVariableValue` instead of `Map<String, Any?>`.** Call sites must wrap values in `CustomVariableValue.String`, `.Number`, and `.Boolean`, so invalid types are caught at compile time instead of dropped at runtime.
> 
> `CheckpointParams` validates keys once via `CustomVariableKeyValidator` and exposes the same map for **targeting rules** (`custom.<key>`) and **paywall rendering**. The UI layer maps each value to `RulesDimensionValue` when calling `Purchases.resolveCheckpoint`; the purchases module no longer converts from loose `Any` values (`customVariableDimensions` removed).
> 
> `RulesDimensionValue` is extracted to a public `@InternalRevenueCatAPI` type in `:purchases`, and `CustomVariableValue` gains an internal `map` / `asRulesDimensionValue` bridge. Examples, api-tester, and tests are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 14de80708eaa7f2646aa1e9e4cd809c6145f3eb7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tonidero:
* **#3960** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3960?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3960?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3949** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3949?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3943** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3943?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -109,6 +109,5 @@ private val RulesDimensionValue.asRulesEngineValue: Value
     get() = when (this) {
         is RulesDimensionValue.StringValue -> Value.StringValue(value)
         is RulesDimensionValue.BoolValue -> Value.BoolValue(value)
-        is RulesDimensionValue.IntValue -> Value.IntValue(value)
```
rickvdl, line 112: I think this may still be used by other audience rules right? This is supported in the rules engine, even if we may not support it for the custom variables

vegaro (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointParams.kt
```diff
@@ -36,9 +35,17 @@ public class CheckpointParams(
...
-internal val CheckpointParams.paywallCustomVariables: Map<String, CustomVariableValue>
-    get() = customVariables.mapValues { (_, value) -> CustomVariableValue.from(value) }
+internal val CustomVariableValue.asRulesDimensionValue: RulesDimensionValue
+    get() = when (this) {
+        is CustomVariableValue.String -> RulesDimensionValue.StringValue(value)
+        is CustomVariableValue.Number -> RulesDimensionValue.DoubleValue(value)
+        is CustomVariableValue.Boolean -> RulesDimensionValue.BoolValue(value)
+        // CustomVariableValue is an abstract class with an internal constructor, so this is unreachable, but the
+        // compiler can't know that.
+        else -> error("Unknown CustomVariableValue type: ${this::class.simpleName}")
```
rickvdl: This is true, but adding a new type without updating this list could potentially crash right? In iOS I was able to make it exhaustive, maybe we can do something similar? https://github.com/RevenueCat/purchases-ios/pull/7400/commits/f16df44829cb2d86e1268e6e7d132de05d615b2a

rickvdl (approved):

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointParams.kt
tonidero: Ohh right... yeah makes sense!

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
tonidero: Yeah... It's currently unused, but I do think we might likely want to use it... so I brought it back. Thanks!

ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/checkpoints/CheckpointParams.kt
tonidero: Good catch! I applied a similar logic in Android 🙏

PR #3931, opened 2026-08-11
build: move baseline profiles to the plugin's default source dir

AlvaroBrey:
Part of the AGP 9 upgrade pre-work: removes a config that breaks library release builds under AGP 9.

- `baselineProfileOutputDir = "."` on `:purchases` and `:ui:revenuecatui` made the baseline-profile plugin register `src/main` itself as a profile source dir; AGP 9's `ProcessLibraryArtProfileTask` parses every file there (no `.txt` filter) and crashes on `res/values/styles.xml`.
- Fix: drop the override, let profiles land in the plugin's default `src/main/generated/baselineProfiles/` dir instead, `git mv` the two committed profile files there.
- Updates the two hardcoded old paths in `fastlane/Fastfile` and `.circleci/config.yml`

### Checklist
- [ ] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids

<details>
<summary>Agent description</summary>

### Motivation

Both public library modules set `baselineProfileOutputDir = "."` in their `baselineProfile { }`
block. This makes the `androidx.baselineprofile` plugin register `src/main` (the module root) as
a baseline-profile source directory. Harmless under AGP 8. Under AGP 9,
`ProcessLibraryArtProfileTask` parses *every* file in a registered profile source dir with no
`.txt` filter, and chokes on `res/values/styles.xml`, breaking every library **release** build.

This is one item pulled out of a broader AGP 9 upgrade reconnaissance pass, landed standalone
since it's AGP-8-safe and has no dependency on anything else in that upgrade.

### Description

- `purchases/build.gradle.kts`, `ui/revenuecatui/build.gradle.kts`: removed
  `baselineProfileOutputDir = "."`. The plugin's default,
  `generated/baselineProfiles` under `src/<variant>/`, is used instead (confirmed from
  `androidx.baselineprofile`'s own `BaselineProfileConsumerExtension.kt` source).
- `purchases/src/main/baseline-prof.txt` and `ui/revenuecatui/src/main/baseline-prof.txt` moved
  (`git mv`) to `.../src/main/generated/baselineProfiles/baseline-prof.txt`.
- `fastlane/Fastfile` (`generate_baseline_profiles` lane) and `.circleci/config.yml`
  (`generate-baseline-profiles` job) updated to the new paths; both hardcode the profile file
  locations for the biweekly regeneration lane and would otherwise break on the next scheduled
  run.

[...]

PR #3961, opened 2026-08-13
fix(paywalls): Expose compose-foundation-layout so PaywallFooter is callable without extra setup

AlvaroBrey:
- `PaywallFooter`, `OriginalTemplatePaywallFooter` and `CloseButton` expose `PaddingValues` and `BoxScope` in their signatures, but every Compose dependency was declared `implementation`, so consumers never received `foundation-layout` on their compile classpath.
- Kotlin 2.0.x tolerated the unresolvable type. Kotlin 2.2 rejects it, so consumers building with a recent Kotlin already cannot call those APIs against released versions unless they happen to pull `foundation-layout` in themselves. Most Compose apps do, via `foundation` or `material`, which is why this went unnoticed.
- The Compose BOM is promoted to `api` as well. Without it the published Gradle metadata lists `foundation-layout` with no version in the api variant, so a consumer without their own Compose BOM could not resolve it.
- Headline caveat: that promotion means our BOM now participates in Gradle consumers' Compose version resolution, as a constraint they can still override. Maven consumers already got exactly this through `dependencyManagement`, so this aligns Gradle with existing Maven behavior rather than adding a new constraint.
- Additive for consumers and no public API change: `scripts/api-check.sh` reports no signature diff.
- Measured effect: `foundation-layout` on `api-tester`'s compile classpath goes from absent to resolving at 1.7.2, with `api-tester` declaring nothing.

### Checklist

### Motivation

`:ui:revenuecatui` declared every Compose dependency as `implementation`, with `api(project(":purchases"))`
as the only exported dependency. Three public declarations nevertheless reference
`androidx.compose.foundation.layout` types:

```
OriginalTemplatePaywallFooter(..., mainContent: Function1<PaddingValues,Unit>?)
PaywallFooter(...)  // deprecated, same signature
CloseButton(BoxScope, ...)  // extension receiver
```

A consumer therefore needs `foundation-layout` on its own compile classpath to call them, even though
nothing in the published metadata says so.

Removing the workaround from `api-tester` and compiling reproduces it:

```
PaywallAPI.kt:23:9 Cannot access class 'PaddingValues'.
  PaywallFooter(options = options)
  OriginalTemplatePaywallFooter(options = options)
```

Note it fails on the overloads that pass no lambda at all, because the compiler still resolves the
default parameter's type.

The consumer compiles with their own Kotlin, not ours, so this is not caused by the SDK's toolchain:
anyone on Kotlin 2.2 or newer is already affected by shipped versions. It surfaced here only because
`api-tester` is an unusually minimal consumer, pulling `compose-ui`, `activity-compose` and
`compose-ui-google-fonts` but neither `foundation` nor `material`, both of which would have dragged
`foundation-layout` in transitively.

### Description

- `api(libs.compose.foundation.layout)` in `:ui:revenuecatui`, plus a catalog entry for it.
- `api(platform(libs.compose.bom))`, replacing the `implementation` declaration of the same platform.
  This is required, not incidental. With the BOM left at `implementation`, the generated POM is fine,
  because the BOM is already published as a `dependencyManagement` import and Maven resolves the
  version from it, but the Gradle module metadata api variant comes out as:

  ```
  androidx.compose.foundation:foundation-layout:NO-VERSION
  ```

  with no accompanying platform dependency or constraint. Gradle consumers are the majority here, so
  the POM alone being correct is not enough.

- No source changes and no signature changes.

### Testing

- `foundation-layout` in `:api-tester`'s `defaultsDebugCompileClasspath`: absent before, resolves to
  1.7.2 after, without `api-tester` declaring it.
- Inspected the generated POM and `module.json` before and after. The POM gains
  `foundation-layout` at `compile` scope; the api variant gains both `compose-bom:2024.09.00` and
  `foundation-layout`.
- `assembleDefaultsRelease`, `:api-tester:compileDefaultsDebugKotlin`, `detektAll`, and
  `scripts/api-check.sh` all pass locally.
- End to end: with #3934 stacked on top of this branch, so Kotlin is 2.2.10 and `api-tester` declares
  no Compose layout dependency of its own, `:api-tester:compileDefaultsDebugKotlin` succeeds. That is
  the exact compilation that fails without this change.
- `api-tester` is the lasting regression gate: it exercises these call sites, and under Kotlin 2.2 it
  cannot compile unless this dependency is exported.

### Related

- #3934 (Kotlin 2.2.10) is stacked on this branch. It originally carried an `api-tester` workaround for
  the same problem; that commit is dropped now that the dependency is exported properly, which is also
  the end to end check described above.

</details>

### DIFF
--- a/ui/revenuecatui/build.gradle.kts
+++ b/ui/revenuecatui/build.gradle.kts
@@ -88,8 +88,6 @@ tasks.withType<KotlinCompile>().configureEach {
 
 dependencies {
     api(project(":purchases"))
-    // PaddingValues and BoxScope appear in this module's public API, so the BOM that versions
-    // them has to be exported too: without it the published api variant has no version for them.
     api(platform(libs.compose.bom))
     api(libs.compose.foundation.layout)
 
--- a/ui/revenuecatui/build.gradle.kts
+++ b/ui/revenuecatui/build.gradle.kts
@@ -88,8 +88,11 @@ tasks.withType<KotlinCompile>().configureEach {
 
 dependencies {
     api(project(":purchases"))
+    // PaddingValues and BoxScope appear in this module's public API, so the BOM that versions
+    // them has to be exported too: without it the published api variant has no version for them.
+    api(platform(libs.compose.bom))
+    api(libs.compose.foundation.layout)
 
-    implementation(platform(libs.compose.bom))
     implementation(libs.compose.ui)
     implementation(libs.compose.ui.util)
     implementation(libs.compose.ui.graphics)
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -126,6 +126,7 @@ commonmark-strikethrough = { module = "org.commonmark:commonmark-ext-gfm-striket
 
 compose-bom = { module = "androidx.compose:compose-bom", version.ref = "composeBom" }
 compose-constraintlayout = { module = "androidx.constraintlayout:constraintlayout-compose", version.ref = "constraintlayout" }
+compose-foundation-layout = { module = "androidx.compose.foundation:foundation-layout" }
 compose-material = { module = "androidx.compose.material:material" }
 compose-material3 = { module = "androidx.compose.material3:material3" }
 compose-material3Adaptive = { module = "androidx.compose.material3.adaptive:adaptive" }


ui/revenuecatui/build.gradle.kts
```diff
@@ -88,8 +88,9 @@ tasks.withType<KotlinCompile>().configureEach {
 
 dependencies {
     api(project(":purchases"))
+    api(platform(libs.compose.bom))
+    api(libs.compose.foundation.layout)
```
tonidero, line 92: [Not related to this line]

Hmm so exposing those types in our API was a mistake 😅 This means that we are now tied to the compose API, and if they have breaking changes, we would need to do a major as well theoretically :( 

In my mind the appropriate fix would be to, since we need to do a major. Change the public API of the PaywallFooter types to avoid exposing any compose types + hide (make internal) the `CloseButton` that seems like something we shouldn't have ever exposed. Wdyt @AlvaroBrey @JayShortway?

AlvaroBrey:
* **#3965** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3965?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3963** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3963?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3961** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3961?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3961?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ui/revenuecatui/build.gradle.kts
AlvaroBrey: Don't we expose a bunch of Compose types in the ui module anyway? I don't think these are the only ones by far (and also it's a compose-aware UI lib, we are kinda supposed to?)

tonidero: Well... we do expose annotations like `@Composable` or `@Immutable` and some types like `FontFamily`... It's not great since it means that, if we want to update compose, and that update includes a breaking change, we could potentially be breaking someone that might be using an older version of compose, so not ideal...

But yeah, you do have a good point, and I don't have a good thought on how to workaround those annotations and types we already expose... So with that in mind, I guess I'm ok with this then. 👍

tonidero (approved):

ui/revenuecatui/build.gradle.kts
JayShortway: >if we want to update compose, and that update includes a breaking change, we could potentially be breaking someone that might be using an older version of compose

I think this is true regardless of `implementation`/`api`, because we will pull in the newer compose version and that will win over the older version the app has declared (in a standard setup). 

It's the most correct (tm) to declare Compose as `api` indeed because it is part of our public API. One question: do we need to make the bom `api` too, or is foundation.layout enough?

tonidero: Hmm you're right... so yeah please ignore me 😅 🙇 . Not sure if we need to also `api` the bom TBH... But in the PR description:
```
Without it the published Gradle metadata lists foundation-layout with no version in the api
variant, so a consumer without their own Compose BOM could not resolve it.
```
So it seems it is?

AlvaroBrey: Right, for the reasons above we need to either set the bom as api, or set a explicit version for `foundation-layout`

PR #3945, opened 2026-08-12
build: fix api-tester customEntitlementComputation variant resolution after Kover bump

AlvaroBrey:
- `:api-tester`'s `customEntitlementComputation` flavor now falls back to `defaults` when resolving library modules, which only publish that flavor.
- Fixes `androidDependencies` failing with "No matching variant of project :feature:amazon" (also `:feature:galaxy`, `:ui:debugview`, `:ui:revenuecatui`) since #3932.
- The bug predates the Kover bump: Kover 0.7.0's `koverArtifact` configuration declared no usage/category attributes, so Gradle silently bound those modules to an empty coverage artifact instead of the real library. 0.9.9 gives it a `kover` usage attribute, so the accidental match is gone and resolution fails.

<details>
<summary>Agent description</summary>

### Motivation

`assemble-admob-integration-sample-app` started failing on every branch at the "Download Dependencies"
step (`./gradlew androidDependencies`). First red run on main is the Kover 0.9.9 merge (#3932); the run
immediately before it is green.

`:api-tester` declares the `apis` dimension with both `defaults` and `customEntitlementComputation`.
Every library module except `:purchases` declares the dimension with only `defaults`, and the app's CEC
flavor never declared `matchingFallbacks`, so there has never been a matching variant for
`:feature:amazon`, `:feature:galaxy`, `:ui:debugview` or `:ui:revenuecatui` on the
`customEntitlementComputationDebugCompileClasspath`.

It resolved anyway because Kover 0.7.0's `koverArtifact` consumable configuration declared no
`org.gradle.usage` or `org.gradle.category`, so Gradle scored it compatible on every requested attribute
and picked it. `dependencyInsight` on 0.7.0 confirms it: `project :feature:amazon` resolved to
`Variant koverArtifact`. Kover 0.9.x tags that configuration with a `kover` usage, it becomes
incompatible, no candidate is left and the resolution fails.

### Description

- `ConfigureApisFlavorsForApplication`: `matchingFallbacks += "defaults"` on the
  `customEntitlementComputation` flavor. That plugin is applied only by `:api-tester`.
- Verified by bisecting the version in isolation: repo-wide `androidDependencies` fails on 0.9.9 and
  passes on 0.7.0 with no other change, and passes on 0.9.9 with this fallback.
- `dependencyInsight --configuration customEntitlementComputationDebugCompileClasspath --dependency
  :feature:amazon` now reports `defaultsDebugApiElements` instead of `koverArtifact`.
[...]

PR #3963, opened 2026-08-13
build: apply the AGP 9 config migrations that already work on AGP 8

AlvaroBrey:
Four config changes that AGP 9 requires but that are already valid on AGP 8.13, so they can land ahead of the version bump instead of inflating it:

- Library `targetSdk` moves out of `defaultConfig`, which AGP 9 removes for libraries, into `testOptions.targetSdk` and `lint.targetSdk`.
- `aarMetadata.minCompileSdk = 1` is now explicit.  AGP 9 defaults it to the module's `compileSdk`, which would silently tighten the published floor from the shipped `minCompileSdk=1` to `36`. 
- `proguard-android.txt` becomes `proguard-android-optimize.txt` in two internal app modules, since AGP 9 rejects the old file. Not consumer facing.
- `managedDevices.devices` becomes `localDevices`. 
- No consumer-visible change and no public API change.

### Motivation

The AGP 9 upgrade inventory grouped these under "land with the AGP bump", but that grouping is
more conservative than it needs to be. #3945 already landed `matchingFallbacks += "defaults"` on
AGP 8.13 for the same reason, and the same holds for these four: each is a config shape that AGP 9
requires and AGP 8.13 already accepts.

Splitting them out keeps the eventual AGP 9 PR to genuinely coupled changes. That PR is unusually
large because Gradle 9, AGP 9, Kotlin 2.3 and Poko cannot be separated from each other, so anything
that can be removed from it is worth removing.

### Description

- **`targetSdk`**: for a library module it only ever fed instrumentation tests and lint. AGP 9 drops
  it from `defaultConfig`, and both consumers now take it directly, so it moves to
  `testOptions.targetSdk` and `lint.targetSdk` in `ConfigureAndroidLibrary.kt`.
- **`aarMetadata.minCompileSdk = 1`**: AGP 8 defaults this to 1, so setting it explicitly is a no-op
  today. Under AGP 9 the default becomes the module's `compileSdk`, so leaving it implicit would
  raise the floor consumers must compile against from 1 to 36. Pinning it now means the AGP bump
  cannot change the published contract by accident.
- **proguard**: `getDefaultProguardFile("proguard-android.txt")` to the `-optimize` variant in
  `integration-tests` and `examples/purchase-tester`. AGP 9 rejects the non-optimize file.
- **managed devices**: `testOptions.managedDevices.devices` is deprecated in favor of `localDevices`.
  `localDevices` is a `NamedDomainObjectContainer<ManagedVirtualDevice>` rather than a polymorphic
  container, so `create<ManagedVirtualDevice>("pixel6Api34")` fails with
  `Container 'ManagedVirtualDevice container' is not polymorphic`. The type argument and the now
  unused import are removed.

### Not needed, contrary to the inventory

The inventory flagged `:feature:admob` as possibly needing its own `matchingFallbacks += "defaults"`,
alongside `:feature:amazon` and `:feature:galaxy`. It does not. `matchingFallbacks` is declared once
on the `customEntitlementComputation` **application** flavor in
`ConfigureApisFlavorsForApplication.kt`, not per library, so all three defaults-only feature modules
are already covered by the change that shipped in #3945.

### Testing

Everything below on the current AGP 8.13.2 and Gradle 8.14.5, since the point is that these do not
need AGP 9:

- `assembleDebug` across all modules, plus `:integration-tests:assembleDebug` and
  `:examples:purchase-tester:assembleDebug` for the proguard change specifically.
- `detektAll`.
- `scripts/api-check.sh` with no signature diff.
- Unpacked the release AAR and confirmed `aar-metadata.properties` still contains `minCompileSdk=1`,
  which is the regression this change exists to prevent.

</details>

---

> [!NOTE]
> **Low Risk**
> Build and tooling-only changes with no public API impact; integration-test ProGuard tweaks affect an internal minified test app only.
> 
> **Overview**
> Prepares the repo for **AGP 9** by landing config shape changes that are already valid on **AGP 8.13**, so they do not need to ride with the larger version bump.
> 
> In shared **library** convention (`ConfigureAndroidLibrary.kt`), **`targetSdk` is removed from `defaultConfig`** and set on **`testOptions.targetSdk`** and **`lint.targetSdk`**, matching AGP 9’s library model. **`aarMetadata.minCompileSdk = 1`** is set explicitly so AGP 9 cannot default it to `compileSdk` and raise the published compile floor.
> 
> **`integration-tests`** and **`examples/purchase-tester`** switch release minification to **`proguard-android-optimize.txt`** (AGP 9 rejects the old default). For integration tests, **ProGuard keeps `kotlin.**` and `androidx.tracing.**`** so R8 optimization does not strip classes the **androidTest** APK still needs at runtime.
> 
> **`baselineprofile`** uses **`testOptions.managedDevices.localDevices`** instead of deprecated **`devices`**, with a non-polymorphic **`create("pixel6Api34")`** call.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d4e50cc41bd2e7ca0fc4393d4017fca3c29c31a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
@@ -29,16 +29,23 @@ internal fun Project.configureAndroidLibrary() {
 
         defaultConfig {
             minSdk = obtainMinSdkVersion()
-            targetSdk = targetSdkVersion
             testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
             consumerProguardFiles("consumer-rules.pro")
+            // AGP 9 defaults this to compileSdk, tightening the published floor from 1 to 36.
+            aarMetadata.minCompileSdk = 1
         }
 
+        // AGP 9 removes targetSdk from a library's defaultConfig.
         testOptions {
+            targetSdk = targetSdkVersion
             unitTests.isIncludeAndroidResources = true
             unitTests.all {
                 it.maxHeapSize = "1024m"
             }
+        }
+
+        lint {
+            targetSdk = targetSdkVersion
         }
 
         compileOptions {
--- a/baselineprofile/build.gradle.kts
+++ b/baselineprofile/build.gradle.kts
@@ -1,7 +1,5 @@
 @file:Suppress("UnstableApiUsage")
 
-import com.android.build.api.dsl.ManagedVirtualDevice
-
 plugins {
     alias(libs.plugins.android.test)
     alias(libs.plugins.kotlin.android)
@@ -52,8 +50,8 @@ android {
     // This code creates the gradle managed device used to generate baseline profiles.
     // To use GMD please invoke generation through the command line:
     // ./gradlew :examples:paywall-tester:generateBaselineProfile
-    testOptions.managedDevices.devices {
-        create<ManagedVirtualDevice>("pixel6Api34") {
+    testOptions.managedDevices.localDevices {
+        create("pixel6Api34") {
             device = "Pixel 6"
             apiLevel = 34
             systemImageSource = "aosp"
--- a/examples/purchase-tester/build.gradle.kts
+++ b/examples/purchase-tester/build.gradle.kts
@@ -55,11 +55,11 @@ android {
         getByName("release") {
             isMinifyEnabled = true
             proguardFiles(
-                getDefaultProguardFile("proguard-android.txt"),
+                getDefaultProguardFile("proguard-android-optimize.txt"),
                 "proguard-rules.pro",
             )
             testProguardFiles(
-                getDefaultProguardFile("proguard-android.txt"),
+                getDefaultProguardFile("proguard-android-optimize.txt"),
                 "proguard-test-rules.pro",
             )
             signingConfig = signingConfigs.getByName("release")
--- a/integration-tests/build.gradle.kts
+++ b/integration-tests/build.gradle.kts
@@ -62,8 +62,8 @@ android {
     buildTypes {
         release {
             isMinifyEnabled = true
-            proguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")
-            testProguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")
+            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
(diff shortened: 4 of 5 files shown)


build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
```diff
@@ -29,18 +29,25 @@ internal fun Project.configureAndroidLibrary() {
 
         defaultConfig {
             minSdk = obtainMinSdkVersion()
-            targetSdk = targetSdkVersion
             testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
             consumerProguardFiles("consumer-rules.pro")
+            // AGP 9 defaults this to compileSdk, tightening the published floor from 1 to 36.
+            aarMetadata.minCompileSdk = 1
```
tonidero, line 35: Hmm I'm not sure I understand why we set it to 1... Since this defaults to `compileSdk`, should we keep it to the current `compileSdk` to avoid changes?

tonidero (commented):
Thanks for iterating on this! I think it makes sense!

tonidero (approved):

AlvaroBrey:
* **#3965** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3965?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3963** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3963?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3963?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3961** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3961?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
AlvaroBrey: Right now we're publishing `minCompileSdk=1`, this just keeps it as is. If we default it to `compileSdk` ie 36,  then consumers with lower compile SDK would not be able to compile

AlvaroBrey: TBF this should likely be equal to our minSdk, not 1, but our compile is strictly worse

PR #3964, opened 2026-08-13
build: update to AGP 9 (AGP 9.2.1, Gradle 9.4.1, Kotlin 2.2.21)

AlvaroBrey:
> [!NOTE]
> This targets branch `11.0-dev` which will become the epic branch for the upcoming major release. Once this first PR is merged, I'll open a continuous one from that one to `main` until we are ready to release it altogether. 

- Moves the SDK's build toolchain to AGP 9:

| Dependency | From | To |
|---|---|---|
| AGP | 8.13.2 | **9.2.1** |
| Gradle | 8.14.5 | **9.4.1** |
| Kotlin | 2.0.21 | **2.2.21** |
| Poko | 0.17.2 | **0.20.2** |
| Paparazzi | 2.0.0-alpha02 | **2.0.0-alpha05** |

- No public API change and no `api*.txt` diff.
- **Why this becomes a major**: this transitively raises the miminum kotlin version for consumers to 2.1.0. up from 1.8.0.
- Kotlin lands on **2.2.21, the minimum required by AGP 9**,
- AGP is capped at **9.2.1** rather than latest: AGP 9.3.x requires Gradle 9.5+, and Gradle 9.5+ embeds a Kotlin that a 2.2 compiler cannot read, bumping consumers to an even higher minimum Kotlin version.
- Nothing downstream needs AGP 9 or Gradle 9, just the Kotlin bump. The published AAR metadata is unchanged (`minCompileSdk=1`, `minAndroidGradlePluginVersion=1.0.0`).

Companion docs PR: https://github.com/RevenueCat/docs/pull/1942
Companion PHC PR: https://github.com/RevenueCat/purchases-hybrid-common/pull/1844

### Checklist
- [ ] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-ios` and hybrids

<details><summary>Agent description</summary>

### Motivation

AGP 8 is increasingly behind, and AGP 9 removes a number of DSL surfaces we still used. The
prerequisite work already landed separately in #3931, #3961 and #3963.

AGP 9 has a runtime dependency on Kotlin Gradle Plugin **2.2.10 or higher**, so a Kotlin bump is not
optional here. Everything above that floor is a choice, and this PR deliberately takes the minimum.

### Why not Kotlin 2.3.x

An earlier revision of this branch used Kotlin 2.3.21. Measuring the cost changed the decision:

- **It breaks every React Native consumer.** We publish `kotlin-stdlib` at `compile` scope and Gradle
[...]

PR #3965, opened 2026-08-13
build: migrate to AGP 9 built-in Kotlin

AlvaroBrey:
- Removes the `android.builtInKotlin=false` and `android.newDsl=false` opt-out that #3964 added, and takes the `kotlin-android` plugin out of 13 build scripts and the 3 convention plugins. AGP 10 removes the ability to opt out at all, so this is the migration rather than a deferral.
- Headline caveat: `ui/debugview/api.txt` loses `InternalDebugRevenueCatScreenViewModelFactory`. Metalava now analyses the release variant, and that class lives in `src/debug`. The release AAR contains zero occurrences of it, so it never shipped and the entry was wrong, but it reads as an API removal and deserves a second opinion.
- Bumps `androidx.navigation` 2.5.3 to 2.9.8. safeargs 2.5.3 detects AGP through the old DSL and hard fails, so the navigation bump is a prerequisite for the new DSL rather than the independent change the upgrade inventory described.
- Reworks the Dokka config in `:purchases` and `:ui:revenuecatui`. Built-in Kotlin registers one source set per Android variant instead of the main and flavor source sets, and registers them lazily, so both modules now select the documented variant through `configureEach`.
- Stacked on #3964, which is the AGP 9 bump itself.

### Checklist

### Motivation

AGP 9 enables built-in Kotlin and rejects the `org.jetbrains.kotlin.android` plugin outright:

```
The 'org.jetbrains.kotlin.android' plugin is no longer required for Kotlin support since AGP 9.0.
```

Opting out needs two flags together, `android.builtInKotlin=false` and `android.newDsl=false`, because
the kotlin-android plugin is not compatible with the new DSL. #3964 takes that opt-out so the version
bump stays reviewable. AGP 10 removes the opt-out, so the migration has to happen regardless; doing it
separately keeps the two diffs legible.

### Description

Three things blocked the migration, each hidden behind the previous one.

**The old DSL extension types are gone.** `com.android.build.gradle.BaseExtension` and
`com.android.build.gradle.LibraryExtension` do not exist once the new DSL is active. The convention
plugins now configure `com.android.build.api.dsl.ApplicationExtension` and
`com.android.build.api.dsl.LibraryExtension`, and set `compileSdk` directly rather than calling
`compileSdkVersion()`.

**safeargs 2.5.3 cannot see AGP 9.** It fails with `safeargs plugin must be used with android plugin`,
because its detection goes through the old DSL. `androidx.navigation` moves to 2.9.8, the current
stable. Worth flagging that the upgrade inventory lists the navigation bump as independent and
optional; it is neither.

**Dokka source sets changed shape.** With the kotlin-android plugin, Dokka saw the Kotlin source sets
and the config addressed `main`, `defaults` and `customEntitlementComputation`. Built-in Kotlin instead
registers one source set per Android variant:

```
defaultsDebug   customEntitlementComputationDebug
defaultsRelease customEntitlementComputationRelease
plus the AndroidTest and UnitTest variants
```

They are also registered after the `dokka { }` block runs, so `named("defaultsRelease")` throws even
though the name is valid later. Both modules now use `configureEach` with a name check, suppressing
everything except the documented variant.

`:purchases` keeps `reportUndocumented`, `skipDeprecated`, the Android external documentation link, the
two source links, and the `paywalls.components` suppression, all moved onto the selected variant.

### Testing

- `assembleDebug` across all modules, `:purchases` unit tests, `detektAll`, and `scripts/api-check.sh`.
- Dokka generates 956 pages, the public entry points are present, and the internal
  `com.revenuecat.purchases.paywalls.components` package is still suppressed. Checked explicitly,
  because a bad rewrite of that block would silently produce empty or over-broad docs rather than fail.
- For the signature change: unpacked `debugview-defaults-release.aar` and confirmed it contains no
  `InternalDebugRevenueCatScreenViewModelFactory`, so the class was only ever in debug builds.

</details>

### DIFF
--- a/baselineprofile/build.gradle.kts
+++ b/baselineprofile/build.gradle.kts
@@ -2,7 +2,6 @@
 
 plugins {
     alias(libs.plugins.android.test)
-    alias(libs.plugins.kotlin.android)
     alias(libs.plugins.baselineprofile)
 }
 
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidApplication.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidApplication.kt
@@ -1,6 +1,6 @@
 package com.revenuecat.purchases.android.buildlogic.convention
 
-import com.android.build.gradle.BaseExtension
+import com.android.build.api.dsl.ApplicationExtension
 import com.revenuecat.purchases.android.buildlogic.ktx.getVersion
 import com.revenuecat.purchases.android.buildlogic.ktx.versionCatalog
 import org.gradle.api.JavaVersion
@@ -14,8 +14,8 @@ internal fun Project.configureAndroidApplication() {
     val minSdkVersion = libs.getVersion("android-minSdk").toInt()
     val targetSdkVersion = libs.getVersion("android-targetSdk").toInt()
 
-    extensions.configure<BaseExtension> {
-        compileSdkVersion(compileSdkVersion)
+    extensions.configure<ApplicationExtension> {
+        compileSdk = compileSdkVersion
 
         defaultConfig {
             minSdk = minSdkVersion
--- a/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
+++ b/build-logic/convention/src/main/kotlin/com/revenuecat/purchases/android/buildlogic/convention/ConfigureAndroidLibrary.kt
@@ -1,8 +1,8 @@
 package com.revenuecat.purchases.android.buildlogic.convention
 
+import com.android.build.api.dsl.LibraryExtension
 import com.android.build.api.variant.HasUnitTestBuilder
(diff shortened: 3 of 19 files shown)


AlvaroBrey:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3965?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3965** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3965?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3965?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3963** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3963?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3961** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-android/3961?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #3969, opened 2026-08-14
feat(checkpoints): Add StoreDimensionProvider with store.country

tonidero:
### Description

Adds the store as a third dimension source, with one dimension for now: `store.country`.

Note that what we get in Android is a different code than in iOS. For example, we get the 2-letter `US` instead of the 3-letter `USA` as in iOS. In this PR, I try to mimick the format in iOS, so we can consolidate the rules.

It fetches through `awaitStorefrontCountryCode` rather than reading the cached property, so the dimension is available on the first checkpoint of a session instead of missing until the billing client happens to connect. A store that can't answer contributes nothing instead of failing the snapshot — that is a routine outcome (the Galaxy Store never answers), while a failure would abort the evaluation and take an otherwise resolvable checkpoint down with it.

Similar to RevenueCat/purchases-ios#7399.

Also fixes `RulesDimensionResolver` nesting an empty object for a source with nothing to contribute. An empty object is truthy in JSON Logic, so `{"var": "store"}` would have read as present for a customer whose store never answered — reachable for the first time now that a provider routinely returns nothing.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint rule evaluation semantics (new dimension and absent-vs-empty namespace behavior) and adds an async storefront fetch on the evaluation path, though store failures are isolated and do not abort snapshots.
> 
> **Overview**
> Checkpoint **local rules** can now read **`store.country`** (ISO 3166-1 alpha-3, e.g. `USA`) so predicates can align with iOS. Android alpha-2 storefront codes are converted at evaluation time; unmapped or invalid codes are omitted.
> 
> **`StoreDimensionProvider`** is registered on **`LocalRulesEvaluator`** and resolves the country via **`awaitStorefrontCountryCode()`** on each evaluation (not a cached snapshot), so the first checkpoint in a session can still get a value once the store responds. Store errors, missing answers, or teardown mid-evaluation yield **no store namespace** instead of failing the whole dimension snapshot.
> 
> **`RulesDimensionResolver`** no longer nests an empty object when a source has nothing to contribute (including empty **custom** variables). Empty objects are truthy in JSON Logic, so absent namespaces correctly mean “store/custom not present” for predicates like `{"var": "store"}`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee6973ec9a534cf6caa670171c703c2fe35f0cbc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -32,6 +32,7 @@ import com.revenuecat.purchases.common.isDeviceProtectedStorageCompat
 import com.revenuecat.purchases.common.localrules.DeviceDimensionProvider
 import com.revenuecat.purchases.common.localrules.LocalRulesEvaluator
 import com.revenuecat.purchases.common.localrules.RulesEngineLoggerBridge
+import com.revenuecat.purchases.common.localrules.StoreDimensionProvider
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.common.networking.APISourceFailover
 import com.revenuecat.purchases.common.networking.DeviceConnectivityChecker
@@ -355,7 +356,12 @@ internal class PurchasesFactory(
             }
             RulesEngine.setLogger(RulesEngineLoggerBridge)
             val localRulesEvaluator = LocalRulesEvaluator(
-                providers = listOf(DeviceDimensionProvider(appConfig, localeProvider)),
+                providers = listOf(
+                    DeviceDimensionProvider(appConfig, localeProvider),
+                    // Only read during a checkpoint evaluation, so the instance is configured by then. Same
+                    // reasoning as CheckpointWorkflowResolverImpl's getOfferings.
+                    StoreDimensionProvider { Purchases.sharedInstance.awaitStorefrontCountryCode() },
+                ),
             )
             if (remoteConfigManager != null && uiConfigProvider != null && workflowsConfigProvider != null) {
                 remoteConfigManager.registerListener(uiConfigProvider)
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
@@ -47,8 +47,7 @@ internal class RulesDimensionResolver(
 
     /**
      * [customVariables] are the caller's own values for this one evaluation, exposed under
-     * [RulesDimensionNamespace.Custom]. An empty map contributes no namespace at all rather than an empty object,
-     * which is truthy in JSON Logic: a call with no values should read as absent, not as present-but-empty.
+     * [RulesDimensionNamespace.Custom].
      */
     @Suppress("ReturnCount")
     suspend fun snapshot(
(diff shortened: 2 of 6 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -91,10 +88,14 @@ internal class RulesDimensionResolver(
...
 
 /** Nests [dimensions] under [namespace], or returns the conflict that stops the whole snapshot. */
+@Suppress("ReturnCount")
 private fun MutableMap<String, MutableMap<String, Value>>.addDimensions(
     namespace: RulesDimensionNamespace,
     dimensions: Map<String, RulesDimensionValue>,
 ): RulesDimensionResolutionException.ConflictingDimension? {
+    // A source with nothing to say contributes no namespace at all: an empty object is truthy in JSON Logic, so
+    // `{"var": "store"}` would read as present for a customer whose store never answered.
+    if (dimensions.isEmpty()) return null
```
tonidero, line 98: This seemed like something we were missing @rickvdl

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -355,7 +356,12 @@ internal class PurchasesFactory(
             }
             RulesEngine.setLogger(RulesEngineLoggerBridge)
             val localRulesEvaluator = LocalRulesEvaluator(
-                providers = listOf(DeviceDimensionProvider(appConfig, localeProvider)),
+                providers = listOf(
+                    DeviceDimensionProvider(appConfig, localeProvider),
+                    // Only read during a checkpoint evaluation, so the instance is configured by then. Same
+                    // reasoning as CheckpointWorkflowResolverImpl's getOfferings.
+                    StoreDimensionProvider { Purchases.sharedInstance.awaitStorefrontCountryCode() },
```
tonidero, line 363: I thought about just returning the cached value... But I wanted to have an accurate value, and once the billing is connected, it should normally just be synchronous to get it.

rickvdl: FWIW did the same on iOS

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
rickvdl: Good point! Will address that one as well

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
rickvdl: Just thinking; what happens if the storefront call hangs (if the BC isn't connected) or for any other reason? Could we have a test for that? I think we have the same issue on iOS, maybe we need to (more generally) see how to handle these?

rickvdl (approved):
Just one question

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Well, if there is an error we do handle that. But if there is indeed some hang connecting to the BC, it would mean we would get stuck here indeed... We could potentially add some timeout for these. But I will take note so it can be addressed in follow-up PRs. Good catch!

rickvdl: Great! Did the same :)

PR #3970, opened 2026-08-14
Deserialize audience rules for rules engine evaluation

vegaro:
The rules engine accepts rules as a string, so decoding the whole audience as an opaque blob would make us deserialize it again later.

Implements [WFL-464](https://linear.app/revenuecat/issue/WFL-464/android-deserialize-audience-rules-for-rules-engine-evaluation).

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/Audience.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/Audience.kt
@@ -1,0 +1,40 @@
+package com.revenuecat.purchases.common.audiences
+
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.json.JsonDecoder
+import kotlinx.serialization.json.JsonEncoder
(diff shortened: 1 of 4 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/Audience.kt
```diff
@@ -0,0 +1,40 @@
...
+    val rules: String,
+)
+
+internal object JsonObjectStringSerializer : KSerializer<String> {
+    override val descriptor: SerialDescriptor = JsonObject.serializer().descriptor
+
+    override fun deserialize(decoder: Decoder): String {
+        val jsonDecoder = decoder as? JsonDecoder
+            ?: throw SerializationException("Audience rules can only be deserialized from JSON.")
+        val element = jsonDecoder.decodeJsonElement()
```
tonidero, line 26: Hmm I guess with this we're first decoding to json, then later transforming to string... Probably NABD, but would have been great if we could have skipped that part I guess 😅

tonidero (commented):
A couple comments. ~~I think the only one we need to solve is the name of the topic... The other one is not a blocker IMO.~~ Sorry got confused. This is right I think!

tonidero (approved):
:shipit:

purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/Audience.kt
vegaro: yeah, `decodeString()` requires a quoted string unfortunately

PR #3971, opened 2026-08-14
Evaluate checkpoint audiences before resolving workflows

vegaro:
Evaluates the audience explicitly before the checkpoint answers with a workflow.

Audience definitions are stored inline in `ConfigItem.metadata`, without a `blob_ref`, so `AudiencesConfigProvider` now reads and decodes the topic metadata directly.

[WFL-464](https://linear.app/revenuecat/issue/WFL-464/android-deserialize-audience-rules-for-rules-engine-evaluation).

Validation: `:purchases:compileDefaultsDebugKotlin` and the pre-commit Detekt check. The existing blob-backed `AudiencesConfigProviderTest` cases were not updated as part of this fix.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint targeting logic and remote-config read path for audiences; misconfiguration or parse failures alter which workflow (or no action) customers see, though behavior is fail-closed toward configuration unavailable rather than silent wrong matches.
> 
> **Overview**
> Checkpoint resolution no longer treats every checkpoint rule as matching via a placeholder predicate. Each rule’s `audience_id` is resolved through **`AudiencesConfigProvider`**, and **`LocalRulesEvaluator`** evaluates the audience’s rules JSON against device/store/custom dimensions until the first match picks a workflow.
> 
> **`AudiencesConfigProvider`** now loads audiences from inline remote-config topic **`metadata`** (not blob storage) and decodes them to **`Audience`** for the rules engine.
> 
> **`PurchasesFactory`** wires **`AudiencesConfigProvider`** into the orchestrator when remote config is enabled (absent for custom entitlement computation). Missing or unreadable audiences surface **`CONFIGURATION_UNAVAILABLE`** instead of **`NO_MATCH`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fdbcf963e24eb0531497078acd04185ab399fae5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/AudiencesConfigProvider.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/AudiencesConfigProvider.kt
@@ -1,11 +1,29 @@
 package com.revenuecat.purchases.common.audiences
 
+import com.revenuecat.purchases.JsonTools
+import com.revenuecat.purchases.common.debugLog
+import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigManager
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigTopic
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.decodeFromJsonElement
 
 internal class AudiencesConfigProvider(
     private val manager: RemoteConfigManager,
 ) {
-    suspend fun getAudience(identifier: String): Audience? =
-        manager.blobData(RemoteConfigTopic.Audiences, identifier)
+    suspend fun getAudience(identifier: String): Audience? {
+        val metadata = manager.topic(RemoteConfigTopic.Audiences)
+            ?.get(identifier)
+            ?.metadata
+            ?: return null
+        debugLog {
+            "Raw audience remote config metadata for '$identifier' before SDK deserialization: $metadata"
+        }
+        return try {
+            JsonTools.json.decodeFromJsonElement<Audience>(metadata)
+        } catch (e: SerializationException) {
+            errorLog(e) { "Failed to parse remote config metadata for audience '$identifier' as JSON." }
+            null
+        }
+    }
 }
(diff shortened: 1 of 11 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/LocalRulesEvaluator.kt
```diff
@@ -60,7 +67,8 @@ internal class LocalRulesEvaluator(
 
         var firstFailure: LocalRulesEvaluationException.PredicateEvaluation? = null
         for ((index, rule) in rules.withIndex()) {
-            val result = RulesEngine.evaluate(rule.predicate, snapshot.values)
+            val predicate = predicateFor(rule).getOrElse { error -> return Result.failure(error) }
```
tonidero, line 70: Hmm do we return if there is an error 🤔 I guess it depends... Not sure I see what types of error we could have... Just wondering if we should continue with the next rule before returning a failure, but if it's an actual error, this could make sense!

tonidero (approved):
Looks good to me! Hopefully we can test if it works from the backend 🙏 Just left a question, but I don't think it's a blocker.

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/LocalRulesEvaluator.kt
vegaro: My thinking here was that a valid audience that doesn't match evaluates to `false` and continues to the next rule, but since rules are ordered, continuing after a load/parse failure could select a later rule even though the unavailable earlier one should have won. And if the audiences topic itself is unavailable, trying the next rule won't help anyway?

tonidero: Yeah... I think it's ok for now 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/audiences/AudiencesConfigProvider.kt
```diff
@@ -1,11 +1,29 @@
...
+        debugLog {
+            "Raw audience remote config metadata for '$identifier' before SDK deserialization: $metadata"
+        }
+        return try {
+            JsonTools.json.decodeFromJsonElement<Audience>(metadata)
+        } catch (e: SerializationException) {
+            errorLog(e) { "Failed to parse remote config metadata for audience '$identifier' as JSON." }
+            null
+        }
+    }
```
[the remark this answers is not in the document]
vegaro: This is accepted. The data comes through metadata not blobData

tonidero (approved):

PR #3972, opened 2026-08-14
test(checkpoints): Hardcode audience rules to exercise rule selection [DO NOT MERGE]

tonidero:
### ⚠️ Not for main

Testing scaffolding only. It replaces config-driven rule data with constants so the audience walk can be exercised on a device before the topic that serves real predicates exists.

### Description

Substitutes three hardcoded rules for the published ones, in priority order:

| # | Predicate | Presents the workflow whose offering is |
| --- | --- | --- |
| 1 | `custom.name == "Antonio"` and `device.locale == "en_us"` | `default` |
| 2 | `custom.name == "Antonio"` | `rick` |
| 3 | `true` | `no_name` |

Targets are named by offering rather than by workflow, because workflow ids are generated per project while the offering ids are the ones a tester recognises — the workflow that renders each is found by inverting `offeringIdByWorkflowId`. A rule whose offering no workflow renders is skipped, so the rest stay testable on a project that lacks one of the three.

The `checkpoint_rules` read is untouched, so `DISABLED` / `UNKNOWN_CHECKPOINT` / `Unavailable` still behave as they do today; only the rule data is replaced. The switch is a defaulted constructor param, and the existing resolver tests opt out so they keep covering the real published-rule walk.

`custom.name` comes from a new field on the checkpoint tester's onboarding Personalize step. An empty field sends no `custom.name` at all, so a rule reading it sees an absent dimension rather than an empty string.

### How to test

In `examples/checkpointtester` → Onboarding, with `onboarding_complete` configured and offerings `default`, `rick` and `no_name` each backed by a workflow:

| Input | Expected |
| --- | --- |
| `Antonio`, en-US device | rule 1 → `default` |
| `Antonio`, any other locale | rule 2 → `rick` |
| any other name | rule 3 → `no_name` |
| empty | rule 3 → `no_name` |

Which rule won shows in logcat: `Checkpoint resolved to workflow '<id>' (offering: <id>)`.

Stacked on #3969.

### DIFF
--- a/examples/checkpointtester/src/main/java/com/revenuecat/checkpointtester/ui/screens/onboarding/OnboardingScreen.kt
+++ b/examples/checkpointtester/src/main/java/com/revenuecat/checkpointtester/ui/screens/onboarding/OnboardingScreen.kt
@@ -10,6 +10,7 @@ import androidx.compose.material3.Button
 import androidx.compose.material3.LinearProgressIndicator
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.OutlinedButton
+import androidx.compose.material3.OutlinedTextField
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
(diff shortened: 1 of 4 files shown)


tonidero:
This was only for a quick demo's purposes so closing now.

PR #3976, opened 2026-08-17
Report public API changes on PRs and in the SDK API feed

AlvaroBrey:
Brings the public API detector to Android: a Danger rule diffs the metalava signature files against the PR base, posts a collapsible diff on the PR, and mirrors a summary into the SDK API feed channel.

Companion change on `purchases-ios`: https://github.com/RevenueCat/purchases-ios/pull/7425

### Checklist

### Motivation

`purchases-ios` announces public API changes on every PR that touches its API surface. Android already has the hard half: metalava signature files committed to the repo, and the `metalava` job that fails when they drift.

What was missing is the "what changed in this PR" view. The drift gate cannot provide it: once the author regenerates the dump, the working-tree diff is empty and the evidence is gone. `purchases-ios` solves this by extracting its baselines at the merge base. Danger supplies it for free here, since its diff is already taken against the PR base.

### Description

- `danger/api_diff_report.rb` parses metalava signature diffs. It drops `}`, `package … {` and the format banner rather than allowlisting `method`/`ctor`/`field`, so a keyword metalava adds in a future release cannot silently go unreported.
- Declarations are deduplicated across `purchases/api-defauts.txt` and `purchases/api-entitlement.txt`, which hold two flavours of largely the same surface, so a shared addition is counted once. File order is preserved so a new type reads before its own members.
- `ApiDiffReport.run` is the single entry point: the Dangerfile passes in Danger's changed-file list, a patch lookup and the PR link, and gets back the comment body. Any failure in the block warns instead of failing the run, so a Slack outage or a parser bug cannot take the rest of the Dangerfile down with it.
- The channel name comes from the context rather than the CI config: the orb's `danger` job accepts no `environment:` block, and keeping an internal channel name out of a public repo is preferable regardless.
- The unit tests run as `post-steps` on the existing `revenuecat/danger` job, which already has Ruby and the checkout, so no job gains a toolchain it did not need.
- Test fixtures use deliberately synthetic declarations (`apiDiffDemoPing`), since a plausible-looking unreleased API name in a public repo reads as a roadmap hint.

Verification not visible in the diff: `danger dry_run` always feeds an empty file list, so it only proves the Dangerfile parses, `require_relative` resolves under Danger's `instance_eval`, and the rule is a no-op when nothing changed. The populated path was exercised by feeding a real `git diff` of `ui/debugview/api.txt` straight into `ApiDiffReport`. Two seams neither covers, both needing a PR that actually changes a signature file: `git.diff_for_file(...).patch` returning that unified diff, and `github.pr_json` supplying the PR link.

</details>

---

> [!NOTE]
> **Low Risk**
> CI/Danger-only automation with isolated error handling; no runtime SDK or customer-facing behavior changes.
> 
> **Overview**
> Adds Android parity with **purchases-ios** for surfacing **public API** changes on PRs that touch metalava signature files (`api.txt`, `api-*.txt`).
> 
> **Danger** loads `ApiDiffReport`, diffs those files against the PR base, posts a **collapsible diff** on the PR, and **best-effort** posts a summary to the SDK API feed Slack channel via `slack-secrets`. Slack failures only **warn**; they do not fail Danger or block the PR comment.
> 
> **`danger/api_diff_report.rb`** parses unified patches (filters metalava noise, dedupes across `purchases` flavour files, caps size for GitHub/Slack). **`danger/api_diff_report_test.rb`** covers parsing, formatting, and Slack posting. CircleCI runs those tests as a **post-step** on the existing Danger job and wires the **`slack-secrets`** context for tokens/channel env vars.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47b0f82c064521ff11d7942bf163376408e79149. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -967,6 +967,12 @@ workflows:
     jobs:
       - revenuecat/danger:
           ruby_version: "3.2.0"
+          context:
+            - slack-secrets
+          post-steps:
+            - run:
+                name: Run api diff report unit tests
+                command: ruby danger/api_diff_report_test.rb
 
   snapshot-deploy-manual:
     when:
--- a/danger/api_diff_report.rb
+++ b/danger/api_diff_report.rb
@@ -1,0 +1,203 @@
+# Reports the public API a PR changes, from the metalava signature diffs. Loaded by the Dangerfile.
+
+require 'json'
+require 'net/http'
+require 'uri'
+
+module ApiDiffReport
+  PLATFORM_LABEL = "Android :android:".freeze
+
+  # The same two patterns scripts/api-check.sh gates on.
+  SIGNATURE_GLOBS = ["api.txt", "api-*.txt"].freeze
+
+  # A metalava 4.0 file holds nothing but the banner, `package … {`, `}`, types and members.
+  NOISE = %r{\A(\}|package\s|//)}.freeze
+
+  SLACK_LIMIT = 10
+
+  # Slack stops wrapping code blocks past this; the full list is in the PR comment.
+  SLACK_WIDTH = 160
+
+  # GitHub answers 422 over 65536 characters, and Danger shares one comment across every rule.
+  MARKDOWN_LIMIT = 200
+  MARKDOWN_WIDTH = 200
+
+  module_function
+
+  def signature_file?(path)
+    basename = File.basename(path.to_s)
+
+    SIGNATURE_GLOBS.any? { |glob| File.fnmatch?(glob, basename) }
+  end
+
+  def module_name(path)
+    File.dirname(path.to_s).tr("/", ":")
+  end
+
+  def declarations(patch)
+    added = []
+    removed = []
+
+    patch.to_s.each_line do |raw_line|
+      sign = raw_line[0]
+      next unless sign == "+" || sign == "-"
+      next if raw_line.start_with?("+++", "---")
+
+      line = raw_line[1..].to_s.strip
+      next if line.empty? || NOISE.match?(line)
+
+      # Members end in `;`; only a type header carries the opening brace of its body.
+      (sign == "+" ? added : removed) << line.sub(/\s*\{\z/, "")
+    end
+
+    [added, removed]
+  end
+
+  # A member line omits its owning type, so two types in one file can add the same text. Identity is
+  # therefore the text plus how many times it has already appeared in that file.
+  def occurrences(declarations)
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -967,6 +967,12 @@ workflows:
     jobs:
       - revenuecat/danger:
           ruby_version: "3.2.0"
+          context:
+            - slack-secrets
+          post-steps:
+            - run:
+                name: Run api diff report unit tests
+                command: ruby danger/api_diff_report_test.rb
```
tonidero, line 975: Interesting... I don't think we had tests for danger rules yet... I guess it's nice to have them :) (I thought of just moving the logic to the fastlane plugin where it can be better tested, but danger ideally runs fast without having to install fastlane)

danger/api_diff_report.rb
```diff
@@ -0,0 +1,203 @@
...
+
+  def slack_credentials
+    token = TOKEN_VARIABLES.map { |name| ENV[name].to_s }.find { |value| !value.empty? }
+    channel = ENV["SLACK_CHANNEL_SDK_NEW_API"].to_s
+    return nil if token.nil? || channel.empty?
+
+    [token, channel]
+  end
+
+  def slack_request(message, bot_token:, channel:)
```
tonidero, line 155: Hmm so I feel a lot of this logic is likely duplicated with iOS... Wonder if we should try to extract to our danger repo to share it? Ideally only the creation of the API diff message is platform specific, and all the rest is shared.

tonidero (commented):
Very cool! The only thing I would like to try is avoiding the duplication on the danger files with iOS and Android... Wdyt?

danger/api_diff_report.rb
AlvaroBrey: iOS is currently running all of this in fastlane, not Danger. I'd have to refactor it a bit, will check

tonidero: Ohhh ok... Then I think it's ok to keep as is and we can try to refactor later if needed then... Just ideally we would have the same mechanisms I guess, but not a blocker for this!

tonidero: As in, eventually we probably want to do this for hybrids as well

tonidero (approved):

PR #3978, opened 2026-08-17
feat(checkpoints): Add date and record collection rules dimension values

tonidero:
### Description

`RulesDimensionValue` was limited to basic primitives. This PR adds support for 2 more types: Dates and Records. These will be used in the follow-up PR.

---

> [!NOTE]
> **Low Risk**
> Additive sealed-class variants and resolver mapping in local rules with no changes to purchase or auth flows; risk is limited to how new dimension types are serialized for rules evaluation.
> 
> **Overview**
> Extends **`RulesDimensionValue`** beyond string/bool/int/double so checkpoint rules can express **dates** and **collections of records** (e.g. purchases, entitlements).
> 
> **`DateValue`** is converted to epoch milliseconds in the rules engine, so predicates use normal numeric comparison without a separate date type.
> 
> **`ObjectListValue`** becomes an array of objects for JSON Logic iteration (`some`, `all`, `none`, `filter`); nested record fields map recursively, and an empty list stays a present empty array so “no records” is explicit.
> 
> **`RulesDimensionResolver`** mapping and **`RulesDimensionResolverTest`** cover conversion, date ordering, per-record predicates, and empty lists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b171d10a0edb1370d172c97814402ec213f552f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
@@ -112,4 +112,8 @@ private val RulesDimensionValue.asRulesEngineValue: Value
         is RulesDimensionValue.BoolValue -> Value.BoolValue(value)
         is RulesDimensionValue.IntValue -> Value.IntValue(value)
         is RulesDimensionValue.DoubleValue -> Value.FloatValue(value)
+        is RulesDimensionValue.DateValue -> Value.IntValue(value.time)
+        is RulesDimensionValue.ObjectListValue -> Value.ArrayValue(
+            value.map { record -> Value.ObjectValue(record.mapValues { (_, item) -> item.asRulesEngineValue }) },
+        )
     }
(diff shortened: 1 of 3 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -112,4 +112,8 @@ private val RulesDimensionValue.asRulesEngineValue: Value
         is RulesDimensionValue.BoolValue -> Value.BoolValue(value)
         is RulesDimensionValue.IntValue -> Value.IntValue(value)
         is RulesDimensionValue.DoubleValue -> Value.FloatValue(value)
+        is RulesDimensionValue.DateValue -> Value.IntValue(value.time)
```
tonidero, line 115: TBH, we could just pass the `value.time` epoch milliseconds as a long/double so it felt a bit unnecessary... But this allows us to read `Date` types in the injected data and transform it to the expected type in one place... Lmk what you think!

rickvdl: I think this is nice to have, as it solidifies the underlying type both platforms use for this

rickvdl (approved):

PR #3980, opened 2026-08-17
feat(checkpoints): Add SubscriberDimensionProvider with the customer's purchases

tonidero:
### Description

Adds the customer's subscription state as a third dimension source, under a `subscriber` namespace: 21 of the 23 fields `dimensions.csv` lists for that group, all derived on-device from `CustomerInfo`.

| Key | Derived from |
|---|---|
| `customerId` / `originalAppUserId` | the app user ID / `CustomerInfo.originalAppUserId` |
| `allPurchasedProductIds` | `CustomerInfo.allPurchasedProductIds` |
| `anyActiveStore` | the stores of the subscriptions that are active or in a grace period |
| `latestStore`, `latestOwnershipType`, `isRcPromo`, `latestEntitlements` | the most recent purchase of any kind (its subscription, where only a subscription can answer) |
| `latestAutoRenewIntent`, `mostRecentRenewalAt`, `subscriptionOptOutAt`, `trialOptOutAt` | the most recent subscription |
| `isCurrentlyTrialing`, `trialStartAt`, `trialEndAt` | the subscription whose current period is a trial |
| `hasMadeNonSubscriptionPurchase`, `hasMadeSandboxPurchase`, `mostRecentPurchaseAt` | every purchase |
| `firstSeenAt`, `lastSeenAt`, `latestExpirationAt` | `firstSeen` + the earliest purchase, `requestDate`, `latestExpirationDate` |

Read on every evaluation rather than snapshotted: a purchase, a renewal or a cancellation lands mid-session, and an audience keyed on subscription state has to agree with what the customer's access actually is. It goes through the orchestrator rather than `Purchases.awaitCustomerInfo`, which only exists in the `defaults` source set, and asks for the cache with a fetch fallback so the first checkpoint of a session is not left without subscriber dimensions on a fresh install.

The app user ID is read separately from the rest, so a rule on `subscriber.customerId` still gets an answer when the customer info read fails. A failed read contributes no other dimensions instead of failing the snapshot, like `StoreDimensionProvider`.

#### Deliberate differences from the dashboard

- **Sandbox purchases count.** The dashboard evaluates these fields over production purchases only. Mirroring that would leave every subscriber rule unmatchable in a debug build or against a license tester, where every purchase is a sandbox one. `hasMadeSandboxPurchase` is the one dimension that reads the flag itself.
- **`status` and `latestProduct` are left out.** `status`'s value set isn't settled, and `latestProduct`'s exact operators compare the RevenueCat product ID, which the SDK never sees — it only has the store identifier. A predicate reading either resolves to null in the engine, which is an ordinary non-match rather than an error.

#### Known local limitations

- `trialStartAt` / `trialEndAt` are knowable only while the trial is the current period. Once it converts, the SDK no longer carries the trial window while the backend still does.
- `firstSeenAt` cannot include the backend's customer creation date; `firstSeen` and the earliest purchase are the closest local facts.
- `lastSeenAt` is `CustomerInfo.requestDate` — when the backend last answered us, which is when it last saw this customer through this device.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -33,6 +33,7 @@ import com.revenuecat.purchases.common.localrules.DeviceDimensionProvider
 import com.revenuecat.purchases.common.localrules.LocalRulesEvaluator
 import com.revenuecat.purchases.common.localrules.RulesEngineLoggerBridge
 import com.revenuecat.purchases.common.localrules.StoreDimensionProvider
+import com.revenuecat.purchases.common.localrules.SubscriberDimensionProvider
 import com.revenuecat.purchases.common.log
 import com.revenuecat.purchases.common.networking.APISourceFailover
 import com.revenuecat.purchases.common.networking.DeviceConnectivityChecker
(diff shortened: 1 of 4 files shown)


tonidero:
We're going with a different approach so we won't be needing this.

PR #3982, opened 2026-08-17
ci(danger): fail PRs that grow a suppression baseline

AlvaroBrey:
- Adds a Danger rule that fails a PR when the net count of `<ID>` entries in `config/detekt/detekt-baseline.xml` goes up, so new findings get fixed or `@Suppress`'d at the smallest scope instead of silently baselined.
- Bypass  is the `skip-baseline-growth-check` label, mirroring the existing `skip-pr-lines-changed-check`.
- Records the rule in `AGENTS.md` Guardrails, and annotates the `detektAllBaseline` command so agents stop reaching for the baseline first.
- Second commit deletes `purchases/detekt-baseline.xml` and `examples/purchase-tester/detekt-baseline.xml`, which were **unused** (no gradle task configures them).

### Checklist

**Why Danger and not detekt**

detekt cannot express this. Custom rules extend `io.gitlab.arturbosch.detekt.api.Rule`, a `KtTreeVisitorVoid` whose callbacks are all Kotlin PSI nodes, and both detekt tasks filter the source set to `**/*.kt`. There is no file-system input surface and no git awareness, which is needed to know the file grew relative to the base branch. detekt applies the baseline as a post-analysis filter, so a "baseline grew" finding would itself be baseline-suppressible.

`scripts/api-check.sh`'s regenerate-and-diff pattern does not transfer either: regenerating the baseline reproduces the current violations, so the diff is clean even when the baseline grew in the PR.

**Implementation notes**

- Counts `<ID>` entries rather than diffing lines, because `detektAllBaseline` overwrites the whole file rather than merging, so a reformat or reorder would otherwise read as growth. A rename (one added, one removed) and a pure reorder both pass; two added against one removed fails.
- Matches on `File.basename` rather than a pinned path, so a baseline reintroduced elsewhere is still covered.
- Uses `git.diff_for_file`, which resolves against the PR's actual base. Worth noting the shared `RevenueCat/Dangerfile` helper `size_of_file_in_main` uses a bare `main` ref, which `revenuecat-mobile-ios` had to patch to `origin/main` because CI checkouts have no local `main`.

**Deleted baselines**

The detekt plugin is applied only to the root project (`build.gradle.kts:14`); the `subprojects {}` block only wires Dokka. Both `detektAll` and `detektAllBaseline` set `baseline` to `config/detekt/detekt-baseline.xml`, and nothing else in the repo references a baseline path. `purchases/detekt-baseline.xml` is still in detekt's pre-1.0 `<SmellBaseline>` schema with `Blacklist`/`Whitelist` elements; someone added an entry to it in #3150 in March 2026, which had no effect since the same entry was also added to the live baseline. The examples one is a 162-byte empty stub from 2021. Verified by deleting both and running `./gradlew detektAll`, which still passes.

**Verification**

The rule was exercised end to end by temporarily pushing a commit that added one baseline entry, then reverting it:

| Case | Danger |
| --- | --- |
| Baseline grows, no label | fails, one error, correct message |
| Same commit plus bypass label | passes |
| No baseline change | passes |

Plus six local assertions on the counting logic against a real baseline-growing patch and synthetic edge cases: removal-only, rename, pure reorder, `+++`/`---` diff headers, and net growth despite a removal.

**Follow-ups not in this PR**

- The 241 existing entries are untouched. This only stops new growth.
- `purchases-ios` has the same unguarded problem with `swiftlint-baseline.json`.
- Porting `rerun_danger_on_label.yml` from `purchases-ios` would remove the manual re-run step above.

</details>

---

> [!NOTE]
> **Low Risk**
> CI and documentation changes only; no runtime SDK or production Kotlin/Java logic is modified.
> 
> **Overview**
> Adds a **Danger** check that fails PRs when any `detekt-baseline.xml` gains more `<ID>` entries than it removes (net growth), so new detekt findings must be fixed or narrowly `@Suppress`'d instead of baselined. Bypass uses the `skip-baseline-growth-check` label, similar to the existing production-line limit.
> 
> **AGENTS.md** documents the guardrail and notes that `detektAllBaseline` is a last resort because CI enforces baseline shrink-only behavior.
> 
> Removes unused `purchases/detekt-baseline.xml` and `examples/purchase-tester/detekt-baseline.xml`; Gradle only uses `config/detekt/detekt-baseline.xml`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4314deb5d2b912f8ca7774cfa83984c5a4961806. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ajpallares (approved):
Nice! Thanks for adding this!

tonidero (approved):

PR #3983, opened 2026-08-17
feat(admob): add next-gen suspend interstitial loading

polmiro:
## Summary

- Add the coroutine counterpart to callback-based Next-Gen interstitial loading.
- Track successful and failed loads while returning Google's original result and preserving lifecycle callbacks.
- Require a callback for the callback-based overload so loaded ads cannot be silently discarded.

## Testing

- `./gradlew :feature:admob-next-gen:testDefaultsDebugUnitTest detektAll :feature:admob-next-gen:metalavaCheckCompatibilityDefaultsRelease`
- `./scripts/api-check.sh`


polmiro:
Folded into #3985.

PR #3985, opened 2026-08-17
feat(admob): add next-gen interstitial ad tracking

polmiro:
## Motivation

Expose Google Mobile Ads Next-Gen interstitial loading and event tracking through a public RevenueCat API.

This PR consumes the load- and event-tracking building blocks from #3955 and #3957 and wires them into the first format-specific integration.

## Description

- Add callback-based and suspending `AdTracker.loadAndTrackInterstitialAd` APIs for request-based loading through Google's Next-Gen `AdRequest` API.
- Add callback-based `loadAndTrackInterstitialAdFromResponse` for server-to-server ad responses, with an explicit ad unit ID for attribution.
- Track successful and failed loads while preserving Google's original callbacks and suspending result.
- Install lifecycle and paid-event tracking before a loaded ad reaches application code.
- Add `InterstitialAd.show(activity, placement)` for overriding the tracking placement when the display location is only known at show time.
- Add `setTrackingAdEventCallback` so applications can replace their callback without removing RevenueCat's wrapper.
- Publish and document the experimental Kotlin API.

## Tracking behavior

The load-time placement is used for load events and as the default for subsequent lifecycle events. A placement passed to `show` takes precedence for display, click, and revenue events.

Full-screen display is tracked from `onAdShowedFullScreenContent`; impressions are not also counted as displays. RevenueCat tracking runs before application callbacks.

Assigning `adEventCallback` directly after a tracked load replaces RevenueCat's wrapper. Applications should provide the callback while loading or use `setTrackingAdEventCallback`.

If `show(activity, placement)` is called after the callback was manually replaced, the ad is still shown but the placement override is ignored and a warning is logged.

---

> [!NOTE]
> **Medium Risk**
> New analytics surface on ad load/show paths; incorrect callback replacement or placement override could skew attribution, but changes are additive with documented guardrails.
> 
> **Overview**
> Adds **public RevenueCat tracking** for Google Mobile Ads Next-Gen **interstitial** ads: load helpers on `AdTracker`, show-time placement override, and safe callback replacement without dropping the tracking wrapper.
> 
> **Load paths:** callback-based `loadAndTrackInterstitialAd` (request and server-to-server `loadAndTrackInterstitialAdFromResponse` with explicit `adUnitId`), plus a **suspending** overload that returns Google's unchanged `AdLoadResult`. Successful loads install lifecycle/paid tracking **before** the app callback or return; failures still emit RevenueCat load-failure events with optional `placement`.
> 
[...]

PR #3986, opened 2026-08-17
feat(admob): track next-gen interstitial response loading

polmiro:
Adds callback-based Google Mobile Ads Next-Gen interstitial loading from server-to-server ad responses, including RevenueCat load tracking and lifecycle callback wiring.

`adUnitId` is required because the opaque response and failed-load result do not expose a reliable ad unit ID for RevenueCat attribution.

Validation: `:feature:admob-next-gen:testDefaultsDebugUnitTest`, `detektAll`, and `scripts/api-check.sh`.


polmiro:
Folded into #3985.

PR #3987, opened 2026-08-18
feat(checkpoints): Add the customer's purchases as a rules dimension

tonidero:
### Description

This PR adds a new dimension provider for customer info information that will contain all purchases facts themselves, without any processing, so the rules that come from the backend can use this info to process the data as needed.

The evaluation instant is a single `evaluatedAt` at the root of the scope, alongside the namespaces, rather than repeated on every record. Note that `some`/`all`/`none`/`filter`/`map` rebind `var` to the current item with no parent scope, so a predicate comparing a purchase against the instant reads that purchase by index (`customerInfo.purchases.0.expiresAt`) instead of iterating.

The app user ID is read once per evaluation and used both to request the customer info and as the reported `appUserId`, so a login or logout while the request is in flight cannot file one customer's purchases under another's ID. Same fix as iOS: https://github.com/RevenueCat/purchases-ios/pull/7461.

This ends up added to the rules engine like:
```
{
  "customerInfo": {
    "appUserId": "current_user",
    "entitlements": [
      {
        "billingIssueDetectedAt": 1714780800000,
        "expiresAt": 4102444800000,
        "identifier": "extra",
        "isActive": true,
        "isSandbox": true,
        "latestPurchasedAt": 1714521600000,
        "originalPurchasedAt": 1609459200000,
        "ownershipType": "PURCHASED",
        "periodType": "trial",
        "productIdentifier": "premium",
        "productPlanIdentifier": "monthly",
        "purchasedProductIdentifier": "premium:monthly",
        "store": "play_store",
        "unsubscribeDetectedAt": 1714694400000,
        "willRenew": false
      },
      {
        "billingIssueDetectedAt": 1714780800000,
        "expiresAt": 4102444800000,
        "identifier": "premium",
        "isActive": true,
        "isSandbox": true,
        "latestPurchasedAt": 1714521600000,
        "originalPurchasedAt": 1609459200000,
        "ownershipType": "PURCHASED",
        "periodType": "trial",
        "productIdentifier": "premium",
        "productPlanIdentifier": "monthly",
        "purchasedProductIdentifier": "premium:monthly",
        "store": "play_store",
        "unsubscribeDetectedAt": 1714694400000,
        "willRenew": false
      }
    ],
    "firstSeenAt": 1640995200000,
    "lastSeenAt": 1717200000000,
    "originalAppUserId": "original_user",
    "originalPurchasedAt": 1609459200000,
    "purchases": [
      {
        "autoResumeAt": 1717200000000,
        "billingIssueDetectedAt": 1714780800000,
        "displayName": "Premium Monthly",
        "expiresAt": 4102444800000,
        "gracePeriodExpiresAt": 1715299200000,
        "isActive": true,
        "isInGracePeriod": false,
        "isPaused": true,
        "isRefunded": true,
        "isSandbox": true,
        "kind": "subscription",
        "originalPurchasedAt": 1609459200000,
        "ownershipType": "PURCHASED",
        "periodType": "trial",
        "priceAmountMicros": 4990000,
        "priceCurrency": "USD",
        "productIdentifier": "premium",
        "productPlanIdentifier": "monthly",
        "purchasedAt": 1714521600000,
        "purchasedProductIdentifier": "premium:monthly",
        "refundedAt": 1714867200000,
        "status": "paused",
        "store": "play_store",
        "storeTransactionId": "GPA.0000-0000-0000-00000",
        "unsubscribeDetectedAt": 1714694400000,
        "willRenew": false
      },
      {
        "displayName": "100 Coins",
        "isSandbox": false,
        "kind": "nonSubscription",
        "originalPurchasedAt": 1677801600000,
        "priceAmountMicros": 1990000,
        "priceCurrency": "EUR",
        "productIdentifier": "coins",
        "purchasedAt": 1677801600000,
        "purchasedProductIdentifier": "coins",
        "store": "amazon",
        "storeTransactionId": "amzn.1234",
        "transactionIdentifier": "abc123"
      }
    ]
  },
  "evaluatedAt": 1718452800000
}
```

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint targeting and CustomerInfo lookup during rule evaluation, including identity pinning so a login mid-snapshot cannot mix two customers. Failures degrade to missing customer dimensions rather than changing purchase or auth flows.
> 
> **Overview**
> Local checkpoint rules can now target the current customer: identity, last-seen, purchases (newest first), and entitlements, under `customerInfo`. Fetch failures omit those fields instead of failing the whole snapshot; `appUserId` is still available without a network round-trip.
> 
> Snapshots pin one app user and one `evaluatedAt` at the root. Subscriber attributes are read for that same ID, and a login/logout mid-collection discards the snapshot. `last_seen` is parsed internally (not part of `CustomerInfo` equality) so “time since last open” is not the SDK’s own refresh time.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4876e0b1f8eb9af0d8dd5dd6ee699e7277df6ccd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -30,6 +30,7 @@ import com.revenuecat.purchases.common.errorLog
 import com.revenuecat.purchases.common.events.BackendStoredEvent
 import com.revenuecat.purchases.common.events.EventsManager
 import com.revenuecat.purchases.common.isDeviceProtectedStorageCompat
+import com.revenuecat.purchases.common.localrules.CustomerInfoDimensionProvider
 import com.revenuecat.purchases.common.localrules.DeviceDimensionProvider
 import com.revenuecat.purchases.common.localrules.LocalRulesEvaluator
 import com.revenuecat.purchases.common.localrules.RulesEngineLoggerBridge
@@ -53,6 +54,8 @@ import com.revenuecat.purchases.common.remoteconfig.RemoteConfigBlobStore
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigDiskCache
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigManager
 import com.revenuecat.purchases.common.remoteconfig.RemoteConfigTopicStore
+import com.revenuecat.purchases.common.safeResume
+import com.revenuecat.purchases.common.safeResumeWithException
 import com.revenuecat.purchases.common.uiconfig.UiConfigProvider
 import com.revenuecat.purchases.common.verification.SignatureVerificationMode
 import com.revenuecat.purchases.common.verification.SigningManager
@@ -81,6 +84,7 @@ import com.revenuecat.purchases.utils.PurchaseParamsValidator
 import com.revenuecat.purchases.utils.UrlConnectionFactory
 import com.revenuecat.purchases.utils.isAndroidNOrNewer
 import com.revenuecat.purchases.virtualcurrencies.VirtualCurrencyManager
+import kotlinx.coroutines.suspendCancellableCoroutine
 import java.net.URL
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.Executors
@@ -398,6 +402,10 @@ internal class PurchasesFactory(
                             identityManager.currentAppUserID,
                         )
                     },
+                    CustomerInfoDimensionProvider(
+                        appUserId = { identityManager.currentAppUserID },
+                        customerInfo = { Purchases.sharedInstance.purchasesOrchestrator.awaitCustomerInfo() },
+                    ),
                 ),
             )
 
@@ -693,3 +701,22 @@ internal class PurchasesFactory(
         ): Boolean = diagnosticsEnabled && !uiPreviewMode
     }
 }
+
+/**
+ * The cache when it is warm, the initial fetch otherwise, so the first checkpoint of a session can still read
+ * customer dimensions.
+ *
+ * Goes through the orchestrator rather than `Purchases.awaitCustomerInfo`, which only exists in the `defaults`
+ * source set: local rule evaluation is wired for both flavors.
+ */
+private suspend fun PurchasesOrchestrator.awaitCustomerInfo(): CustomerInfo =
+    suspendCancellableCoroutine { continuation ->
+        getCustomerInfo(
+            fetchPolicy = CacheFetchPolicy.default(),
+            trackDiagnostics = false,
+            callback = receiveCustomerInfoCallback(
+                onSuccess = { continuation.safeResume(it) },
+                onError = { continuation.safeResumeWithException(PurchasesException(it)) },
+            ),
+        )
+    }
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -404,7 +404,9 @@ internal class PurchasesFactory(
                     },
                     CustomerInfoDimensionProvider(
                         appUserId = { identityManager.currentAppUserID },
-                        customerInfo = { Purchases.sharedInstance.purchasesOrchestrator.awaitCustomerInfo() },
+                        customerInfo = { appUserID ->
+                            Purchases.sharedInstance.purchasesOrchestrator.awaitCustomerInfo(appUserID)
+                        },
                     ),
                 ),
             )
@@ -706,9 +708,10 @@ internal class PurchasesFactory(
  * The cache when it is warm, the initial fetch otherwise. Goes through the orchestrator because
  * `Purchases.awaitCustomerInfo` only exists in the `defaults` source set, while this is compiled for both flavors.
  */
-private suspend fun PurchasesOrchestrator.awaitCustomerInfo(): CustomerInfo =
+private suspend fun PurchasesOrchestrator.awaitCustomerInfo(appUserID: String): CustomerInfo =
     suspendCancellableCoroutine { continuation ->
         getCustomerInfo(
+            appUserID = appUserID,
             fetchPolicy = CacheFetchPolicy.default(),
             trackDiagnostics = false,
             callback = receiveCustomerInfoCallback(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -403,7 +403,7 @@ internal class PurchasesFactory(
                         )
                     },
                     CustomerInfoDimensionProvider(
-                        appUserId = { identityManager.currentAppUserID },
+                        currentAppUserId = { identityManager.currentAppUserID },
                         customerInfo = { appUserID ->
                             Purchases.sharedInstance.purchasesOrchestrator.awaitCustomerInfo(appUserID)
                         },
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -703,11 +703,8 @@ internal class PurchasesFactory(
 }
 
 /**
- * The cache when it is warm, the initial fetch otherwise, so the first checkpoint of a session can still read
- * customer dimensions.
- *
- * Goes through the orchestrator rather than `Purchases.awaitCustomerInfo`, which only exists in the `defaults`
- * source set: local rule evaluation is wired for both flavors.
+ * The cache when it is warm, the initial fetch otherwise. Goes through the orchestrator because
+ * `Purchases.awaitCustomerInfo` only exists in the `defaults` source set, while this is compiled for both flavors.
  */
 private suspend fun PurchasesOrchestrator.awaitCustomerInfo(): CustomerInfo =
     suspendCancellableCoroutine { continuation ->
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -409,6 +409,7 @@ internal class PurchasesFactory(
                         },
                     ),
                 ),
+                currentAppUserId = { identityManager.currentAppUserID },
             )
 
             val customerInfoUpdateHandler = CustomerInfoUpdateHandler(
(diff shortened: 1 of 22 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/CustomerInfoDimensionProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+     * no iteration at all. Subscriptions are ordered by product before the merge, so purchases that share a date
+     * still come out in a defined order.
+     */
+    private fun CustomerInfo.purchaseRecords(date: Date): List<Map<String, RulesDimensionValue>> {
+        val subscriptions = subscriptionsByProductIdentifier.values
+            .sortedBy { subscription -> subscription.productIdentifier }
+            .map { subscription -> subscription.record(date) }
+        // Already sorted by purchase date by `CustomerInfo`.
+        val transactions = nonSubscriptionTransactions.map { transaction -> transaction.record(date) }
+        return (subscriptions + transactions).sortedByDescending { record -> record.dateOrNull(KEY_PURCHASED_AT) }
```
tonidero, line 74: Note how we sort by  productId first, but then we resort by purchase date, so newer ones come first. Might not be needed, but felt like a good convention to have.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -687,3 +695,19 @@ internal class PurchasesFactory(
         ): Boolean = diagnosticsEnabled && !uiPreviewMode
     }
 }
+
+/**
+ * The cache when it is warm, the initial fetch otherwise. Goes through the orchestrator because
+ * `Purchases.awaitCustomerInfo` only exists in the `defaults` source set, while this is compiled for both flavors.
+ */
+private suspend fun PurchasesOrchestrator.awaitCustomerInfo(): CustomerInfo =
```
tonidero: Had to keep this since for the CEC mode, we don't have the `awaitCustomerInfo` function in the public API. Otherwise, I could just call `Purchases.sharedInstance.awaitCustomerInfo`.

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/CustomerInfoDimensionProvider.kt
```diff
@@ -0,0 +1,302 @@
...
+
+internal class CustomerInfoDimensionProvider(
+    private val appUserId: () -> String?,
+    private val customerInfo: suspend () -> CustomerInfo,
+) : RulesDimensionProvider {
+
+    override val namespace: RulesDimensionNamespace = RulesDimensionNamespace.CustomerInfo
+
+    override suspend fun dimensions(date: Date): Map<String, RulesDimensionValue> =
+        appUserIdDimension() + customerInfoDimensions(date)
```
facumenzella: If a login happens in between these two (which reading this I am assuming they go sequentially) don't we end up combining two different users?

we hit this on iOS in RevenueCat/purchases-ios#7461, fix was reading the id once and passing it into the fetch (0700e43cd)

facumenzella: This makes it more visible I think https://github.com/RevenueCat/purchases-android/pull/3987/changes#diff-a5d1b55d8040d4e96c30987e585413a0e34716b97e6e7f8871525f23f6d7c5e1R405-R408

tonidero: Hmm it's indeed possible... FWIW, getting the appUserId is synchronous... So I'm thinking the login would need to happen before starting the customer info request. Will try to refactor this. Thanks!

tonidero: Fixed as part of https://github.com/RevenueCat/purchases-android/pull/3987/commits/3abca830c7254a1f7ff4c1ccb87cc4ad4f28cfcf

tonidero: @facumenzella I had to do some additional changes due to Cursor's reported issue which was also right: https://github.com/RevenueCat/purchases-android/pull/3987/commits/da29281833b1ab7262b9469a6bfde3fe8e845271

Basically, if the user changes midway, we basically ignore it and don't fill the value in the SDK. The evaluation will likely fail, but I think that is fine in this situation for now.

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -28,32 +28,60 @@ internal sealed class RulesDimensionResolutionException(message: String) : Excep
     internal data class ConflictingDimension(
         val path: String,
     ) : RulesDimensionResolutionException("two dimension providers supplied '$path'")
+
+    internal object AppUserChanged : RulesDimensionResolutionException(
+        "the app user kept changing while the dimensions were being collected",
```
facumenzella:
```suggestion
the app user changed while the dimensions were being collected
```

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -28,31 +28,60 @@ internal sealed class RulesDimensionResolutionException(message: String) : Excep
...
+ * The three failure modes fail the whole snapshot instead of silently degrading a rule to a non-match, which
+ * would be indistinguishable from a customer who genuinely does not match: a provider that cannot produce its
+ * values, two providers claiming the same path, and an app user that would not stay still. Cancellation is none
+ * of them, and propagates.
  */
 internal class RulesDimensionResolver(
     private val providers: List<RulesDimensionProvider>,
     private val dateProvider: DateProvider = DefaultDateProvider(),
+    /** No identity source means nothing can change: the same value the SDK reports for an uncached user. */
+    private val currentAppUserId: () -> String = { "" },
```
facumenzella, line 54: following up on the slack thread so it's not lost here: what if the resolver reads the app user once and passes it down, instead of each provider reading the live value?

```kotlin
internal data class RulesDimensionContext(val date: Date, val appUserId: String)
```

`dimensions(date)` becomes `dimensions(context)`, and then there's nothing to compare, no retry and no extra error reason. same shape as ios `DimensionContext`. or is there a case where the retry earns it that I'm not seeing? 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -28,31 +28,60 @@ internal sealed class RulesDimensionResolutionException(message: String) : Excep
...
+            // A provider that cannot produce its values and two providers claiming the same path are both
+            // configuration bugs. Asking again does not fix either.
+            if (result.isFailure) return result
+            if (currentAppUserId() == appUserId) return result
+            warnLog { "The app user changed while the dimensions were being collected, so they were discarded." }
+        }
+        return Result.failure(RulesDimensionResolutionException.AppUserChanged)
+    }
+
+    /** One pass over every provider, against one reference instant of its own. */
```
facumenzella: not a real problem right now, just to have it in mind: the retry redoes device and store too, and those can't change with the user. one more small reason to pass the id in instead

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/RulesDimensionResolver.kt
```diff
@@ -81,11 +110,29 @@ internal class RulesDimensionResolver(
...
+        private const val ATTEMPTS = 2
+
+        /**
+         * The instant the snapshot was taken, at the root of the scope rather than repeated on every record.
+         *
+         * Not readable from inside an iteration operator: `some` and its siblings rebind `var` to the current item
+         * with no parent scope, so a predicate comparing a purchase against it reads that purchase by index, as
+         * `customerInfo.purchases.0.expiresAt`.
+         */
+        const val KEY_EVALUATED_AT = "evaluatedAt"
```
facumenzella, line 120: better at the root than copied onto every record. I'll follow on ios so it reads the same on both sides

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/CustomerInfoDimensionProvider.kt
```diff
@@ -0,0 +1,301 @@
...
+     *
+     * The records are built inside the same guard as the read, because a [CustomerInfo]'s purchases are parsed out
+     * of its raw payload on first access rather than when it is constructed, so that is where a malformed payload
+     * surfaces.
+     *
+     * Asking for one app user is not enough to be answered about them: on a cold cache the SDK syncs pending
+     * purchases first, and that sync reads the current app user for itself. Catching that is
+     * [RulesDimensionResolver]'s job, since it watches the whole collection rather than this read alone, and it
+     * can take the snapshot again instead of leaving a customer half-described.
+     */
```
vegaro, line 37: Can we make this comment a bit more concrete? It's a bit verbose. I had a bit of trouble following it.

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/CustomerInfoDimensionProvider.kt
```diff
@@ -0,0 +1,301 @@
...
+            warnLog { "The customer info is unavailable, so customer dimensions can't be evaluated: $e" }
+            emptyMap()
+        }
+    }
+
+    private fun CustomerInfo.dimensions(date: Date): Map<String, RulesDimensionValue> = buildMap {
+        putString(KEY_ORIGINAL_APP_USER_ID, originalAppUserId)
+        putDate(KEY_FIRST_SEEN_AT, firstSeen)
+        // The date the backend last answered us, which is when it last saw this customer through this device.
+        putDate(KEY_LAST_SEEN_AT, requestDate)
```
vegaro: I am unsure this is what we want to do.<br><br>`requestDate` is when the backend last answered this device, not when the customer was last seen, so I think this ends up meaning something different from what the name promises, as it means "the moment the user opens the app" since we refresh CustomerInfo on app foreground. Any rule evaluated in that session will see "seen just now". The only way to get an old value is to be offline which is when we would use a cached CustomerInfo with an older request date, or for the checkpoint to beat the startup refresh.

Also as a developer what I'd want from `lastSeenAt` is "how long since they last opened the app", so `lastSeenAt older than 7 days` targets a user that hasn't opened the app in a while.<br><br>Looking at the backend, it looks like we are actually sending `subscriber.last_seen` is in the payload, we just don't parse it into `CustomerInfo`. Would that work better?

tonidero: Hmm you're indeed right... I think it makes sense, will do those changes 🙏

tonidero:
We're taking a different approach here. We might need something similar in the future, but for now, will close it and can reopen later if/when needed

PR #3884, opened 2026-08-03
feat(ads): fire reward-verification tracking events from the poll

peterporfy:
### Description

This change introduces a new `RewardedAdTrackingMetadata` class as a way to pass ad information to the reward polling function - and so to make it possible to _actually_ track the lifecycle of rewarded ads.

The metadata is optional and it is driving if the tracking is happening or not (someone can simply omit it like today to not have tracking).

---

> [!NOTE]
> **Medium Risk**
> Changes the reward-verification poll path and ties it to ad analytics, but optional metadata and default-null preserve existing callers; mis-typed metadata could skew event attribution.
> 
> **Overview**
> **Reward polling can now emit ad lifecycle analytics** when callers pass optional **`RewardedAdTrackingMetadata`** (network, mediator, format, placement, ad unit, impression id). **`pollRewardVerification`** and **`awaitPollRewardVerification`** accept this metadata; omitting it keeps today’s behavior with **no reward-tracking events**.
> 
> During a tracked poll, **`Purchases`** fires **`AdTracker`** events: earned-unverified at start, then verified / per-grant granted / failed-to-verify based on the poll **`Outcome`**. **`NoReward`** does not produce a granted event. Poll **cancellation** still propagates **`CancellationException`** but records a failed-to-verify with a **Cancelled** reason when metadata was provided.
> 
> Internal **`Outcome.Failed`** cases now map to **`AdRewardFailureReason`** for analytics, including a new **`Cancelled`** failure type. Public API surface and AdMob integration tests were updated for the expanded **`pollRewardVerification`** signature (**`AdCaptureMethod.MANUAL`** on manual polls).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5559b3b96dd41122c2527b2e67c85b8f59484693. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #3991, opened 2026-08-18
Announce a public API change in the SDK API feed once per PR

AlvaroBrey:
- The feed announced the same summary on every CI run of a PR touching a signature file. Now it announces once per distinct summary.
- Dedup compares against the channel's newest message about this PR, so a PR that reverts to a surface it announced earlier is announced again.
- When the channel says nothing , a `<!-- api-diff:<hash> -->` marker on the PR comment decides. It is written only after the post succeeds, so a Slack outage cannot silence the next run.
- Needs `channels:history` on the bot token and the channel ID, not `#name`, in `SLACK_CHANNEL_SDK_NEW_API`. Without either, dedup still works across pushes through the marker and the PR gets a warning naming the reason.

- Companion change on `purchases-ios`: https://github.com/RevenueCat/purchases-ios/pull/7431

### Checklist

### Motivation

`#3976` shipped the announcement without the dedup: `run` posted whenever the diff against the PR base was non-empty, which is every run of a PR that already changed the API surface. `#3884` was announced three times with byte-identical content, twice within seven seconds.

Those seven seconds are a second cause. Each push creates two pipelines (26535 and 26537, 26496 and 26498, and so on): `build-test-deploy` on the older one is auto-canceled, but `danger` finishes in ~25s and posts first. That rules out the PR comment as the only store, since Danger writes it at the end of the run and two concurrent runs would both read the previous push's state.

### Description

- `announcement_state` returns `:same` / `:different` / `:unknown`. Identity is the PR link plus the platform label, both already in the message; modules are excluded because they change with what the PR touches.
- `:unknown` is the only state that consults the comment marker, so both stores answer one question: what was announced last.
- `announce` returns `[:posted | :duplicate | :failed, reason]`, and the marker is recorded for the first two only.
- The comment lookup is injected from the `Dangerfile`, since Danger's Octokit client is the only thing that can list comments, and it receives the marker text rather than the bare hash so the helper owns the marker shape on both sides.

Not visible in the diff: the marker sits in the `<details>` block Danger rewrites every run, so it always reflects the last announcement rather than accumulating.

</details>

---

> [!NOTE]
> **Medium Risk**
> Changes CI Danger behavior and Slack integration for API announcements; misconfiguration (channel name vs ID or missing scopes) can cause duplicate posts or warnings, but no runtime SDK or auth impact.
> 
> **Overview**
> Stops **duplicate Slack posts** to the SDK API feed when Danger reruns on the same public API diff. Announcements now go out **once per distinct summary** for a PR, while still **re-announcing** if the API surface changes again (including reverting to an earlier announced state).
> 
> Dedup first checks Slack **`conversations.history`** for this PR’s latest message (matched by PR link and Android platform label) and skips posting when the text matches. When history is unavailable—wrong channel format, missing `channels:history`, errors, or concurrent CI siblings—it falls back to a **`<!-- api-diff:<hash> -->`** marker on the PR comment (lookup injected from the `Dangerfile`). The hash fingerprints the **full Slack message**, not raw declaration lists, so removals and additions don’t collide. The marker is recorded only after a **successful** post so a Slack outage doesn’t block a later announcement.
> 
> Operational notes: **`SLACK_CHANNEL_SDK_NEW_API` must be a channel ID** (not `#name`) for history. Warnings replace the old `slack_error` field when Slack fails or when a post may duplicate because neither Slack nor the comment marker could confirm prior announcement.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4c4c48909a304a8be82c1becc170f20622b0f605. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -112,8 +112,7 @@ begin
 
   if api_diff
     markdown(api_diff[:comment])
-    warn("The public API changed, but it was not announced in the SDK API feed: #{api_diff[:slack_error]}.") if api_diff[:slack_error]
-    warn("Could not read the SDK API feed channel, so this change may be announced twice: #{api_diff[:degraded_dedupe]}.") if api_diff[:degraded_dedupe]
+    warn(api_diff[:warning]) if api_diff[:warning]
   end
 rescue StandardError => e
   # `warn` is Danger's DSL: surfaces on the PR without failing the run.
--- a/Dangerfile
+++ b/Dangerfile
@@ -102,11 +102,18 @@ begin
     changed_files: git.modified_files + git.added_files + git.deleted_files,
     patch_for: ->(file) { git.diff_for_file(file)&.patch },
     pull_request_link: "<#{github.pr_json['html_url']}|##{github.pr_json['number']}>",
+    announced_in_comment: lambda do |marker|
+      github.api.issue_comments(github.pr_json["base"]["repo"]["full_name"], github.pr_json["number"])
+            .any? { |comment| comment.body.to_s.include?(marker) }
+    rescue StandardError
+      false
+    end,
   )
 
   if api_diff
     markdown(api_diff[:comment])
     warn("The public API changed, but it was not announced in the SDK API feed: #{api_diff[:slack_error]}.") if api_diff[:slack_error]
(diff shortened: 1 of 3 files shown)


tonidero (approved):
🙇 Thank you!!

PR #3994, opened 2026-08-18
Track checkpoint hits through the analytics events pipeline

vegaro:
Send checkpoint hits through the existing batched events endpoint so Khepri can register checkpoints and publish analytics without adding another SDK endpoint.

https://github.com/RevenueCat/khepri/pull/24035
https://github.com/RevenueCat/purchases-ios/pull/7447

---

> [!NOTE]
> **Low Risk**
> Adds a new analytics event type following existing paywall/workflow patterns; no changes to purchase, auth, or billing logic.
> 
> **Overview**
> **Checkpoint hits are now analytics events.** When `resolveCheckpoint` runs, it records a `CheckpointEvent` (identifier + timestamp) via the same `track` path used for paywalls and workflows, *before* the workflow resolver runs.
> 
> The PR adds the full serialization stack for Khepri: `BackendEvent.Checkpoint` (`checkpoint_hit`, schema v1), `BackendStoredEvent.Checkpoint`, polymorphic JSON registration in `JsonProvider` and `EventsManager`, and `CheckpointEvent.toBackendStoredEvent` with app user and session IDs.
> 
> Tests cover Khepri-compatible JSON encoding/decoding, file persistence in `EventsManager`, and that resolving a checkpoint triggers exactly one tracked event.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b71bc6919bf83ec25e49d504ece83f4edcd75540. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -436,7 +437,15 @@ internal class PurchasesOrchestrator(
...
         customVariables: Map<String, RulesDimensionValue>,
-    ): CheckpointResolution = checkpointWorkflowResolver.resolve(checkpointIdentifier, customVariables)
+    ): CheckpointResolution {
+        track(
+            CheckpointEvent(
+                identifier = checkpointIdentifier,
+                timestamp = dateProvider.now,
+            ),
+        )
+        return checkpointWorkflowResolver.resolve(checkpointIdentifier, customVariables)
```
facumenzella, line 447: Shouldn't we check if there was an error before tracking? or track the error? 🤔

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -436,7 +437,15 @@ internal class PurchasesOrchestrator(
     suspend fun resolveCheckpoint(
         checkpointIdentifier: String,
         customVariables: Map<String, RulesDimensionValue>,
```
facumenzella, line 439: No need to track custom variables?

vegaro: in what sense?

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
vegaro: The intention is to track the checkpoint hit itself, regardless of the resolution outcome. Khepri uses `checkpoint_hit` both to register the `(app_config_id, identifier)` pair and to publish the user event, so moving this after `resolve` would drop hits whenever resolution throws. `NoAction` also includes expected outcomes such as disabled, unknown checkpoint, and no audience match, so it is not a success/error signal. The current backend schema has no result or error field; if we want resolution observability, I think that should be modeled separately without replacing the hit event.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
facumenzella: I know it's not in the definition but I wondered if we might want to track those on every hit

facumenzella: but we can add that later

facumenzella (approved):
legit to me

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: FWIW, I think tracking the outcome would also be VERY useful. But of course can come on a separate PR. I think we could have 2 more events. One for the outcome of the rule evaluation, but before executing anything (presenting workflow/returning offering). Another one after the outcome is fully completed... At least for those presenting a workflow paywall.

vegaro: Right, resolution and result. That would be perfect

PR #3996, opened 2026-08-18
feat(checkpoints): Add the app's subscriber attributes as a rules dimension

tonidero:
### Description

Adds a `subscriberAttributes` dimension provider, so everything the app has told the SDK about the customer via `setAttributes` is readable by a local rule. It would look like:

```json
{
  "subscriberAttributes": {
    "$email": {
      "evaluatedAt": 1718452800000,
      "updatedAt": 1714521600000,
      "value": "jane@example.com"
    },
    "goal": {
      "evaluatedAt": 1718452800000,
      "updatedAt": 1718366400000,
      "value": "lose_weight"
    },
    "seats": {
      "evaluatedAt": 1718452800000,
      "updatedAt": 1714780800000,
      "value": "3"
    }
  }
}
```

---

> [!NOTE]
> **Medium Risk**
> Changes how checkpoint audiences resolve using live subscriber attributes and identity timing; mis-wiring could target the wrong user, but attribute read failures are isolated and do not abort evaluation.
> 
> **Overview**
> Adds a **`subscriberAttributes`** rules dimension so checkpoint audiences can target values from `Purchases.setAttributes`. Each attribute appears as a nested object (`value`, `updatedAt`, `evaluatedAt`) readable via paths like `subscriberAttributes.goal.value`.
> 
> **Rules engine plumbing:** Introduces `RulesDimensionValue.ObjectValue` for dot-path access (distinct from list iteration). `RulesDimensionResolver` drops unreachable attribute names (empty or containing `.`) with a warning instead of failing the snapshot. Provider diagnostics now use **`namespace`** instead of a separate `identifier`.
> 
> **Wiring:** `LocalRulesEvaluator` is built **after** `IdentityManager` so attributes are loaded for `identityManager.currentAppUserID`, not a stale singleton. Read failures for subscriber attributes degrade to an empty namespace (other dimensions still evaluate), matching store dimension behavior.
> 
> Unit and integration tests cover predicates, deletions/tombstones, per-user isolation, and the `setAttributes` → disk cache → resolver path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 600517359fafc4a58d5a58d171f19f1223d361a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -355,18 +355,6 @@ internal class PurchasesFactory(
             val checkpointsConfigProvider = remoteConfigManager?.let {
                 CheckpointsConfigProvider(it)
             }
-            RulesEngine.setLogger(RulesEngineLoggerBridge)
-            val localRulesEvaluator = LocalRulesEvaluator(
-                providers = listOf(
-                    DeviceDimensionProvider(appConfig, localeProvider),
-                    StoreDimensionProvider { Purchases.sharedInstance.awaitStorefrontCountryCode() },
-                    SubscriberAttributesDimensionProvider {
-                        subscriberAttributesCache.getAllStoredSubscriberAttributes(
-                            Purchases.sharedInstance.purchasesOrchestrator.appUserID,
-                        )
-                    },
-                ),
-            )
             if (remoteConfigManager != null && uiConfigProvider != null && workflowsConfigProvider != null) {
                 remoteConfigManager.registerListener(uiConfigProvider)
                 remoteConfigManager.registerListener(workflowsConfigProvider)
@@ -388,6 +376,24 @@ internal class PurchasesFactory(
                 offlineEntitlementsManager,
                 dispatcher,
                 uiPreviewMode = appConfig.uiPreviewMode,
+            )
+
+            // Built after the identity manager so a dimension source reads the app user ID from this instance
+            // rather than from whichever one is the singleton by the time a checkpoint is resolved. The evaluator
+            // is a leaf, only consumed from there, so where it is built is otherwise unconstrained.
+            RulesEngine.setLogger(RulesEngineLoggerBridge)
+            val localRulesEvaluator = LocalRulesEvaluator(
+                providers = listOf(
+                    DeviceDimensionProvider(appConfig, localeProvider),
+                    // Only read during a checkpoint evaluation, so the instance is configured by then. Same
+                    // reasoning as CheckpointWorkflowResolverImpl's getOfferings.
+                    StoreDimensionProvider { Purchases.sharedInstance.awaitStorefrontCountryCode() },
+                    SubscriberAttributesDimensionProvider {
+                        subscriberAttributesCache.getAllStoredSubscriberAttributes(
+                            identityManager.currentAppUserID,
+                        )
+                    },
+                ),
             )
 
             val customerInfoUpdateHandler = CustomerInfoUpdateHandler(
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -359,11 +359,7 @@ internal class PurchasesFactory(
             val localRulesEvaluator = LocalRulesEvaluator(
                 providers = listOf(
                     DeviceDimensionProvider(appConfig, localeProvider),
-                    // Only read during a checkpoint evaluation, so the instance is configured by then. Same
-                    // reasoning as CheckpointWorkflowResolverImpl's getOfferings.
                     StoreDimensionProvider { Purchases.sharedInstance.awaitStorefrontCountryCode() },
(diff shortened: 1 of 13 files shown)


purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/SubscriberAttributesDimensionProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.revenuecat.purchases.subscriberattributes.SubscriberAttribute
+import java.util.Date
+
+internal class SubscriberAttributesDimensionProvider(
+    private val storedAttributes: () -> Map<String, SubscriberAttribute>,
+) : RulesDimensionProvider {
+
+    override val identifier: String = "subscriberAttributes"
+
+    override val namespace: RulesDimensionNamespace = RulesDimensionNamespace.SubscriberAttributes
```
rickvdl, line 14: Should we combine the identifier and namespace? Having an identifier on the namespace type? That's what we did on iOS recently

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
```diff
@@ -358,9 +359,12 @@ internal class PurchasesFactory(
             val localRulesEvaluator = LocalRulesEvaluator(
                 providers = listOf(
                     DeviceDimensionProvider(appConfig, localeProvider),
-                    // Only read during a checkpoint evaluation, so the instance is configured by then. Same
-                    // reasoning as CheckpointWorkflowResolverImpl's getOfferings.
                     StoreDimensionProvider { Purchases.sharedInstance.awaitStorefrontCountryCode() },
+                    SubscriberAttributesDimensionProvider {
+                        subscriberAttributesCache.getAllStoredSubscriberAttributes(
+                            Purchases.sharedInstance.purchasesOrchestrator.appUserID,
```
rickvdl: Could / should we use the identityManager directly here rather than going through the shared instance? May improve testability?

rickvdl: After that we should also be able to add a small integration test

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/SubscriberAttributesDimensionProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+            warnLog { "The subscriber attributes are unavailable, so they can't be evaluated: $e" }
+            return emptyMap()
+        }
+        return attributes.values.mapNotNull { attribute -> attribute.dimension(date) }.toMap()
+    }
+
+    @Suppress("ReturnCount")
+    private fun SubscriberAttribute.dimension(date: Date): Pair<String, RulesDimensionValue>? {
+        val name = key.backendKey
+        if (name.isEmpty() || name.contains(DIMENSION_PATH_SEPARATOR)) {
```
rickvdl: I know the `name.contains(DIMENSION_PATH_SEPARATOR)` is just an extra check since the attributes shouldn't contain it to begin with, but should we move this to `RulesDimensionResolver.addDimensions()`? I think it would benefit all dimension providers?

rickvdl (approved):
Nice! Couple of small points

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/SubscriberAttributesDimensionProvider.kt
tonidero: Good call! Will move this 👍

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Hmm you're probably right. Will try to do that 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/localrules/SubscriberAttributesDimensionProvider.kt
tonidero: Ohh right, I forgot to apply it in Android 🤦 thanks!

tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3996/commits/4c478e24d557dec69111696a37fbf807711c587b

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
tonidero: Changed in https://github.com/RevenueCat/purchases-android/pull/3996/commits/935f54767bc361338aa3af3e4814764d936a2e4f and added test in https://github.com/RevenueCat/purchases-android/pull/3996/commits/da051e9ad0be87ec07a147755095fceb9a5c3fba

PR #4019, opened 2026-08-19
Add reward verification runtime

polmiro:
### Checklist

### Motivation
Reward verification needs lifecycle-safe state and polling before the next-generation AdMob adapter can expose it publicly.
This is the first reward verification layer and targets `next-gen-admob-sdk`.

### Description
Adds a per-configuration runtime that correlates AdMob response IDs with client transaction IDs, polls RevenueCat, delivers callbacks once on the main thread, and fails safely on cancellation or missing correlation.
Runtime tests cover verified results, missing IDs, callback threading, and cancellation cleanup; validated with the focused runtime test suite.

---

> [!NOTE]
> **Medium Risk**
> New reward-verification flow affects when developers grant rewards, but behavior is internal, fails closed, and is covered by unit tests; full adapter integration is still ahead.
> 
> **Overview**
> Introduces **lifecycle-scoped reward verification** for the next-gen AdMob module: a new `RewardVerificationRuntime` that ties each ad response ID to a client transaction ID, polls RevenueCat via `awaitPollRewardVerification` on a background coroutine scope, and reports **started/completed** callbacks **once** on the main thread.
> 
> If verification was never set up (missing ad response ID or unregistered transaction ID), it logs guidance and completes with **failed** without starting the poll. **Teardown** via `close()` cancels in-flight work, clears mappings, and still delivers a single failed completion when verification was cancelled mid-poll.
> 
> Adds a small **`runOnMainIfPresent`** helper for main-thread delivery and shared log strings. **Robolectric tests** cover happy path threading, missing setup, and cancellation during poll.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 097bf2e90dbb7bea1c806e40388d1011570006c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/feature/admob-next-gen/src/main/kotlin/com/revenuecat/purchases/admob/nextgen/rewardverification/RewardVerificationRuntime.kt
+++ b/feature/admob-next-gen/src/main/kotlin/com/revenuecat/purchases/admob/nextgen/rewardverification/RewardVerificationRuntime.kt
@@ -44,7 +44,6 @@ internal class RewardVerificationRuntime(
         rewardVerificationCompleted: (RewardVerificationResult) -> Unit,
     ) {
         val clientTransactionId = adResponseId?.let { getClientTransactionId(it) }
-        warnAndAssertIfMissingClientTransactionId(clientTransactionId)
 
         val completionDelivered = AtomicBoolean(false)
         fun deliverOnce(result: RewardVerificationResult) {
@@ -53,7 +52,8 @@ internal class RewardVerificationRuntime(
             }
         }
 
-        if (adResponseId == null || clientTransactionId == null) {
+        if (clientTransactionId == null) {
+            Logger.w(RewardVerificationStrings.NOT_SET_UP_FOR_AD)
             deliverOnce(RewardVerificationResult.failed)
             return
         }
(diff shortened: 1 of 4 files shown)


peterporfy (commented):
looks good to me!

feature/admob-next-gen/src/main/kotlin/com/revenuecat/purchases/admob/nextgen/rewardverification/RewardVerificationRuntime.kt
```diff
@@ -0,0 +1,117 @@
...
+    private fun notifyCompleted(
+        result: RewardVerificationResult,
+        rewardVerificationCompleted: (RewardVerificationResult) -> Unit,
+    ) {
+        runOnMainIfPresent(mainHandler) {
+            rewardVerificationCompleted(result)
+        }
+    }
+
+    private fun warnAndAssertIfMissingClientTransactionId(clientTransactionId: String?) {
```
AlvaroBrey: perhaps I'm missing something but
- the `Assert` is throwing me off, this doesn't really assert anything in the traditional search, it just logs if something is missing
- this will log `ENABLE_REQUIRED_BEFORE_SHOW` even if the cause is `adResponseId == null`

AlvaroBrey (approved):
one minor comment, otherwise nice

feature/admob-next-gen/src/main/kotlin/com/revenuecat/purchases/admob/nextgen/rewardverification/RewardVerificationRuntime.kt
polmiro: Good catch, I have removed that assertion terminology but also improved the naming of this case so it's clearer. It's not possible for this code to execute if there is no ad instance, and there will aways be an adResponseId in those cases. Updated the string key to NOT_SET_UP_FOR_AD which is more fitting and fired it only in the one spot it should.

PR #4014, opened 2026-08-19
feat(admob): add next-gen banner ad tracking

polmiro:
## Motivation

Expose Google Mobile Ads Next-Gen banner loading and event tracking through a public RevenueCat API.

## Description

- Add `AdView.loadAndTrackAd` and `AdTracker.loadAndTrackBannerAd` for request-based loading through the recommended Next-Gen `AdView` API.
- Track successful and failed loads while preserving Google callbacks.
- Install lifecycle and refresh tracking before a loaded banner reaches application code.
- Track automatic refresh successes and failures against the current creative and original load placement.
- Add tracking-safe setters for banner event and refresh callbacks.
- Publish and document the experimental Kotlin API.

## Tracking behavior

The load-time placement is used throughout the banner lifecycle, including automatic refreshes. Banner display is tracked from `onAdImpression`; full-screen overlay callbacks are forwarded without also counting a display. RevenueCat tracking runs before application callbacks.

Assigning `adEventCallback` or `bannerAdRefreshCallback` directly after a tracked load replaces the RevenueCat wrapper. Applications should provide callbacks while loading or use the tracking-safe setters.

---

> [!NOTE]
> **Low Risk**
> Additive experimental API on the AdMob Next-Gen adapter with tests; main integration risk is apps bypassing tracking-safe setters and dropping analytics.
> 
> **Overview**
> Adds **RevenueCat ad tracking for Google Mobile Ads Next-Gen banners** loaded through `AdView`, complementing the existing banner preloader helpers.
> 
> New entry points are **`AdView.loadAndTrackAd`** and **`AdTracker.loadAndTrackBannerAd`**. They wrap `AdView.loadAd` with `TrackingAdLoadCallback`, install lifecycle and refresh wrappers on the loaded `BannerAd` before the app’s load callback runs, and emit load success/failure (including placement and ad unit) while still forwarding Google callbacks.
> 
> **`BannerAd.setTrackingAdEventCallback`** and **`setTrackingBannerAdRefreshCallback`** let apps swap delegates without replacing RevenueCat’s wrappers. Shared **`installTrackingCallbacks`** is extracted so **`BannerAdPreloader.pollAndTrackAd`** uses the same path as direct loads.
> 
> Public **`api.txt`** and README cover banner placement (sticky for refreshes), usage examples, and the warning not to assign callbacks directly. **`BannerAdFlowTest`** covers load tracking order, safe callback replacement, and failed-load attribution.
> 
[...]

PR #4028, opened 2026-08-19
feat(admob): track next-gen banner response loading

polmiro:
## Motivation

Expose server-to-server Google Mobile Ads Next-Gen banner response loading through the RevenueCat tracking API.

This PR is stacked on #4014, which provides the core banner tracking integration and callback installation.

## Description

- Add callback-based `AdView.loadAndTrackAdFromResponse` and `AdTracker.loadAndTrackBannerAdFromResponse` APIs.
- Add suspending overloads that return the original Google `AdLoadResult` unchanged.
- Require an explicit ad unit ID so successful and failed response loads retain attribution.
- Install lifecycle and refresh tracking before loaded banners reach application code.
- Preserve application callbacks while tracking load success and failure.
- Publish and document the experimental response-loading API.

## Tracking behavior

The supplied placement and ad unit ID are used for response load events and subsequent banner lifecycle events. RevenueCat tracking and callback installation happen before application callbacks or suspending results are returned.

---

> [!NOTE]
> **Low Risk**
> Additive experimental API on top of existing banner tracking patterns, with focused unit tests and no changes to core purchase or auth flows.
> 
> **Overview**
> Adds **server-to-server banner loading** with RevenueCat tracking for Google Mobile Ads Next-Gen, matching the existing `FromResponse` APIs on other formats.
> 
> New public entry points are `AdView.loadAndTrackAdFromResponse` and `AdTracker.loadAndTrackBannerAdFromResponse`, each with callback and suspending overloads. Callers must pass an explicit **`adUnitId`** (and optional placement) so load success and failure events keep correct attribution when the opaque response does not expose it.
> 
> Implementation wires `AdView.loadFromAdResponse` through the same tracking path as request-based banner loads: `TrackingAdLoadCallback` for callbacks, `trackAndConfigureAdLoadResult` for suspend, with lifecycle and refresh wrappers installed before app code sees the loaded `BannerAd`. Suspending APIs return Google’s `AdLoadResult` unchanged.
> 
> README and `api.txt` document the new surface; `BannerAdResponseFlowTest` covers callback ordering, failure attribution, and suspending success/failure behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a172fc56330de5378f5fc7fe7d12139d259f405. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


feature/admob-next-gen/src/main/kotlin/com/revenuecat/purchases/admob/nextgen/AdTrackerAdMobExtensions.kt
```diff
@@ -44,3 +45,66 @@ public fun AdTracker.loadAndTrackBannerAd(
...
+ */
+@ExperimentalPreviewRevenueCatPurchasesAPI
+@JvmSynthetic
+@Suppress("LongParameterList")
+public fun AdTracker.loadAndTrackBannerAdFromResponse(
+    adView: AdView,
+    adResponse: String,
+    adUnitId: String,
+    placement: String? = null,
+    loadCallback: AdLoadCallback<BannerAd>,
```
AlvaroBrey, line 530: Minor: `loadCallback` is mandatory in this PR's signatures but optional in [the parent PR](https://github.com/RevenueCat/purchases-android/pull/4014/changes). Unless there's  something I'm missing, they should be nullable here too?

AlvaroBrey (approved):

feature/admob-next-gen/src/main/kotlin/com/revenuecat/purchases/admob/nextgen/AdTrackerAdMobExtensions.kt
polmiro: The response API has a same-named suspending overload, so loadCallback is marked required to distinguish them. The regular loader has no suspending overload, so loadCallback can remain optional without requiring callers to pass loadCallback = null. I think the result fits the ergonomics for each case well. Let me know what do you think.

AlvaroBrey: Hmm that makes sense, hadn't noticed.

Not gonna block on this, but the typical SDK language for suspend variants is `awaitXYZ` when the non-suspend is `XYZ` ( `awaitOfferings`, `awaitPurchase`, etc).  I'd argue to switch to it for three reasons:

- Same name overrides gated by a single param are brittle in that if e.g. we added a default value to `loadCallback`, then a caller with no loadCallback would always be the suspend
- Just consistency with the rest of the SDK
- Would allow nullability of `loadCallback` for fire-and-forget usecases

polmiro: Sounds good. I'll do it in a follow up PR and against this and the native case as well. Thanks for the heads up

PR #4041, opened 2026-08-20
Don't jitter developer-initiated subscriber attribute posts in foreground

tonidero:
Posting subscriber attributes always waited out a random 0-5s jitter before the request left the device. That jitter exists to spread out syncs that many apps perform at the same lifecycle moment, but it was hardcoded in the poster, so it also applied to syncs a developer explicitly asked for while the app is foregrounded.

This PR fixes that so we know only add the jittering if the app is foregrounded.

---

> [!NOTE]
> **Low Risk**
> Timing-only change for network enqueue; foreground attribute syncs become faster without altering auth or purchase logic.
> 
> **Overview**
> Subscriber attribute POSTs no longer always wait on the fixed **0–5s** dispatcher jitter. **`Delay.jitterOnlyIfInBackground`** centralizes the rule: **background** → `DEFAULT` jitter, **foreground** → `Delay.NONE`.
> 
> **`synchronizeSubscriberAttributesForAllUsers`** and **`SubscriberAttributesPoster`** now take an explicit **`delay`** and pass it to the backend instead of hardcoding jitter in the poster. Call sites pick delay from app state: e.g. **`syncAttributesAndOfferingsIfNeeded`** and login/logout in **`IdentityManager`** use background-aware jitter; lifecycle sync uses **`NONE`** on background and **`DEFAULT`** on foreground.
> 
> **`Backend`** callback enqueue paths that previously inlined the same if/else now call **`jitterOnlyIfInBackground`** for consistency.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47f9d89b1a6e8751b313ed1a6f8db027b341fa1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt
```diff
@@ -24,6 +24,13 @@ internal enum class Delay(val minDelay: Duration, val maxDelay: Duration) {
     NONE(0.milliseconds, 0.milliseconds),
     DEFAULT(0.milliseconds, DispatcherConstants.jitterDelay),
     LONG(DispatcherConstants.jitterDelay, DispatcherConstants.jitterLongDelay),
+    ;
+
+    companion object {
+        // Backgrounded apps tend to wake up together, so we spread their requests out. A foregrounded app is
+        // reacting to something the user or the developer just did, so it should not be held back.
+        fun jitterOnlyIfInBackground(appInBackground: Boolean): Delay = if (appInBackground) DEFAULT else NONE
+    }
```
ajpallares, line 33: Nice cleanup 👌

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -408,7 +408,7 @@ internal class PurchasesOrchestrator(
             }
             offeringsManager.onAppForeground(identityManager.currentAppUserID)
             postPendingTransactionsHelper.syncPendingPurchaseQueue(allowSharingPlayStoreAccount)
-            synchronizeSubscriberAttributesIfNeeded()
+            synchronizeSubscriberAttributesIfNeeded(Delay.DEFAULT)
```
tonidero, line 411: I debated a bit with myself on whether to use `Delay.DEFAULT` and `Delay.NONE`... on app foreground shouldn't really be a problem to use `Delay.NONE`, but still, it's not user initiated so I thought this was ok.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
```diff
@@ -360,7 +360,7 @@ internal class PurchasesOrchestrator(
         )
         appConfig.isAppBackgrounded = true
         if (!appConfig.uiPreviewMode) {
-            synchronizeSubscriberAttributesIfNeeded()
+            synchronizeSubscriberAttributesIfNeeded(Delay.NONE)
```
tonidero, line 363: On app background, decided to have no delay, so we can avoid issues of the attributes not being synced. Also checked that this won't interfere with the events flushing right after this, since those run on a different dispatcher, it shouldn't be enqueued.

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
vegaro: This will be triggered on notifications for example right? if so I think we should not do `NONE` here and this is fine

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: Oh, you just answered to the question I had 😅 
And it's a good decision. It should reduce the possibility of attributes not being synced 👍

purchases/src/main/kotlin/com/revenuecat/purchases/common/Dispatcher.kt
```diff
@@ -24,6 +24,13 @@ internal enum class Delay(val minDelay: Duration, val maxDelay: Duration) {
     NONE(0.milliseconds, 0.milliseconds),
     DEFAULT(0.milliseconds, DispatcherConstants.jitterDelay),
     LONG(DispatcherConstants.jitterDelay, DispatcherConstants.jitterLongDelay),
+    ;
+
+    companion object {
+        // Backgrounded apps tend to wake up together, so we spread their requests out. A foregrounded app is
+        // reacting to something the user or the developer just did, so it should not be held back.
+        fun jitterOnlyIfInBackground(appInBackground: Boolean): Delay = if (appInBackground) DEFAULT else NONE
```
vegaro, line 32: This is very nice touch

vegaro (approved):

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
ajpallares: > This will be triggered on notifications for example right?

I don't think so? Only if people opens the notification I'd say. So `NONE` would be ok in that regard. The question would be whether there's some rush to sync them. And arguably it can be benefitial to do `NONE` and make sure attributes are synced as quickly as possible...

tonidero: not really. For this to be called, the app needs to be foregrounded. So it's not enough with just waking the app with a notification. So yeah, a bit more open I would say... Still, felt like not urgent, and I didn't want to overload the app foreground process too much if possible so basically I left it as it was.

ajpallares (approved):
Looks great! Thanks for fixing!

purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesOrchestrator.kt
tonidero: > The question would be whether there's some rush to sync them. And arguably it can be benefitial to do NONE and make sure attributes are synced as quickly as possible...

Yeah, my counter argument is to avoid having too many requests immediately after onappforeground, or at least prioritizing the ones that are more important. In any case, for the majority of cases, these should be synced when backgrounding I would say.

PR #4047, opened 2026-08-20
Add Scope plumbing and custom operator extension point

ajpallares:
### Checklist

### Motivation
Port of RevenueCat/purchases-ios#7433, the first of the rules-engine custom operator PRs. Custom `rc.*` operators need somewhere to live, and `rc.rootVar` needs the predicate's root data to survive iteration, which today is lost as soon as `some` / `map` / `reduce` rebind the scope to the current item.

### Description
- Thread a `Scope` (current data plus the never-replaced root) through the evaluator instead of a bare `Value`, and delegate unknown operators to an empty `CustomOperators` dispatcher.
- Behavior-preserving: `var` still sees only the current item inside iteration, no fixtures were added or changed, and the pinned fixture count stays at 382.

### Notes
The `@Suppress("UnusedParameter")` on the empty dispatcher is temporary — `args` and `vars` become used as soon as the first `rc.*` operator lands.

---

> [!NOTE]
> **Medium Risk**
> Touches the rules-engine evaluator and every operator signature, so a scope mix-up could change predicate results. Intended to be behavior-preserving for existing var/iteration semantics.
> 
> **Overview**
> Prepares the rules engine for RevenueCat `rc.*` operators by replacing a bare data `Value` with a `Scope` that keeps both **current** data and the original **root**.
> 
> Iteration (`some`/`all`/`none`/`map`/`filter`/`reduce`) now rebinds only `current` via `scoped()`, so `var` still sees the item (or reduce’s `{current, accumulator}`) while custom ops can later read top-level data. Unknown operators fall through to an empty `CustomOperators` dispatcher (still `UnsupportedOperator`). `firstArgEvaluated` moves onto `Operators`.
> 
> No new operator behavior; tests wrap existing calls in `Scope`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa1fa3c1a99686840262c89a70c0ddc5dccd79e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/rules/Evaluator.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/rules/Evaluator.kt
@@ -33,7 +33,7 @@ internal object Evaluator {
         predicate: Value,
         variables: Map<String, Value>,
     ): Boolean {
-        val scope = Value.ObjectValue(variables)
+        val scope = Scope(root = Value.ObjectValue(variables))
         return evaluateValue(predicate, scope).isTruthy
     }
 
@@ -43,7 +43,7 @@ internal object Evaluator {
      */
     internal fun evaluateValue(
         predicate: Value,
-        vars: Value,
+        vars: Scope,
     ): Value = when (predicate) {
         Value.Null,
         Value.Undefined,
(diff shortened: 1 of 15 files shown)


tonidero (approved):
Nice!

PR #4048, opened 2026-08-20
Fix(Paywalls): Avoid POSIX localizations

JZDesign:
### Checklist

### Motivation

Certain android devices seem to present a POSIX locale and it results in unexpected string output to the user.

For example, the expectation would be "$1.99" but the user sees "$ 1.99"
[linear issue PW-1345](https://linear.app/revenuecat/issue/PW-1345/normalize-android-generated-price-locale-selection)
### Description

Filters out posix if possible

---

> [!NOTE]
> **Low Risk**
> Localized paywall currency formatting only; behavior is narrowed to avoid POSIX locales with no auth or payment impact.
> 
> **Overview**
> Fixes incorrect paywall price strings on some Android devices (e.g. **$ 1.99** instead of **$1.99**) by changing how the paywall picks a **currency locale** for the storefront country.
> 
> Storefront-country locale lookup is moved into **`getAvailableStorefrontCountryLocalesByLanguage`**, which **drops POSIX variants** when building the per-language map so currency formatting matches store-style spacing. Non-POSIX regional variants (e.g. Valencian) keep the same selection behavior.
> 
> Unit tests cover POSIX exclusion and non-POSIX variant preservation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90d109a54c62910a3cf5ff02b3af23b1854ec718. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/PaywallState.kt
```diff
@@ -462,6 +452,31 @@ internal sealed interface PaywallState {
...
+                val isStorefrontCountry = availableLocale.country.equals(
+                    storefrontCountryCode,
+                    ignoreCase = true,
+                )
+                val isPosix = availableLocale.variant.equals("POSIX", ignoreCase = true)
+                if (isStorefrontCountry && !isPosix) {
+                    put(availableLocale.language.lowercase(), availableLocale)
+                }
+            }
+        }
```
AlvaroBrey:
```suggestion
        availableLocales
            .filter { locale ->
                locale.country.equals(storefrontCountryCode, ignoreCase = true) &&
                    !locale.variant.equals("POSIX", ignoreCase = true)
            }
            .associateBy { it.language.lowercase() }
```

suggestion: I know this was preexisting code but still, thought it was worth this change for readability

ui/revenuecatui/src/test/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/CurrencyLocaleForStorefrontTests.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun `preserves existing selection behavior for non-POSIX variants`() {
+        val valencian = Locale.forLanguageTag("ca-ES-VALENCIA")
+
+        val result = getAvailableStorefrontCountryLocalesByLanguage(
+            storefrontCountryCode = "ES",
+            availableLocales = arrayOf(Locale.forLanguageTag("ca-ES"), valencian),
+        )["ca"]
+
+        assertThat(result).isEqualTo(valencian)
+    }
```
AlvaroBrey, line 43: This is a bit weird, because it reads like "most specific variant wins" when really the implementation is "the last variant on the input wins", as we clobber the same key over and over for each country. That's kind of an odd behaviour we have, out of the scope for this PR tho

AlvaroBrey (approved):
Fix is good, just one non-blocking suggestion

PR #4050, opened 2026-08-21
Add rc.length operator for strings and arrays

ajpallares:
### Checklist

### Motivation
Port of RevenueCat/purchases-ios#7437. String and array length are only reachable today through awkward `substr` probes or reduce counters, neither of which supports readable rules or arithmetic on the length.

### Description
- Add `rc.length`: UTF-16 code units for strings (JS `String.length` parity, including emoji), top-level element count for arrays.
- Anything else, including the null from a missing `var`, throws a type mismatch rather than returning 0, so an absent value can't match a length check by accident.
- Fourteen fixtures copied byte-for-byte from iOS (396 total).

### Notes
Stacked on #4047. The fixture JSON is the shared cross-platform spec, so it is intentionally identical to the iOS file rather than reformatted.

---

> [!NOTE]
> **Low Risk**
> Additive rules-engine operator with strict type errors and shared fixtures; existing predicates are unchanged unless they start using `rc.length`.
> 
> **Overview**
> Adds **`rc.length`** so targeting rules can get a numeric length for strings and arrays (and use it in comparisons/arithmetic).
> 
> Strings use **UTF-16 code units** (JS `String.length` / emoji as 2). Arrays use top-level element count. Other types, including a missing `var` resolving to null, throw `TypeMismatch` instead of returning `0`, so absent data cannot accidentally match a length check.
> 
> Literal arrays must be wrapped (`{"rc.length": [[1,2,3]]}`). Extra args follow json-logic-js first-arg spread. Covered by 15 shared predicate fixtures (suite 416 → 431).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 74a93324deb8496b827c6ceb178ee841809741af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/testDefaults/resources/predicate-fixtures/rc_length.json
```diff
@@ -0,0 +1,217 @@
...
+          },
+          2
+        ]
+      },
+      "variables": {},
+      "expected": true
+    },
+    {
+      "id": "length_empty_array_is_zero",
+      "description": "Literal array operands must be wrapped: {\"rc.length\": [[]]}",
```
tonidero, line 45: Could be worth having another test where it’s not wrapped? 
```suggestion
      "description": "Literal array operands must be wrapped: {\"rc.length\": []}",
```

ajpallares: Good suggestion! Added in https://github.com/RevenueCat/purchases-android/commit/9a89bf01ac071fc1b8024d9d57696e4e43156237
Thanks!

tonidero (approved):

PR #4051, opened 2026-08-21
Add rc.lower and rc.upper custom operators

ajpallares:
### Checklist

### Motivation
Port of RevenueCat/purchases-ios#7436. Case-insensitive audience matching is impossible today — a rule comparing a country code or substring regardless of case either fails or becomes silently case-sensitive.

### Description
- Add `rc.lower` and `rc.upper`: locale-independent Unicode case mapping via the parameterless `lowercase()` / `uppercase()`, matching JS `toLowerCase` / `toUpperCase` rather than the locale-sensitive variants.
- Non-string input throws a type mismatch instead of being coerced, so a missing `var` can't compare equal to `"null"`. Consistent with `rc.length`.
- Eighteen fixtures copied byte-for-byte from iOS (400 total), pinning the Turkish *I* mapping in both directions.

### Notes
Stacked on #4047. The `jsNumberString` doc correction is doc-only, with Kotlin's own formatting examples since JVM and Swift diverge from JS differently.

---

> [!NOTE]
> **Medium Risk**
> Touches the rules engine used for audience matching: new operators change which predicates evaluate, and type-mismatch-on-non-string is a deliberate JS divergence.
> 
> **Overview**
> Adds **`rc.lower`** and **`rc.upper`** so JSON Logic rules can do case-insensitive matching (e.g. country codes, substring `in`) without depending on device locale.
> 
> Conversion uses parameterless Kotlin `lowercase()`/`uppercase()` (Unicode default mapping, matching JS `toLowerCase`/`toUpperCase`, not locale variants). Non-strings throw `TypeMismatch` instead of coercing, so a missing `var` cannot silently equal `"null"`. Extra args follow json-logic-js first-arg-only spread.
> 
> Also expands `jsNumberString` docs with Kotlin vs JS scientific-notation examples. Conformance fixtures cover ASCII, Turkish I, non-ASCII, type errors, and motivating equality/`in` cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88a80bf9cef8097933c664ecfd4a2387dcd4b929. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


purchases/src/testDefaults/resources/predicate-fixtures/rc_case.json
```diff
@@ -0,0 +1,262 @@
...
+      "expected": true
+    },
+    {
+      "id": "lower_case_insensitive_country_code_equality",
+      "description": "Motivating case: compare a variable country code case-insensitively against a lowercase literal",
+      "predicate": {
+        "===": [
+          {
+            "rc.lower": {
+              "var": "country"
```
tonidero, line 230: I’m pretty sure it works, but maybe having an additional test with a nested variable? Something like 
```suggestion
              "var": "country.code"
```

tonidero (approved):

purchases/src/testDefaults/resources/predicate-fixtures/rc_case.json
ajpallares: Sure! Done in https://github.com/RevenueCat/purchases-android/pull/4051/commits/806b95a41b1c962c944675a8ee3669494e88124d

PR #4054, opened 2026-08-21
Key the persisted remote config state by domain

tonidero:
### Description

Part 1 of 2 of subdomain support for the `/v1/config` endpoint. A config response can declare `subdomains`: other domains the SDK should also sync to assemble the full configuration, so the backend can move a topic to a subdomain without an SDK update. We already parse the field but drop it at commit. **Scope note: only one level of subdomains is supported** — a subdomain's own `subdomains` list is ignored.

This PR reshapes the persisted state so it can hold a domain tree, with **no behavior change** for the single-domain trees the backend serves today:

- `PersistedRemoteConfigurationState` becomes `{ rootDomain, domains: Map<domain, PersistedDomainState> }`, still one atomic file write. Each domain entry carries its own `manifest`/`subdomains`/`activeTopics`/`prefetchBlobs`/`topics`/`lastRefreshTime`.
- Lazy derived views on the state: `domainsInPrecedenceOrder` (the root followed by its subdomains in declaration order, filtered to persisted entries), `mergedTopics` (parent-wins union across those domains — what `RemoteConfigTopicStore` now serves), and `liveBlobRefs` (union over **all** persisted domains).
- `blobStore.retainOnly` now takes `liveBlobRefs`, so one domain's sync can never evict another domain's blobs (identical set with a single domain, but the critical groundwork).
- Requests replay only the requested domain's own manifest / last-refresh-time / `prefetched_blobs`.
- A legacy single-domain file lifts into the new shape on read, preserving the manifest so the next sync after an SDK update stays an incremental diff instead of a cold re-fetch.

Follow-up: PR 2 (#4056) adds the sync engine (fetch root → sync its subdomains → commit children first, root last, with deletion deferral).

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt
@@ -222,7 +222,7 @@ internal class PurchasesFactory(
             val remoteConfigEnabled = !appConfig.customEntitlementComputation
             val remoteConfigDiskCache = if (remoteConfigEnabled) RemoteConfigDiskCache(contextForStorage) else null
             val remoteConfigTopicStore = RemoteConfigTopicStore {
-                remoteConfigDiskCache?.read()?.topics?.get(it.wireName)
+                remoteConfigDiskCache?.read()?.mergedTopics?.get(it.wireName)
             }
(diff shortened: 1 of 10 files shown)


tonidero:
We changed audiences topic to be a single blob so we don't need this for now. We will bring it over if/when we think the config endpoint may grow but not a priority right now

PR #4056, opened 2026-08-21
Sync the remote config subdomain tree

tonidero:
### Description

Part 2 of 2 of subdomain support for `/v1/config` (stacked on #4054). This adds the sync engine. It is inert against the current backend, which declares no `subdomains` yet: a single-domain tree behaves exactly as before.

A refresh is now one pass over the domain tree:

- **Fetch root → sync its subdomains → commit children first, root last.** Only one level is supported: duplicates and self-references in the root's list are dropped, and a subdomain's own list is ignored (logged). Each subdomain syncs with its own manifest/refresh-time/`prefetched_blobs`. `RemoteConfigDomainFetcher` is the new suspend bridge over the callback-based backend endpoints; a root 4xx still applies the session kill-switch at callback time even if the pass was cancelled.
- **Deletion deferral (the migration-no-gap guarantee).** The root's 200 commits only when every subdomain it lists ended the pass fresh. Otherwise the root keeps its old state and old manifest, so the next pass re-diffs cheaply and committed subdomains answer 204. Combined with parent-wins merging, a topic migrating from the root to a subdomain never disappears from the merged view at any read interleaving (there's a test asserting the intermediate state).
- **Pass-wide finalize.** Per-domain 204 refresh times land in one batched write; `blobStore.retainOnly` runs once per pass with the cross-domain union; the generation bumps once per pass so listeners only re-warm from tree-consistent states; staleness bookkeeping advances only when the pass ended fresh, so a partial pass retries on the next trigger.
- **Failure rules.** Any subdomain failure — retryable or 4xx — defers the root's commit; a subdomain 4xx does *not* trip the whole-system kill-switch and self-heals once the server stops listing the refused domain. The cold-start fallback stays root-only (the fallback response's subdomains list is persisted, and the next regular pass completes the tree once the main API recovers).

Known limitations (deliberate, to keep the change small): a 4xx-ing subdomain is refetched once per pass (bounded by the staleness gate and the 1-minute attempt cooldown) rather than session-disabled, and a subdomain dropped from the root's list keeps its inert persisted entry (invisible to reads, blobs pinned) until an identity change wipes the cache.

### DIFF
--- a/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDomainFetcher.kt
+++ b/purchases/src/main/kotlin/com/revenuecat/purchases/common/remoteconfig/RemoteConfigDomainFetcher.kt
@@ -1,0 +1,119 @@
+package com.revenuecat.purchases.common.remoteconfig
+
+import com.revenuecat.purchases.PurchasesError
+import com.revenuecat.purchases.common.Backend
+import com.revenuecat.purchases.common.GetRemoteConfigErrorHandlingBehavior
+import com.revenuecat.purchases.common.LogIntent
+import com.revenuecat.purchases.common.errorLog
+import com.revenuecat.purchases.common.log
+import com.revenuecat.purchases.common.networking.HTTPResult
+import com.revenuecat.purchases.common.networking.RCContainer
+import com.revenuecat.purchases.common.verboseLog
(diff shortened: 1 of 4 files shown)


tonidero:
We changed audiences topic to be a single blob so we don't need this for now. We will bring it over if/when we think the config endpoint may grow but not a priority right now

